PR #576, opened 2019-02-27
(testpart2) Remove Security Vulnerabilities

DarylThayil:
Upgrade versions and remove extraneous dependencies to remove security vulnerabilties.

The only remaining are as follows
 
1 critical - growl  - removed in grunt pr #569 
1 high - handlebars - removed in grunt pr #569 
1 low - braces - no plan to remove yet, In karma, can remove if we switch to mocha


DarylThayil:
Also pin travis to use 10 as its environment as this is the highest node we support

PR #578, opened 2019-02-28
API Changes: Error Object

pkanher617:
This PR is for the addition of the custom error object to the library. This will enable us to throw detailed errors with stack trace in the cases of error. This PR does not include integration with the UserAgentApplication file - this will come in a separate PR.


lib/msal-core/src/AuthError.ts
```diff
@@ -0,0 +1,240 @@
...
+/**
+* @hidden
+*
+* General error class thrown by the MSAL.js library.
+*/
+export class AuthError extends Error {
+    error: string;
+    errorDesc: string;
+
+    constructor(error: string, detailMessage: string) {
```
DarylThayil: error is a little generic and non descript.
What do you think of `errorCode`?

DarylThayil: if the class has `errorDesc` why not match instead of detailed message

lib/msal-core/src/AuthError.ts
```diff
@@ -0,0 +1,240 @@
...
+import { ErrorMessage } from "./ErrorMessage";
+import { Utils } from "./Utils";
+/**
+* @hidden
+*
+* General error class thrown by the MSAL.js library.
+*/
+export class AuthError extends Error {
+    error: string;
+    errorDesc: string;
```
DarylThayil: how about `errorMessage`?

lib/msal-core/src/AuthError.ts
```diff
@@ -0,0 +1,240 @@
...
+
+    constructor(error: string, detailMessage: string) {
+        super(detailMessage);
+        this.error = error;
+        this.errorDesc = detailMessage;
+        this.name = "AuthError";
+        this.stack = new Error().stack;
```
DarylThayil: will this bring the stack from the caller? like if we throw the error in UserAgentApplication, the stack root is there?

lib/msal-core/src/AuthError.ts
```diff
@@ -0,0 +1,240 @@
...
+
+        Object.setPrototypeOf(this, AuthError.prototype);
```
DarylThayil: what does this do ?

DarylThayil: like what does in accomplish in usage?

lib/msal-core/src/AuthError.ts
pkanher617: error is the name of the var coming from the server, so we decided to keep it uniform

lib/msal-core/src/AuthError.ts
```diff
@@ -21,220 +21,220 @@
...
+        return new AuthError(ErrorMessage.unexpectedError.code, errDesc);
+    }
+}
+
+/**
+ * @hidden
+ *
+ * Error thrown when there is an error in the client code running on the browser.
+ */
+export class ClientAuthError extends AuthError {
```
DarylThayil: Any Generic Client Auth Error needed?

lib/msal-core/src/AuthError.ts
pkanher617: Noted, should be the same. Adding this in a commit.

lib/msal-core/src/AuthError.ts
pkanher617: errorDesc is the name of the var that comes from the server, so we decided to keep it uniform

lib/msal-core/src/AuthError.ts
```diff
@@ -21,220 +21,220 @@
...
+    constructor(error: string, errorDesc: string) {
+        super(error, errorDesc);
+        this.name = "ClientAuthError";
+
+        Object.setPrototypeOf(this, ClientAuthError.prototype);
+    }
+
+    static createEndpointResolutionError(errDesc: string): ClientAuthError {
+        var errorMessage = "Error: could not resolve endpoints. Please check network and try again.";
```
DarylThayil: why is this not in the config file for errors

lib/msal-core/src/AuthError.ts
pkanher617: Yes. It is essentially mimicking the stack that the JS built-in Error object creates. We have to do this because inheritance for built-ins is a little bit more convoluted in Typescript 2.2+

lib/msal-core/src/AuthError.ts
```diff
@@ -21,220 +21,220 @@
...
+        if (!Utils.isEmpty(errDesc)) {
```
DarylThayil: could write a util for appendErrorDetailsToDescription?

DarylThayil: nitpick unimportant in long run

lib/msal-core/src/AuthError.ts
DarylThayil: seems like some are in config some are not

lib/msal-core/src/AuthError.ts
pkanher617: This enabled us to extend the JS built-in Error object. We have to do this because inheritance for built-ins is a little bit more convoluted in Typescript 2.2+. https://github.com/Microsoft/TypeScript-wiki/blob/master/Breaking-Changes.md#extending-built-ins-like-error-array-and-map-may-no-longer-work

lib/msal-core/src/AuthError.ts
pkanher617: something like appendErrorDetailsToDescription(errDesc, details)? I wasn't sure it was needed but can be added pretty easily

lib/msal-core/src/AuthError.ts
pkanher617: This should probably be added to the ErrorMessage.ts file.

lib/msal-core/src/AuthError.ts
pkanher617: Not sure what the ask here is...?

lib/msal-core/src/AuthError.ts
```diff
@@ -0,0 +1,240 @@
...
+            ErrorMessage.loginProgressError.desc);
+    }
+
+    static createAcquireTokenInProgressError(): ClientAuthError {
+        return new ClientAuthError(ErrorMessage.acquireTokenProgressError.code,
+            ErrorMessage.acquireTokenProgressError.desc);
+    }
+
+    static createUserCancelledError(): ClientAuthError {
+        return new ClientAuthError(ErrorMessage.userCancelledError.code,
```
DarylThayil: I do appreciate the config of the errors quite a bit, makes it easy to change and reference, but I wonder if there is a way to get that but also get readability of errors in code.

When I am getting a SOMETHING_ERROR it is nice to be able to easily trace that and see oh here is where that error is being created, instead of ending up in a config file and looking around.

its more of a dev nicety, and I am not sure I have a great answer for an alternative off the top of my head, but food for thought

pkanher617: The error.stack should have the trace where it occurred in the code

pkanher617: Is there something additional beyond a verbal description of the error and the stack trace that we should be giving back?

lib/msal-core/src/AuthError.ts
```diff
@@ -0,0 +1,240 @@
...
+        var errorMessage = "Error: could not resolve endpoints. Please check network and try again.";
+        if (!Utils.isEmpty(errDesc)) {
+            errorMessage += " Details: " + errDesc;
+        }
+        return new ClientAuthError(ErrorMessage.endpointResolutionError.code, errorMessage);
+    }
+
+    static createMultipleMatchingTokensInCacheError(scope: string): ClientAuthError {
+        return new ClientAuthError(ErrorMessage.multipleMatchingTokens.code,
+            "Cache error for scope " + scope + ": " + ErrorMessage.multipleMatchingTokens.desc);
```
DarylThayil: in all of this file, instead of using + signs to concat use string interpolation

`Cache error for scope ${scope}: ${ErrorMessage.multiplematchingTokens.desc}`;

lib/msal-core/src/AuthError.ts
```diff
@@ -0,0 +1,240 @@
...
+            ErrorMessage.userCancelledError.desc);
+    }
+}
+
+/**
+ * @hidden
+ *
+ * Error thrown when there is an error in configuration of the .js library.
+ */
+export class ConfigurationError extends ClientAuthError {
```
DarylThayil: Should this extend AuthError not ClientAuth Error?

lib/msal-core/src/AuthError.ts
pkanher617: +1, string interpolation for the rest of the code should be coming in a separate PR.

lib/msal-core/src/AuthError.ts
```diff
@@ -0,0 +1,240 @@
...
+    //         userState);
+    // }
+
+    // static createSuccessCallbackParametersError(numArgs: number, tokenType: string, userState: string) : ClientAuthError {
+    //     return new ConfigurationAuthError("Error occurred in callback for successful token response - incorrect number of arguments, expected 1, got " + numArgs + ".", tokenType, userState);
+    // }
+
+    // static createErrorCallbackParametersError(numArgs: number, tokenType: string, userState: string) : ClientAuthError {
+    //     return new ConfigurationAuthError("Error occurred in callback for error response - incorrect number of arguments, expected 1, got " + numArgs + ".", tokenType, userState);
+    // }
```
DarylThayil: should delete when we can

lib/msal-core/src/AuthError.ts
pkanher617: It was agreed we would have AuthError as the base, ClientAuthError, ServerError extending this base class, and other error types deriving from these two types.

lib/msal-core/src/AuthError.ts
```diff
@@ -0,0 +1,240 @@
...
+            errorDesc);
+    }
+}
+
+/**
+ * @hidden
+ *
+ * Error thrown when the user is required to perform an interactive token request.
+ */
+export class InteractionRequiredAuthError extends ServerError {
```
DarylThayil: What does extending server error get us ? There is no chain right?

lib/msal-core/src/AuthError.ts
```diff
@@ -0,0 +1,240 @@
+/**
+  * Copyright (c) Microsoft Corporation
```
DarylThayil: 1 class per file please 
So much easier to read 
lets add a src/error folder 
index.ts // Exports all the classes so they can be imported like this
```js
import { AuthError, InteractionError } from './errors';
```

then have files fore
AuthError.ts
ClientInteractionError.ts
ServerError.ts
 etc...

lib/msal-core/src/ErrorMessage.ts
```diff
@@ -21,13 +21,127 @@
...
+  unexpectedError: {
+    code: "unexpected_error",
+    desc: "unexpected error in authentication"
+  }
+
+  // static get authorityUriInvalidPath(): string { return "AuthorityUriInvalidPath"; }
+  // static get authorityUriInsecure(): string { return "AuthorityUriInsecure"; }
+  // static get invalidAuthorityType(): string { return "InvalidAuthorityType"; }
+  // static get unsupportedAuthorityValidation(): string { return "UnsupportedAuthorityValidation"; }
+  // static get b2cAuthorityUriInvalidPath(): string { return "B2cAuthorityUriInvalidPath"; }
```
DarylThayil: remove comments ?

pkanher617: +1

lib/msal-core/src/ErrorMessage.ts
```diff
@@ -21,13 +21,127 @@
...
+    code: "server_unavailable",
+    desc: "Server is temporarily unavailable"
+  },
+  unknownServerError: {
+    code: "unknown_server_error",
+    desc: "Unknown server error"
+  },
+
+  // Interaction Required Errors
+  loginRequired: "login_required",
```
DarylThayil: everything in here is an object, why aren't these

lib/msal-core/src/AuthError.ts
pkanher617: We do have a prototype chain. That is the entire point of setPrototypeOf().

lib/msal-core/src/AuthError.ts
pkanher617: +1

lib/msal-core/src/ErrorMessage.ts
pkanher617: The detailed errors are sent from the server, just need these as constants to be able to detect and throw the correct error type.

lib/msal-core/src/ErrorMessage.ts
```diff
@@ -21,13 +21,127 @@
...
- * @hidden
- */
-export class ErrorMessage {
-  static get authorityUriInvalidPath(): string { return "AuthorityUriInvalidPath"; }
-  static get authorityUriInsecure(): string { return "AuthorityUriInsecure"; }
-  static get invalidAuthorityType(): string { return "InvalidAuthorityType"; }
-  static get unsupportedAuthorityValidation(): string { return "UnsupportedAuthorityValidation"; }
-  static get b2cAuthorityUriInvalidPath(): string { return "B2cAuthorityUriInvalidPath"; }
-}
+export const ErrorMessage = {
```
DarylThayil: If we break up every error class, i actually like the idea better of putting these config in the individual files 

so `ClientInteractionError.ts` would have the errors for it, and its still defined in a const at the top of the file, but its also really easy to see the definition

lib/msal-core/src/AuthError.ts
pkanher617: Ok I will update and separate these files.

lib/msal-core/src/AuthError.ts
DarylThayil: code and message are a little more descriptive though.  error indicates error object, code indicates a string const representation of an error, and message indicates the details of that given code

lib/msal-core/src/ErrorMessage.ts
pkanher617: I think we wanted to consolidate strings to a single file, bu this could work too. I'd like the rest of the team's feedback as well.

lib/msal-core/src/AuthError.ts
DarylThayil: +1 for interesting weird stuff ... haha it's kinda gross but not your fault

lib/msal-core/src/AuthError.ts
DarylThayil: I think my future comments have a better idea

DarylThayil: interpolation!

lib/msal-core/src/AuthError.ts
DarylThayil: I do remember that but there is no benefit out of that right? We don't represent in an way to developer.

You dont get an error with any indication that a config error is a client error right ?

lib/msal-core/src/ErrorMessage.ts
DarylThayil: I think for consistency its better to have an object with a code but not message

lib/msal-core/src/AuthError.ts
pkanher617: @manoj-rath @sameerag Looking at other libraries, here are their error object signatures:
Android: errorCode, errorMessage
iOS: code, desc
C#: StatusCode, ExceptionDetail

I am going to use errorCode and errorMessage similar to Android since it seems like there isn't much uniformity across MSAL currently.

lib/msal-core/src/AuthError.ts
pkanher617: ConfigError instanceof ClientAuthError = false
ClientAuthError instanceof ConfigError = true

usage:

try {
msal.loginRedirect(...);
} catch (err) {
    if (err instanceof ClientAuthError) {
         if(err instanceof ConfigError) {

         }
    }
}

The benefit is that errors are classified as coming from the client or server before being classified as configuration or client library errors. Happy to discuss changing these, but I will keep it as is for now.

lib/msal-core/src/ErrorMessage.ts
DarylThayil: Yeah i think having one file has its benefits but I really like the idea of small digestable files 

FakeErrorClass.ts
```js
import { AuthError } from './AuthError';

const  fakeErrors = {
   errorone: {
          code: 'error_one',
          message: 'I want to update error one and its here with the error class!'
   },
  errortwo:  {
          code: 'error_one',
          message: 'And since there are only two errors here its super easy to read and digest these errors'
   },
};

export default class FakeErrorClass extends AuthError { 
   constructor(cool) {
      super(cool)
   }

   static createErrorOne(message: string) {
            ....
   } 

   static createErrorTwo(message: string) {
            ....
   } 
}
```

lib/msal-core/src/ErrorMessage.ts
pkanher617: ok I can do that.

lib/msal-core/src/AuthError.ts
DarylThayil: +1

lib/msal-core/src/AuthError.ts
DarylThayil: never mind !

lib/msal-core/src/AuthError.ts
DarylThayil: while you have error here why not just update this file use interpolation?

lib/msal-core/src/AuthError.ts
DarylThayil: that makes sense, I am fine with that.

lib/msal-core/src/AuthError.ts
pkanher617: I think @sameerag has these updates in a separate PR.

lib/msal-core/src/AuthError.ts
DarylThayil: I mean a chain in terms of telling errors parents from eachother, but you answered my question above for the value. Marking resolved

lib/msal-core/src/AuthError.ts
pkanher617: Will change this to display errorCode and errorMessage in line with Android.

DarylThayil:
Notes from offline
[ ] Unit Tests
[ ] Error Classes Not Hidden
[ ] Fix String Interpolation

lib/msal-core/src/Constants.ts
```diff
@@ -84,9 +84,11 @@ export class Constants {
...
   static get unknown(): string { return "UNKNOWN"; }
   static get urlHash(): string { return "msal.urlHash"; }
-    static get angularLoginRequest(): string { return "msal.angular.login.request"; }
-    static get userIdentifier(): string { return "userIdentifier"; }
-    static get common(): string { return "common"; }
+  static get angularLoginRequest(): string { return "msal.angular.login.request"; }
+  static get userIdentifier(): string { return "userIdentifier"; }
+  static get common(): string { return "common"; }
+  static get cacheLocationLocal(): string { return "localStorage"; }
+  static get cacheLocationSession(): string { return "sessionStorage"; }
```
DarylThayil: why are these being added in an error pr ?

lib/msal-core/src/error/AuthError.ts
```diff
@@ -0,0 +1,59 @@
+/**
```
DarylThayil: We should check and see if this is actually necessary in every file. Maybe look at other open source ms projects like Vscode?

lib/msal-core/src/error/AuthError.ts
```diff
@@ -0,0 +1,59 @@
...
+        desc: "unexpected error in authentication"
+    }
+};
+
+/**
+* @hidden
+*
+* General error class thrown by the MSAL.js library.
+*/
+export class AuthError extends Error {
```
DarylThayil: We should export default the class so that this can be true true

import AuthError,  { AuthErrorMessage } from './AuthError'

lib/msal-core/src/error/ClaimsRequiredError.ts
```diff
@@ -21,13 +21,20 @@
...
  */
-export class ErrorMessage {
-  static get authorityUriInvalidPath(): string { return "AuthorityUriInvalidPath"; }
-  static get authorityUriInsecure(): string { return "AuthorityUriInsecure"; }
-  static get invalidAuthorityType(): string { return "InvalidAuthorityType"; }
-  static get unsupportedAuthorityValidation(): string { return "UnsupportedAuthorityValidation"; }
-  static get b2cAuthorityUriInvalidPath(): string { return "B2cAuthorityUriInvalidPath"; }
+export class ClaimsRequiredAuthError extends InteractionRequiredAuthError {
+    claimsRequired: string;
+    constructor(errorCode: string, errorMessage: string, claimsRequired: string) {
```
DarylThayil: do we not have any claims required error?

lib/msal-core/src/error/ClientAuthError.ts
```diff
@@ -0,0 +1,147 @@
...
+  * OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+  * WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT
+  * OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+  */
+
+import { AuthError } from "./AuthError";
+import { Utils } from "../Utils";
+
+export const ClientAuthErrorMessage = {
+
```
DarylThayil: nitpick, unecessary space, when we tighten our linting this will probably get fixed

lib/msal-core/src/error/ClaimsRequiredError.ts
pkanher617: Not currently, coming in a future feature when we implement Conditional Access.

lib/msal-core/src/Constants.ts
pkanher617: Required for the error messages.

lib/msal-core/src/error/ClaimsRequiredError.ts
DarylThayil: Is this a feature we are eventually going to implement? or that you have code ready to review for?
Its such a small amount of code that if we are not using it, I would prefer we just write it when we have a use case. 

If for some un forseen reason feature priorities change or something like that, pushing unused code becomes debt

lib/msal-core/src/error/AuthError.ts
sameerag: Can we track this as a separate follow up somewhere? May be @navyasric, @manoj-rath  can help.

lib/msal-core/src/error/ClaimsRequiredError.ts
sameerag: There is a lot of work that is put in understanding Claims @DarylThayil . We have a use case and it is implemented in most MSALs. We are currently in the middle of untangling some requirements and should be able to implement this soon.

DarylThayil: @sameerag  my feedback was just to avoid comitting code that we aren't using yet.  It's fairly simple to add the 20 lines to create the error class when we need it in code

DarylThayil:
please squash commits before merging

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -35,7 +35,6 @@ import { TokenResponse } from "./RequestInfo";
 import { User } from "./User";
```
manoj-rath: This file is essentially a revert-back, correct?

pkanher617: Yes that is correct. UserAgentApp changes are coming in a future PR.

PR #579, opened 2019-02-28
Add String Interpolation for Cache Keys generated

sameerag:
Throughout UserAgentApplication.ts, we generate 'authorityKey' and 'acquireTokenUserKey' to generate keys in the Cache. The code now uses a utility function for better readability.

- Not functional change
- Added the new functions in Storage.ts
- Unit Tests pass


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -808,17 +806,13 @@ export class UserAgentApplication {
...
 
         // Cache acquireTokenUserKey
-        var acquireTokenUserKey;
-        if (userObject) {
-          acquireTokenUserKey = Constants.acquireTokenUser + Constants.resourceDelimeter + userObject.userIdentifier + Constants.resourceDelimeter + authenticationRequest.state;
-        }
-        else {
-          acquireTokenUserKey = Constants.acquireTokenUser + Constants.resourceDelimeter  + Constants.no_user + Constants.resourceDelimeter + authenticationRequest.state;
-        }
+        const userId = (userObject) ? userObject.userIdentifier : Constants.no_user;
```
DarylThayil: no need to wrap userObect in parens

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1773,17 +1767,13 @@ export class UserAgentApplication {
...
 
     // Cache acquireTokenUserKey
-    var acquireTokenUserKey;
-    if (user) {
-        acquireTokenUserKey = Constants.acquireTokenUser + Constants.resourceDelimeter + user.userIdentifier + Constants.resourceDelimeter + authenticationRequest.state;
-    }
-    else {
-        acquireTokenUserKey = Constants.acquireTokenUser + Constants.resourceDelimeter  + Constants.no_user + Constants.resourceDelimeter + authenticationRequest.state;
-    }
+    const userId = (user) ? user.userIdentifier : Constants.no_user;
```
DarylThayil: same, no need for parens

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1817,18 +1807,14 @@ export class UserAgentApplication {
...
-    // Cacher acquireTokenUserKey
-    var acquireTokenUserKey;
-    if (user) {
-        acquireTokenUserKey = Constants.acquireTokenUser + Constants.resourceDelimeter + user.userIdentifier + Constants.resourceDelimeter + authenticationRequest.state;
-    }
-    else {
-        acquireTokenUserKey = Constants.acquireTokenUser + Constants.resourceDelimeter + Constants.no_user + Constants.resourceDelimeter + authenticationRequest.state;
-    }
+    // Cache acquireTokenUserKey
+    const userId = (user) ? user.userIdentifier : Constants.no_user;
```
DarylThayil: parens

DarylThayil:
I LOVE LOVE LOVE that we are using a util for this instead of concatenating strings everywhere.  
I also love how small and focused this is 
A few small comments but once those are fixed, approved!

lib/msal-core/src/UserAgentApplication.ts
sameerag: True. I guess I overlooked it. 
+1

sameerag:
@manoj-rath  and @pkanher617, once you sign off, will merge it to the main api branch.

sameerag:
Thanks @DarylThayil . Focused changes for easy testing. :)

sameerag:
All, removed the unneeded parenthesis.

lib/msal-core/src/Storage.ts
```diff
@@ -175,4 +175,20 @@ export class Storage {// Singleton
         this.setItemCookie(Constants.loginRequest, "", -1);
         this.setItemCookie(Constants.stateAcquireToken, "", -1);
     }
+
+    /**
+     * Create acquireTokenUserKey to cache user object
+     * acquireTokenUserKey = Constants.acquireTokenUser + Constants.resourceDelimeter + userId + Constants.resourceDelimeter + authenticationRequest.state;
+     */
+    static generateATUserKey(userId: any, state: string) {
+        return `msal.acquireTokenUser|${userId}|${state}`;
```
manoj-rath: any reason we are not using the constants here?

lib/msal-core/src/Storage.ts
```diff
@@ -175,4 +175,20 @@ export class Storage {// Singleton
...
+    }
+
+    /**
+     * Create authorityKey to cache authority
+     * const authorityKey = Constants.authority + Constants.resourceDelimeter + authenticationRequest.state;
+     */
+    static generateAuthKey(state: string) {
+        return `msal.authority|${state}`;
```
manoj-rath: any reason we are not using the constants here?

lib/msal-core/src/Storage.ts
```diff
@@ -175,4 +175,20 @@ export class Storage {// Singleton
         this.setItemCookie(Constants.loginRequest, "", -1);
         this.setItemCookie(Constants.stateAcquireToken, "", -1);
     }
+
+    /**
+     * Create acquireTokenUserKey to cache user object
+     * acquireTokenUserKey = Constants.acquireTokenUser + Constants.resourceDelimeter + userId + Constants.resourceDelimeter + authenticationRequest.state;
+     */
+    static generateATUserKey(userId: any, state: string) {
```
manoj-rath: can we change the method name, so that it's not cryptic? something like generateCacheKeyForAcquireToken / generateAcquireTokenUserKey

sameerag: Sure manoj. Will do.

lib/msal-core/src/Storage.ts
sameerag: Nothing particular. I thought this style is simpler, picked it from some String Interpolation examples. We are moving constants from the central location piece by piece so I added text. I can use constants too in the same utils function if we want to adhere to that standard.

manoj-rath: if the const is specific to a file, then it might make sense to move locally. However if it's used in different files, centralized location is better, correct?

DarylThayil: technically  this is not a const, it is just a string being generated at run time.

DarylThayil: @sameerag  can you add return type to the fn

lib/msal-core/src/Storage.ts
DarylThayil: its a string generated at run

DarylThayil: @sameerag  can you add return type as well

sameerag: Done

lib/msal-core/src/Storage.ts
sameerag: Done

sameerag: So I have created a new constant in Constants.ts file for CacheKeys and used them. There is more clean up needed for Constants but we will take that up after the current activity. I have added an internal tracker for this.

sameerag:
@manoj-rath Added the changes incorporating your feedback. I have also pulled some changes from dev to resolve the travis issue.

sameerag:
Rebased this to dev. @pkanher617 after you sign off, will merge this to dev.

PR #580, opened 2019-03-02
'var' cleanup - replaced with const/let as applicable

sameerag:
- Cleaned up all the source code to use 'const' and 'let' instead of 'var' wherever applicable. 
- In the class Logger, renamed all '_name' to 'name' to stop using '_underscore' before private variables. UserAgentApplication.ts still contains some variables with the same convention, that change will be a part of config API changes.


sameerag:
@DarylThayil You mean separating 'const' and 'let' changes? Just curious.

DarylThayil:
No the let const together are fine, but the removing underscore rename. It’s a small nitpick, it just makes prs super easy to review when they are one type of change

lib/msal-core/src/ErrorMessage.ts
```diff
@@ -21,6 +21,7 @@
   * OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
   */
```
manoj-rath: This will conflict with Prithvi's error PR, so you may want to remove this change.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1896,19 +1898,13 @@ export class UserAgentApplication {
...
   protected saveTokenFromHash(tokenResponse: TokenResponse): void {
+
     this._logger.info("State status:" + tokenResponse.stateMatch + "; Request type:" + tokenResponse.requestType);
     this._cacheStorage.setItem(Constants.msalError, "");
     this._cacheStorage.setItem(Constants.msalErrorDescription, "");
-    var scope: string = "";
-    var authorityKey: string = "";
-    var acquireTokenUserKey: string = "";
-    // if response has "scope" set it, else set the scope to client_id
-    if (tokenResponse.parameters.hasOwnProperty("scope")) {
```
manoj-rath: why did you remove this? did I miss something?

sameerag: Thanks for catching this. I think it is an oversight while merging. Added the code back. Will commit soon.

sameerag: Actually may be not. Sorry, let me check. I think scope was never used. Give me some time, will update.

sameerag: @manoj-rath 'scope' in saveTokenFromHash is never used. Hence removed the un-used code fragment.

manoj-rath: I also couldn't see scope being used in this method, however I know that this logic "if response has "scope" set it, else set the scope to client_id" exists. So I am not sure if the scope being "var" (global scope) is making it work. please double check if it works correctly by removing this code.

sameerag: I checked the code again and here is what is happening: 
saveTokenFromHash() is the function that has the above code which processes the response from the server, this internally calls saveAccessToken() which has the above logic of "if response has "scope" set it, else set the scope to client_id". 

The deleted code is just redundant and never used.
I think it is safe to remove this.

@pkanher617 can you also check and confirm to ensure no mistakes?

lib/msal-core/src/ErrorMessage.ts
sameerag: I am coordinating with him for the merge, there will be conflicts in UserAgentApplication.ts too and we will handle this at the same time.

PR #582, opened 2019-03-04
Api iter error scope validation rework

pkanher617:
This pull request updates the scope validation and filtering actions in our library, combining them into a single function and utilizing the error objects now created.

Please review PR #578 before this one.

The current base branch is api_iter_error_object until the merge has been completed.

### DIFF
--- a/lib/msal-angular/package.json
+++ b/lib/msal-angular/package.json
@@ -74,7 +74,7 @@
     "ts-loader": "^2.0.3",
     "tslib": "^1.7.1",
     "tslint": "^5.8.0",
-    "typedoc": "^0.6.0",
+    "typedoc": "^0.14.2",
     "typescript": "^2.7.2",
     "typings": "^2.1.1",
     "webpack": "3.6.0",
@@ -90,7 +90,7 @@
   },
   "scripts": {
     "clean": "shx rm -rf dist docs",
-    "doc": "typedoc --out ./docs ./src/**/* --gitRevision dev",
+    "doc": "typedoc --out ./docs ./src/ --gitRevision dev",
     "build:modules": "tsc && tsc -m es6 --outDir lib-es6",
     "build": "npm run clean && npm run doc && npm run build:modules && webpack",
     "test": "typings install && karma start karma.conf.js",
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -369,7 +369,6 @@ export class UserAgentApplication {
 
     // Validate and filter scopes (the validate function will throw if validation fails)
     this.validateInputScope(scopes, false);
-    scopes = this.filterScopes(scopes);
 
     // extract ADAL id_token if exists
     var idTokenObject;
@@ -465,7 +464,6 @@ export class UserAgentApplication {
   acquireTokenRedirect(scopes: Array<string>, authority?: string, user?: User, extraQueryParameters?: string): void {
     // Validate and filter scopes (the validate function will throw if validation fails)
     this.validateInputScope(scopes, true);
-    scopes = this.filterScopes(scopes);
 
     // Get the user object if a session exists
     const userObject = user ? user : this.getUser();
@@ -585,7 +583,6 @@ export class UserAgentApplication {
       }
       // Validate and filter scopes (the validate function will throw if validation fails)
       this.validateInputScope(scopes, false);
-      scopes = this.filterScopes(scopes);
 
       // Extract ADAL id_token if it exists
       var idTokenObject;
@@ -718,7 +715,6 @@ export class UserAgentApplication {
     return new Promise<string>((resolve, reject) => {
       // Validate and filter scopes (the validate function will throw if validation fails)
       this.validateInputScope(scopes, true);
-      scopes = this.filterScopes(scopes);
 
       const scope = scopes.join(" ").toLowerCase();
 
@@ -968,7 +964,6 @@ export class UserAgentApplication {
     return new Promise<string>((resolve, reject) => {
       // Validate and filter scopes (the validate function will throw if validation fails)
       this.validateInputScope(scopes, true);
-      scopes = this.filterScopes(scopes);
 
       const scope = scopes.join(" ").toLowerCase();
       const userObject = user ? user : this.getUser();
@@ -2132,7 +2127,6 @@ export class UserAgentApplication {
    * @ignore
    * @hidden
    */
-  // TODO: Check if this can be combined with filterScopes()
   private validateInputScope(scopes: Array<string>, scopesRequired: boolean): void {
     if (!scopes) {
       if (scopesRequired) {
@@ -2158,25 +2152,6 @@ export class UserAgentApplication {
         throw ClientConfigurationError.createClientIdSingleScopeError(scopes.toString());
       }
     }
-  }
-
-  /**
-  * Used to remove openid and profile from the list of scopes passed by the developer.These scopes are added by default
-  * @hidden
-  */
-  // TODO: Check if this can be combined with validateInputScope()
-  private filterScopes(scopes: Array<string>): Array<string> {
-    if (scopes) {
-      scopes = scopes.filter(function (element) {
-        return element !== Constants.openidScope;
-      });
-
-      scopes = scopes.filter(function (element) {
-        return element !== Constants.profileScope;
-      });
-    }
-
-    return scopes;
   }
 
   /**
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -367,12 +367,8 @@ export class UserAgentApplication {
       // TODO: Should we throw noCallback error here?
     }
 
-    // Validate and filter scopes
-    try {
-      this.validateInputScope(scopes, false);
-    } catch (err) {
-      throw err;
-    }
+    // Validate and filter scopes (the validate function will throw if validation fails)
+    this.validateInputScope(scopes, false);
     scopes = this.filterScopes(scopes);
 
     // extract ADAL id_token if exists
@@ -467,12 +463,8 @@ export class UserAgentApplication {
   acquireTokenRedirect(scopes: Array<string>, authority: string, user: User): void;
   acquireTokenRedirect(scopes: Array<string>, authority: string, user: User, extraQueryParameters: string): void;
   acquireTokenRedirect(scopes: Array<string>, authority?: string, user?: User, extraQueryParameters?: string): void {
-    // Validate and filter scopes
-    try {
-      this.validateInputScope(scopes, true);
-    } catch (err) {
-      throw err;
-    }
+    // Validate and filter scopes (the validate function will throw if validation fails)
+    this.validateInputScope(scopes, true);
     scopes = this.filterScopes(scopes);
 
     // Get the user object if a session exists
@@ -591,12 +583,8 @@ export class UserAgentApplication {
         reject(ErrorCodes.loginProgressError + Constants.resourceDelimeter + ErrorDescription.loginProgressError);
         return;
       }
-      // Validate and filter scopes
-      try {
-        this.validateInputScope(scopes, false);
-      } catch (err) {
-        throw err;
-      }
+      // Validate and filter scopes (the validate function will throw if validation fails)
+      this.validateInputScope(scopes, false);
       scopes = this.filterScopes(scopes);
 
       // Extract ADAL id_token if it exists
@@ -728,12 +716,8 @@ export class UserAgentApplication {
   acquireTokenPopup(scopes: Array<string>, authority: string, user: User, extraQueryParameters: string): Promise<string>;
   acquireTokenPopup(scopes: Array<string>, authority?: string, user?: User, extraQueryParameters?: string): Promise<string> {
     return new Promise<string>((resolve, reject) => {
-      // Validate and filter scopes
-      try {
-        this.validateInputScope(scopes, true);
-      } catch (err) {
-        throw err;
-      }
+      // Validate and filter scopes (the validate function will throw if validation fails)
+      this.validateInputScope(scopes, true);
       scopes = this.filterScopes(scopes);
 
       const scope = scopes.join(" ").toLowerCase();
@@ -982,12 +966,8 @@ export class UserAgentApplication {
   @resolveTokenOnlyIfOutOfIframe
   acquireTokenSilent(scopes: Array<string>, authority?: string, user?: User, extraQueryParameters?: string): Promise<string> {
     return new Promise<string>((resolve, reject) => {
-      // Validate and filter scopes
-      try {
-        this.validateInputScope(scopes, true);
-      } catch (err) {
-        throw err;
-      }
+      // Validate and filter scopes (the validate function will throw if validation fails)
+      this.validateInputScope(scopes, true);
       scopes = this.filterScopes(scopes);
 
       const scope = scopes.join(" ").toLowerCase();
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -369,7 +369,7 @@ export class UserAgentApplication {
 
     // Validate and filter scopes
     try {
-      this.validateInputScope(scopes);
+      this.validateInputScope(scopes, false);
     } catch (err) {
       throw err;
     }
@@ -469,7 +469,7 @@ export class UserAgentApplication {
   acquireTokenRedirect(scopes: Array<string>, authority?: string, user?: User, extraQueryParameters?: string): void {
     // Validate and filter scopes
     try {
-      this.validateInputScope(scopes);
+      this.validateInputScope(scopes, true);
     } catch (err) {
       throw err;
     }
@@ -593,7 +593,7 @@ export class UserAgentApplication {
       }
       // Validate and filter scopes
       try {
-        this.validateInputScope(scopes);
+        this.validateInputScope(scopes, false);
       } catch (err) {
         throw err;
       }
@@ -730,7 +730,7 @@ export class UserAgentApplication {
     return new Promise<string>((resolve, reject) => {
       // Validate and filter scopes
       try {
-        this.validateInputScope(scopes);
+        this.validateInputScope(scopes, true);
       } catch (err) {
         throw err;
       }
@@ -984,7 +984,7 @@ export class UserAgentApplication {
     return new Promise<string>((resolve, reject) => {
       // Validate and filter scopes
       try {
-        this.validateInputScope(scopes);
+        this.validateInputScope(scopes, true);
       } catch (err) {
         throw err;
       }
@@ -2148,18 +2148,27 @@ export class UserAgentApplication {
    * Used to validate the scopes input parameter requested  by the developer.
    *
    * @param {Array<string>} scopes - Developer requested permissions. Not all scopes are guaranteed to be included in the access token returned.
+   * @param {boolean} scopesRequired - Boolean indicating whether the scopes array is required or not
    * @ignore
    * @hidden
    */
   // TODO: Check if this can be combined with filterScopes()
-  private validateInputScope(scopes: Array<string>): void {
+  private validateInputScope(scopes: Array<string>, scopesRequired: boolean): void {
+    if (!scopes) {
+      if (scopesRequired) {
+        throw ClientConfigurationError.createScopesRequiredError(scopes);
+      } else {
+        return;
+      }
+    }
+
     // Check that scopes is an array object (also throws error if scopes == null)
     if (!Array.isArray(scopes)) {
       throw ClientConfigurationError.createScopesNonArrayError(scopes);
     }
 
     // Check that scopes is not an empty array
-    if (!scopes || scopes.length < 1) {
+    if (scopes.length < 1) {
       throw ClientConfigurationError.createEmptyScopesArrayError(scopes.toString());
     }
 
@@ -2177,6 +2186,7 @@ export class UserAgentApplication {
   */
   // TODO: Check if this can be combined with validateInputScope()
   private filterScopes(scopes: Array<string>): Array<string> {
+    if (scopes) {
       scopes = scopes.filter(function (element) {
         return element !== Constants.openidScope;
       });
@@ -2184,6 +2194,7 @@ export class UserAgentApplication {
       scopes = scopes.filter(function (element) {
         return element !== Constants.profileScope;
       });
+    }
 
     return scopes;
   }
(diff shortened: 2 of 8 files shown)


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -366,16 +367,20 @@ export class UserAgentApplication {
...
-              this._tokenReceivedCallback(ErrorDescription.inputScopesError, null, ErrorCodes.inputScopesError, Constants.idToken, this.getUserState(this._cacheStorage.getItem(Constants.stateLogin, this.storeAuthStateInCookie)));
+    try {
+      scopes = this.validateAndFilterInputScope(scopes);
+    } catch (err) {
+      if (err instanceof ClientConfigurationError) {
+        if (this._tokenReceivedCallback) {
+          this._tokenReceivedCallback(ErrorDescription.inputScopesError, null, ErrorCodes.inputScopesError, Constants.idToken, this.getUserState(this._cacheStorage.getItem(Constants.stateLogin, this.storeAuthStateInCookie)));
           return;
+        } else {
+          // We should be checking for callback earlier
```
DarylThayil: should we throw a more specific type of error here? Configuration error?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -366,16 +367,20 @@ export class UserAgentApplication {
...
+          throw err;
         }
+      } else {
+        // TODO: Should create an error and return in error callback here
```
DarylThayil: can we do that in this code?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -471,20 +476,22 @@ export class UserAgentApplication {
   acquireTokenRedirect(scopes: Array<string>, authority: string, user: User, extraQueryParameters: string): void;
   acquireTokenRedirect(scopes: Array<string>, authority?: string, user?: User, extraQueryParameters?: string): void {
     // Validate scopes
-    // TODO: Change to new validation pattern
-    // TODO: is this always access token?
-    const isValidScope = this.validateInputScope(scopes);
-    if (isValidScope && !Utils.isEmpty(isValidScope)) {
+    try {
```
DarylThayil: this is duplicate code, can we factor it out into a util ?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -2173,38 +2178,37 @@ export class UserAgentApplication {
...
-      return "Scopes cannot be passed as an empty array";
+  private validateAndFilterInputScope(scopes: Array<string>): Array<string> {
+    if (scopes == null) {
+      return null;
     }
 
     // Check that scopes is an array object
     if (!Array.isArray(scopes)) {
-      throw new Error("API does not accept non-array scopes");
+      throw ClientConfigurationError.createScopesNonArrayError(scopes);
```
DarylThayil: so its okay to pass in nothing, but not okay to pass in an empty array?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -2173,38 +2178,37 @@ export class UserAgentApplication {
...
-  }
 
-  /**
-  * Used to remove openid and profile from the list of scopes passed by the developer.These scopes are added by default
-  * @hidden
-  */
-  // TODO: Check if this can be combined with validateInputScope()
-  private filterScopes(scopes: Array<string>): Array<string> {
     scopes = scopes.filter(function (element) {
       return element !== "openid";
```
DarylThayil: openid should be a const somewhere and referenced here

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -2173,38 +2178,37 @@ export class UserAgentApplication {
...
    * @param {Array<string>} scopes - Developer requested permissions. Not all scopes are guaranteed to be included in the access token returned.
    * @ignore
    * @hidden
    */
   // TODO: Check if this can be combined with filterScopes()
-  private validateInputScope(scopes: Array<string>): string {
-    // Check that scopes is not an empty array
-    if (!scopes || scopes.length < 1) {
-      return "Scopes cannot be passed as an empty array";
+  private validateAndFilterInputScope(scopes: Array<string>): Array<string> {
```
DarylThayil: This function seems responsible for too much maybe... 
It's not a huge deal, but I think you could break up responsibilities and outcomes of this function by having a isValidScopes() function and filterValidScopes() function 
is valid scopes could return a boolean true/false and instead of try catch code you could do this 
```js
if (!isValidScopes(scopes)) {
   throw ClientConfigurationError.createClientIdSingleScopeError(scopes.toString());
}
const filteredScopes = filterValidScopes(scopes)
```

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1004,105 +1011,103 @@ export class UserAgentApplication {
...
-      if (isValidScope && !Utils.isEmpty(isValidScope)) {
+      try {
+        scopes = this.validateAndFilterInputScope(scopes);
+      } catch (err) {
+        if (err instanceof ClientConfigurationError) {
+          // These will return errors instead of strings in the future
+          // Should we throw instead of reject here?
+          reject(err.message);
+        } else {
+          reject(ErrorCodes.inputScopesError + Constants.resourceDelimeter + ErrorDescription.inputScopesError);
```
DarylThayil: interpolate this string, and if its used elsewhere , factor it out

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1004,105 +1011,103 @@ export class UserAgentApplication {
...
-      // TODO: Replace with new validation pattern
-      const isValidScope = this.validateInputScope(scopes);
-      if (isValidScope && !Utils.isEmpty(isValidScope)) {
+      try {
+        scopes = this.validateAndFilterInputScope(scopes);
+      } catch (err) {
+        if (err instanceof ClientConfigurationError) {
+          // These will return errors instead of strings in the future
+          // Should we throw instead of reject here?
+          reject(err.message);
```
DarylThayil: it is interesting to mix reject and throws, are there reasons why?
my gut tells me this
callback => code should handle in the wild
exception => something unexpected is going on and you need to fix your code

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1004,105 +1011,103 @@ export class UserAgentApplication {
...
+        } else {
+          reject(ErrorCodes.inputScopesError + Constants.resourceDelimeter + ErrorDescription.inputScopesError);
+        }
+      }
+
+      const scope = scopes.join(" ").toLowerCase();
```
DarylThayil: pretty sure this is either factored into a util or should be I have seen it all over

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1004,105 +1011,103 @@ export class UserAgentApplication {
...
+      const userObject = user ? user : this.getUser();
+      const adalIdToken = this._cacheStorage.getItem(Constants.adalIdToken);
+      //if user is not currently logged in and no login_hint/sid is passed as an extraQueryParamater
+      if (!userObject && Utils.checkSSO(extraQueryParameters) && Utils.isEmpty(adalIdToken) ) {
+        this._logger.info("User login is required");
         // TODO: Reject with custom error here
-        // TODO: Is this always accessToken?
-        reject(ErrorCodes.inputScopesError + "|" + isValidScope);
+        reject(ErrorCodes.userLoginError + Constants.resourceDelimeter + ErrorDescription.userLoginError);
```
DarylThayil: same comment here about interp / factor common strings you are using

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1004,105 +1011,103 @@ export class UserAgentApplication {
...
-      } else {
-        // TODO: Remove this from the else block, it is unnecessary
-        if (scopes) {
-          scopes = this.filterScopes(scopes);
-        }
+      }
+      //if user didn't passes the login_hint and adal's idtoken is present and no userobject, use the login_hint from adal's idToken
+      else if (!userObject && !Utils.isEmpty(adalIdToken)) {
+        const idTokenObject = Utils.extractIdToken(adalIdToken);
+        console.log("ADAL's idToken exists. Extracting login information from ADAL's idToken ");
```
DarylThayil: why are we using console and not logger

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1004,105 +1011,103 @@ export class UserAgentApplication {
...
+        extraQueryParameters = Utils.constructUnifiedCacheExtraQueryParameter(idTokenObject, extraQueryParameters);
```
DarylThayil: I am not a huge fan of mutating parameters, it makes it really hard to reason about / trace code.

something gets passed in, and then in some conditional logic may or may not change and this is forwarded on.
This becomes even harder to keep track of when you have shared memory references flying around.

Something we should aim for is that whatever comes into a method, stays pure, and we create new vars as needed.
```js
const unifiedExtraQueryParameters = Utils.constructUnifiedCacheExtraQueryParameter(idTokenObject, extraQueryParameters);
```

Then where ever we need that we can use it, but the original is still pure 
The hope is also that `constructUnifiedCacheExtraQueryParameter` is also creating a new object instead of mutating what is passed in and passing it back

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1004,105 +1011,103 @@ export class UserAgentApplication {
...
-        const scope = scopes.join(" ").toLowerCase();
-        const userObject = user ? user : this.getUser();
-        const adalIdToken = this._cacheStorage.getItem(Constants.adalIdToken);
-        //if user is not currently logged in and no login_hint/sid is passed as an extraQueryParamater
-        if (!userObject && Utils.checkSSO(extraQueryParameters) && Utils.isEmpty(adalIdToken) ) {
-          this._logger.info("User login is required");
-          // TODO: Reject with custom error here
-          reject(ErrorCodes.userLoginError + Constants.resourceDelimeter + ErrorDescription.userLoginError);
-          return null;
+      let authenticationRequest: AuthenticationRequestParameters;
```
DarylThayil: lets can sometimes be good examples of things to factor out.

instead of let and then follow it getting filled, let the logic look like this
```js
const authenticationRequest: AuthenticationRequestParameters = constructAuthRequest(whateverIneed);
```

this then becomes really easy to test right, we have a function that we expect given certain inputs to produce certain results instead of in a big file, hoping the right thing is happening

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1004,105 +1011,103 @@ export class UserAgentApplication {
...
+      if (Utils.compareObjects(userObject, this.getUser())) {
+        if (scopes.indexOf(this.clientId) > -1) {
```
DarylThayil: use array.includes instead of scopes.indexOf

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1004,105 +1011,103 @@ export class UserAgentApplication {
...
-            reject(cacheResult.errorDesc + Constants.resourceDelimeter + cacheResult.error);
-            return null;
-          }
+      const cacheResult = this.getCachedToken(authenticationRequest, userObject);
+      // resolve/reject based on cacheResult
+      if (cacheResult) {
+        if (cacheResult.token) {
+          this._logger.info("Token is already in cache for scope:" + scope);
+          resolve(cacheResult.token);
+          return null;
```
DarylThayil: why return null? why not just `return`? or `return resolve()`

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1004,105 +1011,103 @@ export class UserAgentApplication {
...
         }
-        // else proceed with login
-        else {
-          this._logger.verbose("Token is not in cache for scope:" + scope);
+        else if (cacheResult.errorDesc || cacheResult.error) {
+          this._logger.infoPii(cacheResult.errorDesc + ":" + cacheResult.error);
+          reject(cacheResult.errorDesc + Constants.resourceDelimeter + cacheResult.error);
```
DarylThayil: this can defintelyBeAUtil :D

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1004,105 +1011,103 @@ export class UserAgentApplication {
...
+          return null;
```
DarylThayil: same question about return null

DarylThayil: maybe it's the fn expectation just checking we are not replacing undefined with null

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1004,105 +1011,103 @@ export class UserAgentApplication {
...
-              // Default scope is tracked as clientId to store this token
-              this._logger.verbose("renewing idToken");
-              this.renewIdToken(scopes, resolve, reject, userObject, authenticationRequest, extraQueryParameters);
-            } else {
-              // renew access token
-              this._logger.verbose("renewing accesstoken");
-              this.renewToken(scopes, resolve, reject, userObject, authenticationRequest, extraQueryParameters);
-            }
+        else {
+          if (scopes && scopes.indexOf(this.clientId) > -1 && scopes.length === 1) {
```
DarylThayil: use array.includes

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1004,105 +1011,103 @@ export class UserAgentApplication {
...
+            // App uses idToken to send to api endpoints
+            // Default scope is tracked as clientId to store this token
+            this._logger.verbose("renewing idToken");
```
DarylThayil: how many of these consoles are valuable to leave in? do we have opt in logger messages ?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1004,105 +1011,103 @@ export class UserAgentApplication {
...
-          this._logger.warning("could not resolve endpoints");
-          reject(err);
-          return null;
-        });
-      }
+        }
+      }).catch((err) => {
+        this._logger.warning("could not resolve endpoints");
+        reject(err);
+        return null;
```
DarylThayil: I don't think returning in here does what you think it does, because this happens asynchronously

pkanher617: Not my code.

lib/msal-core/src/UserAgentApplication.ts
pkanher617: .includes does not work on some browsers

lib/msal-core/src/UserAgentApplication.ts
pkanher617: Will resolve offline

lib/msal-core/src/UserAgentApplication.ts
pkanher617: There will be a function that checks for redirect callbacks in a future PR

lib/msal-core/src/UserAgentApplication.ts
pkanher617: Refactoring, please check new commit

lib/msal-core/src/UserAgentApplication.ts
pkanher617: Resolving offline.

lib/msal-core/src/UserAgentApplication.ts
pkanher617: Refactoring, please check new commit

lib/msal-core/src/UserAgentApplication.ts
pkanher617: Refactoring, please check new commit

lib/msal-core/src/UserAgentApplication.ts
pkanher617: Resolving offline.

lib/msal-core/src/UserAgentApplication.ts
pkanher617: Please create an item in VSTS to track this, I will resolve this here.

lib/msal-core/src/UserAgentApplication.ts
pkanher617: Please create an item in VSTS for this. I am resolving this here.

lib/msal-core/src/UserAgentApplication.ts
pkanher617: .includes does not work in certain browsers

lib/msal-core/src/UserAgentApplication.ts
pkanher617: will be refactored in future commit

lib/msal-core/src/UserAgentApplication.ts
pkanher617: Resolving offline

lib/msal-core/src/UserAgentApplication.ts
pkanher617: Resolving offline

lib/msal-core/src/UserAgentApplication.ts
pkanher617: Yeah this was a misunderstanding, I will fix this in a future commit.

lib/msal-core/src/UserAgentApplication.ts
DarylThayil: I am almost positive that is the benefit the typescript compiler gives us, going down to es5

DarylThayil: open to being wrong !

lib/msal-core/src/UserAgentApplication.ts
DarylThayil: discussed in person, doesn't work on ie edge

lib/msal-core/src/UserAgentApplication.ts
pkanher617: Can you try branching and testing this? Happy to integrate it if that is the case.

lib/msal-core/src/UserAgentApplication.ts
pkanher617: I will do this as part of a separate PR.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1003,106 +982,99 @@ export class UserAgentApplication {
   @resolveTokenOnlyIfOutOfIframe
   acquireTokenSilent(scopes: Array<string>, authority?: string, user?: User, extraQueryParameters?: string): Promise<string> {
     return new Promise<string>((resolve, reject) => {
-      // Validate scopes
-      // TODO: Replace with new validation pattern
-      const isValidScope = this.validateInputScope(scopes);
-      if (isValidScope && !Utils.isEmpty(isValidScope)) {
+      // Validate and filter scopes
```
pkanher617: The changes here are simply that the code below was extracted out of the else block of the condition. No other code logic changes were made.

lib/msal-core/src/UserAgentApplication.ts
sameerag: Yes, this is because String[] in es5 does not support include.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -366,17 +367,9 @@ export class UserAgentApplication {
...
-          if (this._tokenReceivedCallback) {
-              this._tokenReceivedCallback(ErrorDescription.inputScopesError, null, ErrorCodes.inputScopesError, Constants.idToken, this.getUserState(this._cacheStorage.getItem(Constants.stateLogin, this.storeAuthStateInCookie)));
-          return;
-        }
-      }
-      scopes = this.filterScopes(scopes);
-    }
+    // Validate and filter scopes (the validate function will throw if validation fails)
+    this.validateInputScope(scopes, false);
+    scopes = this.filterScopes(scopes);
```
sameerag: I know we have discussed this but I believe 'state' has to be a part of error response too as per: https://tools.ietf.org/html/rfc6749#section-10.12 
@manoj-rath Can you check the spec and comment?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -601,19 +583,9 @@ export class UserAgentApplication {
         reject(ErrorCodes.loginProgressError + Constants.resourceDelimeter + ErrorDescription.loginProgressError);
         return;
       }
```
sameerag: I edited Constants.resourceDelimeter in my PR #579, please check that before merging these changes.

lib/msal-core/src/UserAgentApplication.ts
sameerag: This part of the code will change with request. This pattern for extraQueryParameters is seen in Utils.ts too, hence changing it in an irrelevant PR may not make sense. Also touching this would mean core functionality testing, hence a VSTS AI with the aim of evaluating the usage of this pattern in the code makes more sense.

lib/msal-angular/package.json
```diff
@@ -74,7 +74,7 @@
     "ts-loader": "^2.0.3",
     "tslib": "^1.7.1",
     "tslint": "^5.8.0",
-    "typedoc": "^0.6.0",
+    "typedoc": "^0.14.2",
```
DarylThayil: maybe pull dev into your integration branch so this doesn't show up

lib/msal-core/src/UserAgentApplication.ts
pkanher617: This is edited out in a future PR, focus on the changes only please.

lib/msal-angular/package.json
pkanher617: I'm not sure why this is showing up. I have merged the changes from dev into the integration branch...

DarylThayil:
this branch has conflicts which must be resolved

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -2205,13 +2166,15 @@ export class UserAgentApplication {
   */
   // TODO: Check if this can be combined with validateInputScope()
   private filterScopes(scopes: Array<string>): Array<string> {
```
manoj-rath: looks like we remove them here, and add it to every request before we send to server. Ideally we just add if needed, and the code seem to be doing that already, so not sure the value add of this method. We may be able to remove this, but if you are not sure, leave it as is

lib/msal-core/src/UserAgentApplication.ts
manoj-rath: Prithvi and I looked at it. It looks like it matters only in case of successes, not errors because the state is useless if it failed. Also the exiting code was already returning the state in case of call back errors. So we decided to avoid state in case of clientValidation errors and return in case of callback errors.

sameerag: Okay! Sounds good.

lib/msal-core/src/UserAgentApplication.ts
pkanher617: removed the function as it seems like we don't use it anywhere else. We already filter before we send to the server.

PR #583, opened 2019-03-04
API Changes: Configuration

sameerag:
This PR focuses on the "Configuration" passed to the MSAL JS library by the users when the app initializes UserAgentApplication() class. This simplifies the interface by:

- [x] Introducing a "Configuration" object
- [x] Classifying the  "Configuration" object parameters into simpler groups based on their functionality
- [x] Allowing developers to just pass only the needed parameters and an API that builds the object with defaults
- [x] Simplifies existing UserAgentApplication() constructor

**Work Completed** 

- [x] Configuration object added to the code and modified the MSAL interface accordingly introducing 'Configuration.ts'
- [x] Includes UserAgentApplication.ts aligning the new Config object

**Testing**

- [x] Unit tests modified for Configuration Object
- [x] All Unit tests pass

Notes:

- Unit tests are not refactored or modified in functionality. The existing tests are just modified to take the Config object as a parameter and tested to work.
- The streamlining of testing is in pipeline


lib/msal-core/src/Configuration.ts
```diff
@@ -0,0 +1,181 @@
...
+export function buildConfiguration(
+  {
+    clientId = "",
+    authority = null,
+    validateAuthority = true,
+    redirectUri = () => window.location.href.split("?")[0].split("#")[0],
+    postLogoutRedirectUri = () => window.location.href.split("?")[0].split("#")[0],
+    state = "",
+    navigateToLoginRequestUrl = true
+  }: AuthOptions = { clientId },
```
DarylThayil: not sure I full understand this, making sure
define the defaults for authOptions destructure so you can use
what does = { clientId } do?

lib/msal-core/src/Configuration.ts
```diff
@@ -0,0 +1,181 @@
...
+ * @param TAuthOptions
+ * @param TCacheOptions
+ * @param TSystemOptions
+ * @param TFrameworkOptions
+ *
+ * @returns TConfiguration object
+ */
+
+// destructure with default settings
+export function buildConfiguration(
```
DarylThayil: I think I still believe this is something that should be used by new UserAuthAgent(config) to construct the configuration object

lib/msal-core/src/Configuration.ts
```diff
@@ -0,0 +1,181 @@
...
+  {
+    clientId = "",
+    authority = null,
+    validateAuthority = true,
+    redirectUri = () => window.location.href.split("?")[0].split("#")[0],
```
DarylThayil: any chance anywhere through this long call you get a null pointer or undefined?

lib/msal-core/src/Configuration.ts
```diff
@@ -0,0 +1,181 @@
...
+    postLogoutRedirectUri = () => window.location.href.split("?")[0].split("#")[0],
```
DarylThayil: this can be defined above and referenced here especially since this fn is duplicated

lib/msal-core/src/Configuration.ts
```diff
@@ -0,0 +1,181 @@
...
+    state = "",
+    navigateToLoginRequestUrl = true
+  }: AuthOptions = { clientId },
+  {
+    cacheLocation = "sessionStorage",
+    storeAuthStateInCookie = false
+  }: CacheOptions = {},
+  {
+    logger = new Logger(null),
```
DarylThayil: is this in the case where a user might pass in their own logger? what if they pass in a logger that doesn't have the same api as ours ? 
What if they pass in a logger that does have the `logPii` function and expose pii information ?

lib/msal-core/src/Configuration.ts
```diff
@@ -0,0 +1,181 @@
...
+
+import { Logger } from "./Logger";
+
+// make CacheStorage a fixed type to limit it to specific inputs
+export type CacheLocation = "localStorage" | "sessionStorage";
+
+/**
+ * Defaults for the Configuration Options
+ */
+const FRAME_TIMEOUT = 6000;
```
DarylThayil: +1 for keeping consts specific to the file in here instead of in the constants file

lib/msal-core/src/Configuration.ts
sameerag: It can. We can play around when we are building samples. This code can still remain the same. I will discuss this on a wider platform and get back on this.

lib/msal-core/src/Configuration.ts
sameerag: +1.

lib/msal-core/src/Configuration.ts
```diff
@@ -0,0 +1,181 @@
...
+    tokenRenewalOffsetSeconds = OFFSET
+  }: SystemOptions = {},
+  {
+    isAngular = false,
+    unprotectedResources = new Array<string>(),
+    protectedResourceMap = new Map<string, Array<string>>()
+  }: FrameworkOptions = {}
+): Configuration {
+  // restructure
+  let config: Configuration = {
```
DarylThayil: ill make the case one more time for why I do think destructing params here is harder to read and maybe a little duplicative 
 
```js
const DEFAULT_PARAMS = {
    auth: DEFAULT_AUTH_PARMAMS,
    cache: DEFAULT_CACHE_PARAMS,
    system: DEFAULT_SYSTEM_PARAMS,
    framework: DEAFULT_FRAMEWORK_PARAMS
};

...

export const buildConfig = config => ({ ...DEFAULT_CONFIG, config}); 
```

lib/msal-core/src/Configuration.ts
sameerag: They have to adhere to our api today. It is a mutually agreed upon structure as it stands. This PR is just mimicking what is working, an overhaul in future is already in the pipeline from what I understand.

sameerag: When they do not pass a logger, we initialize it for them with our defaults.

lib/msal-core/src/Configuration.ts
sameerag: The de-structuring allows me the pain of merging initiated params. It allows sparse param initialization and pick up the delta with defaults.The user can choose to:

- initialize only select params and have the remaining fill in default
- maintain the structure that would be a json may be eventually

I can write some test code and see if the above format works the same way. One thing that will be working out here for us is that even if the initialization style changes, we still can keep the base structure same. So I would like to keep this open for now and revisit it soon.

lib/msal-core/src/Configuration.ts
sameerag: Changes done. Will close this after I push it.

lib/msal-core/src/Configuration.ts
sameerag: I do not think so. This sure begs for simplification but at this point I am not sure of there is a better way to achieve the same functionality. Can we tag this as to be considered internally?

lib/msal-core/src/Configuration.ts
DarylThayil: This should do the same, essentially use default as a base and spread any passed in params over it.
Not a big deal now if you don't want to, it's a style thing and we can come back and clean up if we want,

DarylThayil:
@sameerag  what code outside of configuration.ts needs review, anything you are concerned about ?

lib/msal-core/src/Configuration.ts
sameerag: Since clientId is a mandatory field, the User should definitely pass the client Id. This code ensures that.

DarylThayil: I think I understand, but I also think that the = { clientId } might be unencessary 
its essentially destructuring the input and making all of the variables have defaults that are also available to the function to use. It is not clear to me what the = { clientId } does. I don't think it makes it available to the fn because it already is

sameerag: Yeah I was just being uniform across my declarations. I tested removing the initializations to empty objects and it seems to work fine. I also need to fine tune the initialization of Config object for better flexibility to the developer, that can be experimented during samples. 

I added the change now removing the {clientId}. Please review and let me know.

lib/msal-core/src/Configuration.ts
sameerag: Tagging this with 'Samples' activity.

lib/msal-core/src/Configuration.ts
sameerag: Pushed the changes. Please review.

lib/msal-core/src/Configuration.ts
sameerag: Resolving this conversation. Please tag this with the upcoming logging testing.

lib/msal-core/src/Configuration.ts
sameerag: Tag this with samples, if we can get the same structure without extra code, I don't mind a different style. I need the fallback to defaults to be:

1) Readable
2) No extra checking of each parameter if it is initialized or not
3) No change in the substructures we created in the Config object

As long as the criteria are satisfied, I am fine. Resolving it here.

lib/msal-core/tests/MSALSpec.js
```diff
@@ -116,20 +116,50 @@ describe('Msal', function () {
...
         global.localStorage = storageFake;
         global.sessionStorage = storageFake;
         global.document = documentMock;
         global.Math = mathMock;
 
-        msal = new MsalModule.UserAgentApplication("0813e1d1-ad72-46a9-8665-399bba48c201", null, function (errorDes, token, error) {
+        // Setting a default value in JSON format
+        var config = {
+            "auth": {
+                "clientId": "0813e1d1-ad72-46a9-8665-399bba48c201",
```
manoj-rath: we need more tests to have good coverage of tests using various combinations of input params.

lib/msal-core/tests/MSALSpec.js
```diff
@@ -116,20 +116,50 @@ describe('Msal', function () {
         };
         window.localStorage = storageFake;
         window.sessionStorage = storageFake;
-        // Init adal 
+        // Init adal
```
manoj-rath: nit: update the comment to msal

lib/msal-core/tests/MSAL.spec.ts
```diff
@@ -166,12 +167,12 @@ describe('Msal', function (): any {
         global.document = documentMock;
         global.Math = mathMock;
 
-        msal = new UserAgentApplication("0813e1d1-ad72-46a9-8665-399bba48c201", null, function (errorDes, token, error) {
-        });
-        msal._user = null;
-        msal._renewStates = [];
-        msal._activeRenewals = {};
-        msal._cacheStorage = storageFake;
+        let config = buildConfiguration({clientId: "0813e1d1-ad72-46a9-8665-399bba48c201" }, {}, {}, {});
```
manoj-rath: is there a way to not have to pass these {} objects?

lib/msal-core/src/Utils.ts
```diff
@@ -457,6 +457,10 @@ export class Utils {
 
   //#region URL Processing (Extract to UrlProcessing.ts?)
 
+  static getDefaultRedirectUri(): string {
```
manoj-rath: is this method used anywhere? i didn't see in this PR.

manoj-rath: never mind, found it in Configuration.ts

lib/msal-core/tests/MSALSpec.js
sameerag: Yes. Probably next week will get a separate PR out for this.

lib/msal-core/tests/MSAL.spec.ts
sameerag: This is in my pending AI, I will write some sample code and confirm this next week.

lib/msal-core/tests/MSALSpec.js
pkanher617: I will take care of this in another PR

lib/msal-core/tests/MSALSpec.js
pkanher617: Resolving for merge.

lib/msal-core/tests/MSAL.spec.ts
pkanher617: Resolving for merge, we can update with a separate PR next week.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -254,89 +170,51 @@ export class UserAgentApplication {
...
+    this.inCookie = this.config.cache.storeAuthStateInCookie;
+
+    // if no authority is passed, set the default: "https://login.microsoftonline.com/common"
+    this.authority = this.config.auth.authority || DEFAULT_AUTHORITY;
+
+    // track login and acquireToken in progress
+    this.userLoginInProgress = false;
+    this.acquireTokenInProgress = false;
+
+    // cache keys msal - typescript throws an error if any value other than "localStorage" or "sessionStorage" is passed
```
DarylThayil: question, do we throw anything if typescript does not? not all consumers are using typescript

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -254,89 +170,51 @@ export class UserAgentApplication {
...
+    try {
+        this.cacheStorage = new Storage(this.config.cache.cacheLocation);
+    } catch (e) {
+        this.config.system.logger.error("CacheLocation can be set only to 'localStorage' or 'sessionStorage' ");
```
DarylThayil: don't we set the local logger when config comes in?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -493,26 +371,26 @@ export class UserAgentApplication {
...
 
     // If no session exists, prompt the user to login.
     if (!userObject && !(extraQueryParameters && (extraQueryParameters.indexOf(Constants.login_hint) !== -1 ))) {
     // TODO: This should be replaced with error callback
     // TODO: Is this always accessToken?
-      if (this._tokenReceivedCallback) {
-        this._logger.info("User login is required");
-        this._tokenReceivedCallback(ErrorDescription.userLoginError, null, ErrorCodes.userLoginError, Constants.accessToken, this.getUserState(this._cacheStorage.getItem(Constants.stateLogin, this.storeAuthStateInCookie)));
+      if (this.tokenReceivedCallback) {
+        this.logger.info("User login is required");
```
DarylThayil: yeah we call logger here, but above call this.config.system.logger

lib/msal-core/src/UserAgentApplication.ts
sameerag: Yes. Need to refer the local logger here. Thanks for the catch.

lib/msal-core/src/UserAgentApplication.ts
sameerag: In that case may be we should not make this language specific?

lib/msal-core/src/UserAgentApplication.ts
sameerag: The above is an oversight. This reference is what is correct. Will update this in the next checkin.

lib/msal-core/src/UserAgentApplication.ts
DarylThayil: When you say language specific what do you mean? Ts vs Js?
I think we should write code for the lowest common denominator we compile down to (es5) and treat the typesafety of typescript as sugar on top for those using the msal lib in vscode 

use case: 

I am using the msal.js file in my spa, downloaded from cdn.  I put in the wrong storage type,  maybe storage: 'database'
Msal.js throws a ClientConfigurationError

usecase2
I am using the msal lib in my project, and in vscode when I try and set storage to 'database' It throws an angry little redline and I fix this before committing code.  (keep in mind here that even then, not all people who use are lib are using typescript to check types before deploying, so it is a dev convenience and we still require the ClientConfigurationError)

sameerag: Yes. That is what I meant. I initially coded that way but got carried away with the convenience of TS doing it for me. Let me make these changes and send out for review sometime today to close this PR and merge it to the API branch.

PR #586, opened 2019-03-06
API Changes: Error - First set of changes to UAA

pkanher617:
Please review PR #582 before reviewing this.

This is the first set of changes in the UserAgentApplication class that utilizes the newly added Error objects for MSAL. These changes include:
- login and acquireToken APIs now throw errors directly for configuration errors, and will return errors in callbacks for all errors returned by the server.
- The openWindow and openPopup functions now use the error objects to communicate popupWindow errors

### DIFF
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -36,7 +36,8 @@ import { User } from "./User";
 import { Utils } from "./Utils";
 import { AuthorityFactory } from "./AuthorityFactory";
 import { ClientConfigurationError } from "./error/ClientConfigurationError";
-
+import { AuthError } from "./error/AuthError";
+import { ClientAuthError } from "./error/ClientAuthError";
 
 /**
  * Interface to handle iFrame generation, Popup Window creation and redirect handling
@@ -90,7 +91,11 @@ export interface CacheResult {
 // TODO: Rework the callback as per new design - handleRedirectCallbacks() implementation etc.
 export type tokenReceivedCallback = (errorDesc: string, token: string, error: string, tokenType: string, userState: string ) => void;
 
-// TODO: Add second callback for error cases
+/**
+ * A type alias of for a errorReceivedCallback function.
+ * @param errorReceivedCallback.err error object created by library that contains the message returned from the STS if API call fails.
+ */
+export type errorReceivedCallback = (err: AuthError) => void;
 
 /**
  * A wrapper to handle the token response/error within the iFrame always
@@ -169,6 +174,11 @@ export class UserAgentApplication {
    * @hidden
    */
   private _tokenReceivedCallback: tokenReceivedCallback = null;
+
+  /**
+   * @hidden
+   */
+  private _errorReceivedCallback: errorReceivedCallback = null;
 
   /**
    * @hidden
@@ -259,6 +269,7 @@ export class UserAgentApplication {
     clientId: string,
     authority: string | null,
     tokenReceivedCallback: tokenReceivedCallback,
+    errorReceivedCallback: errorReceivedCallback,
     options:
       {
         validateAuthority?: boolean,
@@ -305,8 +316,8 @@ export class UserAgentApplication {
     this._logger = logger;
     this.storeAuthStateInCookie = storeAuthStateInCookie;
 
-    // TODO: This should be replaced with two different callbacks for success and error cases
     this._tokenReceivedCallback = tokenReceivedCallback;
+    this._errorReceivedCallback = errorReceivedCallback;
 
     // Track login and acquireToken in progress
     this._loginInProgress = false;
@@ -359,12 +370,7 @@ export class UserAgentApplication {
     3. redirect user to AAD
      */
     if (this._loginInProgress) {
-      // TODO: use error callback here
-      if (this._tokenReceivedCallback) {
-            this._tokenReceivedCallback(ErrorDescription.loginProgressError, null, ErrorCodes.loginProgressError, Constants.idToken, this.getUserState(this._cacheStorage.getItem(Constants.stateLogin, this.storeAuthStateInCookie)));
-        return;
-      }
-      // TODO: Should we throw noCallback error here?
+      throw ClientAuthError.createLoginInProgressError();
     }
 
     // Validate and filter scopes (the validate function will throw if validation fails)
@@ -479,13 +485,8 @@ export class UserAgentApplication {
     // If no session exists, prompt the user to login.
     const scope = scopes.join(" ").toLowerCase();
     if (!userObject && !(extraQueryParameters && (extraQueryParameters.indexOf(Constants.login_hint) !== -1 ))) {
-    // TODO: This should be replaced with error callback
-    // TODO: Is this always accessToken?
-      if (this._tokenReceivedCallback) {
-        this._logger.info("User login is required");
-        this._tokenReceivedCallback(ErrorDescription.userLoginError, null, ErrorCodes.userLoginError, Constants.accessToken, this.getUserState(this._cacheStorage.getItem(Constants.stateLogin, this.storeAuthStateInCookie)));
-        return;
-      }
+      this._logger.info("User login is required");
+      throw ClientAuthError.createUserLoginRequiredError();
     }
 
     // Track the acquireToken progress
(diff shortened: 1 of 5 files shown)


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -837,19 +821,18 @@ export class UserAgentApplication {
    */
   private openWindow(urlNavigate: string, title: string, interval: number, instance: this, resolve?: Function, reject?: Function): Window {
     // Generate a popup window
-    var popupWindow = this.openPopup(urlNavigate, title, Constants.popUpWidth, Constants.popUpHeight);
-
-    // if popupWindow is not valid, throw error
-    if (popupWindow == null) {
+    var popupWindow: Window;
+    try {
+      popupWindow = this.openPopup(urlNavigate, title, Constants.popUpWidth, Constants.popUpHeight);
```
DarylThayil: does this now throw if its null?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -931,6 +913,9 @@ export class UserAgentApplication {
 
       // open the window
       const popupWindow = window.open(urlNavigate, title, "width=" + popUpWidth + ", height=" + popUpHeight + ", top=" + top + ", left=" + left);
+      if (!popupWindow || popupWindow == null) {
+        throw ClientAuthError.createPopupWindowError();
```
DarylThayil: I wonder if this is a client auth error or something that is a problem with our lib.  If A pop up window cant be created thats a big red flag right>?

lib/msal-core/src/error/ClientAuthError.ts
```diff
@@ -84,10 +92,10 @@ export class ClientAuthError extends AuthError {
         Object.setPrototypeOf(this, ClientAuthError.prototype);
     }
 
-    static createEndpointResolutionError(errDesc: string): ClientAuthError {
+    static createEndpointResolutionError(errDetail?: string): ClientAuthError {
         var errorMessage = ClientAuthErrorMessage.endpointResolutionError.desc;
```
DarylThayil: make this a let

lib/msal-core/src/error/ClientAuthError.ts
```diff
@@ -102,9 +110,12 @@ export class ClientAuthError extends AuthError {
             `Cache error for scope ${scope}: ${ClientAuthErrorMessage.multipleMatchingAuthorities.desc}.`);
     }
 
-    static createPopupWindowError(): ClientAuthError {
-        return new ClientAuthError(ClientAuthErrorMessage.popUpWindowError.code,
-            ClientAuthErrorMessage.popUpWindowError.desc);
+    static createPopupWindowError(errDetail?: string): ClientAuthError {
+        var errorMessage = ClientAuthErrorMessage.endpointResolutionError.desc;
```
DarylThayil: this is a little duplicated could make small createErrMessage(description, errDetail? ) fn but its a nitpick not a big deal

lib/msal-core/src/error/ClientConfigurationError.ts
```diff
@@ -23,17 +23,22 @@
...
     invalidCacheLocation: {
         code: "invalid_cache_location",
         desc: "The cache contains multiple tokens satisfying the requirements. " +
             "Call AcquireToken again providing more requirements like authority."
     },
-    noCallback: {
-        code: "no_callback",
-        desc: "Error in configuration: no callback(s) registered for login/acquireTokenRedirect flows. " +
-            "Please call handleRedirectCallbacks() with the appropriate callback signatures. " +
+    noSuccessCallback: {
```
DarylThayil: I am a really big fan about how the splitting of these error classes makes them readible all within one file

lib/msal-core/tests/MSAL.spec.ts
```diff
@@ -679,49 +669,50 @@ describe('Msal', function (): any {
...
-        msal._tokenReceivedCallback = callback;
-        msal.loginRedirect();
-        expect(errDesc).toBe(ErrorDescription.loginProgressError);
-        expect(err).toBe(ErrorCodes.loginProgressError);
-        expect(token).toBe(null);
-        expect(tokenType).toBe(Constants.idToken);
-        expect(state).toBe('https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-implicit-grant-flow?name=value&name2=value2');
-        msal._loginInProgress = false;
-    });
+    // We no longer return state as a part of the error, so these tests need to be updated
```
DarylThayil: delete comments

lib/msal-core/src/UserAgentApplication.ts
pkanher617: Yes, before we were checking the nullity in this function. I have moved that to the previous function where we create the popupWindow and that function now throws if we do not create a window.

lib/msal-core/src/UserAgentApplication.ts
pkanher617: Window.open returns null if the window isn't opened. I haven't found a way to get more information about why it wasn't opened, but we should throw an error if we can't open one.

lib/msal-core/tests/MSAL.spec.ts
pkanher617: Had some conversations about returning state as part of the error object, so I will be uncommenting this in a future PR and fixing this functionality

lib/msal-core/src/error/ClientAuthError.ts
pkanher617: Resolving offline

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -931,6 +913,9 @@ export class UserAgentApplication {
 
       // open the window
       const popupWindow = window.open(urlNavigate, title, "width=" + popUpWidth + ", height=" + popUpHeight + ", top=" + top + ", left=" + left);
+      if (!popupWindow || popupWindow == null) {
```
DarylThayil: 2 things 
1.) if popup window is null !popupWindow will be true, the second predicate is unnecessary
2.) (Almost) Always use === strict equals in javascript

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -479,13 +475,8 @@ export class UserAgentApplication {
...
     const scope = scopes.join(" ").toLowerCase();
     if (!userObject && !(extraQueryParameters && (extraQueryParameters.indexOf(Constants.login_hint) !== -1 ))) {
-    // TODO: This should be replaced with error callback
-    // TODO: Is this always accessToken?
-      if (this._tokenReceivedCallback) {
-        this._logger.info("User login is required");
-        this._tokenReceivedCallback(ErrorDescription.userLoginError, null, ErrorCodes.userLoginError, Constants.accessToken, this.getUserState(this._cacheStorage.getItem(Constants.stateLogin, this.storeAuthStateInCookie)));
-        return;
-      }
+      this._logger.info("User login is required");
```
manoj-rath: ideally this log statement should move to the ClientAuthError class (where you create different types of errors)

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -683,7 +672,7 @@ export class UserAgentApplication {
 
       // What is this? Is this the reject that is passed in?? -- REDO this in the subsequent refactor, passing reject is confusing
       if (reject) {
-        reject(ErrorCodes.endpointResolutionError + ":" + ErrorDescription.endpointResolutionError);
+        reject(ClientAuthError.createEndpointResolutionError());
```
manoj-rath: double check throw vs. reject.
also should we be doing return reject();

lib/msal-core/src/error/ClientAuthError.ts
manoj-rath: endpointResolutionError: typo, you meant popUpWindowError

lib/msal-core/src/UserAgentApplication.ts
pkanher617: We will leave it as reject for now, we should change our guidance to consumers to use .catch for the error cases rather than the reject function pointer.

lib/msal-core/src/UserAgentApplication.ts
pkanher617: We are tracking an item for telemetry, we will revisit the logging when we come to this item.

PR #587, opened 2019-03-06
API Error Changes: Updates to Error Object

pkanher617:
Please review PR #586 before reviewing this.

This set of changes includes:
- Fixed stack variable and added tests to ensure stack trace was working
- Made errorMessage optional

### DIFF
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -478,9 +478,8 @@ export class UserAgentApplication {
     const userObject = user ? user : this.getUser();
 
     // If already in progress, do not proceed
-    // TODO: Should we throw or return an error here?
     if (this._acquireTokenInProgress) {
-      return;
+      throw ClientAuthError.createAcquireTokenInProgressError();
     }
 
     // If no session exists, prompt the user to login.
@@ -577,14 +576,14 @@ export class UserAgentApplication {
     return new Promise<string>((resolve, reject) => {
       // Fail if login is already in progress
       if (this._loginInProgress) {
-        throw ClientAuthError.createLoginInProgressError();
+        reject(ClientAuthError.createLoginInProgressError());
       }
       // Validate and filter scopes (the validate function will throw if validation fails)
       try {
         this.validateInputScope(scopes, false);
       } catch (e) {
         // Rethrow for better error tracking
-        throw e;
+        reject(e);
       }
       scopes = this.filterScopes(scopes);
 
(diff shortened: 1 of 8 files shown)


lib/msal-core/src/error/AuthError.ts
```diff
@@ -36,12 +36,11 @@ export class AuthError extends Error {
     errorCode: string;
     errorMessage: string;
 
-    constructor(errorCode: string, errorMessage: string) {
+    constructor(errorCode: string, errorMessage?: string) {
```
DarylThayil: just doublechecking that we don't wanna set a default here? does anything assume error message is required in this class / subclass

lib/msal-core/tests/error/error.spec.ts
```diff
@@ -8,74 +8,105 @@ describe("Error", () => {
     const TEST_ERROR_CODE: string = "test";
     const TEST_ERROR_MSG: string = "This is a test error";
 
+    let throwFunc = (err: AuthError): void => {
```
DarylThayil: const

lib/msal-core/tests/error/error.spec.ts
```diff
@@ -8,74 +8,105 @@ describe("Error", () => {
...
     describe("ClientAuthError", () => {
         it("can be created", () => {
             let clientAuthError = new ClientAuthError(TEST_ERROR_CODE, TEST_ERROR_MSG);
+            let err: AuthError;
+            try {
+                throwFunc(clientAuthError);
+            } catch(error) {
+                err = error;
+            }
+            console.log("Stack trace: " + err.stack);
```
DarylThayil: do we want to leave the logs in the tests?

lib/msal-core/tests/error/error.spec.ts
DarylThayil: what does this do? cant you just throw an AuthError below if all you are doing is try catching it

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -364,7 +364,12 @@ export class UserAgentApplication {
     }
 
     // Validate and filter scopes (the validate function will throw if validation fails)
-    this.validateInputScope(scopes, false);
+    try {
```
manoj-rath: rethrowing adds unnecessary complexity in the code. We shouldn't use this, and there are multiple locations for this.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -714,7 +731,7 @@ export class UserAgentApplication {
 
       // If already in progress, throw an error and reject the request
       if (this._acquireTokenInProgress) {
-        return reject(ClientAuthError.createAcquireTokenInProgressError());
+        reject(ClientAuthError.createAcquireTokenInProgressError());
```
manoj-rath: is this intentional to remove the return?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -952,7 +973,7 @@ export class UserAgentApplication {
       //if user is not currently logged in and no login_hint/sid is passed as an extraQueryParamater
       if (!userObject && Utils.checkSSO(extraQueryParameters) && Utils.isEmpty(adalIdToken) ) {
         this._logger.info("User login is required");
-        return reject(ClientAuthError.createUserLoginRequiredError());
+        reject(ClientAuthError.createUserLoginRequiredError());
```
manoj-rath: same as above

PR #600, opened 2019-03-19
remove lib files from git

DarylThayil:
pr to remove lib files from git.  The only thing they are used for right now is in the npm package publishing.

Package json only uses lib-commonjs/index.js and this is when someone imports the lib after it has been built and published

Clean up build pt 1  for final preview


DarylThayil:
https://github.com/zeit/pkg
example of another lib where npm build creates a lib-es5 folder but does not commit it
package json references it because it is used only in the published package

DarylThayil:
@sameerag  I am hoping if someone is dependent on them that we break them!

DarylThayil:
merging

sameerag:
@DarylThayil Yeah well! :)

PR #610, opened 2019-03-22
update sample routes so its easy to test other flows without restarti…

DarylThayil:
…ng server


DarylThayil:
wont work with tenant redirect, closing

PR #618, opened 2019-03-27
API changes: Request - Prompt related checks added

sameerag:
Some delta changes needed for Request object to be complete.
Changes include:
- prompt validation checks added.
- Some prettifying the code.


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -231,7 +231,7 @@ export class UserAgentApplication {
    * @param {Array.<string>} scopes - Permissions you want included in the access token. Not all scopes are guaranteed to be included in the access token returned.
    * @param {string} extraQueryParameters - Key-value pairs to pass to the authentication server during the interactive authentication flow.
    */
-  loginRedirect(request: AuthenticationParameters): void {
+  loginRedirect(request?: AuthenticationParameters): void {
```
DarylThayil: is this nullable?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -240,24 +240,17 @@ export class UserAgentApplication {
...
-
-    if (request) {
-      if (request.extraScopesToConsent) {
-        scopes = [...request.scopes, ...request.extraScopesToConsent];
-      }
-      else {
-        scopes = request.scopes;
-      }
-    }
+    scopes = this.appendScopes(request);
```
DarylThayil: why not just say `let scopes = this.appendScopes(request)`

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -474,23 +466,17 @@ export class UserAgentApplication {
...
       let scopes: Array<string>;
-      if (request) {
-        if (request.extraScopesToConsent) {
-          scopes = [...request.scopes, ...request.extraScopesToConsent];
-        }
-        else {
-          scopes = request.scopes;
-        }
-      }
+      scopes = this.appendScopes(request);
```
DarylThayil: same comment here

lib/msal-core/src/Utils.ts
```diff
@@ -712,13 +712,22 @@ export class Utils {
...
+
   /**
    * Utility to test if valid prompt value is passed in the request
    * @param request
    */
-  static validateRequestParameters (request: AuthenticationParameters) {
-    if (!([PromptState.LOGIN, PromptState.SELECT_ACCOUNT, PromptState.CONSENT, PromptState.NONE].indexOf(request.prompt) >= 0)) {
-      console.log("Invalid Prompt Value");
+  static validatePromptParameter (prompt: string) {
+    if (!([PromptState.LOGIN, PromptState.SELECT_ACCOUNT, PromptState.CONSENT, PromptState.NONE].indexOf(prompt) >= 0)) {
```
DarylThayil: can this just be Object.keys(PromptState).indexOf(prompt >= 0)

lib/msal-core/src/UserAgentApplication.ts
sameerag: It is! All the params in request are optional. This is because we enabled users to call loginRedirect() and loginPopup() without any arguments when they just need id_token with no specific scopes attached. I just missed making the main argument optional, hence the correction.

lib/msal-core/src/UserAgentApplication.ts
sameerag: Yeah. Can be. Must have been lost in the merge. Can edit.

lib/msal-core/src/UserAgentApplication.ts
sameerag: Ditto above.

lib/msal-core/src/Utils.ts
sameerag: Actually yes. There was an error seen when I made some ErrorObject call and that will come in a different PR later. Will change it up then.

sameerag: Actually I have checked this, Object.keys(PromptState) keys off the key not the value. User passes "login" but this keys off LOGIN. There could be some other way to match this, resolving this for now. Will revisit when I add the error.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -2360,6 +2369,17 @@ export class UserAgentApplication {
     return serverAuthenticationRequest;
   }
 
+  /**
+   * Utility to test if valid prompt value is passed in the request
+   * @param request
+   */
+  private validatePromptParameter (prompt: string) {
+    console.log("I am here: " + prompt);
```
manoj-rath: remove the log?

sameerag: Done

PR #594, opened 2019-03-13
API Changes Response: Redirect Callback changes

pkanher617:
Here is the first set of changes for the response object, starting with changes for the redirect callbacks and how they will be set and checked in the public APIs.

This PR also includes:
- Separation of redirect and popup/silent flow logic
- Enhanced response handling logic
- Renaming of variables to better reflect usage
- Removal of unnecessary code

PR #619, opened 2019-03-28
API Changes Response: Final Changes

pkanher617:
This PR is dependent on PR #594. Please review that PR before this one. 

This pull request covers:
- Changes to AuthResponse type to class
- Renaming setRedirectCallbacks to handleRedirectCallbacks
- Additional response logic fixes
- Updated tests

### DIFF
--- a/lib/msal-core/src/AuthResponse.ts
+++ b/lib/msal-core/src/AuthResponse.ts
@@ -4,98 +4,14 @@ import { IdToken } from "./IdToken";
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-export class AuthResponse  {
-    private uniqueId: string;
-    private tenantId: string;
-    private tokenType: string;
-    private idToken: IdToken;
-    private accessToken: object;
-    private scopes: Array<string>;
-    private expiresIn: string;
-    private account: User;
-    private userState: string;
-
-    constructor() {
-        this.uniqueId = "";
-        this.tenantId = "";
-        this.tokenType = "";
-        this.idToken = null;
-        this.accessToken = null;
-        this.scopes = null;
-        this.expiresIn = "";
-        this.account = null;
-        this.userState = "";
-    }
-
-    public getUniqueId() {
-        return this.uniqueId;
-    }
-
-    public setUniqueId(inUniqueId: string) : void {
-        this.uniqueId = inUniqueId;
-    }
-
-    public getTenantId() : string {
-        return this.tenantId;
-    }
-
-    public setTenantId(inTenantId: string) : void {
-        this.tenantId = inTenantId;
-    }
-
-    public getTokenType() : string {
-        return this.tokenType;
-    }
-
-    public setTokenType(inTokenType: string) : void {
-        this.tokenType = inTokenType;
-    }
-
-    public getIdToken() : IdToken {
-        return this.idToken;
-    }
-
-    public setIdToken(inIdToken: IdToken) : void {
-        this.idToken = inIdToken;
-    }
-
-    public getAccessToken() : object {
-        return this.accessToken;
-    }
-
-    public setAccessToken(inAccessToken: object) : void {
-        this.accessToken = inAccessToken;
-    }
-
-    public getScopes() : Array<string> {
-        return this.scopes;
-    }
-
-    public setScopes(inScopes: Array<string>) : void {
-        this.scopes = inScopes;
-    }
-
-    public getExpiresIn() : string {
-        return this.expiresIn;
-    }
-
-    public setExpiresIn(inExpiresIn: string) : void {
-        this.expiresIn = inExpiresIn;
-    }
-
-    public getAccount() : User {
-        return this.account;
-    }
-
-    public setAccount(inAccount: User) : void {
-        this.account = inAccount;
-    }
-
-    public getUserState() : string {
-        return this.userState;
-    }
-
-    public setUserState(inUserState: string) : void {
-        this.userState = inUserState;
-    }
-}
+export type AuthResponse = {
+    uniqueId: string,
+    tenantId: string,
+    tokenType: string,
+    idToken: IdToken,
+    accessToken: object,
+    scopes: Array<string>,
+    expiresIn: string,
+    account: User,
+    userState: string
+};
--- a/lib/msal-core/src/AuthResponse.ts
+++ b/lib/msal-core/src/AuthResponse.ts
@@ -5,13 +5,13 @@ import { IdToken } from "./IdToken";
 // Licensed under the MIT License.
 
 export type AuthResponse = {
-    uniqueId: string,
-    tenantId: string,
-    tokenType: string,
-    idToken: IdToken,
-    accessToken: object,
-    scopes: Array<string>,
-    expiresIn: string,
-    account: User,
-    userState: string
+    uniqueId: string;
+    tenantId: string;
+    tokenType: string;
+    idToken: IdToken;
+    accessToken: object;
+    scopes: Array<string>;
+    expiresIn: string;
+    account: User;
+    userState: string;
 };
--- a/lib/msal-core/src/AuthResponse.ts
+++ b/lib/msal-core/src/AuthResponse.ts
@@ -4,14 +4,98 @@ import { IdToken } from "./IdToken";
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-export type AuthResponse = {
-    uniqueId: string;
-    tenantId: string;
-    tokenType: string;
-    idToken: IdToken;
-    accessToken: object;
-    scopes: Array<string>;
-    expiresIn: string;
-    account: User;
-    userState: string;
-};
+export class AuthResponse  {
+    private uniqueId: string;
+    private tenantId: string;
+    private tokenType: string;
+    private idToken: IdToken;
+    private accessToken: object;
+    private scopes: Array<string>;
+    private expiresIn: string;
+    private account: User;
+    private userState: string;
+
+    constructor() {
+        this.uniqueId = "";
+        this.tenantId = "";
+        this.tokenType = "";
+        this.idToken = null;
+        this.accessToken = null;
+        this.scopes = null;
+        this.expiresIn = "";
+        this.account = null;
+        this.userState = "";
+    }
+
+    public getUniqueId() {
+        return this.uniqueId;
+    }
+
+    public setUniqueId(inUniqueId: string) : void {
+        this.uniqueId = inUniqueId;
+    }
+
+    public getTenantId() : string {
+        return this.tenantId;
+    }
+
+    public setTenantId(inTenantId: string) : void {
+        this.tenantId = inTenantId;
+    }
+
+    public getTokenType() : string {
+        return this.tokenType;
+    }
+
+    public setTokenType(inTokenType: string) : void {
+        this.tokenType = inTokenType;
+    }
+
+    public getIdToken() : IdToken {
+        return this.idToken;
+    }
+
+    public setIdToken(inIdToken: IdToken) : void {
+        this.idToken = inIdToken;
+    }
+
+    public getAccessToken() : object {
+        return this.accessToken;
+    }
+
+    public setAccessToken(inAccessToken: object) : void {
+        this.accessToken = inAccessToken;
+    }
+
+    public getScopes() : Array<string> {
+        return this.scopes;
+    }
+
+    public setScopes(inScopes: Array<string>) : void {
+        this.scopes = inScopes;
+    }
+
+    public getExpiresIn() : string {
+        return this.expiresIn;
+    }
+
+    public setExpiresIn(inExpiresIn: string) : void {
+        this.expiresIn = inExpiresIn;
+    }
+
+    public getAccount() : User {
+        return this.account;
+    }
+
+    public setAccount(inAccount: User) : void {
+        this.account = inAccount;
+    }
+
+    public getUserState() : string {
+        return this.userState;
+    }
+
+    public setUserState(inUserState: string) : void {
+        this.userState = inUserState;
+    }
+}
--- a/lib/msal-core/src/ServerRequestParameters.ts
+++ b/lib/msal-core/src/ServerRequestParameters.ts
@@ -137,10 +137,12 @@ export class ServerRequestParameters {
     str.push(`x-client-SKU=${this.xClientSku}`);
     str.push(`x-client-Ver=${this.xClientVer}`);
 
-    str.push("prompt=" + encodeURI(this.promptValue));
+    if (this.promptValue) {
+      str.push("prompt=" + encodeURI(this.promptValue));
+    }
 
     if (this.queryParameters) {
-        str.push(this.queryParameters);
+      str.push(this.queryParameters);
     }
 
     if (this.extraQueryParameters) {
--- a/lib/msal-core/src/ServerRequestParameters.ts
+++ b/lib/msal-core/src/ServerRequestParameters.ts
@@ -23,16 +23,13 @@
 
 import { Authority } from "./Authority";
 import { Utils } from "./Utils";
-import { Constants } from "./Constants";
 
 /**
- * TODO: Change this to ServerRequestParameters soon after Request changes are in.
- *
  * Nonce: OIDC Nonce definition: https://openid.net/specs/openid-connect-core-1_0.html#IDToken
  * State: OAuth Spec: https://tools.ietf.org/html/rfc6749#section-10.12
  * @hidden
  */
-export class AuthenticationRequestParameters {
+export class ServerRequestParameters {
 
   authorityInstance: Authority;
   clientId: string;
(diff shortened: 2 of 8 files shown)


lib/msal-core/src/AuthResponse.ts
```diff
@@ -4,14 +4,98 @@ import { IdToken } from "./IdToken";
...
-    expiresIn: string;
-    account: User;
-    userState: string;
-};
+export class AuthResponse  {
+    private uniqueId: string;
+    private tenantId: string;
+    private tokenType: string;
+    private idToken: IdToken;
+    private accessToken: object;
```
DarylThayil: should we define a type for accessToken

lib/msal-core/src/AuthResponse.ts
```diff
@@ -4,14 +4,98 @@ import { IdToken } from "./IdToken";
...
+    private scopes: Array<string>;
+    private expiresIn: string;
```
DarylThayil: what is an example value here?

lib/msal-core/src/AuthResponse.ts
```diff
@@ -4,14 +4,98 @@ import { IdToken } from "./IdToken";
...
-    tenantId: string;
-    tokenType: string;
-    idToken: IdToken;
-    accessToken: object;
-    scopes: Array<string>;
-    expiresIn: string;
-    account: User;
-    userState: string;
-};
+export class AuthResponse  {
```
DarylThayil: what if instead of a class, with no need for a constructor and a bunch of getters and setters with no tranformations, we did a type?

lib/msal-core/src/ServerRequestParameters.ts
```diff
@@ -137,10 +137,12 @@ export class ServerRequestParameters {
     str.push(`x-client-SKU=${this.xClientSku}`);
     str.push(`x-client-Ver=${this.xClientVer}`);
 
-    str.push("prompt=" + encodeURI(this.promptValue));
+    if (this.promptValue) {
```
DarylThayil: is there a valid set of valid promptValues ?

lib/msal-core/src/ServerRequestParameters.ts
```diff
@@ -137,10 +137,12 @@ export class ServerRequestParameters {
...
+      str.push("prompt=" + encodeURI(this.promptValue));
```
DarylThayil: does this need an & after ?

lib/msal-core/src/ServerRequestParameters.ts
```diff
@@ -137,10 +137,12 @@ export class ServerRequestParameters {
...
+    }
 
     if (this.queryParameters) {
-        str.push(this.queryParameters);
+      str.push(this.queryParameters);
```
DarylThayil: assuming these are properly formatted?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -245,7 +239,7 @@ export class UserAgentApplication {
    * @param {errorReceivedCallback} errorCallback - Callback which contains a AuthError object, containing error data from either the server
    * or the library, depending on the origin of the error.
    */
-  setRedirectCallbacks(successCallback: tokenReceivedCallback, errorCallback: errorReceivedCallback): void {
+  handleRedirectCallbacks(successCallback: tokenReceivedCallback, errorCallback: errorReceivedCallback): void {
```
DarylThayil: setRedirectHandlers ?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1723,7 +1737,7 @@ export class UserAgentApplication {
       const accessTokenCacheItems = this.cacheStorage.getAllAccessTokens(this.clientId, authority);
       for (let i = 0; i < accessTokenCacheItems.length; i++) {
         const accessTokenCacheItem = accessTokenCacheItems[i];
-        if (accessTokenCacheItem.key.userIdentifier === response.account.userIdentifier) {
+        if (accessTokenCacheItem.key.userIdentifier === response.getAccount().userIdentifier) {
```
DarylThayil: is account ever null ?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1733,11 +1747,11 @@ export class UserAgentApplication {
       // Generate and cache accessTokenKey and accessTokenValue
       const expiresIn = Utils.expiresIn(parameters[Constants.expiresIn]).toString();
       const accessTokenKey = new AccessTokenKey(authority, this.clientId, scope, clientObj.uid, clientObj.utid);
-      const accessTokenValue = new AccessTokenValue(parameters[Constants.accessToken], response.idToken.rawIdToken, expiresIn, clientInfo);
+      const accessTokenValue = new AccessTokenValue(parameters[Constants.accessToken], response.getIdToken().rawIdToken, expiresIn, clientInfo);
```
DarylThayil: is idToken ever null

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1746,11 +1760,11 @@ export class UserAgentApplication {
       // Generate and cache accessTokenKey and accessTokenValue
       const accessTokenKey = new AccessTokenKey(authority, this.clientId, scope, clientObj.uid, clientObj.utid);
       // TODO: since there is no access_token, this is also set to id_token?
-      const accessTokenValue = new AccessTokenValue(parameters[Constants.idToken], parameters[Constants.idToken], response.idToken.expiration, clientInfo);
+      const accessTokenValue = new AccessTokenValue(parameters[Constants.idToken], parameters[Constants.idToken], response.getIdToken().expiration, clientInfo);
```
DarylThayil: idToken ever null ?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1914,7 +1930,7 @@ export class UserAgentApplication {
               parameters["error_description"] = "Invalid idToken. idToken: " + parameters[Constants.idToken];
               this.cacheStorage.setItem(Constants.msalError, "invalid idToken");
               this.cacheStorage.setItem(Constants.msalErrorDescription, "Invalid idToken. idToken: " + parameters[Constants.idToken]);
-              error = ClientAuthError.createInvalidIdTokenError(response.idToken);
+              error = ClientAuthError.createInvalidIdTokenError(response.getIdToken());
```
DarylThayil: I think if you got rid of all the getters and setters you could just figure out your type and varaibles and then create an object of type

lib/msal-core/src/AuthResponse.ts
sameerag: No. We are not supposed to open-up tokens, we do for idToken as we need to read some identifier info like upn etc. So even if we create a type, it will be of the type 'object' - intentional blackbox.

lib/msal-core/src/ServerRequestParameters.ts
sameerag: Yes, this came as a part of a different PR #618  which you just reviewed.

lib/msal-core/src/ServerRequestParameters.ts
sameerag: There is a generic API that joins all the strings with '&' following this function.

lib/msal-core/src/ServerRequestParameters.ts
sameerag: Yes. The above comment addresses this.

lib/msal-core/src/AuthResponse.ts
DarylThayil: why don't we do that? isnt it documentation to the developer what the type is

sameerag: It is by design. Developers or anyone is not supposed to pry open the access_token and base their code on the same, since it is a token and not a set of fixed values, which can change anytime depending on the IDP. Hence we discourage them by not defining a type.

DarylThayil: makes sense

lib/msal-core/src/AuthResponse.ts
pkanher617: I had a type before. I changed to class to make initialization of the response object simpler. We can also do any response-based logic in this class in the future, as well as set defaults for values if needed. I am open to using a type, but this was more for future changes as well as code readability.

lib/msal-core/src/AuthResponse.ts
pkanher617: This comes from the server:
&expires_in=3600

We can make this a number instead of string, but I figured for simplicity we just pass back as a string.

lib/msal-core/src/UserAgentApplication.ts
pkanher617: much better name. +1

lib/msal-core/src/UserAgentApplication.ts
pkanher617: saveAccessToken is called when an accessToken has been sent by the server, which would require an idToken and, subsequently, the account object. It is also called when the app saves the idToken as an access token for the application.

lib/msal-core/src/UserAgentApplication.ts
pkanher617: see above comment

lib/msal-core/src/UserAgentApplication.ts
pkanher617: See above comment

lib/msal-core/src/UserAgentApplication.ts
pkanher617: The getters and setters behave functionally the same as a type and variables would. The main benefit of using a class is that instantiation is easier. Using a type would require knowing the values beforehand and setting them at time of creation. With a class we can encapsulate this variable setting and create defaults outside of the UAA class, while keeping the same functionality of a type.

pkanher617: I'm happy to revisit this if needed, but that was my justification for using class over type.

lib/msal-core/src/UserAgentApplication.ts
DarylThayil: we are handling auth response, but here we are setting the handler right ? not actually handling, it makes it clearer we are setting something that will handle at some future point imo

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1944,6 +1905,62 @@ export class UserAgentApplication {
...
+      scope = this.clientId;
+
+      // Generate and cache accessTokenKey and accessTokenValue
+      const accessTokenKey = new AccessTokenKey(authority, this.clientId, scope, clientObj.uid, clientObj.utid);
+      // TODO: since there is no access_token, this is also set to id_token?
+      const accessTokenValue = new AccessTokenValue(parameters[Constants.idToken], parameters[Constants.idToken], response.idToken.expiration, clientInfo);
+      this.cacheStorage.setItem(JSON.stringify(accessTokenKey), JSON.stringify(accessTokenValue));
+      accessTokenResponse.scopes = [scope];
+      accessTokenResponse.accessToken = parameters[Constants.idToken];
+      accessTokenResponse.expiresIn = response.idToken.expiration;
```
DarylThayil: Can we create a new object and not mutate what is passed in, this causes side effects to the caller of the fn

lib/msal-core/src/Utils.ts
```diff
@@ -724,4 +726,18 @@ export class Utils {
 
   //#endregion
 
+  //#region Response Helpers
+
+  static setResponseIdToken(response: AuthResponse, idToken: IdToken) : void {
+    response.idToken = idToken;
```
DarylThayil: same thing here, lets have our functions be pure so the same input always produces the same output

this could be withResponseIdToken(response, idToken)

manoj-rath: the name is still confusing to me.. I don't see a verb

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1816,14 +1787,8 @@ export class UserAgentApplication {
           if (hashParams.hasOwnProperty(Constants.idToken)) {
             response.idToken = new IdToken(hashParams[Constants.idToken]);
           } else {
-            response.idToken = new IdToken(this.cacheStorage.getItem(Constants.idTokenKey));
```
manoj-rath: Is this a merge issue? or why it was removed?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1878,19 +1843,14 @@ export class UserAgentApplication {
             if (!Utils.isEmpty(authority)) {
               authority = Utils.replaceTenantPath(authority, response.idToken.tenantId);
             }
-            if (response.idToken.objectId) {
-              response.uniqueId = response.idToken.objectId;
```
manoj-rath: why removed?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1937,6 +1897,62 @@ export class UserAgentApplication {
...
+   * @param {string} authority authority received in the redirect response from AAD.
+   * @param {TokenResponse} requestInfo an object created from the redirect response from AAD comprising of the keys - parameters, requestType, stateMatch, stateResponse and valid.
+   * @param {User} user user object for which scopes are consented for. The default user is the logged in user.
+   * @param {ClientInfo} clientInfo clientInfo received as part of the response comprising of fields uid and utid.
+   * @param {IdToken} idToken idToken received as part of the response.
+   * @ignore
+   * @private
+   * @hidden
+   */
+  private saveAccessToken(response: AuthResponse, authority: string, parameters: any, clientInfo: string): AuthResponse {
```
manoj-rath: location change of the method? or something changed here. Github diff is not helping me.

pkanher617: Yes, no other changes.

lib/msal-core/src/Utils.ts
pkanher617: I am not going to debate naming here. I will call this setResponseIdToken and we can discuss naming later if it is critical.

lib/msal-core/src/UserAgentApplication.ts
pkanher617: This was not removed. This was extracted into the responseWithIdToken function.

lib/msal-core/src/UserAgentApplication.ts
pkanher617: Not removed. Extracted into the responseWithIdToken function.

lib/msal-core/src/AuthResponse.ts
DarylThayil: if we are just returning the number, lets return the number as a number,
Let's developers create time check on their own end if they want

lib/msal-core/src/UserAgentApplication.ts
DarylThayil: 👍

lib/msal-core/src/UserAgentApplication.ts
DarylThayil: 👍

lib/msal-core/src/Utils.ts
DarylThayil: to me the naming is not the important part, it's that we have functions with side effects

```js
const obj = { a: 'a' };

sideEffectFunction(objectPassed) : void {
    objectPassed.b = 'b';
}

sideEffectFunction(obj);

```
Side effects are hard to reason about, and it makes it hard to tell the state of an application
my suggestion is this

```js
const myObj = { a: 'a' };

// returns new object, old one is never altered
nonSideEffectFunction(passedObject): object {
   return { ...passedObject, b: 'b' };
}

const myNewDecoratedObj = nonSideEffectFunction(myObj);

console.log(myObj); // { a: 'a' }
console.log(myNewDecoratedObj); // { a: 'a', b:'b' } 
```

pkanher617: I have changed it to return an object.

DarylThayil: +1

lib/msal-core/src/UserAgentApplication.ts
pkanher617: Line 257-272 also handles the callback. That is why I changed it from set before

lib/msal-core/src/Utils.ts
DarylThayil: Did it get pushed? I don't see it

lib/msal-core/src/UserAgentApplication.ts
DarylThayil: just a naming opinion, just want to make sure in the public api, it clearly and concisely conveys what the function is doing.  If everyone feels good about that, I am fine with it

lib/msal-core/src/Utils.ts
DarylThayil: name could still be clearer, that its more of a responseWithThisThing, but I am fine with it, its internal fn if we want to change in the future.

lib/msal-core/src/UserAgentApplication.ts
pkanher617: Resolve for now, we can revisit later if needed

lib/msal-core/src/AuthResponse.ts
pkanher617: +1. Eventually we should be returning a Date imo

DarylThayil: agreed, probably want server to return the date, so we aren't trying to guess now + milliseconds

lib/msal-core/src/UserAgentApplication.ts
pkanher617: moved this back to its original location for the purposes of this PR

PR #621, opened 2019-03-29
API Changes: Account (User is being moved to Account with some updates)

sameerag:
> There is one pending change for this PR - accountIdentifier: requirements finalization in progress with the server.

This PR consists of introducing Account object (previously known as User)

Changes made:
- Account.ts - new parameters introduced standardizing the object and compatibility with other MSALs
- getUser() API now changes to getAccount()
- All the user references in the library shifted to account. 
- All dependencies in other .ts files addressed

After this migration any reference to the "User" in the documentation/comments would mean the developer using the library. The logged in context is referred to as "account" as it is possible for a single user to have multiple accounts and hence more realistic nomenclature.


lib/msal-core/src/Account.ts
```diff
@@ -0,0 +1,101 @@
+/**
+ * Copyright (c) Microsoft Corporation
+ *  All Rights Reserved
+ *  MIT License
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of this
```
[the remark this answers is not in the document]
sameerag: Yes. I will do this once I merge #621 to integration branch and pull dev into the integration branch.

lib/msal-core/src/Account.ts
```diff
@@ -0,0 +1,101 @@
+/**
+ * Copyright (c) Microsoft Corporation
```
DarylThayil: Use new copyright in all the other files

lib/msal-core/src/Account.ts
```diff
@@ -0,0 +1,101 @@
...
+    /**
+     * @hidden
+     * @param idToken
+     * @param clientInfo
+     */
+    static createAccount(idToken: IdToken, clientInfo: ClientInfo): Account {
+
+        // create accountIdentifier
+        // TODO: should we advocate idToken.oid+idToken.tid for all MSALs??
+        let accountIdentifier: string;
```
DarylThayil:
```js
const accountIdentifier: string = idToken.objectId ||  idToken.subject;
```

lib/msal-core/src/Account.ts
```diff
@@ -0,0 +1,101 @@
...
+
+        if (idToken.objectId) {
+          accountIdentifier = idToken.objectId;
+        }
+        else {
+          accountIdentifier = idToken.subject;
+        }
+
+        // create homeAccountIdentifier
+        let uid: string;
```
DarylThayil: const uid: string = clientInfo  ? clientInfo.uid : '';
const utid: string = clientInfo ? clientInfo.utid: ''

lib/msal-core/src/Account.ts
```diff
@@ -0,0 +1,101 @@
...
+        let utid: string;
+
+        if (!clientInfo) {
+            uid = "";
```
DarylThayil: we should stick to single quotes, it is convention in js community

lib/msal-core/src/Storage.ts
```diff
@@ -179,15 +179,18 @@ export class Storage {// Singleton
     }
 
     /**
-     * Create acquireTokenUserKey to cache user object
+     * Create acquireTokenAccountKey to cache account object
+     * @param accountId
+     * @param state
      */
-    static generateAcquireTokenUserKey(userId: any, state: string): string {
+    static generateAcquireTokenAccountKey(accountId: any, state: string): string {
```
DarylThayil: should this have a type

lib/msal-core/src/Account.ts
sameerag: Yes. As I mentioned above, I plan to rebase the final integration branch to dev soon and this should be covered in the same.

lib/msal-core/src/Account.ts
sameerag: Was keeping the code pattern constant. Will change this up soon.

lib/msal-core/src/Account.ts
sameerag: Sure.

lib/msal-core/src/Storage.ts
sameerag: no. It is not supposed to as it is a key off with no structure reference from what I understand. I can check again.

lib/msal-core/src/Account.ts
sameerag: Done.

Issue #526, opened 2019-01-06
acquireTokenSilent cause page refresh


Gtyrande:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[x] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Other... Please describe:
</code></pre>

## Browser:
- [x] Center Browser version 3.7.2.33 (Chromium 70.0.3538.110)
- [ ] Chrome version XX
- [ ] Firefox version XX
- [ ] IE version XX
- [ ] Edge version XX
- [ ] Safari version XX

## Library version

<pre><code>
Library version: 0.2.3

## Current behavior
My vue page refreshed after every time acquireTokenSilent has been called, it causes my created() method in App.vue to be called multiple times.

## Expected behavior
My page will not refresh after acquireTokenSilent.

## Minimal reproduction of the problem with instructions
// store.js
state.msalService.getApiToken().then(token => {}
// msal.service.js
getApiToken() {
    return this.app.acquireTokenSilent(this.applicationConfig.apiScopes).then(
      accessToken => {
        return accessToken;
      },
      error => {
        console.log(error);
        return this.app
          .acquireTokenPopup(this.applicationConfig.apiScopes)
          .then(
            accessToken => {
              return accessToken;
            },
            err => {
              console.error(err);
            }
          );
      }
    );
  };

</code></pre>

Ralms:
I'm having the same issue. 
Using Angular, I've noticed that all classes gets re-instantiated when I request a token silently.

Is this reload really required?

vramasam:
We are calling acquireTokenSilently to get the access Token (To extend the token). 

But it causes all classes in the angular gets initiated. Because of that it loses the immemory contents and the UI contents.

Our requirement , we are adding all the REST API request in queue. When we get unauthorised we have to get the token silently and make the pending queue api success. 

But because of entire Angular APP gets reinitiated when we call acquireTokenSilently. All the information(Queue) gets lost.

sameerag:
@vramasam we started investigating this issue. Will update as soon as we find the rootcause for the above behavior stated.

sameerag:
So here is how MSAL JS is designed:
[...]

Issue #393, opened 2018-08-14
Azure AD B2C Msal.js acquireTokenSilent Performance issue


srivathsanvlb:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[x] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Other... Please describe:
</code></pre>

## Browser:
- [x] Chrome version 68.0.3440.106
- [ ] Firefox version XX
- [ ] IE version XX
- [ ] Edge version XX
- [ ] Safari version XX

## Library version

<pre><code>
Library version: msal.js 0.1.3

## Current behavior

Works fine, able to get the access token and call my web api endpoints. But it takes 4-5 seconds to get the access token by using acquireTokenSilent  

## Expected behavior

acquireTokenSilent  should be able to get the access token faster

## Minimal reproduction of the problem with instructions

Have the details here
https://social.msdn.microsoft.com/Forums/en-US/22a53bc6-cd7d-4226-ba61-3ba84368dd8b/azure-ad-b2c-msaljs-acquiretokensilent-performance-issue?forum=WindowsAzureAD#1551122f-9a36-43c6-8d64-f94201e7c877

Have also opened one here
https://github.com/Azure-Samples/active-directory-b2c-javascript-msal-singlepageapp/issues/20

</code></pre>

srivathsanvlb:
Hi @nehaagrawal 

1. I researched a  bit and found that <script src="https://secure.aadcdn.microsoftonline-p.com/lib/0.1.7/js/msal.min.js"></script> is the latest version I am able to get via CDN. Even though Release notes says that the latest version is 0.2.1, I am not able to get it via CDN. Correct if I am wrong.

2. I am NOT able to see any considerable performance improvement after upgrade from 0.1.3 to 0.1.7

3. According to documentation, when app launches, I initialize
this.userAgentApplication = new Msal.UserAgentApplication(ServiceUrl.clientId, ServiceUrl.authority, this.authCallback, { cacheLocation: ServiceUrl.cacheLocation, loadFrameTimeout: ServiceUrl.loadFrameTimeout, navigateToLoginRequestUrl: false });

For every subsequent API call I use
return window.msal.acquireTokenSilent(ServiceUrl.b2cScopes).then((accessToken) => {});
To retrieve the access token and call my web api endpoints. 

[...]

Issue #267, opened 2018-03-09
Unable to use acquireTokenSilent() method for Live Accounts via Azure B2C 


tvperez76:
It seems that the Out of Box experience for using MS Live as a Identity Provider isn't working properly.  I'm able to log in correctly.  However, the acquireTokenSilent() method forces a redirect in my Angular application.  The problem is, MSAL 0.1.5 doesn't set a 'msal.state.acquireToken' entry to the Sesson or Local storage cache, thus causing the URL hash and subsequently the newly acquired access token to be systematically disregarded.  I have included a really nasty hack that seems to address this issue from the client's perspective, but I'd image there must be a much better solution that can be implemented from within the MSAL code base:

```
const hackToSupportLive = 'Renew token Expected state:';
const storage = new Storage('localStorage');
        const logger = new Msal.Logger(
            (level: Msal.LogLevel, message: string, containsPii: boolean) => {
                const index = message.indexOf(AuthenticationService.hackToSupportLive);
                if (index > -1) {
                    const neededState = message.substring(index + 28);
                    storage.setItem(Constants.stateAcquireToken, neededState.trim());
                }
            },
            { level: Msal.LogLevel.Verbose, piiLoggingEnabled: true });

        this.authClient$ = new Msal.UserAgentApplication(
            this.appConfig$.clientID,
            this.appConfig$.authority,
            (errorDesc, token, error, tokenType) => {
                if (!error && token) {
                    console.log(`Token: ${token} - ${tokenType}`);
                } else {
                    console.log(`Error: ${error} - ${errorDesc}`);
                }
            },
            { cacheLocation: 'localStorage', loadFrameTimeout: 20000, navigateToLoginRequestUrl: false, logger: logger });
```

My questions are

1. What's causing MSAL to force a redirect whenever an attempt to acquire an access token from a Live account?
2.  Why doesn't this occur whenever other 3rd party identification providers are used (i.e. Facebook)?
3.  Though the acquireTokenSilent apparently sets a new randomized State ID, why is it not being stored in the Local or Session storage areas so that it can be properly processed during the next callback?

rohitnarula7176:
[...]

PR #622, opened 2019-03-29
fix iFrame redirect not calling handleAuthenticationResponse

step-lee:
The invisible iFrame created by `acquireTokenSilent` will try to load the token url then gets redirect. However, in some case, the redirect will not trigger msal. Resulting in the hash never gets properly parsed. I have changed the behavior to rely on the iFrame calling the `onload` instead of relying on reinitializing msal.

This PR should only effect the `acquireTokenSilent` flow. The `onload` handler on iFrame will trigger 5 times with 100ms delay on each to make sure the iFrame is redirected from login.microsoft.com. After the redirect, it will run the `recursiveDelayCheckFrameHash`. I've also added additional security and clean up on the iFrame. This should prevent the iFrame from loading / initialing the bundle again

### DIFF
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -1126,10 +1126,17 @@ export class UserAgentApplication {
     this._logger.info("LoadFrame: " + frameName);
     const frameCheck = frameName;
     const frameHandle = this.addAdalFrame(frameCheck);
-    // Check hash on iframe
-    // 100ms delay each check for 5 times. 500ms total delay before function times out.
+
     frameHandle.onload = () => {
-      this.recursiveDelayCheckFrameHash(frameHandle, this.IFRAME_HANDLE_RETRY_COUNT);
+        // clear iframe document
+        const frameDoc =
+        frameHandle.contentDocument || frameHandle.contentWindow.document;
+        frameDoc.documentElement.innerHTML = "";
+        this._logger.info("Remove frame");
+        
+        // Check hash on iframe
+        // 100ms delay each check for 5 times. 500ms total delay before function times out.
+        this.recursiveDelayCheckFrameHash(frameHandle, 5);
     };
 
     // TODO: VSTS AI, work on either removing the 500ms timeout or making it optional for IE??
@@ -1151,27 +1158,22 @@ export class UserAgentApplication {
     frameHandle: HTMLIFrameElement,
     numberOfExecutionLeft: number
   ): void {
-    const timeout = setTimeout(() => {
-      const urlHash = getUrlHashFromFrame(frameHandle);
-      const isCallback = this.isCallback(urlHash);
+    const urlHash = getUrlHashFromFrame(frameHandle);
+    const isCallback = this.isCallback(urlHash);
+  
+    // If redirect is successfully and we have the hash
+    if (isCallback) {
+      this.handleAuthenticationResponse(urlHash);
+      return;
+    }
 
-      // If redirect is successfully and we have the hash
-      if (isCallback) {
+    const timeout = setTimeout(() => {
+      if (numberOfExecutionLeft) {
+        this.recursiveDelayCheckFrameHash(frameHandle, numberOfExecutionLeft - 1);
+      } else {
         // clear timeout and proceed
         clearTimeout(timeout);
-
-        // clear iframe so we are not loading any scripts
-        const frameDoc =
-          frameHandle.contentDocument || frameHandle.contentWindow.document;
-        frameDoc.documentElement.innerHTML = "";
-        this._logger.info("Remove frame");
         frameHandle.parentNode.removeChild(frameHandle);
-
-        this.handleAuthenticationResponse(urlHash);
-      } else {
-        if (numberOfExecutionLeft) {
-          this.recursiveDelayCheckFrameHash(frameHandle, numberOfExecutionLeft - 1);
-        }
       }
     }, 100);
   }
@@ -1195,7 +1197,8 @@ export class UserAgentApplication {
         (window.navigator.userAgent.indexOf("MSIE 5.0") === -1)) {
         const ifr = document.createElement("iframe");
         ifr.setAttribute("id", iframeId);
-        ifr.setAttribute("sandbox", "allow-same-origin allow-scripts"); // add additional security to sandbox iFrame
+        // Add additional security to sandbox iFrame (also prevent iFrame executing external scripts)
+        ifr.setAttribute("sandbox", "allow-same-origin");
         ifr.style.visibility = "hidden";
         ifr.style.position = "absolute";
         ifr.style.width = ifr.style.height = "0";
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -14,7 +14,7 @@ import { Logger } from "./Logger";
 import { TokenResponse } from "./RequestInfo";
 import { Storage } from "./Storage";
 import { User } from "./User";
-import { Utils } from "./Utils";
+import { getUrlHashFromFrame, Utils } from "./Utils";
 
 /**
  * Interface to handle iFrame generation, Popup Window creation and redirect handling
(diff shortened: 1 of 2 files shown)


step-lee:
This PR can potentially resolve the below issues:
- #526 Because the iFrame is no longer reloading the whole app. This should prevent re-initializing all the app classes. Therefore, no need to catch the iFrame in the application code.
- #267 The bug are related to: iFrame redirecting the parent and iFrame redirects but hash is not parsed. This PR should fix the second part. In addition, by preventing the iFrame reloading the whole app, the redirect shouldn't happen. At least it should help us narrow down the root cause for the redirect.
- #393 Should help with the latency. Before the iFrame will always wait 500ms. Also not re-initializing the whole app helps the iFrame.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1116,21 +1115,63 @@ export class UserAgentApplication {
...
+  private loadAndRemoveFrame(urlNavigate: string, frameName: string): void {
     // This trick overcomes iframe navigation in IE
     // IE does not load the page consistently in iframe
     this._logger.info("LoadFrame: " + frameName);
     const frameCheck = frameName;
+    const frameHandle = this.addAdalFrame(frameCheck);
+    // Check hash on iframe
+    // 100ms delay each check for 5 times. 500ms total delay before function times out.
+    frameHandle.onload = () => {
+      this.recursiveDelayCheckFrameHash(frameHandle, 5);
```
DarylThayil: maybe pull magic number up into a const IFRAME_HANDLE_RETRY_COUNT or something

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1116,21 +1115,63 @@ export class UserAgentApplication {
...
+      if (isCallback) {
+        // clear timeout and proceed
+        clearTimeout(timeout);
+
+        // clear iframe so we are not loading any scripts
+        const frameDoc =
+          frameHandle.contentDocument || frameHandle.contentWindow.document;
+        frameDoc.documentElement.innerHTML = "";
+        this._logger.info("Remove frame");
+        frameHandle.parentNode.removeChild(frameHandle);
```
DarylThayil: is parent node always here?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1116,21 +1115,63 @@ export class UserAgentApplication {
...
+
+  /**
+   * Recursively check if iFrame has been redirected with hash
+   * Each check has 100ms delay
+   * @ignore
+   * @hidden
+   */
+  private recursiveDelayCheckFrameHash(
+    frameHandle: HTMLIFrameElement,
+    executionsNumber: number
```
DarylThayil: executionNumber instead of executionsNumber

DarylThayil: Maybe a more descriptive name, because its really representing numberOfExecutionsLeft

or retryExecutionLeftToRun or something like that

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1116,21 +1115,63 @@ export class UserAgentApplication {
...
+  ): void {
+    const timeout = setTimeout(() => {
+      const urlHash =
```
DarylThayil: is this in a util or could it be?

DarylThayil: const urlHash = getUrlHashFromFrame(frameHandle)

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1116,21 +1115,63 @@ export class UserAgentApplication {
...
+
+      // If redirect is successfully and we have the hash
+      if (isCallback) {
+        // clear timeout and proceed
+        clearTimeout(timeout);
+
+        // clear iframe so we are not loading any scripts
+        const frameDoc =
+          frameHandle.contentDocument || frameHandle.contentWindow.document;
+        frameDoc.documentElement.innerHTML = "";
```
DarylThayil: is frameDoc ever null

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1152,11 +1193,12 @@ export class UserAgentApplication {
         (window.navigator.userAgent.indexOf("MSIE 5.0") === -1)) {
         const ifr = document.createElement("iframe");
         ifr.setAttribute("id", iframeId);
+        ifr.setAttribute("sandbox", "allow-same-origin allow-scripts"); // add additional security to sandbox iFrame
```
DarylThayil: this is great, sharepoint asked for similar functionality

lib/msal-core/src/UserAgentApplication.ts
step-lee: Yes. Because we insert the frame into the body. Besides `addAdalFrame` will always return a `frameHandle` in this case

lib/msal-core/src/UserAgentApplication.ts
step-lee: Yeah. const magic number is always better

lib/msal-core/src/UserAgentApplication.ts
step-lee: Yes. Definitely could be a util. Not sure where I would put it. In `Utils.ts`?

lib/msal-core/src/UserAgentApplication.ts
step-lee: I don't think so. iFrame should always have a document. It's a DOM element.

lib/msal-core/src/UserAgentApplication.ts
step-lee: Move the function into `Utils.ts`. I put it outside of the utils class because it can be a standalone function. Not need to move it into the class

sameerag:
Thanks @step-lee for the initiative/submission. This is an issue we had in our pipeline and this fix sounds interesting.

I have a couple of questions/observations.
- @pkanher617 This may be either complementing or contrasting your changes on redirectCallBacks, the one that handles the iFrame redirect. Please take a look. We do know we want to resolve this but pushed it until API changes are resolved.

- Removing 500 ms fixed delay (from code inherited from ADAL JS), as the response time for most browsers improved is in the pipeline. It is great if this resolves the same, but since we are trying 100ms *5 times, we many need to check how this works for folks whose ask is to totally remove the delay.

- I am interested to see how this helps #526.

Checking all the issues attached to this PR and would get back on timelines to approve. This submitted code impacts the dev cycle we are on and hence need sometime for impact analysis.

Finally I would recommend very thorough testing in all browsers before accepting this PR, especially IE and Edge.

step-lee:
@sameerag:

- @pkanher617 I would love to connect with you and see the your changes on iFrame redirectCallBacks.

- The 500ms is the worst case scenario. In my app, the callback will happen the first time we call that function. I can change the code to call the function synchronously. And use the recursive timer as a fallback. We will still have 500ms for the worst case but if the redirect in the iFrame happens instantly, we won't get into the 500ms timer.

- In #526, Some of the users reported re-initializing their angular app is causing problem in their code. This PR should address that because the iFrame is not loading any bundle / code. I'm not sure if this will resolve the original problem the user reported. (page reload) But `acquireTokenSilent` will not recreating Angular services(singleton classes) and user don't need to put in the workaround / hack to prevent that.

step-lee:
@sameerag @DarylThayil Updated my PR to address some of the comments.
The `recursiveDelayCheckFrameHash()` will immediately check for the urlHash. If the redirect is not successful, we will then go into the timer (5 * 100ms) to make sure we wait for the redirect. I've also change the sanbox attribute to NOT allow scripts. This way we can guaranteed the iFrame is not executing any of the downloaded external scripts.

DarylThayil:
Big fan of the security feature. 
Have you confirmed that all other flows run with your code? 
also you have some linting errors in build.

benjaminmillhouse:
Is there any kind of timeframe available for this release?

jasonnutter:
@step-lee Do you want to fix up this PR?

step-lee:
@jasonnutter This needs quite a rebase. I'm going to close this PR and submit another one.
When I talked to @sameerag back in May. She's concerned about changing the iframe behavior. So I'll need to spend more time testing it. I'm going to break this PR up into two:
1. security and reduce the default timeout.
2. Change iframe behavior so it doesn't reload the whole app.

sameerag:
Update:  #939 is now merged to dev.

PR #634, opened 2019-04-10
Fix Callback Logic for Redirect and Popup

pkanher617:
This PR is to fix the following bug:
- calling login/acquireTokenPopup or acquireTokenSilent and also setting the handleRedirectCallbacks function would cause a bug where the redirect callbacks would also be triggered in addition to the promise resolve/reject

### DIFF
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -240,14 +240,9 @@ export class UserAgentApplication {
 
     // On the server 302 - Redirect, handle this
     if (!this.config.framework.isAngular) {
-      if (isCallback) {
-        this.handleAuthenticationResponse(urlHash);
-      }
-      else {
-        const pendingCallback = this.cacheStorage.getItem(Constants.urlHash);
-        if (pendingCallback) {
-          this.processCallBack(pendingCallback, null);
-        }
+      const pendingCallback = this.cacheStorage.getItem(Constants.urlHash);
+      if (pendingCallback) {
+        this.processCallBack(pendingCallback, null);
       }
     }
   }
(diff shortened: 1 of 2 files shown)


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -235,19 +235,11 @@ export class UserAgentApplication {
...
     if (!this.config.framework.isAngular) {
-      if (isCallback) {
-        this.handleAuthenticationResponse(urlHash);
-      }
-      else {
-        const pendingCallback = this.cacheStorage.getItem(Constants.urlHash);
-        if (pendingCallback) {
-          this.processCallBack(pendingCallback, null);
-        }
+      const pendingCallback = this.cacheStorage.getItem(Constants.urlHash);
```
step-lee: This will break the current flow for `loginSilent` as we no longer parsing the hash through `handleAuthenticationResponse`

pkanher617: @step-lee Thanks for the review.

Do you mean acquireTokenSilent? We don't have any loginSilent APIs.

We are still parsing the hash in the constructor - this removes that logic from the handleRedirectCallbacks method. I removed the handleAuthenticationResponse call to avoid parsing the hash multiple times in the case of popup or iframe and the user mistakenly uses the handleRedirectCallbacks function.

Our code has evolved quite a bit, so the diff may be hiding some context for you. Feel free to ping me in Teams if you want to dive a bit deeper into this.

step-lee: Yes I mean the `acquireTokenSilent`. My apology on not reading the changes more carefully. You can discard my comment as this behavior is not happening in the constructor.

PR #635, opened 2019-04-10
Introduce response object to getCachedToken

pkanher617:
This PR introduces the response and error to the getCachedToken code path. This will allow all ATS calls to return response and error objects instead of strings, as before. This PR also introduces minor changes to setting expiration date in the response object.


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1714,12 +1718,16 @@ export class UserAgentApplication {
...
 
-      // TODO: since there is no access_token, this is also set to id_token?
       const accessTokenValue = new AccessTokenValue(parameters[Constants.idToken], parameters[Constants.idToken], response.idToken.expiration, clientInfo);
       this.cacheStorage.setItem(JSON.stringify(accessTokenKey), JSON.stringify(accessTokenValue));
       accessTokenResponse.scopes = [scope];
       accessTokenResponse.accessToken = parameters[Constants.idToken];
-      accessTokenResponse.expiresIn = response.idToken.expiration;
+      let exp = Number(response.idToken.expiration);
+      if (exp) {
+        accessTokenResponse.expiresIn = new Date(exp * 1000);
```
sameerag: Should this be new Date(Date.now() + (exp * 1000)); or new Date(exp * 1000);?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1704,8 +1703,13 @@ export class UserAgentApplication {
       this.cacheStorage.setItem(JSON.stringify(accessTokenKey), JSON.stringify(accessTokenValue));
 
       accessTokenResponse.accessToken  = parameters[Constants.accessToken];
-      accessTokenResponse.expiresIn = expiresIn;
       accessTokenResponse.scopes = consentedScopes;
+      let exp = Number(expiresIn);
+      if (exp) {
+        accessTokenResponse.expiresIn = new Date(Date.now() + (exp * 1000));
```
sameerag: Is there a reason why this isn't consistent here and in the else case?
Should this be new Date(Date.now() + (exp * 1000)); or new Date(exp * 1000);?

lib/msal-core/package.json
```diff
@@ -29,7 +29,7 @@
     "doc": "typedoc --out ./docs ./src/ --gitRevision dev",
     "build:modules": "tsc && tsc -m es6 --outDir lib-es6",
     "build": "npm run clean && npm run doc && npm run build:modules && npm run lint && npm run build:webpack && npm test",
-    "test": "karma start && jasmine tests/MSALSpec.js",
+    "test": "karma start",
```
sameerag: I know this change is made because the tests are duplicating but travis will still fail for this. :)

lib/msal-core/src/AuthResponse.ts
```diff
@@ -12,9 +12,9 @@ export type AuthResponse = {
     tenantId: string;
     tokenType: string;
     idToken: IdToken;
-    accessToken: object;
+    accessToken: string;
     scopes: Array<string>;
-    expiresIn: string;
+    expiresIn: Date;
```
manoj-rath: if we are calling expiresIn, it should have the relative time period (e.g. 300 secs)
if you return date object, then we should name it different

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1533,26 +1522,36 @@ export class UserAgentApplication {
...
-          error: null
+        const idToken = new IdToken(accessTokenCacheItem.value.idToken);
+        if (!account) {
+          account = this.getAccount();
+          if (!account) {
+            throw AuthError.createUnexpectedError("Account should not be null here.");
+          }
+        }
+        const aState = this.getAccountState(this.cacheStorage.getItem(Constants.stateLogin, this.inCookie));
+        let response : AuthResponse = {
```
manoj-rath: can we pls discuss this?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1533,26 +1522,36 @@ export class UserAgentApplication {
...
+          uniqueId: "",
+          tenantId: "",
+          tokenType: (accessTokenCacheItem.value.idToken === accessTokenCacheItem.value.accessToken) ? Constants.idToken : Constants.accessToken,
+          idToken: idToken,
+          accessToken: accessTokenCacheItem.value.accessToken,
+          scopes: serverAuthenticationRequest.scopes,
+          expiresIn: new Date(expired * 1000),
```
manoj-rath: ?

lib/msal-core/src/UserAgentApplication.ts
pkanher617: In the else case we are getting the exp param from the token. That is the Numeric Date (in relation to 1-1-1970). In the if case we get expiresIn, a param from AAD which gives the lifetime of the token in seconds (i.e. 3600). The else case is more accurate, but we can only if it is an idToken since we do not unpack accessTokens.

lib/msal-core/src/UserAgentApplication.ts
pkanher617: See above comment

lib/msal-core/src/AuthResponse.ts
DarylThayil: should just be a number type for milliseconds

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -298,13 +298,12 @@ export class UserAgentApplication {
         let tokenRequest: AuthenticationParameters = this.buildIDTokenRequest(request);
 
         this.silentLogin = true;
-        this.acquireTokenSilent(tokenRequest).then((idToken) => {
+        this.acquireTokenSilent(tokenRequest).then((response) => {
```
DarylThayil: no parens needed around single param

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -519,12 +518,11 @@ export class UserAgentApplication {
 
           this.silentLogin = true;
           this.acquireTokenSilent(tokenRequest)
-              .then((idToken) => {
+              .then((response) => {
```
DarylThayil: ditto

lib/msal-core/src/UserAgentApplication.ts
DarylThayil: why not just use a ms time ?

lib/msal-core/src/UserAgentApplication.ts
DarylThayil: what is exp if not set ? would the output of this be NaN? 
what if exp is 0 this will be falsey
what is difference between Number(input) vs parseInt(num, 10)

pkanher617: Number will either return a number or NaN. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number
For empty or null values, Number(x) will return 0. 
For any string values, Number(x) will return NaN. 
NaN evaluates to falsey. https://developer.mozilla.org/en-US/docs/Glossary/Falsy

lib/msal-core/src/AuthResponse.ts
pkanher617: Trying to be consistent with other MSAL platforms here. MSAL Android and MSAL Obj-C are using the corresponding Date format for their platforms, and returning a var called 'expiresOn'. I will change this var here to be expiresOn instead of expiresIn to better reflect its usage.

lib/msal-core/src/UserAgentApplication.ts
pkanher617: Want to be consistent with other MSALs here, where we are returning Date format.

lib/msal-core/package.json
pkanher617: Let me take a look and try to fix. thanks.

lib/msal-core/src/UserAgentApplication.ts
pkanher617: Number will parse the entire string, whereas parseInt parses up to the first non-digit and then returns what has been parsed up until then. I think Number is a better fit for this use case.

lib/msal-core/package.json
sameerag: You can also change travis config if you are sure the tests are duplicating.

lib/msal-core/src/AuthResponse.ts
```diff
@@ -12,9 +12,9 @@ export type AuthResponse = {
     tenantId: string;
     tokenType: string;
     idToken: IdToken;
-    accessToken: object;
+    accessToken: string;
     scopes: Array<string>;
-    expiresIn: string;
+    expiresOn: Date;
```
manoj-rath: did you confirm what other teams are using?

lib/msal-core/package.json
pkanher617: @sameerag managed to fix it somehow, thanks!

lib/msal-core/src/UserAgentApplication.ts
pkanher617: Any open questions?

lib/msal-core/src/UserAgentApplication.ts
pkanher617: What is the question here?

lib/msal-core/src/UserAgentApplication.ts
pkanher617: resolved in latest commit

lib/msal-core/src/UserAgentApplication.ts
pkanher617: resolved in latest commit

lib/msal-core/src/UserAgentApplication.ts
manoj-rath: we should standardize the storage of the expiry in the cache such that it's consistent, something like seconds since epoch. looks like currently, it's either epoch time in case of id_token, and duration in case of access_token. With that, please make sure every caller/usage of this is modified appropriately.

lib/msal-core/src/AuthResponse.ts
pkanher617: other teams are also using expiresOn

lib/msal-core/src/UserAgentApplication.ts
pkanher617: After reviewing this further, Utils.now() returns time in Unix time (seconds), which is the same as the format of the exp claim in the id token.

pkanher617: Therefore, as long as we use utils.now when we are storing the access tokens, we can always be sure that the cache is storing expiration time in seconds.

PR #641, opened 2019-04-12
Security Vulnerabilties

DarylThayil:
1 moderate leftover, but it is in typedoc and a part of our build chain.
The vulnerabiltity has to do with poor performance internals creating the opportunity to hammer and
cause denial of service, however it is never deploted and therfore is not a threat from attackers.


lib/msal-core/package.json
```diff
@@ -48,16 +48,16 @@
...
     "jsdom-global": "3.0.2",
     "jshint": "^2.9.7",
-    "karma": "^2.0.5",
+    "karma": "^4.0.1",
     "karma-chai": "^0.1.0",
     "karma-chrome-launcher": "^2.2.0",
     "karma-cli": "^1.0.1",
     "karma-coverage-istanbul-reporter": "^2.0.1",
     "karma-firefox-launcher": "^0.1.7",
-    "karma-html-reporter": "^0.2.7",
```
sameerag: We don't need this? Or is this removed to avoid vulnerabilities? Or does js-yaml replace this?

DarylThayil: un needed.  We don't do any html reporting with karma that I can see.  There are a ton of these we can prune but this one had vulnerabilities

sameerag: Great. Thanks. Yes, we need to plan to prune all unneeded packages eventually.

DarylThayil:
Yes, I did. It was all test deps

PR #643, opened 2019-04-13
Telemetry Manager and Base Events

DarylThayil:
This pull request adds the following

- TelemetryManager - the class that will be responsible for taking events from the code and starting, stopping the, and storing them to be returned to developer.  It is configured with platform /app info + a callback that is created by the developer

- TelemetryPlatform / Config - Config for Telemetry Manager

- TelemetryEvent - The base event that all events will inherit from (apievent, uievent, cacheevent etc... coming in next pr) It is responsible for storing start and end time about a specific event along with metadata about a specific event. 

- DefaultEvent - A special event that is probably poorly named but follows convention of other adal/msal.  It is the "AggregateEvent". On a flush for a correlationId, the default event is appended to the end of all the events, with aggregate counts of certain types of events + platform info on the set of events. 

- TelemetryConstants - A set of constants used for the Telemetry classes 

- TelemetryUtils  - A set of utils used specifically for telemetry

-TelemetryTypes - Interfaces / Types for different Objects 
- Tests - !Yay


lib/msal-core/src/telemetry/TelemetryTypes.ts
```diff
@@ -0,0 +1,20 @@
+import TelemetryEvent from "./TelemetryEvent";
+
+export interface InProgressEvents { [key: string] : TelemetryEvent; }
+export interface EventCount { [eventName: string] : number; }
+export interface EventCountByCorrelationId { [correlationId: string] : EventCount; }
+export interface CompletedEvents { [correlationId: string ] : Array<TelemetryEvent>; }
+
+// SDK SHOULD BE DEFAULTED and Pulled from Package
+export interface TelemetryPlatform {
+    sdk: string;
```
pkanher617: Is this to record whether we are using angular, react, etc.?

package-lock.json
```diff
@@ -0,0 +1,2738 @@
+{
```
pkanher617: Why do we have a package-lock in the base folder?

lib/msal-core/package.json
```diff
@@ -38,8 +38,9 @@
     "@types/chai": "^4.1.7",
     "@types/handlebars": "4.0.33",
     "@types/mocha": "^5.2.6",
-    "@types/node": "^11.11.5",
+    "@types/node": "^11.13.4",
```
pkanher617: Any breaking changes?

lib/msal-core/src/telemetry/TelemetryTypes.ts
DarylThayil: no this is for sdk: 'javascript' or msaljs. To differentiate between other msals  in telemetry aggregates

package-lock.json
DarylThayil: eww whoops
killing

DarylThayil: removed

lib/msal-core/src/telemetry/TelemetryConstants.ts
```diff
@@ -0,0 +1,19 @@
+export const EVENT_NAME_PREFIX: String = "msal.";
```
pkanher617: I like the code style here, but we should match our constants format across the project. 

Also should we start splitting constants up by usage across other files as well?

lib/msal-core/package.json
DarylThayil: no, its a minor update, just wanted to update types for node libs

lib/msal-core/src/telemetry/TelemetryTypes.ts
pkanher617: Shouldn't we be setting this as a readonly then? Are we allowing devs to modify this value?

lib/msal-core/src/telemetry/TelemetryConstants.ts
DarylThayil: Tbd how we split up, but I dont want to add to the monster const file we have, so I am modularizing for telemetry. Also there is a good chance this code will be modularized and reused for node as well and so I don't want to have to rip it out in the future again.

lib/msal-core/src/telemetry/TelemetryTypes.ts
DarylThayil: yeah that is probably better

DarylThayil: I actually think I am going to wait till I get to the point where this is wired through to Configuration to see where the default should be.  TelemetryManager will be constructed by UAA so I just want to make sure I put the defaults at the right level.  Again this code should be able to be reused for any js platform lib we build

lib/msal-core/src/telemetry/TelemetryConstants.ts
pkanher617: We should probably stop growing that file at some point then, and try to separate it out so we can have better control and readability. Maybe a constants subdirectory or something.

lib/msal-core/src/telemetry/TelemetryTypes.ts
pkanher617: No problem, just want to make sure we don't leave any room for the developer to mess up the log format we expect

pkanher617: setting default isn't what I'm worried about here, its more of can the developer change the value and possibly change the format of expected logs.

DarylThayil: Yep, make sense.  I think that will happen in the config object based on what we expose right?

```js
new UserAgenApplication({
   system: { 
      telemetry: {  // dont expose sdk here so that we can configure it internally. Dev has no access
          telemetryCallback,
          applicationName,
          applicationVersion,
          onlySendFailureTelemetry?
      }
   }
})
```

lib/msal-core/src/telemetry/TelemetryConstants.ts
DarylThayil: yep big fan of that, as we modularize, we should namespace constants to their own files / objects

PR #645, opened 2019-04-17
Improved Unit Testing - First Cut

pkanher617:
The following PR adds additional Unit Tests, as well as fixes or notes for bugs found during the first cut of testing.
Major Changes:
- Tests for merged changes now do not overwrite private functions or give false positives
- isInIframe made public
- Notes for future bug fixes


lib/msal-core/karma.conf.js
```diff
@@ -7,16 +7,23 @@ module.exports = function (config) {
         frameworks: ['mocha', 'chai'],
         files: [
             'node_modules/babel-polyfill/dist/polyfill.js',
-            'tests/mochaTests/**/*.ts',
+            'tests/**/*.ts',
+        ],
+        exclude: [
+            'tests/MSAL*',
```
DarylThayil: should this have a file extension?

lib/msal-core/karma.conf.js
```diff
@@ -7,16 +7,23 @@ module.exports = function (config) {
...
+            'tests/Authority.spec.ts'
```
DarylThayil: what happened to tests in mocha tests?

pkanher617: Moved into the UAA.spec file.

lib/msal-core/karma.conf.js
pkanher617: this is just temporary, this will be gone in the future.

PR #658, opened 2019-04-23
Use Error-first callbacks

pkanher617:
In order to comply with most common JS practices, instead of using separate callbacks we will use an error-first callback instead. This will allow us to simplify some of our callback logic.


lib/msal-core/src/AuthResponse.ts
```diff
@@ -18,3 +18,19 @@ export type AuthResponse = {
     account: Account;
     accountState: string;
 };
+
+export function buildResponseStateOnly(state: string) : AuthResponse {
+    let response : AuthResponse = {
```
DarylThayil: could just 
```js
return { 
  fields
};
```

instead of defining a var you are not using

lib/msal-core/test/error/ClientConfigurationError.spec.ts
```diff
@@ -83,9 +83,8 @@ describe("ClientConfigurationError", () => {
 
   it("createInvalidCallbackObjectError creates a ClientConfigurationError object", () => {
 
-    const callbackType = "successCallback";
-    const callbackFunction: Object = "";
-    const invalidCallBackObject = ClientConfigurationError.createInvalidCallbackObjectError(callbackType, callbackFunction);
+    const callbackFunction: Object = null;
```
DarylThayil: should be type function?

lib/msal-core/test/UserAgentApplication.spec.ts
```diff
@@ -53,6 +53,15 @@ describe("UserAgentApplication", function () {
 
     let msal: UserAgentApplication;
 
+    let authCallback = function (error, response) {
```
DarylThayil: all these lets can be consts

lib/msal-core/test/UserAgentApplication.spec.ts
```diff
@@ -53,6 +53,15 @@ describe("UserAgentApplication", function () {
...
+        if (error) {
+            console.log("Error: " + error);
```
DarylThayil: why have console logs in spec files?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -395,7 +383,7 @@ export class UserAgentApplication {
 
     // If already in progress, do not proceed
     if (this.acquireTokenInProgress) {
-      this.errorReceivedCallback(ClientAuthError.createAcquireTokenInProgressError(), this.getAccountState(this.silentAuthenticationState));
+      this.authResponseCallback(ClientAuthError.createAcquireTokenInProgressError(), buildResponseStateOnly(this.getAccountState(this.silentAuthenticationState)));
```
DarylThayil: so the return is (err, resp) ?

DarylThayil: have we updated docs with this usage

lib/msal-core/test/UserAgentApplication.spec.ts
pkanher617: More of a placeholder than anything, these only get called if there is any issue with the tests or the tests fail.

lib/msal-core/src/UserAgentApplication.ts
pkanher617: No not yet, presenting the PR first, we will update docs before i merge this.

pkanher617: Yes that is correct for the function signature

lib/msal-core/test/UserAgentApplication.spec.ts
DarylThayil: can we just throw the error so it shows up?

lib/msal-core/src/UserAgentApplication.ts
DarylThayil: What is the impact on graph and railyard?

DarylThayil: and the pax team?

pkanher617: It will be a breaking change, we discussed with state as well. We can take this discussion offline and merge this PR later if needed. Resolving for now.

lib/msal-core/src/AuthResponse.ts
pkanher617: updated, resolving comments. unresolve if you feel they arent addressed.

PR #659, opened 2019-04-24
Minor SSOParams fix

pkanher617:
Makes ssoParam in addSSOParameter optional


DarylThayil:
lgtm, should we add a test for this?

pkanher617:
Tests for the query param outputs exist in UserAgentApplication.spec.ts. @sameerag or I will be adding Utils tests in the near future I believe.

PR #664, opened 2019-04-26
Pass Claims in Request Object and Skip cache lookup

pkanher617:
This PR covers the claims request feature, and is a prerequisite for the Conditional Access feature. 

This PR implements the following features:
- Pass claims string to the request object
- Removes any strings passed to extra query parameters if passed in the request object
- Skips cache lookup if claims parameter is passed in by user

Unit testing for the above features are also included.


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -2391,16 +2404,27 @@ export class UserAgentApplication {
...
+  private sanitizeEQParams(request: AuthenticationParameters) : QPDict {
+    let eQParams : QPDict = request.extraQueryParameters;
+    if (!eQParams) {
+      return null;
+    }
+    if (request.claimsRequest) {
+      this.logger.warning("Removed duplicate claims from extraQueryParameters. Please use either the claimsRequest field OR pass as extraQueryParameter - not both.");
+      delete eQParams[Constants.claims];
+    }
+    if (request.sid) {
```
sameerag: Please remove the if check. The idea is to remove any sid, login_hint (and may be claims too) from extraQueryParameters as we do not support the pattern anymore. 

Also this check will be a problem for anyone passing sid by accident in extraQueryParameters and contrasts with our library populating the hints in addHintParameters.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -2391,16 +2404,27 @@ export class UserAgentApplication {
...
+      this.logger.warning("Removed duplicate sid from extraQueryParameters. Please use either the sid field OR pass as extraQueryParameter - not both.");
       delete eQParams[SSOTypes.SID];
+    }
+    if (request.loginHint) {
```
sameerag: Same comment as above.

PR #665, opened 2019-04-28
update acquireTokenAccountKey generation when client_info is empty

sameerag:
- Fix the case where the account has no "homeAccountIdentifier" while generating an acquireTokenAccountKey in the code. 
- Minor name change to adhere to "account" instead of "user"
- Test cases modified for Storage class

### DIFF
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -1744,7 +1744,14 @@ export class UserAgentApplication {
         authorityKey = Storage.generateAuthorityKey(stateInfo.state);
 
         const account: Account = this.getAccount();
-        const accountId: string = account ? this.getAccountId(account) : "";
+        let accountId;
+
+        if (account && !Utils.isEmpty(account.homeAccountIdentifier)) {
+            accountId = account.homeAccountIdentifier;
+        }
+        else {
+            accountId = Constants.no_account;
+        }
 
         acquireTokenAccountKey = Storage.generateAcquireTokenAccountKey(accountId, stateInfo.state);
       }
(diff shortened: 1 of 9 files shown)


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -2295,8 +2315,15 @@ export class UserAgentApplication {
    * @param state
    */
   private setAccountCache(account: Account, state: string) {
+
     // Cache acquireTokenAccountKey
-    let accountId = account ? this.getAccountId(account) : Constants.no_account;
+    let accountId: string;
```
pkanher617: Why are we expanding the ternary operators?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1792,15 +1799,24 @@ export class UserAgentApplication {
...
+          }
+          else {
+            accountKey = Constants.no_account;
+          }
 
           acquireTokenAccountKey = Storage.generateAcquireTokenAccountKey(accountKey, stateInfo.state);
           const acquireTokenAccountKey_noaccount = Storage.generateAcquireTokenAccountKey(Constants.no_account, stateInfo.state);
 
           let cachedAccount: string = this.cacheStorage.getItem(acquireTokenAccountKey);
+          console.log("cache acquireTokenAccountKey", cachedAccount);
```
pkanher617: Is this supposed to be in here?

sameerag: No. Thanks, will remove it. Was testing.

lib/msal-core/src/UserAgentApplication.ts
sameerag: This can be reverted. There are multiple iterations that this code went through during requirements clarity on where we want to put the check on client_info. Reverting in a new commit.

PR #666, opened 2019-04-28
Handle 'undefined' case for SSO Params generation

sameerag:
In error cases where homeAccountIdentifier is undefined from the server response, need to add a 'undefined' case for SSO Params generation.


sameerag:
Merged this with PR #665

PR #672, opened 2019-05-01
update readme code patterns

DarylThayil:
update code patterns in readme, will be followed by improved content, probably splitting apart confusing parts of the code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,11 +2,11 @@
 Microsoft Authentication Library Preview for JavaScript (MSAL.js)
 =========================================================
 
-| [Getting Started](https://docs.microsoft.com/en-us/azure/active-directory/develop/guidedsetups/active-directory-javascriptspa)| [Docs](https://aka.ms/aaddevv2) | [Library Reference](https://htmlpreview.github.io/?https://raw.githubusercontent.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/docs/classes/_useragentapplication_.useragentapplication.html) | [Support](README.md#community-help-and-support) | [Samples](https://github.com/AzureAD/microsoft-authentication-library-for-js/wiki/Samples)
+| [Getting Started](https://docs.microsoft.com/en-us/azure/active-directory/develop/guidedsetups/active-directory-javascriptspa)| [AAD Docs](https://aka.ms/aaddevv2) | [Library Reference](https://htmlpreview.github.io/?https://raw.githubusercontent.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/docs/classes/_useragentapplication_.useragentapplication.html) | [Support](README.md#community-help-and-support) | [Samples](https://github.com/AzureAD/microsoft-authentication-library-for-js/wiki/Samples)
 | --- | --- | --- | --- | --- |
 
 
-The MSAL library preview for JavaScript enables client-side JavaScript web applications, running in a web browser, to authenticate users using [Azure AD](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-overview) work and school accounts (AAD), Microsoft personal accounts (MSA) and social identity providers like Facebook, Google, LinkedIn, Microsoft accounts, etc. through [Azure AD B2C](https://docs.microsoft.com/en-us/azure/active-directory-b2c/active-directory-b2c-overview#identity-providers) service. It also enables your app to get tokens to access [Microsoft Cloud](https://cloud.microsoft.com) services such as [Microsoft Graph](https://graph.microsoft.io).
+The MSAL library for JavaScript enables client-side JavaScript web applications, running in a web browser, to authenticate users using [Azure AD](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-overview) work and school accounts (AAD), Microsoft personal accounts (MSA) and social identity providers like Facebook, Google, LinkedIn, Microsoft accounts, etc. through [Azure AD B2C](https://docs.microsoft.com/en-us/azure/active-directory-b2c/active-directory-b2c-overview#identity-providers) service. It also enables your app to get tokens to access [Microsoft Cloud](https://cloud.microsoft.com) services such as [Microsoft Graph](https://graph.microsoft.io).
 
 [![Build Status](https://travis-ci.org/AzureAD/microsoft-authentication-library-for-js.png?branch=dev)](https://travis-ci.org/AzureAD/microsoft-authentication-library-for-js)[![npm version](https://img.shields.io/npm/v/msal.svg?style=flat)](https://www.npmjs.com/package/msal)[![npm version](https://img.shields.io/npm/dm/msal.svg)](https://nodei.co/npm/msal/)
 
@@ -22,36 +22,34 @@ Via CDN:
     <script src="https://secure.aadcdn.microsoftonline-p.com/lib/<version>/js/msal.js"></script>
     <script src="https://secure.aadcdn.microsoftonline-p.com/lib/<version>/js/msal.min.js"></script>
 
-Note that msal.js is built for ES5, therefore enabling support for Internet Explorer 11. If you want to target Internet Explorer, you'll need to add a reference to promises polyfill.
+Internet Explorer does not have native `Promise` support, and so you will need to include a polyfill for promises such as `bluebird`.
 
     <!-- IE support: add promises polyfill before msal.js  -->
     <script src="https://cdnjs.cloudflare.com/ajax/libs/bluebird/3.3.4/bluebird.min.js" class="pre"></script>
 
-See here for more details on [supported browsers](https://github.com/AzureAD/microsoft-authentication-library-for-js/wiki/FAQs#q4-what-browsers-is-msaljs-supported-on).
+See here for more details on [supported browsers and known compatability issues](https://github.com/AzureAD/microsoft-authentication-library-for-js/wiki/FAQs#q4-what-browsers-is-msaljs-supported-on).
 
 ## Usage
-The example snippets below show how to acquire a token for Microsoft Graph.
+The example below walks you through how to login a user and acquire a token to be used for Microsoft's Graph Api.
 
 #### Prerequisite
 
-Before using MSAL.js, [register an application in Azure AD](https://docs.microsoft.com/en-us/azure/active-directory/develop/quickstart-register-app) to get your clientID.
+Before using MSAL.js you will need to [register an application in Azure AD](https://docs.microsoft.com/en-us/azure/active-directory/develop/quickstart-register-app) to get a valid `clientId` for configuration, and to register the routes that your app will accept redirect traffic on.
 
 #### 1. Instantiate the UserAgentApplication
 
-Instantiate the UserAgentApplication with a minimal required configuration of clientId.
+`UserAgentApplication` can be configured with a variety of different options, detailed in our [Wiki](https://github.com/AzureAD/microsoft-authentication-library-for-js/wiki/MSAL.js-1.0.0-preview-api-release#configuration-options), but the only required parameter is `auth.clientId`.
 
-You must register a callback handler if you are using redirect flow(loginRedirect and acquireTokenRedirect). This callback function is called after the authentication request is completed either successfully or with a failure. This is not required for the popup flows since they return promises.
-
-UserAgentApplication has other optional parameters like redirectUri which can be assigned. Please refer to the [Wiki](https://github.com/AzureAD/microsoft-authentication-library-for-js/wiki/MSAL-basics#configuration-options) to see the full list and their default values.
+After instantiating your instance, if you plan on using a redirect flow (`loginRedirect` and `acquireTokenRedirect`), you must register a callback handlers  using `handleRedirectCallbacks(successHandler, errorHandler)`. The callback function is called after the authentication request is completed either successfully or with a failure. This is not required for the popup flows since they return promises
 
 ```JavaScript
-    var applicationConfig = {
+    var msalConfig = {
         auth: {
             clientId: 'your_client_id'
         }
     };
 
-    var msalInstance = new Msal.UserAgentApplication(applicationConfig);
+    var msalInstance = new Msal.UserAgentApplication(msalConfig);
 
     msalInstance.handleRedirectCallbacks(response => {
         // handle redirect response
@@ -63,11 +61,15 @@ UserAgentApplication has other optional parameters like redirectUri which can be
 
 #### 2. Login the user
 
-Your app must login the user with either loginPopup or the loginRedirect method to establish user context. When the login methods are called and the authentication of the user is completed by the Azure AD service, an [id token](https://docs.microsoft.com/en-us/azure/active-directory/develop/id-tokens) is returned which is used to identify the user with some basic information.
+Your app must login the user with either the `loginPopup` or the `loginRedirect` method to establish user context.
+
+When the login methods are called and the authentication of the user is completed by the Azure AD service, an [id token](https://docs.microsoft.com/en-us/azure/active-directory/develop/id-tokens) is returned which is used to identify the user with some basic information.
+
+When you login a user, you can request scopes up front if you know they will be needed, but it is not required, and login can be called with no params. It is best practice to only request scopes you need when you need them, a concept called dynamic consent. AAD will only allow you to get consent for 3 resources at a time, although you can request many scopes within a resource.
 
 ```JavaScript
    var loginRequest = {
-       scopes: ["user.read", "mail.send"] // optional
+       scopes: ["user.read", "mail.send"] // optional Array<string>
    };
 
     msalInstance.loginPopup(loginRequest)
@@ -79,43 +81,42 @@ Your app must login the user with either loginPopup or the loginRedirect method 
         });
 
 ```
-> Note: The scopes passed to the login method are optional. In this example, the graphScopes are passed in the login method to obtain consent upfront from the user for your app to access certain Graph API scopes. The idtoken returned here does not contain the scopes. In the next step, you can see how to get an access token which will contain the consented scopes.
 
 #### 3. Get an access token to call an API
 
-In MSAL, you can get access tokens for the APIs your app needs to call using the acquireTokenSilent method which makes a silent request(without prompting the user) to Azure AD to obtain an access token. The Azure AD service then returns an [access token](https://docs.microsoft.com/en-us/azure/active-directory/develop/access-tokens) containing the user consented scopes to allow your app to securely call the API.
+In MSAL, you can get access tokens for the APIs your app needs to call using the `acquireTokenSilent` method which makes a silent request(without prompting the user with UI) to Azure AD to obtain an access token. The Azure AD service then returns an [access token](https://docs.microsoft.com/en-us/azure/active-directory/develop/access-tokens) containing the user consented scopes to allow your app to securely call the API.
 
-If the silent token acquisition fails for some reasons such as an expired token or password change, you will need to invoke an interactive method to acquire tokens such as acquireTokenPopup or acquireTokenRedirect.
+If the `acquireTokenSilent` call fails with an error of type `InteractionRequiredAuthError` (requiring the user to consent to something), you will need to initiate an interactive request. You can use `acquireTokenRedirect` or `acquireTokenPopup`.
 
- ```JavaScript
+`acquireTokenSilent` will look for a valid token in the cache, and if it is close to expiring or does not exist, will automatically refresh it for you.
 
-    var tokenRequest = {
-        scopes: ["user.read", "mail.send"]
-    };
-
-    msalInstance.loginPopup(loginRequest)
-        .then(response => msalInstance.acquireTokenSilent(graphScopes)
-        .then(response {
-            // get access token from response
-            // response.accessToken
-        })
-        .catch(err => {
-            // if the error requires user consent
-            if (err.errorCode.indexOf("consent_required") !== -1 ||err.errorCode.indexOf("interaction_required") !== -1 ||err.errorCode.indexOf("login_required") !== -1 ) {
-                return msalnstance.acquireTokenPopup(tokenRequest)
-                    .then(response => {
-                        // get access token from response
-                        // response.accessToken
-                    })
-                    .catch(err => {
-                        // handle error
-                    })
-            } else {
-                // handle error
-            }
-        });
-
-
+```JavaScript
+    // if the user is already logged in you can acquire a token
+    if (msalInstance.getAccount()) {
+        var tokenRequest = {
+            scopes: ["user.read", "mail.send"]
+        };
+        msalInstance.acquireTokenSilent(tokenRequest)
+            .then(response => {
+                // get access token from response
+                // response.accessToken
+            })
+            .catch(err => {
+                // could also check if err instance of InteractionRequiredAuthError if you can import the class.
+                if (err.name === "InteractionRequiredAuthError") {
+                    return msalnstance.acquireTokenPopup(tokenRequest)
+                        .then(response => {
+                            // get access token from response
+                            // response.accessToken
+                        })
+                        .catch(err => {
+                            // handle error
+                        });
+                }
+            });
+    } else {
+        // user is not logged in, you will need to log them in to acquire a token
+    }
 ```
 
 #### 4. Use the token as a bearer in an HTTP request to call the Microsoft Graph or a Web API
--- a/README.md
+++ b/README.md
@@ -1,17 +1,15 @@
+
 Microsoft Authentication Library Preview for JavaScript (MSAL.js)
 =========================================================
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -38,74 +36,86 @@ The example snippets below show how to acquire a token for Microsoft Graph.
...
-            }, function (error) {
-                console.log(error);
-            });
+
+    var tokenRequest = {
+        scopes: ["user.read", "mail.send"]
+    };
+
+    msalInstance.loginPopup(loginRequest)
+        .then(response => msalInstance.acquireTokenSilent(graphScopes)
```
sameerag: This should be acquireTokenSilent(tokenRequest) right?

README.md
```diff
@@ -38,74 +36,86 @@ The example snippets below show how to acquire a token for Microsoft Graph.
...
-                updateUI();
-            }, function (error) {
-                console.log(error);
-            });
+
+    var tokenRequest = {
+        scopes: ["user.read", "mail.send"]
+    };
+
+    msalInstance.loginPopup(loginRequest)
```
sameerag: where is loginRequest defined?

README.md
```diff
@@ -38,74 +36,86 @@ The example snippets below show how to acquire a token for Microsoft Graph.
...
+            // get access token from response
+            // response.accessToken
         })
-    }, function (error) {
-        //login failure
-        console.log(error);
-    });
+        .catch(err => {
+            // if the error requires user consent
+            if (err.errorCode.indexOf("consent_required") !== -1 ||err.errorCode.indexOf("interaction_required") !== -1 ||err.errorCode.indexOf("login_required") !== -1 ) {
```
sameerag: as in the Quickstart can we have a wrapped up function. Also briefly summarize why these three errors are called out in the sample and how they can be handled. This is the most popular error we see for usage issues.

lib/msal-core/README.md
```diff
@@ -41,71 +38,84 @@ Before using MSAL.js, [register an application in Azure AD](https://docs.microso
...
 
-A callback function must be passed in for the redirect flows(loginRedirect and acquireTokenRedirect). This callback function is called after the authentication request is completed either successfully or with a failure. This is not required for the popup flows since they return promises.
+You must register a callback handler if you are using redirect flow(loginRedirect and acquireTokenRedirect). This callback function is called after the authentication request is completed either successfully or with a failure. This is not required for the popup flows since they return promises.
 
 UserAgentApplication has other optional parameters like redirectUri which can be assigned. Please refer to the [Wiki](https://github.com/AzureAD/microsoft-authentication-library-for-js/wiki/MSAL-basics#configuration-options) to see the full list and their default values.
 
 ```JavaScript
     var applicationConfig = {
-        clientID: 'your_client_id'
+        auth: {
```
sameerag: same comments as above, please change the name 'applicationConfig'

lib/msal-core/README.md
```diff
@@ -41,71 +38,84 @@ Before using MSAL.js, [register an application in Azure AD](https://docs.microso
...
-            }, function (error) {
-                console.log(error);
-            });
+
+    var tokenRequest = {
+        scopes: ["user.read", "mail.send"]
+    };
+
+    msalInstance.loginPopup(loginRequest)
+        .then(response => msalInstance.acquireTokenSilent(graphScopes)
```
sameerag: This should be acquireTokenSilent(tokenRequest)

lib/msal-core/README.md
```diff
@@ -41,71 +38,84 @@ Before using MSAL.js, [register an application in Azure AD](https://docs.microso
...
+            // get access token from response
+            // response.accessToken
         })
-    }, function (error) {
-        //login failure
-        console.log(error);
-    });
+        .catch(err => {
+            // if the error requires user consent
+            if (err.errorCode.indexOf("consent_required") !== -1 ||err.errorCode.indexOf("interaction_required") !== -1 ||err.errorCode.indexOf("login_required") !== -1 ) {
```
sameerag: Same comment as above on errors

README.md
```diff
@@ -38,74 +36,86 @@ The example snippets below show how to acquire a token for Microsoft Graph.
...
 
 #### 2. Login the user
 
 Your app must login the user with either loginPopup or the loginRedirect method to establish user context. When the login methods are called and the authentication of the user is completed by the Azure AD service, an [id token](https://docs.microsoft.com/en-us/azure/active-directory/develop/id-tokens) is returned which is used to identify the user with some basic information.
 
 ```JavaScript
-   var graphScopes = ["user.read", "mail.send"];
+   var loginRequest = {
+       scopes: ["user.read", "mail.send"] // optional
+   };
```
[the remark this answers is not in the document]
DarylThayil: what do you mean by this ?

README.md
```diff
@@ -38,74 +36,86 @@ The example snippets below show how to acquire a token for Microsoft Graph.
...
-A callback function must be passed in for the redirect flows(loginRedirect and acquireTokenRedirect). This callback function is called after the authentication request is completed either successfully or with a failure. This is not required for the popup flows since they return promises.
+You must register a callback handler if you are using redirect flow(loginRedirect and acquireTokenRedirect). This callback function is called after the authentication request is completed either successfully or with a failure. This is not required for the popup flows since they return promises.
 
 UserAgentApplication has other optional parameters like redirectUri which can be assigned. Please refer to the [Wiki](https://github.com/AzureAD/microsoft-authentication-library-for-js/wiki/MSAL-basics#configuration-options) to see the full list and their default values.
 
 ```JavaScript
     var applicationConfig = {
-        clientID: 'your_client_id'
+        auth: {
+            clientId: 'your_client_id'
```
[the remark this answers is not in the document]
DarylThayil: adding new link

README.md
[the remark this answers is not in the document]
sameerag: We throw errors if scopes are not sent as an array or an empty array is sent (I think). If Readme is to advocate the patterns, adding a note insisting the same would be a good addition IMHO/

README.md
DarylThayil: yes! good catch

README.md
DarylThayil: removing and adding comment, that if you are not calling with scopes, you don't have to pass an obj

README.md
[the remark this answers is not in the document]
pkanher617: just config is probably fine.

README.md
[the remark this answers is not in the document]
pkanher617: +1 just add in comments that it must be string[]

README.md
pkanher617: We don't need to use this. we can use error.name === "InteractionRequiredAuthError"

README.md
```diff
@@ -24,88 +22,117 @@ Via CDN:
...
+
+Please check our [roadmap](https://github.com/AzureAD/microsoft-authentication-library-for-js/wiki#roadmap) to see what we are workign on and what we are tracking next.
+
+## Oauth and the Implict Flow
+Msal implments the [Implicit Grant Flow](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-implicit-grant-flow), as defined by the Oauth 2.0 protocol and is [OpenID](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-protocols-oidc) compliant.
+
+Our goal is that the library abstracts enough of this away that you can get plug and play authentication, but it is important to know for users that need AAD functionality outside of simple auth, and also important to understand from a security perspective.  There is no client secret, and one less hop between client and server, and so tokens are naturally less secure.  This is mitigated by the use of tokens with a short time to live. We encourage you to read about the flows, and the spec, so that you have a strong understanding of the use case.
```
navyasric: It is not clear what simple auth is here. Can we rephrase as "  the library abstracts enough of the protocol away so that you can get plug and play authentication, but it is important to know and understand the implicit flow from a security perspective. " 

Elaborating the security aspect a little more.
" The implicit flow runs in the context of a web browser which cannot manage client secrets securely. It is optimized for single page apps and has one less hop between client and server so tokens are returned directly to the browser. These aspects make it naturally less secure. 

These security concerns are mitigated per standard practices such as- use of short lived tokens (and so no refresh tokens are returned), the library requiring a registered redirect URI for the app, library matching the request and response with a unique nonce and state parameter."

README.md
```diff
@@ -24,88 +22,117 @@ Via CDN:
...
+
+## Roadmap and What To Expect From This Library
+Msal support on Javascript is a collection of libraries. msal-core or just simply msal, is the framework agnostic we library. Once our 1.X+ is stabilized, we are going to bring our msal-angular library to 1.x parity.  We are currently no planning on supporting msal-angularjs for 1.x, because the feedback we have heard is that community has moved on, however we will provide some code patterns of how you can use msal-core in an angularjs app. After our current libraries are up to standards, we will begin balancing new feature requests, with new platforms such as `react`, server side `node.js`, `electron`, etc...
+
+Our goal is communicate extremely well with the community and to take their opinions into account. We would like to get to a monthly minor release schedule, with patches comming as often as needed.  The level of communication, planning, and granularity we want to get to will be a work in progress.
+
+Please check our [roadmap](https://github.com/AzureAD/microsoft-authentication-library-for-js/wiki#roadmap) to see what we are workign on and what we are tracking next.
+
+## Oauth and the Implict Flow
```
navyasric: OAuth 2.0 to be explicit (since there was OAuth 1.0)

README.md
```diff
@@ -24,88 +22,117 @@ Via CDN:
...
+Msal implments the [Implicit Grant Flow](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-implicit-grant-flow), as defined by the Oauth 2.0 protocol and is [OpenID](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-protocols-oidc) compliant.
+
+Our goal is that the library abstracts enough of this away that you can get plug and play authentication, but it is important to know for users that need AAD functionality outside of simple auth, and also important to understand from a security perspective.  There is no client secret, and one less hop between client and server, and so tokens are naturally less secure.  This is mitigated by the use of tokens with a short time to live. We encourage you to read about the flows, and the spec, so that you have a strong understanding of the use case.
+
+## Cache Storage
```
navyasric: I suggest we move the topics cache storage, oauth 2.0 implicit details, Roadmap sections after the Usage section. 
It is common in most Readmes to expect to see a Installation followed by Usage for those who want to get quickly started.

README.md
sameerag: "Working" instead of "workign" - typo

README.md
```diff
@@ -24,88 +22,117 @@ Via CDN:
...
+
+Our goal is communicate extremely well with the community and to take their opinions into account. We would like to get to a monthly minor release schedule, with patches comming as often as needed.  The level of communication, planning, and granularity we want to get to will be a work in progress.
+
+Please check our [roadmap](https://github.com/AzureAD/microsoft-authentication-library-for-js/wiki#roadmap) to see what we are workign on and what we are tracking next.
+
+## Oauth and the Implict Flow
+Msal implments the [Implicit Grant Flow](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-implicit-grant-flow), as defined by the Oauth 2.0 protocol and is [OpenID](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-protocols-oidc) compliant.
+
```
sameerag: Are we 'OpenID' compliant completely?

README.md
sameerag: Can we add hyperlinks to these and add more data as needed?

README.md
```diff
@@ -24,88 +22,117 @@ Via CDN:
...
 ```
 
 #### 2. Login the user
 
-Your app must login the user with either loginPopup or the loginRedirect method to establish user context. When the login methods are called and the authentication of the user is completed by the Azure AD service, an [id token](https://docs.microsoft.com/en-us/azure/active-directory/develop/id-tokens) is returned which is used to identify the user with some basic information.
+Your app must login the user with either the `loginPopup` or the `loginRedirect` method to establish user context.
+
+When the login methods are called and the authentication of the user is completed by the Azure AD service, an [id token](https://docs.microsoft.com/en-us/azure/active-directory/develop/id-tokens) is returned which is used to identify the user with some basic information.
+
+When you login a user, you can request scopes up front if you know they will be needed, but it is not required, and login can be called with no params. It is best practice to only request scopes you need when you need them, a concept called dynamic consent. AAD will only allow you to get consent for 3 resources at a time, although you can request many scopes within a resource.
```
sameerag: This text is a little confusing for usage. I am adding a detailed text here - may be you can summarize and add a link on scopes usage in login/acquire calls etc and expand it in the link:

When you login a user, and do not know what scopes you would need eventually, you can call the login() APIs without any scopes. For usecases where you may want the user to pre-consent for resource access later, you can pass the scopes during login. 
    - Please note that the passing of scopes at login allows you to request consent for the scopes, but the token returned will not contain the scopes. 
    - The user still would be calling acquireToken() for the 'token' with scopes. However this can be a non-interactive silent call if the user has preconsented the scopes in the previous login call.

It is usually best practice to have 'dynamic consent' app model. If your app model fits in 'dynamic consent' which is where you only request scopes you need when you need them, any number of acquireToken() calls can be made per need basis. Please expect interactive consent screen for those calls whose scopes are not preconsented.

AAD will only allow you to get consent for 3 resources at a time, although you can request many scopes within a resource. In this context:
    - When the user makes a login request, multiple resources and their corresponding scopes can be passed since the service issues an idToken preconsenting those scopes
    - However acquireToken calls are valid only for one resource/multiple scopes. If you have a usecase of accessing multiple resources, please split the acquireToken calls one per resource

README.md
```diff
@@ -24,88 +22,117 @@ Via CDN:
...
+In MSAL, you can get access tokens for the APIs your app needs to call using the `acquireTokenSilent` method which makes a silent request(without prompting the user with UI) to Azure AD to obtain an access token. The Azure AD service then returns an [access token](https://docs.microsoft.com/en-us/azure/active-directory/develop/access-tokens) containing the user consented scopes to allow your app to securely call the API.
 
-If the silent token acquisition fails for some reasons such as an expired token or password change, you will need to invoke an interactive method to acquire tokens such as acquireTokenPopup or acquireTokenRedirect.
+If the `acquireTokenSilent` call fails with an error of type `InteractionRequiredAuthError` (requiring the user to consent to something), you will need to initiate an interactive request. You can use `acquireTokenRedirect` or `acquireTokenPopup`.
```
sameerag: If the `acquireTokenSilent` call fails with an error of type `InteractionRequiredAuthError` (requiring the user to consent to something)

or
"if you choose to have an interactive accessToken call"

Also I have a question here on if we should mention loginPopup -> acquireTokenPopup and loginRedirect -> acquireTokenRedirect flows. Can they mix these today?

README.md
```diff
@@ -24,88 +22,117 @@ Via CDN:
...
 
- ```JavaScript
-    var graphScopes = ["user.read", "mail.send"];
+`acquireTokenSilent` will look for a valid token in the cache, and if it is close to expiring or does not exist, will automatically refresh it for you.
```
sameerag: @navyasric can you confirm if any more data is needed here regarding the 'tokenRenewalOffsetSeconds'?

README.md
sameerag: Also "for some reasons such as an expired token or password change" - password change needs to be mentioned IMHO.

lib/msal-core/README.md
```diff
@@ -2,18 +2,15 @@
...
 
 [![Build Status](https://travis-ci.org/AzureAD/microsoft-authentication-library-for-js.png?branch=dev)](https://travis-ci.org/AzureAD/microsoft-authentication-library-for-js)[![npm version](https://img.shields.io/npm/v/msal.svg?style=flat)](https://www.npmjs.com/package/msal)[![npm version](https://img.shields.io/npm/dm/msal.svg)](https://nodei.co/npm/msal/)
 
 
-## Important Note about the MSAL Preview
-The MSAL.js core library is suitable for use in a production environment. We provide the same production level support for this library as we do our current production libraries. During the preview we may make changes to the API, internal cache format, and other mechanisms of this library, which you will be required to take along with bug fixes or feature improvements. This may impact your application. For instance, a change to the cache format may impact your users, such as requiring them to sign in again. An API change may require you to update your code. When we provide the General Availability release we will require you to update to the General Availability version within six months, as applications written using a preview version of library may no longer work.
```
sameerag: Again, we may need some version of this until we go GA.

lib/msal-core/README.md
```diff
@@ -25,87 +22,117 @@ Via CDN:
...
     <script src="https://cdnjs.cloudflare.com/ajax/libs/bluebird/3.3.4/bluebird.min.js" class="pre"></script>
 
-See here for more details on [supported browsers](https://github.com/AzureAD/microsoft-authentication-library-for-js/wiki/FAQs#q4-what-browsers-is-msaljs-supported-on).
+See here for more details on [supported browsers and known compatability issues](https://github.com/AzureAD/microsoft-authentication-library-for-js/wiki/FAQs#q4-what-browsers-is-msaljs-supported-on).
+
+## Roadmap and What To Expect From This Library
+Msal support on Javascript is a collection of libraries. msal-core or just simply msal, is the framework agnostic we library. Once our 1.X+ is stabilized, we are going to bring our msal-angular library to 1.x parity.  We are currently no planning on supporting msal-angularjs for 1.x, because the feedback we have heard is that community has moved on, however we will provide some code patterns of how you can use msal-core in an angularjs app. After our current libraries are up to standards, we will begin balancing new feature requests, with new platforms such as `react`, server side `node.js`, `electron`, etc...
+
```
sameerag: looks like same text as above, same comments apply.

lib/msal-core/README.md
```diff
@@ -25,87 +22,117 @@ Via CDN:
...
+Our goal is that the library abstracts enough of this away that you can get plug and play authentication, but it is important to know for users that need AAD functionality outside of simple auth, and also important to understand from a security perspective.  There is no client secret, and one less hop between client and server, and so tokens are naturally less secure.  This is mitigated by the use of tokens with a short time to live. We encourage you to read about the flows, and the spec, so that you have a strong understanding of the use case.
+
+## Cache Storage
+
+We offer two methods of storage for Msal, `localStorage` and `sessionStorage`.  Our recommendation is to use `sessionStorage` because it is more secure in storing tokens that are acquired by your users, but `localStorage` will give you Single Sign On accross tabs and user sessions.  We encourge you to explore the options and make the best decision for your application.
```
sameerag: same comments as above

lib/msal-core/README.md
```diff
@@ -25,87 +22,117 @@ Via CDN:
...
+In MSAL, you can get access tokens for the APIs your app needs to call using the `acquireTokenSilent` method which makes a silent request(without prompting the user with UI) to Azure AD to obtain an access token. The Azure AD service then returns an [access token](https://docs.microsoft.com/en-us/azure/active-directory/develop/access-tokens) containing the user consented scopes to allow your app to securely call the API.
 
-If the silent token acquisition fails for some reasons such as an expired token or password change, you will need to invoke an interactive method to acquire tokens such as acquireTokenPopup or acquireTokenRedirect.
+If the `acquireTokenSilent` call fails with an error of type `InteractionRequiredAuthError` (requiring the user to consent to something), you will need to initiate an interactive request. You can use `acquireTokenRedirect` or `acquireTokenPopup`.
```
sameerag: please mention password change

lib/msal-core/README.md
```diff
@@ -25,87 +22,117 @@ Via CDN:
...
+                }
             });
-        })
-    }, function (error) {
-        //login failure
-        console.log(error);
-    });
+    } else {
+        // user is not logged in, you will need to log them in to acquire a token
+    }
```
sameerag: Add a link to how 'response' looks like at some place.

README.md
```diff
@@ -24,88 +22,117 @@ Via CDN:
...
     <script src="https://cdnjs.cloudflare.com/ajax/libs/bluebird/3.3.4/bluebird.min.js" class="pre"></script>
 
-See here for more details on [supported browsers](https://github.com/AzureAD/microsoft-authentication-library-for-js/wiki/FAQs#q4-what-browsers-is-msaljs-supported-on).
+See here for more details on [supported browsers and known compatability issues](https://github.com/AzureAD/microsoft-authentication-library-for-js/wiki/FAQs#q4-what-browsers-is-msaljs-supported-on).
+
+## Roadmap and What To Expect From This Library
+Msal support on Javascript is a collection of libraries. msal-core or just simply msal, is the framework agnostic we library. Once our 1.X+ is stabilized, we are going to bring our msal-angular library to 1.x parity.  We are currently no planning on supporting msal-angularjs for 1.x, because the feedback we have heard is that community has moved on, however we will provide some code patterns of how you can use msal-core in an angularjs app. After our current libraries are up to standards, we will begin balancing new feature requests, with new platforms such as `react`, server side `node.js`, `electron`, etc...
```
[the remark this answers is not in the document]
DarylThayil: updating, leaving node because i think it is closer than electron and in my view is at same level as react in terms of importance.

README.md
```diff
@@ -24,88 +22,117 @@ Via CDN:
...
+
+Our goal is communicate extremely well with the community and to take their opinions into account. We would like to get to a monthly minor release schedule, with patches comming as often as needed.  The level of communication, planning, and granularity we want to get to will be a work in progress.
```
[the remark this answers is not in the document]
DarylThayil: updating

README.md
DarylThayil: this is great updating

README.md
DarylThayil: updating good catch

README.md
DarylThayil: @navyasric  I thought about that, but some of the feedback we heard was that we should be more upfront with some of this information.  If you still think below usage is better I am find with that.

DarylThayil: @sameerag  do we have links for cache ?

README.md
```diff
@@ -1,18 +1,16 @@
...
 
 [![Build Status](https://travis-ci.org/AzureAD/microsoft-authentication-library-for-js.png?branch=dev)](https://travis-ci.org/AzureAD/microsoft-authentication-library-for-js)[![npm version](https://img.shields.io/npm/v/msal.svg?style=flat)](https://www.npmjs.com/package/msal)[![npm version](https://img.shields.io/npm/dm/msal.svg)](https://nodei.co/npm/msal/)
 
 
-## Important Note about the MSAL Preview
-The MSAL.js core library is suitable for use in a production environment. We provide the same production level support for this library as we do our current production libraries. During the preview we may make changes to the API, internal cache format, and other mechanisms of this library, which you will be required to take along with bug fixes or feature improvements. This may impact your application. For instance, a change to the cache format may impact your users, such as requiring them to sign in again. An API change may require you to update your code. When we provide the General Availability release we will require you to update to the General Availability version within six months, as applications written using a preview version of library may no longer work.
```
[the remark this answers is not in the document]
DarylThayil: adding this back fill we take GA

README.md
sameerag: no, we should document what we cache and where. I am not sure if there is a specific section except the 'localStorage' vs 'sessionStorage' setting. @navyasric can correct me if I an wrong.

README.md
DarylThayil: When you login a user, you can pass in scopes that the user can pre consent to on login, however this is not required. Please note that consenting to scopes on login, does not return an access_token for these scopes, but gives you the opportunity to obtain a token silently with these scopes passed in, with no further interaction from the user.  

It is best practice to only request scopes you need when you need them, a concept called dynamic consent. While this can create more interactive consent for users in your application, it also reduces drop-off from users that may be uneasy granting a large list of permissions for features they are not yet using.

AAD will only allow you to get consent for 3 resources at a time, although you can request many scopes within a resource.
When the user makes a login request, you can pass in multiple resources and their corresponding scopes because AAD issues an idToken pre consenting those scopes. However acquireToken calls are valid only for one resource / multiple scopes. If you need to access multiple resources, please make separate acquireToken calls per resource.

DarylThayil: incorporated feedback

README.md
DarylThayil: You can use `acquireTokenRedirect` or `acquireTokenPopup` to initiate interactive requests, although, it is best practice to only show interactive experiences if you are unable to obtain a token silently due to interaction required errors. If you are using an interactive token call, it must match the login method used in your application. (`loginPopup`=> `acquireTokenPopup`, `loginRedirect` => `acquireTokenRedirect`).

If the `acquireTokenSilent` call fails with an error of type `InteractionRequiredAuthError` you will need to initiate an interactive request.  This could happen for many reasons including scopes that have been revoked, expired tokens, or password changes.

`acquireTokenSilent` will look for a valid token in the cache, and if it is close to expiring or does not exist, will automatically try to refresh it for you.

DarylThayil: incorporating feedback.

lib/msal-core/README.md
DarylThayil: adding

README.md
navyasric: v2.0 endpoint is OIDC compliant.

README.md
navyasric: No section describing how and what we cache. Only info on how to configure cache. Do we really think developers should know internal details of how library caches? Probably an advanced wiki topic than on the Readme page.

navyasric: Yes having these info on Readme is being more visible and upfront. But I still think installation and usage should be near each other and the first things most developers want to see to get quickly started. Those who need further info will read further.

PR #674, opened 2019-05-01
Move state parameter from Configuration to Request

pkanher617:
This PR moves the state parameter from the Configuration object to the request object. This more closely follows the Open ID Connect protocol [here](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest). 

Also includes name change for cachedHash.


sameerag:
Nevermind, The error PR that is merged has some references to config.state, creating a new PR to close that miss asap.

PR #544, opened 2019-01-22
fix: clean acquireToken caching so it doesn't accrue through states

zamzowd:
Fix for #527 

The cache entries for the Acquire Token process aren't properly removed, resulting in a build-up of entries (with different state values) over time.

Update to `Storage.removeAcquireTokenEntries` so it removes all entries that are not In Progress, rather than only specified entries.

Update to `Storage.resetCacheItems` so it removes all the cache entries for the Acquire Token process.

Create private function in `UserAgentApplication.updateAcquireTokenCache` to set the Acquire Token cache values from the user and the authenticationRequest objects.  Update to Acquire Token processes to use that function, eliminating duplicate code.

Minor cleanup to `UserAgentApplication.saveTokenFromHash`.

Update tests to reflect cache-clearing responsibility getting moved to `Storage`.

Issue #527, opened 2019-01-06
Lots of old "msal." entires in local storage


dwatty:
## I'm submitting a...
- [ ] Regression (a behavior that used to work and stopped working in a new release)
- [X] Bug report 
- [ ] Performance issue
- [ ] Feature request
- [ ] Documentation issue or request
- [ ] Other... Please describe:

## Browser:
- [X] Chrome version  71.0.3578.98
- [ ] Firefox version XX
- [ ] IE version XX
- [ ] Edge version XX
- [ ] Safari version XX

## Library version
Library version: 0.2.3

## Current behavior
If the user is never logged out and they continue to use the webapp long enough to have the iFrame renew the token, a new entry appears to be added to local storage and no entries are ever being cleaned up.  In the event that the user stays logged in for several days, the local storage begins to really fill up with old entries.

My current localhost session has 171 entries for msal.acquireTokenUser and a new one is added every time acquireTokenSilent is called with an expired token and it makes a request for a new token.  Of those 171, only 5 of them have actual values - the rest have blank values.

I also see entries for msal.token.renew.status and msal.authority piling up in my local storage as well. 

I've run across issues which describe what I'm seeing, but based on the comments in those tickets this sounds like it would have already been resolved.

https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/327
https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/370

In reviewing the logout behavior, I noticed these entries aren't cleared from the local storage even when logout is called.  The resetCacheItems function in [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-core/src/Storage.ts) is only looking for two keys to clear, and it's not actually removing them in most cases, just emptying the value.

## Expected behavior
The msal library should be cleaning up the local storage and only keeping around what it needs.

abhisheksoni27:
We are using this library in an internal project (soon to be publicly released) and would appreciate if this issue is resolved. 

I'd be happy to help out, if needed.

freeman-g:
We are having this same issue, so we added a bit of cleanup code until this is fixed in the published MSAL package.

```
  Object.keys(window.localStorage).forEach(function (key) {
    if (key.startsWith('msal.acquireTokenUser') || key.startsWith('msal.token.renew')) {
      window.localStorage.removeItem(key)
    }
  })
```

We're running this code everytime we renew the tokens.

sameerag:
[...]

PR #675, opened 2019-05-01
Fix MSAL cache cleaning method

pkanher617:
This PR is a fix for issue #527 where cache keys are not being cleared out. 

A previous PR (#544) was submitted for this, but was failing some use cases. This fix works for manually tested use cases and passes all unit tests.

### DIFF
--- a/lib/msal-core/src/Storage.ts
+++ b/lib/msal-core/src/Storage.ts
@@ -101,6 +101,8 @@ export class Storage {// Singleton
                         if (state && !this.tokenRenewalInProgress(state)) {
                             this.removeItem(key);
                             this.removeItem(Constants.renewStatus + state);
+                            this.removeItem(Constants.stateLogin);
+                            this.removeItem(Constants.stateAcquireToken);
                             this.setItemCookie(key, "", -1);
                         }
                     }
--- a/lib/msal-core/src/Storage.ts
+++ b/lib/msal-core/src/Storage.ts
@@ -98,7 +98,7 @@ export class Storage {// Singleton
                 if (storage.hasOwnProperty(key)) {
                     if (key.indexOf(CacheKeys.AUTHORITY) !== -1 || key.indexOf(CacheKeys.ACQUIRE_TOKEN_USER) !== 1) {
                         const state = key.split(Constants.resourceDelimiter)[1];
-                        if (!this.tokenRenewalInProgress(state)) {
+                        if (state && !this.tokenRenewalInProgress(state)) {
                             this.removeItem(key);
                             this.removeItem(Constants.renewStatus + state);
                             this.setItemCookie(key, "", -1);
--- a/lib/msal-core/src/Storage.ts
+++ b/lib/msal-core/src/Storage.ts
@@ -90,18 +90,31 @@ export class Storage {// Singleton
         return results;
     }
 
-    removeAcquireTokenEntries(authorityKey: string, acquireTokenAccountKey: string): void {
+    removeAcquireTokenEntries(): void {
         const storage = window[this.cacheLocation];
         if (storage) {
             let key: string;
             for (key in storage) {
                 if (storage.hasOwnProperty(key)) {
-                    if ((authorityKey !== "" && key.indexOf(authorityKey) > -1) || (acquireTokenAccountKey !== "" && key.indexOf(acquireTokenAccountKey) > -1)) {
-                        this.removeItem(key);
+                    if (key.indexOf(CacheKeys.AUTHORITY) !== -1 || key.indexOf(CacheKeys.ACQUIRE_TOKEN_USER) !== 1) {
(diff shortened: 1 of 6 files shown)


lib/msal-core/src/Storage.ts
```diff
@@ -90,18 +90,31 @@ export class Storage {// Singleton
...
+    removeAcquireTokenEntries(): void {
         const storage = window[this.cacheLocation];
         if (storage) {
             let key: string;
             for (key in storage) {
                 if (storage.hasOwnProperty(key)) {
-                    if ((authorityKey !== "" && key.indexOf(authorityKey) > -1) || (acquireTokenAccountKey !== "" && key.indexOf(acquireTokenAccountKey) > -1)) {
-                        this.removeItem(key);
+                    if (key.indexOf(CacheKeys.AUTHORITY) !== -1 || key.indexOf(CacheKeys.ACQUIRE_TOKEN_USER) !== 1) {
+                        const state = key.split(Constants.resourceDelimiter)[1];
```
DarylThayil: any case where the array does not have element [1]

lib/msal-core/src/Storage.ts
```diff
@@ -90,18 +90,31 @@ export class Storage {// Singleton
...
+                        if (state && !this.tokenRenewalInProgress(state)) {
```
DarylThayil: what happens if a tokenRenewalInProgress?

DarylThayil:
few questions / comments but looks good to me

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -430,12 +430,7 @@ export class UserAgentApplication {
         request.state
       );
 
-      // Cache nonce
-      this.cacheStorage.setItem(Constants.nonceIdToken, serverAuthenticationRequest.nonce, this.inCookie);
-
-      // Cache account and authority
-      this.setAccountCache(account, serverAuthenticationRequest.state);
-      this.setAuthorityCache(serverAuthenticationRequest.state, acquireTokenAuthority.CanonicalAuthority);
+      this.updateAcquireTokenCacheEntries(serverAuthenticationRequest, account);
```
sameerag: @pkanher617 As discussed you may want to combine the login calls caching code for "authority" and "nonce" also into this API. Please factor that login calls do not have "acquireTokenAccountKey".

lib/msal-core/src/Storage.ts
pkanher617: We always store CacheKeys.Authority or CacheKeys.AcquireToken with the resource delim and state. I can add a check though.

lib/msal-core/src/Storage.ts
pkanher617: Then we don't remove the entries.

lib/msal-core/src/UserAgentApplication.ts
pkanher617: added to updateCacheEntries. Please check new code.

lib/msal-core/src/Storage.ts
pkanher617: updated

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -674,13 +658,7 @@ export class UserAgentApplication {
         // populate QueryParameters (sid/login_hint/domain_hint) and any other extraQueryParameters set by the developer
         serverAuthenticationRequest = this.populateQueryParams(account, request, serverAuthenticationRequest);
 
-        // Cache nonce
-        this.cacheStorage.setItem(Constants.nonceIdToken, serverAuthenticationRequest.nonce, this.inCookie);
-        serverAuthenticationRequest.state = serverAuthenticationRequest.state;
-
-        // Cache account and authority
-        this.setAccountCache(account, serverAuthenticationRequest.state);
-        this.setAuthorityCache(serverAuthenticationRequest.state, acquireTokenAuthority.CanonicalAuthority);
```
manoj-rath: it's canonicalAuthority here, however in the updateCacheEntries is using authority

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -430,12 +419,7 @@ export class UserAgentApplication {
         request.state
       );
 
-      // Cache nonce
-      this.cacheStorage.setItem(Constants.nonceIdToken, serverAuthenticationRequest.nonce, this.inCookie);
-
-      // Cache account and authority
-      this.setAccountCache(account, serverAuthenticationRequest.state);
-      this.setAuthorityCache(serverAuthenticationRequest.state, acquireTokenAuthority.CanonicalAuthority);
```
manoj-rath: CanonicalAuthority - same as the other comment

DarylThayil:
lgtm approved

PR #676, opened 2019-05-02
First stab to make the wait time configurable

sameerag:
Today we are hardcoding the wait time of 500ms when we load an iFrame. Added this to make it configurable.

Future rollout:
To make this more accurate by dynamically checking if the iFrame is loaded and further improvements if possible.

### DIFF
--- a/lib/msal-core/src/Configuration.ts
+++ b/lib/msal-core/src/Configuration.ts
@@ -12,7 +12,7 @@ export type CacheLocation = "localStorage" | "sessionStorage";
  */
 const FRAME_TIMEOUT = 6000;
 const OFFSET = 300;
-const LOAD_FRAME_WAIT_TIME = 500;
+const NAVIGATE_FRAME_WAIT = 500;
 
 
 /**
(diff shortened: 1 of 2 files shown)


lib/msal-core/src/Configuration.ts
```diff
@@ -63,6 +65,7 @@ export type SystemOptions = {
   logger?: Logger;
   loadFrameTimeout?: number;
   tokenRenewalOffsetSeconds?: number;
+  waitTimeToLoadFrame?: number;
```
pkanher617: So the naming is a little confusing here with load, frame, and time. I think the usage here is a little different.

loadFrameTimeout is actually the time it will wait to load the frame. This parameter is being used as the timeout for frame navigation, so you could call it silentAuthTimeout or something. I think the name of the function loadFrame is probably misleading as well, we could change it to navigateFrame instead.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1017,7 +1017,7 @@ export class UserAgentApplication {
         this.logger.infoPii("Frame Name : " + frameName + " Navigated to: " + urlNavigate);
       }
     },
-    500);
+    this.config.system.waitTimeToLoadFrame);
```
pkanher617: See comments above in Configuration.ts

lib/msal-core/src/Configuration.ts
sameerag: 👍 I feel the confusion too. Changed it now.

Issue #689, opened 2019-05-04
redirectCallback method is not called if navigateToLoginRequestUrl is false


kosjanne:
## I'm submitting a...

<pre><code>
[x ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Other... Please describe:
</code></pre>

## Browser:
- [x ] Chrome version 73
- [ ] Firefox version XX
- [ ] IE version XX
- [ ] Edge version XX
- [ ] Safari version XX

## Library version

<pre><code>
Library version: 1.0.0

## Current behavior

If I set navigateToLoginRequestUrl to false, redirectCallback is not called.

## Expected behavior

I would expect that my redirectCallback method gets called even if navigateToLoginRequestUrl is false, as it was working still in 0.2.4 version.

## Minimal reproduction of the problem with instructions

Create user agent as follows:
const msalConfig: Msal.Configuration = {
      auth: {
        ...
        navigateToLoginRequestUrl: false
      }};
userAgentApplication = new Msal.UserAgentApplication(msalConfig);
userAgentApplication.handleRedirectCallback(authCallback);

authCallback method will never get called after we return from userAgentApplication.loginRedirect(…);
</code></pre>

pkanher617:
@kosjanne Thank you for the submission. I was able to reproduce this, looking at the fix now.

pkanher617:
@kosjanne I have introduced a PR to fix this. Feel free to pull the branch and test.

PR #696, opened 2019-05-08
Fix: navigateToLoginRequestUrl set to false does not call callbacks

pkanher617:
This PR is a fix for issue #689. This was a bug introduced with changes to the API.

We now cache the hash briefly before processing the callback in handleRedirectCallback.

### DIFF
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -1383,7 +1383,8 @@ export class UserAgentApplication {
         window.location.hash = "";
       }
       if (!this.redirectCallbacksSet) {
-        // We reached this point too early, return and come back later
+        // We reached this point too early - cache hash, return and process in handleRedirectCallbacks
+        self.cacheStorage.setItem(Constants.urlHash, hash);
         return;
       }
     }


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1383,7 +1383,8 @@ export class UserAgentApplication {
         window.location.hash = "";
       }
       if (!this.redirectCallbacksSet) {
-        // We reached this point too early, return and come back later
+        // We reached this point too early - cache hash, return and process in handleRedirectCallbacks
+        self.cacheStorage.setItem(Constants.urlHash, hash);
```
sameerag: Looks good. Can we add unit test case/s  toggling navigateToLoginRequestUrl ?

DarylThayil:
lgtm, same comment on unit tests

pkanher617:
Due to the 'self' references in our code, it is a somewhat extended effort to add these unit tests currently. This should become easier once we refactor our code.

bcary:
I just found this issue as well performing our upgrade to the new version, thank you for fixing this!  When can we expect a minor release containing the fix?

sameerag:
@bcary Yes. Sometime early next week there will be a patch release.

Issue #691, opened 2019-05-05
ClientAuthError: Error: could not resolve endpoints. Please check network and try again when not supplying authenticationParameters


Tovli:
## I'm submitting a...

<pre><code>
[x] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Other... Please describe:
</code></pre>

## Browser:
- [ ] Chrome version XX
- [ ] Firefox version XX
- [ ] IE version XX
- [ ] Edge version XX
- [ ] Safari version XX

## Library version

<pre><code>
Library version: 1.0.0

## Current behavior

When using loginPopup without an AuthenticationParameters parameter then an exception is thrown due to undefined request parameter

## Expected behavior

The loginPopup window should be displayed

## Minimal reproduction of the problem with instructions

Login using this code:
this.app.loginPopup().then(function () {
      return this.app.acquireTokenSilent();
    }).catch(function (error) { return Promise.reject(error) }

    );

</code></pre>

sameerag:
@Tovli Thanks for raising this. loginPopup() should have been working without any request. Will reach out once I debug this.

sameerag:
@Tovli  I just tested this and it works as expected. loginPopup() works without a parameter. From the error you posted I do see that this could be probably the endPoint Verification that is failing. Can you please send us the authority that you are using to make this call?

"Error: could not resolve endpoints." is seen only when we make a call to the end point before sending a real request to the service to ensure that it is a reachable and valid. A call is made to [Open-Id well known configuration](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-protocols-oidc#fetch-the-openid-connect-metadata-document) and any failure in this path causes the above error.

Tovli:
Hi man,
you are right, I forgot to add my config.
I'm sending authority : https://login.microsoftonline.com/tfp/<tenant_name>.
onmicrosoft.com/<SignInSignUp_User_Flow_Name>

And the ClientId of course.
Sorry for the disinformation
Thanks for checking this

Now I send the same authority in the loginRequest as well and it works as
expected (Would change that later on of course :) )

Tovli:
so this is the expected behavior?

sameerag:
[...]

PR #698, opened 2019-05-08
Fix cases where request is undefined for user passed 'state' checks

sameerag:
Fix cases where request is undefined for user passed 'state' checks before passing it ServerRequestParameters

### DIFF
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -302,11 +302,7 @@ export class UserAgentApplication {
 
     // Creates navigate url; saves value in cache; redirect user to AAD
     if (this.loginInProgress) {
-      let reqState;
-      if (request) {
-        reqState = request.state;
-      }
-      this.redirectErrorHandler(ClientAuthError.createLoginInProgressError(), buildResponseStateOnly(reqState));
+      this.redirectErrorHandler(ClientAuthError.createLoginInProgressError(), buildResponseStateOnly(request && request.state));
       return;
     }
 
@@ -371,12 +367,6 @@ export class UserAgentApplication {
     // Track login in progress
     this.loginInProgress = true;
 
-    // request can be optional for login calls, hence check for the same
-    let reqState: string;
-    if (request) {
-      reqState = request.state;
-    }
-
     this.authorityInstance.resolveEndpointsAsync().then(() => {
 
       // create the Request to be sent to the Server
@@ -385,7 +375,7 @@ export class UserAgentApplication {
         this.clientId, scopes,
         ResponseTypes.id_token,
         this.getRedirectUri(),
-        reqState
+        request && request.state
       );
 
       // populate QueryParameters (sid/login_hint/domain_hint) and any other extraQueryParameters set by the developer
@@ -408,7 +398,7 @@ export class UserAgentApplication {
       this.promptUser(urlNavigate);
     }).catch((err) => {
       this.logger.warning("could not resolve endpoints");
-      this.redirectErrorHandler(ClientAuthError.createEndpointResolutionError(err.toString), buildResponseStateOnly(reqState));
+      this.redirectErrorHandler(ClientAuthError.createEndpointResolutionError(err.toString), buildResponseStateOnly(request && request.state));
     });
   }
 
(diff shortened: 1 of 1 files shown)


sameerag:
#691

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -379,7 +385,7 @@ export class UserAgentApplication {
         this.clientId, scopes,
         ResponseTypes.id_token,
         this.getRedirectUri(),
-        request.state
+        reqState
```
DarylThayil: reqState could still be undefined right?

DarylThayil: Oh i see what you are trying to guard against, could do this

```javascript
request && request.state
```

sameerag: It can but we cover that use case if it isn't defined in ServerRequestParameters. The issue is when we are referencing undefined.value

DarylThayil: instead of the if block, the inline && will be safe for this.

sameerag: I would like to stick to one style, @pkanher617 made some changes already with the above code. Let me redo all the checks in one pattern and send out for the review, once I test it.

DarylThayil: okay consistency is fine, I just think the if check and set adds some mental overhead, and inline is a little more expressive, basically saying, we aren't sure this is set but if it is, grab state

pkanher617: state can be undefined, but if request is undefined we get the error.

pkanher617: so we don't want the request.state check in the if condition

pkanher617: In the end it doesn't really matter, either way is functionally the same.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -371,6 +371,12 @@ export class UserAgentApplication {
     // Track login in progress
     this.loginInProgress = true;
 
+    // request can be optional for login calls, hence check for the same
+    let reqState: string;
+    if (request) {
```
AE-MS: Should the function parameter be declared as
`request?: AuthenticationParameters`
instead to indicate that it is valid for it to be undefined?

pkanher617: @AE-MS Thanks for reviewing! It is true that making the request parameter optional would allow us to pass it as undefined, which is what we have exposed in the loginRedirect function. However, that doesn't solve the issue of trying to retrieve state from a possibly null request object, which is what the code here does. We will try to find a more elegant solution for this when we perform some refactoring in the future.

sameerag: @AE-MS Updated the login helper functions too to have optional parameters, to clear any confusion extending what @pkanher617 mentioned - "request" is optional only for login() functions but not for acquireToken() calls.

DarylThayil:
lgtm

mmarshad:
I'm having the same issue today with following error
Error in logging in:  ClientAuthError: endpoints_resolution_error: Error: could not resolve endpoints. Please check network and try again. Detail: ClientConfigurationError: untrusted_authority: The provided authority is not a trusted authority. Please include this authority in the knownAuthorities config parameter.
    at t [as constructor] (https://xxxx/static/js/2.8c752529.chunk.js:2:136326)
    at new t (https://xxxxx/static/js/2.8c752529.chunk.js:2:103927)
    at Function.t.createEndpointDiscoveryIncompleteError (https://xxxxx/static/js/2.8c752529.chunk.js:2:104619)
    at Function.<anonymous> (https://xxxxx/static/js/2.8c752529.chunk.js:2:1872752)
    at https://xxxxx/static/js/2.8c752529.chunk.js:2:62868
    at Object.throw (https://xxxx/static/js/2.8c752529.chunk.js:2:62973)
    at s (https://etdcorestrdevne.z16.web.core.windows.net/static/js/2.8c752529.chunk.js:2:61762)

It was working until yesterday but today when we deploy the build to server, starting getting above error. from localhost its still working fine.

PR #700, opened 2019-05-09
add a missed sid check when adding login_hint

sameerag:
While adding login_hint in addHintParameters, need to check if sid is already populated. Adding the check.

### DIFF
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -672,7 +672,7 @@ export class UserAgentApplication {
       }
 
       // If no session exists, prompt the user to login.
-      if (!account && !!(request.sid  || request.loginHint)) {
+      if (!account && !(request.sid  || request.loginHint)) {
         this.logger.info("User login is required");
         return reject(ClientAuthError.createUserLoginRequiredError());
       }
@@ -895,7 +895,7 @@ export class UserAgentApplication {
       const adalIdToken = this.cacheStorage.getItem(Constants.adalIdToken);
 
       //if there is no account logged in and no login_hint/sid is passed in the request
-      if (!account && !!(request.sid  || request.loginHint) && Utils.isEmpty(adalIdToken) ) {
+      if (!account && !(request.sid  || request.loginHint) && Utils.isEmpty(adalIdToken) ) {
         this.logger.info("User login is required");
         return reject(ClientAuthError.createUserLoginRequiredError());
       }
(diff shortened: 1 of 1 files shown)


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1133,7 +1133,7 @@ export class UserAgentApplication {
       // login_hint
       else {
         // login_hint is account.userName
-        if (!qParams[SSOTypes.LOGIN_HINT]  && account.userName && !Utils.isEmpty(account.userName)) {
+        if (!qParams[SSOTypes.SID] && !qParams[SSOTypes.LOGIN_HINT] && account.userName && !Utils.isEmpty(account.userName)) {
```
DarylThayil: this looks find, I wonder if factoring this into variable will add to readability

```javascript
const thingWeAreCheckingFor = predicate && predicate && predicate; 
if(!thingWeAreCheckingFor) {  ... }
```

sameerag: no ... so this is a test branch, hence I haven't yet sent it for a review. This is for the customer to test the change. Any style that simplifies readability, can always be incorporated.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1122,18 +1122,18 @@ export class UserAgentApplication {
...
         }
       }
       // login_hint
       else {
         // login_hint is account.userName
-        if (!qParams[SSOTypes.LOGIN_HINT]  && account.userName && !Utils.isEmpty(account.userName)) {
+        if (!qParams[SSOTypes.LOGIN_HINT] && account.userName && !Utils.isEmpty(account.userName)) {
           qParams = Utils.addSSOParameter(SSOTypes.LOGIN_HINT, account.userName, qParams);
         }
       }
```
pkanher617: Should we be throwing an error here if they pass both? looks like we silently remove or don't add a parameter if both are present without saying anything to the caller.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1122,18 +1122,18 @@ export class UserAgentApplication {
...
-        if (!qParams[SSOTypes.SID]  && !qParams[SSOTypes.LOGIN_HINT]) {
+        if (!qParams[SSOTypes.LOGIN_HINT]) {
           qParams = Utils.addSSOParameter(SSOTypes.SID, account.sid, qParams);
         }
       }
       // login_hint
       else {
         // login_hint is account.userName
-        if (!qParams[SSOTypes.LOGIN_HINT]  && account.userName && !Utils.isEmpty(account.userName)) {
+        if (!qParams[SSOTypes.LOGIN_HINT] && account.userName && !Utils.isEmpty(account.userName)) {
```
pkanher617: You can simplify this by moving the '!qParams[SSOTypes.LOGIN_HINT]' check to line 1126.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1122,18 +1122,18 @@ export class UserAgentApplication {
     const account: Account = accountObj || this.getAccount();
 
     // This is a final check for all queryParams added so far; preference order: sid > login_hint
-    // sid cannot be passed along with login_hint, hence we check both are not populated yet in queryParameters so far
-    if (account) {
+    // sid cannot be passed along with login_hint or domain_hint, hence we check both are not populated yet in queryParameters so far
```
pkanher617: The comment says sid cannot be passed along with login_hint or domain_hint, but where do we actually check this and remove the hint if sid is available?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1122,18 +1122,18 @@ export class UserAgentApplication {
...
+    if (account && !qParams[SSOTypes.SID]) {
       // sid
       if (account.sid && serverReqParams.promptValue === PromptState.NONE) {
-        if (!qParams[SSOTypes.SID]  && !qParams[SSOTypes.LOGIN_HINT]) {
+        if (!qParams[SSOTypes.LOGIN_HINT]) {
```
pkanher617: You can simplify this by moving the '!qParams[SSOTypes.LOGIN_HINT]' check to line 1126.

lib/msal-core/src/UserAgentApplication.ts
sameerag: That check is done at a different place in populateQueryParams.

sameerag: This is only fall back when the user does not pass anything, that is why we check if it is already populated.

lib/msal-core/src/UserAgentApplication.ts
sameerag: +1. I am also checking if we can optimize more of hints code to simplify all checks we make regarding hints.

lib/msal-core/src/UserAgentApplication.ts
sameerag: +1

DarylThayil:
Is this still a test branch? or is it good to be added to dev?

sameerag:
@DarylThayil go ahead and finish the review. This is good to be added to dev now. I incorporated your review comments.

PR #707, opened 2019-05-14
add user cancelled error handling for acquireTokenPopup

sameerag:


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -775,7 +775,7 @@ export class UserAgentApplication {
 
     const pollTimer = window.setInterval(() => {
       // If popup closed or login in progress, cancel login
-      if (popupWindow && popupWindow.closed && instance.loginInProgress) {
+      if (popupWindow && popupWindow.closed && (instance.loginInProgress || instance.acquireTokenInProgress)) {
```
DarylThayil: can we add a test for this?

pkanher617: +1

sameerag: I am tracking testing this in a VSTS item. I did not find an easy way to stub this. Will follow up this week.

PR #712, opened 2019-05-15
Migrating to base64.js instead of window.atob for base64 encode/decode

sameerag:
Will add more tests later in e2e or UT

### DIFF
--- a/lib/msal-core/package.json
+++ b/lib/msal-core/package.json
@@ -49,7 +49,6 @@
     "ajv": "6.6.1",
     "awesome-typescript-loader": "^3.2.3",
     "babel-polyfill": "^6.26.0",
-    "babel-preset-env": "^1.7.0",
     "chai": "^4.2.0",
     "fork-ts-checker-webpack-plugin": "^0.5.1",
     "handlebars": "^4.1.2",
--- a/lib/msal-core/package.json
+++ b/lib/msal-core/package.json
@@ -36,7 +36,7 @@
     "prepack": "npm run build"
   },
   "dependencies": {
-    "atob": "~2.1.1",
+    "js-base64": "^2.5.1",
     "tslib": "^1.9.3"
   },
   "devDependencies": {
@@ -45,9 +45,11 @@
     "@types/mocha": "^5.2.6",
     "@types/node": "^11.11.5",
     "@types/sinon": "^7.0.11",
+    "@types/js-base64": "^2.3.1",
     "ajv": "6.6.1",
     "awesome-typescript-loader": "^3.2.3",
     "babel-polyfill": "^6.26.0",
+    "babel-preset-env": "^1.7.0",
     "chai": "^4.2.0",
     "fork-ts-checker-webpack-plugin": "^0.5.1",
     "handlebars": "^4.1.2",
--- a/lib/msal-core/src/Utils.ts
+++ b/lib/msal-core/src/Utils.ts
@@ -231,12 +231,7 @@ export class Utils {
    */
   static base64EncodeStringUrlSafe(input: string): string {
     // html5 should support atob function for decoding
-    if (Base64) {
-      return Base64.encode(input);
-    }
-    else {
-      return this.encode(input);
-    }
+    return Base64.encode(input);
   }
 
   /**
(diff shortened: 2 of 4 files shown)


lib/msal-core/package.json
```diff
@@ -45,9 +45,11 @@
     "@types/mocha": "^5.2.6",
     "@types/node": "^11.11.5",
     "@types/sinon": "^7.0.11",
+    "@types/js-base64": "^2.3.1",
     "ajv": "6.6.1",
     "awesome-typescript-loader": "^3.2.3",
     "babel-polyfill": "^6.26.0",
+    "babel-preset-env": "^1.7.0",
```
DarylThayil: This should not be added right?

lib/msal-core/src/Utils.ts
```diff
@@ -231,8 +231,8 @@ export class Utils {
    */
   static base64EncodeStringUrlSafe(input: string): string {
     // html5 should support atob function for decoding
-    if (window.btoa) {
-      return window.btoa(input);
+    if (Base64) {
```
DarylThayil: When is Base64 not defined?

lib/msal-core/package.json
sameerag: This was stated as a requirement in their release page: "Note js-base64 itself is stand-alone so its package.json has no dependencies.  However, it is also tested on ES6 environment so "babel-preset-env": "^1.7.0" is on devDependencies."

lib/msal-core/src/Utils.ts
sameerag: This should be true always, I added this to retain the existing code structure where we fall back to custom encode decode if needed. I can remove the fall back if you think we need to optimize it now. There is much of legacy code in Utils.ts which we can slowly move away from.

DarylThayil: to me it's less optimization and more removing confusion.
If its always true, lets remove that check

lib/msal-core/package.json
DarylThayil: hmmm ... 
I think what it means is that they need this to be able to run their tests, not that we need it ?
Did our tests break without it ?

lib/msal-core/src/Utils.ts
sameerag: So always is true is because we are importing it from the package.json; makes sense as this will standardize all our base64 encode and decode. Made the changes.

lib/msal-core/package.json
sameerag: We are not testing this in E2E today, we are manually verifying. This will be in automation later. I included it for folks using ES6 to not have this issue. I am okay either way.

DarylThayil: I don’t believe you get anything for free including this. If it does solve a problem, it means Babel transpilation would need to be used somewhere. I think they Babel transpile their tests which is why they need it, I’ll look

lib/msal-core/src/Utils.ts
sameerag: Have added UT now. Please review.

lib/msal-core/test/Utils.spec.ts
```diff
@@ -14,4 +15,22 @@ describe("Utils.ts class", () => {
         console.log(Utils.replaceTenantPath("http://a.com/common/", "1234-56778"));
         console.log(Utils.replaceTenantPath("http://a.com/common", "1234-5678"));
     });
+
+    it("test Base64 encode decode", () => {
```
DarylThayil: I think the right thing to test here would be `Utils.base64EncodeStringUrlSafe`  instead of `Base64.encode`

sameerag: Well, this was intentional. Those tests are coming as a part of Utils.ts testing, base64DecodeStringUrlSafe is used for idToken and no one calls base64EncodeStringUrlSafe as far as I can see. I wanted to construct more meaningful inputs for these APIs that are realistic as a part of the Utils.ts UT suite, hence was testing only Base64.encode() and Base64.decode() in this specific case.

DarylThayil: I would say us testing another libs api is less useful than testing our own. We did talk offline, but if api is unused, we can delete. otherwise, if we leave it in, lets test that interface instead of Base64.<>

sameerag: Done, please review. Code removal will be done later in a patch, raised a VSTS for the same.

DarylThayil:
lgtm

PR #716, opened 2019-05-17
fix: use authority.EndSessionEndpoint for logout

verbedr:
The library was using a fixed logout url which will not work in all
azure ad cases. According to the oidc specifications the log out url
is published in the discovery document as end_session_endpoint.
This information is fetched and stored in the
TenantDiscoveryResponse.EndSessionEndpoint.

As the discovery property is only required if the OpenId Provider
supports session managment. We will fall back to the original
solution when the property is undefined, null or empty.

### DIFF
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -125,7 +125,7 @@ const resolveTokenOnlyIfOutOfIframe = (target: any, propertyKey: string, descrip
 
 /**
  * UserAgentApplication class
- * 
+ *
  * Object Instance that the developer can use to make loginXX OR acquireTokenXX functions
  */
 export class UserAgentApplication {
@@ -169,7 +169,7 @@ export class UserAgentApplication {
   /**
    * Method to manage the authority URL.
    *
-   * @returns {string} authority 
+   * @returns {string} authority
    */
   public get authority(): string {
     return this.authorityInstance.CanonicalAuthority;
@@ -200,7 +200,7 @@ export class UserAgentApplication {
    * If your application supports Accounts in any organizational directory and personal Microsoft accounts, replace "Enter_the_Tenant_Info_Here" value with common.
    * To restrict support to Personal Microsoft accounts only, replace "Enter_the_Tenant_Info_Here" value with consumers.
    *
-   * 
+   *
    * In Azure B2C, authority is of the form https://&lt;instance&gt;/tfp/&lt;tenant&gt;/&lt;policyName&gt;/
 
    * @param {@link (Configuration:type)} configuration object for the MSAL UserAgentApplication instance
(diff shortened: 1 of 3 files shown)


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1247,8 +1247,12 @@ export class UserAgentApplication {
     if (this.getPostLogoutRedirectUri()) {
       logout = "post_logout_redirect_uri=" + encodeURIComponent(this.getPostLogoutRedirectUri());
     }
-    const urlNavigate = this.authority + "oauth2/v2.0/logout?" + logout;
-    this.promptUser(urlNavigate);
+    this.authorityInstance.resolveEndpointsAsync().then(a => {
```
DarylThayil: can we use a more descriptive param name than a?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1247,8 +1247,12 @@ export class UserAgentApplication {
...
+        const urlNavigate = a.EndSessionEndpoint
```
DarylThayil: can we use string interpolation her instead of concatenating?

lib/msal-core/test/UserAgentApplication.spec.ts
```diff
@@ -902,12 +902,29 @@ describe("UserAgentApplication.ts Class", function () {
...
+        });
+        it("falls back to default url when there is no end_session_endpoint ", function (done) {
+            sinon.stub(msal.getAuthorityInstance(), "EndSessionEndpoint").reset();
+            sinon.stub(msal.getAuthorityInstance(), "EndSessionEndpoint").value("");
+            sinon.stub(window.location, "replace").callsFake(function (url) {
+                expect(url).to.include(DEFAULT_INSTANCE + TENANT + '/oauth2/v2.0/logout?');
+                done();
+            });
+            msal.logout();
+        });
```
DarylThayil: thanks for adding tests

DarylThayil:
added a few comments on code, @sameerag  is confirming the product should work this way.
Thanks for contributing!

lib/msal-core/src/UserAgentApplication.ts
verbedr: Done. 

I originally looked at the existing way of working. But I also like string interpolation more

lib/msal-core/src/UserAgentApplication.ts
verbedr: Of course I've used 'authority' as it is of type 'Authority'. We could change it to something in respect to the async method name (e.g. endpoints)?

DarylThayil:
@verbedr  Resolved my comments, waiting for sign off on functionality!

sameerag:
@verbedr Service confirmed the changes your introduced are aligned with the requirements. Merging this PR.

DarylThayil:
@verbedr  thanks so much for the contribution. Expect this released soon!

verbedr:
@DarylThayil thank you and your team, it was a very positive experience!

sameerag:
Glad to be of help!

PR #720, opened 2019-05-21
remove .only from utils.spec.ts test

sameerag:


jasonnutter:
LGTM! When we update the linting rules, we should add some of the rules from [eslint-plugin-mocha](https://www.npmjs.com/package/eslint-plugin-mocha), which includes a rule disallowing exclusive tests.

sameerag:
Yep! Miss on my part. :-(

PR #727, opened 2019-05-24
Add react sample app

jasonnutter:
First draft of the React sample app. Tested in Edge, Edge Dev, IE11, Chrome, Firefox. 

Notes:
- Due to the redirect issue noted below, there are a few changes to the functionality compared to the other samples: every time the page loads, it will check to see if there's a cached token with the `Read.Mail` scope, and if so, fetch and display messages. 
- I've included some basic error handling for when something goes wrong on the login page, or if the user cancels.

A few todos:

- [x] Make the UI a little nicer
- [x] Update sample README with content specific to our library
- [x] Test other browsers
- [x] Remove all of the CRA boilerplate
- [x] Error messages

Known issues:
- When using redirect flow, the application will get mounted twice when coming back from the login screen. This makes it impossible to make API requests or change state in the `msalApp.handleRedirectCallback` function.

CRA provides polyfills for fetch and promises, so I decided to use `async/await` for promise control flow. Can go over this with the team tomorrow.

reactsample1

reactsample2

Inkedreactsample3_LI


lib/msal-core/samples/react-sample-app/package.json
```diff
@@ -0,0 +1,39 @@
+{
+    "name": "react-sample-app",
+    "version": "0.1.0",
+    "private": true,
+    "dependencies": {
+        "msal": "^1.0.1",
```
jasonnutter: I think this should actually pull from the locally built artifact, instead of the published version.

lib/msal-core/samples/react-sample-app/src/App.js
```diff
@@ -0,0 +1,43 @@
...
+            <div>
+                <section>
+                    <p>
+                        Welcome to the Microsoft Authentication Library For
+                        Javascript - React Quickstart
+                    </p>
+                    {!this.props.account ? (
+                        <button onClick={this.props.signIn}>Sign In</button>
+                    ) : (
+                        <>
```
jasonnutter: Shorthand syntax for `React.Fragment`

lib/msal-core/samples/react-sample-app/src/AuthProvider.js
```diff
@@ -0,0 +1,173 @@
...
+const msalApp = new Msal.UserAgentApplication({
+    auth: {
+        clientId: "245e9392-c666-4d51-8f8a-bfd9e55b2456",
+        authority: "https://login.microsoftonline.com/common",
+        validateAuthority: true,
+        postLogoutRedirectUri: "http://localhost:3000"
+    },
+    cache: {
+        cacheLocation: "localStorage",
+        storeAuthStateInCookie: true
```
sameerag: Have the 'readme' add that this is mandatory only for IE/Edge. Essentially for any browser that has trusted zones enabled and the browser storage is cleared on redirect. Making this a part of the sample may generalize it for all the users which we are not supposed to.

lib/msal-core/samples/react-sample-app/src/AuthProvider.js
```diff
@@ -0,0 +1,173 @@
...
+
+const msalApp = new Msal.UserAgentApplication({
+    auth: {
+        clientId: "245e9392-c666-4d51-8f8a-bfd9e55b2456",
+        authority: "https://login.microsoftonline.com/common",
+        validateAuthority: true,
+        postLogoutRedirectUri: "http://localhost:3000"
+    },
+    cache: {
+        cacheLocation: "localStorage",
```
sameerag: The default setting is 'sessionStorage'. Can we stick to that? I know we are going off existing samples. Let us discuss this and decide what patterns we want to establish to the devs in our samples. @DarylThayil.

lib/msal-core/samples/react-sample-app/package.json
DarylThayil: +1, since we use this in development testing as well.

Maybe add a build script that reaches back and out of the sample to make sure the dist file exists?

lib/msal-core/samples/react-sample-app/src/AuthProvider.js
DarylThayil: Should we add as a comment in the code instead / as well?

lib/msal-core/samples/react-sample-app/src/AuthProvider.js
DarylThayil: Sounds good

DarylThayil:
Oh nice I did not know WIP blocked merge!

jasonnutter:
Adding "WIP" to the title isn't what's actually blocking the merge, I open this PR as a draft: https://github.blog/2019-02-14-introducing-draft-pull-requests/

lib/msal-core/samples/react-sample-app/README.md
```diff
@@ -0,0 +1,68 @@
+This project was bootstrapped with [Create React App](https://github.com/facebook/create-react-app).
```
DarylThayil: Should we add some msal / react specific things to the read me ?

lib/msal-core/samples/react-sample-app/src/App.js
```diff
@@ -0,0 +1,43 @@
+import React from "react";
+import "./App.css";
+
+import AuthProvider from "./AuthProvider";
+
+class App extends React.Component {
```
DarylThayil: Probably should define prop types so its clear we are passing these down to the app

lib/msal-core/samples/react-sample-app/src/App.js
```diff
@@ -0,0 +1,43 @@
...
+                                Sign Out
+                            </button>
+                            <button onClick={this.props.readMail}>
+                                Read Email
+                            </button>
+                        </>
+                    )}
+                </section>
+                <section>
+                    {this.props.account && (
```
DarylThayil: Should we add headers to the pre blocks? 
So its clear what the json data represents

lib/msal-core/samples/react-sample-app/src/App.js
```diff
@@ -0,0 +1,43 @@
...
+                        <pre>{JSON.stringify(this.props.account, null, 4)}</pre>
+                    )}
+                    {this.props.graphData && (
+                        <pre>
+                            {JSON.stringify(this.props.graphData, null, 4)}
```
DarylThayil: can we rename graphData to userMail or something more descriptive of the data?

lib/msal-core/samples/react-sample-app/src/App.js
```diff
@@ -0,0 +1,43 @@
...
+                        </pre>
+                    )}
+                </section>
+            </div>
+        );
+    }
+}
+
+export default AuthProvider(App);
```
DarylThayil: I wonder if Hocs are new enough to most that something like this might be more descriptive of what `AuthProvider` is 

```js
export default () => (
    <AuthProivder>
      <App /> 
     </AuthProvider>
);
```

lib/msal-core/samples/react-sample-app/src/AuthProvider.js
```diff
@@ -0,0 +1,173 @@
...
+                    account: msalApp.getAccount()
+                });
+            }
+        };
+
+        signOut = () => {
+            msalApp.logout();
+        };
+
+        readMail = async () =>
```
DarylThayil: I would still encourage breaking this out a little bit more

getAccessTokenForScopes(scopes: Array<String>)

and then the token requests and what not can live in the piece responsible 

```
await const { accessToken } =  this.props.getAccessTokenForScopes(["user.read", "mail.read"]);
await const userMail = this.props.fetchMailFromGraph(accessToken);
this.props.setMailData(userMail);
```

DarylThayil: maybe I am overcomplicating for a sample
If you think its overkill, maybe just leave a good comment that outlines that 

```js 
// In your application it would be a good idea to generalize the acquisition of tokens away from the fetching of data but for the purpose of this sample we are do them together. 
```

lib/msal-core/samples/react-sample-app/src/AuthProvider.js
```diff
@@ -0,0 +1,173 @@
...
+                    ).then(this.graphAPICallback);
+                } else {
+                    console.log("token type is:" + response.tokenType);
+                }
+            }
+        };
+
+        graphAPICallback = graphData => {
+            this.setState({
+                graphData
```
DarylThayil: same comment on renaming this to be more descriptive of the data.

lib/msal-core/samples/react-sample-app/src/AuthProvider.js
```diff
@@ -0,0 +1,173 @@
...
+const isIE = () => {
+    const ua = window.navigator.userAgent;
+    const msie = ua.indexOf("MSIE ");
+    const msie11 = ua.indexOf("Trident/");
+
+    return msie > 0 || msie11 > 0;
+};
+
+export default C =>
+    class AuthProvider extends Component {
```
DarylThayil: Thoughts on adding a prop that is `autoLogin`? 

maybe not because you would need an error page if they don't authenticate?

DarylThayil:
made some comments, but overall great start!

lib/msal-core/samples/react-sample-app/src/AuthProvider.js
jasonnutter: Yeah I think that is something we could add incrementally.

lib/msal-core/samples/react-sample-app/src/AuthProvider.js
```diff
@@ -0,0 +1,173 @@
...
+
+const requiresInteraction = errorMessage => {
+    if (!errorMessage || !errorMessage.length) {
+        return false;
+    }
+
+    return (
+        errorMessage.indexOf("consent_required") !== -1 ||
+        errorMessage.indexOf("interaction_required") !== -1 ||
+        errorMessage.indexOf("login_required") !== -1
```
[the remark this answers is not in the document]
DarylThayil: or something like 

`msal.errorRequiresInteraction(err: AuthError ): Boolean`

In my heart of hearts, I really want to expose a simple api that wraps all of these parts into plug and play for the 90% who just don't care and leave the building blocks for those who need them


`msal.login(config)` /// does the browser checks and all that for you
`msal.acquireToken(config)` // Does all the interaction require fallbacks for you, lets you pass in some config on it, but abstracts boiler plate code away.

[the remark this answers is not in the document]
jasonnutter: Or even a `requiresIteraction` property on the error itself.

lib/msal-core/samples/react-sample-app/src/AuthProvider.js
```diff
@@ -0,0 +1,173 @@
...
+    );
+};
+
+const isIE = () => {
+    const ua = window.navigator.userAgent;
+    const msie = ua.indexOf("MSIE ");
+    const msie11 = ua.indexOf("Trident/");
+
+    return msie > 0 || msie11 > 0;
```
jasonnutter: These should be `> -1` not `0`

lib/msal-core/samples/react-sample-app/src/AuthProvider.js
```diff
@@ -0,0 +1,173 @@
...
+        errorMessage.indexOf("consent_required") !== -1 ||
+        errorMessage.indexOf("interaction_required") !== -1 ||
+        errorMessage.indexOf("login_required") !== -1
+    );
+};
+
+const isIE = () => {
+    const ua = window.navigator.userAgent;
+    const msie = ua.indexOf("MSIE ");
+    const msie11 = ua.indexOf("Trident/");
```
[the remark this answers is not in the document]
DarylThayil: I had looked at this in our internal lib code, we should add polyfill to our code as well so we can use it

lib/msal-core/samples/react-sample-app/package.json
```diff
@@ -0,0 +1,39 @@
...
+            ">0.2%",
+            "not dead",
+            "not op_mini all",
+            "ie 11"
+        ],
+        "development": [
+            "last 1 chrome version",
+            "last 1 firefox version",
+            "last 1 safari version",
+            "ie 11"
```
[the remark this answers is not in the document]
DarylThayil: what does this package member get used for? Webpack?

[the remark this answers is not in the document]
jasonnutter: Yeah, `babel-preset-env` uses it to determine what features it needs to transpile.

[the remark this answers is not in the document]
jasonnutter: (looks like its also used in the CSS toolchain that CRA ships with)

lib/msal-core/samples/react-sample-app/src/App.js
jasonnutter: Since we're passing down props from the provider, I don't think this will work. I would have to use context, or do render props, neither of which I'm a fan of for this sample.

jasonnutter:
> One question I had, can we isolate the redirect refresh in 'react'?

@sameerag I'm not sure what you mean by isolate, but there is an issue with this sample related to the redirect and page loading more than once.

lib/msal-core/samples/react-sample-app/README.md
DarylThayil: still planning to make this change?

lib/msal-core/samples/react-sample-app/src/AuthProvider.js
```diff
@@ -0,0 +1,160 @@
...
+
+            this.state = {
+                account: null,
+                error: null,
+                emailMessages: null,
+                graphProfile: null
+            };
+        }
+
+        async acquireToken(request, redirect) {
```
DarylThayil: nitpick, feel free to ignore, but maybe instead of `AuthProvider.acquireToken` you expose `AuthProvider.acquireMailToken` and `AuthProvider.acquireMeToken`  That way the caller doesn't have to construct a request

DarylThayil: never mind I see that this is onlu called internally

lib/msal-core/samples/react-sample-app/src/AuthProvider.js
```diff
@@ -0,0 +1,160 @@
...
+                // due to consent or interaction required ONLY
+                if (requiresInteraction(error.errorCode)) {
+                    return redirect
+                        ? msalApp.acquireTokenRedirect(request)
+                        : msalApp.acquireTokenPopup(request);
+                }
+            });
+        }
+
+        async onSignIn(redirect) {
```
DarylThayil: how does redirect get passed in, the app looks like it just passes the handler to the onClick without binding an option

lib/msal-core/samples/react-sample-app/src/AuthProvider.js
```diff
@@ -0,0 +1,160 @@
...
+            if (redirect) {
+                return msalApp.loginRedirect(GRAPH_REQUESTS.LOGIN);
+            }
+
+            const loginResponse = await msalApp
+                .loginPopup(GRAPH_REQUESTS.LOGIN)
+                .catch(() => {
```
DarylThayil: do you want to define the err here and log / set error

```js
...
.catch(error=> {
    this.setState({
          error // or maybe construct a message from it 
    })
})
```

lib/msal-core/samples/react-sample-app/src/AuthProvider.js
```diff
@@ -0,0 +1,160 @@
...
+                    this.setState({
+                        error: "Unable to acquire access token."
+                    });
+                });
+
+            if (loginResponse) {
```
DarylThayil: what is the case that login response is falsey?

DarylThayil: Oh i see, its because you are catching and not returning essentially

lib/msal-core/samples/react-sample-app/src/AuthProvider.js
```diff
@@ -0,0 +1,160 @@
...
+                });
+            });
+
+            if (tokenResponse) {
+                return this.readMail(tokenResponse.accessToken);
+            }
+        }
+
+        async readMail(accessToken) {
+            const emailMessages = await fetchMsGraph(
```
DarylThayil: no catch?

lib/msal-core/samples/react-sample-app/src/AuthProvider.js
```diff
@@ -0,0 +1,160 @@
...
+            msalApp.logout();
+        }
+
+        async onRequestEmailToken() {
+            const tokenResponse = await this.acquireToken(
+                GRAPH_REQUESTS.EMAIL,
+                useRedirectFlow
+            ).catch(e => {
+                this.setState({
+                    error: "Unable to acquire access token for reading email."
```
DarylThayil: Do you want to "reset" error to null on new requests somewherE?

lib/msal-core/samples/react-sample-app/src/AuthProvider.js
```diff
@@ -0,0 +1,160 @@
...
+            );
+
+            this.setState({
+                emailMessages,
+                error: null
+            });
+        }
+
+        async componentDidMount() {
+            msalApp.handleRedirectCallback(error => {
```
DarylThayil: does the token actually come back here?

DarylThayil: if we are doing nothing maybe just set a n00p with a comment 

would be nice if we knew why it mounted multiple times

lib/msal-core/samples/react-sample-app/README.md
jasonnutter: Yeah, working on it now.

lib/msal-core/samples/react-sample-app/src/AuthProvider.js
jasonnutter: In the render function: 
```jsx
onSignIn={() => this.onSignIn(useRedirectFlow)}
```

lib/msal-core/samples/react-sample-app/src/AuthProvider.js
jasonnutter: For this error looks like I could do that yeah.

jasonnutter: (that isn't the case with all the error scenarios, which is why I wasn't doing so originally).

lib/msal-core/samples/react-sample-app/src/AuthProvider.js
jasonnutter: Yeah, I reset this in the other places when API calls are made.

lib/msal-core/samples/react-sample-app/src/AuthProvider.js
jasonnutter: Good catch.

lib/msal-core/samples/react-sample-app/src/AuthProvider.js
jasonnutter: The token does come back here, but because of the refresh that happens, we lose the ability to do anything it, such as making API calls. And yeah, I'm going to look into why the refresh happens.

jasonnutter: Also, I think the type definitions are wrong for this method, as it says that the signature is `(response, error)`, when it's really `(error, response)` (as you would expect).

jasonnutter: I'll PR a fix for that separately.

pkanher617:
Running into an issue when I run an npm install:
image

sameerag:
@jasonnutter  Seeing the same issue @pkanher617 mentioned above in my local download too!

jasonnutter:
@sameerag @pkanher617 Issue was caused by the lockfile, which has been removed from teh `react-sample-app` folder. Pull the latest, should be able to run `npm install` and run teh sample.

DarylThayil:
lgtm!

pkanher617:
Works like a charm, lgtm.

AndrewCraswell:
@jasonnutter should the wiki have been updated with a link to the new sample?
https://github.com/AzureAD/microsoft-authentication-library-for-js/wiki/Samples

jasonnutter:
@AndrewCraswell Yeah I'll make sure that page gets updated.

PR #737, opened 2019-05-31
Telemetry Base Event and Telemetry Manager

DarylThayil:
This pull request adds the following

TelemetryManager - the class that will be responsible for taking events from the code and starting, stopping the, and storing them to be returned to developer. It is configured with platform /app info + a callback that is created by the developer

TelemetryPlatform / Config - Config for Telemetry Manager

TelemetryEvent - The base event that all events will inherit from (apievent, uievent, cacheevent etc... coming in next pr) It is responsible for storing start and end time about a specific event along with metadata about a specific event.

DefaultEvent - A special event that is probably poorly named but follows convention of other adal/msal. It is the "AggregateEvent". On a flush for a correlationId, the default event is appended to the end of all the events, with aggregate counts of certain types of events + platform info on the set of events.

TelemetryConstants - A set of constants used for the Telemetry classes

TelemetryUtils - A set of utils used specifically for telemetry

-TelemetryTypes - Interfaces / Types for different Objects

Tests - !Yay


lib/msal-core/src/telemetry/TelemetryEvent.ts
```diff
@@ -0,0 +1,56 @@
...
+    public eventId: string;
+
+    constructor(eventName: string, correlationId: string) {
+
+        this.startTime = new Date().getTime();
+        this.eventId = uuid();
+        this.event = {
+            [`${EVENT_NAME_PREFIX}${EVENT_NAME_KEY}`]: eventName,
+            [`${EVENT_NAME_PREFIX}${START_TIME_KEY}`]: this.startTime,
+            [`${EVENT_NAME_PREFIX}${ELAPSED_TIME_KEY}`]: -1,
```
jasonnutter: What does `-1` represent here? Is it just used for semantics, or there a programmatic reason?

lib/msal-core/src/telemetry/TelemetryManager.ts
```diff
@@ -0,0 +1,131 @@
...
+
+    constructor(config: TelemetryConfig, cb: Function) {
+        // TODO THROW if bad options
+        this.telemetryPlatform = config.platform;
+        this.clientId = config.clientId;
+        this.onlySendFailureTelemetry = config.onlySendFailureTelemetry;
+        // TODO, when i get to wiring this through, think about what it means if
+        // a developer does not implement telem at all, we still instrument, but cb can be
+        // optional?
+        this.telemetryCallback = cb;
```
jasonnutter: You could default it to a noop function, which would remove the need downstream to always be checking the truthiness of the `telemetryCallback` function.

jasonnutter: I suppose unless you don't want the code after those checks to run at all if the function isn't passed.

lib/msal-core/src/telemetry/TelemetryUtils.ts
```diff
@@ -0,0 +1,35 @@
...
+    // validate trusted host
+    if (!TrustedHostList[url.HostNameAndPort.toLocaleLowerCase()]) {
+        // Should this return null or what was passed?
+        return null;
+    }
+
+    const pathParams = url.PathSegments;
+
+    if (pathParams && pathParams.length >= 2) {
+        const tenantPosition = pathParams[1] ===  B2cAuthority.B2C_PREFIX ? 2 : 1;
```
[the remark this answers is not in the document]
jasonnutter: I think I see what it means now...this could probably be a one-liner with a regex yeah? :)

lib/msal-core/test/telemetry/TelemetryUtils.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+        expect(uri).to.eq("https://login.microsoftonline.com/abc-123");
+    });
+    it("TelemtryUtils.scrubTenantFromUri works for b2c", () => {
+        const uri = scrubTenantFromUri("https://login.microsoftonline.com/abc-123/tfp/banana");
+        expect(uri).to.eq("https://login.microsoftonline.com/abc-123/tfp/<tenant>");
+    });
+    it("TelemtryUtils.scrubTenantFromUri works for b2c", () => {
+        const uri = scrubTenantFromUri("https://login.microsoftonline.com/abc-123/tfp/banana");
+        expect(uri).to.eq("https://login.microsoftonline.com/abc-123/tfp/<tenant>");
+    });
```
jasonnutter: I think this is the same test twice in a row?

jasonnutter: I'm guessing you meant to create a second test without the b2c prefix?

lib/msal-core/package.json
```diff
@@ -85,6 +87,7 @@
     "typedoc": "^0.14.2",
     "typescript": "^3.2.1",
     "uglifyjs-webpack-plugin": "^2.1.1",
+    "uuid": "^3.3.2",
```
jasonnutter: Seems like this should be a regular dependency?

DarylThayil:
@jasonnutter  on usage examples, since this just creates some of the base classes needed, and doesn't instrument or expose the api to the dev yet, that will follow with the second set.

lib/msal-core/src/telemetry/TelemetryEvent.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export default class TelemetryEvent {
+
+    private startTime: number;
+    protected event: any; // TODO TYPE THIS
+    public eventId: string;
+
+    constructor(eventName: string, correlationId: string) {
+
+        this.startTime = new Date().getTime();
```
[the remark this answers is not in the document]
DarylThayil: good call, updating

lib/msal-core/src/telemetry/TelemetryManager.ts
```diff
@@ -0,0 +1,131 @@
+import TelemetryEvent from "./TelemetryEvent";
+import { CompletedEvents, EventCount, EventCountByCorrelationId, InProgressEvents, TelemetryConfig, TelemetryPlatform } from "./TelemetryTypes";
+import DefaultEvent from "./DefaultEvent";
+import { platform } from "os";
```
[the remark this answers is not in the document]
DarylThayil: ahh must have code complete imported while working on this. good catch

lib/msal-core/src/telemetry/TelemetryManager.ts
```diff
@@ -0,0 +1,131 @@
...
+        event.stop();
+        this.incrementEventCount(event);
+
+        const completedEvents = this.completedEvents[event.telemetryCorrelationId];
+
+        if (completedEvents && completedEvents.length) {
+            this.completedEvents[event.telemetryCorrelationId] = [...completedEvents, event];
+        } else {
+            this.completedEvents[event.telemetryCorrelationId] = [event];
+        }
```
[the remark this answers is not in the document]
DarylThayil: love it, updating

lib/msal-core/src/telemetry/TelemetryManager.ts
```diff
@@ -0,0 +1,131 @@
...
+    startEvent(event: TelemetryEvent) {
+        if (!this.telemetryCallback) {
+            return;
+        }
+        const eventKey = createEventKey(event);
+        this.inProgressEvents[eventKey] = event;
+    }
+
+    stopEvent(event: TelemetryEvent) {
+        if (!this.telemetryCallback || !this.inProgressEvents[createEventKey(event)]) {
```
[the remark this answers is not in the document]
DarylThayil: +1

lib/msal-core/src/telemetry/TelemetryManager.ts
```diff
@@ -0,0 +1,131 @@
...
+        if (!eventCount) {
+            this.eventCountByCorrelationId[event.telemetryCorrelationId] = {
+                [eventName]: 1
+            };
+        } else {
+            if (eventCount[eventName]) {
+                eventCount[eventName] = eventCount[eventName] + 1;
+            } else {
+                eventCount[eventName] = 1;
+            }
```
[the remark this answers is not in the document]
DarylThayil: +1

lib/msal-core/src/telemetry/DefaultEvent.ts
```diff
@@ -0,0 +1,26 @@
...
+        this.event[`${EVENT_NAME_PREFIX}application_version`] = platform.applicationVersion;
+        this.event[`${TELEMETRY_BLOB_EVENT_NAMES.UiEventCountTelemetryBatchKey}`] = this.getEventCount(`${EVENT_NAME_PREFIX}ui_event`, eventCount);
+        this.event[`${TELEMETRY_BLOB_EVENT_NAMES.HttpEventCountTelemetryBatchKey}`] = this.getEventCount(`${EVENT_NAME_PREFIX}http_event`, eventCount);
+        this.event[`${TELEMETRY_BLOB_EVENT_NAMES.CacheEventCountConstStrKey}`] = this.getEventCount(`${EVENT_NAME_PREFIX}cache_event`, eventCount);
+        /// Device id?
+    }
+
+    private getEventCount(eventName: string, eventCount: EventCount): String {
+        if (!eventCount[eventName]) {
+            return "0";
```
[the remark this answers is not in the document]
DarylThayil: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/master/src/Microsoft.Identity.Client/TelemetryCore/Internal/Events/DefaultEvent.cs

Based on this code, for consistent output with other msal telemtry

lib/msal-core/src/telemetry/TelemetryEvent.ts
DarylThayil: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/master/src/Microsoft.Identity.Client/TelemetryCore/Internal/Events/EventBase.cs#L28

Default used to denote that an elapsed time has not been set on this event, for consistency with other msal libs

lib/msal-core/src/telemetry/TelemetryManager.ts
DarylThayil: Yeah I think for now that if the callback is not passed, we don't want to track any of this data.

lib/msal-core/test/telemetry/TelemetryUtils.spec.ts
DarylThayil: good catch, not sure what I meant there!

lib/msal-core/package.json
DarylThayil: good call

lib/msal-core/src/telemetry/TelemetryConstants.ts
```diff
@@ -0,0 +1,19 @@
+export const EVENT_NAME_PREFIX: String = "msal.";
+export const EVENT_NAME_KEY: String = "event_name";
+export const START_TIME_KEY: String = "start_time";
+export const ELAPSED_TIME_KEY : String = "elapsed_time";
+
+export const TELEMETRY_BLOB_EVENT_NAMES =  {
+    MsalCorrelationIdConstStrKey: "Microsoft.MSAL.correlation_id",
+    ApiTelemIdConstStrKey: "msal.api_telem_id",
+    ApiIdConstStrKey: "msal.api_id",
+    BrokerAppConstStrKey: "Microsoft_MSAL_broker_app",
```
sameerag: Why do we need Broker APP?

lib/msal-core/src/telemetry/TelemetryUtils.ts
```diff
@@ -0,0 +1,35 @@
+import { B2cAuthority } from "../B2cAuthority";
+import { TrustedHostList } from "../Constants";
+import { Utils } from "../Utils";
+
+// This is used to replace the real tenant in telemetry info
```
sameerag: Can be moved to TelemetryConstant.ts may be?

lib/msal-core/src/telemetry/TelemetryUtils.ts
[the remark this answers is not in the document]
sameerag: Not adding parenthesis is confusing on the first look.

lib/msal-core/src/AadAuthority.ts
```diff
@@ -3,6 +3,7 @@
 
 import { Authority, AuthorityType } from "./Authority";
 import { XhrClient } from "./XHRClient";
+import { TrustedHostList } from './Constants';
```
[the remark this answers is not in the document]
DarylThayil: Will rename to AADTrustedHostList

lib/msal-core/src/AadAuthority.ts
```diff
@@ -61,6 +53,6 @@ export class AadAuthority extends Authority {
    * @param {string} The host to look up
    */
   public IsInTrustedHostList(host: string): boolean {
-    return AadAuthority.TrustedHostList[host.toLowerCase()];
+    return TrustedHostList[host.toLowerCase()];
```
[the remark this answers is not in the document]
DarylThayil: will rename

lib/msal-core/src/telemetry/TelemetryConstants.ts
DarylThayil: we probably do not, once I wire up all of the events, I will be pruning what is unused

lib/msal-core/src/telemetry/TelemetryUtils.ts
DarylThayil: updating

lib/msal-core/src/telemetry/TelemetryUtils.ts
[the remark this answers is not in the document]
DarylThayil: what would the parens do?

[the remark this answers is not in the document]
DarylThayil: what would the regex be? I am leaving numbers in for now, only because its an index of an array we are going after

lib/msal-core/src/telemetry/TelemetryUtils.ts
DarylThayil: updated

lib/msal-core/package-lock.json
```diff
@@ -290,6 +290,15 @@
       "integrity": "sha512-6ee09Ugx6GyEr0opUIakmxIWFNmqYPjkqa3/BuxCBokA0klsOLPgMD5K4q40lH7/yZVuJVzOfQpd7pipwjngkQ==",
       "dev": true
     },
+    "@types/uuid": {
+      "version": "3.4.4",
+      "resolved": "https://registry.npmjs.org/@types/uuid/-/uuid-3.4.4.tgz",
+      "integrity": "sha512-tPIgT0GUmdJQNSHxp0X2jnpQfBSTfGxUMc/2CXBU2mnyTFVYVa2ojpoQ74w0U2yn2vw3jnC640+77lkFFpdVDw==",
+      "dev": true,
+      "requires": {
+        "@types/node": "*"
```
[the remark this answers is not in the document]
DarylThayil: unclear, its a generated file by npm

sameerag:
@DarylThayil Looks fine now.

sameerag:
Please resolve the conflicts - I know it is a generated file but we may need to check why the conflicts are generated.

lib/msal-core/package.json
```diff
@@ -37,14 +37,17 @@
   },
   "dependencies": {
     "js-base64": "^2.5.1",
-    "tslib": "^1.9.3"
+    "sha256": "^0.2.0",
```
pkanher617: I might consider a different sha-256 library here. I know it's a small thing but encryption libraries are notorious for being attack vectors. It worries me that this only has three published versions, and I also don't see any developers that I can trust here. Something like [js-crypto](https://www.npmjs.com/package/crypto-js) or [Stanford's Javascript Crypto Library](https://www.npmjs.com/package/sjcl) might be a more trustworthy option since they have more releases and trusted developers behind them.

lib/msal-core/src/telemetry/DefaultEvent.ts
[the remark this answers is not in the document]
pkanher617: String seems weird to me too, couldn't you just return 0 and string convert if needed? If someone calls if(getEventCount(x, y) > 0) they may end up with errors or false positives.

lib/msal-core/test/telemetry/TelemetryEvent.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+import TelemetryEvent from "../../src/telemetry/TelemetryEvent";
+import { v4 as uuid } from "uuid";
+import { idText } from "typescript";
+
+
+describe("TelemetryEvent", () =>{
+    it("constructed with correct params", () => {
+        const correlationId = uuid();
+        const eventName = "fakeEvent";
+        const telemetyEvent: TelemetryEvent = new TelemetryEvent(
```
pkanher617: typo: telemetyEvent -> telemetryEvent

lib/msal-core/test/telemetry/TelemetryEvent.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+        );
+        expect(telemetyEvent.telemetryCorrelationId).to.eql(correlationId);
+        expect(telemetyEvent.eventName).to.eql(eventName);
+    });
+
+    it("stop event and get elapsed time", done => {
+        const time = 500;
+        const correlationId = uuid();
+        const eventName = "coolEvent";
+        const telemetyEvent: TelemetryEvent = new TelemetryEvent(
```
pkanher617: typo: telemetyEvent -> telemetryEvent

lib/msal-core/test/telemetry/TelemetryEvent.spec.ts
pkanher617: typo: telemetyEvent -> telemetryEvent

lib/msal-core/src/telemetry/DefaultEvent.ts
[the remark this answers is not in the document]
pkanher617: This looks internal, so I guess it is understandable. It makes more sense to me to make it an int but if other MSALs are doing it, this shouldn't hold us back.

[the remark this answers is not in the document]
sameerag: I agree. Unless we are breaking something on a surface telemetry level where eventually we want benefits from collective data combing and analytics, changing this to 'int' should not be an issue @DarylThayil

lib/msal-core/src/AadAuthority.ts
```diff
@@ -3,6 +3,7 @@
 
 import { Authority, AuthorityType } from "./Authority";
 import { XhrClient } from "./XHRClient";
+import { AADTrustedHostList } from './Constants';
```
pkanher617: I am getting a linting error since you are using single quotes here instead of double quotes, this will cause the build to fail.

pkanher617: Fixing this doesn't cause any further build issues.

lib/msal-core/src/telemetry/DefaultEvent.ts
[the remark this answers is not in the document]
DarylThayil: updating

lib/msal-core/test/telemetry/TelemetryEvent.spec.ts
DarylThayil: fixing

lib/msal-core/src/AadAuthority.ts
DarylThayil: weird ... wonder why it didnt break me!

DarylThayil: fixing

lib/msal-core/package.json
DarylThayil: uninstalling for now

DarylThayil:
fixing sorry tests wont run on my machine, I think its wsl

PR #742, opened 2019-06-05
Adding IdToken claims to MSAL JS interfaces : AuthResponse and Account

sameerag:
This PR:

- adds all the claims in an id_token in the API surface for the developer reference; idToken in 'Account' is to map a request to the corresponding session. This PR deprecates the current parameter 'idToken' in the 'Account' class and adds a more meaningful parameter: 'idTokenClaims'. Since both are same, for backward compatibility until 1.1.0, both the parameters are supported.
- adds 'idTokenClaims' in response to map the claims within a request-response cycle.
- Today we have two different data types for IdToken and AccessToken. Ideally, both should have been tokens (JWT format as referenced in Microsoft AAD docs). Hence as a part of this PR the changes in UAA and AuthResponse are made.

Relevant changes:
- [x] Corresponding changes in response in UserAgentApplication
- [x] Unit tests added for claims; more enhancement in UserAgentApplication test cases
- [x] Unit tests added for IdToken and Account classes

Todo:

- [ ] Documentation changes needed for the Surface changes will be edited in the Readme once this PR is ready to be Merged.


lib/msal-core/src/Account.ts
```diff
@@ -20,7 +21,7 @@ export class Account {
     homeAccountIdentifier: string;
     userName: string;
     name: string;
-    idToken: Object;
+    claims: QPDict;
```
DarylThayil: Not sure this is a Query Parameter Dictionary, same type but the name is not right. 
Should we rename QP dictionary to Dict which is a generic <String, String> interface

DarylThayil:
looks fairly straightforward to me, other than type naming.

DarylThayil:
can we add tests ?

lib/msal-core/src/Account.ts
sameerag: Agree. 👍

pkanher617: Is this called claims in other MSAL libraries as well? I would prefer something like idTokenClaims to be more descriptive.

pkanher617: We use idTokenClaims down below as well.

lib/msal-core/src/AuthResponse.ts
```diff
@@ -17,6 +18,7 @@ export type AuthResponse = {
     expiresOn: Date;
     account: Account;
     accountState: string;
+    claims: QPDict;
```
pkanher617: If we already include these as part of account, is there any value in adding this to the response object?

lib/msal-core/src/Account.ts
pkanher617: +1 for changing QPDict to somethign else.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1831,6 +1833,7 @@ export class UserAgentApplication {
           // retrieve the id_token from response if present :
           if (hashParams.hasOwnProperty(Constants.idToken)) {
             response.idToken = new IdToken(hashParams[Constants.idToken]);
+            response.claims = response.idToken.decodedIdToken;
```
pkanher617: Why reassign a value that is already in the object? Seems like we are starting to have duplicate values in multiple places. This will start bloating objects, I think we should try to move away from this practice. If we need to restructure account and response to do this I am all for it.

lib/msal-core/src/Utils.ts
```diff
@@ -705,6 +705,7 @@ export class Utils {
   static setResponseIdToken(originalResponse: AuthResponse, idToken: IdToken) : AuthResponse {
     var response = { ...originalResponse };
     response.idToken = idToken;
+    response.claims = idToken.decodedIdToken;
```
pkanher617: Same comment as above.

lib/msal-core/src/Constants.ts
```diff
@@ -120,6 +120,11 @@ export const PromptState = {
 	NONE: "none",
 };
 
+/**
+ * Key-Value type to support queryParams and extraQueryParams
+ */
+export type QPDict = {[key: string]: string};
```
pkanher617: We should try to separate string constants and types that we want to export. Would a separate file for types be better?

lib/msal-core/src/AuthResponse.ts
sameerag: We can choose not to, it is to tie up request-response as most people look up for the response for claims. Account is one step further.

lib/msal-core/src/Constants.ts
sameerag: May be. If we have more of these. Let me make a call once I review how many generic types we have in our code to separate them from Constants.ts file.

lib/msal-core/src/UserAgentApplication.ts
sameerag: One option is to have it only in 'Account' but as stated in the above comment, it ties up in two different flows, request-response  (and the corresponding look up for claims); Account for generic session information. Let me get back on this after further evaluation.

sameerag:
As synced up offline, 
1) We need to be uniform in our response 'token' type, for both idToken and accessToken, I will make changes that reflect the same. Today we have two different data types for both of them and I think we should be more consistent. I will make this change soon.
2) W.r.t idToken support in Account, it is to map a request to the corresponding session, we have always traditionally maintained idToken claims in the account, the field is just named 'idToken' instead of 'Claims'. 
3) Adding 'claims' in response is to map the claims within a request-response cycle.

Let me add the pending changes and tests, we can revisit this.

lib/msal-core/src/AuthResponse.ts
DarylThayil: I like having it on the response as well, otherwise people have to get the response, and then call get account vs just grabbing it.  

Can still get from account if its out of band from request

lib/msal-core/src/UserAgentApplication.ts
DarylThayil: @pkanher617  I think what was talked about was getting rid of 'idToken' and so having 'claims' also available here lets us not break those looking at id token, and we can deprecate it, next major version.

@sameerag  if that is infact what we are doing, can we add a comment here explaining it ?

DarylThayil: also will need to update docs to reflect this.

lib/msal-core/src/AuthResponse.ts
pkanher617: the account object is already part of the response object, so they wouldn't have to call any other function. It would just be accessing the account object we return in the response.

pkanher617: Line 18

lib/msal-core/src/UserAgentApplication.ts
pkanher617: I think what we agreed upon was to remove the IdToken object type and change it to a string. Then adding the claims object makes sense to me. We should only use IdToken internally since we are not exporting it as a type.

sameerag: Yes. What @DarylThayil is stating is different, he is mentioning the parameter in 'Account' and not the 'Response'.

I added what we agreed as a comment at the end of this PR. Will alert you guys when the code changes are available.

lib/msal-core/src/Account.ts
sameerag: Done. Changed to Dict.

lib/msal-core/src/AuthResponse.ts
sameerag: Resolved to both "Account" and "Response" having claims.

lib/msal-core/src/UserAgentApplication.ts
sameerag: Code changes added.

lib/msal-core/src/Utils.ts
sameerag: Done. As agreed, changed idToken in response to "String" and added "claims" in 'reponse' and 'account' objects.

lib/msal-core/src/Constants.ts
sameerag: Done, introduced a new file for 'MsalTypes'

lib/msal-core/src/Utils.ts
```diff
@@ -703,17 +704,20 @@ export class Utils {
...
+    if (idToken.objectId) {
+      response.uniqueId = idToken.objectId;
     } else {
-      response.uniqueId = response.idToken.subject;
+      response.uniqueId = idToken.subject;
     }
-    response.tenantId = response.idToken.tenantId;
+    response.tenantId = idToken.tenantId;
+
     return response;
```
jasonnutter: This block could be simplified a bit:

```js
return {
  ...originResponse,
  idToken: idToken.rawIdToken,
  claims: idToken.claims,
  uniqueId: idToken.objectId || idToken.subject,
  tenantId: idToken.tenantId
}
```

lib/msal-core/src/AuthResponse.ts
```diff
@@ -11,12 +12,13 @@ export type AuthResponse = {
     uniqueId: string;
     tenantId: string;
     tokenType: string;
-    idToken: IdToken;
+    idToken: string;
```
jasonnutter: Seems like this is a breaking change, going from an object to a string?

sameerag: Yes, it is. This is a conscious one to keep the idToken object private:

- our internal naming of some claims in idToken can confuse users.
- cleaner interfacing, uniformity with accessToken format.

jasonnutter: Could this instead add a property (`idTokenString` or something) instead of changing the shape?

pkanher617: We had that as "rawIdToken" part of idToken, but since we expose accessToken as a string it makes sense to do the same with idToken.

DarylThayil: okay so just to be clear... 
this used to be a raw id token and is still a raw id token right?

sameerag: nope, It used to be 'idToken: IdToken', which is a class in our library. What @pkanher617 is referring to is that that class also had a property, 'IdToken.rawIdToken' which is a "string".
The catch is that we never exported 'IdToken', hence it was just a js object.

lib/msal-core/src/MsalTypes.ts
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+/**
+ * Key-Value type to support queryParams and extraQueryParams
+ */
+export type Dict = {[key: string]: string};
```
DarylThayil: i am fine with this but dict is like a hashmap, which can be key valued by any Generic, Dict<T, T> where as this is string specific.

I wonder if what we want is the typescript object type

lib/msal-core/src/AuthResponse.ts
DarylThayil: okay, my misunderstanding. I think I would agree that we should not introduce a breaking change. 
IdToken should stay an object, and we should deprecate it in major version.

DarylThayil: can also add rawIdToken separately

sameerag: @DarylThayil This is exactly the point I was making the other day before we decided on the interface.

- Though we have idToken as a JS object, we are not exporting its type.
- We need to either add 'IdToken' class as an exported type or change it to a rawToken.

Cons of exposing the internal non-standard IdToken class:

- The problem with exposing a non-standard  [IdToken](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-core/src/IdToken.ts) class is that we are choosing to part open the idToken for our internal code and having the developer see it.
- At the end of the day, all the dev needs is the idToken and the claims with it.
- This also gives the freedom to change/add to the internal IdToken class without affecting the developer.

This is the reason why I was advocating to change the current response to idToken: string instead of idToken: object. The object has been mysterious anyways unless someone opens up the code as we stand today.

Follow this up with good documentation, we should be able to push it.

sameerag: If you want to break this in a major version, I am fine with it. Let the PR stay and integrate it in 1.1.x!

jasonnutter: > We need to either add 'IdToken' class as an exported type or change it to a rawToken.

This is because since the `IdToken` class isn't exported, it doesn't get included in the exported types for the library, and thus doesn't show up in intellisense?

I would personally advocate for exporting the class (and having teh types work), instead of introducing a breaking change and creating work for our users.

sameerag: @jasonnutter From a code perspective, I get your stand. My concern is more about exposing the non-standard idToken class type that MSAL JS introduced. No MSAL exposes an IdToken class that picks only certain claims in the class structure and renames them internally.

It should not have been returned to the user in the firs place IMO.

lib/msal-core/src/Utils.ts
sameerag: Done.

lib/msal-core/src/MsalTypes.ts
sameerag: Any name suggestions?

lib/msal-core/src/AuthResponse.ts
sameerag: Made the changes non-breaking. Please review.

lib/msal-core/src/MsalTypes.ts
sameerag: Any ideas for naming?

sameerag:
Closing this in favor of #804

PR #745, opened 2019-06-06

pkanher617:


azure-pipelines.yml
```diff
@@ -0,0 +1,21 @@
...
+trigger:
+- master
+
+pool:
+  vmImage: 'ubuntu-latest'
+
+steps:
+- task: NodeTool@0
+  inputs:
+    versionSpec: '10.x'
```
jasonnutter: It might be good to pin this to a specific version, so that our personal machines can match whats used in CI.

DarylThayil: Is this the node version?

azure-pipelines.yml
```diff
@@ -0,0 +1,21 @@
...
+  displayName: 'Install Node.js'
+
+- script: |
```
DarylThayil: are we running lint and test here?

DarylThayil: Is this just running install for the parent? how does the mono repo stuff play here?

azure-pipelines.yml
pkanher617: Sorry I didn't mean to create this PR. Although it would be good to agree on a node version.

azure-pipelines.yml
jasonnutter: We would want to add a post install hook to the root package to run `lerna bootstrap` to have all the subpackages be installed correctly.

sameerag: As far as I know, this one is also unintentional PR opened when @pkanher617 is trying some Azure pipelines integration in our code. FYI.

Issue #733, opened 2019-05-28
acquireTokenSilent error not InteractionRequiredAuthError as documented


TiagoBrenck:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x ] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Other... Please describe:
</code></pre>

## Browser:
- [x ] Chrome version 74
- [x ] Firefox version 67

## Library version

Library version: 1.0.0

## Current behavior

According to the documentation, the error handling when acquiring a token could be done like this:

(code from: https://github.com/AzureAD/microsoft-authentication-library-for-js/wiki/MSAL.js-1.0.0-api-release#signing-in-and-getting-tokens-with-msaljs)

```javascript
msalInstance.acquireTokenSilent(tokenRequest)
            .then(response => {
                // get access token from response
                // response.accessToken
            })
            .catch(err => {
                // could also check if err instance of InteractionRequiredAuthError if you can import the class.
                if (err.name === "InteractionRequiredAuthError") {
                    return msalnstance.acquireTokenPopup(tokenRequest)
                        .then(response => {
                            // get access token from response
                            // response.accessToken
                        })
                        .catch(err => {
                            // handle error
                        });
                }
            });
```

However, if errorCode == "login_required" or "consent_required", the error.name is actually "ServerError" not "InteractionRequiredAuthError"

## Expected behavior
I would expect the error.name to be "InteractionRequiredAuthError" or, update the documentation and remove the misleading example.

## Minimal reproduction of the problem with instructions
Execute the following code, in a scenario where an exception will the thrown for "login_required" or "consent_required" and observes the value on err.name:

```javascript
msalInstance.acquireTokenSilent(tokenRequest)
            .then(response => {
                // get access token from response
                // response.accessToken
            })
            .catch(err => {
                // could also check if err instance of InteractionRequiredAuthError if you can import the class.
                if (err.name === "InteractionRequiredAuthError") {
                    return msalnstance.acquireTokenPopup(tokenRequest)
                        .then(response => {
                            // get access token from response
                            // response.accessToken
                        })
                        .catch(err => {
                            // handle error
                        });
                }
            });
```

pkanher617:
[...]

PR #753, opened 2019-06-10
Fix Issue #733: acquireTokenSilent error not InteractionRequiredAuthError as documented

pkanher617:
This is a PR to fix the issue described in #733 where the acquireTokenSilent method was throwing errors of the incorrect type. I have updated the type checking and added tests to ensure we are correctly throwing the error.

### DIFF
--- a/lib/msal-core/test/UserAgentApplication.spec.ts
+++ b/lib/msal-core/test/UserAgentApplication.spec.ts
@@ -11,7 +11,8 @@ import {
     ClientConfigurationError,
     ServerError,
     Authority,
-    AuthResponse
+    AuthResponse,
+    InteractionRequiredAuthError
 } from '../src/index';
 import sinon from "sinon";
 import { ITenantDiscoveryResponse } from "../src/ITenantDiscoveryResponse";
@@ -34,6 +35,9 @@ describe("UserAgentApplication.ts Class", function () {
     const TEST_REDIR_URI = "https://localhost:8081/redirect.html";
     const TEST_LOGOUT_URI = "https://localhost:8081/logout.html";
     const TEST_ERROR_CODE = "error_code";
+    const TEST_INTERACTION_REQ_ERROR_CODE = "interaction_required";
+    const TEST_LOGIN_REQ_ERROR_CODE = "login_required";
+    const TEST_CONSENT_REQ_ERROR_CODE = "consent_required";
     const TEST_ERROR_DESC = "msal error description"
     const TEST_USER_STATE_NUM = "1234";
     const TEST_USER_STATE_URL = "https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-implicit-grant-flow/scope1";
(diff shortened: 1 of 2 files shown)


lib/msal-core/test/UserAgentApplication.spec.ts
```diff
@@ -34,6 +35,9 @@ describe("UserAgentApplication.ts Class", function () {
     const TEST_REDIR_URI = "https://localhost:8081/redirect.html";
     const TEST_LOGOUT_URI = "https://localhost:8081/logout.html";
     const TEST_ERROR_CODE = "error_code";
+    const TEST_INTERACTION_REQ_ERROR_CODE = "interaction_required";
+    const TEST_LOGIN_REQ_ERROR_CODE = "login_required";
+    const TEST_CONSENT_REQ_ERROR_CODE = "consent_required";
```
jasonnutter: Do these need to be duplicated here, or could you pull in this? https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/f8e0eb4c11cb355212c2fa2d2c158a985d748326/lib/msal-core/src/error/InteractionRequiredAuthError.ts#L6-L16

lib/msal-core/test/UserAgentApplication.spec.ts
```diff
@@ -56,6 +60,12 @@ describe("UserAgentApplication.ts Class", function () {
...
     const TEST_SUCCESS_HASH = `#id_token=${TEST_ID_TOKEN}&client_info=${TEST_RAW_CLIENT_INFO}&state=RANDOM-GUID-HERE|`;
     const TEST_ERROR_HASH = "#error=error_code&error_description=msal+error+description&state=RANDOM-GUID-HERE|";
+    const TEST_INTERACTION_REQ_ERROR_HASH1 = "#error=interaction_required&error_description=msal+error+description&state=RANDOM-GUID-HERE|";
+    const TEST_INTERACTION_REQ_ERROR_HASH2 = "#error=interaction_required&error_description=msal+error+description+interaction_required&state=RANDOM-GUID-HERE|";
+    const TEST_LOGIN_REQ_ERROR_HASH1 = "#error=login_required&error_description=msal+error+description&state=RANDOM-GUID-HERE|";
+    const TEST_LOGIN_REQ_ERROR_HASH2 = "#error=login_required&error_description=msal+error+description+login_required&state=RANDOM-GUID-HERE|";
+    const TEST_CONSENT_REQ_ERROR_HASH1 = "#error=consent_required&error_description=msal+error+description&state=RANDOM-GUID-HERE|";
+    const TEST_CONSENT_REQ_ERROR_HASH2 = "#error=consent_required&error_description=msal+error+description+consent_required&state=RANDOM-GUID-HERE|";
```
sameerag: Not specific to this PR but may be we should categorize all our test constants too later for readability. This can wait though.

lib/msal-core/test/UserAgentApplication.spec.ts
```diff
@@ -864,6 +874,130 @@ describe("UserAgentApplication.ts Class", function () {
...
+                expect(error.errorMessage).to.include(TEST_CONSENT_REQ_ERROR_CODE);
+                expect(error.message).to.include(TEST_CONSENT_REQ_ERROR_CODE);
+                expect(error.stack).to.include("UserAgentApplication.spec.js");
+                done();
+            };
+            msal.handleRedirectCallback(checkErrorFromServer);
+        });
+
+    });
+
```
sameerag: +1 for adding tests.

lib/msal-core/test/UserAgentApplication.spec.ts
pkanher617: No that's a good point. I will do that.

lib/msal-core/test/UserAgentApplication.spec.ts
pkanher617: I tried to separate them with comments. See if it is more readable?

lib/msal-core/test/UserAgentApplication.spec.ts
pkanher617: Done, please check and resolve if it looks good.

lib/msal-core/test/UserAgentApplication.spec.ts
sameerag: Definitely yes. Thanks.

lib/msal-core/test/UserAgentApplication.spec.ts
sameerag: Lgtm.

PR #761, opened 2019-06-13
Faced an issue of cyclic dependency while using Msal for angular out of the box

tbudukh:
We faced an issue while using Msal service right out of the box. Issue was regarding cyclic dependency. We were trying to bootup a plugin using Msal service with correct config and added app_initializer in plugin module to use msal service. the project using the plugin would not build because of the cyclic dependencies. Router is actually never available before app_initializer. We figured Msal service doesnt really need the router to create its instance and we can make router available when its ready. 
We pulled out router dependency from Msal service, put it in its own container and provided config. We also bound the router events after router is init. At the bootup, no, Msal service doesnt need router and can bootup without cyclic dependency errors.


jasonnutter:
General comment: This PR includes a lot of changes, including things unrelated to the dependency issue detailed in the description. Would it be possible to limit this PR to just those changes, and include the other  changes (e.g. README, etc) to separate PRs? It will make this PR easier to review.

jasonnutter:
Does seem like pulling in latest `dev` will help clean this PR up a bit, but previous question still stands.

tbudukh:
Cleaned it up and kept this PR to just code changes.

tbudukh:
> General comment: This PR includes a lot of changes, including things unrelated to the dependency issue detailed in the description. Would it be possible to limit this PR to just those changes, and include the other changes (e.g. README, etc) in separate PRs? It will make this PR easier to review. Thanks!

I cleaned the commits a day after the pull request was raised, do I have to do anything else?

jasonnutter:
The diff still includes a large number of unrelated changes. I would recommend closing this PR, updating your fork's dev branch with the latest from upstream's dev, and then reopening your PR.

Issue #736, opened 2019-05-31
acquireTokensilent return null accessToken randomly.


mmuarc:
## I'm submitting a...

[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ x] Other... Please describe:
Technical doubt and possibly a bug.

Browser:
- [ X] Chrome 
- [ ] Firefox version XX
- [ ] IE version XX
- [ ] Edge version XX
- [ ] Safari version XX

## Library version

<
Library version: 1.0.0

## Current behavior
accessToken from acquireTokenSilent is null sometimes

## Expected behavior
accessToken from acquireTokenSilent is never null. Any error should come in the error function.

## Minimal reproduction of the problem with instructions
I have an apihelper that every time is going to do a call contacts my clientApplication wrapper like this:

```
 getAuthorizationCookie = async function () {
        var token = await Ab2c.getTokenAsync()
        if (!token)
        {
            console.log(token);
        }
        return 'Bearer ' + token;
    }
```
The implementation of Ab2c.getTokenAsync is this:

```
let authCallback = function (error, response) {
  if (error)
    console.log(error);
};
const clientApplication = new UserAgentApplication(config);
clientApplication.handleRedirectCallback(authCallback);
//FIXME: this temporary catch token is wrong and should be  removed. It is here as a workaround because ab2c to sometimes returns a valid response with a null token.
let tempCacheToken="";**

class Ab2c {
....

static getTokenAsync = async () => {
    let tokenRequest = {
      scopes: config.auth.b2cScopes,
      authority: config.auth.authority,
      extraQueryParameters: config.auth.extraQueryParameters
    };
    return new Promise((resolve,reject) =>{
    clientApplication.acquireTokenSilent(tokenRequest).then(function (loginResponse) {
      if (!loginResponse.accessToken)
      {
         **//This If happens sometimes when I refresh the page**
          console.log("null token");
          loginResponse.accessToken = tempCacheToken;
      }
      else
      {
             console.log("valid token")
          tempCacheToken = loginResponse.accessToken;
      }
      return resolve(loginResponse.accessToken);
    }).catch(function (error) {
      console.log(error);
      Ab2c.logout()
      let newError = new Error(error);
      newError.fromMsal = true;
      return reject(newError);
    });
    });
  }
....
}
```

[...]

Issue #722, opened 2019-05-23
Fetching multiple access tokens silently only succeeds for one token


raikoss:
## I'm submitting a...

<pre><code>
[x] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Other... Please describe:
</code></pre>

## Browser:
- [x] Chrome version XX
- [ ] Firefox version XX
- [ ] IE version XX
- [ ] Edge version XX
- [ ] Safari version XX

## Library version

<pre><code>
Library version: 1.0.0

</code></pre>

## Current behavior

When trying to fetch two access tokens silently at the same time after a login, only one of the responses actually includes the access token. When refreshing, one token gets fetched from the cache which makes the other token response succeed. 

## Expected behavior

Both token responses should succeed in getting the token.

## Minimal reproduction of the problem with instructions

```js
const msal = new Msal.UserAgentApplication(config);

// after a user has logged in
msal.acquireTokenSilent({ scopes: ['user.read'] })
.then(response => {
  // do something with access token
})

msal.acquireTokenSilent({ scopes: ['api://myapiguid/Read'] })
.then(response => {
  // do something with access token
})
```

When doing the above, either of the responses will not contain an access token, as they are being run at the same time. This used to work in the previous version of msal, 0.2.4 if I recall correctly. 

When doing the following workaround, the tokens are being acquired correctly: 

```js
msal.acquireTokenSilent({ scopes: ['user.read'] })
.then(response => {
  // do something with access token

  return msal.acquireTokenSilent({ scopes: ['api://myapiguid/Read'] });
})
.then(response => {
  // do something with access token
})
```

Which suggests that the tokens are not being successfully acquired when attempting to do so at the same time. 

Edit:
After some more testing, it correctly delays one request until the other one is done. I'm not really sure why, but before the issue happened as explained above. I'm not sure if something changed that affected this, so I'm leaving the issue open in case it happens again.

ThomasJacob:
Another case, for multiple scope token graph scope and custom api scope
In case of first login it properly returns the token properly. But if there was a cache token available which has expired, After adding some delay after login the token gets returned but the token will be for the graph scope and not for the request one. If there is no delay it returns empty.

Can we move this to a high priority issue?

DarylThayil:
@raikoss  @ThomasJacob  
[...]

PR #768, opened 2019-06-18
Fix bug preventing asynchronous silent token requests

pkanher617:
This PR is a fix for issue #736 and #722. 

In order to allow for asynchronous silent token calls, we need to use the state variable for cache clearing to ensure the correct entries are used.


lib/msal-core/src/Storage.ts
```diff
@@ -90,13 +90,17 @@ export class Storage {// Singleton
...
+        let removeState = true;
         if (storage) {
             let key: string;
             for (key in storage) {
                 if (storage.hasOwnProperty(key)) {
-                    if (key.indexOf(CacheKeys.AUTHORITY) !== -1 || key.indexOf(CacheKeys.ACQUIRE_TOKEN_ACCOUNT) !== 1) {
+                    if (state) {
+                        removeState = key.indexOf(state) !== -1;
+                    }
+                    if ((key.indexOf(CacheKeys.AUTHORITY) !== -1 || key.indexOf(CacheKeys.ACQUIRE_TOKEN_ACCOUNT) !== 1) && removeState) {
```
jasonnutter: Could this be this?

```js
if ((key.indexOf(CacheKeys.AUTHORITY) !== -1 || key.indexOf(CacheKeys.ACQUIRE_TOKEN_ACCOUNT) !== 1) && (removeState && key.indexOf(state) !== -1) {
```

pkanher617: pushed the fix. Thanks!

PR #775, opened 2019-06-20
Set up CI with Azure Pipelines for Dev PRs

pkanher617:
This PR is to replace the CI product we are using with Azure Pipelines.


pkanher617:
> Do we want to wait to remove travis until the ADO pipeline in functional? Also how are external submissions going to tested rather the test report shared? Travis today is public.

pkanher617:
> Do we want to wait to remove travis until the ADO pipeline in functional? Also how are external submissions going to tested rather the test report shared? Travis today is public.

Just removing so that we don't have clutter, but eventually we are going to be moving away from travis, so this shouldn't be an issue.

jasonnutter:
Would prefer not to have both...could we make part of the ADO viewable publicly? At least the pipeline status for a given PR.

sameerag:
@jasonnutter yep! As long as someone who submits a PR can see the test reports.

pkanher617:
Most contributors should have access to view test reports. We can verify before we merge this.

jasonnutter:
Yep, looks like anyone can view the status (and the output). Looks good once you are able to get the build passing.

pkanher617:
Moving to PR #912

PR #785, opened 2019-06-24
Fix msal-angular exports to properly support IE11, fixes #765 #414

### LINKED ISSUE #765: IE 11 not Working with Angular MSAL Library
I still see this issue with MSAL Angular library while launching the application in IE11 and console has below Error:

SCRIPT1002: Syntax error
vendor.js (148561,1)

class AuthenticationResult { constructor(token, tokenType) { this._token = ""; this._tokenType = ""; this._token = token; if (tokenType) { this._tokenType = tokenType; } } get token() { return this._token; } set token(value) { this._token = value; } get tokenType() { return this._tokenType; } set tokenType(value) { this._tokenType = value; } }

isn't the classes supposed to be transpiled to es5?

jasonnutter:
The main export for the `msal-angular` library was the ES6 version of the library, which is not compatible with IE11 (I was able to reproduce this with the MSAL Angular sample). This updates the library's exports to properly export the CommonJS version of the library by default, with the ES6 version available under `module` ([which matches](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-core/package.json#L22-L24) the main `msal` package).

Before:

msal-angular-ie11-before

After:

msal-angular-ie11

### DIFF
--- a/lib/msal-angular/samples/MSALAngularDemoApp/package.json
+++ b/lib/msal-angular/samples/MSALAngularDemoApp/package.json
@@ -24,7 +24,7 @@
     "rxjs": "^5.1.0",
     "ts-helpers": "^1.1.1",
     "zone.js": "^0.8.4",
-    "@azure/msal-angular": "0.1.2"
+    "@azure/msal-angular": "file:../.."
   },
   "devDependencies": {
     "@angular/cli": "^1.0.1",
--- a/lib/msal-angular/package.json
+++ b/lib/msal-angular/package.json
@@ -18,8 +18,9 @@
     "msal",
     "oauth"
   ],
-  "main": "./dist/index.js",
-  "typings": "./dist/index.d.ts",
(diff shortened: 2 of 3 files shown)


lib/msal-angular/samples/MSALAngularDemoApp/package.json
```diff
@@ -24,7 +24,7 @@
     "rxjs": "^5.1.0",
     "ts-helpers": "^1.1.1",
     "zone.js": "^0.8.4",
-    "@azure/msal-angular": "0.1.2"
+    "@azure/msal-angular": "file:../.."
```
sameerag: Does this help with picking the latest?

jasonnutter: Yeah, this will install the locally built version of the library, instead of what's published to npm.

Issue #790, opened 2019-06-26
Incorrect calculation of accessTokenResponse.expiresOn date


sergey-tihon:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Other... Please describe:
</code></pre>

## Browser:
- [x] Chrome version XX
- [ ] Firefox version XX
- [ ] IE version XX
- [ ] Edge version XX
- [ ] Safari version XX

## Library version
1.0.2

## Current behavior
Library incorrectly calculate access token expiration date
Token should leave about 1h (not 49 years)
![Skype](https://user-images.githubusercontent.com/1197905/60192421-7a5fbe00-983e-11e9-8b7a-49fc5053f7e1.png)

## Minimal reproduction of the problem with instructions

You do have `Utils. expiresIn` that add `this.now()` to expiration window and technically return `expiresOn`.

https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/0648a237f774e71f0a10e95eeafea92c0df7b8a4/lib/msal-core/src/Utils.ts#L141-L147

but later in the code you add `Utils.now()` one more time and add extra `2019-1970=49` years to expiration.

```js
const expiresIn = Utils.expiresIn(parameters[Constants.expiresIn]).toString();
let exp = Number(expiresIn);
accessTokenResponse.expiresOn = new Date((Utils.now() + exp) * 1000);
```

here is the buggy code

https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/2fefca80827b2c408dad28d79de6f10664aa4031/lib/msal-core/src/UserAgentApplication.ts#L1714-L1724

sameerag:
Thanks, @sergey-tihon for the fix @ #791. We will try to close loop once testing is complete and have it released asap.

DarylThayil:
@sergey-tihon  fixed in #799 released in 1.1.1

PR #791, opened 2019-06-26
Fix for expiration date calculation

sergey-tihon:
Potential fix for #790 

`expiresIn` function is used in only one place, so the fix should be 'safe'
that why I think would be better to remove `this.now()` or we have to rename this function to `expiresOn`

### DIFF
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -1721,7 +1721,7 @@ export class UserAgentApplication {
       accessTokenResponse.scopes = consentedScopes;
       let exp = Number(expiresIn);
       if (exp) {
-        accessTokenResponse.expiresOn = new Date((Utils.now() + exp) * 1000);
+        accessTokenResponse.expiresOn = new Date(exp * 1000);
       } else {
         this.logger.error("Could not parse expiresIn parameter. Given value: " + expiresIn);
       }
--- a/lib/msal-core/src/Utils.ts
+++ b/lib/msal-core/src/Utils.ts
@@ -143,7 +143,7 @@ export class Utils {
      if (!expires) {
          expires = "3599";
       }
-    return parseInt(expires, 10);
+    return this.now() + parseInt(expires, 10);
   }
 
   /**
--- a/lib/msal-core/src/Utils.ts
+++ b/lib/msal-core/src/Utils.ts
@@ -143,7 +143,7 @@ export class Utils {
      if (!expires) {
          expires = "3599";
       }
-    return this.now() + parseInt(expires, 10);
+    return parseInt(expires, 10);
   }
 
   /**


jasonnutter:
I think the better fix would be to fix this line (as opposed to `Utils.expiresIn`):

```js
accessTokenResponse.expiresOn = new Date((Utils.now() + exp) * 1000); 
```

which would be

```js
accessTokenResponse.expiresOn = new Date(exp * 1000); 
```

(I'm also surprised to not see any unit tests failing as a result of this PR, meaning we should probably add some)

sergey-tihon:
Done.
@jasonnutter Can you please help find the right place for the assertion? (if you want me to add test to this PR before the merge).  I cannot find tests that verify that JWT access token is interpreted correctly.

sameerag:
@sergey-tihon There are some added with #742; Also @pkanher617 when this change is made I thought we considered this use case? Did we not add tests?

pkanher617:
Unfortunately this won't fix your issue. expiresIn comes back from the server as the lifetime of the token in seconds (i.e. 3599 seconds, or one hour), which we convert this to milliseconds and add to the current time to get the expiration Date object. Date constructor takes Unix time, which is in milliseconds. I think the parentheses are just off, it should be Utils.now() + (exp * 1000).

I am surprised we don't have unit tests covering this, I will make sure to add those. I can add a separate PR for these changes, or you can add it here and merge this, and I can just add the unit tests after.

jasonnutter:
@pkanher617 In this context, [the value of `exp` is the value](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-core/src/UserAgentApplication.ts#L1714-L1722) that comes back from `Utils.expiresIn` [already includes the current timestamp in its value](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/1ba05b01d40c6525f530f5418cb9102635515c13/lib/msal-core/src/Utils.ts#L146).

sergey-tihon:
>  I think a better strategy is to fix Utils.ts to just return parseInt(expires, 10) in the expiresIn function.

@pkamalamsft That was my initial idea and first commit in this PR.
But than @jasonnutter point me out that we use `const expiresIn = Utils.expiresIn` in [accessTokenValue construction](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-core/src/UserAgentApplication.ts#L1716) and it will be wrong there if we remove `Utils.now()` from `Utils.expiresIn`.

> I will make sure to add those. I can add a separate PR for these changes, or you can add it here and merge this, and I can just add the unit tests after.

@pkamalamsft it would be awesome if you can help me with tests for this case.

__One more strange thing__

If I understand correctly, msal.js persist `accessTokenKey+ accessTokenValue` in `localStorage`. 
But why token value (for login/identity token) store `expiresIn` as a Number
Presales_Portal

but accessTokenValue  for resource/graph token is stored as a string 
Presales_Portal

Just checking, it is by design or maybe fixed in one of upcoming (not merged) PRs?
If this is an unexpected behavior I can create new issue.

pkanher617:
@sergey-tihon You may have tagged the wrong person lol.

I have added a PR #799 to fix this issue.

> That was my initial idea and first commit in this PR.
But than @jasonnutter point me out that we use const expiresIn = Utils.expiresIn in accessTokenValue construction and it will be wrong there if we remove Utils.now() from Utils.expiresIn.

@jasonnutter was correct, our caching logic breaks if we change it there completely, so I have moved that calculation out of the Utils function and calculate at cache time the expiration.

> it would be awesome if you can help me with tests for this case.

No problem, I have added tests.

> Just checking, is it by design or maybe fixed in one of upcoming (not merged) PRs?
If this is an unexpected behavior I can create new issue.

Yes this was fixed in an unmerged PR. This happens because in the case of an id token, we do not receive the expiration lifetime from the server, so we extract this from the token. I have fixed this to always store as a string. Please let me know if these changes work. I will merge my PR if the changes seem to work for you.

sergey-tihon:
Close this PR (😢) in favor of #799

PR #799, opened 2019-06-28
Fix expiresIn calculation

pkanher617:
This PR is to fix the issue raised in PR #790 and add unit tests to ensure future calculations are as expected.

This issue was attempted to be fixed in PR #791 but 791 did not have all of the changes and tests necessary to ensure this use case worked moving forward.

Summary of Changes:
- Changes to expiration time calculation for access token and id token flows
- Utils.expiresIn now returns expiresIn time, not expiresOn
- Utils.setResponseIdToken now also sets expiration
- Simplified saveAccessToken expiration calculation
- Ensure that expiration is always saved as a string in the cache
- Added tests

### DIFF
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -1712,15 +1712,15 @@ export class UserAgentApplication {
       }
 
       // Generate and cache accessTokenKey and accessTokenValue
-      const expiresIn = Utils.expiresIn(parameters[Constants.expiresIn]);
+      const expiresIn = Utils.parseExpiresIn(parameters[Constants.expiresIn]);
+      expiration = Utils.now() + expiresIn;
       const accessTokenKey = new AccessTokenKey(authority, this.clientId, scope, clientObj.uid, clientObj.utid);
-      const accessTokenValue = new AccessTokenValue(parameters[Constants.accessToken], response.idToken.rawIdToken, (expiresIn + Utils.now()).toString(), clientInfo);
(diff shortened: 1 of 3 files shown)


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1711,20 +1712,15 @@ export class UserAgentApplication {
...
 
       accessTokenResponse.accessToken  = parameters[Constants.accessToken];
       accessTokenResponse.scopes = consentedScopes;
-      let exp = Number(expiresIn);
-      if (exp) {
-        accessTokenResponse.expiresOn = new Date((Utils.now() + exp) * 1000);
-      } else {
-        this.logger.error("Could not parse expiresIn parameter. Given value: " + expiresIn);
-      }
+      expiration = Utils.now() + expiresIn;
```
jasonnutter: Looks like this is done twice (on line 1717 also), you should create a `const` for this and use it there as well.

lib/msal-core/src/Utils.ts
```diff
@@ -140,10 +140,10 @@ export class Utils {
    */
   static expiresIn(expires: string): number {
```
jasonnutter: Could this be `parseExpiresIn`, as the name doesn't make it clear what is happening. And I think the variable could be `expiresIn` to be even more clear.

lib/msal-core/src/UserAgentApplication.ts
pkanher617: Fixed this. Can't use const because this value is set after determining whether or not scopes are present.

lib/msal-core/src/Utils.ts
pkanher617: Changed the names

PR #801, opened 2019-07-01
WIP: Revamping and updating samples (includes use cases for external APIs, conditional access)

pkanher617:
This PR is to update our current samples and add new samples for different JS use cases.

We have added and updated the following:
- external API sample
- conditional access sample


pkanher617:
Closing this for now, will update in a separate PR.

PR #792, opened 2019-06-26
Refactor Acquire Token APIs Internally

pkanher617:
This PR is to refactor our code for interactive acquire token calls, which will avoid repetition. There are some considerations to make here, including:

- allowing request.account in login calls
- removal of silentLogin
- remove requirement to call login before acquireToken
- acquireToken and loginInProgress call changes

The above changes can come in separate PRs that will be released in a major version, since these will be breaking changes.

Here is a summary of changes in this PR:
- Refactored all interactive calls to a single function
- fixed bug when sending clientId as only scope (?)
- Renamed isCallback function to isStsResponse
- Added new success and error handlers which will handle all interactive errors (will remove the old success handlers for redirect)
- promptUser function renamed to navigateWindow, now takes a popup window, will navigate popups when given a window object
- Moved around functions

PR #802, opened 2019-07-01
add telemetry event types

DarylThayil:
Add event types that will allow us to start instrumenting code.


lib/msal-core/src/telemetry/ApiEvent.ts
```diff
@@ -0,0 +1,91 @@
...
+const AUTHORITY_KEY = EVENT_NAME_PREFIX + "authority";
+const AUTHORITY_TYPE_KEY = EVENT_NAME_PREFIX + "authority_type";
+const PROMPT_KEY = EVENT_NAME_PREFIX + "ui_behavior";
+const TENANT_ID_KEY = EVENT_NAME_PREFIX + "tenant_id";
+const USER_ID_KEY = EVENT_NAME_PREFIX + "user_id";
+const WAS_SUCESSFUL_KEY = EVENT_NAME_PREFIX + "was_successful";
+const API_ERROR_CODE_KEY = EVENT_NAME_PREFIX + "api_error_code";
+const LOGIN_HINT_KEY = EVENT_NAME_PREFIX + "login_hint";
+
+export const API_CODE = {
```
pkanher617: We could possibly simplify this with the changes in PR #792 so we are just tracking silent and interactive events.

lib/msal-core/src/telemetry/CacheEvent.ts
```diff
@@ -0,0 +1,29 @@
...
+    TokenCacheWrite: `${EVENT_NAME_PREFIX}token_cache_write`,
+    TokenCacheBeforeAccess: `${EVENT_NAME_PREFIX}token_cache_before_access`,
+    TokenCacheAfterAccess: `${EVENT_NAME_PREFIX}token_cache_after_access`,
+    TokenCacheBeforeWrite: `${EVENT_NAME_PREFIX}token_cache_before_write`,
+    TokenCacheDelete: `${EVENT_NAME_PREFIX}token_cache_delete`,
+    TokenType: `${EVENT_NAME_PREFIX}token_type`
+};
+
+export const TOKEN_TYPES = {
+    AT: "at",
```
pkanher617: This is being really nitpicky, but I'd prefer "access" instead of "at", just to be clear.

lib/msal-core/src/telemetry/UiEvent.ts
```diff
@@ -0,0 +1,19 @@
+import TelemetryEvent from "./TelemetryEvent";
+import { EVENT_NAME_PREFIX } from "./TelemetryConstants";
+
+const USER_CANCELLED = EVENT_NAME_PREFIX + "user_cancelled";
+const ACCESS_DENIED = EVENT_NAME_PREFIX + "access_denied";
+
+export default class UiEvent extends TelemetryEvent {
```
pkanher617: Would popup window and iframe events be part of this class?

lib/msal-core/src/telemetry/UiEvent.ts
```diff
@@ -0,0 +1,19 @@
...
+    constructor(correlationId: string) {
+        super(`${EVENT_NAME_PREFIX}ui_event`, correlationId);
+    }
+
+    public set userCancelled(userCancelled: boolean) {
+        this.event[USER_CANCELLED] = userCancelled.toString();
+    }
+
+    public set accessDenied(accessDenied: boolean) {
```
pkanher617: When does this event get triggered?

lib/msal-core/src/telemetry/ApiEvent.ts
DarylThayil: Could you expand a little more :D

lib/msal-core/src/telemetry/CacheEvent.ts
DarylThayil: I think this is what dot net is setting in their telemetry, keeping for consistency. I appreciate nit picks anyway :p

lib/msal-core/src/telemetry/UiEvent.ts
DarylThayil: Potentially, those are not in any other telem examples, but as we instrument, we can add to it

lib/msal-core/src/telemetry/UiEvent.ts
DarylThayil: opposite of user cancelled? the admin required or something?

DarylThayil:
added tests

lib/msal-core/src/telemetry/ApiEvent.ts
pkanher617: #792 changes our internal structure so we just have acquireTokenInteractive and acquireTokenSilent, while the interaction type (popup or redirect) is handled by a UI function, so we have separated duties a little bit. We can simplify our telemetry events by just having "interactive" and "silent" events and just logging whether they are getting an id_token, access_token, and using redirect or popup.

lib/msal-core/src/telemetry/UiEvent.ts
pkanher617: I think we should probably try to add those sooner rather than later, since UI related issues are common.

lib/msal-core/src/telemetry/UiEvent.ts
pkanher617: @sameerag doesn't the PR in #811 fix the issue that occurs when accessDenied is given? If this is something we are detecting for telemetry why aren't we handling it as an error in our library?

lib/msal-core/src/telemetry/CacheEvent.ts
```diff
@@ -0,0 +1,30 @@
+import TelemetryEvent from "./TelemetryEvent";
+import { EVENT_NAME_PREFIX } from "./TelemetryConstants";
+
+export const CACHE_EVENT_TYPES = {
+    TokenCacheLookup: `${EVENT_NAME_PREFIX}token_cache_lookup`,
+    TokenCacheWrite: `${EVENT_NAME_PREFIX}token_cache_write`,
+    TokenCacheBeforeAccess: `${EVENT_NAME_PREFIX}token_cache_before_access`,
+    TokenCacheAfterAccess: `${EVENT_NAME_PREFIX}token_cache_after_access`,
+    TokenCacheBeforeWrite: `${EVENT_NAME_PREFIX}token_cache_before_write`,
+    TokenCacheDelete: `${EVENT_NAME_PREFIX}token_cache_delete`,
```
sameerag: We are appending EVENT_NAME_PREFIX at many places to constant strings. Is there an easier way to do that?

lib/msal-core/src/telemetry/HttpEvent.ts
```diff
@@ -0,0 +1,73 @@
+import TelemetryEvent from "./TelemetryEvent";
+import { EVENT_NAME_PREFIX } from "./TelemetryConstants";
+import { QPDict } from "../AuthenticationParameters";
```
sameerag: This will change with PR #804. Maybe StringDict?

lib/msal-core/src/telemetry/UiEvent.ts
sameerag: @pkanher617 We can. For access_denied we may need to read the sub-error sent by the server team, apparently, their messaging is vague intentionally as a security smoke screen. I am following up with Adrian Frei on what would be an appropriate error messaging in this case. Will close loop once I have their response.

lib/msal-core/src/telemetry/ApiEvent.ts
```diff
@@ -0,0 +1,92 @@
...
+export const API_ERROR_CODE_KEY = EVENT_NAME_PREFIX + "api_error_code";
+export const LOGIN_HINT_KEY = EVENT_NAME_PREFIX + "login_hint";
+
+export const API_CODE = {
+    AcquireTokenRedirect: 2001,
+    AcquireTokenSilent: 2002,
+    AcquireTokenPopup: 2003,
+    LoginRedirect: 2004,
+    LoginPopup: 2005
+};
```
jasonnutter: This could be an enum instead of an object (which allows for more type safety).

lib/msal-core/src/telemetry/ApiEvent.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export const API_EVENT_IDENTIFIER = {
+    AcquireTokenRedirect: "AcquireTokenRedirect",
+    AcquireTokenSilent: "AcquireTokenSilent",
+    AcquireTokenPopup: "AcquireTokenPopup",
+    LoginRedirect: "LoginRedirect",
+    LoginPopup: "LoginPopup"
+};
```
jasonnutter: This can also be an enum.

lib/msal-core/src/telemetry/ApiEvent.ts
```diff
@@ -0,0 +1,92 @@
...
+        this.event[USER_ID_KEY] = this.logger.piiLoggingIsEnabled() && accountId ?
+            hashPersonalIdentifier(accountId)
+            : null;
+    }
+
+    public set wasSuccessful(wasSuccessful: boolean) {
+        this.event[WAS_SUCESSFUL_KEY] = wasSuccessful;
+    }
+
+    public get wasSuccessful() {
```
jasonnutter: Nitpick: `isSuccessful` or just `successful`.

lib/msal-core/src/telemetry/ApiEvent.ts
```diff
@@ -0,0 +1,92 @@
...
+        this.logger = logger;
+    }
+
+    public set apiEventIdentifier(apiEventIdentifier: string) {
+        this.event[TELEMETRY_BLOB_EVENT_NAMES.ApiTelemIdConstStrKey] = apiEventIdentifier;
+    }
+
+    public set apiCode(apiCode: number) {
+        this.event[TELEMETRY_BLOB_EVENT_NAMES.ApiIdConstStrKey] = apiCode;
+    }
```
jasonnutter: General comment: if you `return this;` in all of these setters it allows for chaining.

lib/msal-core/src/telemetry/ApiEvent.ts
```diff
@@ -0,0 +1,92 @@
...
+
+// export these to test... could namespace them, or find another way?
+export const AUTHORITY_KEY = EVENT_NAME_PREFIX + "authority";
+export const AUTHORITY_TYPE_KEY = EVENT_NAME_PREFIX + "authority_type";
+export const PROMPT_KEY = EVENT_NAME_PREFIX + "ui_behavior";
+export const TENANT_ID_KEY = EVENT_NAME_PREFIX + "tenant_id";
+export const USER_ID_KEY = EVENT_NAME_PREFIX + "user_id";
+export const WAS_SUCESSFUL_KEY = EVENT_NAME_PREFIX + "was_successful";
+export const API_ERROR_CODE_KEY = EVENT_NAME_PREFIX + "api_error_code";
+export const LOGIN_HINT_KEY = EVENT_NAME_PREFIX + "login_hint";
```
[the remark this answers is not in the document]
jasonnutter: Since you can't use an enum with a computer value, I think an object would be better:

```js
export const EVENT_KEYS = {
   AUTHORITY: `${EVENT_NAME_PREFIX}authority`,
  ...
};
```

lib/msal-core/src/telemetry/ApiEvent.ts
```diff
@@ -0,0 +1,92 @@
...
+    AcquireTokenSilent: "AcquireTokenSilent",
+    AcquireTokenPopup: "AcquireTokenPopup",
+    LoginRedirect: "LoginRedirect",
+    LoginPopup: "LoginPopup"
+};
+
+
+export default class ApiEvent extends TelemetryEvent {
+
+    private logger: Logger;
```
jasonnutter: When we tiddy up the lint rules we should decide on whether this should be `this.logger` or `this._logger`.

lib/msal-core/src/telemetry/ApiEvent.ts
```diff
@@ -0,0 +1,92 @@
...
+            hashPersonalIdentifier(accountId)
+            : null;
+    }
+
+    public set wasSuccessful(wasSuccessful: boolean) {
+        this.event[WAS_SUCESSFUL_KEY] = wasSuccessful;
+    }
+
+    public get wasSuccessful() {
+        return this.event[WAS_SUCESSFUL_KEY] === true;
```
jasonnutter: Don't  need `=== true` part.

jasonnutter: Unless if you are expecting this to be potentially be a truthy non-boolean value (which doesn't appear to be the case based on the setter).

lib/msal-core/src/telemetry/HttpEvent.ts
```diff
@@ -0,0 +1,73 @@
...
+export const USER_AGENT_KEY = `${EVENT_NAME_PREFIX}user_agent`;
+export const QUERY_PARAMETERS_KEY = `${EVENT_NAME_PREFIX}query_parameters`;
+export const API_VERSION_KEY = `${EVENT_NAME_PREFIX}api_version`;
+export const RESPONSE_CODE_KEY = `${EVENT_NAME_PREFIX}response_code`;
+export const O_AUTH_ERROR_CODE_KEY = `${EVENT_NAME_PREFIX}oauth_error_code`;
+export const HTTP_METHOD_KEY = `${EVENT_NAME_PREFIX}http_method`;
+export const REQUEST_ID_HEADER_KEY = `${EVENT_NAME_PREFIX}request_id_header`;
+export const SPE_INFO_KEY = `${EVENT_NAME_PREFIX}spe_info`;
+export const SERVER_ERROR_CODE_KEY = `${EVENT_NAME_PREFIX}server_error_code`;
+export const SERVER_SUB_ERROR_CODE_KEY = `${EVENT_NAME_PREFIX}server_sub_error_code`;
```
jasonnutter: Same comment about having be an object.

lib/msal-core/src/telemetry/CacheEvent.ts
```diff
@@ -0,0 +1,30 @@
...
+};
+
+export const TOKEN_TYPES = {
+    AT: "at",
+    ID: "id",
+    ACCOUNT: "account"
+};
```
jasonnutter: This can also be an enum.

lib/msal-core/src/telemetry/ApiEvent.ts
jasonnutter: I would understand teh `set` function if you had matching `get` functions, but since you don't, I don't think a lot of value is added by having them be `set` functions.

jasonnutter:
Could you also please update teh description with a summary of the changes?

lib/msal-core/src/telemetry/ApiEvent.ts
DarylThayil: I think it is valuable to track both, Api events will tell us what users are using

lib/msal-core/src/telemetry/UiEvent.ts
DarylThayil: if we do not see this from server that is fine, for now it is supported in the Event

lib/msal-core/src/telemetry/ApiEvent.ts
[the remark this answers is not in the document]
DarylThayil: fixing

lib/msal-core/src/telemetry/ApiEvent.ts
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    public set loginHint(loginHint: string) {
```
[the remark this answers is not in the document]
DarylThayil: Specifically on the telemetry event they are 3 different fields. 
I think having the even setters not contain any of the logic makes sense.  

When setting the event properties is when we would determine is it loginHint, accountId, or tenantId we are setting

lib/msal-core/src/telemetry/CacheEvent.ts
DarylThayil: yes! utilitizing

DarylThayil: added util fixed everywhere, good suggestion

lib/msal-core/src/telemetry/HttpEvent.ts
DarylThayil: removing

lib/msal-core/src/telemetry/ApiEvent.ts
DarylThayil: following naming from other msal telemtries

lib/msal-core/src/telemetry/ApiEvent.ts
DarylThayil: updating

lib/msal-core/src/telemetry/ApiEvent.ts
DarylThayil: updating

lib/msal-core/src/telemetry/ApiEvent.ts
DarylThayil: so the usage i see is something like this 

```js
const event = new HttpEvent(params);
TelemetryManager.startEvent(Event);

event.url = url;

event.
fetch(url).then(res => {
    event.httpResponse = 200;
})
.catch(err  => {
    event.errorCode = "FAIL"
    event.httpResponse = 500;
})

TelemetryManager.endEvent(event);

TelemetryManager.flush();


```


I think set is kinda shorthand for setter functions, no?

lib/msal-core/src/telemetry/ApiEvent.ts
DarylThayil: yeah, probably not necessary, just wanted to make sure that in case something get set there, that what we are returning is an actual true

lib/msal-core/src/telemetry/HttpEvent.ts
DarylThayil: fixing

lib/msal-core/src/telemetry/HttpEvent.ts
```diff
@@ -0,0 +1,73 @@
...
+    public set queryParams(queryParams: QPDict) {
+        this.event[QUERY_PARAMETERS_KEY] = Utils.generateQueryParametersString(queryParams);
+    }
+
+    public set apiVersion(apiVersion: string) {
+        this.event[API_VERSION_KEY] = apiVersion.toLowerCase();
+    }
+
+    public set httpResponseStatus(statusCode: number) {
+        this.event[RESPONSE_CODE_KEY] = statusCode;
```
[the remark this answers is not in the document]
DarylThayil: undecided to benefit of setters / chaining, willing to be convinced

lib/msal-core/src/telemetry/HttpEvent.ts
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    public set requestIdHeader(requestIdHeader: string) {
+        this.event[REQUEST_ID_HEADER_KEY] = requestIdHeader;
+    }
+
+    ///  Indicates whether the request was executed on a ring serving SPE traffic.
+    ///  An empty string indicates this occurred on an outer ring, and the string "I"
+    ///  indicates the request occurred on the inner ring
+    public set speInfo(speInfo: string) {
```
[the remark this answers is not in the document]
DarylThayil: When I go to instrument, I may do that

lib/msal-core/src/telemetry/CacheEvent.ts
DarylThayil: adding

lib/msal-core/src/telemetry/ApiEvent.ts
jasonnutter: I believe it would be the difference between that and this: 

```js
event.setErrorCode("FAIL").setHttpResponse(500);
```

Not a big deal, ultimately.

lib/msal-core/src/telemetry/UiEvent.ts
pkanher617: Ok just keep it in mind as we move forward.

lib/msal-core/src/telemetry/HttpEvent.ts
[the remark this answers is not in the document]
pkanher617: While I don't feel strongly about this one way or the other, and there are some benefits to chaining,  I think it is better to follow JS Best practices here. It may mess up inheritance and in general doesn't follow command-query separation principle, so I would suggest using setters.

lib/msal-core/src/telemetry/ApiEvent.ts
pkanher617: Same comment as before:

I think it is better to follow JS Best practices here. Implementing chaining may mess up inheritance and in general doesn't follow command-query separation principle, so I would suggest using setters.

lib/msal-core/src/telemetry/HttpEvent.ts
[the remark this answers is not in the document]
DarylThayil: As an internal implementation, I am fine shipping this, and if we want to update implementation to be cleaner I am fine changing down the road

lib/msal-core/src/telemetry/UiEvent.ts
DarylThayil: Will add as we start to instrument the apis

lib/msal-core/src/telemetry/ApiEvent.ts
DarylThayil: internal implementation developers never touch, so if we decide to change internally we can. leaving for now

lib/msal-core/src/Logger.ts
```diff
@@ -143,4 +143,8 @@ export class Logger {// Singleton Class
   verbosePii(message: string): void {
     this.logMessage(LogLevel.Verbose, message, true);
   }
+
+  piiLoggingIsEnabled(): boolean {
```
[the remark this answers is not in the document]
DarylThayil: updated

PR #804, opened 2019-07-03
Adding IdToken claims to MSAL JS interfaces : AuthResponse and Account

sameerag:
This PR:

- adds all the claims in an id_token in the API surface for the developer reference; idToken in 'Account' is to map a request to the corresponding session. This PR deprecates the current parameter 'idToken' in the 'Account' class and adds a more meaningful parameter: 'idTokenClaims'. Since both are same, for backward compatibility until 1.1.0, both the parameters are supported.
- adds 'idTokenClaims' in response to map the claims within a request-response cycle.
- remove dependency on response.idToken in UserAgentApplication code.

Relevant changes:
- [x] Corresponding changes in response in UserAgentApplication
- [x] Unit tests added for claims; more enhancement in UserAgentApplication test cases
- [x] Unit tests added for IdToken and Account classes

Todo:

- [ ] Documentation changes needed for the Surface changes will be edited in the Readme once this PR is ready to be Merged. 
- [ ] Today we have two different data types for IdToken and AccessToken. Ideally, both should have been tokens (JWT format as referenced in Microsoft AAD docs). Since this is a breaking change, we are preserving the IdToken type in the response for this PR but will track this in the future major release.


lib/msal-core/src/Account.ts
```diff
@@ -20,7 +21,9 @@ export class Account {
     homeAccountIdentifier: string;
     userName: string;
     name: string;
-    idToken: Object;
+    // will be deprecated soon
+    idToken: Dict;
```
pkanher617: We will be changing idToken to a string in a future release correct?

lib/msal-core/src/AuthResponse.ts
```diff
@@ -12,6 +13,7 @@ export type AuthResponse = {
     tenantId: string;
     tokenType: string;
     idToken: IdToken;
```
pkanher617: We will be changing idToken to a string in a future release correct?

sameerag: Yes, in a "major" release as this will be a breaking change.

lib/msal-core/src/MsalTypes.ts
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+/**
+ * Key-Value type to support queryParams and extraQueryParams
+ */
+export type Dict = {[key: string]: string};
```
jasonnutter: I don't think `Dict` is specific enough, I would call this `StringDict` or something to make it clear values can only be strings.

sameerag: Agreed. This is something I wanted to change. Any standard naming for this "type" in JS world?

jasonnutter: Not sure, I'm kinda surprised Typescript doesn't have a built-in type you can use for this.

jasonnutter: Looks like `Record<string, string>` would work? https://www.typescriptlang.org/docs/handbook/advanced-types.html#mapped-types

DarylThayil: +1 if TS already has a type for this.

sameerag: Record<string, string> may be overkill. As per this SO [post]https://stackoverflow.com/questions/51936369/what-is-the-record-type-in-typescriptl) it  is better suited for not just string but any datatype and a great way to get typesafety in TS in usecases mentioned in the post. We are looking at a free flowing map<string, string>.

Naming it StringDict<>.

PR #805, opened 2019-07-03
Split Utils.ts into different files

pkanher617:
PR #792 must be approved before this one can be merged.

This PR splits the massive Utils.ts file into separate smaller files with well-defined task sets. Eventually these files can lend a more object-oriented code style.

Newly-created files from Utils.ts:
- ScopeSet.ts (this will eventually be an object representation of scopes)
- TokenProcessor.ts (handles most token operations, we may want to start coming up with internal structure for token objects?)
- UrlProcessor (this will handle the url operations, eventually should move the createNavigateUrls to this file)

Major modified files:
- AuthResponse (now exports a function originally in Utils.ts)
- UserAgentApplication (replaced Utils with new calls)

Other file imports were fixed as well.


DarylThayil:
for all utils you are splitting up, can they go in a utils folder?

/Utils/UrlUtils
/Utils/ScopeUtils 
etc...

DarylThayil:
any logic changes or Just moving files?

lib/msal-core/src/AuthResponse.ts
```diff
@@ -32,3 +32,15 @@ export function buildResponseStateOnly(state: string) : AuthResponse {
         accountState: state
     };
 }
+
+export function setResponseIdToken(originalResponse: AuthResponse, idToken: IdToken) : AuthResponse {
```
sameerag: Can we move this to a different file? Restricting AuthResponse.ts is for just the response type may be more ideal going forward. Create a new file if needed.

lib/msal-core/src/ScopeSet.ts
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+export class ScopeSet {
```
sameerag: Better name? May be 'ScopesUtils'?

lib/msal-core/src/ScopeSet.ts
```diff
@@ -0,0 +1,54 @@
...
+    return scopes.every((value: any): boolean => cachedScopes.indexOf(value.toString().toLowerCase()) >= 0);
+  }
+
+  /**
+   * toLower
+   *
+   * @param scopes
+   */
+  // TODO: Rename this, too generic name for a function that only deals with scopes
+  static convertToLowerCase(scopes: Array<string>): Array<string> {
```
sameerag: rename please as per the comment.

lib/msal-core/src/ScopeSet.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  /**
+   * remove one element from a scope array
+   *
+   * @param scopes
+   * @param scope
+   */
+  // TODO: Rename this, too generic name for a function that only deals with scopes
+  static removeElement(scopes: Array<string>, scope: string): Array<string> {
```
sameerag: Same as above, rename conveying scope parsing, Or better yet since it is part of the class, if the class is renamed,  maybe we can leave this alone.

lib/msal-core/src/AuthResponse.ts
pkanher617: I disagree, this is a helper for AuthResponse and naturally fits in this class. Too many files will start cluttering the workspace.

lib/msal-core/src/ScopeSet.ts
pkanher617: This file will be an object that represents the set of scopes passed to MSAL by the developer. Ideally it will perform all scope related functions, and act upon its own object, so we won't need a ScopesUtils.

lib/msal-core/src/ScopeSet.ts
pkanher617: Just moving functions first to keep git diff similar. I will rename in a future PR.

lib/msal-core/src/ScopeSet.ts
pkanher617: Just moving functions first to keep git diff similar. I will rename in a future PR.

lib/msal-core/src/AuthResponse.ts
pkanher617: Make ResponseUtils

lib/msal-core/src/utils/Utils.ts
```diff
@@ -1,14 +1,10 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
```
jasonnutter: What about naming this `utils/index.ts`?

lib/msal-core/src/utils/UrlUtils.ts
```diff
@@ -0,0 +1,138 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
jasonnutter: I'd argue you don't need to include `Utils` in the name, since its in the path already: `utils/Url.ts`

lib/msal-core/src/utils/Utils.ts
pkanher617: This file exports the Utils class object, which is why it is named Utils.ts. Do we turn all of the functions into globals?

lib/msal-core/src/utils/UrlUtils.ts
pkanher617: I'd prefer to name files what they are, regardless of folder. It is easier to read in stacktraces and such, but I'm open to this if the team feels it is more concise.

lib/msal-core/src/AadAuthority.ts
```diff
@@ -3,7 +3,7 @@
 
 import { Authority, AuthorityType } from "./Authority";
 import { XhrClient } from "./XHRClient";
-import { AADTrustedHostList } from "./Constants";
+import { AADTrustedHostList } from "./utils/Constants";
```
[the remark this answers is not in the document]
pkanher617: We can refactor this later, but for now I have just moved the constants file.

PR #806, opened 2019-07-03
Refactor ServerRequestParameters and UrlProcessor to perform functions themselves, instead of in UserAgentApplication

pkanher617:
PR #805 should be reviewed before this PR.

This PR is to refactor functions in Utils, ServerRequestParameters and userAgentApplication to be in classes which better fit their description.

Tests have also been updated.


lib/msal-core/src/ScopeSet.ts
```diff
@@ -51,4 +51,19 @@ export class ScopeSet {
   static removeElement(scopes: Array<string>, scope: string): Array<string> {
     return scopes.filter(value => value !== scope);
   }
+
+  /**
+   * Parse the scopes into a formatted scopeList
+   * @param scopes
+   */
+  static parseScope(scopes: Array<string>): string {
+    let scopeList: string = "";
```
DarylThayil: return scopes.reduce((memo, val, index) => (
    ${memo}${index !== scopes.length - 1 ? `${val} `: val }`
), "")

jasonnutter: Yeah was going comment same thing, but this is just copy/pasta, so I'm cool with not touching this as part of this PR.

pkanher617: +1, this will come in the scopes refactor.

lib/msal-core/test/telemetry/HttpEvent.spec.ts
```diff
@@ -50,7 +50,7 @@ describe("HttpEvent", () => {
         expect(event[EVENT_KEYS.SERVER_SUB_ERROR_CODE]).to.eq(fakeSubErrorCode);
     });
 
-    it("sets values that are changed", () => {
+    it.skip("sets values that are changed", () => {
```
jasonnutter: Intentional?

pkanher617: nope. fixing now.

jasonnutter: I'll make sure we get a lint rule to catch this.

lib/msal-core/src/telemetry/HttpEvent.ts
```diff
@@ -33,7 +33,7 @@ export default class HttpEvent extends TelemetryEvent {
     }
 
     public set queryParams(queryParams: any) {
-        this.event[EVENT_KEYS.QUERY_PARAMETERS] = Utils.generateQueryParametersString(queryParams);
+        // this.event[EVENT_KEYS.QUERY_PARAMETERS] = Utils.generateQueryParametersString(queryParams);
```
jasonnutter: Can you add a comment about why this is commented out (I assume this is temporary)?

lib/msal-core/test/telemetry/HttpEvent.spec.ts
pkanher617: Actually never mind - yes, this is intentional. The refactor changes break the functionality in this test.

lib/msal-core/src/telemetry/HttpEvent.ts
pkanher617: The refactor changed the visibility of a function in the telemetry work. I wanted @DarylThayil to take a look at this.

DarylThayil: why can't that function be public

lib/msal-core/test/telemetry/HttpEvent.spec.ts
DarylThayil: I would prefer we not skip tests, either delete em or keep em. 
Tough to tell when a test is skipped whether it should be fixed, or deleted, is there a hole in coverage etc.

lib/msal-core/src/telemetry/HttpEvent.ts
pkanher617: It is a member of serverRequestParameters class now. I'd prefer to keep it private so we don't start using the functionality outside. Is there another way to represent this event?

lib/msal-core/test/telemetry/HttpEvent.spec.ts
pkanher617: This is skipped temporarily until we resolve the issue above. Skipping allows us to at least see that we have tests that aren't being run.

lib/msal-core/src/telemetry/HttpEvent.ts
pkanher617: I've changed generateQueryParameters to a static function, can we get this merged?

PR #811, opened 2019-07-05
Handle null errorDesc for a few Server Errors

sameerag:
When the user chooses to go back to the home page without proceeding with authentication flow in some usecases, we see improper handling of error from the server in the msal js code. Two of those cases are:

1. When the user chooses not to grant consent; The Popup has an option to “Return to the application without granting consent”
2. The user clicks “Back” when the multi-account selection tile appears in acquireTokenPopup(); 

This is because the service is not populating the error description in these specific cases.
While we discuss how to handle this case with more meaningful messaging from the service, we can still handle it gracefully and throw a "ServerError" from msal js.
 
Fix includes:

- Graceful handling of null error description
- Unit test cases for this use case


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1805,7 +1805,10 @@ export class UserAgentApplication {
         acquireTokenAccountKey = Storage.generateAcquireTokenAccountKey(accountId, stateInfo.state);
       }
 
-      if (this.isInteractionRequired(hashParams[Constants.error]) || this.isInteractionRequired(hashParams[Constants.errorDescription])) {
+      const hashErr = hashParams[Constants.error];
+      const hashErrDesc = hashParams[Constants.errorDescription];
```
jasonnutter: Cool ES6 trick:

```js
const {
  [Constants.error]: hashErr,
  [Constants.errorDescription]: hasErrDesc
} = hashParams;
```

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1805,7 +1805,10 @@ export class UserAgentApplication {
...
+
+      if ((hashErr && this.isInteractionRequired(hashErr)) || (hashErrDesc && this.isInteractionRequired(hashErrDesc))) {
```
DarylThayil: Small nitpick, could pull predicate into a variable for readability

jasonnutter: Could `this.isInteractionRequired` be updated to check the truthiness of whatever is passed to it, that way you don't have to do the truthiness check here?

pkanher617: +1, I think we should check in the function rather than beforehand

lib/msal-core/src/UserAgentApplication.ts
sameerag: Done! Thanks.

lib/msal-core/src/UserAgentApplication.ts
sameerag: Please review the changes. Added this.

sameerag: Please review. Added this.

sameerag:
Incorporated the feedback and also simplified some code. Please review. Would like to merge this today for a release early next week.

lib/msal-core/src/Constants.ts
```diff
@@ -133,6 +141,19 @@ export const PromptState = {
...
+ */
 export const Library = {
   version: "1.0.2"
 };
+
+/**
+ * @hidden
+ * Readable shortcut for Interaction Required Errors
+ */
+export const InteractionErrorType = {
```
DarylThayil: could this be made an enum?

lib/msal-core/test/UserAgentApplication.spec.ts
```diff
@@ -881,6 +884,24 @@ describe("UserAgentApplication.ts Class", function () {
...
+                expect(error.name).to.include("ServerError");
+                expect(error.errorCode).to.include(TEST_ACCESS_DENIED);
+                expect(error.stack).to.include("UserAgentApplication.spec.js");
+                done();
+            };
+            msal.handleRedirectCallback(checkErrorFromServer);
+        });
+
+
+
```
DarylThayil: only 1 empty linebetween

sameerag: Done and pushed the changes.

lib/msal-core/src/Constants.ts
sameerag: More for consistency sake. I can push a PR right after making enums the NORM in constants file if that is what we want to do. I do not want to break consistency, it makes the code look ugly if you see what I mean :)

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1016,10 +1017,10 @@ export class UserAgentApplication {
...
-    if (errorString.indexOf("interaction_required") !== -1 ||
-    errorString.indexOf("consent_required") !== -1 ||
-    errorString.indexOf("login_required") !== -1) {
-      return true;
+
+    const errorTypes = [InteractionErrorType.INTERACTION, InteractionErrorType.CONSENT, InteractionErrorType.LOGIN];
+    if (errorString && errorTypes.indexOf(errorString) > -1) {
+        return true;
     }
     return false;
```
jasonnutter: Can be simplified to `return errorString && errorTypes.indexOf(errorString) > -1`

sameerag: Done.

sameerag:
Merging. Incorporated all the feedback.

PR #812, opened 2019-07-08
WIP: Initial proof of possesion implementation

jasonnutter:
Initial implementation for proof of possession tokens.

PoP flow:

1. When the client application wants an access token, it can indicate that it wants an access token protected by PoP by passing `authenticationType: 'pop'` in the request object. It will also be required to provide the request uri and method that the access token will be used for.
2. If that value is set, `UserAgentApplication` will use `PopTokenGenerator.getNewToken()` to generate a new pop token. This will store the public/private keys in the Pop cache, and return back to `UserAgentApplication` the value to be used for the `pop_jwk` query parameter in `ServerRequestParameters` that needs to be added to access tokens request to receive an access token that is of type `pop`. This string is intended to be opaque to `UserAgentApplication`.
3. When the access token response is received, a new field will be added to the token cache called `accessTokenType`, will be either `pop` or `bearer`. This is so that when the token is retrieved from the cache, we can tell whether or not it is a pop token without inspecting the token itself.
4. When we read the access token out of the cache, we check the `accessTokenType` field, and if it is a pop token, create a new wrapping token that contains the original access token and is signed with the private key that was used to generate the original request.

TODO:

- [ ] Check that the signed tokens are in the correct format
- [ ] Complete error handling
- [ ] Telemetry?
- [ ] Test with mobile devices
- [ ] Documentation
- [ ] Samples
- [ ] Comments

### DIFF
--- a/lib/msal-core/package.json
+++ b/lib/msal-core/package.json
@@ -29,6 +29,7 @@
     "clean": "shx rm -rf dist docs lib-commonjs lib-es6",
     "doc": "typedoc --mode modules --excludePrivate --excludeProtected --out ./docs ./src/ --gitRevision dev",
     "build:modules": "tsc && tsc -m es6 --outDir lib-es6",
+    "build:modules:watch": "tsc -m es6 --outDir lib-es6 --watch",
     "build": "npm run clean && npm run doc && npm run build:modules && npm run lint && npm run build:webpack && npm test",
     "test": "karma start --single-run --browsers PhantomJS karma.conf.js",
     "lint": "tslint -c tslint.json src/**/*.ts",
--- a/lib/msal-core/src/AccessTokenValue.ts
+++ b/lib/msal-core/src/AccessTokenValue.ts
@@ -7,12 +7,14 @@
 export class AccessTokenValue {
 
   accessToken: string;
+  accessTokenType: string;
   idToken: string;
   expiresIn: string;
   homeAccountIdentifier: string;
 
-  constructor(accessToken: string, idToken: string, expiresIn: string, homeAccountIdentifier: string) {
+  constructor(accessToken: string, idToken: string, expiresIn: string, homeAccountIdentifier: string, accessTokenType: string) {
     this.accessToken = accessToken;
+    this.accessTokenType = accessTokenType;
     this.idToken = idToken;
     this.expiresIn = expiresIn;
     this.homeAccountIdentifier = homeAccountIdentifier;
--- a/lib/msal-core/src/AuthenticationParameters.ts
+++ b/lib/msal-core/src/AuthenticationParameters.ts
@@ -9,6 +9,11 @@ import { ClientConfigurationError } from "./error/ClientConfigurationError";
  */
 export type QPDict = {[key: string]: string};
 
+export enum AuthenticationType {
+    PoP = 'pop',
+    Bearer = 'bearer'
+}
+
 /**
  * @link AuthenticationParameters}AuthenticationParameters
  */
@@ -24,6 +29,9 @@ export type AuthenticationParameters = {
     account?: Account;
     sid?: string;
     loginHint?: string;
+    requestMethod?: string;
+    requestUri?: string;
+    authenticationType?: AuthenticationType;
 };
 
 export function validateClaimsRequest(request: AuthenticationParameters) {
--- a/lib/msal-core/src/BrowserCrypto.ts
+++ b/lib/msal-core/src/BrowserCrypto.ts
@@ -1,0 +1,134 @@
+import { KeyGenAlgorithm, HashAlgorithm, KeyFormat, getArrayBufferFromString, getStringFromArrayBuffer, isIE11, getJwkString } from './PopTokenCommon';
+
+/**
+ * Thin wrapper around browser Web Crypto APIs.
+ * Docs: https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto
+ *
+ * Includes support for IE11 msCrypto methods.
+ * Docs: https://msdn.microsoft.com/en-us/ie/dn904640(v=vs.94)
+ */
+export class BrowserCrypto {
+    private _algorithmOptions: RsaHashedKeyGenParams;
+
+    constructor(keyGenAlgorithm: KeyGenAlgorithm, hashAlgorithm: HashAlgorithm, modulusLength: number, publicExponent: Uint8Array) {
+        this._algorithmOptions = {
+            name: keyGenAlgorithm,
+            hash: hashAlgorithm,
+            modulusLength,
+            publicExponent
+        }
+    }
+
+    async sign(key: CryptoKey, data: ArrayBuffer): Promise<ArrayBuffer> {
+        if (isIE11()) {
+            return new Promise((resolve, reject) => {
+                const signOperation = window['msCrypto'].subtle.sign(this._algorithmOptions, key, data);
+                signOperation.addEventListener('complete', (e: { target: { result: ArrayBuffer | PromiseLike<ArrayBuffer>; }; }) => {
+                    resolve(e.target.result);
+                });
+                signOperation.addEventListener('error', (error: any) => {
+                    reject(error);
+                });
+            });
+        }
+
+        return window.crypto.subtle.sign(this._algorithmOptions, key, data);
+    }
+
+    async verify(key: CryptoKey, signature: ArrayBuffer, data: ArrayBuffer): Promise<boolean> {
+        if (isIE11()) {
+            return new Promise((resolve, reject) => {
+                const verifySignatureOperation = window['msCrypto'].subtle.verify(this._algorithmOptions, key, signature, data);
+                verifySignatureOperation.addEventListener('complete', (e: { target: { result: boolean | PromiseLike<boolean>; }; }) => {
+                    resolve(e.target.result);
+                });
+                verifySignatureOperation.addEventListener('error', (error: any) => {
+                    reject(error);
+                });
+            });
+        }
+
+        return window.crypto.subtle.verify(this._algorithmOptions, key, signature, data);
+    }
+
+    async generateKey(extractable: boolean, usages: Array<string>): Promise<CryptoKeyPair> {
+        if (isIE11()) {
+            return new Promise((resolve, reject) => {
+                const generateKeyOperation = window['msCrypto'].subtle.generateKey(this._algorithmOptions, extractable, usages);
+                generateKeyOperation.addEventListener('complete', (e: { target: { result: CryptoKeyPair | PromiseLike<CryptoKeyPair>; }; }) => {
+                    resolve(e.target.result);
+                });
+                generateKeyOperation.addEventListener('error', (error: any) => {
+                    reject(error);
+                });
+            });
+        }
+
+        return window.crypto.subtle.generateKey(this._algorithmOptions, extractable, usages);
+    }
+
+    async exportKey(key: CryptoKey, format: KeyFormat): Promise<JsonWebKey> {
+        if (isIE11()) {
+            return new Promise((resolve, reject) => {
+                const exportKeyOperation = window['msCrypto'].subtle.exportKey(format, key);
+                exportKeyOperation.addEventListener('complete', (e: { target: { result: ArrayBuffer; }; }) => {
+                    const resultBuffer: ArrayBuffer = e.target.result;
+                    const resultString = getStringFromArrayBuffer(resultBuffer)
+                        .replace(/\r/g, '')
+                        .replace(/\n/g, '')
+                        .replace(/\t/g, '')
+                        .split(' ').join('')
+                        .replace("\u0000", '');
+
+                    try {
+                        resolve(JSON.parse(resultString));
+                    } catch (e) {
+                        reject(e);
+                    }
+                });
+                exportKeyOperation.addEventListener('error', (error: any) => {
+                    reject(error);
+                })
+            });
+        }
+
+        return window.crypto.subtle.exportKey(format, key);
+    }
+
+    async importKey(key: JsonWebKey, format: KeyFormat, extractable: boolean, usages: Array<string>): Promise<CryptoKey> {
+        if (isIE11()) {
+            const keyString = getJwkString(key);
+            const keyBuffer = getArrayBufferFromString(keyString);
+
+            return new Promise((resolve, reject) => {
+                const importOperation = window['msCrypto'].subtle.importKey(format, keyBuffer, this._algorithmOptions, extractable, usages);
+                importOperation.addEventListener('complete', (e: { target: { result: CryptoKey | PromiseLike<CryptoKey>; }; }) => {
+                    resolve(e.target.result);
+                });
+                importOperation.addEventListener('error', (error: any) => {
+                    reject(error);
+                });
+            });
+        }
+
+        return window.crypto.subtle.importKey(format, key, this._algorithmOptions, extractable, usages);
+    }
+
+    async digest(algo: HashAlgorithm, dataString: string): Promise<ArrayBuffer> {
+        const data = getArrayBufferFromString(dataString);
+
+        if (isIE11()) {
+            return new Promise((resolve, reject) => {
+                const digestOperation = window['msCrypto'].subtle.digest(algo, data);
+                digestOperation.addEventListener('complete', (e: { target: { result: ArrayBuffer | PromiseLike<ArrayBuffer>; }; }) => {
+                    resolve(e.target.result);
+                });
+                digestOperation.addEventListener('error', (error: any) => {
+                    reject(error);
+                });
+            });
+        }
+
+        return window.crypto.subtle.digest(algo, data);
+    }
+}
--- a/lib/msal-core/src/DatabaseCache.ts
+++ b/lib/msal-core/src/DatabaseCache.ts
@@ -1,0 +1,65 @@
+interface IDBOpenDBRequestEvent extends Event {
+    target: IDBOpenDBRequest & EventTarget;
+}
+
+interface IDBOpenOnUpgradeNeededEvent extends IDBVersionChangeEvent {
+    target: IDBOpenDBRequest & EventTarget;
+}
+
+interface IDBRequestEvent extends Event {
+    target: IDBRequest & EventTarget;
+}
+
+export class DatabaseCache<T> {
+    private _db : IDBDatabase;
+    private _dbName: string;
+    private _tableName: string;
+    private _version: number;
+
+    constructor(dbName: string, tableName: string, version: number) {
+        this._dbName = dbName;
+        this._tableName = tableName;
+        this._version = version;
+    }
+
+    async open(): Promise<void> {
+        const openOperation = window.indexedDB.open(this._dbName, this._version);
+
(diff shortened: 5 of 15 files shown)


lib/msal-core/src/PopTokenGenerator.ts
```diff
@@ -0,0 +1,153 @@
...
+            privateKey: unextractablePrivateKey,
+            requestMethod: request.requestMethod,
+            requestUri: request.requestUri
+        });
+
+        const popTokenString = JSON.stringify({
+            kid: publicKeyHash
+        })
+
+        return popTokenString;
```
jasonnutter: I'm thinking that this method should catch any errors and return back `null` if so, that way `UserAgentApplication` falls back gracefully if generating a pop token goes wrong (e.g. crypto isn't supported on the browser, etc).

juho-hanhimaki:
Sounds interesting. Is there end-to-end support for this? Get token from Azure AD and use it to authenticate to ASP.NET Core API?

lib/msal-core/package.json
```diff
@@ -29,6 +29,7 @@
     "clean": "shx rm -rf dist docs lib-commonjs lib-es6",
     "doc": "typedoc --mode modules --excludePrivate --excludeProtected --out ./docs ./src/ --gitRevision dev",
     "build:modules": "tsc && tsc -m es6 --outDir lib-es6",
+    "build:modules:watch": "tsc -m es6 --outDir lib-es6 --watch",
```
DarylThayil: just so I understand, as you make changes this will update the build files?

jasonnutter: Yep.

lib/msal-core/src/AccessTokenValue.ts
```diff
@@ -7,12 +7,14 @@
 export class AccessTokenValue {
 
   accessToken: string;
+  accessTokenType: string;
   idToken: string;
   expiresIn: string;
   homeAccountIdentifier: string;
 
-  constructor(accessToken: string, idToken: string, expiresIn: string, homeAccountIdentifier: string) {
+  constructor(accessToken: string, idToken: string, expiresIn: string, homeAccountIdentifier: string, accessTokenType: string) {
```
DarylThayil: is type always provided?

lib/msal-core/src/AuthenticationParameters.ts
```diff
@@ -9,6 +9,11 @@ import { ClientConfigurationError } from "./error/ClientConfigurationError";
  */
 export type QPDict = {[key: string]: string};
 
+export enum AuthenticationType {
+    PoP = 'pop',
```
DarylThayil: Is this case consistent?

DarylThayil: +1 on using enums

lib/msal-core/src/BrowserCrypto.ts
```diff
@@ -0,0 +1,134 @@
...
+                    reject(error);
+                });
+            });
+        }
+
+        return window.crypto.subtle.verify(this._algorithmOptions, key, signature, data);
+    }
+
+    async generateKey(extractable: boolean, usages: Array<string>): Promise<CryptoKeyPair> {
+        if (isIE11()) {
```
DarylThayil: maybe breakout the if(isIe11()) then do this logic in to functions so that your top level apis read cleaner 

```js
async generateKey(extractable: boolean, usages: Array<string>): Promise<CryptoKeyPair> {
  return isIE11() ? generateIe11Key(extractable. usages) : window.crypto.subtle.generateKey(this._algorithmOptions, extractable, usages);

}
```

DarylThayil: What about sub IE 11?

lib/msal-core/src/BrowserCrypto.ts
```diff
@@ -0,0 +1,134 @@
...
+            return new Promise((resolve, reject) => {
+                const exportKeyOperation = window['msCrypto'].subtle.exportKey(format, key);
+                exportKeyOperation.addEventListener('complete', (e: { target: { result: ArrayBuffer; }; }) => {
+                    const resultBuffer: ArrayBuffer = e.target.result;
+                    const resultString = getStringFromArrayBuffer(resultBuffer)
+                        .replace(/\r/g, '')
+                        .replace(/\n/g, '')
+                        .replace(/\t/g, '')
+                        .split(' ').join('')
+                        .replace("\u0000", '');
```
DarylThayil: can this be a single regex? seems like there is something you need to know about the result to format it, so if it changes, this string manipulation may break?

DarylThayil: maybe factor out the string manipulation into a format method you can test in isolation

lib/msal-core/src/DatabaseCache.ts
```diff
@@ -0,0 +1,65 @@
+interface IDBOpenDBRequestEvent extends Event {
```
DarylThayil: Can you add some context here on the design and the flexibility you are trying to add ?

lib/msal-core/src/PopTokenCommon.ts
```diff
@@ -0,0 +1,61 @@
+export enum HashAlgorithm {
```
DarylThayil: maybe a src/proof-of-possesion folder that has all your pop pieces, and then you don need a common, you could have a types.ts, utils.ts etc...

lib/msal-core/src/PopTokenCommon.ts
```diff
@@ -0,0 +1,61 @@
...
+    sha256 = 'SHA-256'
+};
+
+export enum KeyGenAlgorithm {
+    rsassa_pkcs1_v15 = "RSASSA-PKCS1-v1_5"
```
DarylThayil: your enum casing here is different than your other, any reason?

lib/msal-core/src/PopTokenCommon.ts
```diff
@@ -0,0 +1,61 @@
...
+    jwk = "jwk"
+}
+
+export function getStringFromArrayBuffer(data: ArrayBuffer): string {
+    return String.fromCharCode.apply(null, new Uint8Array(data));
+}
+
+export function getArrayBufferFromString(dataString: string): ArrayBuffer {
+    const data = new ArrayBuffer(dataString.length);
+    const dataView = new Uint8Array(data);
```
DarylThayil:
```js
const dataView = dataString.split('').reduce((memo, val, i) => {
    return [...memo, [i]: dataString.charCodeAt(i)];
}, new Uint8Array(data));
```

DarylThayil: maybe this is more confusing haha

lib/msal-core/src/PopTokenCommon.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+    return data;
+}
+
+export function getJwkString(key: JsonWebKey): string {
+    return JSON.stringify(key, Object.keys(key).sort());
+}
+
+export function utf8Encode(input: string): string {
+    input = input.replace(/\r\n/g, "\n");
```
DarylThayil: dont mutate arg

lib/msal-core/src/PopTokenCommon.ts
```diff
@@ -0,0 +1,61 @@
...
+    var utftext = "";
```
DarylThayil: let vs var

DarylThayil: can probably reduce this down as well instead of mutating

lib/msal-core/src/PopTokenCommon.ts
```diff
@@ -0,0 +1,61 @@
...
+            utftext += String.fromCharCode(((c >> 6) & 63) | 128);
+            utftext += String.fromCharCode((c & 63) | 128);
+        }
+    }
+
+    return utftext;
+}
+
+export function isIE11(): boolean {
+    return "msCrypto" in window;
```
DarylThayil: maybe a little misleading? this is checking for the presence of msCrypto. is it 100% always in ie11 and not in other versions?

DarylThayil: this check is simple enough you could argue that 

if(window.msCrypto) is more explicit than isIE11() (which as written is misleading)

lib/msal-core/src/PopTokenGenerator.ts
```diff
@@ -0,0 +1,153 @@
+import { BrowserCrypto } from './BrowserCrypto';
+import { Base64 } from 'js-base64';
+import { HashAlgorithm, getStringFromArrayBuffer, KeyFormat, utf8Encode, getArrayBufferFromString, KeyUsages, KeyGenAlgorithm, getJwkString } from './PopTokenCommon';
+import { Utils } from './Utils';
+import { DatabaseCache } from './DatabaseCache';
+import { AuthenticationParameters } from './AuthenticationParameters';
```
DarylThayil: same comment here on wanting to encapsulate this all in a folder inside src

lib/msal-core/src/PopTokenGenerator.ts
```diff
@@ -0,0 +1,153 @@
...
+
+type CachedKeyPair = {
+    publicKey: CryptoKey,
+    privateKey: CryptoKey,
+    requestMethod: string,
+    requestUri: string
+}
+
+export class PopTokenGenerator {
```
DarylThayil: This class so I understand, is essentially a browser based glue class for  Crypto, and Database interfaces? 
if so, should this be interacting with interfaces and constructed with implementations?

lib/msal-core/src/PopTokenGenerator.ts
```diff
@@ -0,0 +1,153 @@
...
+    private _crypto: BrowserCrypto;
```
DarylThayil: do we have a naming convention one way or the other with _underscore?

lib/msal-core/src/PopTokenGenerator.ts
```diff
@@ -0,0 +1,153 @@
...
+
+    private static DB_VERSION = 1;
+    private static DB_NAME = 'msal.db';
+    private static TABLE_NAME =`${PopTokenGenerator.DB_NAME}.keys`;
+    private _cache: DatabaseCache<CachedKeyPair>;
```
DarylThayil: maybe naming this cache to be more specific so it is not confused with our other caches

lib/msal-core/src/PopTokenGenerator.ts
```diff
@@ -0,0 +1,153 @@
...
+        const encodedPayload = Utils.base64EncodeStringUrlSafe(utf8Encode(JSON.stringify(payload)));
+
+        const tokenString = `${encodedHeader}.${encodedPayload}`;
+        const tokenBuffer = getArrayBufferFromString(tokenString);
+
+        const signatureBuffer = await this._crypto.sign(privateKey, tokenBuffer);
+        const encodedSignature = Utils.base64EncodeStringUrlSafe(getStringFromArrayBuffer(signatureBuffer));
+        const signedToken = `${tokenString}.${encodedSignature}`;
+
+        return signedToken;
```
DarylThayil: decent amount of awaits here, is there any errors that could come through the control flow to here?

lib/msal-core/src/PopTokenGenerator.ts
```diff
@@ -0,0 +1,153 @@
...
+
+        return popTokenString;
+    }
+
+    /**
+     * Signs an access token with the given pop key.
+     * @param publicKeyHash Public key to sign payload with
+     * @param payload Payload to sign
+     */
+    async signToken(accessToken: string): Promise<string> {
```
DarylThayil: rather large function, can any be broken up into helpers that can be tested?

lib/msal-core/src/ServerRequestParameters.ts
```diff
@@ -45,10 +46,11 @@ export class ServerRequestParameters {
    * @param redirectUri
    * @param state
    */
-  constructor (authority: Authority, clientId: string, scope: Array<string>, responseType: string, redirectUri: string, state: string ) {
+  constructor (authority: Authority, clientId: string, scope: Array<string>, responseType: string, redirectUri: string, state: string, popToken?: string ) {
```
DarylThayil: would be nice to eventually refactor this to take options even on required fields 


new ServerRequestParameters({
   authority,
   clientId,
   .....
   popToken: 'fdav32r4raacd4rt4a......'
})

DarylThayil: not for this pr

DarylThayil: is popToken the thing that is being passed here?

lib/msal-core/src/PopTokenGenerator.ts
DarylThayil: Should this class be a singleton?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -944,6 +961,18 @@ export class UserAgentApplication {
...
+
+        if (cacheResultResponse.accessTokenType === AuthenticationType.PoP) {
+            const signedPopToken = await this.popTokenGenerator.signToken(cacheResultResponse.accessToken);
+
+            resolve({
+                ...cacheResultResponse,
+                accessToken: signedPopToken
+            });
+
+            return null;
```
DarylThayil: either 

return; or return resolve({ ... 
returning null seems to say empty value instead of what I think is meant here which is void return

lib/msal-core/src/PopTokenGenerator.ts
```diff
@@ -0,0 +1,153 @@
...
+            requestMethod,
+            requestUri
+        } = await this._cache.get(publicKeyHash);
+
+        const publicKeyJwk = await this._crypto.exportKey(publicKey, KeyFormat.jwk);
+        const publicJwkString = getJwkString(publicKeyJwk);
+
+        const header = {
+            alg: publicKeyJwk.alg,
+            type: KeyFormat.jwk,
```
jasonnutter: Should be `typ`

lib/msal-core/src/PopTokenGenerator.ts
```diff
@@ -0,0 +1,153 @@
...
+
+        const encodedHeader = Utils.base64EncodeStringUrlSafe(utf8Encode(JSON.stringify(header)));
+
+        const urlComponents = Utils.GetUrlComponents(requestUri);
+        const uriPath = (urlComponents.AbsolutePath || "").replace(/\/+$/, "");
+        const uriHost = urlComponents.HostNameAndPort || "";
+        const uriQueryString = urlComponents.QueryString || "";
+
+        const pS256 = await this._crypto.digest(HashAlgorithm.sha256, uriPath);
+        const qS256 = await this._crypto.digest(HashAlgorithm.sha256, uriQueryString);
```
jasonnutter: Off by default, not needed.

lib/msal-core/src/BrowserCrypto.ts
jasonnutter: Yeah, I could probably create an abstraction for the `msCrypto` operations, might revisit that. 

And if a browser doesn't have either `window.crypto` or `window.msCrypto` (which is in IE11 only, I believe), it will not be able to support PoP.

There is also `webkitSubtleCrypto` [in older Safari versions](https://webkit.org/blog/7790/update-on-web-cryptography/), but I don't think I'm going to worry about supporting that.

lib/msal-core/src/AuthenticationParameters.ts
jasonnutter: Yeah I'll probably fix this.

lib/msal-core/src/AccessTokenValue.ts
jasonnutter: Yes, from what I've seen.

lib/msal-core/src/BrowserCrypto.ts
jasonnutter: It could be a single regex, although I like that this is a little easier to comprehend. Given this is to normalize the response from IE11, I don't expect this to change, but agree about testing.

lib/msal-core/src/DatabaseCache.ts
jasonnutter: Yeah, will do.

lib/msal-core/src/PopTokenCommon.ts
jasonnutter: Agreed.

lib/msal-core/src/PopTokenCommon.ts
jasonnutter: Laziness, mostly. Will clean these enums up on my next pass.

lib/msal-core/src/PopTokenCommon.ts
jasonnutter: Yeah, for loops with mutations isn't my favorite either. Was copy/pasta from the POC, might revisit later.

lib/msal-core/src/PopTokenCommon.ts
jasonnutter: Copy/pasta, but yeah I agree.

lib/msal-core/src/PopTokenCommon.ts
jasonnutter: Yeah, not a big fan of the naming. Will revisit.

lib/msal-core/src/PopTokenGenerator.ts
jasonnutter: Yeah I thought about that (especially thinking ahead to Node), but decided not to overcomplicate things to start. Can always do that incrementally.

jasonnutter: `UserAgentApplication` will new up a singleton instance of this class.

lib/msal-core/src/PopTokenGenerator.ts
jasonnutter: I believe we actually don't want to do underscores, I'll clean this up.

lib/msal-core/src/PopTokenGenerator.ts
jasonnutter: Taking about `_cache` or `DatabaseCache`?

lib/msal-core/src/PopTokenGenerator.ts
jasonnutter: Not a lot of value from that, as this is the only place most of the functionality is used.

lib/msal-core/src/PopTokenGenerator.ts
jasonnutter: Certainly, and `UserAgentApplication` will have to catch any errors and fail gracefully.

lib/msal-core/src/ServerRequestParameters.ts
jasonnutter: Not sure what you mean. `popToken` will be the public key hash that we generate and send to STS.

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: Yeah, think I was just trying to be consistent with teh code around it.

jasonnutter:
This work will be redone at a later time, closing.

PR #814, opened 2019-07-09
Issue of cyclic dependency while using Msal for angular out of the box

tbudukh:
We faced an issue while using Msal service right out of the box. Issue was regarding cyclic dependency. We were trying to bootup a plugin using Msal service with correct config and added app_initializer in plugin module to use msal service. the project using the plugin would not build because of the cyclic dependencies. Router is actually never available before app_initializer. We figured Msal service doesnt really need the router to create its instance and we can make router available when its ready.
We pulled out router dependency from Msal service, put it in its own container and provided config. We also bound the router events after router is init. At the bootup, no, Msal service doesnt need router and can bootup without cyclic dependency errors.


tbudukh:
Reopened the pull request based on jsonnutter's comments on previous PR -https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/761

lib/msal-angular/src/msal.service.ts
```diff
@@ -24,7 +24,7 @@ export class MsalService extends UserAgentApplication {
     private loginScopes: string[];
     _renewActive: boolean;
 
-    constructor(@Inject(MSAL_CONFIG) private config: MsalConfig, private router: Router, private broadcastService: BroadcastService) {
+    constructor(@Inject(MSAL_CONFIG) private config: MsalConfig, private broadcastService: BroadcastService) {
```
jasonnutter: Is this a breaking change (changing the constructor function signature)? Not familiar enough with Angular to know.

tbudukh: It is if users are instantiating the msal service themselves! For the most part, the injector will take care of instantiating the service and users dont have to know about this change.

jasonnutter: Gotya. Given `MsalService` is exported and thus part of the public API, I'd like to avoid a breaking change, even if we don't think anyone is actually using the API in that way. 

If `router` was left in the function signature (to keep the same function signature), presumably if an app was instantiating `MsalService` manually, that `router` would no longer be used, correct?

sameerag: I agree with @jasonnutter to keep the surface intact as the release (once this PR goes through approval) wouldn't be a major version release.

tbudukh: changed the constructor definition back to include router, defaulted it to null..
Can you please review it soon so that we can make our build pick it up from the public registry?

tbudukh: Hi @jasonnutter can you please let me know if there is anything pending on this PR from our side?

jasonnutter: I'll have the team take a look at this PR and get it approved, apologies for the delay!

jasonnutter:
Approved, @sameerag @pkanher617 can one of you please take a look this PR when you get a chance? Thanks!

tbudukh:
Thanks for the update Jason!

tbudukh:
What's happening with this request..
@jasonnutter

jasonnutter:
@tbudukh Apologies for the delay, we've had other Angular issues that were more pressing. Once we get those sorted out, we'll take a look at this again.

tbudukh:
Merged and updated my branch to include latest changes on parent

jasonnutter:
@tbudukh Thanks for your patience. I'm going include this change in the work I'm doing for MSAL Angular 1.0. I'll likely end up merging this into the WIP branch I have open for that, instead of `dev`.

tbudukh:
Thanks!

kylej-89:
Where are we with this? It's blocking me from using msal

tbudukh:
They wouldn't merge this request, its been pending for more than 8 months now! We moved on with our own alpha release internally, so that we're not blocked! 
@kylej-89 - may be you'd wanna do that as well and occasionally merge latest on the master branch to keep it updated

kylej-89:
Ya I might need to, can't wait around for this forever especially considering the work is basically done.

sameerag:
cc @jasonnutter; Wouldn't the angular-beta you are releasing solve this already?

kylej-89:
Seems like Angular 9 actually solves this issue altogether

jasonnutter:
@sameerag No, this would be different. I haven't forgotten about this, will take a look at incorporating it now.

jasonnutter:
@kylej-89 Interesting, good to know. We still want earlier versions to work properly, so we'll still want to address this.

PR #817, opened 2019-07-10
Allowing responseType as an optional configurable parameter

anjiuidev:
## Current behavior
Using msal.js I wanted to pass responseType in the configuration, but msal.js not allowing me to do that.

## Expected behavior
Add responseType as an optional configurable parameter in AuthOptions, so that I can pass responseType in the configuration.

### DIFF
--- a/lib/msal-core/src/Configuration.ts
+++ b/lib/msal-core/src/Configuration.ts
@@ -29,7 +29,7 @@ const NAVIGATE_FRAME_WAIT = 500;
  *  - postLogoutRedirectUri       - Used to redirect the user to this location after logout. Defaults to `window.location.href`.
  *  - state                       - Use to send the state parameter with authentication request
  *  - navigateToLoginRequestUrl   - Used to turn off default navigation to start page after login. Default is true. This is used only for redirect flows.
- *
+ *  - responseType                - used to configure response type. Default is 'id_token'. Available options are 'id_token', 'token', 'id_token token'.
  */
 export type AuthOptions = {
   clientId: string;
@@ -38,6 +38,7 @@ export type AuthOptions = {
   redirectUri?: string | (() => string);
   postLogoutRedirectUri?: string | (() => string);
   navigateToLoginRequestUrl?: boolean;
+  responseType? : string;
 };
 
 /**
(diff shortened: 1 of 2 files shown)


sameerag:
@anjiuidev The responseType is not a user parameter by design in MSAL JS. We determine the token type based on the cache state, call type made, current account state, etc. We also have acquireTokenSilent honor both types depending on the above parameters.

We cannot approve of this change as it stands. If there is a use case where you would need this functionality, please let us know. We might be able to help you achieve that without changing the surface.

anjiuidev:
@sameerag In msal library **loginPopup** function is giving only **idtoken** and to get **access token** I have to call **acquireTokenSilent** function. The second call(acquireTokenSilent) is taking over _30sec time_ to get the response. By default in the _loginPopup_ function, it is taking **responseType** as **'id_token**', If I pass **'id_token token'**, I was able to get both _idtoken_ and _accesstoken_ at a time.

lib/msal-core/src/Configuration.ts
```diff
@@ -38,6 +38,7 @@ export type AuthOptions = {
   redirectUri?: string | (() => string);
   postLogoutRedirectUri?: string | (() => string);
   navigateToLoginRequestUrl?: boolean;
+  responseType? : string;
```
DarylThayil: I dont think this belongs in the top level config, if it were to be added wouldn't it make sense in the AuthenticationParameters request for a single request ?

DarylThayil:
@anjiuidev I am suprised Acquire token silent is taking 30 seconds! how is it not timing out?! 
also I believe login popup is calling acquire token silent under the hood.

lib/msal-core/src/Configuration.ts
sameerag: We need a more streamlined approach to handle idToken only cases but I don't believe we can make it a surface param. I can check with the PMs on what our guidance is here. Also as I stated above,  @anjiuidev use case could be solved without this if we figure out why the delay.

sameerag:
@anjiuidev "In msal library loginPopup function is giving only idtoken and to get access token I have to call acquireTokenSilent function"

Expanding on the above statement, login calls are specifically and intentionally limited to idTokens as it is to facilitate the user to signin. This is needed "only once" and as long as the user has session, you can keep acquiring tokens for multiple resources.  Our [ReadMe](https://github.com/AzureAD/microsoft-authentication-library-for-js) has details on the same.

Regarding ATS taking 30sec every time, I would like to have this as a github issue if you notice significant performance delay. Can you please raise one with sample code for us to reproduce at our end? I am closing this PR for now, and will have someone in my team respond to the github issue if you raise one and tag this PR.

anjiuidev:
@DarylThayil Thanks for your time. I will create a new issue with a sample code.

PR #820, opened 2019-07-10
demonstrate working error catching after redirect by disabling naviga…

jfbloom22:
The current React sample application does not demonstrate how to display errors from failed redirect login attempts.  This is a show stopper for any production application.  The solution here is to disable navigateToLoginRequestUrl.  This prevents refreshing and loosing component state.  

The downside is loosing the functionality of navigateToLoginRequestUrl, however this functionality can easily be implemented outside of MSAL.

### DIFF
--- a/lib/msal-core/samples/react-sample-app/src/AuthProvider.js
+++ b/lib/msal-core/samples/react-sample-app/src/AuthProvider.js
@@ -117,12 +117,11 @@ export default C =>
         async componentDidMount() {
             msalApp.handleRedirectCallback(error => {
                 if (error) {
-                    // This currently doesn't work,
-                    // as the component mounts multiple times
-                    // and state gets lost.
-                    // this.setState({
-                    //     error: "Unable to acquire access token."
-                    // });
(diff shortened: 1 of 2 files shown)


lib/msal-core/samples/react-sample-app/src/AuthProvider.js
```diff
@@ -117,12 +117,11 @@ export default C =>
         async componentDidMount() {
             msalApp.handleRedirectCallback(error => {
                 if (error) {
-                    // This currently doesn't work,
-                    // as the component mounts multiple times
-                    // and state gets lost.
-                    // this.setState({
-                    //     error: "Unable to acquire access token."
-                    // });
+                    const errorMessage = error && error.errorMessage ? error.errorMessage : "Unable to acquire access token.";
```
jasonnutter: This can be simplified, as line 119 already checks for the truthiness of `error`:

```js
const errorMessage = error.errorMessage || "Unable to acquire access token.";
```

jfbloom22: thanks for catching that!

DarylThayil:
LGTM @jasonnutter  to merge

PR #823, opened 2019-07-10
add forceRefresh to api

DarylThayil:
Add a force option to api that will allow cache to be skipped

before merge
[x] Do I need other tests?
[x] Update documentation


lib/msal-core/src/AuthenticationParameters.ts
```diff
@@ -24,6 +24,7 @@ export type AuthenticationParameters = {
     account?: Account;
     sid?: string;
     loginHint?: string;
+    force?: boolean;
```
jasonnutter: I think I would prefer `skipCache` over `force`, is more clear what it will be used for.

DarylThayil: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/976b5e26798d04716e82ea215b12e3374afa7d40/src/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs

How about this for consistency with .NET 

`forceRefresh`

lib/msal-core/test/UserAgentApplication.spec.ts
```diff
@@ -827,8 +827,43 @@ describe("UserAgentApplication.ts Class", function () {
...
+                expect(url).to.include('&client_info=1');
+                done();
+                return {};
+            });
+
+            cacheStorage.setItem(JSON.stringify(accessTokenKey), JSON.stringify(accessTokenValue));
+
+            msal.acquireTokenSilent(tokenRequest).then(function(response) {
+                // Won't happen - we are not testing response here
+                console.error("Shouldn't have response here. Data: " + JSON.stringify(response));
```
technical-boy: Should the test break if you do get a response? Or is this just for reference?

jasonnutter: Yeah, might be worth asserting/throwing here?

jasonnutter: Otherwise just remove the `.then`.

jasonnutter: Also, you can return the promise chain instead of calling `done.

lib/msal-core/test/UserAgentApplication.spec.ts
```diff
@@ -827,8 +827,43 @@ describe("UserAgentApplication.ts Class", function () {
                 console.error("Error in assertion: " + JSON.stringify(err));
             });
         });
+
+        it("tests getCachedToken is skipped when force is set true", function (done) {
```
jasonnutter: Do you need another test for when `forceRefresh: false`?

lib/msal-core/test/UserAgentApplication.spec.ts
DarylThayil: @technical-boy  great question / suggestion
good call updating this stuff now

DarylThayil: actually cant remove done @jasonnutter , it is called in a callback outside of promise chain

lib/msal-core/README.md
```diff
@@ -46,6 +46,9 @@ These security concerns are mitigated per standard practices such as- use of sho
 
 We offer two methods of storage for Msal, `localStorage` and `sessionStorage`.  Our recommendation is to use `sessionStorage` because it is more secure in storing tokens that are acquired by your users, but `localStorage` will give you Single Sign On accross tabs and user sessions.  We encourge you to explore the options and make the best decision for your application.
 
+### forceRefresh to skip cache
+If you would like to skip a cached token and go to the server, please pass in the option `forceRefresh`: boolean into the `AuthenticationParameters` object used to make a login / token request.  **WARNING** this should not be used by default, because of the performance impact on your application.  Relying on the cache will give your users a better experience, and skipping it should onlu be used in scenarios where you know the current cached data does not have up to date information.  Example: Admin tool to add roles to a user, wants to get a new token with said roles.
```
technical-boy:
```suggestion
If you would like to skip a cached token and go to the server, please pass in the option `forceRefresh`: boolean into the `AuthenticationParameters` object used to make a login / token request.  **WARNING** this should not be used by default, because of the performance impact on your application.  Relying on the cache will give your users a better experience, and skipping it should only be used in scenarios where you know the current cached data does not have up-to-date information.  Example: Admin tool to add roles to a user, wants to get a new token with said roles.
```

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -934,7 +934,7 @@ export class UserAgentApplication {
       let authErr: AuthError;
       let cacheResultResponse;
 
-      if (!userContainedClaims) {
+      if (!userContainedClaims && !request.forceRefresh) {
```
sameerag: Any possibility for the request to be null?  login() calls -> acquireTokenSilent() is a path that can have a null request, are we handling this? I hope we did but please cross check.

lib/msal-core/test/UserAgentApplication.spec.ts
sameerag: +1

DarylThayil: I don't think so, since it defaults to false and all other tests don't pass in.  The check is falsey

lib/msal-core/README.md
technical-boy: Just a couple spelling fixes (onlu -> only and up to date -> up-to-date)

lib/msal-core/src/UserAgentApplication.ts
DarylThayil: ATS on it's own checks params in request above this , so if it is broken, I do not believe it is by this

lib/msal-core/README.md
DarylThayil: great catch thanks

lib/msal-core/src/UserAgentApplication.ts
pkanher617: We require a request object for ATS calls, it is not an optional parameter.

lib/msal-core/test/UserAgentApplication.spec.ts
pkanher617: Default case is the normal case for ATS, so we should have some testing for that already.

lib/msal-core/src/UserAgentApplication.ts
pkanher617: You may want to add a log message for the forceRefresh case, as we have for the userContainedClaims on line 959.

pkanher617: Actually this is a bug, since we don't null check the object. May want to add this.

DarylThayil: adding

DarylThayil: It is not clear to me what the right behavior is here.
If a request object is not provided, should we throw from ATS?

sameerag: You just need to check against request being null. The code flow here even if request is null is when ats is called for adal idtoken cases

sameerag: Mostly to avoid referencing undefined objects

DarylThayil: What i mean is if you look at the code above it, there are lots of accessors to the the request object, so if it is null, it should break anyway right ?

DarylThayil: updating to fix my code

lib/msal-core/README.md
jasonnutter: Couple things:

1. "please pass in the boolean `forceRefresh`...
2. **WARNING**: This should not be used by default...
3. should only
4. Example: Admin tool to add roles to a user that needs to get a new token with updates roles.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -934,7 +934,9 @@ export class UserAgentApplication {
       let authErr: AuthError;
       let cacheResultResponse;
 
-      if (!userContainedClaims) {
+      const forceCacheSkip = request ? request.forceRefresh : false;
```
jasonnutter: This could be `!!request.forceRefresh` (as `request` should always be present, correct)?

lib/msal-core/README.md
DarylThayil: updated

lib/msal-core/src/UserAgentApplication.ts
DarylThayil: no the feedback was that it could be null

jasonnutter: I doubt that, as there is code above this that access properties on `request` with null checking, which would break if that was the case.

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: Do we have consumers who call ATS without passing in a request object?

jasonnutter: Because looking at the code, that would not be supported.

jasonnutter: That does not appear to be supported, and in terms of code readability we shouldn't added null checks just because it could be null if its not actually supported. If a null request object isn't supported, we should probably add an assertion at the beginning of the method. If it is supposed to be supported, there are more checks that need to be added (and probably some documentation).

lib/msal-core/src/UserAgentApplication.ts
sameerag: So one case where ATS is called from login calls where request is optional. It is not specific to this case, we have to in general check for that path where request can be null. That was the intent of the original comment. If it is referenced before this statement, we need to check if that is a bug in itself.

lib/msal-core/src/UserAgentApplication.ts
sameerag: Yes. That was the intent of the original comment, it is to ensure we have the code guarded against null request to avoid undefined references. It need not necessarily be a null check and as you said, a generic assertion suffices and is a better approach.

PR #824, opened 2019-07-11
Add abstraction for using custom XhrClient implementation

daviwil:
Hi all!  I work on the [`@azure/identity` library](https://github.com/Azure/azure-sdk-for-js/tree/master/sdk/identity/identity) that's part of the new wave of Azure SDK libraries in development by my team.  We've just started looking into browser-based authentication and wanted to integrate MSAL.js for that purpose since you already have a complete implementation.

@schaabs and I were wondering if there is a way to use our own HTTP pipeline for any requests that MSAL.js makes, but based on what I found in the code, `XhrClient` use is hardwired.  This PR is an attempt to make the `XhrClient` implementation configurable so that our own can be swapped in.

I'm not thrilled with calling the interface `IXhrClient` but thought it'd be the best starting point for discussion.  Perhaps it should be `RestClient` or something to that effect?

### DIFF
--- a/lib/msal-core/src/AadAuthority.ts
+++ b/lib/msal-core/src/AadAuthority.ts
@@ -2,7 +2,7 @@
 // Licensed under the MIT License.
 
 import { Authority, AuthorityType } from "./Authority";
-import { XhrClient } from "./XHRClient";
+import { IXhrClient } from "./XHRClient";
 import { AADTrustedHostList } from "./Constants";
 
 /**
@@ -15,8 +15,8 @@ export class AadAuthority extends Authority {
       return `${AadAuthority.AadInstanceDiscoveryEndpoint}?api-version=1.0&authorization_endpoint=${this.CanonicalAuthority}oauth2/v2.0/authorize`;
   }
 
-  public constructor(authority: string, validateAuthority: boolean) {
-    super(authority, validateAuthority);
+  public constructor(authority: string, validateAuthority: boolean, xhrClient?: IXhrClient) {
+    super(authority, validateAuthority, xhrClient);
   }
 
   public get AuthorityType(): AuthorityType {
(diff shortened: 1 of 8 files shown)


jasonnutter:
Yeah `XhrClient` may not be the best name, but its what we got, so `IXhrClient` is reasonable given that.

Out of curiosity, how would your custom `IXhrClient` implementation differ?

daviwil:
In our implementation, we would leverage the HTTP pipeline in [`@azure/core-http`](https://github.com/Azure/azure-sdk-for-js/tree/master/sdk/core/core-http) which allows us to inject various policies that are relevant to our SDK libraries (telemetry, logging, retries, proxies, etc).  We want to make sure that all HTTP traffic coming out of the use of our libraries is going through such a pipeline so that we have the ability to tune and monitor it.

More specifically, our `IXhrClient` implementation would wrap a [`ServiceClient`](https://github.com/Azure/azure-sdk-for-js/blob/6d4c34ae5f7a642e05241f66d0bcb323de90f5c7/sdk/core/core-http/lib/serviceClient.ts#L110) , configured with the appropriate policies, and then implement the `IXhrClient.sendRequestAsync` method by calling `ServiceClient.sendRequest` and returning the deserialized JSON response as an object.

jasonnutter:
Gotya. While the code here is reasonable in-and-of itself, there's a more philosophical discussion to be had about whether or not we want to delegate something as critical as http calls over to the consuming application. This is both in terms of maintenance on our side, and that this is arguably a footgun as it could result in unexpected or broken behavior if not implemented correctly.

Note, that we are in the process of adding telemetry hooks to the library, which is intended to help 1st party apps such as yours.

cc: @DarylThayil

DarylThayil:
I would agree with Jason, that there are some concerns about whether or not this is something we want to make configurable.  

We have some things coming in the next minor release that will return telemetry back to the dev, things such as http requests status, performance, etc.

My initial preference is see how far we can get with that, and then look at what we can add to our library to get you the insights you need.

daviwil:
Thanks all!  Is there any work in progress for those hook points that I could take a look at?  I'll chat with Scott and see if it'd be enough to just hook in to those or if there are other needs we might have.

DarylThayil:
https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev/lib/msal-core/src/telemetry these are currently the type of data we will be sending back, in the middle of instrumenting the actual code.

Going to reach out to Scott to set up a meeting to better understand needs and how we can help!

jasonnutter:
We have made this possible in `@azure/msal-browser` via `system.networkClient`. Closing.

Issue #746, opened 2019-06-06
The base64.js dependency is not an es6 module and does not work in the browser


nmetulev:
## I'm submitting a...

<pre><code>
[x] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Other... Please describe:
</code></pre>

## Browser:
- [ ] Chrome version XX
- [ ] Firefox version XX
- [ ] IE version XX
- [x] Edge dev latest
- [ ] Safari version XX

## Library version

Library version: 1.0.1

## Current behavior

Using the es modules directly in a browser throws an error in the browser console trying to load base64.js. This issue was introduced in #712. [base64.js does not provide](https://github.com/dankogai/js-base64/blob/master/base64.js#L10) an es module export and a browser can not load the file. See https://github.com/microsoftgraph/microsoft-graph-toolkit/issues/103 for more details. 

## Expected behavior

All es modules in the project should work directly in the browser.

## Minimal reproduction of the problem with instructions

Follow the steps in https://github.com/microsoftgraph/microsoft-graph-toolkit/issues/103 to repro

sameerag:
@nmetulev Thanks for raising this. Will adding {"babel-preset-env": "^1.7.0"} in devdependecies solve this issue? We kind of debated within the team keeping it vs removing it and ultimately decided to not go ahead to keep the dependencies lean.

Let me know and maybe I can push for that.

sameerag:
@DarylThayil FYI - the ES6 dependency I was talking about when we pulled in this change. Looks like it is affecting some customers. Can you take a look too and see if there is another approach apart from what I suggested above?

nmetulev:
Thanks for the response @sameerag. I don't believe "babel-preset-env" will solve this  since you are generating the es modules directly through `tsc` and it's not passing through babel

jasonnutter:
Looks like we need a base64 implementation that is compatible with es modules, correct @nmetulev? Taking a look to see if I can find any that would match that.

Also, I think it would be good to add a sample/test case to our repo to catch Graph Toolkit's use case.

sameerag:
@jasonnutter So you also need to take into account that we need approvals from our sec team to approve of any external encoding/decoding library we use. base64.js is approved, we may need to go through all that process if we are deciding to pick a new library.

jasonnutter:
[...]

PR #826, opened 2019-07-12
Remove base64-js dependency due to incompatibility with es modules, remove unused encoding functions

jasonnutter:
Removes the dependency on `js-base64` due to incompatibility with ES modules, as reported in #746. Replaced base64 encoding functionality with implementation provided on [MDN](https://developer.mozilla.org/en-US/docs/Web/API/WindowBase64/Base64_encoding_and_decoding#Solution_4_%E2%80%93_escaping_the_string_before_encoding_it).

Also removes unused encoding functions that performed similar functionality.

### DIFF
--- a/lib/msal-core/src/Utils.ts
+++ b/lib/msal-core/src/Utils.ts
@@ -7,9 +7,7 @@ import {Constants, SSOTypes, PromptState} from "./Constants";
 import { AuthenticationParameters } from "./AuthenticationParameters";
 import { AuthResponse } from "./AuthResponse";
 import { IdToken } from "./IdToken";
-import { ClientAuthError } from "./error/ClientAuthError";
 import { Library } from "./Constants";
-import { Base64 } from "js-base64";
 import { StringDict } from "./MsalTypes";
 
 /**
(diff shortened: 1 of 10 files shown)


jasonnutter:
Before merging I want to test this with my PoP branch, which relies heavily on base64 encoding.

lib/msal-core/src/Utils.ts
```diff
@@ -225,136 +223,29 @@ export class Utils {
...
    *
    * @param input
    */
-  static base64EncodeStringUrlSafe(input: string): string {
-    // html5 should support atob function for decoding
-    return Base64.encode(input);
+  static base64Encode(input: string): string {
+    return btoa(encodeURIComponent(input).replace(/%([0-9A-F]{2})/g,
+        function toSolidBytes(match, p1) {
+            return String.fromCharCode(Number('0x' + p1));
```
pkanher617: This build is failing for me locally because of the single quoted strings here and below.

ERROR: src/Utils.ts:236:47 - ' should be "
ERROR: src/Utils.ts:246:49 - ' should be "
ERROR: src/Utils.ts:247:16 - ' should be "
ERROR: src/Utils.ts:247:23 - ' should be "
ERROR: src/Utils.ts:248:13 - ' should be "

I'm not sure how travis is passing, but might want to double check your local build.

pkanher617:
Tested in IE, Edge, Chrome and Firefox. Looks like it works fine, but I would ask that you run some end-to-end tests manually to make sure. If you need a set of use cases let me know.

DarylThayil:
LGTM @jasonnutter

PR #828, opened 2019-07-15
Regenerate package-lock to pick up latest version of mixin-deep to resolve npm audit warnings

jasonnutter:
This will resolve the 179 `npm audit` security warnings due to `mixin-deep` and a couple other transitive dependencies. 

There is still one moderate warning left for `marked` via `typedoc`, [which should be addressed soon](https://github.com/TypeStrong/typedoc/issues/1009).

### DIFF
--- a/lib/msal-core/package.json
+++ b/lib/msal-core/package.json
@@ -43,11 +43,11 @@
   "devDependencies": {
     "@types/chai": "^4.1.7",
     "@types/handlebars": "4.0.33",
+    "@types/js-base64": "^2.3.1",
     "@types/mocha": "^5.2.6",
     "@types/node": "^11.13.4",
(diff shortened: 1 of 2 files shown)


lib/msal-core/package.json
```diff
@@ -43,11 +43,11 @@
   "devDependencies": {
     "@types/chai": "^4.1.7",
     "@types/handlebars": "4.0.33",
+    "@types/js-base64": "^2.3.1",
```
sameerag: Any specific reason for the change of order?

jasonnutter: `npm` puts dependencies in alphabetical order (they were not before).

DarylThayil:
lgtm, good to merge

PR #831, opened 2019-07-15
Error check request

DarylThayil:
add checks for empty request and throw new error

### DIFF
--- a/lib/msal-core/src/error/ClientConfigurationError.ts
+++ b/lib/msal-core/src/error/ClientConfigurationError.ts
@@ -70,6 +70,10 @@ export const ClientConfigurationErrorMessage = {
     claimsRequestParsingError: {
         code: "claims_request_parsing_error",
         desc: "Could not parse the given claims request object."
+    },
+    emptyRequestError: {
+        code: "empty_request_error",
+        desc: "Request object is required."
     }
 };
 
@@ -136,5 +140,10 @@ export class ClientConfigurationError extends ClientAuthError {
     static createClaimsRequestParsingError(claimsRequestParseError: string): ClientConfigurationError {
         return new ClientConfigurationError(ClientConfigurationErrorMessage.claimsRequestParsingError.code,
             `${ClientConfigurationErrorMessage.claimsRequestParsingError.desc} Given value: ${claimsRequestParseError}`);
+    }
+
+    static createEmptyRequestError(): ClientConfigurationError {
+        const { code, desc } = ClientConfigurationErrorMessage.emptyRequestError;
+        return new ClientConfigurationError(code, desc);
     }
 }
--- a/lib/msal-core/test/UserAgentApplication.spec.ts
+++ b/lib/msal-core/test/UserAgentApplication.spec.ts
@@ -486,6 +486,17 @@ describe("UserAgentApplication.ts Class", function () {
             expect(authErr.stack).to.include("UserAgentApplication.spec.js");
             done();
         });
+
+        it("throws an error if configured with a null request", () => {
+            let correctError;
+            try {
+                msal.acquireTokenRedirect();
+            } catch (e) {
+                expect(e).to.be.instanceOf(ClientConfigurationError);
+                correctError = true;
+            }
+            expect(correctError).to.be.true;
+        });
     });
 
     describe("Different Callback Signatures", function () {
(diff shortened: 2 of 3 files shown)


lib/msal-core/test/UserAgentApplication.spec.ts
```diff
@@ -486,6 +486,14 @@ describe("UserAgentApplication.ts Class", function () {
             expect(authErr.stack).to.include("UserAgentApplication.spec.js");
             done();
         });
+
+        it("throws an error if configured with a null request", () => {
+            try {
+                msal.acquireTokenRedirect();
```
jasonnutter: I'm not sure this is what you want to do, as this test will pass if it doesn't throw.

jasonnutter: Checkout https://github.com/domenic/chai-as-promised provides a nice set of chai asserts for tests such as this.

DarylThayil: good call ... updating!

jasonnutter: e.g.
```js
return msal.acquireTokenRedirect.should.be.rejectedWith(ClientConfigurationError);
```

technical-boy: Could you do this to assert the error details and then add a separate test that does something like:

```typescript
it("throws an error if configured...", () => {
    ...
   expect(msal.acquireTokenRedirect).to.throw(ClientConfigurationError);
});
```
?

DarylThayil: There is some weird error happening with the iframe wrapping code that would prevent this 
pushing test that works

DarylThayil: what do you mean hector?

DarylThayil: not sure I fully understood

lib/msal-core/src/error/ClientConfigurationError.ts
```diff
@@ -70,6 +70,10 @@ export const ClientConfigurationErrorMessage = {
     claimsRequestParsingError: {
         code: "claims_request_parsing_error",
         desc: "Could not parse the given claims request object."
+    },
+    emptyRequestError: {
+        code: "empty_request_error",
+        desc: "Request object is required"
```
jasonnutter: Nit pick: Period at the end (for consistency).

DarylThayil: updating

lib/msal-core/test/UserAgentApplication.spec.ts
```diff
@@ -1281,6 +1292,17 @@ describe("UserAgentApplication.ts Class", function () {
             expect(acquireTokenPromise instanceof Promise).to.be.true;
             acquireTokenPromise.catch(error => {});
         });
+
+        it("throws an error if configured with a null request", () => {
+            let correctError;
+            try {
+                msal.acquireTokenPopup();
```
jasonnutter: I think you need to await this?

lib/msal-core/test/UserAgentApplication.spec.ts
technical-boy: Basically, one test that checks that the error being thrown *is the right one* (accomplished by your original code) and then another test that checks that *the error is actually being thrown* (the example code I suggested).

lib/msal-core/test/UserAgentApplication.spec.ts
jasonnutter: Or you should, to be safe.

DarylThayil: hmm the error will happen before the promise right?

lib/msal-core/test/UserAgentApplication.spec.ts
technical-boy: Although the `correctError = true` hack makes it so you test both at the same time, which is fine (maybe even better).

lib/msal-core/test/UserAgentApplication.spec.ts
DarylThayil: so there is no return type

DarylThayil: in the scenario it doesnt throw before the promise train goes off on its own, I think that it is good that it continues on to the non catch expect that will fail

jasonnutter: That's not the point. Since its a promise, it will happen async, and in theory the error could happen and not get caught by the catch (I'm surprised what you have works). Read in the yellow in [this article](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function#Rewriting_a_promise_chain_with_an_async_function) on the differences between `return await promise()` and `return promise()` when used in a try/catch.

jasonnutter: I disagree, as that would be potentially a false negative. Adding await ensures your catch will be executed if the promise throws.

e.g.
```js
try {
  // we dont care about the return value of the resolve promise
  await msal.acquireTokenPopup();
} catch (e) {
 // this will always get run if the promise throws
}
```

jasonnutter: Ah so I misread that you put the throws before/outside of the `new Promise`. I would personally like to see the assertion inside the promise.

PR #836, opened 2019-07-17
Removes claim dependency 

sameerag:
Since the claim name is not required to exist in the id token (https://openid.net/specs/openid-connect-core-1_0.html#IDToken), having the check depending on that claim is breaking some angular users.
In case of this claim name been not returned, the MsalGuard process to acquire a token silently just fails and the redirect to the identity provider takes place.

Fixing the code to check if the idToken is populated than a specific claim.

### DIFF
--- a/lib/msal-angular/src/msal-guard.service.ts
+++ b/lib/msal-angular/src/msal-guard.service.ts
@@ -24,7 +24,7 @@ export class MsalGuard implements CanActivate {
         this.authService.getLogger().verbose("location change event from old url to new url");
 
         this.authService.updateDataFromCache([this.config.clientID]);
-        if (!this.authService._oauthData.isAuthenticated && !this.authService._oauthData.userName) {
+        if (!this.authService._oauthData.isAuthenticated && !this.isObjectEmpty(this.authService._oauthData.idToken)) {
             if (state.url) {
 
                 if (!this.authService._renewActive && !this.authService.loginInProgress()) {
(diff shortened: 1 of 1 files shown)


lib/msal-angular/src/msal-guard.service.ts
```diff
@@ -90,4 +90,8 @@ export class MsalGuard implements CanActivate {
         return (typeof str === "undefined" || !str || 0 === str.length);
     };
 
+    isObjectEmpty(obj: Object) {
```
DarylThayil: should this be in a util?

sameerag: Ideally yes but there is no util currently for angular and we can move a bunch of functionality when we get there. Did not want to introduce it as a part of this PR.

lib/msal-angular/src/msal-guard.service.ts
```diff
@@ -90,4 +90,8 @@ export class MsalGuard implements CanActivate {
         return (typeof str === "undefined" || !str || 0 === str.length);
```
DarylThayil: This should be a util

AlteredConstants:
Hey, I figured I'd mention what I think might be a bug in this change. I'm currently vendoring this project (to fix other bugs) and could not get this latest code to work under a normal login flow that worked before the change.

Regarding this part:
```ts
!this.isObjectEmpty(this.authService._oauthData.idToken)
```

I believe this should not be negated. Looks like an unintentional double negative? Removing the `!` seemed to work for me, at least, though I'm not certain how to reproduce the original error that prompted this change.

sameerag:
@AlteredConstants Yes,  we are checking for a case where idToken does not exist instead of depending on a specific claim. I will also create a new PR that needs to address one more use case in this context, Will update the PR here once ready.

PR #840, opened 2019-07-18
add telemetry config and construct telem manager

DarylThayil:
add telemetry config and construct telem manager
This will be the instance used to instrument code

### DIFF
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -14,7 +14,7 @@ import { Storage } from "./Storage";
 import { Account } from "./Account";
 import { Utils } from "./Utils";
 import { AuthorityFactory } from "./AuthorityFactory";
-import { Configuration, buildConfiguration } from "./Configuration";
+import { Configuration, buildConfiguration, TelemetryOptions } from "./Configuration";
 import { AuthenticationParameters, validateClaimsRequest } from "./AuthenticationParameters";
 import { StringDict } from "./MsalTypes";
 import { ClientConfigurationError } from "./error/ClientConfigurationError";
@@ -23,8 +23,9 @@ import { ClientAuthError, ClientAuthErrorMessage } from "./error/ClientAuthError
 import { ServerError } from "./error/ServerError";
 import { InteractionRequiredAuthError } from "./error/InteractionRequiredAuthError";
 import { AuthResponse, buildResponseStateOnly } from "./AuthResponse";
-
-// default authority
+import TelemetryManager from "./telemetry/TelemetryManager";
+import { TelemetryPlatform, TelemetryConfig } from './telemetry/TelemetryTypes';
+ // default authority
 const DEFAULT_AUTHORITY = "https://login.microsoftonline.com/common";
 
 /**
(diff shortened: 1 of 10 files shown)


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -2617,5 +2624,34 @@ export class UserAgentApplication {
...
+    }
+    if (!applicationVersion) {
+      missingVals.push("applicationVersion (string)");
+    }
+    if (telemetryEmitter) {
+      missingVals.push("telemetryEmitter (function)");
+    }
+    if (missingVals.length > 0) {
+      throw ClientConfigurationError.createTelemetryConfigError(missingVals);
+    }
```
jasonnutter: Another approach for this validation code would be to simplified the code here, and then pass `config` to `createTelemetryConfigError`. One thing I don't like about this code if you have error message business logic (i.e. the strings themselves) mixed into this function, when they could be abstracted into the function that creates the error.

```js
if (!applicationName || !applicationVersion || ! telemetryEmitter) {
 throw ClientConfigurationError.createTelemetryConfigError(config);
}
```

```js
static createTelemetryConfigError(config: TelemetryOptions): ClientConfigurationError {
        const { code, desc } = ClientConfigurationErrorMessage.telemetryConfigError;
        const requiredKeys = {
            applicationName: "string",
            applicationVersion: "string",
            telemetryEmitter: "function"
        }

       const missingKeys = Object.keys(requiredKeys).reduce((keys, key) => config[key] ? keys : keys.concat([ `${key} (${requiredKeys[key]})` ], []);

        const errorMessage = 
        return new ClientConfigurationError(code, `${desc} mising values: ${missingKeys.join(",")}`);
    }
```

DarylThayil: sold

lib/msal-core/test/Configuration.spec.ts
```diff
@@ -118,4 +118,31 @@ describe("Configuration.ts Type", () => {
...
+            }
+        };
+        const configWithDefaults: Configuration = buildConfiguration(config);
+        const { system } = configWithDefaults;
+        expect(system.telemetry).to.be.undefined;
+    });
+    it("does not set defaults for telemetry object memebers", () => {
+        const config: Configuration = {
+            auth: {
+                clientId: "iamnotreal"
```
pkanher617: Can we use string constant variables instead of hardcoding strings in tests?

lib/msal-core/test/UserAgentApplication.spec.ts
```diff
@@ -134,6 +134,62 @@ describe("UserAgentApplication.ts Class", function () {
...
+            });
+        });
+        it("configure telemtry in UAA missing configuration throws config error", () => {
+            const configureTestCase = () => msal = new UserAgentApplication({
+                auth: {
+                    clientId: TEST_CONFIG.MSAL_CLIENT_ID
+                },
+                system: {
+                    telemetry: {
+                        applicationName: "msal.jsTest",
```
pkanher617: Same comment about string constant variables.

DarylThayil: updating

lib/msal-core/test/Configuration.spec.ts
DarylThayil: updating

PR #850, opened 2019-07-19
errors for spacing and quotes

DarylThayil:
turn eslint rules on for double quote and spacing.
No fixes needed

### DIFF
--- a/lib/msal-core/.eslintrc.json
+++ b/lib/msal-core/.eslintrc.json
@@ -18,7 +18,7 @@
         "@typescript-eslint/ban-types": 0,
         "@typescript-eslint/camelcase": 0,
         "@typescript-eslint/explicit-member-accessibility": 0,
-        "@typescript-eslint/indent": [0, 4, { "SwitchCase": 1 }],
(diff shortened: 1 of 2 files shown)


lib/msal-core/.eslintrc.json
```diff
@@ -18,7 +18,7 @@
         "@typescript-eslint/ban-types": 0,
         "@typescript-eslint/camelcase": 0,
         "@typescript-eslint/explicit-member-accessibility": 0,
-        "@typescript-eslint/indent": [0, 4, { "SwitchCase": 1 }],
+        "@typescript-eslint/indent": [2, 4, { "SwitchCase": 1 }],
```
jasonnutter: Does anything change when you enable the normal `indent` rule? I think that's what I did originally that appear to caused a bunch of changes.

DarylThayil: I think we saw some conflicts there right? so we are defaulting to @ts version?
Theoretically ts lint should have been enforcing that?

PR #851, opened 2019-07-20
adding CacheLocationChoices so clients don't need to worry about type widening

jodonnell:


jodonnell:
running npm run endToEnd was broken, npm run test failed one time and passed the next, so seems like there might be a flaky test?

i didn't want to change CacheLocation since probably a lot of users are relying on that so I just added a new enum that can be used instead.

sameerag:
Merging as we got the team's approval.

pkanher617:
This ended up breaking surface configuration for cacheLocation, we will revert this PR and revisit the change in a future release.

pkanher617:
@jodonnell This was failing  in typescript because it did not allow users to set string values for localstorage and sessionstorage and the enums weren't exported. Could you please test this in typescript and javascript to ensure the behavior is working correctly? You can open another PR, unfortunately I can't reopen this one.

jodonnell:
I will try to get this done in the near future.

PR #861, opened 2019-07-23
proof of concept login provider

DarylThayil:
Proof of concept of how we can split the business logic out of UAA

Terms:

*Resource* - The piece of the application responsible as the interface to consumers.  UserAgentApplication would be the resource / client.  Its sole responsibility would be to expose methods like `acquireTokenSilent` or `loginPopup`, not to implement them.

*Provider* - Instantiated by the resource, but not responsible for exposing anything externally to customers.  It encapsulates the business logic / recipe for what `loginPopup` means, or the steps needed to execute that.

*Store* [Not Seen Here]  - Responsible for the applicaiton state of the entire app.  Instead of UAA holding references like this.loginInProgress, it would be encasulated and managed by a <something>Store class.  This could eventually be broken up into many stores, but it keeps state out of resources and providers but allows them to be accessed. 

What would moving to this pattern look like

1. All public apis in UAA are simply wrappers, and the business logic is pulled into providers.
we would have.
```js
/providers
   LoginProvider.ts [loginRedirect, loginPopup]  // these can all be separate fils
   TokenProvider.ts [acquireTokenSilent, popup, Redirect] 
```

For now, UAA would still have state, but we could pass an instance of UAA into each provider so they have what they need to keep working

- When step 1 is done, the code is separated but the problems that still exist are 
 * state in UAA with an instance being passed around
* Utils that dont belong in UAA are still there 
* Things are now public that don't belong in the public scope.

2. Factor out all the state independent utilities from UAA 
    now that the business logic functions are gone, we can pull all the state independent things into utilities.  Basically every function we had to make public that doesn't use `this`

Problems that still exist after this 
*   UAA still has state and every provider needs to know about its calling resource

3.  Encapsulate app state in a store.
pull out app state into a singleton class responsible for managing state.  UAA no longer has any state, but is responsible for initializing the store and provders

```js
class UAA {
   constructor(config){
        this.applicationStore = new ApplicationStore(config);
       
        this.tokenProvder = new TokenProvider(this.applicationStore, this.logger);
        this.loginProvider = new LoginProvider(this.applicationStore, this.logger, this.tokenProvider);
    }


    loginRedirect() {
       return this.loginProvdider.loginRedirect
   }
   
   loginPopup() { 
       return this.loginProvdider.loginPopup();
   }

   acquireTokenSilen() {
       return this.tokenProvider.acquireTokenSilent();
   }
}
```

UAA is now free from state, utils, and business logic.
Providers encapsulate business logic and can be cleanly understood and changed

### DIFF
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -25,7 +25,9 @@ import { InteractionRequiredAuthError } from "./error/InteractionRequiredAuthErr
 import { AuthResponse, buildResponseStateOnly } from "./AuthResponse";
 import TelemetryManager from "./telemetry/TelemetryManager";
 import { TelemetryPlatform, TelemetryConfig } from './telemetry/TelemetryTypes';
- // default authority
+import LoginProvider from "./login/LoginProvider";
+
+// default authority
 const DEFAULT_AUTHORITY = "https://login.microsoftonline.com/common";
 
 /**
@@ -53,7 +55,7 @@ declare global {
  * Since we support only implicit flow in this library, we restrict the response_type support to only 'token' and 'id_token'
  *
  */
-const ResponseTypes = {
+export const ResponseTypes = {
   id_token: "id_token",
   token: "token",
   id_token_token: "id_token token"
@@ -141,13 +143,13 @@ export class UserAgentApplication {
   private errorReceivedCallback: errorReceivedCallback = null;
 
   // Added for readability as these params are very frequently used
-  private logger: Logger;
-  private clientId: string;
+  public logger: Logger;
+  public clientId: string;
   private inCookie: boolean;
   private telemetryManager: TelemetryManager;
 
   // Cache and Account info referred across token grant flow
-  protected cacheStorage: Storage;
+  public cacheStorage: Storage;
   private account: Account;
 
   // state variables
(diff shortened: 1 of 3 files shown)


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -2644,4 +2545,19 @@ export class UserAgentApplication {
     };
     return new TelemetryManager(telemetryManagerConfig, telemetryEmitter);
   }
-}
+
+  public redirectCallbacksAreSet(): boolean  {
```
DarylThayil: these would all be in the eventual store

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -2539,7 +2440,7 @@ export class UserAgentApplication {
    * @param request
    * @param serverAuthenticationRequest
    */
-  private populateQueryParams(account: Account, request: AuthenticationParameters, serverAuthenticationRequest: ServerRequestParameters, adalIdTokenObject?: any): ServerRequestParameters {
+  public populateQueryParams(account: Account, request: AuthenticationParameters, serverAuthenticationRequest: ServerRequestParameters, adalIdTokenObject?: any): ServerRequestParameters {
```
DarylThayil: these are all utilties

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -2470,7 +2371,7 @@ export class UserAgentApplication {
    * @hidden
    * @ignore
    */
-  private updateCacheEntries(serverAuthenticationRequest: ServerRequestParameters, account: Account, loginStartPage?: any) {
+  public updateCacheEntries(serverAuthenticationRequest: ServerRequestParameters, account: Account, loginStartPage?: any) {
```
DarylThayil: cache stuff should be behind a singleton encapsulation

PR #868, opened 2019-07-25
Change msal-angular to output es5, fixing compatibility with IE11

jasonnutter:
Fixes MSAL Angular compatibility with IE11 by changing the output to `es5`.

Key changes:
- Changed the compilation configuration to ouput es5/commonjs modules instead of es6/es2015 modules, which was not compatible with IE11. The diff is a little confusing, as I renamed [the old tsconfig-aot.json](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/5bc9828ae9d32dbc2965f24b85ab1cc74fbbe0b5/lib/msal-angular/tsconfig-aot.json) to `tsconfig.json`, and also changed the options accordingly.
- Removed the webpack scripts and config, as they were not being used (and were a source of confusion), and updated the build scripts to use the `ngc` command.
- Removed the `dist` folder from the repo.
- Update msal-angular to use `msal` core `0.2.2` which introduced the `storeAuthStateInCookie` config flag for IE11.

Once approved, I plan to release as a beta to ensure we haven't introduced any regression like we did last time.


lib/msal-angular/dist/msal-angular.js
```diff
@@ -30433,6 +30433,10 @@ var MsalGuard = /** @class */ (function () {
             return currentAbsoluteUrl.substring(0, index);
         }
     };
+    MsalGuard.prototype.isObjectEmpty = function (obj) {
```
DarylThayil: is this duplicated somewhere else? I remember commenting on it in @sameerag  pr somewhere

DarylThayil: nm this is dist

lib/msal-angular/src/msal-guard.service.ts
```diff
@@ -34,13 +34,12 @@ export class MsalGuard implements CanActivate {
                         this.authService.getCacheStorage().setItem(Constants.angularLoginRequest, loginStartPage);
                     }
                     if (this.config.popUp) {
-                        return new Promise((resolve, reject) => {
-                            this.authService.loginPopup(this.config.consentScopes, this.config.extraQueryParameters).then(function (token) {
-                                resolve(true);
+                        return this.authService.loginPopup(this.config.consentScopes, this.config.extraQueryParameters)
```
DarylThayil: does this solve ie 11?

DarylThayil:
will this effect any who are using es6 modules / any other side effects?

lib/msal-angular/src/msal-guard.service.ts
jasonnutter: This change is really fix to type errors in typescript...I could fix it the same way `Map` was fixed (by importing a promise polyfill from `core-js-pure` but decided not to since newing up a `Promise` was actually needed from a functional perspective.

jasonnutter:
I suppose it might impact es modules in theory, but I'm not sure if anyone uses es modules with Angular in a way that would break (or if that's even possible?).

DarylThayil:
looks like tests are failing

lib/msal-angular/package.json
```diff
@@ -92,9 +93,11 @@
     "clean": "shx rm -rf dist docs",
     "doc": "typedoc --out ./docs ./src/ --gitRevision dev",
     "build:modules": "tsc && tsc -m es6 --outDir lib-es6",
+    "build:ngc": "ngc -p tsconfig-aot.json",
     "build": "npm run clean && npm run doc && npm run build:modules && webpack",
-    "test": "typings install && karma start karma.conf.js",
-    "ngcompile": "npm run clean && npm run doc && ngc -p tsconfig-aot.json && webpack",
+    "test": "typings install && npm run test:verify-build && karma start karma.conf.js",
```
sameerag: not related but at somepoint, we may need to move to chai-mocha here too?

lib/msal-angular/src/msal-guard.service.ts
```diff
@@ -50,21 +49,22 @@ export class MsalGuard implements CanActivate {
         }
         //token is expired/deleted but userdata still exists in _oauthData object
         else if (!this.authService._oauthData.isAuthenticated && this.authService._oauthData.userName) {
-            return new Promise((resolve, reject) => {
-                this.authService.acquireTokenSilent([this.config.clientID]).then((token: any) => {
+            return this.authService.acquireTokenSilent([this.config.clientID])
```
sameerag: +1. Also looks more optimal.

lib/msal-angular/src/msal-guard.service.ts
```diff
@@ -50,21 +49,22 @@ export class MsalGuard implements CanActivate {
...
+                .then((token: any) => {
                     if (token) {
                         this.authService._oauthData.isAuthenticated = true;
                         var authenticationResult = new AuthenticationResult(token );
                         this.broadcastService.broadcast("msal:loginSuccess",  authenticationResult);
-                        resolve (true);
+                        return true
```
sameerag: semicolon missing. I know JS doesn't care but good for syntactic uniformity across the code

lib/msal-angular/tsconfig-aot.json
```diff
@@ -1,6 +1,6 @@
 {
     "compilerOptions": {
-      "target": "es6",
+      "target": "es5",
```
sameerag: So is this what fixes the IE11 and the above ngc tool for AOT compatibility?

jasonnutter: Yeah, this was the first step. Then I had to make the other changes (removing `Promise` usage, polyfill `Map`, etc) to fix type errors.

jasonnutter:
@sameerag Yeah we can probably remove the dist files from the repo.

lib/msal-angular/package.json
```diff
@@ -24,7 +24,7 @@
     "node": ">=0.8.0"
   },
   "dependencies": {
-    "msal": "0.2.1",
+    "msal": "0.2.2",
```
sameerag: why not 2.3?

jasonnutter: I tried it but it didn't work. `0.2.2` has `storeAuthStateinCookie`, which was all I needed, so didn't try to dig deeper into why `0.2.3` didn't work.

PR #873, opened 2019-07-26
Fixing issue with tokenType when renewing Id Token

pkanher617:
This PR is to fix part of an issue describe in issue #736 where the tokenType in the response for a acquireTokenSilent call which was renewing an id token was incorrectly displaying the "access_token" type.

### DIFF
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -655,6 +655,7 @@ export class UserAgentApplication {
               // App uses idToken to send to api endpoints
               // Default scope is tracked as clientId to store this token
               this.logger.verbose("renewing idToken");
+              this.silentLogin = true;
               this.renewIdToken(request.scopes, resolve, reject, account, serverAuthenticationRequest);
             } else {
               // renew access token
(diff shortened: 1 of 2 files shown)


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -655,6 +655,7 @@ export class UserAgentApplication {
               // App uses idToken to send to api endpoints
               // Default scope is tracked as clientId to store this token
               this.logger.verbose("renewing idToken");
+              this.silentLogin = true;
```
jasonnutter: Can you explain what this option does?

pkanher617: Essentially it sets the fact that we are logging in (requesting only an id token) silently. It isn't a great design, and we are looking at refactoring this to take this out, but for now it exists so I am leveraging it here.

pkanher617:
This is not something we can unit test, which is why I have added a sample.

lib/msal-core/src/UserAgentApplication.ts
sameerag: Looks okay but the name bugs me. ATS would have to be ideally silentLogin - but looks like we are adding it for cases of adal idToken -> ATS route only previously. Using the same for an idToken renewal is confusing, Having said that I cannot yet think of a better option. Let's discuss this in person once.

pkanher617: I think this variable doesn't even need to be there, but we can discuss that another time. For now we should approve and merge this.

sameerag: That I agree. Maybe no sense in renaming it now.

PR #875, opened 2019-07-26
Fixing issue where input scopes parameter was being mutated

pkanher617:
Please review PR #805 and #806 before reviewing this one.

This fixes the other part of the issue in #736 where the scopes object that is passed in by the user is being mutated to add "openid" and "profile". This behavior has been moved rectified by using the Array.slice() method.

### DIFF
--- a/lib/msal-core/src/ServerRequestParameters.ts
+++ b/lib/msal-core/src/ServerRequestParameters.ts
@@ -54,7 +54,11 @@ export class ServerRequestParameters {
   constructor (authority: Authority, clientId: string, scope: Array<string>, responseType: string, redirectUri: string, state: string) {
     this.authorityInstance = authority;
     this.clientId = clientId;
-    this.scopes = scope;
+    if (!scope) {
+      this.scopes = [clientId];
+    } else {
+      this.scopes = scope.slice(0);
+    }
 
     this.nonce = Utils.createNewGuid();
     this.state = state && !Utils.isEmpty(state) ?  Utils.createNewGuid() + "|" + state   : Utils.createNewGuid();
--- a/lib/msal-core/src/ServerRequestParameters.ts
+++ b/lib/msal-core/src/ServerRequestParameters.ts
@@ -57,7 +57,7 @@ export class ServerRequestParameters {
     if (!scope) {
       this.scopes = [clientId];
     } else {
-      this.scopes = scope.slice(0);
+      this.scopes = scope.slice();
     }
 
     this.nonce = Utils.createNewGuid();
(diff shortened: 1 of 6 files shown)


lib/msal-core/src/ServerRequestParameters.ts
```diff
@@ -54,7 +54,11 @@ export class ServerRequestParameters {
   constructor (authority: Authority, clientId: string, scope: Array<string>, responseType: string, redirectUri: string, state: string) {
     this.authorityInstance = authority;
     this.clientId = clientId;
-    this.scopes = scope;
+    if (!scope) {
+      this.scopes = [clientId];
+    } else {
+      this.scopes = scope.slice();
```
DarylThayil: what does slice with no params do here? return back the original array?

lib/msal-core/src/ServerRequestParameters.ts
```diff
@@ -54,7 +54,11 @@ export class ServerRequestParameters {
   constructor (authority: Authority, clientId: string, scope: Array<string>, responseType: string, redirectUri: string, state: string) {
     this.authorityInstance = authority;
     this.clientId = clientId;
-    this.scopes = scope;
+    if (!scope) {
```
DarylThayil: can we add a test to make sure that what we are fixing is fixed

lib/msal-core/src/ServerRequestParameters.ts
pkanher617: Yes that's correct:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice

DarylThayil: If it does nothing why call it

lib/msal-core/src/ServerRequestParameters.ts
pkanher617: Added tests

lib/msal-core/src/ServerRequestParameters.ts
jasonnutter: So that any mutations on it don't mutate the passed in array. 

You can do this via ES6 like so:

```js
this.scopes = [ ...scope ];
```

pkanher617: It returns a copy instead of the original, so we don't have any mutation. @jasonnutter is it better to use ES6 notation than .slice()?

jasonnutter: It looks cooler :) should be functionally equivalent.

PR #886, opened 2019-07-30
Completely remove Utils.ts and make smaller Utility files

pkanher617:
This is a continuation of efforts to refactor, this was an ask by @jasonnutter to make the Utils.ts file a little cleaner. This should be the last major change to our utils structure.


lib/msal-core/src/utils/Constants.ts
```diff
@@ -146,8 +146,8 @@ export const PromptState = {
...
 
 /**
- * @hidden
+ * MSAL JS Library Version
  */
-export const Library = {
-  version: "1.1.1"
-};
+export function libraryVersion(): string {
+  return "1.1.1";
```
jasonnutter: At some point we should pull this directly from the `package.json`.

Issue #899, opened 2019-07-31
Msal Js Supports Iframe In Iframe scenarios


DarylThayil:
As it currently stands msal js blocks certain scenarios when rendered inside of an iframe because there are edge case scenarios that can not be supported.

The goal is to unblock these so that it can be supported. 

Linking all the relevant PRs for this ask:

**Beta release**
Released `msal@1.2.0-beta.0` [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/wiki/MSAL-JS:-1.2.0-beta.0)
- [x] #939: Allows iframed applications call loginPopup(), acquireTokenPopup() and acquireTokenSilent().
- [x] #983: Cache cleanup

**Next beta release contents**
- [x] #975: Allows iframed applications call loginRedirect(), acquireTokenRedirect() by establishing a trust-relationship model with the top frame. This assumes the top-frame also runs MSAL JS.
- [x] #1010 , #994  : Supporting multiple instances of MSAL JS and relevant cache cleanup. 
This is also a prerequisite for #975 and allows concurrent `acquireTokenSilent()` calls.
- [x] #1042: Clear temporary cache on error exit paths
- [ ] Documentation of the protocols/logic introduced for the above functionality to work

mozart27:
What's the time-frame for this to be available for us?

NickSevens:
@mozart27 they mention end of September [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/504#issuecomment-521039037). Hope they hold up to it 😄

DarylThayil:
We are on track to deliver at end of September!  We have some beta versions being  tested with customers.  Please let us know If you would like to test some betas with us! @sameerag

yuriys-kentico:
Hi @DarylThayil, what would it take for me to use a beta version that supports this? I want to see if my use case applies.

AndrewCraswell:
Would this unblock the scenarios where MSAL authentication in Teams apps breaks? If I recall Teams renders tab app inside an iframe so there were authentication issues trying to use MSAL...

sameerag:
@AndrewCraswell Teams app's requirements have some specific restrictions on the iframe applications, we are in internal discussions with them. Our solution is generic, and ideally should work but your specific scenario may need some fine tuning from the "Teams" app too.

sameerag:
[...]

PR #909, opened 2019-08-02
Do not merge : When rendered in an iframe, silent rejects

### LINKED ISSUE #902: Does not follow reject Promise calling acquireTokenSilent  if in Iframe
## I'm submitting a...

<pre><code>
[X ] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Other... Please describe:
</code></pre>

## Browser:
- [X ] Chrome version Version 76.0.3809.87 (Official Build) (64-bit)
- [ ] Firefox version XX
- [ ] IE version XX
- [X ] Edge version DevEdge 77.0.230.2 (Official build) dev (64-bit)
- [ ] Safari version XX

## Library version

<pre><code>
Library version: 1.1.1

</code></pre>

## Current behavior

When making a call to msal.acquireTokenSilent,   if the page is embedded in an Iframe it does not work and neither does it return an error.

## Expected behavior

Now I know that you do not want to support IFRAMEs in this library from earlier tickets, but when this is called it needs to obey the promise pattern and either succeed or fail.

In this case it should return an error, preferably with a new error code to say we do not support iframes as part of the promise.

In the UserAgentApplication.ts the decorator resolveTokenOnlyIfOutOfIframe
checks if in Iframe and returns a new promise,  but does nothing with that promise subsequently,    if not in an iframe a differnt promise is called and resolves or rejects correctly.
That promise needs to be rejected with an iframe error warning please.

If you take the  QuickStart Code and make that work stand alone,   then embed that page inside an Iframe in another Website (pref not both localhost, I have not tried on localhost or same websites so this is how I set it up).     you can actually login it will popup the Dialog box, but it will not do anything.

DarylThayil:
... with an appropriate error.
* Get rid of the decorator function because it is only used once
* Add EnvironmentError
* tests

resolves #902

### DIFF
--- a/lib/msal-core/src/error/EnvironmentError.ts
+++ b/lib/msal-core/src/error/EnvironmentError.ts
@@ -1,0 +1,33 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { AuthError } from "./AuthError";
+import { IdToken } from "../IdToken";
+import { StringUtils } from "../utils/StringUtils";
+
+export const EnvironmentErrorMessage = {
+    noResolveInIframe: {
+        code: "no_resolve_in_iframe",
+        desc: `Msal.js does not currently support resolving a silent token request in an iframe`
+    }
+};
+
+/**
+ * Error thrown when there is an error in the client code running on the browser.
+ */
+export class EnvironmentError extends AuthError {
+
+    constructor(errorCode: string, errorMessage?: string) {
+        super(errorCode, errorMessage);
+        this.name = "EnvironmentError";
+
+        Object.setPrototypeOf(this, EnvironmentError.prototype);
+    }
+
+    static createNoResolveInIframeError(): EnvironmentError {
+        const { code, desc } = EnvironmentErrorMessage.noResolveInIframe
+        return new EnvironmentError(code, desc);
+    }
+
+
+}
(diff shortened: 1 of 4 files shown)


lib/msal-core/src/error/EnvironmentError.ts
```diff
@@ -0,0 +1,33 @@
...
+
+        Object.setPrototypeOf(this, EnvironmentError.prototype);
+    }
+
+    static createNoResolveInIframeError(): EnvironmentError {
+        const { code, desc } = EnvironmentErrorMessage.noResolveInIframe
+        return new EnvironmentError(code, desc);
+    }
+
+
```
DarylThayil: remove extra space

DarylThayil:
This causes issues with the way that pages are rerendered within popup on return from aad

binaryjam:
Is there a way I can help test this, as I get the code from a cdn normally and am not sure what to do here.  I would have difficulty compiling and building this locally.

DarylThayil:
@binaryjam  no problem. I think there are unfortunately some architectural issues with that prevent us from making thing change right now

@sameera is looking to remove the reliance on not allowing to be rendered in an iframe.  Sameera this is something we should consider because when you remove the restriction, and the original app is rendered in the iframe to get the token response, there may be another ATS call and create an endless loop

sameerag:
@DarylThayil Yes, I am considering that and indeed it is a well-known issue in ADAL which needed the developer to add a condition to prevent the loop, that was remedied by this decorator in MSAL.

jasonnutter:
Closing, as this will be addressed by the work to support iframes in #899.

PR #910, opened 2019-08-02
fix lerna vulnerability

DarylThayil:
update package lock to deep resolve a safe version of lodash in lerna


jasonnutter:
@DarylThayil I think you need to pull master to get the ADO pipelines working properly so this can be merged.

PR #912, opened 2019-08-05

pkanher617:


lib/msal-core/azure-pipelines.yml
```diff
@@ -0,0 +1,29 @@
+# Node.js
+
+# Build a general Node.js project with npm.
+# Add steps that analyze code, save build artifacts, deploy, and more:
+# https://docs.microsoft.com/azure/devops/pipelines/languages/javascript
+
+jobs:
+  - job: rolling_VS2017_build
+    displayName: 'Build Job for JS (VS2017)'
```
DarylThayil: what is vs 2017

DarylThayil:
is publish in this or only build?

lib/msal-core/azure-pipelines.yml
jasonnutter: Visual Studio 2017

pkanher617: This is the server that we are using to build.

pkanher617:
Publish is in VSTS as a separate task.

lib/msal-core/build/npm-build-steps.yml
```diff
@@ -0,0 +1,19 @@
...
+
+steps:
+- task: NodeTool@0
+  inputs:
+    versionSpec: '10.x'
+  displayName: 'Install Node.js'
+
+- task: Npm@1
+  inputs:
+    command: 'install'
```
jasonnutter: If you have it run `npm ci` instead of `npm install` it'll go faster.

PR #925, opened 2019-08-14
Adding docs and samples for Conditional access and claims request

pkanher617:


lib/msal-core/README.md
```diff
@@ -49,6 +49,9 @@ We offer two methods of storage for Msal, `localStorage` and `sessionStorage`.
 ### forceRefresh to skip cache
 If you would like to skip a cached token and go to the server, please pass in the boolean `forceRefresh` into the `AuthenticationParameters` object used to make a login / token request. **WARNING:** This should not be used by default, because of the performance impact on your application.  Relying on the cache will give your users a better experience, and skipping it should only be used in scenarios where you know the current cached data does not have up to date information.  Example: Admin tool to add roles to a user that needs to get a new token with updates roles.
 
+### Conditional Access and Claims Requests
+MSAL.js supports [conditional access scenarios](https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-handling-exceptions#javascript) enabled in Azure AD. You may also [request additional claims]() for your application. Please review the linked MSAL Wiki pages. To understand more about how to enable conditional access for your application, or common conditional access scenarios, please review the docs [here](https://docs.microsoft.com/en-us/azure/active-directory/conditional-access/overview).
```
jasonnutter: Looks like you forgot the url here: `[request additional claims]()`

lib/msal-core/test/AuthenticationParameters.spec.ts
```diff
@@ -1,7 +1,43 @@
 import { expect } from "chai";
 import sinon from "sinon";
+import { AuthenticationParameters, validateClaimsRequest } from "../src/AuthenticationParameters"
+import { TEST_CLAIMS_REQ } from "./TestConstants";
+import { ClientConfigurationError, ClientConfigurationErrorMessage } from "../src/error/ClientConfigurationError";
 
 describe("Authentication Parameters (A.K.A. Request Object)", function () {
 
+    it("doesn't throw error if claims string can be parsed into JSON", () => {
+        let req: AuthenticationParameters = {
```
DarylThayil: use consts where possible

lib/msal-core/README.md
pkanher617: Yes waiting until the page is up.

pkanher617:
> General comment: So does CA work end-to-end now in MSAL JS? Publishing this readme would mean folks will start trying end-to-end cases, hence the question.

@sameerag CA has worked for a while now, this is supporting the use case for requesting additional claims as well as a use case for resources to present claims challenges.

AndrewCraswell:
@pkanher617 is there any guidance for how to use the `claimsRequest` property in the Authentication Parameters? I ran into an issue this week where someone needed to request country code from AAD optional claims to implement EU cookie compliance banner. Was hoping maybe this property was a mechanism to accomplish that?
https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-optional-claims

pkanher617:
@AndrewCraswell sorry for the delayed response. It's a bit of a complicated answer, but if the token that is being requested is coming from a resource that the user controls, they can edit the manifest to send optional claims in the token without the library needing to know about it. However, if the resource is outside of the users control, and the desired claim isn't present in the token, you should ideally be able to request that claim by passing it in as a stringified key:value pair to the `claimsRequest` property. However I haven't been able to get the latter scenario to work, and I am working with the server to understand the extent of the claimsRequest feature from the client side.

jasonnutter:
@pkanher617 Can this be closed?

pkanher617:
Yes this can be closed, I will follow this up at a later date.

PR #928, opened 2019-08-15
Fix error with no application in edge

tomaszzmuda:
There is a problem when we try to open new window with "about:blank" in url. We get error: 
"You'll need a new app to open this about" with link to the app store with empty list of applications supports this.

image

I used it to wrote an Add-In to Ms Office products so popup window in edge/internet explorer is the only option to login.

What have I changed:
Move a bit lines with opening popup and navigate directly to login page, not about:blank.

### DIFF
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -481,16 +481,6 @@ export class UserAgentApplication {
     let serverAuthenticationRequest: ServerRequestParameters;
     const acquireTokenAuthority = (!isLoginCall && request && request.authority) ? AuthorityFactory.CreateInstance(request.authority, this.config.auth.validateAuthority) : this.authorityInstance;
 
-    let popUpWindow: Window;
-    if (interactionType === Constants.interactionTypePopup) {
-      // Generate a popup window
-      popUpWindow = this.openWindow("about:blank", "_blank", 1, this, resolve, reject);
-      if (!popUpWindow) {
-        // We pass reject in openWindow, we reject there during an error
-        return;
-      }
-    }
-
     acquireTokenAuthority.resolveEndpointsAsync().then(() => {
       // On Fulfillment
       const responseType: string = isLoginCall ? ResponseTypes.id_token : this.getTokenType(account, scopes, false);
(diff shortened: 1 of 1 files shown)


jasonnutter:
Yeah I'll take a look and test this out.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -538,14 +528,22 @@ export class UserAgentApplication {
...
+      }
+      else {
+        // prompt user for interaction
+        this.navigateWindow(urlNavigate, popUpWindow);
+      }
     }).catch((err) => {
       this.logger.warning("could not resolve endpoints");
       this.authErrorHandler(interactionType, ClientAuthError.createEndpointResolutionError(err.toString), buildResponseStateOnly(request.state), reject);
-      if (popUpWindow) {
-        popUpWindow.close();
```
jasonnutter: Why was this removed?

jasonnutter:
@tomaszzmuda I'm not able to reproduce this on my machine, as Edge is able to open the pop up with `about:blank` with no issue. As the issue templates asks, can you please provide more details about how to reproduce this (e.g. you mention you are making an office add-in)? Thanks!

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -538,14 +528,22 @@ export class UserAgentApplication {
...
 
-      // prompt user for interaction
-      this.navigateWindow(urlNavigate, popUpWindow);
+      let popUpWindow: Window;
+      if (interactionType === Constants.interactionTypePopup) {
+        // Generate a popup window
+        popUpWindow = this.openWindow(urlNavigate, "_blank", 1, this, resolve, reject);
+        if (!popUpWindow) {
+          // We pass reject in openWindow, we reject there during an error
+          return;
```
jasonnutter: My understanding is that this code is attempting to open a popup to a blank page, to check if popups are allowed (i.e. `popUpWindow` will be `null` if popups are blocked), and then short-circuits the `acquireTokenHelper` function if that happens. This ensures none that `acquireTokenAuthority.resolveEndpointsAsync` and the code in its `.then` don't run unnecessarily. I'm assuming this is done intentionally, to prevent orphaned callbacks (which are registered above).

tomaszzmuda: If the browser cannot open the link, the code hangs on line with "openWindow".

lib/msal-core/src/UserAgentApplication.ts
tomaszzmuda: I assume that was added to close popUp if some of the previous actions fail (access to cache, build url). Right now, show popup is the last action so if popUp won't open it doesn't have to be closed.

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: > If the browser cannot open the link, the code hangs on line with "openWindow".

Specifically in the scenario you describe above, correct? I'm referring to scenarios such as the browser having popups blocked, for example.

tomaszzmuda: Yes. I didn't find any other scenarios when it doesn't work. I still didn't find why it hangs ;/

pkanher617:
@tomaszzmuda I haven't seen this issue on Edge on my devices either. Is this happening with a specific version of Edge? I would also recommend filing an issue, as we can track it as a bug rather than a PR.

jasonnutter:
@pkanher617 I have seen other issues related to `about:blank` (such as #1006), so it might be worth seeing if we can determine why we navigate to `about:blank` first, and if it still needed, given it seems to cause issues in certain situations.

jasonnutter:
Closing, as this has been addressed.

PR #934, opened 2019-08-19
add prefer const rule

DarylThayil:
Add prefer const rule and fix for vars never reassigned

### DIFF
--- a/lib/msal-core/src/ServerRequestParameters.ts
+++ b/lib/msal-core/src/ServerRequestParameters.ts
@@ -275,7 +275,7 @@ export class ServerRequestParameters {
                 break;
             }
             case SSOTypes.HOMEACCOUNT_ID: {
-                let homeAccountId = ssoData.split(".");
+                const homeAccountId = ssoData.split(".");
                 const uid = CryptoUtils.base64Decode(homeAccountId[0]);
                 const utid = CryptoUtils.base64Decode(homeAccountId[1]);
 
@@ -311,7 +311,7 @@ export class ServerRequestParameters {
    * @param request
    */
     private sanitizeEQParams(request: AuthenticationParameters) : StringDict {
-        let eQParams : StringDict = request.extraQueryParameters;
+        const eQParams : StringDict = request.extraQueryParameters;
         if (!eQParams) {
             return null;
         }
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -423,12 +423,12 @@ export class UserAgentApplication {
         if (!account && !ServerRequestParameters.isSSOParam(request)) {
             if (isLoginCall) {
                 // extract ADAL id_token if exists
-                let adalIdToken = this.extractADALIdToken();
+                const adalIdToken = this.extractADALIdToken();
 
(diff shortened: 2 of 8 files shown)


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1474,7 +1474,7 @@ export class UserAgentApplication {
   /* tslint:disable:no-string-literal */
   private saveAccessToken(response: AuthResponse, authority: string, parameters: any, clientInfo: string, idTokenObj: IdToken): AuthResponse {
       let scope: string;
-      let accessTokenResponse = { ...response };
+      const accessTokenResponse = { ...response };
```
pkanher617: On line 1510, we assign new values to accessTokenResponse. We can't make this a const.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1337,19 +1337,19 @@ export class UserAgentApplication {
...
+              const idTokenObj = new IdToken(accessTokenCacheItem.value.idToken);
               if (!account) {
                   account = this.getAccount();
                   if (!account) {
                       throw AuthError.createUnexpectedError("Account should not be null here.");
                   }
               }
               const aState = this.getAccountState(serverAuthenticationRequest.state);
-              let response : AuthResponse = {
+              const response : AuthResponse = {
```
pkanher617: We can't make this const, in setResponseIdToken we modify the idToken field.

lib/msal-core/src/UserAgentApplication.ts
DarylThayil: Const only restricts the memory reference of the top level item so in this case an object is created with a memory reference

When we assign values to the references inside that object we are not changing the memory reference of the accessTokenResponse 
```js
const accessTokenResponse = {};
accessTokenResponse = {} // not allowed
acessTokenResponse.foo = 'bar' //allowed
```

lib/msal-core/src/UserAgentApplication.ts
DarylThayil: see previous response

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -587,7 +587,7 @@ export class UserAgentApplication {
 
             const responseType = this.getTokenType(account, request.scopes, true);
 
-            let serverAuthenticationRequest = new ServerRequestParameters(
+            const serverAuthenticationRequest = new ServerRequestParameters(
```
pkanher617: if we make this const, does changing fields in the object throw an error?

pkanher617: I believe we change fields in some of the functions.

lib/msal-core/src/ServerRequestParameters.ts
```diff
@@ -311,7 +311,7 @@ export class ServerRequestParameters {
    * @param request
    */
     private sanitizeEQParams(request: AuthenticationParameters) : StringDict {
-        let eQParams : StringDict = request.extraQueryParameters;
+        const eQParams : StringDict = request.extraQueryParameters;
```
pkanher617: We modify eqParams down below, we can't make this const.

PR #935, opened 2019-08-19
add no var rule to eslint

DarylThayil:
and update instances of it


pkanher617:
Seems like there are some duplicates with #934?

DarylThayil:
@pkanher617  this enforces a different rule, it says no var, and the fix is turning them from var to either let or const

lib/msal-core/.eslintrc.json
```diff
@@ -26,8 +26,8 @@
         "@typescript-eslint/no-object-literal-type-assertion": 0,
         "@typescript-eslint/prefer-interface": 0,
         "@typescript-eslint/type-annotation-spacing": 0,
-        "no-var": 0,
-        "prefer-const": 0,
+        "no-var": 2,
+        "prefer-const": 2,
```
jasonnutter: Looks like this is doing both rules, I'm guessing you created this branch from the other branch?

jasonnutter: I think pulling master into this branch will clean this up.

DarylThayil: oh snap, updated

Issue #884, opened 2019-07-30
acquireTokenSilent raise a page refresh


m4ss1m0g:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[X] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Other... Please describe:
</code></pre>

## Browser:
- [X] Chrome version 75
- [ ] Firefox version XX
- [ ] IE version XX
- [ ] Edge version XX
- [ ] Safari version XX

## Library version

<pre><code>
Library version: 1.1.1
</code></pre>

## Current behavior
When calling the `acquireTokenSilent` **AND** specify more than ONE scope on ``AuthenticationPramaters` the lib cannot find the cached token

I found a possible bug on **Utils.js** on method **containsScope** that return always false when requiring `scopes: [`api://${this._settings!.clientId}/access_as_user`, 'openid']`

`
Utils.containsScope = function (cachedScopes, scopes) {
        cachedScopes = this.convertToLowerCase(cachedScopes);
        return scopes.every(function (value) { return cachedScopes.indexOf(value.toString().toLowerCase()) >= 0; });
    };
`

## Expected behavior
Do not refresh the page

## Minimal reproduction of the problem with instructions
1. Specify more that one scope on request
2. Complete the authentication
3. Hit the refresh button
4. You see the log that the page is refreshed

## TEMP Workaround
Specify ONLY one scope
 `scopes: [`api://${this._settings!.clientId}/access_as_user`]`

## See also
#526

DarylThayil:
@pkanher617  please comment on status of fix you are working on

jasonnutter:
Closing a duplicate, see the PR linked above for progress on a fix.

PR #939, opened 2019-08-20
Poll popups and iframes for hash instead of relying on nested MSAL instance to call back into parent

jasonnutter:
Currently, for ATS and popup flows, we rely on the instance of MSAL running inside of the iframe/popup to callback into the top-level MSAL when it is loaded with a hash url (which contains the token response). This requires MSAL to be running in that frame, which has performance overhead (both to load the page server-side and to load and execute the Javascript client-side).

This PR changes the paradigm to have the MSAL instance which created the popup/iframe monitor the frame until it receives a hash response, at which points it parses the response and then closes the frame. This should be faster than the previous approach, as we don't have to wait until the MSAL running in the frame is loaded and executed to receive the response.

To prevent the instance of MSAL running in hidden iframes from executing (which could cause unintended side-effects, and also has a performance penalty), hidden iframes will be sandboxed, which prevents all JS from running inside the frame.

This also means app developers can set the redirect uri for their requests to a page that doesn't include MSAL or any content on it, which would result in further performance gains.

**Open questions**

When the app is loaded in the hidden/sandboxed iframe, there is a scary error message that gets shown in the dev console:

Annotation 2019-08-20 142338

Not sure how much we can do about this, other than providing good documentation warning about this, and also letting developers know they can change their redirect to a page with no JS on it at all.

**Impacted issues**
- #884 
- #526 

**To do**
- [x] Complete error handling
- [x] Tests
- [x] Merge conflicts 

**Prior art**
Thanks to @step-lee for his work in #622, which served as a POC for this work.


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -540,6 +540,9 @@ export class UserAgentApplication {
 
       // prompt user for interaction
       this.navigateWindow(urlNavigate, popUpWindow);
+
+      const hash = await this.pollWindowForHash(popUpWindow);
```
DarylThayil: would this ever throw / catch?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -852,27 +846,35 @@ export class UserAgentApplication {
...
 
-    setTimeout(() => {
-      const frameHandle = this.addHiddenIFrame(frameCheck);
-      if (frameHandle.src === "" || frameHandle.src === "about:blank") {
-        frameHandle.src = urlNavigate;
-        this.logger.infoPii("Frame Name : " + frameName + " Navigated to: " + urlNavigate);
-      }
-    },
-    this.config.system.navigateFrameWait);
+    return new Promise((resolve) => {
```
DarylThayil: never reject?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -910,6 +913,39 @@ export class UserAgentApplication {
     return adalFrame;
   }
 
+  private pollWindowForHash(contentWindow: Window): Promise<string> {
+      return new Promise((resolve, reject) => {
+          const intervalTimeMs = 50;
```
DarylThayil: Is this the right place for this? is it configurable?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -910,6 +913,39 @@ export class UserAgentApplication {
     return adalFrame;
   }
 
+  private pollWindowForHash(contentWindow: Window): Promise<string> {
```
DarylThayil: This doesn't rely on anything in UAA, can it be in its own file ?

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: I couldn't think of an obvious scenario where this would happen, but I'll definitely add something just in case.

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: It will throw a timeout error if it doesn't find a hash after 6 seconds.

jasonnutter: Which will get caught in the `catch` block below.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -910,6 +913,39 @@ export class UserAgentApplication {
...
+                  if (contentWindow.closed) {
+                      clearInterval(intervalId);
+                 } else if (this.urlContainsHash(href)) {
+                      clearInterval(intervalId);
+                      resolve(contentWindow.location.hash);
+                  } else if (ticks > maxTicks) {
+                      clearInterval(intervalId);
+                      reject(ClientAuthError.createTokenRenewalTimeoutError()); // better error?
+                  }
+              } catch (e) {}
```
DarylThayil: what happens here?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -540,6 +540,9 @@ export class UserAgentApplication {
 
       // prompt user for interaction
       this.navigateWindow(urlNavigate, popUpWindow);
+
+      const hash = await this.pollWindowForHash(popUpWindow);
+      this.handleAuthenticationResponse(hash);
```
sameerag: So shouldn't we restrict the handleAuthenticationResponse() call in the constructor now to only "redirect" cases?

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: This is to catch the cross origin errors that will happen when we try to access the location object of the STS login UI pages, which we want to catch and ignore.

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: Yeah, it would be good to put in an iframe utils file.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -910,6 +913,39 @@ export class UserAgentApplication {
...
+                  }
+
+                  // Only start clock when we are on same domain
+                  ticks++;
+
+                  if (contentWindow.closed) {
+                      clearInterval(intervalId);
+                 } else if (this.urlContainsHash(href)) {
+                      clearInterval(intervalId);
+                      resolve(contentWindow.location.hash);
```
sameerag: How does this work in a sandboxed iframe? Does this not come under scripts?

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: If we move this function to its own file it would probably be defined as a const outside the function, and no, wasn't planning on making it configurable, as that would be leaking implementation details.

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: This will be running in the parent window, which has access to the location object if the iframe is on the same domain.

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: Yeah, good catch.

jasonnutter: Updated the constructor to check that we are not in an iframe, see above.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -491,7 +489,7 @@ export class UserAgentApplication {
       }
     }
 
-    acquireTokenAuthority.resolveEndpointsAsync().then(() => {
+    acquireTokenAuthority.resolveEndpointsAsync().then(async () => {
```
step-lee: nit: prefer not mixing async and Promise.then().
`acquireTokenAuthority.resolveEndpointAsync().then(() => {
....
return this.pollWindowForHash(popUpWindow);
}).then(hash => {
...
});`
Functionality they are the same but consistency IMO

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -817,29 +818,20 @@ export class UserAgentApplication {
...
-  }
-
-  /**
-   * @hidden
-   * Returns whether parent window exists and has msal
-   */
-  private parentIsMsal() {
-    return window.parent !== window && window.parent.msal;
-  }
+    return window.parent !== window;
```
step-lee: nit: consider using `window.frameElement` for additional security.
https://developer.mozilla.org/en-US/docs/Web/API/Window/frameElement

It also returns null if the parent and the embedded element is not the same origin, supporting the case where we want to load an iFrame inside of an iFrame.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -852,27 +844,35 @@ export class UserAgentApplication {
...
-    setTimeout(() => {
-      const frameHandle = this.addHiddenIFrame(frameCheck);
-      if (frameHandle.src === "" || frameHandle.src === "about:blank") {
-        frameHandle.src = urlNavigate;
-        this.logger.infoPii("Frame Name : " + frameName + " Navigated to: " + urlNavigate);
-      }
-    },
-    this.config.system.navigateFrameWait);
+    return new Promise((resolve) => {
+        setTimeout(() => {
```
step-lee: You don't need this setTimeout anymore right? if your create frame and navigate frame become synchronous

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -852,27 +844,35 @@ export class UserAgentApplication {
...
+    this.handleAuthenticationResponse(hash);
   }
 
   /**
    * @hidden
    * Loads iframe with authorization endpoint URL
    * @ignore
    */
-  private loadFrame(urlNavigate: string, frameName: string): void {
+  private loadFrame(urlNavigate: string, frameName: string): Promise<HTMLIFrameElement> {
```
step-lee: nit: [consistency] You can also use async await instead of `new Promise`.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -910,6 +911,39 @@ export class UserAgentApplication {
     return adalFrame;
   }
 
+  private pollWindowForHash(contentWindow: Window): Promise<string> {
+      return new Promise((resolve, reject) => {
```
step-lee: nit: [consistency] You can also use async await instead of `new Promise`.

lib/msal-core/src/UserAgentApplication.ts
step-lee: move the `try{} catch {}` to smaller scope. Maybe only try and catch the `contentWindow.location.href`? And leave a comment explaining the silent `catch`

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1125,53 +1159,36 @@ export class UserAgentApplication {
...
-    // retrieve the hash
-    if (hash == null) {
-      hash = window.location.hash;
-    }
+    const locationHash = hash || window.location.hash;
 
-    let self = null;
-    let isPopup: boolean = false;
-    let isWindowOpenerMsal = false;
+    let isPopupOrIframe = false;
```
step-lee: nit: consider rename let to `isTopWindow`? not clear to me what the boolean is indicating.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1180,18 +1197,16 @@ export class UserAgentApplication {
...
-    self.processCallBack(hash, stateInfo, tokenResponseCallback);
+    this.processCallBack(locationHash, stateInfo, tokenResponseCallback);
 
     // If current window is opener, close all windows
-    if (isWindowOpenerMsal) {
-      for (let i = 0; i < window.opener.openedWindows.length; i++) {
-        window.opener.openedWindows[i].close();
-      }
+    if (isPopupOrIframe) {
+        window.openedWindows.forEach(openedWindow => openedWindow.close());
```
step-lee: Should we clean up iFrame here as well? `.close()` only close the popup window

lib/msal-core/samples/react-sample-app/src/AuthProvider.js
```diff
@@ -11,6 +11,7 @@ import {
 
 // If you support IE, our recommendation is that you sign-in using Redirect APIs
 const useRedirectFlow = isIE();
+// const useRedirectFlow = true;
```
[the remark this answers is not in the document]
jasonnutter: This is in a sample, so I'm gonna keep it, since I use this a lot for testing.

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: No, we still want to include the delay for IE.

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: I can't actually, since the promise will be resolved inside the `setInterval`.

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: Need the new `Promise` since we need the `setTimeout`.

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: I'll take another look to make sure we are closing the iframes.

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: `isPopupOrIframe` represents whether or not the flow is a hidden iframe or a popup, as opposed to redirect (which would be the top frame). I'll see if there's a better name.

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: I get what you're saying, I think I was trying to be safe here, as the only errors we want to happen here are timeout errors that we create.

jasonnutter: Agree about adding the comment, though.

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: Ah didn't know about that, I'll take a look at it.

jasonnutter: Although I think we would want to know about the parent if we are in a cross domain scenario?

lib/msal-core/azure-pipelines.yml
```diff
@@ -27,3 +27,8 @@ jobs:
           archiveType: 'zip'
           archiveFile: '$(Build.ArtifactStagingDirectory)/$(Build.BuildId).zip'
           replaceExistingArchive: true
+      - task: PublishBuildArtifacts@1
+        inputs:
+          PathtoPublish: '$(Build.ArtifactStagingDirectory)'
+          ArtifactName: 'drop'
+          publishLocation: 'Container'
```
jasonnutter: @pkanher617 I accidently pulled master into my branch instead of dev, any issues with these changes going into dev?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -253,13 +253,11 @@ export class UserAgentApplication {
...
         const urlHash = window.location.hash;
-        const urlContainsHash = this.urlContainsHash(urlHash);
+        const urlContainsHash = UrlUtils.urlContainsHash(urlHash);
 
         // On the server 302 - Redirect, handle this
-        if (!this.config.framework.isAngular) {
-            if (urlContainsHash) {
-                this.handleAuthenticationResponse(urlHash);
-            }
+        if (!this.config.framework.isAngular && urlContainsHash && !IframeUtils.isInIframe()) {
```
jasonnutter: May need to specifically prevent the popup scenario instead of the iframe scenario.

jasonnutter: This should only run in the redirect case.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1135,85 +1054,58 @@ export class UserAgentApplication {
...
-            isWindowOpenerMsal = false;
-        }
-
-        // Set the self to the window that created the popup/iframe
-        if (isWindowOpenerMsal) {
-            self = window.opener.msal;
-            isPopup = true;
-        } else if (window.parent && window.parent.msal) {
-            self = window.parent.msal;
+            isPopupOrIframe = false;
```
jasonnutter: Double check this `try/catch` is needed still.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1135,85 +1054,58 @@ export class UserAgentApplication {
...
+                if (window.parent === window) {
+                    window.location.href = this.cacheStorage.getItem(Constants.loginRequest, this.inCookie);
                 }
                 return;
             }
-            else {
-                window.location.hash = "";
+            // Current window is window opener (popup)
+            else if (isPopupOrIframe) {
+                tokenResponseCallback = window.opener.callbackMappedToRenewStates[stateInfo.state];
```
jasonnutter: This can just be `window.callbackMappedToRenewStates`

jasonnutter: Double check if needed at all.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1135,85 +1054,58 @@ export class UserAgentApplication {
...
-        self.processCallBack(hash, stateInfo, tokenResponseCallback);
+        this.processCallBack(locationHash, stateInfo, tokenResponseCallback);
 
         // If current window is opener, close all windows
-        if (isWindowOpenerMsal) {
-            for (let i = 0; i < window.opener.openedWindows.length; i++) {
-                window.opener.openedWindows[i].close();
-            }
+        if (isPopupOrIframe) {
+            window.openedWindows.forEach(openedWindow => openedWindow.close());
```
jasonnutter: Revisit that we want to close both iframes and popups.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -541,6 +539,9 @@ export class UserAgentApplication {
 
             // prompt user for interaction
             this.navigateWindow(urlNavigate, popUpWindow);
+
+            const hash = await IframeUtils.monitorWindowForHash(popUpWindow, this.config.system.loadFrameTimeout);
```
jasonnutter: Won't run for redirect. Can add check for truthiness of `popUpWindow` (and comment).

lib/msal-core/src/utils/IframeUtils.ts
```diff
@@ -0,0 +1,119 @@
+import { ClientAuthError } from "../error/ClientAuthError";
+import { UrlUtils } from "./UrlUtils";
+import { Logger } from "../Logger";
+
+export class IframeUtils {
+    private static POLLING_INTERAL_MS = 50;
```
jasonnutter: Fix spelling.

lib/msal-core/src/utils/IframeUtils.ts
```diff
@@ -0,0 +1,119 @@
...
+
+                // Only start clock when we are on same domain
+                ticks++;
+
+                if (contentWindow.closed) {
+                    clearInterval(intervalId);
+                } else if (UrlUtils.urlContainsHash(href)) {
+                    clearInterval(intervalId);
+                    resolve(contentWindow.location.hash);
+                } else if (ticks > maxTicks) {
```
jasonnutter: Review for any race cases.

jasonnutter: Look at swapping first two if statements.

lib/msal-core/src/utils/IframeUtils.ts
```diff
@@ -0,0 +1,119 @@
...
+
+        return new Promise((resolve) => {
+            setTimeout(() => {
+                const frameHandle = IframeUtils.addHiddenIFrame(frameCheck, logger);
+                if (frameHandle.src === "" || frameHandle.src === "about:blank") {
+                    frameHandle.src = urlNavigate;
+                    logger.infoPii("Frame Name : " + frameName + " Navigated to: " + urlNavigate);
+                }
+
+                resolve(frameHandle);
```
jasonnutter: Any error scenarios?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1135,85 +1054,49 @@ export class UserAgentApplication {
...
+        const isPopupOrIframe = !!(iframeWithHash || popUpWithHash);
 
         // if (window.parent !== window), by using self, window.parent becomes equal to window in getResponseState method specifically
-        const stateInfo = self.getResponseState(hash);
+        const stateInfo = this.getResponseState(locationHash);
 
         let tokenResponseCallback: (response: AuthResponse, error: AuthError) => void = null;
 
-        self.logger.info("Returned from redirect url");
+        this.logger.info("Returned from redirect url");
```
sameerag: May be change this log statement to "returns from server redirect ..." to not confuse with our redirect flows.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1135,85 +1054,49 @@ export class UserAgentApplication {
...
-        self.processCallBack(hash, stateInfo, tokenResponseCallback);
+        this.processCallBack(locationHash, stateInfo, tokenResponseCallback);
 
         // If current window is opener, close all windows
-        if (isWindowOpenerMsal) {
-            for (let i = 0; i < window.opener.openedWindows.length; i++) {
-                window.opener.openedWindows[i].close();
-            }
+        if (isPopupOrIframe) {
+            WindowUtils.closePopups();
```
sameerag: So is it decided not to close iframes in this PR?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1422,7 +1305,9 @@ export class UserAgentApplication {
     private renewToken(scopes: Array<string>, resolve: Function, reject: Function, account: Account, serverAuthenticationRequest: ServerRequestParameters): void {
         const scope = scopes.join(" ").toLowerCase();
         this.logger.verbose("renewToken is called for scope:" + scope);
-        const frameHandle = this.addHiddenIFrame("msalRenewFrame" + scope);
+
+        const frameName = `msalRenewFrame${scope}`;
```
sameerag: Why are we adding scope? Is it to distinguish between renew calls as we keep the iframe open? I have always had my doubts on using scope as a distinguisher as the developer can make sequential calls to acquireTokenSilent with the same scopes. Can you please check this?

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: For now, no.

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: I think it just reuses the same iframe for calls with the same scopes. Not clear why.

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: I can fix that in another PR. don't want have too much scope creep.

lib/msal-core/src/utils/WindowUtils.ts
```diff
@@ -0,0 +1,202 @@
...
+     * Adds the hidden iframe for silent token renewal.
+     * @ignore
+     */
+    static addHiddenIFrame(iframeId: string, logger: Logger): HTMLIFrameElement {
+        if (typeof iframeId === "undefined") {
+            return null;
+        }
+
+        logger.info("Add msal frame to document:" + iframeId);
+        let adalFrame = document.getElementById(iframeId) as HTMLIFrameElement;
```
DarylThayil: i know you didnt write this, but can we rename the `adalFrame` it has nothing to do with adal and adds a lot of confusion

lib/msal-core/src/utils/WindowUtils.ts
```diff
@@ -0,0 +1,202 @@
...
+
+        return adalFrame;
+    }
+
+    /**
+     * @hidden
+     * Find and return the iframe element with the given hash
+     * @ignore
+     */
+    static getIframeWithHash(hash: string) {
```
DarylThayil: function return type?

lib/msal-core/src/utils/WindowUtils.ts
```diff
@@ -0,0 +1,202 @@
...
+        return window.openedWindows;
+    }
+
+    /**
+     * @hidden
+     * Find and return the popup with the given hash
+     * @ignore
+     */
+    static getPopUpWithHash(hash: string) {
+        return WindowUtils.getPopups().find(popup => {
```
DarylThayil: function return type

lib/msal-core/src/utils/WindowUtils.ts
```diff
@@ -0,0 +1,202 @@
...
+            }
+        });
+    }
+
+    /**
+     * @hidden
+     * Add the popup to the known list of popups
+     * @ignore
+     */
+    static trackPopup(popup: Window) {
```
DarylThayil: : void

lib/msal-core/src/utils/WindowUtils.ts
```diff
@@ -0,0 +1,202 @@
...
+        WindowUtils.getPopups().push(popup);
+    }
+
+    /**
+     * @hidden
+     * Close all popups
+     * @ignore
+     */
+    static closePopups() {
```
DarylThayil: :void

DarylThayil: would there ever be anything else in the openedWindows than popups?

lib/msal-core/src/utils/WindowUtils.ts
jasonnutter: We should add a lint rule for this.

lib/msal-core/src/utils/WindowUtils.ts
jasonnutter: I'm going to come back later and clean up this function, for now going to keep it as-is since its copy/pasta.

lib/msal-core/src/utils/WindowUtils.ts
jasonnutter: We're gonna hopefully remove `openedWindows` soon.

PR #942, opened 2019-08-22
Update ClientConfigurationError.ts

donaldpipowitch:
There was an API change: https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/619


pkanher617:
@donaldpipowitch Thanks for your submission. It looks like there are some issues with our CI builds with accepting merges from forked repositories. I am looking into fixing it, but in the interest of getting this change in, I have created a mergeable PR in #1110. I will close this for now and track it in that PR.

PR #959, opened 2019-08-29
Remove non-working samples and add a redirect sample

sameerag:
Updating the samples:

- removing unused samples under VanillaJSTestApp folder
- adding index_redirect.html for redirect usecases


lib/msal-core/samples/VanillaJSTestApp/index.html
```diff
@@ -19,17 +19,12 @@
...
-            graphEndpoint: "https://graph.microsoft.com/v1.0/me",
-            loginType: 'POPUP'
-    };*/
 
-    var msalConfig = {
+    // initialize MSAL
+    const msalConfig = {
         auth: {
             clientId: "Enter_your_client_id",
+            // clientId: "245e9392-c666-4d51-8f8a-bfd9e55b2456",
```
DarylThayil: do we want to leave this in here?

jasonnutter: I think so, its helpful to have a demo client id available.

DarylThayil: Is it an internal only client id?

DarylThayil: like is there concern around exposing to others

jasonnutter: Nope, its been in the repo for a while, hasn't it? And its only setup for localhost.

sameerag: Yes. I created this application (in the portal) just for demo purposes, so no issues in making this public. I left it alone here in the sample to have it ready for anyone to test it.

PR #951, opened 2019-08-27
Authority Refactor

sameerag:
First of the authority related changes.

This PR will be followed by:
- Returning 'authority' with the Response
- B2C authority: removing 'validateAuthority' check for B2C authority 
- instance_aware 
- Unit tests

PR #968, opened 2019-09-03
Authority refactor

sameerag:
First of the authority related changes.

This PR will be followed by:

- B2C authority: removing 'validateAuthority' check for B2C authority
- instance_aware
- Unit tests

Note: This PR replaces #951 for a cleaner merge with dev


lib/msal-core/src/authority/B2cAuthority.ts
```diff
@@ -32,19 +32,11 @@ export class B2cAuthority extends AadAuthority {
...
-
-        if (!this.IsValidationEnabled) {
-            return resultPromise;
-        }
-
-        if (this.IsInTrustedHostList(this.CanonicalAuthorityUrlComponents.HostNameAndPort)) {
-            return resultPromise;
+    public async GetOpenIdConfigurationEndpointAsync(): Promise<string> {
+        if (!this.IsValidationEnabled || this.IsInTrustedHostList(this.CanonicalAuthorityUrlComponents.HostNameAndPort)) {
+            return this.DefaultOpenIdConfigurationEndpoint;
```
DarylThayil: this need to return a promise doesn't it ?

return new Promise.resolve(this.defaultEndpoint)

jasonnutter: Its an `async` function so it does inherently.

sameerag: Since it is an `async` function, all return statements are wrapped as promises from my understanding.

DarylThayil: ahh okay makes sense

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1772,6 +1763,23 @@ export class UserAgentApplication {
...
+
+    /**
+     * Set Authority when saving Token from the hash
+     * @param state
+     * @param inCookie
+     * @param cacheStorage
+     * @param idTokenObj
+     * @param response
+     */
+    private populateAuthtority(state: string, inCookie: boolean, cacheStorage: Storage, idTokenObj: IdToken): string {
```
pkanher617: Typo here

pkanher617: Should this be in Authority.ts?

sameerag: Fixed. Please check.

sameerag: nah! We are setting it in UAA for UAA reference. There are many small references to different response/cache elements in UAA we want to refactor as a part of saveTokenFromHash(). If Authority.ts feels right then, we can do it.

DarylThayil:
@sameerag  can this be merged

sameerag:
@DarylThayil Merged now

PR #969, opened 2019-09-03
Instance aware

sameerag:
Add the "cloud_instance_host_name" returned by the eSTS in case of "instance_aware: true" passed as a QueryParameter to the idToken claims.

Tested with the sample.

Pending:
Documentation on the usage.


lib/msal-core/src/IdToken.ts
```diff
@@ -78,6 +79,10 @@ export class IdToken {
                 if (this.claims.hasOwnProperty("sid")) {
                     this.sid = this.claims["sid"];
                 }
+
+                if (this.claims.hasOwnProperty("cloud_instance_host_name")) {
+                    this.cloudInstance = this.claims["cloudInstance"];
```
jasonnutter: Should these properties (`cloud_instance_host_name`, `cloudInstance`) be the same or are they intentionally different?

sameerag: No specific intention, "cloud_instance_host_name" is a fixed claim, the corresponding idToken object member name "cloudInstance" can be the same if we want to.

jasonnutter: Sorry, specifically referring to the fact you are checking for the presence of `this.claims.cloud_instance_host_name`, but then using the value of `this.claims.cloudInstance`. Should it be this?

```js
this.cloudInstance = this.claims["cloud_instance_host_name"];
```

sameerag: No, you are right. Changed it now, its a typo on my end.

sameerag:
Merging now.

PR #972, opened 2019-09-05
Adding code coverage

pkanher617:
This PR adds code coverage to the build and test reports. Eventually this will be integrated into the pipeline reporting also.

### DIFF
--- a/lib/msal-core/package.json
+++ b/lib/msal-core/package.json
@@ -25,12 +25,25 @@
   "engines": {
     "node": ">=0.8.0"
   },
+  "nyc": {
+    "extension": [
+      ".ts",
+      ".tsx"
+    ],
+    "include": [
+      "src"
+    ],
+    "exclude": [
+      "**/*.d.ts"
+    ],
+    "all": true  
+  },
   "scripts": {
     "clean": "shx rm -rf dist docs lib-commonjs lib-es6",
     "doc": "typedoc --mode modules --excludePrivate --excludeProtected --out ./docs ./src/ --gitRevision dev",
     "build:modules": "tsc && tsc -m es6 --outDir lib-es6",
     "build": "npm run clean && npm run doc && npm run build:modules && npm run lint && npm run build:webpack && npm test",
-    "test": "mocha -r test/setup.js -r ts-node/register ./test/**/*.spec.ts",
+    "test": "nyc mocha -r test/setup.js -r ts-node/register -r source-map-support/register ./test/**/*.spec.ts",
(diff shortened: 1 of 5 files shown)


lib/msal-core/package.json
```diff
@@ -25,12 +25,25 @@
...
+      ".tsx"
+    ],
+    "include": [
+      "src"
+    ],
+    "exclude": [
+      "**/*.d.ts"
+    ],
+    "all": true  
+  },
```
jasonnutter: Can we instead put this in an `.nycrc` file? I'm personally not the biggest fan of putting configs for tools inside `package.json` files.

lib/msal-core/package.json
```diff
@@ -25,12 +25,25 @@
...
   "scripts": {
     "clean": "shx rm -rf dist docs lib-commonjs lib-es6",
     "doc": "typedoc --mode modules --excludePrivate --excludeProtected --out ./docs ./src/ --gitRevision dev",
     "build:modules": "tsc && tsc -m es6 --outDir lib-es6",
     "build": "npm run clean && npm run doc && npm run build:modules && npm run lint && npm run build:webpack && npm test",
-    "test": "mocha -r test/setup.js -r ts-node/register ./test/**/*.spec.ts",
+    "test": "nyc mocha -r test/setup.js -r ts-node/register -r source-map-support/register ./test/**/*.spec.ts",
```
jasonnutter: I like having two separate test commands, one that runs coverage (which can be run in CI and does both the testing and coverage) and one that just runs the tests (for local dev), as running coverage slows down the test runner and generates more output.

e.g.
```json
"scripts": {
    "test": "mocha ...",
     "test:coverage": "nyc mocha ..."
}
```

lib/msal-core/package.json
```diff
@@ -25,12 +25,25 @@
   "engines": {
     "node": ">=0.8.0"
   },
+  "nyc": {
+    "extension": [
+      ".ts",
+      ".tsx"
```
jasonnutter: Don't think we have any `tsx` files?

lib/msal-core/package.json
pkanher617: Done.

lib/msal-core/package.json
pkanher617: Removed .tsx from the filter

lib/msal-core/package.json
pkanher617: Done.

PR #976, opened 2019-09-07
Fix bugs in cookie storage by setting cookie path

dlgrit:
This fixes two bugs with cookie storage.

The first occurs if MSAL saves its state in a directory that is not a parent of the directory the user is returned to. The cookie will not be active at this path because the cookie path defaults to the current
path.

The second bug occurs if identity provider does not return from a flow (This can occur if the user closes the tab before being redirected back), and then a new flow is started on a different path is started. Because the default path for cookie is the current paths, and cookies with different paths can have the same key, the msal.state.login cookie is written twice with the same key. Under certain conditions, the wrong cookie can be found when returning from a flow, and the state does not match.

I was unable to write any test cases for these conditions, since jsdom does not allow the URL to be changed, and duplicate cookies cannot be created when the path is `/`, which is what the testing environment is currently setup to use. This can be changed, but I was hesitant to do so without deeper understanding of the test environment.

### DIFF
--- a/lib/msal-core/src/cache/BrowserStorage.ts
+++ b/lib/msal-core/src/cache/BrowserStorage.ts
@@ -73,7 +73,7 @@ export class BrowserStorage {// Singleton
      * @param expires
      */
     setItemCookie(cName: string, cValue: string, expires?: number): void {
-        let cookieStr = cName + "=" + cValue + ";";
+        let cookieStr = cName + "=" + cValue + ";path=/;";
         if (expires) {
             const expireTime = this.getCookieExpirationTime(expires);
             cookieStr += "expires=" + expireTime + ";";


jasonnutter:
I'll test this out and let you know if I have any feedback or questions, thanks!

dlgrit:
@jasonnutter Any updates?

jasonnutter:
@dlgrit Sorry about the delay, we've been working on some other things related to storage, and haven't had a chance to revisit. Can you please address the merge conflicts and I'll have us take another look at getting this merged, thanks!

dlgrit:
@jasonnutter Updated as requested.

jasonnutter:
@dlgrit Great, thanks! I'll review this again this week.

sameerag:
@dlgrit Did you manually test this?

AndrewCraswell:
@sameerag I did some manual testing of this, and we're currently running a version of this in our test environments to good effect. The scenarios I could foresee maybe being an issue is if someone has multiple instances of MSAL running from different parts of a domain, and setting a domain-wide cookie might cause conflicts... I'm not sure if that's a valid scenario, or if MSAL is resilient against such a scenario.

jasonnutter:
We could make it configurable (e.g. `cookiePath`), and set the default to `/`

sameerag:
We can make it configurable as @jasonnutter suggested. I am merging this PR for now, will add the configurable option next week.

PR #980, opened 2019-09-11
Multiple Instances in MSAL Pt. 1 - Cache Refactor and Adding Unique Keys

pkanher617:
This PR is the first in a series to support multiple instances of MSAL in the same window stack (i.e. in main window and iframe, two separate iframes, etc.). 

This PR covers a few changes:
- Cleaning up the Storage class to make room for future improvements
- Introducing new functionality to make unique keys in the cache tied to each instance of MSAL
- Cleaning up the Constants.ts file to properly organize string constants and cache keys
- Introduction of new Storage mechanics to UserAgentApplication
- Migration from old cache signatures to new ones
- Unit tests


lib/msal-core/src/cache/MsalStorage.ts
```diff
@@ -0,0 +1,171 @@
+/*
```
sameerag: Better naming with both the base class and the derived msalStorage class.

sameerag:
Feedback:

- Better naming for AuthStorage.ts and MSALStorage.ts
- Retain writing to old cache values to facilitate rollbacks

PR #982, opened 2019-09-12
The x-client-Ver query string in the login URL hides the keep me signed in login flow

ritmos:
This query string doesn't allow to see the 'Do you want to keep the session open?' question when login in at login.microsoftonline.com if you have it enabled in your AzureAD tenant.

I guess this is not just msal.js fault but is also on microsoftonline.com endpoint side

### DIFF
--- a/lib/msal-core/src/utils/UrlUtils.ts
+++ b/lib/msal-core/src/utils/UrlUtils.ts
@@ -55,7 +55,6 @@ export class UrlUtils {
 
         str.push("client_info=1");
         str.push(`x-client-SKU=${serverRequestParams.xClientSku}`);
-        str.push(`x-client-Ver=${serverRequestParams.xClientVer}`);
         if (serverRequestParams.promptValue) {
             str.push("prompt=" + encodeURIComponent(serverRequestParams.promptValue));
         }


ritmos:
I thought the PR could help you find out the real issue in the backend.
But in any case you should consider including this fix as a _temporary solution_ because I can't imagine how many people is facing the issue believing it's a limitation instead of a bug

jasonnutter:
@ritmos We would prefer to make the correct fix once, instead of making a temporary fix without knowing if it is the right one. We'll update this issue once we make that determination. Thanks!

PraveenVerma17:
We are also facing same issue with msal version 1.1.3
Is there a way to disable x-client-ver in query string?

jasonnutter:
@PraveenVerma17 There is currently no way to remove this query string, as its included for telemetry on our side. We are investigating this internally and will update when we have determined a proper solution.

jasonnutter:
Closing, as no further action in the library (I believe this has been fixed).

ivorobioff:
the "keep me signed in" is still hidden and there's no possibility to change that version passing in the url

PR #983, opened 2019-09-13
Cache cleanup: Nonce

sameerag:
This PR fixes two issues in the transient cache (cache items needed during a request-response cycle):

- Nonce is not cleared in the cache in MSAL JS after a response for a token request is processed. This is causing nonce mismatches in some use-cases.
- We are also creating "error" related cache independent of success/failure. This PR restricts the creation of "error" cache only to failure use-cases.

### DIFF
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -1549,8 +1549,6 @@ export class UserAgentApplication {
      */
     protected saveTokenFromHash(hash: string, stateInfo: ResponseStateInfo): AuthResponse {
         this.logger.info("State status:" + stateInfo.stateMatch + "; Request type:" + stateInfo.requestType);
-        this.cacheStorage.setItem(Constants.msalError, "");
-        this.cacheStorage.setItem(Constants.msalErrorDescription, "");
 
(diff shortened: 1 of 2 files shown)


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -2257,12 +2255,7 @@ export class UserAgentApplication {
         if (loginStartPage) {
             // Cache the state, nonce, and login request data
             this.cacheStorage.setItem(Constants.loginRequest, loginStartPage, this.inCookie);
-            this.cacheStorage.setItem(Constants.loginError, "");
-
             this.cacheStorage.setItem(Constants.stateLogin, serverAuthenticationRequest.state, this.inCookie);
-
-            this.cacheStorage.setItem(Constants.msalError, "");
-            this.cacheStorage.setItem(Constants.msalErrorDescription, "");
```
DarylThayil: Do we not want to clear these?

sameerag: We are not clearing these, from what I see we are just creating empty entries even in non-error cases. I am doing some more testing for library created errors and may add it as appropriate. But these lines should go either way.

Issue #985, opened 2019-09-16
Unicode string(japanese-strting) use atob is getting error .


nakawankuma:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Other... Please describe:
</code></pre>

## Browser:
- [x] Chrome version 76.0.3809.132
- [ ] Firefox version XX
- [ ] IE version XX
- [ ] Edge version XX
- [ ] Safari version XX

## Library version

<pre><code>
Library version: 1.1.3

</code></pre>

## Current behavior

atob( "eyJ2ZXIiOiIxLjAiLCJzdWIiOiJBQUFBQUFBQUFBQUFBQUFBQUFBQUFPLTV0Yy13SlhGREt3VWFqNkc4NFJzIiwibmFtZSI6IuWNmuS_iiDkuK0iLCJwcmVmZXJyZWRfdXNlcm5hbWUiOiJuYWthQHdhbmt1bWEuY29tIiwib2lkIjoiMDAwMDAwMDAtMDAwMC0wMDAwLWVhOTctNmQyNjY5NWIxODhlIiwidGlkIjoiOTE4ODA0MGQtNmM2Ny00YzViLWIxMTItMzZhMzA0YjY2ZGFkIiwiaG9tZV9vaWQiOiIwMDAwMDAwMC0wMDAwLTAwMDAtZWE5Ny02ZDI2Njk1YjE4OGUiLCJ1aWQiOiIwMDAwMDAwMC0wMDAwLTAwMDAtZWE5Ny02ZDI2Njk1YjE4OGUiLCJ1dGlkIjoiOTE4ODA0MGQtNmM2Ny00YzViLWIxMTItMzZhMzA0YjY2ZGFkIn0");
I get exception to :
Failed to execute 'atob' on 'Window': The string to be decoded is not correctly encoded.

## Expected behavior

Non Error decode.

>certutil -decode infile outfile
It is safe output.

## Minimal reproduction of the problem with instructions

Signin Japanese DisplayName account.
myMSALObj.loginPopup(requestObj)

nakawankuma:
when other library that js-base64 I got correctry answer.
var Base64 = require('js-base64').Base64;


[...]

PR #986, opened 2019-09-16
japanese token support

nakawankuma:
It pullreq is test implementation.
https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/985

so this made incident in Microsoft ID Platform.
but I care my app.

when this impplements use then signedin (2bytes code) Japanese name account.

### DIFF
--- a/lib/msal-core/src/utils/CryptoUtils.ts
+++ b/lib/msal-core/src/utils/CryptoUtils.ts
@@ -2,7 +2,7 @@
  * Copyright (c) Microsoft Corporation. All rights reserved.
  * Licensed under the MIT License.
  */
-
+import { Base64 } from "js-base64";
 /**
  * @hidden
  */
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -6,5 +6,7 @@
     "test": "cd lib/msal-core && npm install && npm run build && cd ../msal-angular && npm install && npm run ngcompile && npm run test && cd ../msal-angularjs && npm install && npm rebuild && npm run test",
     "endToEnd": "cd lib/msal-core/tests/endToEndTests && npm install && npm run endToEnd && cd ../../../msal-angular/tests/endToEndTests && npm install &&  npm run endToEnd && cd ../../../msal-angularjs/tests/endToEndTests && npm install &&  npm run test"
   },
-  "dependencies": {}
+  "dependencies": {
+    "js-base64": "^2.5.1"
```
jasonnutter: This needs to be installed in the `lib/msal-core` folder, not here.

lib/msal-core/src/utils/CryptoUtils.ts
```diff
@@ -2,7 +2,7 @@
  * Copyright (c) Microsoft Corporation. All rights reserved.
  * Licensed under the MIT License.
  */
-
+import { Base64 } from "js-base64";
```
jasonnutter: As noted in the original Github issue, we previously removed `js-base64` due to compatibility issues, so we may need to come up with another solution for this.

jasonnutter:
Closing, as this will be addressed in #1036

PR #987, opened 2019-09-16
Move samples to top-level, improve lerna setup

jasonnutter:
Changes:

- Upgrades lerna to v3
- Moves all of the samples to a top-level `samples` folder
- When you run `npm install` from the root of the repo, lerna will run its bootstrap process so that the samples will use the local versions of the libraries, and will also build each of the libraries.

image

I will need to edit any links to these samples in the wiki or other documentation.
- https://github.com/AzureAD/microsoft-authentication-library-for-js/wiki
- https://docs.microsoft.com/en-us/azure/active-directory/develop/sample-v2-code


DarylThayil:
Would now be a good time to rename the sample folders to follow some convention so they are easy to find / identify? 
For instance instead of VanillaJsSample-b2c maybe something like msal-core-b2c-sample-app

jasonnutter:
I thought about that, but that could be done incrementally.

PR #994, opened 2019-09-18
Multiple Instances in MSAL Pt. 2: Further improvements to cache strategy

pkanher617:
This PR is to make sure we cache nonce with the state parameter to ensure that there are no inadvertent  mismatches.


pkanher617:
No longer required - work was done in PR #1010

Issue #588, opened 2019-03-07
Cannot navigate "Back" in Login Redirect


zamzowd:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[x] Feature request
[ ] Documentation issue or request
[ ] Other... Please describe:
</code></pre>

## Current behavior
When `loginRedirect` is called, the browser navigation is handled with `window.location.replace(urlNavigate)`.  This means an entry is not added to the browser "Back" history.  If a user clicks a "Sign In" button to trigger the `loginRedirect`, but then has a change of heart when faced with the login page, there is no way for the user to navigate back to the application.

## Desired behavior
Use `window.location.assign(urlNavigate)` instead of `window.location.replace(urlNavigate)`.  This results in an entry added to the browser "Back" history.  When the user gets taken to the login page, the user can use the "Back" button to return to the application.

## References
https://developer.mozilla.org/en-US/docs/Web/API/Location/replace

https://developer.mozilla.org/en-US/docs/Web/API/Location/assign

emilpalsson:
I agree. `location.replace` leads to annoying user experience.
We've done a silly workaround for this:
```
window.history.pushState({ login: true }, '')
userAgentApplication.loginRedirect(...)
```
This way, msal replaces our faked history item instead of the current one, which allows for proper back navigation.

sameerag:
Closing this as the latest msal release `msal 1.2.0-beta.2` has this fix.

PR #1002, opened 2019-09-20
Change location.replace to location.assign

pkanher617:
This PR is to fix an issue raised in #588 where history is not being pushed due to the use of window.location.replace. So far no major issues are identified with using assign instead of replace, but further investigation should be done before merging.

### DIFF
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -934,7 +934,7 @@ export class UserAgentApplication {
             const navigateWindow: Window = popupWindow ? popupWindow : window;
             const logMessage: string = popupWindow ? "Navigated Popup window to:" + urlNavigate : "Navigate to:" + urlNavigate;
             this.logger.infoPii(logMessage);
-            navigateWindow.location.replace(urlNavigate);
+            navigateWindow.location.assign(urlNavigate);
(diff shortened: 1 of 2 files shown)


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -934,7 +934,7 @@ export class UserAgentApplication {
             const navigateWindow: Window = popupWindow ? popupWindow : window;
             const logMessage: string = popupWindow ? "Navigated Popup window to:" + urlNavigate : "Navigate to:" + urlNavigate;
             this.logger.infoPii(logMessage);
-            navigateWindow.location.replace(urlNavigate);
+            navigateWindow.location.assign(urlNavigate);
```
jasonnutter: Did you consider using `window.history.pushState`? https://developer.mozilla.org/en-US/docs/Web/API/History/pushState

pkanher617: history.pushState only works on same domain since it does not reload the page, which wouldn't work here. We could probably do this instead of window.location.href in the handleAuthenticationResponse function however.

jasonnutter: Ah right of course, forgot this would be a new domain.

Issue #916, opened 2019-08-08
Improve error handling for Storage class


jasonnutter:
The `Storage` class will only work if it is instantiated in an environment (the browser) where `window.sessionStorage` or `window.localStorage` exist. Currently, `UserAgentApplication` assumes that any uncaught error from the `Storage` constructor is a result of an invalid `cacheLocation` string, however, an error could be thrown if `window` doesn't exist at all, such as when `msal` is server-side rendered. We should add a new error message to catch this scenario.

See: #914 

AC:
1. When `msal` is server-side rendered, it throws an error message communicating that it cannot be server-side rendered.

jasonnutter:
This is being addressed in MSAL.js v2, closing.

PR #1010, opened 2019-09-24
Multiple instances

sameerag:
This PR is to support multiple instances of MSAL in the same window stack (i.e. in main window and iframe, two separate iframes, etc.).

Changes:

- Cleaning up the Storage class to make room for future improvements
- Introducing new functionality to make unique keys in the cache tied to each instance of MSAL
- Cleaning up the Constants.ts file to properly organize string constants and cache keys
- Introduction of new Storage mechanics to UserAgentApplication
- Migration from old cache signatures to new ones
- Unit tests

Note: @pkanher617 's PR #980 is moved to this PR with some additional changes


lib/msal-core/src/cache/AuthCache.ts
```diff
@@ -0,0 +1,189 @@
...
+
+import { Constants, CacheKeys } from "../utils/Constants";
+import { AccessTokenCacheItem } from "./AccessTokenCacheItem";
+import { CacheLocation } from "../Configuration";
+import { BrowserStorage } from "./BrowserStorage";
+
+/**
+ * @hidden
+ */
+export class AuthCache extends BrowserStorage {// Singleton
```
jasonnutter: Is this a singleton anymore?

lib/msal-core/src/cache/AuthCache.ts
```diff
@@ -0,0 +1,189 @@
...
+    private generateCacheKey(key: string, addInstanceId: boolean): string {
+        try {
+            JSON.parse(key);
+            return key;
+        } catch (e) {
+            if (key.startsWith(`${CacheKeys.PREFIX}.${this.clientId}`)) {
+                return key;
+            } else {
+                return addInstanceId ? `${CacheKeys.PREFIX}.${this.clientId}.${key}` : `${CacheKeys.PREFIX}.${key}`;
+            }
```
jasonnutter: Nit pick: `else` not needed since you are returning.

lib/msal-core/src/cache/AuthCache.ts
```diff
@@ -0,0 +1,189 @@
...
+        if (value) {
+            this.setItem(newKey, value, storeAuthStateInCookie);
+        }
+    }
+
+    // Prepend msal.<client-id> to each key
+    private generateCacheKey(key: string, addInstanceId: boolean): string {
+        try {
+            JSON.parse(key);
+            return key;
```
jasonnutter: I'm confused, why is this just returning `key`?

lib/msal-core/src/cache/BrowserStorage.ts
```diff
@@ -0,0 +1,90 @@
...
+export class BrowserStorage {// Singleton
+
+    protected cacheLocation: CacheLocation;
+
+    constructor(cacheLocation: CacheLocation) {
+        if (!window) {
+            throw AuthError.createNoWindowObjectError("Browser storage class could not find window object");
+        }
+
+        const storageSupported = typeof window[cacheLocation] !== "undefined" && window[cacheLocation] != null;
```
jasonnutter: Can you also add a check for trutiness of `window`? Would address #916

lib/msal-core/src/cache/AuthCache.ts
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    getItemCookie(cName: string): string {
+        return super.getItemCookie(this.generateCacheKey(cName, true));
+    }
+
+    getAllAccessTokens(clientId: string, homeAccountIdentifier: string): Array<AccessTokenCacheItem> {
+        const results: Array<AccessTokenCacheItem> = [];
+        let accessTokenCacheItem: AccessTokenCacheItem;
+        const storage = window[this.cacheLocation];
```
jasonnutter: I think it would be better if `this.cacheLocation` was an actual refer to `localStorage` or `sessionStorage`, so that you don't need to have this line everywhere (and that's also where you validate if the cache location is valid).

lib/msal-core/src/cache/AuthCache.ts
```diff
@@ -0,0 +1,189 @@
...
+                        if (value) {
+                            accessTokenCacheItem = new AccessTokenCacheItem(JSON.parse(key), JSON.parse(value));
+                            results.push(accessTokenCacheItem);
+                        }
+                    }
+                }
+            }
+        }
+
+        return results;
```
jasonnutter: This could be done with `reduce`,

```js

const results = Object.keys(window[this.cacheLocation]).reduce((tokens, key) => {
    if (storage.hasOwnProperty(key) && key.match(clientId) && key.match(homeAccountIdentifier)) {
        const value = this.getItem(key);
        if (value) {
           const newAccessTokenCacheItem = new AccessTokenCacheItem(JSON.parse(key), JSON.parse(value));
           return tokens.concat([ newAccessTokenCacheItem ]);
        }
     }

     return tokens;
}, []);

return results;
```

lib/msal-core/src/cache/AuthCache.ts
```diff
@@ -0,0 +1,189 @@
...
+        return CacheKeys.ACQUIRE_TOKEN_ACCOUNT + Constants.resourceDelimiter +
+            `${accountId}` + Constants.resourceDelimiter  + `${state}`;
+    }
+
+    /**
+     * Create authorityKey to cache authority
+     * @param state
+     */
+    public static generateAuthorityKey(state: string): string {
+        return CacheKeys.AUTHORITY + Constants.resourceDelimiter + `${state}`;
```
jasonnutter: Entire thing can be a string template: 

```js
return `${CacheKeys.AUTHORITY}${Constants.resourceDelimiter}${state}`;
```

lib/msal-core/src/cache/AuthCache.ts
sameerag: I think it is, this PR is not intended to make any changes for the existing storage design, just added a layer for separation on data base functions vs library related key generation.

lib/msal-core/src/cache/AuthCache.ts
sameerag: For JSON objects as keys which implies well defined schema, we do not need any append. I added a comment and will revisit this code if it can be made simpler.

lib/msal-core/src/cache/AuthCache.ts
sameerag: +1

lib/msal-core/src/cache/AuthCache.ts
sameerag: Can we do that as a separate PR? We can target solving the type widening errors seen in specific environments too at the same time.

lib/msal-core/src/cache/AuthCache.ts
sameerag: +1

lib/msal-core/src/cache/AuthCache.ts
sameerag: +1

lib/msal-core/src/cache/AuthCache.ts
jasonnutter: Gotya. I think a comment here would be helpful to explain that.

sameerag:
Done. @pkanher617 @jasonnutter Can we please prioritize this tomorrow?

lib/msal-core/src/cache/AuthCache.ts
sameerag: Resolving, will address this as a separate PR.

lib/msal-core/src/cache/AuthCache.ts
sameerag: Done.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -467,12 +454,11 @@ export class UserAgentApplication {
                 // Push popup window handle onto stack for tracking
                 WindowUtils.trackPopup(popUpWindow);
             } catch (e) {
-                this.loginInProgress = false;
-                this.acquireTokenInProgress = false;
+                this.cacheStorage.setItem(TemporaryCacheKeys.INTERACTION_STATUS, RequestStatus.CANCELLED);
```
pkanher617: Does setting to cancelled here do anything? Doesn't this get wiped out anything during removeAcquireTokenEntries?

sameerag: Resetting interaction status since we bail out from here with a reject. removeAcquireTokenEntries() checks for this status to remove the entires for a request. So setting this to 'Cancelled' or 'Completed' is necessary.

sameerag:
@jasonnutter Added the comment. Can you please approve?

PR #1030, opened 2019-10-04
Fix SRI hashes in readme, also fix some formatting

jasonnutter:
The integrity hashes in the README were incorrect, so I fixed those (verified locally and with https://www.srihash.org/). I also put each script tag into its own code block to make them easier to copy/paste (the bottom script tags were difficult to copy/paste when they are all in the same block).

### DIFF
--- a/lib/msal-core/README.md
+++ b/lib/msal-core/README.md
@@ -45,12 +45,26 @@ The MSAL library for JavaScript enables client-side JavaScript web applications,
 ```
 
 #### Alternate region URLs
+
+To help ensure reliability, Microsoft provides a second CDN:
+
 ```html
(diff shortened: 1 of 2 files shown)


lib/msal-core/README.md
```diff
@@ -17,25 +17,56 @@ The MSAL library for JavaScript enables client-side JavaScript web applications,
...
+```
+
+Below is an example of how to use one CDN as a fallback when the other CDN is not working:
+
+```html
+<script type="text/javascript" src="https://alcdn.msauth.net/lib/1.1.3/js/msal.js"></script>
+<script type="text/javascript">
+    if(typeof Msal === 'undefined')document.write(unescape("%3Cscript src='https://alcdn.msftauth.net/lib/1.1.3/js/msal.js' type='text/javascript' %3E%3C/script%3E"));
+</script>
+```
```
jasonnutter: FYI @negoe

PR #1036, opened 2019-10-07
Fix base64 decoding to properly handle claims with unicode characters such as Japanese

jasonnutter:
Our base64 decoding was not properly handling when STS generated an id_token for a user with Japanese characters, due to unicode characters. Example id_token included as a test case.

Repro by creating a user with a name with unicode characters (such as those in the test case) and having them sign into the Vanilla JS sample.

Fixes: #985

### DIFF
--- a/lib/msal-core/src/utils/CryptoUtils.ts
+++ b/lib/msal-core/src/utils/CryptoUtils.ts
@@ -50,13 +50,13 @@ export class CryptoUtils {
             buffer[8] &= 0xbf; // buffer[8] & 10111111 will set the 6 bit to 0.
 
             return CryptoUtils.decimalToHex(buffer[0]) + CryptoUtils.decimalToHex(buffer[1])
-            + CryptoUtils.decimalToHex(buffer[2]) + CryptoUtils.decimalToHex(buffer[3])
-            + "-" + CryptoUtils.decimalToHex(buffer[4]) + CryptoUtils.decimalToHex(buffer[5])
-            + "-" + CryptoUtils.decimalToHex(buffer[6]) + CryptoUtils.decimalToHex(buffer[7])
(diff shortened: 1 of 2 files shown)


lib/msal-core/src/utils/CryptoUtils.ts
```diff
@@ -50,13 +50,13 @@ export class CryptoUtils {
             buffer[8] &= 0xbf; // buffer[8] & 10111111 will set the 6 bit to 0.
 
             return CryptoUtils.decimalToHex(buffer[0]) + CryptoUtils.decimalToHex(buffer[1])
```
jmckennon: Can this not be done with a loop or iterator?

jasonnutter: Probably, but this code is copy/pasta, and only part of this diff due to the whitespace fixes below.

jmckennon: Oh I blanked on the copy/paste, mb.

sameerag: Is there a reason why it is not done in the firstplace?

jasonnutter: Not sure, I'm guessing it was copied from somewhere else. In any case, not in scope for this PR.

valeeum:
Any idea when this fix will be deployed to master?

jasonnutter:
@valeeum @nakawankuma Please try `msal@1.2.0-beta.1` and let us know if it fixes the issue.

htthao:
@jasonnutter Please release this merged!

jasonnutter:
@htthao This is available in the beta version noted above. `1.2.0` will be released soon as a stable version, but for now you can try the beta version.

PR #1042, opened 2019-10-11
Clear temp cache on error

sameerag:
`msal js` today is not clearing the temporary cache elements created when an error is thrown. Adding a utils function to handle this usecase.

Todo:
Need to clean up old errors and request status when a new interaction request call comes through. Consider how to clean this up in concurrent use cases for `acquireTokenSilent`.


lib/msal-core/src/utils/ErrorUtils.ts
```diff
@@ -0,0 +1,17 @@
...
+ */
+
+import { AuthCache } from "../cache/AuthCache";
+import { AuthError } from "../error/AuthError";
+/**
+ * @hidden
+ */
+export class ErrorUtils {
+
+    static errorHandler(storage: AuthCache, errorFunc: AuthError): AuthError {
```
pkanher617: If we are moving the error handlers from UserAgentApplication.ts to this class, then I agree we should put this here. If this is simply a function to reset errors and the storage items, we should move this to AuthError.ts as a static function and not create a new file.

sameerag: I want to eventually move all error handling here.

lib/msal-core/README.md
```diff
@@ -17,25 +17,56 @@ The MSAL library for JavaScript enables client-side JavaScript web applications,
...
 ### Via Latest Microsoft CDN Version:
 
-    <!-- Latest compiled and minified JavaScript -->
-    <script type="text/javascript" src="https://alcdn.msauth.net/lib/1.1.3/js/msal.js"></script>
-    <script type="text/javascript" src="https://alcdn.msauth.net/lib/1.1.3/js/msal.min.js"></script>
-    
-    <!-- Alternate region URLs -->
-    <script type="text/javascript" src="https://alcdn.msftauth.net/lib/1.1.3/js/msal.js"></script>
-    <script type="text/javascript" src="https://alcdn.msftauth.net/lib/1.1.3/js/msal.min.js"></script>
+#### Latest compiled and minified JavaScript
```
jasonnutter: Pulling dev into `multiple-instances` i think would help clean up this diff.

lib/msal-core/src/cache/AuthCache.ts
```diff
@@ -29,21 +29,12 @@ export class AuthCache extends BrowserStorage {// Singleton
      * @param storeAuthStateInCookie
      */
     private migrateCacheEntries(storeAuthStateInCookie: boolean) {
+        const pCacheSet = Object.keys(PersistentCacheKeys).map(key => super.getItem(`msal.${PersistentCacheKeys[key]}`));
+        const eCacheSet = Object.keys(ErrorCacheKeys).map(key => super.getItem(`msal.${ErrorCacheKeys[key]}`));
```
jasonnutter: Nitpick: `persistentCacheSet`, `errorCacheSet`

lib/msal-core/src/cache/AuthCache.ts
```diff
@@ -124,6 +115,20 @@ export class AuthCache extends BrowserStorage {// Singleton
...
 
+    /**
+     * Reset all temporary cache items
+     */
+    resetTempCacheItems(): void {
+        const storage = window[this.cacheLocation];
+        let key: string;
+        for (key in storage) {
+            // Check if key contains msal prefix; For now, we are clearing all cache items created by MSAL.js
+            if (storage.hasOwnProperty(key) && (Object.keys(TemporaryCacheKeys).indexOf(key) > -1)) {
```
jasonnutter: I don't believe `storage.hasOwnProperty(key)` check is needed, as it will always be true by nature of doing a for/in.

lib/msal-core/src/utils/ErrorUtils.ts
jasonnutter: Why is it called `errorFunc`? It's an actual error, correct?

lib/msal-core/src/utils/ErrorUtils.ts
```diff
@@ -0,0 +1,17 @@
...
+        storage.resetTempCacheItems();
+        return errorFunc;
```
jasonnutter: An alternative design would be to throw the error, instead of returning it.

lib/msal-core/src/utils/ErrorUtils.ts
jasonnutter: And I would consider being more explicit with the name: `throwErrorAndClearTempCache`

lib/msal-core/README.md
sameerag: I did already. Not sure why this diff is showing.

lib/msal-core/src/cache/AuthCache.ts
sameerag: Done.

lib/msal-core/src/cache/AuthCache.ts
sameerag: yes. +1

lib/msal-core/src/utils/ErrorUtils.ts
sameerag: I actually did that first. There is some code in UAA that needs a return path (`openPopupWindow` related) and throwing an error here is causing compiler issues in that function. There could be a different solution, lets talk tomorrow.

lib/msal-core/src/utils/ErrorUtils.ts
sameerag: Renamed and renamed

lib/msal-core/src/utils/Constants.ts
```diff
@@ -91,19 +89,22 @@ export enum TemporaryCacheKeys {
...
+
+export enum ErrorCacheKeys {
+    LOGIN_ERROR = "login.error",
     ERROR = "error",
-    ERROR_DESC = "error.description",
-    LOGIN_ERROR = "login.error"
+    ERROR_DESC = "error.description"
 }
 
+export const INTERACTION_STATUS = "interaction.status";
```
pkanher617: Not sure why this was moved out

lib/msal-core/src/utils/ErrorUtils.ts
```diff
@@ -0,0 +1,17 @@
...
+ */
+
+import { AuthCache } from "../cache/AuthCache";
+import { AuthError } from "../error/AuthError";
+/**
+ * @hidden
+ */
+export class ErrorUtils {
+
+    static throwErrorAndClearTempCache(storage: AuthCache, error: AuthError): AuthError {
```
pkanher617: Same comment as before - this should either be a part of the AuthError class/file, or it we should have a class that does control flow handling, including functions from UserAgentApplication which handle errors and responses. Creating a utils specifically for errors doesn't need to be done in my opinion.

lib/msal-core/test/UserAgentApplication.spec.ts
```diff
@@ -150,10 +150,10 @@ describe("UserAgentApplication.ts Class", function () {
                 }
             });
         });
-        it("configure telemtry in UAA missing configuration throws config error", () => {
+        it.skip("configure telemtry in UAA missing configuration throws config error", () => {
```
pkanher617: is this function being skipped on purpose?

lib/msal-core/src/utils/Constants.ts
sameerag: I separated the existing `PersistentCacheKeys` into `PersistentCacheKeys` and `ErrorCacheKeys` (this is to separate error handing from UAA and the enable cache wipe for error messages as appropriate). Since `INTERACTION_STATUS` did not belong in both, separated it.

lib/msal-core/test/UserAgentApplication.spec.ts
sameerag: Fixed now. It was on purpose as Telemetry error is failing and now fixed it.

lib/msal-core/src/utils/ErrorUtils.ts
sameerag: Addressing this, made it a helper function in UAA.

sameerag:
Addressed all feedback; cc @pkanher617 and @jasonnutter

lib/msal-core/src/utils/ErrorUtils.ts
sameerag: Revised my stand, moved this to UAA as a helper function.

lib/msal-core/README.md
sameerag: I pulled dev into both the branches. :). I think there is a local and remote state mismatch in git on my end (I use github desktop), I can try this differently.

lib/msal-core/src/utils/Constants.ts
pkanher617: I think it should be moved back into TemporaryCacheKeys, this is just to organize the Constants file. We don't use the enums as arrays anyway, so it should be fine. I would like to prevent using one-off constants, and rather group constants into the relevant enums or objects.

lib/msal-core/src/cache/AuthCache.ts
```diff
@@ -126,6 +117,20 @@ export class AuthCache extends BrowserStorage {// Singleton
         }
     }
 
+    /**
+     * Reset all temporary cache items
+     */
+    resetTempCacheItems(): void {
```
pkanher617: I think this function is supposed to be what removeAcquireTokenEntries does. I'd prefer not to have multiple functions doing similar jobs, so let's try to reconcile those differences and come up with a better strategy for cleaning up temporary keys.

pkanher617:
pinging to close or update this PR

lib/msal-core/src/ServerRequestParameters.ts
```diff
@@ -349,23 +355,24 @@ export class ServerRequestParameters {
...
-     *
-     * Validate  scopes/state/correlationId set in the request by the user
+     * generate unique state per request
      * @param request
      */
-    private setRequestServerParams(scopes: Array<string>, state: string, correlationId: string, clientId: string) {
-        // set scope to clientId if null
-        this.scopes = scopes? [ ...scopes] : [clientId];
-
+    static setState(state: string): string {
```
DarylThayil: should this be a static method on this instance class? 
This should probably be a util somewhere. 
Also unsure setState is the right name. 

This is something like `UrlUtils.getRequestState(userProvidedState: string): string `

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -302,9 +314,12 @@ export class UserAgentApplication {
     loginRedirect(request?: AuthenticationParameters): void {
         // Throw error if callbacks are not set before redirect
         if (!this.redirectCallbacksSet) {
+            this.cacheStorage.setItem(INTERACTION_STATUS, RequestStatus.COMPLETED);
             throw ClientConfigurationError.createRedirectCallbacksNotSetError();
         }
-        this.acquireTokenInteractive(Constants.interactionTypeRedirect, true, request);
+        // append GUID to user set state  or set one for the user if null
+        const state = ServerRequestParameters.setState(request && request.state);
+        this.acquireTokenInteractive(Constants.interactionTypeRedirect, true, request,  null, null, state);
```
DarylThayil: can we clean up state on redirect cancelled?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -411,23 +442,29 @@ export class UserAgentApplication {
...
                 else {
-                    this.acquireTokenHelper(null, interactionType, isLoginCall, request, scopes, resolve, reject);
+                    this.acquireTokenHelper(null, interactionType, isLoginCall, request, scopes, resolve, reject, state);
                 }
             }
             // AcquireToken call, but no account or context given, so throw error
             else {
                 this.logger.info("User login is required");
-                throw ClientAuthError.createUserLoginRequiredError();
+                const stateOnlyResponse = buildResponseStateOnly(this.getAccountState(request && request.state));
```
DarylThayil: we use request && request.state here but not the state passed in?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -2185,4 +2227,14 @@ export class UserAgentApplication {
         };
         return new TelemetryManager(telemetryManagerConfig, telemetryEmitter);
     }
+
+    /**
+     * Helper function to clear the cache and throw an error
+     * @param storage
+     * @param error
+     */
+    private returnErrorAndClearTempCache(storage: AuthCache, error: AuthError, state: string) {
```
DarylThayil: this method feels confusing to me, I understand what it is doing but it feels like there must be a cleaner way. 

Really `storage.resetTempCacheItems(state);` is all you are doing.
I would almost say just put that line in every catch instead of this method that takes errors and storage as well.

```js
.catch(err => {
       this.storage.resetTempCacheItems(state)
       throw err; 
 });
```

I do like that you are not coping the same code and can update it in one place, it just feels like it mixes up responsiblities.

sameerag: Yes. The method initially was conceived as we are handling it at each throw. Now that we are bubbling up the errors in all use cases, guess we can do this.

lib/msal-core/src/utils/ErrorUtils.ts
sameerag: Function no longer needed. Please check the latest code.

lib/msal-core/src/UserAgentApplication.ts
sameerag: Changes done.

lib/msal-core/src/cache/AuthCache.ts
sameerag: Done, Replaced removeAcquireTokenEntries with resetTempCacheItems now in the code.

lib/msal-core/src/ServerRequestParameters.ts
sameerag: Done

lib/msal-core/src/UserAgentApplication.ts
sameerag: This fails too early for any cache state to be created.

lib/msal-core/src/utils/UrlUtils.ts
```diff
@@ -244,4 +245,27 @@ export class UrlUtils {
...
+     * @hidden
+     * validate correlationId and generate if not valid or not set by the user
+     * @param correlationId
+     */
+    static getRequestCorrelationId(correlationId: string): string {
+        // validate user set correlationId or set one for the user if null
+        if(correlationId && !CryptoUtils.isGuid(correlationId)) {
+            throw ClientConfigurationError.createInvalidCorrelationIdError();
+        }
+        return correlationId && CryptoUtils.isGuid(correlationId)? correlationId : CryptoUtils.createNewGuid();
```
jasonnutter: Don't need to truthy check `correlationId`, since you are doing that above.

lib/msal-core/src/cache/AuthCache.ts
```diff
@@ -70,7 +70,7 @@ export class AuthCache extends BrowserStorage {// Singleton
             JSON.parse(key);
             return key;
         } catch (e) {
-            if (key.indexOf(`${Constants.cachePrefix}`) === 0 || key.indexOf(PersistentCacheKeys.ADAL_ID_TOKEN) === 0) {
+            if (key.startsWith(`${Constants.cachePrefix}`) || key.startsWith(Constants.adalIdToken)) {
```
jasonnutter: This needs to be indexOf, because `startsWith` does not work in IE.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -346,12 +371,18 @@ export class UserAgentApplication {
...
 
         return new Promise<AuthResponse>((resolve, reject) => {
-            this.acquireTokenInteractive(Constants.interactionTypePopup, false, request, resolve, reject);
+            if (!request) {
+                throw ClientConfigurationError.createEmptyRequestError();
+            }
+            this.acquireTokenInteractive(Constants.interactionTypePopup, false, request, resolve, reject, state);
+        }).catch((error: AuthError) => {
+            this.cacheStorage.removeItem(TemporaryCacheKeys.INTERACTION_STATUS);
+            this.cacheStorage.resetTempCacheItems(state);
```
jasonnutter: These two lines could be factored into a helper function.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -411,23 +443,29 @@ export class UserAgentApplication {
...
             else {
                 this.logger.info("User login is required");
-                throw ClientAuthError.createUserLoginRequiredError();
+                const stateOnlyResponse = buildResponseStateOnly(this.getAccountState(state));
+                this.cacheStorage.resetTempCacheItems(state);
+                this.authErrorHandler(interactionType,
+                    ClientAuthError.createUserLoginRequiredError(),
+                    stateOnlyResponse,
+                    reject);
+                return;
```
jasonnutter: Is this `return` needed?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -2172,7 +2221,8 @@ export class UserAgentApplication {
         }
         // if set then validate
         const { applicationName, applicationVersion, telemetryEmitter } = config;
-        if (!applicationName || !applicationVersion || ! telemetryEmitter) {
+        if (!applicationName || !applicationVersion || !telemetryEmitter) {
+            console.log("Am I here");
```
jasonnutter: Guessing this can be removed?

lib/msal-core/src/utils/UrlUtils.ts
```diff
@@ -244,4 +245,27 @@ export class UrlUtils {
...
     }
+
+    /**
+     * @hidden
+     * generate unique state per request
+     * @param request
+     */
+    static getRequestState(state: string): string {
+        // append GUID to user set state  or set one for the user if null
+        return state && !StringUtils.isEmpty(state) ? CryptoUtils.createNewGuid() + "|" + state : CryptoUtils.createNewGuid();
```
jasonnutter: I think `isEmpty` also checks the truthiness of `state`?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -573,10 +609,20 @@ export class UserAgentApplication {
...
+
         return new Promise<AuthResponse>((resolve, reject) => {
+            // throw on null request
+            if (!request) {
+                throw ClientConfigurationError.createEmptyRequestError();
+            }
+
+            // return an error if called from the hidden iframe created by the msal js silent calls
+            if (UrlUtils.urlContainsHash(window.location.hash) && WindowUtils.isInIframe()) {
+                throw ClientConfigurationError.createBlockTokenRequestsInHiddenIframeError();
```
jasonnutter: This should be `reject(ClientConfigurationError.createBlockTokenRequestsInHiddenIframeError())` so that it returns a rejected promise.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -848,6 +895,8 @@ export class UserAgentApplication {
                   } else if (response) {
                       window.promiseMappedToRenewStates[expectedState][i].resolve(response);
                   } else {
+                      this.cacheStorage.removeItem(TemporaryCacheKeys.INTERACTION_STATUS);
```
pkanher617: Do we clear the cache already before calling registerCallback?

pkanher617: Never mind, we don't.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1898,12 +1947,12 @@ export class UserAgentApplication {
      * Return boolean flag to developer to help inform if login is in progress
      * @returns {boolean} true/false
      */
-    public getLoginInProgress(): boolean {
+    public getLoginInProgress(state?: string): boolean {
```
pkanher617: Why are we passing state here?

lib/msal-core/src/UserAgentApplication.ts
pkanher617: logging statement?

lib/msal-core/src/cache/AuthCache.ts
pkanher617: key.startsWith isn't supported on all browsers, which is why we used indexOf.

lib/msal-core/src/error/ClientConfigurationError.ts
```diff
@@ -169,4 +173,9 @@ export class ClientConfigurationError extends ClientAuthError {
 
         return new ClientConfigurationError(code, `${desc} mising values: ${missingKeys.join(",")}`);
     }
+
+    static createBlockTokenRequestsInHiddenIframeError(): ClientConfigurationError {
```
pkanher617: This should probably be a ClientAuthError, not a ClientConfig error.

lib/msal-core/src/utils/Constants.ts
```diff
@@ -52,17 +50,9 @@ export class Constants {
 
     static get interactionTypeRedirect(): InteractionType { return "redirectInteraction"; }
     static get interactionTypePopup(): InteractionType { return "popupInteraction"; }
+    static get IN_PROGRESS(): string { return "InProgress"; }
```
pkanher617: IN_PROGRESS -> inProgress(). Consistency in code style :)

lib/msal-core/src/utils/UrlUtils.ts
```diff
@@ -244,4 +245,27 @@ export class UrlUtils {
         const hash = UrlUtils.getHashFromUrl(urlFragment);
         return CryptoUtils.deserialize(hash);
     }
+
+    /**
+     * @hidden
+     * generate unique state per request
+     * @param request
+     */
+    static getRequestState(state: string): string {
```
pkanher617: These aren't URL specific functions, so i dont think they should go here. I know we removed this from ServerrequestParams, but I feel that is the most logical place for it. We should either make a request Utils function or find another design.

lib/msal-core/src/utils/UrlUtils.ts
```diff
@@ -244,4 +245,27 @@ export class UrlUtils {
...
+        // append GUID to user set state  or set one for the user if null
+        return state && !StringUtils.isEmpty(state) ? CryptoUtils.createNewGuid() + "|" + state : CryptoUtils.createNewGuid();
+    }
+
+    /**
+     * @hidden
+     * validate correlationId and generate if not valid or not set by the user
+     * @param correlationId
+     */
+    static getRequestCorrelationId(correlationId: string): string {
```
pkanher617: Same comment as above for getRequestState

lib/msal-core/src/utils/Constants.ts
sameerag: Done

lib/msal-core/src/utils/Constants.ts
sameerag: I thought about it, but left it alone as I did not want to change it in UAA. Guess it makes sense. Changing it.

lib/msal-core/src/UserAgentApplication.ts
sameerag: Left in during testing. Took it out.

lib/msal-core/src/cache/AuthCache.ts
sameerag: Done

sameerag: Done.

lib/msal-core/src/utils/Constants.ts
sameerag: Done.

lib/msal-core/src/UserAgentApplication.ts
sameerag: Edited this. Thanks, good catch.

lib/msal-core/src/utils/UrlUtils.ts
sameerag: Yes. Removing it.

lib/msal-core/src/UserAgentApplication.ts
sameerag: Done. Missed checking in removing this, added for tests.

lib/msal-core/src/UserAgentApplication.ts
sameerag: All the callback code in UAA, `authErrorHandler` or `authResponseHandler` are followed by `return` statements, mostly to ensure the function execution is halted I guess. I retained the pattern. We may get rid of them, but I want to test before we do that and let us take it as a separate enhancement.

lib/msal-core/src/utils/UrlUtils.ts
sameerag: Following this up with a requestValidation utility function in a new PR.

lib/msal-core/src/utils/UrlUtils.ts
sameerag: Following this up with a requestValidation utility function in a new PR.

lib/msal-core/src/utils/RequestUtils.ts
```diff
@@ -0,0 +1,145 @@
...
+            correlationId
+        };
+    }
+
+    /**
+     * @ignore
+     *
+     * blocks any login/acquireToken calls to reload from within a hidden iframe (generated for silent calls)
+     */
+    static blockReloadInHiddenIframes() {
```
pkanher617: Probably makes more sense in WindowUtils.

sameerag: Changed this.

lib/msal-core/src/error/ClientConfigurationError.ts
sameerag: Agreed and modified.

lib/msal-core/src/UserAgentApplication.ts
sameerag: This is not needed. Removed state.

lib/msal-core/src/UserAgentApplication.ts
sameerag: We do not need to clear the Interaction status as we are doing it as a part of cache clearance, hence removed it now.

lib/msal-core/src/utils/UrlUtils.ts
sameerag: Changes made.

Issue #1039, opened 2019-10-09
First login fails, second login redirects to ".com/null"


AndrewCraswell:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Other... Please describe:
</code></pre>

## Browser:
- [x] Edge version: 44.18362.387
- [ ] Chrome version XX
- [ ] Firefox version XX
- [ ] IE version XX
- [ ] Safari version XX

## Library version

<pre><code>
Library version: 1.1.3

</code></pre>

## Current behavior

Over the last month, users on Edge have been reporting that they get a sort of "double login issue". When presented with the login screen, they choose the account to authenticate with and are redirected back to the app. At this point, they're redirected back to the login page a second time, they choose their account, and are returned to the site correctly on return. However, getting a consistent repro has been problematic.

Today we finally have a user on Edge (one of our offsite vendors) who can consistently repro the issue. Unfortunately, no errors show in the console, and the user was unable to export the network trace HAR as the option was greyed out in Edge browser. They did however take screenshots:
https://i.imgur.com/TjIShD5.png

On the second redirect back to the site, it seems the user is being sent to `techskillsforbusiness.com/null` rather than the correct `redirectUri`

The application uses the following configuration for MSAL:
```typescript
const isIE = () => {
  const ua = window.navigator.userAgent;
  const msie = ua.indexOf('MSIE ') > -1;
  const msie11 = ua.indexOf('Trident/') > -1;

  return msie || msie11;
};

const configuration = {
      auth: {
        authority: 'https://login.microsoftonline.com/common/',
        clientId: process.env.REACT_AUTHENTICATION_CLIENT_ID,
        navigateToLoginRequestUrl: true,
        redirectUri: window.location.origin,
        postLogoutRedirectUri: window.location.origin,
        validateAuthority: true
      },
      cache: {
        cacheLocation: 'sessionStorage',
        storeAuthStateInCookie: isIE()
      }
    },
    parameters: {
      scopes: ['user.read', 'openid', 'profile'],
      prompt: 'select_account'
    },
    type: LoginType.Redirect
}
```

The redirect to `/null` is a high priority breaking issue for us which is blocking the launch. As an initial step, I've enabled a logger to get a verbose trace of everything that occurs and will update the bug when the user is able to test again.

[...]

Issue #347, opened 2018-07-17
Msal Intermittently returning null in redirect uri on IE/Edge/Safari


adamtay:
Hi,

I understand that this is a known issue as per the library wiki, however, Safari (MacOS X/iOS) appears to also be impacted by this issue.

**Tested browsers:**
- IE 11.1155.15063.0
- Edge 40.15063.674.0
- Safari 11.1.2 (MacOS X High Sierra 10.13.6)
- Safari (iOS 11.4.1)

**Our solution:**
- React v16.4.1 (via create-react-app)
- Msal v0.1.7
- Utilizing sessionStorage (we are unable to utilize localStorage due to our applications use case. Note that we also experience the same issue on localStorage)
- Using the `loginRedirect` method

As per https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/330#issuecomment-403889919, adding `navigateToLoginRedirectUri: false` to the `UserAgentApplication` options appears to somewhat alleviate this issue on IE/Safari but not on Edge. 

However, we are seeing mixed and intermittent results when testing between a normal browser session and private/incognito session for our tests browsers. i.e. Working on normal session Safari but not on private session.

As we are releasing a public facing website, we cannot be asking our end-users to be adding the websites domain to their 'trusted website' lists in order for msal authentication to function as intended.

If a work-around for this issue will not be ready within the upcoming months, please advise a suitable alternative for AADB2C integration. (perhaps hello.js?)

Thanks

nehaagrawal:
@adamtay We are currently working on a fix for this issue and we are planing to release it in our future release.

sayanghosh123:
@nehaagrawal Thanks for the quick response. We would need to understand a timeline for this fix please, as we are smack dab in the middle of a SDLC programme which is dependent on this. We are happy to get into Private previews if those allow us to test out the fix. An update will be appreciated.

jburman:
Also ran into this issue tonight in Firefox 61 (Windows 10).

vladkosarev:
@nehaagrawal what is the timeline for the fix? Are there any other alternatives that do work with b2c and modern browsers?

Thanks.

vladkosarev:
We converted our app to hellojs and that actually mostly works (other than popup+IE). So if you need to make sure that your app actually works with multiple browsers give hellojs a try. It's a but more verbose than MSAL but it's not too far off.

nehaagrawal:
[...]

PR #1047, opened 2019-10-14
Gracefully handle when login request url is null to prevent redirecting to "/null"

jasonnutter:
When reading the login request url from the cache, it is possible to get back `null` ([which coded explicitly](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/246ce3c761e93f56c9a5f1022895fc2222c32c5b/lib/msal-core/src/Storage.ts#L55)) or the string `"null"` (which can happen if you put `null` in the cache directly). This results in MSAL redirecting the application to `/null`, which predictably breaks most apps.

While we should work to determine the root cause of why `null` is being returned, we need to more gracefully handle when this condition occurs, since it could happen more many different reasons.

Fixes:

- #1039 
- #347


pkanher617 (approved):

sameerag (approved):
lgtm

AndrewCraswell:
@jasonnutter Any idea when this might get released? I'm being asked for daily status reports on this issue... any timeline estimate would be helpful. Not sure what the cadence is, think we're talking a week or > a month?

jasonnutter:
@AndrewCraswell Please try `msal@1.2.0-beta.1` and let us know if it fixes this issue.

PR #1053, opened 2019-10-16
Remove sandboxing from hidden iframe, due to change in eSTS redirect

jasonnutter:
STS now performs its redirects client-side in JavaScript, meaning we can no longer sandbox hidden iframes, as it will prevent the redirect from happening at all. Added code to immediately remove the iframe once we have the hash from the url, to mitigate MSAL code rerunning in the iframe (this is similar to what we do for popups, which we also can't sandbox).

Also removed ES6 JS that wasn't working in IE11 and replaced with ES5 code.

### DIFF
--- a/lib/msal-core/src/cache/AuthCache.ts
+++ b/lib/msal-core/src/cache/AuthCache.ts
@@ -69,7 +69,7 @@ export class AuthCache extends BrowserStorage {// Singleton
             JSON.parse(key);
             return key;
         } catch (e) {
-            if (key.startsWith(`${Constants.cachePrefix}`) || key.startsWith(PersistentCacheKeys.ADAL_ID_TOKEN)) {
+            if (key.indexOf(`${Constants.cachePrefix}`) === 0 || key.indexOf(PersistentCacheKeys.ADAL_ID_TOKEN) === 0) {
                 return key;
             }
             return addInstanceId ? `${Constants.cachePrefix}.${this.clientId}.${key}` : `${Constants.cachePrefix}.${key}`;
--- a/lib/msal-core/src/utils/WindowUtils.ts
+++ b/lib/msal-core/src/utils/WindowUtils.ts
@@ -126,7 +126,6 @@ export class WindowUtils {
                 ifr.style.position = "absolute";
                 ifr.style.width = ifr.style.height = "0";
                 ifr.style.border = "0";
-                ifr.setAttribute("sandbox", "allow-same-origin");
                 adalFrame = (document.getElementsByTagName("body")[0].appendChild(ifr) as HTMLIFrameElement);
             } else if (document.body && document.body.insertAdjacentHTML) {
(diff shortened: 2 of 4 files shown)


lib/msal-core/src/utils/WindowUtils.ts
```diff
@@ -140,19 +139,31 @@ export class WindowUtils {
...
+        const iframeArray: Array<HTMLIFrameElement> = Array.apply(null, Array(iframes.length)).map((iframe: HTMLIFrameElement, index: number) => iframes.item(index)); // eslint-disable-line prefer-spread
+
+        return iframeArray.filter((iframe: HTMLIFrameElement) => {
             try {
                 return iframe.contentWindow.location.hash === hash;
             } catch (e) {
                 return false;
             }
-        });
+        })[0];
```
sameerag: why the first element?

jasonnutter: `.find` returns a single element, but `.filter` returns an array, so this just gets the first (and only) iframe with the hash we're looking for.

jasonnutter: i.e. `.find` is functionality equivalent (in happy-path) to `.filter()[0]`

sameerag (approved):
lgtm

lib/msal-core/src/cache/AuthCache.ts
```diff
@@ -69,7 +69,7 @@ export class AuthCache extends BrowserStorage {// Singleton
             JSON.parse(key);
             return key;
         } catch (e) {
-            if (key.startsWith(`${Constants.cachePrefix}`) || key.startsWith(PersistentCacheKeys.ADAL_ID_TOKEN)) {
+            if (key.indexOf(`${Constants.cachePrefix}`) === 0 || key.indexOf(PersistentCacheKeys.ADAL_ID_TOKEN) === 0) {
```
DarylThayil: does that string need to be interpolated? 
could it just be key.indexOf(variable)

lib/msal-core/src/utils/WindowUtils.ts
```diff
@@ -174,13 +185,13 @@ export class WindowUtils {
      * @ignore
      */
     static getPopUpWithHash(hash: string): Window {
-        return WindowUtils.getPopups().find(popup => {
+        return WindowUtils.getPopups().filter(popup => {
```
DarylThayil: does this always return an array? Could it be null?

DarylThayil (approved):
added some small questions but lgtm

lib/msal-core/src/cache/AuthCache.ts
jasonnutter: Ah, no it doesn't, technically.

lib/msal-core/src/utils/WindowUtils.ts
jasonnutter: Get popups will always return an array, yes.

jasonnutter: As will filter.

PR #1061, opened 2019-10-20
Refactor react sample to use functional components and hooks 

james-innes:
App and AuthProvider defined as arrow functions with use of hooks for state and lifecycle management. 
AuthProvider still defined as HOC as couldn't see a better way of implementing it as a custom hook.
I would value any feedback as to why this approach shouldn't be adopted.

### DIFF
--- a/samples/react-sample-app/src/App.js
+++ b/samples/react-sample-app/src/App.js
@@ -6,64 +6,58 @@ import "./App.css";
 
 const Json = ({ data }) => <pre>{JSON.stringify(data, null, 4)}</pre>;
 
-class App extends React.Component {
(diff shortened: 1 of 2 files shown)


isnifer:
> AuthProvider still defined as HOC as couldn't see a better way of implementing it as a custom hook.
I did it in #1206

asos-jamesinnes:
@isnifer nice work, this PR has been left stale a while and with merge conflicts now

james-innes:
This was a neat refactor at the time to adopt a newer approach with some nice syntax but the pull request has been stale and I don't think worth my time resolving conflicts so I'll close this one off  :)

PR #1064, opened 2019-10-21
Create eslint-config-msal package for common lint rules

jasonnutter:
Sets the `msal` package to use a new shared `@azuread/eslint-config-msal` package, which is hosted on Github Packages.

https://github.com/AzureAD/microsoft-authentication-library-for-js/packages/59316

Published under the `@AzureAD` scope because all Github Packages need to be scoped (to the organization or personal account that they are published under).

New development-related step is that you need to generate and login to the npm CLI with a Github access token (which has the `repo` and `read:packages` scopes). When the repo is bootstrapped via lerna, this isn't actually needed, as lerna will use the local copy instead, but everyone should still go through the steps to generate one. I've also tested that since this is a public repo, users can download packages from GH packages without being a member of the `@AzureAD` organization. 

I recommend that we also use GH packages for the shared configurations that we want to make for Babel and nyc, along with any other tools that we need.


pkanher617:
Does this PR need to be updated?

jasonnutter:
Yeah, I'll close/update this PR soon.

pkanher617 (approved):
Reviewed together - looks fine to me, would like to improve lerna so we can have some faster bootstrapping for dev

lib/msal-core/.npmrc
```diff
@@ -0,0 +1 @@
+@azuread:registry=https://npm.pkg.github.com/AzureAD
```
jasonnutter: This tells npm to get `@azuread` scoped packages from the Github registry, instead of the main npm registry.

tools/eslint-config-msal/.npmrc
```diff
@@ -0,0 +1 @@
+registry=https://npm.pkg.github.com/AzureAD
```
jasonnutter: Tells npm to publish this package to Github packages, instead of the main npm registry.

jmckennon: Does this not need the `@azuread` in front of it like in the main `./npmrc`?

jmckennon (approved):
Also an awesome change, looks good.

contributing.md
```diff
@@ -47,6 +47,17 @@ and start hacking:
 $ git checkout -b my-feature-branch
 ```
 
+### GITHUB PACKAGES
+
+We use the Github Package registry to host some of our internal packages for tooling (e.g. linting, testing, etc.). In order to download packages from the Github registry, you need to generate an access token for Github and use it to login to the npm CLI. If you bootstrap the repository via Lerna (as described in the proceeding section), you should not receive any errors related to this, however, we still recommend that you generate an access token and use it with `npm login`:
+
+1. Go to https://github.com/settings/tokens
+2. Generate a new access token with the `repo` and `read:packages` scopes.
+3. Run `npm login --registry=https://npm.pkg.github.com`
```
DarylThayil: does it proxy through git to npm ?

DarylThayil: are these public pkg ?  If so why do you need to login ?

tools/eslint-config-msal/.npmrc
jasonnutter: I'll can check if setting the scope will also publish it correctly.

contributing.md
jasonnutter: The login command will just set the correct access token for the given registry. You still need to tell npm to use that registry as needed, which the `.npmrc` file in the repo will do for packages that are under the `@azuread` scope (i.e. setting an access token for a registry is different from using it).

And yes, I believe the Github registry will proxy to the npm registry for packages it doesn't have.

jasonnutter: Yes, these packages are public, but Github still requires that everyone login with their Github account (probably to help prevent DDOS and other abuse).

lib/msal-core/build/npm-build-steps.yml
```diff
@@ -7,11 +7,12 @@ steps:
   displayName: 'Install Node.js'
 
 - task: Npm@1
+  env:
+    NPM_CONFIG_TOKEN: $(NPM_CONFIG_TOKEN)
```
DarylThayil: Whose token is this ?

DarylThayil:
What value does using github registry over npm give us?

lib/msal-core/build/npm-build-steps.yml
jasonnutter: I generated one from my account to get it working.

jasonnutter:
@DarylThayil Easier to publish and manage, and doesn't pollute our npm account with packages that are just for our team to consume.

DarylThayil:
My counterpoint would be that publishing in npm is not hard to do or manage, and that adding a second publishing repository makes it harder.  I also don't think that these packages are polluting, if anything its a good opportunity for us to think about use outside our team. If we build a project people believe is a good template to use then the tools we use could help them build great projects too.  Eslint Airbnb being a good example. 

I am sure github is a fine repository, I am just asking are the benefits to 2 tools really worth the maintenance of two tools.

lib/msal-core/build/npm-build-steps.yml
DarylThayil: can we get an organizational one to use so that its not tied to an individual ?

jasonnutter:
@DarylThayil Publishing under the `@azure` scope on npm (which we should be doing for all of our packages, arguably) has been hard to manage, given the restrictions that are placed on that scope. 

Also, its not really that much more work, just need to login to the CLI once and setup the `npmrc` files correctly and you're good.

lib/msal-core/build/npm-build-steps.yml
jasonnutter: We could, but that would make managing it harder (for essentially no security benefit, given its currently just for reading packages).

DarylThayil: I think its worth looking into just to make sure we don;t have things break if an individual member moves teams or something

DarylThayil:
I think I still feel it may be extraneous, but I am willing to try it out and see what we learn.
LGTM

DarylThayil (approved):

sameerag:
Can we please merge this after the beta-5 release?

jasonnutter:
Yeah, it can wait until then.

jasonnutter:
I will reopen this PR given all teh changes that have been made since.

Issue #1065, opened 2019-10-21
MSAL throws an error whenever logging in to an Azure B2C enabled application with Facebook credentials


tvperez76:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X ] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Other... Please describe:
</code></pre>

## Browser:
- [X] Chrome version 77.0.3865.120 
- [ ] Firefox version XX
- [ ] IE version XX
- [ ] Edge version XX
- [ ] Safari version XX

## Library version

<pre><code>
Library version: 1.2.0-beta.1
</code></pre>

## Current behavior
Whenever I try to log on to my Azure B2C enabled application with Facebook credentials, the following error is thrown and the 
```console
Error while getting a token! SyntaxError: Unexpected token m in JSON at position 0
    at JSON.parse (<anonymous>)
    at AuthCache.js:143
    at Array.reduce (<anonymous>)
    at e.getAllAccessTokens (AuthCache.js:139)
    at t.getCachedToken (UserAgentApplication.js:929)
    at UserAgentApplication.js:464
    at new j (zone-evergreen.js.pre-build-optimizer.js:876)
    at t.acquireTokenSilent (UserAgentApplication.js:433)
    at d.project (authentication.service.ts:130)
    at d._next (switchMap.js.pre-build-optimizer.js:30)
```
![Error as found in Chrome](https://mlpexpress2.blob.core.windows.net/errors/Error.png)
## Expected behavior
To log in without an error.

## Minimal reproduction of the problem with instructions
Log in to [MLP Express](https://mlp-express.com/) with a Facebook account.

sameerag:
@tvperez76 This is a bug we identified yesterday in the beta, especially for concurrent calls. We are working on a fix and I will update the thread as we have the PR in dev. Please watch out for a new beta release soon. Thanks for raising this.

sameerag:
`msal.1.2.0-beta.2` fixes this. We are working on the announcement and detailed release notes to follow soon, but the package is available in npm now for anyone to test try it.

PR #1067, opened 2019-10-22
Add "scopes" to check access token keys

sameerag:
Fix for #1065

### DIFF
--- a/lib/msal-core/src/cache/AuthCache.ts
+++ b/lib/msal-core/src/cache/AuthCache.ts
@@ -154,7 +154,8 @@ export class AuthCache extends BrowserStorage {// Singleton
      */
     getAllAccessTokens(clientId: string, homeAccountIdentifier: string): Array<AccessTokenCacheItem> {
         const results = Object.keys(window[this.cacheLocation]).reduce((tokens, key) => {
-            if ( window[this.cacheLocation].hasOwnProperty(key) && key.match(clientId) && key.match(homeAccountIdentifier) && key.match(Constants.scopes)) {
+            const keyMatches = key.match(clientId) && key.match(homeAccountIdentifier) && key.match(Constants.scopes);
+            if ( keyMatches ) {
                 const value = this.getItem(key);
                 if (value) {
                     try {
@@ -162,7 +163,7 @@ export class AuthCache extends BrowserStorage {// Singleton
                         const newAccessTokenCacheItem = new AccessTokenCacheItem(parseAtKey, JSON.parse(value));
                         return tokens.concat([ newAccessTokenCacheItem ]);
                     } catch (e) {
-                        throw ClientAuthError.createCacheParseError();
+                        throw ClientAuthError.createCacheParseError(key);
                     }
                 }
             }
--- a/lib/msal-core/src/cache/AuthCache.ts
+++ b/lib/msal-core/src/cache/AuthCache.ts
@@ -7,6 +7,7 @@ import { Constants, PersistentCacheKeys, TemporaryCacheKeys, RequestStatus } fro
 import { AccessTokenCacheItem } from "./AccessTokenCacheItem";
 import { CacheLocation } from "../Configuration";
 import { BrowserStorage } from "./BrowserStorage";
+import { ClientAuthError } from "../error/ClientAuthError";
 
 /**
  * @hidden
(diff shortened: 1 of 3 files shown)


lib/msal-core/src/cache/AuthCache.ts
```diff
@@ -153,11 +154,16 @@ export class AuthCache extends BrowserStorage {// Singleton
      */
     getAllAccessTokens(clientId: string, homeAccountIdentifier: string): Array<AccessTokenCacheItem> {
         const results = Object.keys(window[this.cacheLocation]).reduce((tokens, key) => {
-            if ( window[this.cacheLocation].hasOwnProperty(key) && key.match(clientId) && key.match(homeAccountIdentifier)) {
+            if ( window[this.cacheLocation].hasOwnProperty(key) && key.match(clientId) && key.match(homeAccountIdentifier) && key.match(Constants.scopes)) {
```
pkanher617: It may be better to put the if condition into a variable or function for readability.

lib/msal-core/src/error/ClientAuthError.ts
```diff
@@ -88,6 +88,10 @@ export const ClientAuthErrorMessage = {
     invalidInteractionType: {
         code: "invalid_interaction_type",
         desc: "The interaction type passed to the handler was incorrect or unknown"
+    },
+    cacheParseError: {
+        code: "cannot_parse_cache",
+        desc: "The cached token key is not a JSON, cannot be parsed"
```
jasonnutter: Typo, "The cached token key is not valid JSON and connect be parsed"

lib/msal-core/src/cache/AuthCache.ts
```diff
@@ -153,11 +154,16 @@ export class AuthCache extends BrowserStorage {// Singleton
...
                 const value = this.getItem(key);
                 if (value) {
-                    const newAccessTokenCacheItem = new AccessTokenCacheItem(JSON.parse(key), JSON.parse(value));
-                    return tokens.concat([ newAccessTokenCacheItem ]);
+                    try {
+                        const parseAtKey = JSON.parse(key);
+                        const newAccessTokenCacheItem = new AccessTokenCacheItem(parseAtKey, JSON.parse(value));
+                        return tokens.concat([ newAccessTokenCacheItem ]);
+                    } catch (e) {
+                        throw ClientAuthError.createCacheParseError();
```
pkanher617: We shouldn't swallow the thown error, but pass it in the error we create. Otherwise we lose context of why the error is occurring. Just pass the error object into the createCacheParseError(e) and print it in the error message string.

lib/msal-core/src/cache/AuthCache.ts
jasonnutter: I realize this was there before, but `window[this.cacheLocation].hasOwnProperty(key)` isn't needed as it will always be true given the nature of the `reduce`.

lib/msal-core/src/cache/AuthCache.ts
jasonnutter: Agree, would be helpful to at least include the key that was invalid.

lib/msal-core/src/cache/AuthCache.ts
pkanher617: +1

lib/msal-core/src/error/ClientAuthError.ts
sameerag: Done.

lib/msal-core/src/cache/AuthCache.ts
sameerag: +1

lib/msal-core/src/cache/AuthCache.ts
sameerag: +1 done

sameerag: done

lib/msal-core/src/cache/AuthCache.ts
```diff
@@ -153,11 +154,17 @@ export class AuthCache extends BrowserStorage {// Singleton
...
                 const value = this.getItem(key);
                 if (value) {
-                    const newAccessTokenCacheItem = new AccessTokenCacheItem(JSON.parse(key), JSON.parse(value));
-                    return tokens.concat([ newAccessTokenCacheItem ]);
+                    try {
+                        const parseAtKey = JSON.parse(key);
+                        const newAccessTokenCacheItem = new AccessTokenCacheItem(parseAtKey, JSON.parse(value));
+                        return tokens.concat([ newAccessTokenCacheItem ]);
+                    } catch (e) {
+                        throw ClientAuthError.createCacheParseError(key);
```
pkanher617: I would prefer we just pass the whole error also instead of the just the key since we are doing multiple operations within the 'try' block.

pkanher617: see the idtokenparsing error in the same ClientAuthError.ts file.

PR #1071, opened 2019-10-24
Add MSER Telemetry Data

sameerag:
Payloads from MSAL JS to STS as `queryParameters`


sameerag:
This is being implemented in a different branch for `msal-common` and cannot be supported for `msal-core` for now due to server limitations. Closing the PR.

PR #1083, opened 2019-10-29
Correlation Id - pass the client set correlation-id to service

sameerag:
`msal js` needs to pass the user set correlation-id per request if the application sets it. Currently we are only generating the random guid and ignoring any app passed values.

Todo:
Tests pending

### DIFF
--- a/lib/msal-core/src/ServerRequestParameters.ts
+++ b/lib/msal-core/src/ServerRequestParameters.ts
@@ -65,7 +65,6 @@ export class ServerRequestParameters {
         this.nonce = CryptoUtils.createNewGuid();
         this.state = state && !StringUtils.isEmpty(state) ?  CryptoUtils.createNewGuid() + "|" + state   : CryptoUtils.createNewGuid();
 
-        // TODO: Change this to user passed vs generated with the new PR
         this.correlationId = correlationId || CryptoUtils.createNewGuid();
 
         // telemetry information
--- a/lib/msal-core/src/ServerRequestParameters.ts
+++ b/lib/msal-core/src/ServerRequestParameters.ts
@@ -11,6 +11,7 @@ import { Account } from "./Account";
 import { SSOTypes, Constants, PromptState, BlacklistedEQParams, libraryVersion } from "./utils/Constants";
 import { ClientConfigurationError } from "./error/ClientConfigurationError";
 import { StringUtils } from "./utils/StringUtils";
+import { UserSetRequestParams } from "./UserAgentApplication";
 
 /**
  * Nonce: OIDC Nonce definition: https://openid.net/specs/openid-connect-core-1_0.html#IDToken
@@ -53,19 +54,15 @@ export class ServerRequestParameters {
      * @param redirectUri
      * @param state
      */
-    constructor (authority: Authority, clientId: string, scope: Array<string>, responseType: string, redirectUri: string, state: string, correlationId?: string) {
+    constructor (authority: Authority, clientId: string, responseType: string, redirectUri: string, reqParams: UserSetRequestParams) {
         this.authorityInstance = authority;
         this.clientId = clientId;
-        if (!scope) {
-            this.scopes = [clientId];
-        } else {
-            this.scopes = [ ...scope ];
-        }
-
         this.nonce = CryptoUtils.createNewGuid();
-        this.state = state && !StringUtils.isEmpty(state) ?  CryptoUtils.createNewGuid() + "|" + state   : CryptoUtils.createNewGuid();
 
-        this.correlationId = correlationId || CryptoUtils.createNewGuid();
+        // validate and populate the below parameters from reqParams: UserSetRequestParams
+        this.scopes = reqParams.scopes;
+        this.state  = reqParams.state;
+        this.correlationId = reqParams.correlationId;
 
         // telemetry information
         this.xClientSku = "MSAL.JS";
@@ -352,6 +349,35 @@ export class ServerRequestParameters {
         }
 
         return paramsString;
+    }
+
+    /**
+     * @hidden
+     *
+     * Validate  scopes/state/correlationId set in the request by the user
+     * @param request
+     */
+    static validateUserSetServerParams(scopes: Array<string>, state: string, correlationId: string, clientId: string): UserSetRequestParams {
+
+        // set scope to clientId if null
+        let reqScopes: Array<string>;
+        if (!scopes) {
+            reqScopes = [clientId];
+        } else {
+            reqScopes = [ ...scopes ];
+        }
+
+        // append GUID to user set state  or set one for the user if null
+        const reqState = state && !StringUtils.isEmpty(state) ? CryptoUtils.createNewGuid() + "|" + state : CryptoUtils.createNewGuid();
+
+        // validate user set correlationId or set one for the user if null
+        if(correlationId && !CryptoUtils.isGuid(correlationId)) {
+            throw ClientConfigurationError.createInvalidCorrelationIdError();
+        }
+        const reqCcorrelationId = correlationId && CryptoUtils.isGuid(correlationId)? correlationId : CryptoUtils.createNewGuid();
+
+        const reqParams = {scopes: reqScopes, state: reqState, correlationId: reqCcorrelationId};
+        return reqParams;
     }
 
     // #endregion
--- a/lib/msal-core/src/ServerRequestParameters.ts
+++ b/lib/msal-core/src/ServerRequestParameters.ts
@@ -11,7 +11,6 @@ import { Account } from "./Account";
 import { SSOTypes, Constants, PromptState, BlacklistedEQParams, libraryVersion } from "./utils/Constants";
 import { ClientConfigurationError } from "./error/ClientConfigurationError";
 import { StringUtils } from "./utils/StringUtils";
-import { UserSetRequestParams } from "./UserAgentApplication";
 
 /**
  * Nonce: OIDC Nonce definition: https://openid.net/specs/openid-connect-core-1_0.html#IDToken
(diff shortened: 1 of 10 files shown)


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1805,7 +1807,7 @@ export class UserAgentApplication {
      * @param state
      * @return {@link AuthResponse} AuthResponse
      */
-    protected getCachedTokenInternal(scopes : Array<string> , account: Account, state: string): AuthResponse {
+    protected getCachedTokenInternal(scopes : Array<string> , account: Account, state: string, correlationId?: string): AuthResponse {
```
DarylThayil: Do we use the correlation id when looking up the cached token?

lib/msal-core/src/ServerRequestParameters.ts
```diff
@@ -65,8 +65,7 @@ export class ServerRequestParameters {
         this.nonce = CryptoUtils.createNewGuid();
         this.state = state && !StringUtils.isEmpty(state) ?  CryptoUtils.createNewGuid() + "|" + state   : CryptoUtils.createNewGuid();
 
-        // TODO: Change this to user passed vs generated with the new PR
-        this.correlationId = CryptoUtils.createNewGuid();
+        this.correlationId = correlationId || CryptoUtils.createNewGuid();
```
jasonnutter: What happens if you pass a string that isn't a guid? I think we should validate that the provided string is a valid guid and throw an error.

pkanher617: +1

jasonnutter: e.g. add an `isValidRegex` function to the CryptoUtils file that performs a check with this regex: https://stackoverflow.com/questions/7905929/how-to-test-valid-uuid-guid

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -490,7 +490,8 @@ export class UserAgentApplication {
                 scopes,
                 responseType,
                 this.getRedirectUri(),
-                request && request.state
+                request && request.state,
```
pkanher617: We are passing three different parameters from the request object into ServerRequestParameters - scopes, state and now correlationId. May be better to pass the request to ServerRequestParameters as a whole object.

lib/msal-core/src/ServerRequestParameters.ts
jasonnutter: And then throws a new type of ClientAuth error if it fails.

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: No, as tokens can be used for multiple requests.

pkanher617: I believe it is used in creating the serverRequestParameters, which is constructed in this function. However, this particular function (getCachedTokenInternal) is only used by the Angular wrapper, so it may not be necessary.

jasonnutter:
One further idea: expose an API to generate a valid correlation id, which helps prevent users from generating invalid ones.

lib/msal-core/src/UserAgentApplication.ts
DarylThayil: So should we not pass it in here?

lib/msal-core/src/ServerRequestParameters.ts
DarylThayil: Do we care if it is a guid? Does ests?

DarylThayil: what if the correlation id is 
134u9324324243 and thats how they correlate?

DarylThayil:
can we add tests for this ?

lib/msal-core/src/ServerRequestParameters.ts
sameerag: I can check with STS if the requirement for correlation-id is a GUID and add this check.

lib/msal-core/src/UserAgentApplication.ts
sameerag: Yes. Agreed.

lib/msal-core/src/ServerRequestParameters.ts
sameerag: So confirmed from STS that they look for a GUID when they parse for correlation-id. So I will go with @jasonnutter' s suggestion to add a check for the same for the user input.

lib/msal-core/src/UserAgentApplication.ts
sameerag: I kept this intact for now to keep the interface the same, ideally I think we should rewrite this function and `getCachedToken` to not have dependency on the `ServerRequestParameters`, as they do not need a NW call. But it is too much code change that is irrelevant to the current problem we are solving.

sameerag: cc @jasonnutter, please let me know if we co-ordinate on refactoring this piece of code in `msal-core` as you start working on angular.

lib/msal-core/src/ServerRequestParameters.ts
sameerag: Changes made. Please review.

lib/msal-core/src/UserAgentApplication.ts
sameerag: Added a new type. Please review.

lib/msal-core/src/utils/CryptoUtils.ts
```diff
@@ -83,6 +83,15 @@ export class CryptoUtils {
         }
     }
 
+    /**
+     * verifies if a string is  GUID
+     * @param guid
+     */
+    static isGuid(guid: string) {
```
jasonnutter: Are there tests for this?

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: For 1.0 of Angular, I'm going to leave this code (and the other Angular-specific code). We'll remove it for msal-core v2.

lib/msal-core/src/ServerRequestParameters.ts
```diff
@@ -355,6 +351,35 @@ export class ServerRequestParameters {
         return paramsString;
     }
 
+    /**
+     * @hidden
+     *
+     * Validate  scopes/state/correlationId set in the request by the user
+     * @param request
+     */
+    static validateUserSetServerParams(scopes: Array<string>, state: string, correlationId: string, clientId: string): UserSetRequestParams {
```
jasonnutter: A new type feels a bit heavy-handed here. An alternative would be to have these parameters passed in as before, and then call this function from the constructor to validate?

lib/msal-core/src/ServerRequestParameters.ts
```diff
@@ -11,6 +11,7 @@ import { Account } from "./Account";
 import { SSOTypes, Constants, PromptState, BlacklistedEQParams, libraryVersion } from "./utils/Constants";
 import { ClientConfigurationError } from "./error/ClientConfigurationError";
 import { StringUtils } from "./utils/StringUtils";
+import { UserSetRequestParams } from "./UserAgentApplication";
```
jasonnutter: If we stick with this type, I think it should be defined in this file, not UAA (as we want fewer things in UAA).

lib/msal-core/src/ServerRequestParameters.ts
```diff
@@ -355,6 +351,35 @@ export class ServerRequestParameters {
...
+     * @param request
+     */
+    static validateUserSetServerParams(scopes: Array<string>, state: string, correlationId: string, clientId: string): UserSetRequestParams {
+
+        // set scope to clientId if null
+        let reqScopes: Array<string>;
+        if (!scopes) {
+            reqScopes = [clientId];
+        } else {
+            reqScopes = [ ...scopes ];
```
jasonnutter: This can be a ternary instead (I know its copy pasta, but its chance to make it better :) )

lib/msal-core/src/utils/CryptoUtils.ts
sameerag: Yes. In ServerRequestParameters.spec.ts indirectly. I can add one for cryptoUtils.

lib/msal-core/src/ServerRequestParameters.ts
sameerag: It is getting complicated with the dependencies with `getCachedTokenInternal`. I implemented that and rewrote the whole code.

lib/msal-core/src/ServerRequestParameters.ts
sameerag: Yes :)

lib/msal-core/src/ServerRequestParameters.ts
sameerag: Since we are passing it in `getCachedTokenInternal`. If you are going to fix it in angular 6/7/8 I can move the whole thing to `ServerRequestParameters`. There is no `request` parameter in   `getCachedTokenInternal` and it gets complicated. Only other option is to keep sending the old params as is and I find that not ideal either.

lib/msal-core/src/UserAgentApplication.ts
DarylThayil: ok resolving

lib/msal-core/src/UserAgentApplication.ts
sameerag: Changed it. Since `getTokenCachedInternal` is not being changed in Angular version for the immediate future, cannot pass request to the `ServerRequestParameters`.

lib/msal-core/src/ServerRequestParameters.ts
sameerag: Done.

lib/msal-core/src/utils/CryptoUtils.ts
sameerag: Added tests for CryptoUtils.ts (I added all tests, not just for this API since it is not much of an effort)

lib/msal-core/src/ServerRequestParameters.ts
```diff
@@ -355,6 +351,29 @@ export class ServerRequestParameters {
         return paramsString;
     }
 
+    /**
+     * @hidden
+     *
+     * Validate  scopes/state/correlationId set in the request by the user
+     * @param request
+     */
+    private validateUserSetServerParams(scopes: Array<string>, state: string, correlationId: string, clientId: string) {
```
pkanher617: If the function is called validate, I wouldn't expect it to create a new object (see validateScopes). I would create the objects and pass it in to validate, or rename the function. You can call it createValidServerParams or something and then set the field variables (this.state, this.correlationid, etc.) directly there instead of returning an object from the function.

pkanher617: it should probably be setValidServerParams instead of create"".

sameerag: I agree. I missed that these are this.* params and can be set in a different function. Changes made.

PR #1089, opened 2019-10-30
Remove duplicate readme.md section

karaggeorge:
The example and text below it was duplicated in this section


pkanher617:
@karaggeorge Thanks for the submission!

PR #1090, opened 2019-10-30
Add response_mode=fragment to silent renewal call, fixes #1088

### LINKED ISSUE #1088: acquireTokenSilent times out when user interaction is required
This is a re-opening of issue #699, which is closed, but I'm still facing the issue. The original issue does a great job of explaining the problem, but there was a lot of discussion which I'll try to consolidate here.

## I'm submitting a...

<pre><code>
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Other... Please describe:
</code></pre>

## Browser:
- [x] Chrome version 78.0.3904.70
- [x] Firefox version 70.0
- [ ] IE version XX
- [x] Edge version 44.18362.387.0 and 79.0.309.5 (beta)
- [ ] Safari version XX

## Library version
<pre><code>Library version: 1.1.3 and 1.2.0-beta.3</code></pre>

## Current behavior
When user interaction is required, `acquireTokenSilent()` hangs for a few seconds and then throws the following error:
```
ClientAuthError: Token renewal operation failed due to timeout.
```

## Expected behavior
`acquireTokenSilent()` should reject the promise more quickly and throw an `InteractionRequiredAuthError` with one of the specific error codes (`interaction_required`, `consent_required`, or `login_required`).

Host a page that runs the following JS (fill in YOUR_CLIENT_ID, YOUR_REDIRECT_URI, and YOUR_UPN):
```js
// Create MSAL object
var msalConfig = {
    auth: {
        clientId: "<YOUR_CLIENT_ID>",
        redirectUri: "<YOUR_REDIRECT_URI>"
    },
    cache: {
        storeAuthStateInCookie: true
    }
};

var myMSALObj = new Msal.UserAgentApplication(msalConfig);

// Config for acquiring token silently
var tokenRequestObj = {
    scopes: ["user.read"],
    loginHint: '<YOUR_UPN>',
    // extraQueryParameters: {
    //     response_mode: 'fragment'
    // }
};

// Try to acquire token silently
myMSALObj.acquireTokenSilent(tokenRequestObj).then(authTokenResult => {
    console.info('successfully acquired token silently');
    console.info(authTokenResult);
}).catch(function (error) {
    console.error(error);
});
```

## Additional details
The cause of the issue seems to be that AAD passes the error to the redirect URI as query parameters, but MSAL is looking for fragments.

The [AAD doc for implicit flow](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-implicit-grant-flow#send-the-sign-in-request) states that `response_mode` "defaults to query for an access token, but **fragment if the request includes an id_token**." This is true if the request succeeds - `id_token`, `expires_in`, etc. are passed as fragments. But if it fails, then the error parameters are passed as query parameters. You can see this by opening the following link in a private browser session: [link with response_type=id_token, prompt=none](https://login.microsoftonline.com/common/oauth2/v2.0/authorize?client_id=6731de76-14a6-49ae-97bc-6eba6914391e&response_type=id_token&redirect_uri=http%3A%2F%2Flocalhost%2Fmyapp%2F&scope=openid&state=12345&nonce=678910&prompt=none)
In this case, the error is sent as a query parameter:
```
http://localhost/myapp/?error=interaction_required&error_description=...&state=...
```

If you explicitly set `response_mode=fragment` (see commented section in repro code above), then the error **does** come through in the fragment. You can see this by opening the following link in a private browser session: [link with response_type=id_token, prompt=none, response_mode=fragment](https://login.microsoftonline.com/common/oauth2/v2.0/authorize?client_id=6731de76-14a6-49ae-97bc-6eba6914391e&response_type=id_token&redirect_uri=http%3A%2F%2Flocalhost%2Fmyapp%2F&scope=openid&state=12345&nonce=678910&prompt=none&response_mode=fragment)
Now the error is sent as a fragment and MSAL would pick it up properly:
```
http://localhost/myapp/#error=interaction_required&error_description=...&state=...
```

By default, MSAL **does not** pass the `response_mode` parameter. So the error comes through as a query parameter, MSAL misses it, and the call times out.

It seems like there could be 3 solutions:
- AAD could return errors in the fragment when no `response_mode` is specified
- MSAL could explicitly set `response_mode=fragment` by default
- MSAL could listen for errors in the query string parameters

jasonnutter:
Ensures that when there is an error in the silent call, that the error is returned as a fragment, instead of a query parameter, fixes #1088.

### DIFF
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -1253,7 +1253,7 @@ export class UserAgentApplication {
         this.logger.verbose("Renew token Expected state: " + serverAuthenticationRequest.state);
 
         // Build urlNavigate with "prompt=none" and navigate to URL in hidden iFrame
-        const urlNavigate = UrlUtils.urlRemoveQueryStringParameter(UrlUtils.createNavigateUrl(serverAuthenticationRequest), Constants.prompt) + Constants.prompt_none;
+        const urlNavigate = UrlUtils.urlRemoveQueryStringParameter(UrlUtils.createNavigateUrl(serverAuthenticationRequest), Constants.prompt) + Constants.prompt_none + Constants.response_mode_fragment;
 
         window.renewStates.push(serverAuthenticationRequest.state);
         window.requestType = Constants.renewToken;
@@ -1278,7 +1278,7 @@ export class UserAgentApplication {
         this.logger.verbose("Renew Idtoken Expected state: " + serverAuthenticationRequest.state);
 
         // Build urlNavigate with "prompt=none" and navigate to URL in hidden iFrame
-        const urlNavigate = UrlUtils.urlRemoveQueryStringParameter(UrlUtils.createNavigateUrl(serverAuthenticationRequest), Constants.prompt) + Constants.prompt_none;
+        const urlNavigate = UrlUtils.urlRemoveQueryStringParameter(UrlUtils.createNavigateUrl(serverAuthenticationRequest), Constants.prompt) + Constants.prompt_none + Constants.response_mode_fragment;
 
         if (this.silentLogin) {
             window.requestType = Constants.login;
--- a/lerna.json
+++ b/lerna.json
@@ -4,5 +4,6 @@
     "lib/*",
     "samples/*"
   ],
-  "version": "independent"
+  "version": "independent",
+  "concurrency": 2
 }


DarylThayil (approved):

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1253,7 +1253,7 @@ export class UserAgentApplication {
         this.logger.verbose("Renew token Expected state: " + serverAuthenticationRequest.state);
 
         // Build urlNavigate with "prompt=none" and navigate to URL in hidden iFrame
-        const urlNavigate = UrlUtils.urlRemoveQueryStringParameter(UrlUtils.createNavigateUrl(serverAuthenticationRequest), Constants.prompt) + Constants.prompt_none;
+        const urlNavigate = UrlUtils.urlRemoveQueryStringParameter(UrlUtils.createNavigateUrl(serverAuthenticationRequest), Constants.prompt) + Constants.prompt_none + Constants.response_mode_fragment;
```
pkanher617: I think we should move this inside the urlRemoveQueryStringParameter function.

pkanher617: Nvm, there is a lot of work to change that, doesn't need to be done.

pkanher617 (approved):

sameerag (approved):

Issue #1000, opened 2019-09-19
Msal Js Support Auth Code Flow In The Browser


DarylThayil:
Goals: We would like to support the authorization code flow with PKCE for Single Page Applications in browser environments. 

Milestones: 
- [x] Alpha Release
This is currently available in npm [here](https://www.npmjs.com/package/@azure/msal-browser). 

This is the first version of the library available for private preview. We do NOT recommend anyone use this in their production apps - this is only for testing if anyone is interested. Please feel free to reach out if you run into any issues or have any feedback for improving what we have released. We will have a more detailed announcement and release notes for this when we reach the beta stage.

In the coming weeks, we will be updating this package actively based on internal testing and external feedback.

- [x] Beta Release
Coming soon!

- [ ] Main Release

pkanher617:
Work has started for this item. 
- Basic design work
- Requirement gathering for using auth code, supported use cases, etc.
- Hope to have POC for this work ready within two weeks

garrettm:
Is there an ETA for this landing?

sergey-tihon:
@garrettm  It is inside [October 2019](https://github.com/AzureAD/microsoft-authentication-library-for-js/milestone/12) milestone, so we may expect to try it soon 🤞

DarylThayil:
We are working on it in October! @garethj-msft  @sergey-tihon  however it will likely take us through the rest of the quarter to finish.  Expect to be able to use it in production early 2020.

garrettm:
Alright, thanks.  Might there be a less stable version we can use beforehand?  Or is there an alternative library I could use for this until msal is ready?  https://github.com/AzureAD/azure-activedirectory-library-for-js ?

Romiko:
This will be awesome to be able to use SPA/PWA apps with Microsoft Azure AD and Authorization Code Flow with PKCE, is this going to enable the possibility?

sameerag:
Yes. @pkanher617 to add more details.

pkanher617:
@garrettm @sergey-tihon We are currently developing this, hope to have the beta version ready by Thanksgiving, which will also include backwards compat for implicit flow. 

@garrettm msal for implicit flow is available and currently at v1.1.3, so you can use that instead of the ADAL library for now. 

@Romiko yes, we would like to support the use cases for SPA/PWA apps using Auth Code w/ PKCE.

garrettm:
[...]

PR #1106, opened 2019-11-11
Authorization Code with PKCE in Single-Page Applications: Proof-of-Concept

pkanher617:
This is a Proof of Concept for the using the OAuth 2.0 Authorization Code Flow with PKCE in Single-Page Applications described in issue #1000. This PR is the first of many to continually improve the MSAL library and bring this flow to parity. 

As a part of this POC, we have also separated much of the platform-specific code from the protocol code, and created separate packages. This will enable us to use this flow for other platforms in the future, including Electron, Node.js, and more.

This PR should eventually have a working version of the POC for Electron using the Auth Code implementation in the msal-common project.


lib/msal-browser-poc/rollup.config.js
```diff
@@ -0,0 +1,76 @@
+import typescript from "rollup-plugin-typescript2";
+import { uglify } from "rollup-plugin-uglify";
+import resolve from "rollup-plugin-node-resolve";
+import pkg from "./package.json";
+
+const libraryHeader = `/*! ${pkg.name} v${pkg.version} ${new Date().toISOString().split('T')[0]} */`;
+const useStrictHeader = `'use strict';`;
+const fileHeader = `${libraryHeader}\n${useStrictHeader}`;
+
+export default [
```
DarylThayil: looking forward to hearing / reading your thoughts on why roll up may be a good choice for the lib

lib/msal-browser-poc/src/app/PublicClientApplication.ts
```diff
@@ -0,0 +1,285 @@
...
+        this.config = buildConfiguration(configuration);
+
+        // Create browser storage
+        this.cacheStorage = new BrowserStorage(this.config.auth.clientId, this.config.cache);
+
+        // Initialize the network module
+        this.networkClient = new FetchClient();
+
+        // Initialize crypto module
+        this.crypto = new BrowserCrypto();
```
DarylThayil: do any of these have to be set on the instance can they just be passed along ?

lib/msal-browser-poc/src/error/ClientBrowserAuthError.ts
```diff
@@ -0,0 +1,34 @@
+/*
```
DarylThayil: I wonder if they errors need to be split by browser or non browser.

for instance a common/errors/ConfigurationError, could be used by both and then browser specific classes might life in msal-browser such as IframeError (or something)

lib/msal-browser-poc/src/network/XHRClient.ts
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * XHR client for JSON endpoints
+ * https://www.npmjs.com/package/async-promise
+ * @hidden
+ */
+export class XhrClient implements IXhrClient {
+    public sendRequestAsync(url: string, requestParams: RequestInit, enableCaching?: boolean): Promise<any> {
+        return new Promise<string>((resolve, reject) => {
+            const xhr = new XMLHttpRequest();
```
DarylThayil: can we use fetch?

lib/msal-common-poc/src/app/MsalConfiguration.ts
```diff
@@ -0,0 +1,141 @@
...
+ *
+ * @param TAuthOptions
+ * @param TStorageOptions
+ * @param TSystemOptions
+ * @param TFrameworkOptions
+ *
+ * @returns TConfiguration object
+ */
+export function buildConfiguration({ auth, storageInterface, networkInterface, cryptoInterface }: MsalConfiguration): MsalConfiguration {
+    const overlayedConfig: MsalConfiguration = {
```
DarylThayil: do we think configuration between server and browser will use the same common object?

lib/msal-browser-poc/package.json
```diff
@@ -0,0 +1,62 @@
...
+    "directories": {
+        "test": "test"
+    },
+    "files": [
+        "dist"
+    ],
+    "scripts": {
+        "clean": "shx rm -rf dist lib",
+        "lint": "eslint src --ext .ts",
+        "build:all": "npm run build:common && npm link msal-common && npm run build",
```
jasonnutter: `lerna bootstrap` will take care of the linking for you.

lib/msal-browser-poc/package.json
```diff
@@ -0,0 +1,62 @@
...
+        "@typescript-eslint/parser": "^2.4.0",
+        "eslint": "^6.5.1",
+        "husky": "^3.0.9",
+        "rollup": "^1.24.0",
+        "rollup-plugin-node-resolve": "^5.2.0",
+        "rollup-plugin-typescript2": "^0.24.3",
+        "rollup-plugin-uglify": "^6.0.3",
+        "shx": "^0.3.2",
+        "tslib": "^1.10.0",
+        "tslint": "^5.20.0",
```
jasonnutter: Thought this needs to be a normal dep?

lib/msal-browser-poc/package.json
```diff
@@ -0,0 +1,62 @@
...
+        "js",
+        "AAD",
+        "msal",
+        "oauth"
+    ],
+    "main": "./dist/index.js",
+    "module": "./dist/index.es.js",
+    "types": "./dist/index.d.ts",
+    "engines": {
+        "node": ">=0.8.0"
```
jasonnutter: We should change this to whatever version is run in CI

lib/msal-browser-poc/package.json
```diff
@@ -0,0 +1,62 @@
+{
+    "name": "msal-browser",
```
jasonnutter: We need to decide if we want this or `@azure/msal-browser` (which is probably better for branding/consistency)

lib/msal-browser-poc/src/network/XHRClient.ts
pkanher617: This implementation is for the implicit flow, in the auth code library we will switch to using fetch.

lib/msal-common-poc/src/app/MsalConfiguration.ts
pkanher617: I think we will have different auth options for server and browser, but everything else will be the same. So I am separating that into two different config objects in PR #1109.

lib/msal-browser-poc/package.json
pkanher617: Ok is there a recommendation for what version we should be using? In CI we are just using latest stable.

lib/msal-browser-poc/package.json
pkanher617: You're correct, I will rectify this in PR #1109.

lib/msal-browser-poc/package.json
pkanher617: I will implement lerna in a future PR, just trying to get POC working for now.

lib/msal-browser-poc/package.json
pkanher617: We should decide on this in PR #1108.

lib/msal-browser-poc/src/error/ClientBrowserAuthError.ts
pkanher617: That's the idea, I will flesh this out more in future PRs.

PR #1107, opened 2019-11-12
Authorization Code Flow for Single Page Applications: Base Branch Pull Request

pkanher617:
This PR is the base PR for Authorization Code Flow. This will be the branch which will eventually hold all of the code for the package which supports the Authorization Code Flow in the browser. This PR will be in draft until all other Authorization Code Flow PRs have been accepted.


lib/msal-browser/.eslintrc.json
```diff
@@ -0,0 +1,37 @@
+{
```
DarylThayil: can we have an eslint package for all of our options so we don't have to copy them?

/lib/eslint-config-msal  

cc @jasonnutter

DarylThayil:
+1 for small focussed pr. 
Left comment for eslint being made into a common package.
Would also think about whether anything else should be commonized that we are copying and pasting

lib/msal-browser/.eslintrc.json
jasonnutter: Yes, I'm working on this for Angular.

PR #1108, opened 2019-11-12
Authorization Code Flow for Single Page Applications: Rollup Build Files

pkanher617:
This is the first PR required to be merged into the base branch to support the Authorization Code Flow for Single Page Applications.

This PR introduces [Rollup](https://rollupjs.org/guide/en/) as the module bundler for the Authorization Code Flow instead of webpack as used in the current msal-core implementation. 

# Rollup vs Webpack
- Rollup has node polyfills for import/export, which makes it a little easier to support JS project structures. Webpack does not have these polyfills
- Rollup has support for relative paths, so no need to use path.resolve/path.join
- Use plugins instead of loaders for transformations and bundling in Rollup vs Webpack
- Tree shaking is enabled by default in Rollup, which eliminates dead code and unused exports. There is extra configuration required to do this in Webpack. This ensures we can keep the library as lightweight as possible.
- Rollup does not have comprehensive support for code splitting while Webpack does. This is a noted feature in Webpack, whereas Rollup has recently implemented it and may not have as much feature support.
- Rollup is generally faster than webpack for bundling, which would save us time when we are rolling out updates to multiple packages.
- A general rule of thumb is to use Webpack for apps, and Rollup for libraries (though as both technologies have matured, this has become less true).

### DIFF
--- a/lib/msal-browser/.huskyrc
+++ b/lib/msal-browser/.huskyrc
@@ -1,5 +1,5 @@
 {
   "hooks": {
-      "pre-commit": ""
+      "pre-commit": "npm run build"
   }
 }
--- a/lib/msal-browser/rollup.config.js
+++ b/lib/msal-browser/rollup.config.js
@@ -1,0 +1,76 @@
+import typescript from "rollup-plugin-typescript2";
+import { uglify } from "rollup-plugin-uglify";
+import resolve from "rollup-plugin-node-resolve";
+import pkg from "./package.json";
+
+const libraryHeader = `/*! ${pkg.name} v${pkg.version} ${new Date().toISOString().split('T')[0]} */`;
+const useStrictHeader = `'use strict';`;
+const fileHeader = `${libraryHeader}\n${useStrictHeader}`;
+
+export default [
+    {
+        input: "src/index.ts",
+        output: [
+            {
+                file: pkg.main,
+                format: "cjs",
+                banner: fileHeader,
+                sourcemap: "inline"
+            },
+            {
+                file: pkg.module,
+                format: "es",
+                banner: fileHeader,
+                sourcemap: "inline"
+            },
+            {
+                file: "./lib/msal-browser.js",
+                format: "umd",
+                name: "Msal",
+                banner: fileHeader,
+                sourcemap: "inline"
+            }
+        ],
+        external: [
+            ...Object.keys(pkg.dependencies || {}),
+            ...Object.keys(pkg.peerDependencies || {})
+        ],
+        plugins: [
+            typescript({
+                typescript: require('typescript')
+            }),
+            resolve({
+                only: ["msal-common"]
+            })
+        ]
+    },
+    // Minified version of msal
+    {
+        input: "src/index.ts",
+        output: [
+            {
+                file: "./lib/msal-browser.min.js",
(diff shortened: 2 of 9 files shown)


lib/msal-browser/rollup.config.js
```diff
@@ -0,0 +1,76 @@
+import typescript from "rollup-plugin-typescript2";
+import { uglify } from "rollup-plugin-uglify";
+import resolve from "rollup-plugin-node-resolve";
+import pkg from "./package.json";
+
+const libraryHeader = `/*! ${pkg.name} v${pkg.version} ${new Date().toISOString().split('T')[0]} */`;
+const useStrictHeader = `'use strict';`;
+const fileHeader = `${libraryHeader}\n${useStrictHeader}`;
+
+export default [
```
DarylThayil: would like to understand why you chose roll up and the comparison to webpack.
Also would like to hear plan if we choose this to standardize.

do we want to make a common package ? 

/lib/rollup/webpack-config 

that standardizes the common things and takes the rest as params

```js 
import { createRollupApp } from 'rollup-msal';

export default createRollupApp('msal-browser', {  sourceMaps: true, someOtherOption: true }) 

```

lib/msal-common/src/app/module/AuthModule.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+/**
+ * @hidden
+ * @ignore
+ * Data type to hold information about state returned from the server
+ */
+export type ResponseStateInfo = {
```
DarylThayil: should this be in a different file?

DarylThayil:
again huge fan of the way you are breaking this up! 
1 comment / suggestion on rollup

lib/msal-browser/rollup.config.js
pkanher617: In the main PR description I added some comparisons between the two and the advantages rollup has over webpack that make it a better choice for us.

pkanher617: I think each msal package should have its own rollup config file. This will enable us to use different build definitions depending on the platform.

lib/msal-common/src/app/module/AuthModule.ts
pkanher617: This could potentially go in the respone handler file, I will move it when I release the PR for the response handling.

jmckennon:
From what I can see (and @pkanher617 please correc me if I'm wrong), the biggest advantage here seems to be the tree-shaking and node polyfills in Rollup, as after some research, it seems that other features are similar enough to not matter either way. What are the explicit ways these help our library? What are the cases where those polyfills will be useful? Are we confident that the tree-shaking will be truly useful? Mostly for my own learning, and also looking for specific justification of the change.

lib/msal-browser/.huskyrc
```diff
@@ -1,5 +1,5 @@
 {
   "hooks": {
-      "pre-commit": ""
+      "pre-commit": "npm run build"
```
jasonnutter: I would personally prefer `npm run lint`.

PR #1109, opened 2019-11-12
Authorization Code Flow for Single Page Applications: API and Major Types

pkanher617:
Please review PR #1108 before this PR.

This PR establishes the major public APIs and types required for the Authorization Code Flow for Single Page Applications package. This includes types for requests, responses, configuration and interfaces to implement to complete the token acquisition flows.


lib/msal-browser/src/app/Configuration.ts
```diff
@@ -0,0 +1,141 @@
...
+import { BrowserUtils } from "../utils/BrowserUtils";
+import { AuthOptions } from "msal-common";
+// import { TelemetryEmitter } from "./telemetry/TelemetryTypes";
+
+/**
+ * Cache location options supported by MSAL are:
+ * - local storage: MSAL uses browsers local storage to store its cache
+ * - session storage: MSAL uses the browsers session storage to store its cache
+ */
+export type CacheLocation = "localStorage" | "sessionStorage";
```
sameerag: We should change this. This is causing issues in implicit-flow with type widening in certain frameworks.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -3,6 +3,190 @@
  * Licensed under the MIT License.
  */
 
+import * as msalAuth from "msal-common";
+import { Configuration, buildConfiguration } from "./Configuration";
+
+/**
+ * Interface to handle iFrame generation, Popup Window creation and redirect handling
+ */
+declare global {
```
sameerag: Why should this be in a public interface? Can we have this in an internal file?

lib/msal-browser/src/app/Configuration.ts
pkanher617: Can you explain the issue or provide an alternative?

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: This was copied from UAA, and I'm assuming it's because this is the general entry point for the library. I will do some research around how defining globals should work.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -3,6 +3,190 @@
...
+     * Use to get the post logout redirect uri configured in MSAL or null.
+     * Evaluates postLogoutredirectUri if its a function, otherwise simply returns its value.
+     *
+     * @returns {string} post logout redirect URL
+     */
+    public getPostLogoutRedirectUri(): string {
+        return this.authModule.getPostLogoutRedirectUri();
+    }
+
+    // #endregion
```
sameerag: Can we break this file up? Constructor in one  file / all public APIs only for token acquisition in one file / redirect URIs and callbacks in one?

pkanher617: I'm not sure how that helps really, and in terms of code structure it would be difficult to do. I'm happy to take suggestions on ways to do this, but I'm fairly certain this class won't be longer than 500 lines of code. Iframe and popup code will be moved to separate classes, and most of the protocol logic is in the msal common package. @DarylThayil @jasonnutter @jmckennon Would like to hear your thought as well.

pkanher617: I'm also of the opinion that if this class is really becoming too large, the solution isn't to split the class definition into different files, but to split the class itself into different classes, which is what I plan to do with the browser specific code (i.e. iframe, popup, etc.)

jmckennon: I'd say that yeah, if it's going to be 500 lines, it's likely we'd want so split it up in some way. I don't think I have a strong opinion about how that necessarily happens, but generally sorting ideas into smaller files is good, IMO.

pkanher617: That's fine, I will separate them as I add implementation then. For now just focus on API surface and whether we have logical separation between browser and common.

lib/msal-browser/src/app/Configuration.ts
pkanher617: Make an enum in Constants.ts.

lib/msal-browser/src/app/Configuration.ts
```diff
@@ -0,0 +1,141 @@
...
+ * - framework: this is where you can configure the running mode of angular. More to come here soon.
+ */
+export type Configuration = {
+    auth: AuthOptions,
+    cache?: CacheOptions,
+    system?: SystemOptions,
+    framework?: FrameworkOptions
+};
+
+const DEFAULT_AUTH_OPTIONS: AuthOptions = {
```
jmckennon: Do all of these `DEFAULT_X` really need to be prefixed by default? is there another place we'll store options that aren't the default? seems a bit confusing at surface-level

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -3,6 +3,190 @@
...
+     */
+    loginRedirect(request: msalAuth.AuthenticationParameters): void {
+        throw new Error("Method not implemented.");
+    }
+
+    /**
+     * Use when you want to obtain an access_token for your API by redirecting the user's browser window to the authorization endpoint.
+     * @param {@link (AuthenticationParameters:type)}
+     *
+     * To renew idToken, please pass clientId as the only scope in the Authentication Parameters
```
jmckennon: this statement here feels like we may want a separate function to explicitly to renew an idToken, rather than using parameters as a functionality-select.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -3,6 +3,190 @@
...
+    // #region Popup Flow 
+
+    /**
+     * Use when initiating the login process via opening a popup window in the user's browser
+     *
+     * @param {@link (AuthenticationParameters:type)}
+     *
+     * @returns {Promise.<TokenResponse>} - a promise that is fulfilled when this function has completed, or rejected if an error was raised. Returns the {@link AuthResponse} object
+     */
+    loginPopup(request: msalAuth.AuthenticationParameters): Promise<msalAuth.TokenResponse> {
```
jmckennon: this may be more for my own learning, but can you clarify the difference in functionality between `loginPopup` and `loginRedirect`? Is the latter meant to be silent?

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -3,6 +3,190 @@
...
+    }
+
+    // #endregion
+
+    // #region Window Navigation
+
+    /**
+     * @hidden
+     * Used to redirect the browser to the STS authorization endpoint
+     * @param {string} urlNavigate - URL of the authorization endpoint
```
jmckennon: is this... necessary? does the endpoint vary in this case? even if it does, we may want to enumerate the various calls

lib/msal-common/src/response/CodeResponse.ts
```diff
@@ -0,0 +1,15 @@
...
+ */
+
+import { AuthResponse } from "./AuthResponse";
+
+/**
+ * CodeResponse type returned by library containing authorization code.
+ * - code: authorization code returned from interactive call
+ */
+export type CodeResponse = AuthResponse &
+{
```
jmckennon: nit: bracket should be on previous line

lib/msal-common/src/request/AuthenticationParameters.ts
```diff
@@ -0,0 +1,56 @@
...
+ * - claimsRequest: stringified claims object to request additional claims in a token
+ * - authority: authority to request tokens from
+ * - state: state parameter to ensure request/response integrity
+ * - correlationId: custom correlationId given by user
+ * - account: Account object to perform SSO
+ * - sid: session id for SSO
+ * - loginHint: login hint for SSO
+ * - forceRefresh: Forces silent requests to make network calls if true
+ */
+export type AuthenticationParameters = {
```
jmckennon: is there a reason that this type has no defaults set, whereas many of the others do?

lib/msal-browser/src/app/Configuration.ts
jmckennon: I also think it may be worth just instantiating these as the default using optionals in the constructor rather than rewriting all of this elsewhere

lib/msal-browser/src/app/Configuration.ts
```diff
@@ -0,0 +1,141 @@
+/*
```
DarylThayil: Should this be in common?

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -3,6 +3,190 @@
...
+     *
+     * @param {@link (Configuration:type)} configuration object for the MSAL PublicClientApplication instance
+     */
+    constructor(configuration: Configuration) {
+        // Set the configuration
+        this.config = buildConfiguration(configuration);
+
+        this.authModule = new msalAuth.AuthorizationCodeModule({
+            auth: this.config.auth,
+            cryptoInterface: null,
```
DarylThayil: assuming these are still to be added?

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -3,6 +3,190 @@
...
+    // #region Redirect Flow
+
+    /**
+     * Set the callback functions for the redirect flow to send back the success or error object.
+     * @param {@link (authCallback:type)} authCallback - Callback which contains
+     * an AuthError object, containing error data from either the server
+     * or the library, depending on the origin of the error, or the AuthResponse object 
+     * containing data from the server (returned with a null or non-blocking error).
+     */
+    handleRedirectCallback(authCallback: authCallback): void {
```
DarylThayil: should this be private ?

lib/msal-browser/src/app/PublicClientApplication.ts
DarylThayil: My personal opinion is to split it like this 

login-provider.js
token-provider.js

---

```js 
import { acquireTokenSilentProvider } from './providers/token-provider.js';

export default class PublicClientApplication { 
   acquireTokenSilent(request: Request): Promise<Response> {
      //log / telemtry
      // create correlation id? 
      return acquireTokenSilentProvider(request, correlationId)
              .catch(err => {
                    // able to log or transform
                    throw err;
              })
  }
}

This way you separate api / resource from business logic / provider 
The business logic can change but the api stays the same. 
Also promotes reuse

File: lib/msal-common/src/app/config/MsalConfiguration.ts [hunk 13]
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * Use the configuration object to configure MSAL Modules and initialize the base interfaces for MSAL.
+ *
+ * This object allows you to configure important elements of MSAL functionality:
+ * - storage: this is where you configure storage implementation.
+ * - network: this is where you can configure network implementation.
+ * - crypto: implementation of crypto functions
+ */
+export type MsalConfiguration = {
```
DarylThayil: I would name this something different, like` MsalModuleConfiguration `

lib/msal-common/src/app/config/MsalConfiguration.ts
```diff
@@ -0,0 +1,88 @@
...
+    storageInterface: ICacheStorage,
+    networkInterface: INetworkModule,
+    cryptoInterface: ICrypto
+};
+
+const DEFAULT_STORAGE_OPTIONS: ICacheStorage = {
+    clear: () => {
+        console.log("clear() has not been implemented for the cacheStorage interface.");
```
DarylThayil: might want to do console.warn instead of log

lib/msal-common/src/app/config/MsalConfiguration.ts
```diff
@@ -0,0 +1,88 @@
...
+ * - network: this is where you can configure network implementation.
+ * - crypto: implementation of crypto functions
+ */
+export type MsalConfiguration = {
+    storageInterface: ICacheStorage,
+    networkInterface: INetworkModule,
+    cryptoInterface: ICrypto
+};
+
+const DEFAULT_STORAGE_OPTIONS: ICacheStorage = {
```
DarylThayil: I am assuming some of these are required to be implemented or they will cause error right?

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: This comment should be in ATS - fix @pkanher617

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -3,6 +3,190 @@
...
+
+    // #region Window Navigation
+
+    /**
+     * @hidden
+     * Used to redirect the browser to the STS authorization endpoint
+     * @param {string} urlNavigate - URL of the authorization endpoint
+     */
+    private navigateWindow(urlNavigate: string): void {
+        throw new Error("Method not implemented.");
```
pkanher617: Move navigateWindow to WindowUtils or BrowserUtils

lib/msal-common/src/app/config/MsalConfiguration.ts
```diff
@@ -0,0 +1,88 @@
...
+};
+
+const DEFAULT_NETWORK_OPTIONS: INetworkModule = {
+    sendRequestAsync: (url: string, method: RequestInit, enableCaching?: boolean): Promise<any> => {
+        console.log("Network interface - sendRequestAsync() has not been implemented");
+        return null;
+    }
+};
+
+const DEFAULT_CRYPTO_IMPLEMENTATION: ICrypto = {
```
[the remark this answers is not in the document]
pkanher617: I think we resolved this earlier, but this is mostly just to support defaults. I think breaking this up is a little overkill, but happy to re-open if we have further issues.

lib/msal-common/src/error/AuthError.ts
```diff
@@ -0,0 +1,40 @@
...
+    unexpectedError: {
+        code: "unexpected_error",
+        desc: "Unexpected error in authentication."
+    }
+};
+
+/**
+ * General error class thrown by the MSAL.js library.
+ */
+export class AuthError extends Error {
```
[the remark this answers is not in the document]
pkanher617: Will do in future PRs

lib/msal-common/src/request/AuthenticationParameters.ts
pkanher617: The params are optional instead, and completely depend on desired experience, so we can't really have defaults. We should have a null check function, but that will come in a future PR.

lib/msal-common/src/app/config/MsalConfiguration.ts
pkanher617: Yes, I will add the errors in future PRs when the detail code is added.

lib/msal-common/src/app/config/MsalConfiguration.ts
pkanher617: done

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: I think we should reserve judgement for this until later, when more implementation code is added.

lib/msal-browser/src/app/Configuration.ts
```diff
@@ -0,0 +1,141 @@
...
+/**
+ * App/Framework specific environment support
+ *
+ * - isAngular                - flag set to determine if it is Angular Framework. MSAL uses this to broadcast tokens. More to come here: detangle this dependency from core.
+ * - unprotectedResources     - Array of URI's which are unprotected resources. MSAL will not attach a token to outgoing requests that have these URI. Defaults to 'null'.
+ * - protectedResourceMap     - This is mapping of resources to scopes used by MSAL for automatically attaching access tokens in web API calls.A single access token is obtained for the resource. So you can map a specific resource path as follows: {"https://graph.microsoft.com/v1.0/me", ["user.read"]}, or the app URL of the resource as: {"https://graph.microsoft.com/", ["user.read", "mail.send"]}. This is required for CORS calls.
+ *
+ */
+export type FrameworkOptions = {
+    isAngular?: boolean;
```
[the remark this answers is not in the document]
pkanher617: Does that include the other parameters in FrameworkOptions?

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -3,6 +3,190 @@
...
 
+import * as msalAuth from "msal-common";
+import { Configuration, buildConfiguration } from "./Configuration";
+
+/**
+ * Interface to handle iFrame generation, Popup Window creation and redirect handling
+ */
+declare global {
+    interface Window {
+        openedWindows: Array<Window>;
```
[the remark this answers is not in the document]
pkanher617: removed.

lib/msal-browser/src/app/Configuration.ts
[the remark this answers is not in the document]
jasonnutter: Yes.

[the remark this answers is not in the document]
pkanher617: Done in a separate PR.

pkanher617:
Try to not use "msal" in filename and class names.

Issue #117, opened 2017-08-29
What is the purpose of class attribute on script tag defined in Installation instruction


ggirard07:
**Not sure whether this is the right place to post it or not**

I noticed in the `README.md` that a `class="pre"` is defined on the script tag.

> Note that msal.js is built for ES5, therefore enabling support of Internet Explorer 11. If you want to target Internet Explorer, you'll need to add a reference to promises polyfill. You might want to read more in the FAQ
> 
>     
>     <script src="https://cdnjs.cloudflare.com/ajax/libs/bluebird/3.3.4/bluebird.min.js" class="pre"></script> 
>     <script src="https://secure.aadcdn.microsoftonline-p.com/lib/0.1.1/js/msal.min.js"></script>
> 

As I am pretty new to HTML5 and JavaScript, I can't understand its purpose (no googling neither stackoverflow gave me a real answer).
Is there any real utility for it? Or is that an internal stuff only for `bluebird`?

I now this is not really an issue... but I like to understand what I put in my code :)

jmprieur:
Hi Guillaume.
MSAL is written in TypeScript, which then generates JavaScript. JavaScript has several standard levels ([Ecma Script](https://fr.wikipedia.org/wiki/ECMAScript) 5, 6). Unfortunately, Internet Explorer does not support the latest standards, therefore we have chosen to generate from our TypeScript code, the msal.js library compliant to ECMA Script 5. However since our code uses promises (more about that below) to simplify the usage of the library, to be able to us it with IE11, you need to add a reference to the bluebird library which adds support for promises. [Promises](https://www.promisejs.org/) are a syntax where you can write after a call to an asynchronous method, how you want to process the results (successful or errors). This is the .then() that you see in the code snippet if you want to use your application with JavaScript.
HTH

jmprieur:
I realize I have not answered to your question about pre in the script. @rohitnarula7176: would you know?

rohitnarula7176:
@ggirard07 @jmprieur The <pre> tage defines preformatted text. <pre> element text is displayed in a fixed-width font which preserves both spaces and line breaks. It is just the way the bluebird promise library wanted its content rendered in the browser.

cicorias:
The question is what does `pre` do in this context as shon in the example devapps

```
<script class="pre">
        var applicationConfig = {
            clientID: 'cd1fdffa-0c1e-4490-a070-812701100306',
[...]

PR #1112, opened 2019-11-13
Remove class attribute from script tag

pkanher617:
Small fix. Class attr on script tag doesn't add anything as far as I know, and there was an issue raised a while back (#117).


jmckennon:
LGTM, but is there a reason coveralls says are code coverage went down so much? Does it just have a bit of variance?

pkanher617:
> LGTM, but is there a reason coveralls says are code coverage went down so much? Does it just have a bit of variance?

It looks like there is some variance, I'll take a look at that later. Not sure why there would be variance, but when I re-ran the build it said no change.

jasonnutter:
LGTM

PR #1113, opened 2019-11-14
Ensure errors thrown when monitoring window are caught and handled

jasonnutter:
Currently, when `monitorWindowForHash` threw an error, it wasn't properly caught, which resulted in unhandled rejection errors (and the inability for apps to catch and handle these errors on their own). This PR ensures that these errors are caught, and properly passed up to the call site that initiate the request.

### DIFF
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -521,9 +521,9 @@ export class UserAgentApplication {
 
             // popUpWindow will be null for redirects, so we dont need to attempt to monitor the window
             if (popUpWindow) {
-                const hash = await WindowUtils.monitorWindowForHash(popUpWindow, this.config.system.loadFrameTimeout, urlNavigate);
-                if (hash) {
-                    // Hash found
+                try {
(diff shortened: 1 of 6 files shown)


samples/react-sample-app/src/AuthProvider.js
```diff
@@ -34,6 +34,8 @@ export default C =>
                     return redirect
                         ? msalApp.acquireTokenRedirect(request)
                         : msalApp.acquireTokenPopup(request);
+                } else {
+                    console.error('Non-interactive error:', error.errorCode)
```
jasonnutter: Timeout errors were previously not bubbling up to the call site in the application.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1260,7 +1268,7 @@ export class UserAgentApplication {
         this.registerCallback(serverAuthenticationRequest.state, scope, resolve, reject);
         this.logger.infoPii("Navigate to:" + urlNavigate);
         frameHandle.src = "about:blank";
-        this.loadIframeTimeout(urlNavigate, frameName, scope);
+        this.loadIframeTimeout(urlNavigate, frameName, scope).catch(error => reject(error));
```
jmckennon: is there a reason why in some places we have `if (reject)` after catching the error and sometimes not?

PR #1114, opened 2019-11-14
Add fromCache flag to AuthResponse to indicate token was returned from cache

jasonnutter:
This will allow applications to programmatically tell if the token was returned from the cache or from the network, which can be helpful for performance measurements of token requests.


jasonnutter:
FYI @lahuey

pkanher617:
I would add some tests so coveralls doesn't complain.

Issue #1118, opened 2019-11-15
Casing issues with module


RMacfarlane:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Other... Please describe:
</code></pre>

## Browser: not browser dependent
- [x] Chrome version XX
- [x] Firefox version XX
- [x] IE version XX
- [x] Edge version XX
- [x] Safari version XX

## Current behavior
The `webpack.config.js` script specifies the library as "Msal" instead of "msal". This produces this header in the bundled `msal.js`:
```
(function webpackUniversalModuleDefinition(root, factory) {
	if(typeof exports === 'object' && typeof module === 'object')
		module.exports = factory();
	else if(typeof define === 'function' && define.amd)
		define("Msal", [], factory);
	else if(typeof exports === 'object')
		exports["Msal"] = factory();
	else
		root["Msal"] = factory();
}) 
```

## Expected behavior
Should be 'msal' to match library name.

## Minimal reproduction of the problem with instructions
Try to dynamically import the library in typescript with `import('msal')`. This fails. Using `import('Msal')` succeeds at runtime, but causes issues because of the case mismatch.

jasonnutter:
@RMacfarlane Can you please clarify what you mean when you say `import('Msal')` causes issues? What sort of issues, e.g. typing errors in TypeScript? And if an application is currently doing `import('Msal')`, would changing the casing cause their app to break? Thanks!

TheLarkInn:
@jasonnutter the problem here is that @RMacfarlane is trying to use a bundler (webpack/rollup/etc) the UMD bundle (which has its module name defined as `Msal`). 

What @RMacfarlane should do instead is use the es-modules format of the library instead, which will prevent him from having this issue. This means `npm install msal` for their project and not using it as a script tag.

jasonnutter:
@TheLarkInn Gotya, thanks! We're currently planning to fix this in v2, as it would be a breaking change to switch.

pkanher617:
Closing this as we have changed the export in @azure/msal-browser to match the name of the package.

PR #1119, opened 2019-11-15
Fix casing of module name in webpack

RMacfarlane:
https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/1118


pkanher617:
@RMacfarlane Thanks for the submission. This is a breaking change for people currently using the library as "Msal", so we will try to incorporate this in a future version of msal. Thanks for your patience.

jasonnutter:
Closing, as this will be addressed in v2.

PR #1133, opened 2019-11-25
Authorization Code Flow for Single Page Applications: Authority and Protocol Classes

pkanher617:
Please review PR #1109 before this one.

This PR adds authority and protocol-specific classes to the Authorization Code Flow library. Any changes to the account, IdToken, client info or authority classes should be incorporated in this PR. Most of these classes have been lifted from the implicit flow library, and should be thoroughly reviewed before merging.


lib/msal-common/src/auth/ClientInfo.ts
```diff
@@ -0,0 +1,56 @@
...
+    get utid(): string {
+        return this._utid ? this._utid : "";
+    }
+
+    set utid(utid: string) {
+        this._utid = utid;
+    }
+
+    constructor(rawClientInfo: string, crypto: ICrypto) {
+        if (!rawClientInfo || StringUtils.isEmpty(rawClientInfo)) {
```
DarylThayil: are there times there are empty client info and we want to construct the class?

lib/msal-common/src/auth/ClientInfo.ts
```diff
@@ -0,0 +1,56 @@
...
+            this.uid = "";
+            this.utid = "";
+            return;
+        }
+
+        try {
+            // const decodedClientInfo: string = crypto.base64Decode(rawClientInfo);
```
DarylThayil: remove comment?

lib/msal-common/src/auth/ClientInfo.ts
```diff
@@ -0,0 +1,56 @@
...
+            const clientInfo: ClientInfo = JSON.parse("decodedClientInfo") as ClientInfo;
+            if (clientInfo) {
```
DarylThayil: I am assuming that the part in this that can / would throw is the JSON.parse 
if that is the case you dont need the if(clientInfo) block

lib/msal-common/src/auth/ClientInfo.ts
```diff
@@ -0,0 +1,56 @@
...
+        return this._utid ? this._utid : "";
+    }
+
+    set utid(utid: string) {
+        this._utid = utid;
+    }
+
+    constructor(rawClientInfo: string, crypto: ICrypto) {
+        if (!rawClientInfo || StringUtils.isEmpty(rawClientInfo)) {
+            this.uid = "";
```
DarylThayil: should these be set as empty strings instead of being null ?

lib/msal-common/src/auth/IdToken.ts
```diff
@@ -0,0 +1,123 @@
...
+                }
+
+                if (this.claims.hasOwnProperty("cloud_instance_host_name")) {
+                    this.cloudInstance = this.claims["cloud_instance_host_name"];
+                }
+                /* tslint:enable:no-string-literal */
+            }
+        } catch (e) {
+            /*
+             * TODO: This error here won't really every be thrown, since extractIdToken() returns null if the decodeJwt() fails.
```
DarylThayil: if this never throws should we not use a try catch here/

lib/msal-common/src/auth/IdToken.ts
```diff
@@ -0,0 +1,123 @@
...
+            const base64IdToken = decodedToken.JWSPayload;
+            const base64Decoded = crypto.base64Decode(base64IdToken);
+            if (!base64Decoded) {
+                // this._requestContext.logger.info("The returned id_token could not be base64 url safe decoded.");
+                return null;
+            }
+            // ECMA script has JSON built-in support
+            return JSON.parse(base64Decoded);
+        } catch (err) {
+            // this._requestContext.logger.error("The returned id_token could not be decoded" + err);
```
DarylThayil: remove comment, and can we avoid empty catches?

lib/msal-common/src/auth/ClientInfo.ts
pkanher617: I believe there are cases in B2C where we may not receive the client info object, but I will double check to verify. It would probably be a better pattern to throw a warning and not create the object at all, so I will make some changes to reflect that.

lib/msal-common/src/auth/ClientInfo.ts
pkanher617: uid and utid are set in cache, so I think we do this in order to avoid null pointer exceptions. As I said above, it is probably a better pattern to handle it in the caller, so I will update the functionality.

lib/msal-common/src/auth/ClientInfo.ts
pkanher617: Commented out because this file was copied, I will update.

lib/msal-common/src/auth/ClientInfo.ts
pkanher617: +1, will change

lib/msal-common/src/auth/IdToken.ts
pkanher617: Agreed, wanted to revisit when we added functionality code but I can update it here.

lib/msal-common/src/auth/IdToken.ts
pkanher617: Agreed, wanted to revisit when we added functionality code but I can update it here.

lib/msal-common/src/auth/IdToken.ts
```diff
@@ -0,0 +1,123 @@
...
+    /* tslint:disable:no-string-literal */
+    constructor(rawIdToken: string, crypto: ICrypto) {
+        if (StringUtils.isEmpty(rawIdToken)) {
+            throw ClientAuthError.createIdTokenNullOrEmptyError(rawIdToken);
+        }
+        try {
+            this.rawIdToken = rawIdToken;
+            this.claims = IdToken.extractIdToken(rawIdToken, crypto);
+            if (this.claims) {
+                if (this.claims.hasOwnProperty("iss")) {
```
jmckennon: We might want to consider having enumerations/constants for these strings, even if they aren't being used elsewhere. Disabling the linter like this feels pretty bad if that's the configuration we've set up.

lib/msal-common/src/auth/IdToken.ts
jmckennon: +1

lib/msal-common/src/auth/ClientInfo.ts
pkanher617: Updated.

lib/msal-common/src/auth/ClientInfo.ts
pkanher617: Done.

lib/msal-common/src/auth/ClientInfo.ts
pkanher617: I have changed this, resolving comment.

lib/msal-common/src/auth/IdToken.ts
pkanher617: Removed try/catch.

lib/msal-common/src/auth/MsalAccount.ts
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * accountIdentifier       combination of idToken.uid and idToken.utid
+ * homeAccountIdentifier   combination of clientInfo.uid and clientInfo.utid
+ * userName                idToken.preferred_username
+ * name                    idToken.name
+ * idToken                 idToken
+ * sid                     idToken.sid - session identifier
+ * environment             idtoken.issuer (the authority that issues the token)
+ */
+export class MsalAccount {
```
jmckennon: For my own understanding, this is just basically just an abstraction around an id token, right? This is effectively our idea of some generic account and the information surrounding it?

lib/msal-common/src/auth/IdToken.ts
pkanher617: Updated try/catch

lib/msal-common/src/auth/ClientInfo.ts
pkanher617: Updated, please re-review

lib/msal-common/src/auth/ClientInfo.ts
```diff
@@ -0,0 +1,56 @@
...
+            this.utid = "";
+            return;
+        }
+
+        try {
+            // const decodedClientInfo: string = crypto.base64Decode(rawClientInfo);
+            const clientInfo: ClientInfo = JSON.parse("decodedClientInfo") as ClientInfo;
```
jasonnutter: Not sure I understand why JSON.parse is needed here?

lib/msal-common/src/error/ClientAuthError.ts
```diff
@@ -9,7 +9,26 @@ import { AuthError } from "./AuthError";
...
+        code: "null_or_empty_id_token",
+        desc: "The idToken is null or empty. Please review the trace to determine the root cause."
+    },
+    endpointResolutionError: {
+        code: "endpoints_resolution_error",
+        desc: "Error: could not resolve endpoints. Please check network and try again."
+    },
+    invalidAuthorityType: {
+        code: "invalid_authority_type",
+        desc: "The given authority is not a valid type of authority supported by MSAL. Please see here for valid authorities: <insert URL here>."
```
jmckennon: reminder to actually put that URL there

lib/msal-common/src/url/IUri.ts
```diff
@@ -0,0 +1,16 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+/**
+ * @hidden
+ */
+export interface IUri {
+    Protocol: string;
+    HostNameAndPort: string;
```
jmckennon: Is there any value in separating the hostname and port fields? is there ever a case where we'll have to parse this to only get one or the other?

lib/msal-common/src/auth/ClientInfo.ts
pkanher617: Updated, refresh the file. I made some changes to remove errors, I have reverted them.

lib/msal-common/src/auth/MsalAccount.ts
pkanher617: It's a combination of id token and client info, which is why we persist those two after login is completed.

lib/msal-common/src/error/ClientAuthError.ts
pkanher617: +1, thanks!

lib/msal-common/src/url/IUri.ts
pkanher617: I'm not sure, I believe we do use it when checking for secure uris, but beyond that it is just to make sure the object completely describes the segments of a URI.

lib/msal-common/src/auth/IdToken.ts
pkanher617: +1, I will make some changes.

lib/msal-common/src/auth/ClientInfo.ts
pkanher617: To answer your question, the decoded client info string looks like this:
`{"uid":"5j1345lj3l54j", "utid":"flkdsjf2ujfldskjf"}`

lib/msal-browser/src/utils/BrowserUtils.ts
```diff
@@ -15,7 +15,7 @@ export class BrowserUtils {
      * Used to redirect the browser to the STS authorization endpoint
      * @param {string} urlNavigate - URL of the authorization endpoint
      */
-    private navigateWindow(urlNavigate: string): void {
+    static navigateWindow(urlNavigate: string): void {
         throw new Error("Method not implemented.");
```
jasonnutter: Do we need this method? If so, why not implement?

lib/msal-common/src/app/config/MsalModuleConfiguration.ts
```diff
@@ -22,49 +23,60 @@ export type MsalModuleConfiguration = {
 
 const DEFAULT_STORAGE_OPTIONS: ICacheStorage = {
     clear: () => {
-        console.warn("clear() has not been implemented for the cacheStorage interface.");
+        const notImplErr = "clear() has not been implemented for the cacheStorage interface.";
+        console.warn(notImplErr);
```
jasonnutter: These shuold go to `logger.warn`

lib/msal-common/src/auth/ClientInfo.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    constructor(rawClientInfo: string, crypto: ICrypto) {
+        if (!rawClientInfo || StringUtils.isEmpty(rawClientInfo)) {
+            throw ClientAuthError.createClientInfoEmptyError(rawClientInfo);
+        }
+
+        try {
+            const decodedClientInfo: string = crypto.base64Decode(rawClientInfo);
+            const clientInfo = JSON.parse(decodedClientInfo);
```
jasonnutter: This could be typed to `ClientInfo` yea?

lib/msal-common/src/auth/IdToken.ts
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        if (this.claims.hasOwnProperty("sid")) {
+            this.sid = this.claims["sid"];
+        }
+
+        if (this.claims.hasOwnProperty("cloud_instance_host_name")) {
+            this.cloudInstance = this.claims["cloud_instance_host_name"];
+        }
+        /* tslint:enable:no-string-literal */
```
jasonnutter: Is this eslint override still needed?

lib/msal-common/src/auth/IdToken.ts
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    /**
+     * Extract IdToken by decoding the RAWIdToken
+     *
+     * @param encodedIdToken
+     */
+    static extractIdToken(encodedIdToken: string, crypto: ICrypto): any {
```
jasonnutter: `ClientInfo` on the return type correct?

lib/msal-common/src/auth/IdToken.ts
```diff
@@ -0,0 +1,109 @@
...
+        // id token will be decoded to get the username
+        const decodedToken = StringUtils.decodeJwt(encodedIdToken);
+        if (!decodedToken) {
+            return null;
+        }
+        try {
+            const base64IdToken = decodedToken.JWSPayload;
+            // base64Decode() should throw an error if there is an issue
+            const base64Decoded = crypto.base64Decode(base64IdToken);
+            // ECMA script has JSON built-in support
```
jasonnutter: Can clean up these comments.

lib/msal-common/src/auth/MsalAccount.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    accountIdentifier: string;
+    homeAccountIdentifier: string;
+    userName: string;
+    name: string;
+    idToken: StringDict; // will be deprecated soon
```
jasonnutter: This is the chance to do the deprecation, so I think this should be removed.

lib/msal-common/src/auth/MsalAccount.ts
```diff
@@ -0,0 +1,92 @@
...
+    static compareAccounts(a1: MsalAccount, a2: MsalAccount): boolean {
+        if (!a1 || !a2) {
+            return false;
+        }
+        if (a1.homeAccountIdentifier && a2.homeAccountIdentifier) {
+            if (a1.homeAccountIdentifier === a2.homeAccountIdentifier) {
+                return true;
+            }
+        }
+        return false;
```
jasonnutter: There's gotta be a simpler way to write this function.

lib/msal-common/src/auth/authority/AadAuthority.ts
```diff
@@ -0,0 +1,55 @@
...
+        const response = await this.networkInterface.sendRequestAsync(this.aadInstanceDiscoveryEndpointUrl, { method: "GET" }, true);
+        return response.tenant_discovery_endpoint;
+    }
+
+    /**
+     * Checks to see if the host is in a list of trusted hosts
+     * @param {string} The host to look up
+     */
+    public isInTrustedHostList(host: string): boolean {
+        return AADTrustedHostList[host.toLowerCase()];
```
jasonnutter: This is actually returning a string, not a boolean. Should be cast to boolean `return !!AADTrustedHostList[host.toLowerCase()];`

lib/msal-common/src/auth/authority/Authority.ts
```diff
@@ -0,0 +1,123 @@
...
+        } as ITenantDiscoveryResponse;
+    }
+
+    public abstract async getOpenIdConfigurationAsync(): Promise<string>;
+
+    public async resolveEndpointsAsync(): Promise<Authority> {
+        const openIdConfigEndpoint = await this.getOpenIdConfigurationAsync();
+        this.tenantDiscoveryResponse = await this.discoverEndpoints(openIdConfigEndpoint);
+
+        return this;
```
jasonnutter: Returning `this` here always seemed weird, given none of the other functions do it.

lib/msal-common/src/utils/Constants.ts
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * List of pre-established trusted host URLs. 
+ */
+export const AADTrustedHostList =  {
+    "login.windows.net": "login.windows.net",
+    "login.chinacloudapi.cn": "login.chinacloudapi.cn",
+    "login.cloudgovapi.us": "login.cloudgovapi.us",
+    "login.microsoftonline.com": "login.microsoftonline.com",
+    "login.microsoftonline.de": "login.microsoftonline.de",
+    "login.microsoftonline.us": "login.microsoftonline.us"
```
jasonnutter: I think this could be some other type (enum maybe?)

lib/msal-browser/src/utils/BrowserUtils.ts
pkanher617: Will be implemented in a future PR - just added it for the sake of the PR. I can remove it later.

lib/msal-common/src/app/config/MsalModuleConfiguration.ts
pkanher617: Logger hasn't been implemented yet.

lib/msal-common/src/auth/ClientInfo.ts
pkanher617: Have a question on this, will resolve offline.

lib/msal-common/src/auth/IdToken.ts
pkanher617: Jake had a comment on this, i will add an enum shortly.

lib/msal-common/src/auth/IdToken.ts
pkanher617: no it would be an object with id token parameters. We can fully define Id token if that would make it easier to define this, but for now we just parse the id token object as a JSON.

lib/msal-common/src/auth/MsalAccount.ts
pkanher617: Yes, I will change this to string in future PRs.

lib/msal-common/src/auth/MsalAccount.ts
pkanher617: I will take a look and refactor.

lib/msal-common/src/auth/authority/Authority.ts
pkanher617: I will revisit when we implement the acquireToken functions.

pkanher617: Changing to return the tenantDiscoveryResponse for now.

lib/msal-common/src/auth/authority/AadAuthority.ts
pkanher617: Done

lib/msal-common/src/auth/IdToken.ts
pkanher617: Cleaned up ECMA comment - the other comments should stay.

lib/msal-common/src/auth/IdToken.ts
```diff
@@ -0,0 +1,109 @@
...
+    subject: string;
+    tenantId: string;
+    version: string;
+    preferredName: string;
+    name: string;
+    homeObjectId: string;
+    nonce: string;
+    expiration: string;
+    rawIdToken: string;
+    claims: StringDict;
```
pkanher617: Create type for claims, make fields optional.

pkanher617: Test whether properties are dropped.

lib/msal-common/src/auth/IdToken.ts
pkanher617: See above comment.

lib/msal-browser/src/app/Configuration.ts
```diff
@@ -63,7 +63,6 @@ export type SystemOptions = {
  *
  */
 export type FrameworkOptions = {
-    isAngular?: boolean;
     unprotectedResources?: Array<string>;
     protectedResourceMap?: Map<string, Array<string>>;
 };
```
sameerag: How are we handling "Angular" broadcast messages then?

lib/msal-common/src/auth/ClientInfo.ts
sameerag: Some cases of device flow had this issue but eSTS guarantees client_info and if it is empty, we need to raise a bug against them. However it may be necessary to confirm the expected behavior from msal js in case it returns empty.

lib/msal-common/src/auth/IdTokenClaims.ts
```diff
@@ -0,0 +1,19 @@
...
+    tid: string,
+    ver: string,
+    preferred_username: string,
+    name: string,
+    nonce: string,
+    exp: string,
+    home_oid: string,
+    sid: string,
+    cloud_instance_host_name: string
+};
```
sameerag: +1. I like the breaking down to small classes approach.

lib/msal-common/src/auth/MsalAccount.ts
```diff
@@ -0,0 +1,88 @@
...
+     * @param idToken
+     * @param sid
+     * @param environment
+     */
+    constructor(accountIdentifier: string, homeAccountIdentifier: string, idTokenClaims: IdTokenClaims, rawIdToken: string) {
+        this.accountIdentifier = accountIdentifier;
+        this.homeAccountIdentifier = homeAccountIdentifier;
+        this.userName = idTokenClaims.preferred_username;
+        this.name = idTokenClaims.name;
+        // will be deprecated soon
```
sameerag: Can we remove the `idToken` here now instead of later?

lib/msal-common/src/auth/authority/AuthorityFactory.ts
```diff
@@ -0,0 +1,51 @@
...
+        const type = AuthorityFactory.detectAuthorityFromUrl(authorityUrl);
+        // Depending on above detection, create the right type.
+        switch (type) {
+            case AuthorityType.Aad:
+                return new AadAuthority(authorityUrl, networkInterface);
+            default:
+                throw ClientConfigurationError.createInvalidAuthorityTypeError(`Given Url: ${authorityUrl}`);
+        }
+    }
+}
```
sameerag: Two points here:
- Introduce KnownAuthorities() API here. The user can set a group of authorities during configuration and limit the session to those? We can that way get rid of 'validateAuthority`.
- Add better check for b2c, `tfp` is outdated and we have `b2clogin` now.

lib/msal-common/src/utils/Constants.ts
pkanher617: changed string constant array

lib/msal-browser/src/app/Configuration.ts
pkanher617: This will be handled within the angular library. The common library doesn't need to know about angular broadcasts.

lib/msal-common/src/utils/CryptoUtils.ts
```diff
@@ -0,0 +1,117 @@
...
+/**
+ * @hidden
+ */
+export class CryptoUtils {
+
+    /**
+     * Creates a new random GUID - used to populate state?
+     * @returns string (GUID)
+     */
+    static createNewGuid(): string {
```
sameerag: any plans to update this to UUID version 5?

lib/msal-browser/src/app/Configuration.ts
pkanher617: Rather than implementing the Angular code in common, we should emit events that wrapper libraries can hook into.

lib/msal-common/src/auth/IdToken.ts
```diff
@@ -0,0 +1,47 @@
...
+        this.rawIdToken = rawIdToken;
+        this.claims = IdToken.extractIdToken(rawIdToken, crypto) as IdTokenClaims;
+    }
+
+    /**
+     * Extract IdToken by decoding the RAWIdToken
+     *
+     * @param encodedIdToken
+     */
+    static extractIdToken(encodedIdToken: string, crypto: ICrypto): any {
```
pkanher617: return IdTokenClaims

lib/msal-common/src/auth/MsalAccount.ts
```diff
@@ -0,0 +1,88 @@
...
+ * environment             idtoken.issuer (the authority that issues the token)
+ */
+export class MsalAccount {
+
+    accountIdentifier: string;
+    homeAccountIdentifier: string;
+    userName: string;
+    name: string;
+    idToken: string;
+    idTokenClaims: StringDict;
```
pkanher617: Change to IdTokenClaims type

lib/msal-common/src/auth/MsalAccount.ts
```diff
@@ -0,0 +1,88 @@
...
+ * sid                     idToken.sid - session identifier
+ * environment             idtoken.issuer (the authority that issues the token)
+ */
+export class MsalAccount {
+
+    accountIdentifier: string;
+    homeAccountIdentifier: string;
+    userName: string;
+    name: string;
+    idToken: string;
```
pkanher617: Remove rawidToken from account object

lib/msal-common/src/auth/MsalAccount.ts
pkanher617: Will do some further research.

lib/msal-common/src/auth/authority/AuthorityFactory.ts
pkanher617: Let's have a separate PR for Authority improvements. This can include B2C, KnownAuthorities, discoverEndpoints improvements.

lib/msal-common/src/utils/CryptoUtils.ts
pkanher617: There isn't really a need for us, since version 4 gives us the pseudo-random generator we use UUIDs for. Version 5 will generate this from a hash, meaning it will generate the same UUID given the same input. We don't really need this, so we will stick with version 4.

lib/msal-common/src/response/TokenResponse.ts
```diff
@@ -28,5 +28,5 @@ export type TokenResponse = AuthResponse & {
     accessToken: string;
     refreshToken: string;
     expiresOn: Date;
-    // account: MsalAccount;
+    account: MsalAccount;
 };
```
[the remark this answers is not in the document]
pkanher617: We will review this in a separate PR.

lib/msal-common/src/auth/MsalAccount.ts
pkanher617: Done

lib/msal-common/src/auth/IdToken.ts
pkanher617: Done

lib/msal-common/src/auth/MsalAccount.ts
pkanher617: I am going have a separate PR for Auth Code which will implement Tenant Profiles

PR #1140, opened 2019-12-03
Authorization Code Flow for Single Page Applications: Browser Storage Implementation

pkanher617:
This PR is a continuation of the work for the Authorization Code Flow for the browser. It introduces the implementation of cache storage in the browser. Please review PR #1133 before this PR.

PR #1141, opened 2019-12-04
Authorization Code Flow for Single Page Applications: Browser Implementation of ICrypto

pkanher617:
This PR implements the crypto interface functions exposed by the common package in the browser package. Most functionality needs to be implemented in the browser package since the common package will need to rely on the crypto APIs available on the relevant platform. 

Please review other open PRs before this one: #1109, #1133, #1140.


lib/msal-browser/src/utils/Base64Decode.ts
```diff
@@ -0,0 +1,92 @@
...
+                            : nPart > 223 && nPart < 240 && nIdx + 2 < nLen ? /* three bytes */
+                                (nPart - 224 << 12) + (aBytes[++nIdx] - 128 << 6) + aBytes[++nIdx] - 128
+                                : nPart > 191 && nPart < 224 && nIdx + 1 < nLen ? /* two bytes */
+                                    (nPart - 192 << 6) + aBytes[++nIdx] - 128
+                                    : /* nPart < 127 ? */ /* one byte */
+                                    nPart
+            );
+        }
+        return sView;
+    }
```
DarylThayil: was this from the POC we already looked at ?

lib/msal-browser/src/utils/GuidGenerator.ts
```diff
@@ -0,0 +1,78 @@
...
+                    r |= 0x8; // set pos 3 to 1 as 1???
+                    guidResponse += hex[r];
+                } else {
+                    guidResponse += guidHolder[i];
+                }
+            }
+            return guidResponse;
+        }
+    }
+}
```
DarylThayil: would be great to be able to use a lib instead of copy these but I know we have had issues there

lib/msal-browser/src/utils/PkceGenerator.ts
```diff
@@ -0,0 +1,81 @@
...
+     */
+    private bufferToCVString(buffer: Uint8Array): string {
+        const charArr = [];
+        for (let i = 0; i < buffer.byteLength; i += 1) {
+            const index = buffer[i] % CV_CHARSET.length;
+            charArr.push(CV_CHARSET[index]);
+        }
+        return charArr.join("");
+    }
+}
```
DarylThayil: small nitpick on all these utils, when I think of utils I don't think of individual classes with private members.

lib/msal-common/src/app/config/MsalModuleConfiguration.ts
```diff
@@ -63,6 +63,11 @@ const DEFAULT_NETWORK_OPTIONS: INetworkModule = {
 };
 
 const DEFAULT_CRYPTO_IMPLEMENTATION: ICrypto = {
+    createNewGuid: (): string => {
+        const notImplErr = "Crypto interface - createNewGuid() has not been implemented";
+        console.warn(notImplErr);
```
DarylThayil: any reason to use warn here if we throw an error?

DarylThayil: also why not use logger? Maybe this is leftover from developing

lib/msal-common/src/utils/StringUtils.ts
```diff
@@ -40,4 +39,23 @@ export class StringUtils {
...
+        const pl = /\+/g;
+        const search = /([^&=]+)=([^&]*)/g;
+        const decode = (s: string) => decodeURIComponent(s.replace(pl, " "));
+        const obj: {} = {};
+        match = search.exec(query);
+        while (match) {
+            obj[decode(match[1])] = decode(match[2]);
+            match = search.exec(query);
+        }
+        return obj;
```
DarylThayil: Deserialize a string from what ?

DarylThayil: oh I see its query params

lib/msal-common/src/utils/StringUtils.ts
```diff
@@ -40,4 +39,23 @@ export class StringUtils {
     static isEmpty(str: string): boolean {
         return (typeof str === "undefined" || !str || 0 === str.length);
     }
+
+    /**
+     * deserialize a string
+     *
+     * @param query
+     */
+    static deserialize(query: string): any {
```
DarylThayil: how about queryStringToObject or something

lib/msal-browser/src/utils/Base64Decode.ts
pkanher617: Yes, I have separated the functions into different files.

lib/msal-browser/src/utils/GuidGenerator.ts
pkanher617: We have had problems in the past trying to make this work with an external library, but if anyone has any suggestions I am open to them.

lib/msal-browser/src/utils/PkceGenerator.ts
pkanher617: A lot of these private members are helpers, but I see your point. Do you have a suggestion for how I can organize the files?

lib/msal-common/src/app/config/MsalModuleConfiguration.ts
pkanher617: Logger will be implemented in a separate PR, I am using console for now to make things a little easier. I will replace the console warnings with logger errors.

lib/msal-common/src/utils/StringUtils.ts
pkanher617: Done.

lib/msal-browser/src/math/GuidGenerator.ts
```diff
@@ -0,0 +1,83 @@
...
+     * time-mid               = 2hexOctet
+     * time-high-and-version  = 2hexOctet
+     * clock-seq-and-reserved = hexOctet:
+     * clock-seq-low          = hexOctet
+     * node                   = 6hexOctet
+     * Format: xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx
+     * y could be 1000, 1001, 1010, 1011 since most significant two bits needs to be 10
+     * y values are 8, 9, A, B
+     */
+    generateGuid(): string {
```
jmckennon: Is there not a library that does all this work for us?

lib/msal-browser/src/utils/MathUtils.ts
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+/**
+ * Utility class for math specific functions
+ */
+export class MathUtils {
```
jmckennon: I know we've bit nitpicking names a lot here, but I don't feel like `MathUtils` is really great either. I think this can probably get tossed in one of the other util files. Could also consider importing a library that does this function for us

lib/msal-browser/src/math/GuidGenerator.ts
jmckennon: https://www.npmjs.com/package/uuidv4

lib/msal-browser/src/math/Base64Encode.ts
```diff
@@ -0,0 +1,138 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+/**
+ * Class which exposes APIs to encode plaintext to base64 encoded string. See here for implementation details:
+ * https://developer.mozilla.org/en-US/docs/Web/API/WindowBase64/Base64_encoding_and_decoding#Solution_2_%E2%80%93_JavaScript's_UTF-16_%3E_UTF-8_%3E_base64
+ */
+export class Base64Encode {
```
jmckennon: +1 to separation of encode/decode files

lib/msal-browser/src/error/BrowserAuthError.ts
```diff
@@ -29,4 +33,9 @@ export class BrowserAuthError extends AuthError {
     static createNoWindowObjectError(): BrowserAuthError {
         return new BrowserAuthError(BrowserAuthErrorMessage.noWindowObjectError.code, BrowserAuthErrorMessage.noWindowObjectError.desc);
     }
+
+    static createPKCENotGeneratedError(arg0: string) {
```
jmckennon: want to call this out as there are places where we have `PKCE`, `pkce`, and `Pkce`, we should probably normalize this, not only just for `PKCE` but also other acronyms

lib/msal-browser/src/utils/MathUtils.ts
pkanher617: I disagree about the MathUtils, but we can figure out naming in the future. I feel like we have been trying to go with keeping files small and targeted, so I'm kind of getting mixed signals here in terms of code design. While I am all for importing other libraries, it is important we take into consideration security and performance.

lib/msal-browser/src/error/BrowserAuthError.ts
pkanher617: I will try to make this uniform, thanks for the catch.

lib/msal-browser/src/math/GuidGenerator.ts
pkanher617: This is currently how we create guids today. I am open to using external libraries, but in the past we have implemented similar functions ourselves due to issues in downstream libraries.

pkanher617: Although this library looks pretty stable, so I have no objections to updating usage.

lib/msal-browser/src/error/BrowserAuthError.ts
pkanher617: Done.

lib/msal-browser/src/math/GuidGenerator.ts
sameerag: Depending on someone else would mean we are depended on the js/ts version support/ polyfill / es6 support etc. My recommendation would be to stick to our implementation as this is straightforward.

lib/msal-browser/src/utils/crypto/BrowserCrypto.ts
```diff
@@ -0,0 +1,48 @@
...
+    base64Decode(input: string): string {
+        return this.b64Decode.decode(input);
+    }
+
+    async generatePkceCodes(): Promise<PkceCodes> {
+        return this.pkceGenerator.generateCodes();
+    }
+
+    static getBrowserCryptoObj(): Crypto {
+        return "msCrypto" in window ? window["msCrypto"] : window.crypto;
```
jasonnutter: FYI This works for `getRandomValues`, but does not work for the other Crypto APIs. I personally think the better solution is to create abstractions for each Crypto API, then just an abstraction for the entire crypto interface.

jasonnutter:
Are there tests for all the new utility classes and methods? I think there should be before this is merged.

pkanher617:
> Are there tests for all the new utility classes and methods? I think there should be before this is merged.

These tests will be added today.

lib/msal-browser/src/utils/crypto/BrowserCrypto.ts
pkanher617: I would almost rather use the webcrypto shim than have diverging code for msCrypto and crypto. It's frustrating to get these working side by side, and we only use crypto for `getRandomValues` and `crypto.subtle.digest`.

lib/msal-browser/src/utils/MathUtils.ts
jmckennon: Sorry to confuse--yes, I agree smaller files are almost always better than shoving things into a big file. I don't necessarily have a great suggestion for the name of this file against `MathUtils`, which is part of why I suggested it may belong in one of the other files. I'm not too opinionated on what happens here, just was leaving some thoughts. I think this is fine if it's what we're working with.

lib/msal-browser/src/math/GuidGenerator.ts
jmckennon: I'm fine with leaving this how it is if you all prefer our own implementation--I generally just prefer using libraries to reduce the # of lines of code we own, like in this case.

lib/msal-browser/package.json
```diff
@@ -57,6 +57,7 @@
         "@babel/preset-typescript": "^7.7.2",
         "@babel/register": "^7.7.0",
         "@istanbuljs/nyc-config-babel": "^2.1.1",
+        "@trust/webcrypto": "^0.9.2",
```
jasonnutter: I very much disagree with this. This would also require security review.

jasonnutter: This also appears to be a Node-only package.

jasonnutter: And there's a warning on their Github not to use it.

jasonnutter: https://github.com/anvilresearch/webcrypto

pkanher617: This is only for testing

pkanher617: I need an implementation of browser crypto for mocha tests. Please provide an alternative.

jasonnutter: Ah, just figured that out. Disregard for now.

lib/msal-browser/src/crypto/BrowserCrypto.ts
```diff
@@ -0,0 +1,44 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+const HASH_ALG = "SHA-256";
+
+export class BrowserCrypto {
+
+    constructor() {
+        if (!(window["msCrypto"] || window.crypto)) {
```
jasonnutter: I think you should create a helper function for this check.

lib/msal-browser/src/crypto/CryptoOps.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export class CryptoOps implements ICrypto {
+
+    private guidGenerator: GuidGenerator;
+    private b64Encode: Base64Encode;
+    private b64Decode: Base64Decode;
+    private pkceGenerator: PkceGenerator;
+
+    constructor() {
+        const browserCrypto = new BrowserCrypto();
```
jasonnutter: Might as well also hold on to this as `this.browserCrypto`

lib/msal-browser/src/crypto/BrowserCrypto.ts
```diff
@@ -0,0 +1,44 @@
...
+            throw BrowserAuthError.createCryptoNotAvailableError("Browser crypto or msCrypto object not available.");
+        }
+    }
+
+    async sha256Digest(dataString: string): Promise<ArrayBuffer> {
+        const data = BrowserStringUtils.stringToUtf8Arr(dataString);
+
+        if (window["msCrypto"]) {
```
jasonnutter: And this check as well.

lib/msal-browser/test/mochaSetup.js
```diff
@@ -2,4 +2,6 @@ require("@babel/register")({ extensions: ['.js', '.jsx', '.ts', '.tsx'] });
 require("jsdom-global")("", {
     url: "https://localhost:8081/index.html"
 });
+window.crypto = require("@trust/webcrypto");
+window["msCrypto"] = require("@trust/webcrypto");
```
jasonnutter: I'm not sure you would actually want to do this for `msCrypto`? Since the APIs are different?

pkanher617: Is there a package I can use for this instead? I was really just going to check for existence of the API using this.

lib/msal-browser/src/utils/crypto/BrowserCrypto.ts
pkanher617: Resolved

lib/msal-browser/src/crypto/BrowserCrypto.ts
pkanher617: Done

lib/msal-browser/src/crypto/CryptoOps.ts
pkanher617: done

lib/msal-browser/src/crypto/BrowserCrypto.ts
pkanher617: done

lib/msal-browser/test/mochaSetup.js
pkanher617: Removed for now - I will add mocks and more later.

PR #1144, opened 2019-12-05
Authorization Code Flow for Single Page Applications: Implementation of the Fetch Client in the Browser

pkanher617:
This PR implements the fetch client for asynchronous network requests in the browser library for MSAL.js. This network client will be used for endpoint discovery and code-token exchange. 

Please review PR #1140 before this PR.


lib/msal-browser/src/network/FetchClient.ts
```diff
@@ -0,0 +1,21 @@
...
+
+    /**
+     * XHR client for JSON endpoints
+     * https://www.npmjs.com/package/async-promise
+     * @param url 
+     * @param requestParams 
+     * @param enableCaching 
+     */
+    async sendRequestAsync(url: string, requestParams: RequestInit, enableCaching?: boolean): Promise<any> {
+        return await fetch(url, requestParams).then(resp => {
```
jasonnutter: `fetch` is not supported in IE11. This means we need to either:

1. Provide a fallback for when its not available
2. Use a library like axios or request instead.
3. Provide a polyfill e.g. https://github.com/github/fetch
4. Ask apps to provide a polyfill if they wanted it supported (like we do for Promises).

I would personally prefer we provide the polyfill (we should do this for Promises in V2 as well).

lib/msal-browser/src/network/FetchClient.ts
```diff
@@ -0,0 +1,21 @@
...
+export class FetchClient implements INetworkModule {
+
+    /**
+     * XHR client for JSON endpoints
+     * https://www.npmjs.com/package/async-promise
+     * @param url 
+     * @param requestParams 
+     * @param enableCaching 
+     */
+    async sendRequestAsync(url: string, requestParams: RequestInit, enableCaching?: boolean): Promise<any> {
```
jasonnutter: 1. This signature feels too tied to fetch. I suppose it would be fine if we decided to use `node-fetch` for Node, for example, but I still think we should consider designing our own API, instead of just using fetch's as-is.
2. Don't think `enableCaching` needs to be an option anymore.

lib/msal-browser/src/network/FetchClient.ts
jmckennon: This is a bit unrelated to the fetch comments, but additionally here (Jason correct me if I'm wrong) I believe you never need an `await` with a return statement, as it's effectively baked in to `return` for an `async` function. I also feel like having a `.then` after this return/await kind of defeats the point of using async/await to begin with.

jasonnutter: Yeah, generally speaking its functionally equivalent to not providing it.

And yeah, might as well lean into async/await:

```js
const response = await fetch(url, requestParams);
return resonse.json();
```

lib/msal-browser/src/network/FetchClient.ts
```diff
@@ -0,0 +1,20 @@
...
+export class FetchClient implements INetworkClient {
+
+    /**
+     * XHR client for JSON endpoints
+     * https://www.npmjs.com/package/async-promise
+     * @param url 
+     * @param requestParams 
+     * @param enableCaching 
+     */
+    async sendRequestAsync(url: string, requestParams: RequestInit): Promise<any> {
```
pkanher617: Implement get and post methods instead of generic functions.

lib/msal-common/src/network/INetworkModule.ts
```diff
@@ -14,5 +14,5 @@ export interface INetworkModule {
      * @param requestParams 
      * @param enableCaching 
      */
-    sendRequestAsync(url: string, requestParams: RequestInit, enableCaching?:boolean): Promise<any>
+    sendRequestAsync(url: string, requestParams: RequestInit): Promise<any>;
```
jasonnutter: Instead of `RequestInit`, implement our own type that is platform/api agnostic.

e.g.
```js
type RequestParams {
   body?: object,
   headers?: object
}
```

lib/msal-browser/src/network/FetchClient.ts
jmckennon: +1

lib/msal-browser/src/network/XhrClient.ts
```diff
@@ -0,0 +1,68 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+import { INetworkModule } from "msal-common";
+
+export class XhrClient implements INetworkModule {
+
+    async sendGetRequestAsync(url: string, headers?: Map<string, string>, body?: string): Promise<any> {
```
jasonnutter: Can `body` be an object? And then gets stringify'd on the way out the door?

jasonnutter: Another nice touch for these APIs would be to allow generics for the return type.

```js
async sendGetRequestAsync(url: string, headers?: Map<string, string>, body?: string): Promise<T>

sendGetRequestAsync<AuthResponse>(url, headers, body)
```

lib/msal-browser/src/network/XhrClient.ts
```diff
@@ -0,0 +1,68 @@
...
+        return this.sendRequestAsync(url, "GET", headers, body);
```
jasonnutter: Can you create an enum for the request types (GET, POST, etc)

lib/msal-browser/src/network/XhrClient.ts
pkanher617: 1. Body is usually a char-delimited string, so I don't know if a object is correct? IMO it should be up to the caller what the string looks like, the XhrClient/FetchClient should just receive the string and put it in the body.

2. Are the generics casted from the response, or are they parsed in our code and we return an AuthResponse from sendGetRequestAsync?

lib/msal-browser/src/network/XhrClient.ts
pkanher617: +1

lib/msal-browser/src/network/XhrClient.ts
jasonnutter: 1. The point of doing this abstraction is that we know we will always be sending stringified objects in the body, so we should allow callers to pass objects and this function will do the stringifying.
2. The response would need to casted to the passed in generic, I believe.

pkanher617: So I think both of these improvements are putting a lot of burden on the network client to make logic decisions on input and output data. I don't think this is a good principle. The network client should receive the data it needs, format it, and give the response. The app should decide whether that response whether or not that response is an AuthResponse type. Especially if we end up opening this abstraction out to customers, are we going to expect them to know how to format the objects we are sending, and how to format the responses received by the server?

jasonnutter: I would argue that's kinda the point: to provide our own opinionated abstraction for our use cases, that make it so that our code doesn't have to do thing like stringifying objects or passing in magic strings for request methods if we don't have to.

And sorry, to be clear, this function would do the casting:

```js
jsonResponse = JSON.parse(xhr.responseText) as T;
```

pkanher617: I'd agree if it was a single private implementation that we are using, but we have decided to open this up with an option to pass a custom implementation, and I don't think that forcing the custom implementation to stringify objects and parse responses is a good practice. And there aren't any magic strings, not sure where that is coming from? Magic strings/numbers are direct values that are passed in without variable names, that is never happening here.

Currently, our OAuth request bodies are `client_id=1235...&client_secret=1234...&scopes=...&param1=value1&...`. If this changes, we will have to force all implementations to change as well. If we want to use another body format, we have to make changes to the network client API to allow for the separate body format.

I don't think casting in the way you described above is a great fit either. We get an OAuth response object back that has the hash param properties as an object. We have to perform some transformations on this object to get an AuthResponse. For example, this is from the POC:
```javascript
const response: TokenResponse = {
            uniqueId: "",
            tenantId: "",
            tokenType: acquiredTokenResponse.token_type,
            idToken: acquiredTokenResponse.id_token,
            idTokenClaims: null,
            accessToken: acquiredTokenResponse.access_token,
            refreshToken: acquiredTokenResponse.refresh_token,
            scopes: acquiredTokenResponse.scope,
            expiresOn: TimeUtils.now() + acquiredTokenResponse.expires_in,
            account: this.getAccount(),
            state: this.parseResponseState(tokenExchangeParams.state)
        };
```
where acquiredTokenResponse is the response from the Fetch request.

TLDR - I think we are introducing too many things for the network client to do, that can be abstracted to the caller, which will make future development and custom implementations of interfaces easier to manage.

pkanher617: Made some updates, please review

lib/msal-browser/src/network/FetchClient.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export class FetchClient implements INetworkModule {
+
+    /**
+     * Fetch Client for REST endpoints - Get request
+     * @param url 
+     * @param headers 
+     * @param body 
+     */
+    async sendGetRequestAsync<T>(url: string, options?: NetworkRequestOptions): Promise<T> {
```
jasonnutter: 🔥 🔥 🔥 🚀

lib/msal-browser/src/network/XhrClient.ts
```diff
@@ -0,0 +1,68 @@
...
+import { INetworkModule, NetworkRequestOptions } from "msal-common";
+
+export class XhrClient implements INetworkModule {
+
+    async sendGetRequestAsync<T>(url: string, options?: NetworkRequestOptions): Promise<T> {
+        return this.sendRequestAsync(url, "GET", options);
+    }    
+    
+    async sendPostRequestAsync<T>(url: string, options?: NetworkRequestOptions): Promise<T> {
+        return this.sendRequestAsync(url, "POST", options);
```
jasonnutter: Can we use the enum here also?

lib/msal-browser/src/network/XhrClient.ts
```diff
@@ -0,0 +1,68 @@
...
+                    reject(this.handleError(xhr.responseText));
+                }
+                let jsonResponse: T;
+                try {
+                    jsonResponse = JSON.parse(xhr.responseText) as T;
+                } catch (e) {
+                    reject(this.handleError(xhr.responseText));
+                }
+
+                resolve(jsonResponse);
```
jasonnutter: I think you can put this inside the try, then that way you can put line 26 inside the try as well.

```js
try {
  const jsonResponse = JSON.parse(xhr.responseText) as t;
  return jsonResponse;
} catch (e) {
  reject(this.handleError(xhr.responseText));
}
```

lib/msal-browser/src/network/XhrClient.ts
jasonnutter: Chatted offline, changes look great.

lib/msal-browser/src/network/XhrClient.ts
pkanher617: Updated in PR #1149.

lib/msal-browser/src/network/XhrClient.ts
pkanher617: done

Issue #845, opened 2019-07-19
Leading blank space in scope makes token caching mechanism not work


mmuarc:
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x ] Bug report  
[ x] Performance issue
[ x] Feature request
[ ] Documentation issue or request
[ ] Other... Please describe:
</code></pre>

## Browser:
- [x ] Chrome version 75.0.3770.100
- [ ] Firefox version XX
- [ ] IE version XX
- [ ] Edge version XX
- [ ] Safari version XX

## Library version

Library version: 1.0.2

## Current behavior
AB2c Server is contacted to renew a token

## Expected behavior
Token should be retrieved from the storage

## Minimal reproduction of the problem with instructions

I didn't know exactly how to tag this because I don't think it is exactly a bug, but here is what happens.

I have my scopes in a comma separated string like this:

REACT_APP_SCOPES='https://xxxx/read, https://xxxx/write'

Please notice the blank space after the comma.

When I login with msal I pass the requested scopes doing a string split getting the following:

'https://xxxx/read'
' https://xxxx/write'  (Again, notice the leading blank space)

When I try to login, I assume that somewhere in the flow the scopes are trimmed and I get back a valid access token with the following scopes.

'https://xxxx/read'
'https://xxxx/write'  (The leading blank space is gone!)

Next time I request a token, I again pass it with the blank leading space... and the code for finding a cached token does this:

`// filter by authority and scope
        for (var i = 0; i < tokenCacheItems.length; i++) {
          var cacheItem = tokenCacheItems[i];
          var cachedScopes = cacheItem.key.scopes.split(" ");
 if (Utils.containsScope(cachedScopes, scopes) && Utils.CanonicalizeUri(cacheItem.key.authority) === serverAuthenticationRequest.authority) {
            filteredItems.push(cacheItem);
          }`
So the cached scopes come like this:

[
  "https://xxxx/write",
  "https://xxxx/read"
]  (the space is gone)

but I'm passing the scopes with the leading space, so UserAgentApplication can't find the scopes because the blank space doesn't exist there.

The result is that my application was reaching the server every time for a new token due to this fault, causing it to be really slow on performance.

I understand this doesn't necessarily need to be a bug, but if when getting the access token for the first time the leading blank spaces are trimmed, they should always be trimmed. Another option would be to just refuse the login with uknown scope error due to the leading blank space.

[...]

PR #1147, opened 2019-12-06
[DO NOT MERGE] Trim user scopes before using in any operations

pkanher617:
This PR is to solve the issue raised in #845 where leading or trailing spaces are not trimmed from scope strings.

In this PR, I have changed the scope creation to trim before sending the scopes to be used in URL creation, and also trim them before sending them to the validation function. I have added test cases for this use case.

### DIFF
--- a/lib/msal-core/src/ScopeSet.ts
+++ b/lib/msal-core/src/ScopeSet.ts
@@ -16,7 +16,7 @@ export class ScopeSet {
      */
     // TODO: Rename this, intersecting scopes isn't a great name for duplicate checker
     static isIntersectingScopes(cachedScopes: Array<string>, scopes: Array<string>): boolean {
-        cachedScopes = this.convertToLowerCase(cachedScopes);
+        cachedScopes = this.convertScopesToLowerCase(cachedScopes);
         for (let i = 0; i < scopes.length; i++) {
             if (cachedScopes.indexOf(scopes[i].toLowerCase()) > -1) {
                 return true;
@@ -32,7 +32,7 @@ export class ScopeSet {
      * @param scopes
      */
     static containsScope(cachedScopes: Array<string>, scopes: Array<string>): boolean {
-        cachedScopes = this.convertToLowerCase(cachedScopes);
+        cachedScopes = this.convertScopesToLowerCase(cachedScopes);
         return scopes.every((value: any): boolean => cachedScopes.indexOf(value.toString().toLowerCase()) >= 0);
     }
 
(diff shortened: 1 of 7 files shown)


jasonnutter:
@pkanher617 Something isn't right with coverage on this branch...

pkanher617:
> @pkanher617 Something isn't right with coverage on this branch...

I had a .only in the tests, I have updated again.

lib/msal-core/test/utils/RequestUtils.spec.ts
```diff
@@ -43,6 +43,21 @@ describe("RequestUtils.ts class", () => {
         expect(emptyScopesError.stack).to.include("RequestUtils.spec.ts");
     });
 
+    it.only("Scopes with leading or trailing spaces are trimmed before validation", () => {
```
jasonnutter: Ah, this is why coverage dipped.

jasonnutter: I'll add a lint rule to prevent `it.only`

lib/msal-core/src/utils/RequestUtils.ts
```diff
@@ -46,6 +46,7 @@ export class RequestUtils {
         if(request) {
             // if extraScopesToConsent is passed in loginCall, append them to the login request; Validate and filter scopes (the validate function will throw if validation fails)
             scopes = isLoginCall ? ScopeSet.appendScopes(request.scopes, request.extraScopesToConsent) : request.scopes;
+            scopes = scopes ? scopes.map(scope => scope.trim()) : scopes;
```
jasonnutter: Could this use the helper function you defined above?

pkanher617: The helper is in ServerRequestParameters, not in RequestUtils

jasonnutter: Right, could it go in a place where both can use it?

jasonnutter: e.g. in the `ScopeSet` class

pkanher617: Made some updates, take a look.

lib/msal-core/src/ScopeSet.ts
```diff
@@ -41,19 +41,25 @@ export class ScopeSet {
...
+    static convertScopesToLowerCase(scopes: Array<string>): Array<string> {
         return scopes.map(scope => scope.toLowerCase());
     }
 
+    /**
+     * Trim scope values
+     * @param scopes 
+     */
+    static trimScopes(scopes: Array<string>): Array<string> {
+        return scopes ? scopes.map(scope => scope.trim()) : scopes;
```
jasonnutter: Could be `scopes && scopes.map(...)`.

pkanher617: Done.

PR #1149, opened 2019-12-10
Authorization Code Flow for Single Page Applications: Login URL Creation

pkanher617:
This PR updates the Authorization Code flow library to create a url for interactive login calls. This will include files for server request parameter creation, scope management, and minor updates.


lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -104,8 +105,9 @@ export class PublicClientApplication {
      * Use when initiating the login process by redirecting the user's browser to the authorization endpoint.
      * @param {@link (AuthenticationParameters:type)}
      */
-    loginRedirect(request: msalAuth.AuthenticationParameters): msalAuth.TokenResponse {
-        throw new Error("Method not implemented.");
+    async loginRedirect(request: AuthenticationParameters): Promise<TokenResponse> {
```
jasonnutter: I thought we decided `loginRedirect` would not return a promise, as it is misleading?

lib/msal-browser/src/network/FetchClient.ts
```diff
@@ -46,10 +43,10 @@ export class FetchClient implements INetworkModule {
      * @param inputHeaders 
      */
     private getFetchHeaders(inputHeaders: Map<string, string>): Headers {
+        const headers = new Headers;
```
jasonnutter: `new Headers()`

lib/msal-browser/src/network/XhrClient.ts
```diff
@@ -38,7 +38,7 @@ export class XhrClient implements INetworkModule {
             };
 
             if (method === "GET" || method === "POST") {
-                xhr.send(body);
+                reqBody ? xhr.send(reqBody) : xhr.send();
```
jasonnutter: Prefer if statement over ternary (since you aren't using the return value).

jasonnutter: Also, potentially sending a body with a GET is nonsensical, I would change this as such:

```js
if ((method === "POST"  && reqBody) {
   xhr.send(reqBody);
} else {
  xhr.send();
}
```

lib/msal-browser/src/network/FetchClient.ts
pkanher617: +1

lib/msal-browser/src/network/XhrClient.ts
pkanher617: Yeah that's way better. Will add now.

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: Currently the helpers and other functions have asynchronous steps, such as generating the PKCE codes or waiting for the endpoint discovery to complete. Since we were moving towards async/await I continued to use this pattern, which requires loginRedirect to be async. If we want to change how this works, I am ok with it, but it will mean that we are using the .then/.catch in loginRedirect instead of async/await.

jasonnutter: Ah, right. How about a private async loginRedirect method, which the public method invokes synchronously?

pkanher617: Hm ok, let me see what I can try.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -104,8 +105,9 @@ export class PublicClientApplication {
...
+        console.log(await this.authModule.createLoginUrl(request));
```
jmckennon: Does this `console.log()` need to be here?

lib/msal-browser/src/network/FetchClient.ts
```diff
@@ -29,14 +27,13 @@ export class FetchClient implements INetworkModule {
      * @param headers 
      * @param body 
      */
-    async sendPostRequestAsync(url: string, headers?: Map<string, string>, body?: string): Promise<any> {
+    async sendPostRequestAsync(url: string, headers?: Map<string, string>, reqBody?: string): Promise<any> {
```
jmckennon: since the reqBody is optional here, and we're always making a post in this function, are there are weird scenarios/edge cases that can come up because of that?

lib/msal-common/src/auth/ScopeSet.ts
```diff
@@ -0,0 +1,169 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { ClientConfigurationError } from "../error/ClientConfigurationError";
+import { StringUtils } from "../utils/StringUtils";
+
+export class ScopeSet {
```
jmckennon: question: are scopes standardized? as in, are there a common set of scopes that an app will look to gain access to, or can these basically be anything depending on the implementation?

lib/msal-common/src/cache/CacheHelpers.ts
```diff
@@ -0,0 +1,118 @@
...
+    /**
+     * @hidden
+     * @ignore
+     *
+     * Sets the cachekeys for and stores the account information in cache
+     * @param account
+     * @param state
+     * @hidden
+     */
+    setAccountCache(account: Account, state: string) {
```
jmckennon: another question: what's the use case here? is this idea of an `AccountCache` supposed to be how we're using a single cache but separating different items in the cache for different requests/apps/tabs/etc? hopefully that thought makes sense

lib/msal-common/src/server/CodeRequestParameters.ts
```diff
@@ -0,0 +1,429 @@
...
+        }
+
+        const requestUrl: string = `${authEndpoint}${str.join("&")}`;
+        return requestUrl;
+    }
+
+    /**
+     * Create a query parameter string.
+     */
+    private async createParamString(): Promise<Array<string>> {
```
jmckennon: would it be better to use template strings here instead of a bunch of `str.push()`? or do we need these different values in an array somewhere else, like for validity checks?

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: Only here for testing, this will be removed going forward

lib/msal-browser/src/network/FetchClient.ts
pkanher617: You have the option of sending a blank body in a POST.

lib/msal-common/src/auth/ScopeSet.ts
pkanher617: There are some reserved namespaces, such as openid, profile, etc., but outside of those scopes can be any string value. It is recommended to use a URI format, but this isn't a hard requirement.

lib/msal-common/src/server/CodeRequestParameters.ts
pkanher617: I think this design was used for code readability, but if you have any suggestions for how to improve it I am open to them. These parameters will have been validated by the time this function is called.

lib/msal-common/src/cache/CacheHelpers.ts
pkanher617: I will rename this, but this is just to cache the account object in our current cache. We only use a single cache today, and we currently only support single account scenarios. Multiple account scenarios may call for the design you are describing, but I don't think we have a finalized solution for it yet.

lib/msal-browser/src/network/FetchClient.ts
jmckennon: right, wanted to make sure it was a valid case, and it sounds like it is.

lib/msal-common/src/cache/CacheHelpers.ts
jmckennon: okay--makes sense. I remembered talk about multiple account scenarios, and that's why i asked.

lib/msal-common/src/server/CodeRequestParameters.ts
jmckennon: i'm not sure if others think it's better or more readable, but you'd end up concatenating template literals together to save readability. as an example: 
```
const templateLiteral = `abcdefgh` +
    `ijklmnopqrst` +
    `uvwxyz`;
```

it's really not too different, just a thought.

lib/msal-common/package.json
```diff
@@ -23,7 +23,7 @@
     ],
     "main": "./dist/index.js",
     "module": "./dist/index.es.js",
-    "types": "./dist/index.d.ts",
+    "types": "./dist/src/index.d.ts",
```
sameerag: Check if this impacts user import of `types`

lib/msal-common/src/app/module/AuthorizationCodeModule.ts
```diff
@@ -23,21 +25,47 @@ import { ClientConfigurationError } from "../../error/ClientConfigurationError";
...
+
+        // Update required cache entries for request
+        this.cacheManager.updateCacheEntries(requestParameters, request.account);
+
+        // Populate query parameters (sid/login_hint/domain_hint) and any other extraQueryParameters set by the developer
+        requestParameters.populateQueryParams();
+
+        return requestParameters.createNavigateUrl();
+    }
+
```
sameerag: +1. Cleans up a lot of legacy load. :)

lib/msal-common/src/cache/CacheHelpers.ts
```diff
@@ -0,0 +1,118 @@
...
+     * @hidden
+     * @ignore
+     */
+    updateCacheEntries(serverAuthenticationRequest: CodeRequestParameters, account: Account): void {
+        // Cache account and state
+        if (account) {
+            this.setAccountCache(account, serverAuthenticationRequest.state);
+        } else {
+            // Cache the request state
+            this.cacheStorage.setItem(TemporaryCacheKeys.REQUEST_STATE, serverAuthenticationRequest.state);
```
sameerag: Please check the latest dev branch in `msal-core` and update the temp cache with `state` for all entries. You can do this in a separate PR if you want to.

lib/msal-common/src/server/CodeRequestParameters.ts
pkanher617: I'm not sure if the readability improves with that, and I don't think we are losing anything by using the string array to build the params. interested to hear other's thoughts.

jmckennon: We'll call it fine how it is, no reason to make more work. Resolving

lib/msal-browser/src/network/XhrClient.ts
```diff
@@ -3,28 +3,30 @@
  * Licensed under the MIT License.
  */
```
DarylThayil: Why use xhr versus fetch?

pkanher617: This is only used as a backup if the fetch client is not available in the window.

DarylThayil: cool!

DarylThayil: Can we polyfill fetch so we don't have to write this? 
cc @jasonnutter  if you have any experience with this

jasonnutter: @pkanher617 and I decided against that, so that we don't have to ship a polyfill, which comes with overhead that will only be needed on a minority of browsers. Applications can provide their own `fetch` polyfill or network client implementation if they would prefer not to use built-in XHR for IE.

PR #1155, opened 2019-12-12
Authorization Code Flow for Single Page Applications: Adding Logger class

pkanher617:
This PR adds a logger class to the Authorization Code Flow version of the MSAL library. It also removes instances of console.log from the common package.

### DIFF
--- a/lib/msal-common/src/logger/Logger.ts
+++ b/lib/msal-common/src/logger/Logger.ts
@@ -58,13 +58,8 @@ export class Logger {
             return;
         }
         const timestamp = new Date().toUTCString();
-        let log: string;
-        if (!StringUtils.isEmpty(this.correlationId)) {
-            log = timestamp + ":" + this.correlationId + "-" + pkg.version + "-" + LogLevel[options.logLevel] + " " + logMessage;
-        }
-        else {
-            log = timestamp + ":" + pkg.version + "-" + LogLevel[options.logLevel] + " " + logMessage;
-        }
+        let logHeader: string = StringUtils.isEmpty(this.correlationId) ? `[${timestamp}] : ` : `[${timestamp}] : [${this.correlationId}]`;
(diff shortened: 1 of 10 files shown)


lib/msal-common/src/logger/Logger.ts
```diff
@@ -0,0 +1,171 @@
...
+     * @hidden
+     */
+    private logMessage(logMessage: string, options: LoggerMessageOptions): void {
+        if ((options.logLevel > this.level) || (!this.piiLoggingEnabled && options.containsPii)) {
+            return;
+        }
+        const timestamp = new Date().toUTCString();
+        let log: string;
+        if (!StringUtils.isEmpty(this.correlationId)) {
+            log = timestamp + ":" + this.correlationId + "-" + pkg.version + "-" + LogLevel[options.logLevel] + " " + logMessage;
```
jasonnutter: I have many thoughts on this function. Let's chat tomorrow.

jasonnutter: Can this part be changed to a ternary, with string templates instead of manual concatenation?

jasonnutter: Also, some more whitespace in teh messages would be nice.

pkanher617: will make some changes.

pkanher617: Made some changes, please review again.

lib/msal-common/src/logger/Logger.ts
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    /**
+     * @hidden
+     */
+    error(message: string, correlationId?: string): void {
+        this.logMessage(message, {
+            logLevel: LogLevel.Error,
+            containsPii: false,
+            correlationId: correlationId || ""
```
jasonnutter: Is `correlationId` used? And I would argue to have this fallback logic to empty string inside of `logMessage`, so that you don't have to do it in each log level function.

PR #1160, opened 2019-12-13
Initial upgrade of MSAL 1.2 in MSAL Angular

jasonnutter:
## 1.0.0-alpha.0

Initial upgrade to use `msal@1.2.0`.

* `msal@1.2.0` package is now a peer dependency, and must be installed alongside `@azure/msal-angular`.
* `MsalModule.forRoot` now takes two arguement.
    * The first argument is the configuration object, which is the [same `Configuration` object](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-core/src/Configuration.ts) you would pass to `msal`.
    * The second argument is a `MsalAngularConfiguration` object, containing the values for `consentScopes`, `popUp`, and `extraQueryParameters`.
    * See the [updated sample](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/918830f9750a43567f1dbc01bc6492481d267ed6/samples/MSALAngularDemoApp/src/app/app.module.ts#L41) for an example of how to pass these configuration objects.
* The `acquireToken` and `login` methods now take a single `AuthenticationParameters` object as parameters.
* `getUser()` is now `getAccount()`.
* Broadcast events now emit objects, instead of just strings.


lib/msal-angular/package.json
```diff
@@ -24,9 +24,14 @@
     "node": ">=0.8.0"
```
jasonnutter: TODO: Bump version to `1.0.0-alpha.0`

lib/msal-core/src/Configuration.ts
```diff
@@ -93,6 +94,9 @@ export type FrameworkOptions = {
     isAngular?: boolean;
     unprotectedResources?: Array<string>;
     protectedResourceMap?: Map<string, Array<string>>;
+    consentScopes?: Array<string>;
+    popUp?: boolean;
+    extraQueryParameters?: StringDict;
```
jasonnutter: TODO: Just have Angular hold these values, don't pass to core.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1956,7 +1956,7 @@ export class UserAgentApplication {
      *
      * returns the logger handle
      */
-    protected getLogger() {
+    public getLogger(): Logger {
```
jasonnutter: TODO: Don't require this change in core.

lib/msal-core/src/index.ts
```diff
@@ -2,7 +2,7 @@ export { UserAgentApplication } from "./UserAgentApplication";
 export { Logger } from "./Logger";
 export { LogLevel } from "./Logger";
 export { Account } from "./Account";
-export { Constants } from "./utils/Constants";
+export { Constants, TemporaryCacheKeys } from "./utils/Constants";
```
jasonnutter: TODO: This change is already done in Core, pull in latest dev.

lib/msal-angular/src/msal.interceptor.ts
```diff
@@ -11,56 +11,48 @@ import 'rxjs/add/operator/mergeMap'
...
+                        this.auth.clearCacheForScope(accessToken);
+                        this.broadcastService.broadcast('msal:notAuthorized', err.message);
                     }
-                    var msalError = new MSALError(JSON.stringify(err), "", JSON.stringify(scopes));
-                    this.broadcastService.broadcast('msal:notAuthorized', msalError);
-                }
-            })); //calling next.handle means we are passing control to next interceptor in chain
-        }
+
+                    throw err;
```
jasonnutter: TODO: remove this?

samples/MSALAngularDemoApp/src/app/app.component.ts
```diff
@@ -3,6 +3,7 @@ import {BroadcastService} from "@azure/msal-angular";
 import { MsalService} from "@azure/msal-angular";
 import {ProductService} from "./product/product.service";
 import {Subscription} from "rxjs/Subscription";
+import { AuthError, AuthResponse } from '../../../../lib/msal-core/lib-commonjs';
```
DarylThayil: can this work with lerna and not reaching in?

DarylThayil (approved):
lgtm in terms of what this is trying to short term accomplish

samples/MSALAngularDemoApp/src/app/app.component.ts
jasonnutter: Ah this was VS Code auto-import shenanigans, I'll fix.

lib/msal-angular/src/msal-guard.service.ts
```diff
@@ -8,89 +8,46 @@ import {MSAL_CONFIG, MsalService} from "./msal.service";
...
-                            }, function (error) {
-                                return false;
-                            });
-                    }
-                    else {
-                        this.authService.loginRedirect(this.config.consentScopes, this.config.extraQueryParameters);
-                    }
-                }
+        if (!this.authService.getAccount()) {
+            if (this.msalConfig.framework.popUp) {
```
pkanher617: minor thing, but comments would be good here.

lib/msal-angular/src/msal.interceptor.ts
pkanher617: If you remove this won't errors just get swallowed by this handler?

pkanher617 (approved):
One question and comment, otherwise looks good.

lib/msal-angular/src/msal.interceptor.ts
jasonnutter: It will get broadcast as `msal:notAuthorized`.

jasonnutter: I'll gather feedback on this, but for will keep as-is.

lib/msal-angular/src/msal-guard.service.ts
jasonnutter: Yeah, I will definitely comment this library more. Think I'll do that in a separate PR.

Issue #841, opened 2019-07-18
Where can I see which versions of Angular are supported?


jstallm:
Is there a single location to obtain this information? It appears at the surface that its up to angular 5.x, and that Angular 6+ and WIP. Can we get an official response on this please? Thank you in advance if this information is obtainable somewhere else, I just dont see it being very apparent to me, and its not obvious when looking at the 
https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev/lib/msal-angular readme at all.

sameerag:
@negoe Can we please add this in the Readme under angular?

aiwangmicrosoft:
Here is my suggestion: https://github.com/AzureAD/microsoft-authentication-library-for-js/compare/dev...aiwangmicrosoft:patch-1#diff-3b1cbf21d93b8654b77b5bf8649ebaa5

aiwangmicrosoft:
Should I create a pull request?

jmckennon:
@aiwangmicrosoft this looks fine, coordinate with @jasonnutter for other angular information supported in the next few months so we can close here.

aiwangmicrosoft:
@jasonnutter can you review my suggestion? I also updated the readme.md

PR #1162, opened 2019-12-16
Update README.md for #841

aiwangmicrosoft:
#841 added versioning table to README.md in addition to previous edit

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,12 @@
 
 The Microsoft Authentication Library for JavaScript enables client-side JavaScript web applications, running in a web browser, to authenticate users using [Azure AD](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-overview) for work and school accounts (AAD), Microsoft personal accounts (MSA), and social identity providers like Facebook, Google, LinkedIn, Microsoft accounts, etc. through [Azure AD B2C](https://docs.microsoft.com/en-us/azure/active-directory-b2c/active-directory-b2c-overview#identity-providers) service. It also enables your app to get tokens to access [Microsoft Cloud](https://www.microsoft.com/enterprise) services such as [Microsoft Graph](https://graph.microsoft.io).
 
+## Library versions and support
+
+AngularJS  |  MSAL Angular   |  MSAL.js
+-----------|-----------------|-------------
+ 4.3 to 5  | 1.0.0 (Preview) |    1.3
+
 ## Repository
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3,6 +3,12 @@
 
 The Microsoft Authentication Library for JavaScript enables client-side JavaScript web applications, running in a web browser, to authenticate users using [Azure AD](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-overview) for work and school accounts (AAD), Microsoft personal accounts (MSA), and social identity providers like Facebook, Google, LinkedIn, Microsoft accounts, etc. through [Azure AD B2C](https://docs.microsoft.com/en-us/azure/active-directory-b2c/active-directory-b2c-overview#identity-providers) service. It also enables your app to get tokens to access [Microsoft Cloud](https://www.microsoft.com/enterprise) services such as [Microsoft Graph](https://graph.microsoft.io).
 
+## Library versions and support
+
+AngularJS  |  MSAL Angular   |  MSAL.js
+-----------|-----------------|-------------
+ 4.3 to 5  | 1.0.0 (Preview) |    1.3
```
DarylThayil: @jasonnutter  do these look right to you>

jasonnutter: The original question was about which version of Angular were supported by MSAL Angular. Example:

Angular versions | Library & Version
-------------------|-----------------
AngularJS < 4  | `@azure/msal-angularjs@0.1.1`
Angular 4, 5 | `@azure/msal-angular@0.1.4`
Angular > 6 | `@azure/msal-angular@1.0.0`

jasonnutter: And I'm not sure this should go in the root README. I think we should have a dedicated section/page to framework support.

jasonnutter: I think we also need to clear articulate what our support schedule looks like for the older versions of Angular. 

For example, Angular v6 just exited LTS by the Angular team. Should we still support it? https://angular.io/guide/releases#support-policy-and-schedule

jasonnutter: @aiwangmicrosoft I'm working on the next Angular update now, so I think it would be good for us to talk through all this as were getting ready to release that.

aiwangmicrosoft: @negoe Given that I'm no longer taking over JS, does it make sense to bring in Haley or you to help review with these issues?

jasonnutter:
I will include this info in my work for Angular, so I will close this as longer needed.

PR #1164, opened 2019-12-17
Authorization Code Flow for Single Page Applications: Adding Redirect Handling Code

pkanher617:
This PR adds interaction to the Authorization Code Flow version of the MSAL library. 

The current implementation of MSAL determines the type of interaction at processing time instead of when the specific API is called. This PR introduces an IInteractionHandler abstract class which specifies which functions need to be handled when performing different methods of interaction. There will be a separate PR which extends this class for popup scenarios.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -8,5 +8,8 @@
     ],
     "eslint.quiet": true,
     "files.eol": "\n",
-    "github-pr.targetBranch": "dev"
+    "github-pr.targetBranch": "dev",
+    "editor.codeActionsOnSave": {
+        "source.fixAll.eslint": true
+    }
 }
--- a/lib/msal-browser/src/app/Configuration.ts
+++ b/lib/msal-browser/src/app/Configuration.ts
@@ -18,7 +18,7 @@ const NAVIGATE_FRAME_WAIT = 500;
 
 export type BrowserAuthOptions = AuthOptions & {
     navigateToLoginRequestUrl?: boolean;
-}
+};
 
 /**
  * Use this to configure the below cache configuration options:
--- a/lib/msal-browser/src/app/Configuration.ts
+++ b/lib/msal-browser/src/app/Configuration.ts
@@ -16,6 +16,10 @@ const FRAME_TIMEOUT = 6000;
 const OFFSET = 300;
 const NAVIGATE_FRAME_WAIT = 500;
 
+export type BrowserAuthOptions = AuthOptions & {
+    navigateToLoginRequestUrl?: boolean;
+}
+
 /**
  * Use this to configure the below cache configuration options:
  *
@@ -77,12 +81,12 @@ export type BrowserLoggerOptions = {
  * - framework: this is where you can configure the running mode of angular. More to come here soon.
  */
 export type Configuration = {
-    auth?: AuthOptions,
+    auth?: BrowserAuthOptions,
     cache?: CacheOptions,
     system?: SystemOptions
 };
 
-const DEFAULT_AUTH_OPTIONS: AuthOptions = {
+const DEFAULT_AUTH_OPTIONS: BrowserAuthOptions = {
     clientId: "",
     clientSecret: "",
     authority: null,
--- a/lib/msal-browser/src/app/PublicClientApplication.ts
+++ b/lib/msal-browser/src/app/PublicClientApplication.ts
@@ -7,6 +7,8 @@ import { AuthError, AuthResponse, AuthorizationCodeModule, AuthenticationParamet
 import { BrowserStorage } from "../cache/BrowserStorage";
 import { Configuration, buildConfiguration } from "./Configuration";
 import { CryptoOps } from "../crypto/CryptoOps";
+import { IInteractionHandler } from "../interaction_handler/IInteractionHandler";
+import { RedirectHandler } from "../interaction_handler/RedirectHandler";
 
 /**
  * A type alias for an authResponseCallback function.
@@ -32,6 +34,8 @@ export class PublicClientApplication {
 
     // auth functions imported from msal-common module
     private authModule: AuthorizationCodeModule;
+
+    private interactionHandler: IInteractionHandler;
 
     // callback for error/token response
     private authCallback: AuthCallback = null;
@@ -110,8 +114,9 @@ export class PublicClientApplication {
      * @param {@link (AuthenticationParameters:type)}
      */
     loginRedirect(request: AuthenticationParameters): void {
+        this.interactionHandler = new RedirectHandler(this.authModule);
         this.authModule.createLoginUrl(request).then((urlNavigate) => {
-            this.authModule.logger.info(urlNavigate);
+            this.interactionHandler.showUI(urlNavigate);
         });
     }
 
--- a/lib/msal-browser/src/app/PublicClientApplication.ts
+++ b/lib/msal-browser/src/app/PublicClientApplication.ts
@@ -105,16 +105,16 @@ export class PublicClientApplication {
      * or the library, depending on the origin of the error, or the AuthResponse object 
      * containing data from the server (returned with a null or non-blocking error).
      */
-    handleRedirectCallback(authCallback: AuthCallback): void {
+    async handleRedirectCallback(authCallback: AuthCallback): Promise<void> {
         if(!authCallback) {
             throw BrowserConfigurationAuthError.createInvalidCallbackObjectError(authCallback);
         }
 
         this.authCallback = authCallback;
-        const urlHash = window.location.hash;
-        if (UrlString.hashContainsKnownProperties(urlHash)) {
+        const { location: { hash }} = window;
+        if (UrlString.hashContainsKnownProperties(hash)) {
             this.interactionHandler = new RedirectHandler(this.authModule, this.browserStorage, this.authCallback);
-            this.interactionHandler.handleCodeResponse(urlHash);
+            this.interactionHandler.handleCodeResponse(hash);
         }
     }
 
--- a/lib/msal-browser/src/app/PublicClientApplication.ts
+++ b/lib/msal-browser/src/app/PublicClientApplication.ts
@@ -111,7 +111,7 @@ export class PublicClientApplication {
         }
 
         this.authCallback = authCallback;
-        const { location: { hash }} = window;
+        const { location: { hash } } = window;
         if (UrlString.hashContainsKnownProperties(hash)) {
             this.interactionHandler = new RedirectHandler(this.authModule, this.browserStorage, this.authCallback);
             this.interactionHandler.handleCodeResponse(hash);
--- a/lib/msal-browser/src/app/PublicClientApplication.ts
+++ b/lib/msal-browser/src/app/PublicClientApplication.ts
@@ -36,8 +36,6 @@ export class PublicClientApplication {
     // auth functions imported from msal-common module
     private authModule: AuthorizationCodeModule;
 
-    private interactionHandler: IInteractionHandler;
-
     // callback for error/token response
     private authCallback: AuthCallback = null;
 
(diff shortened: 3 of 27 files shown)


.vscode/settings.json
```diff
@@ -8,5 +8,8 @@
     ],
     "eslint.quiet": true,
     "files.eol": "\n",
-    "github-pr.targetBranch": "dev"
+    "github-pr.targetBranch": "dev",
+    "editor.codeActionsOnSave": {
```
DarylThayil: is this something we all want in our editors?

pkanher617: I think this was a merge from dev

pkanher617: base branch hasn't been updated

lib/msal-browser/src/interaction_handler/IInteractionHandler.ts
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+import { AuthorizationCodeModule, AuthenticationParameters } from "msal-common";
+import { BrowserStorage } from "../cache/BrowserStorage";
+
+export abstract class IInteractionHandler {
```
DarylThayil: does this abstraction hold true for interactions in general? is there carry over to popup? How will it translate to node scenarios?

pkanher617: This abstraction is specific to browsers (which is why it is defined in the browser class). After our discussions today with the Azure SDK team, this may change, but for now this is only defining interactions in a browser. Yes, there will be carryover to popup in the next PR.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -101,7 +106,16 @@ export class PublicClientApplication {
      * containing data from the server (returned with a null or non-blocking error).
      */
     handleRedirectCallback(authCallback: AuthCallback): void {
-        throw new Error("Method not implemented.");
+        if(!authCallback) {
+            throw BrowserConfigurationAuthError.createInvalidCallbackObjectError(authCallback);
+        }
+
+        this.authCallback = authCallback;
+        const urlHash = window.location.hash;
```
jasonnutter: ES6 tip: `const { location: { hash } } = window;`

lib/msal-browser/src/interaction_handler/RedirectHandler.ts
```diff
@@ -0,0 +1,80 @@
...
+        });
+    }
+
+    /**
+     * Handle authorization code response in the window.
+     * @param hash 
+     */
+    handleCodeResponse(hash: string, navigateToLoginRequestUrl?: boolean): void {
+        // retrieve the hash
+        const locationHash = hash || window.location.hash;
```
jasonnutter: Will this ever be null? Seems nonsensical.

lib/msal-browser/src/interaction_handler/RedirectHandler.ts
```diff
@@ -0,0 +1,80 @@
...
+            window.location.hash = "";
+        }
+
+        try {
+            const codeResponse = this.authModule.handleFragmentResponse(locationHash);
+            this.authModule.acquireTokenAuto(codeResponse).then((tokenResponse: TokenResponse) => {
+                this.authCallback(null, tokenResponse);
+            }).catch((err: AuthError) => {
+                this.authCallback(err, null);
+            });
```
jasonnutter: Can we do `async/await` here?

lib/msal-common/src/app/module/AuthorizationCodeModule.ts
```diff
@@ -64,17 +72,73 @@ export class AuthorizationCodeModule extends AuthModule {
...
         throw new Error("Method not implemented.");
     }
 
+    async acquireTokenAuto(codeResponse: CodeResponse): Promise<TokenResponse> {
+        if (!codeResponse || !codeResponse.code) {
+            throw ClientAuthError.createAuthCodeNullOrEmptyError();
+        }
+
+        const encodedTokenRequest = this.cacheStorage.getItem(TemporaryCacheKeys.REQUEST_PARAMS);
+        const tokenRequest = JSON.parse(this.cryptoObj.base64Decode(encodedTokenRequest)) as TokenExchangeParameters;
```
jasonnutter: This should be wrapped in a `try/catch`, as it will throw if the JSON is invalid.

lib/msal-common/src/app/module/AuthorizationCodeModule.ts
```diff
@@ -64,17 +72,73 @@ export class AuthorizationCodeModule extends AuthModule {
...
+        tokenRequest.userRequestState = codeResponse.userRequestState;
+
+        this.cacheStorage.removeItem(TemporaryCacheKeys.REQUEST_PARAMS);
+        return this.acquireToken(tokenRequest);
+    }
+
     async acquireToken(request: TokenExchangeParameters): Promise<TokenResponse> {
-        throw new Error("Method not implemented.");
+        console.log(request);
+        return null;
```
jasonnutter: Mean to commit this?

lib/msal-common/src/utils/StringUtils.ts
```diff
@@ -45,7 +45,7 @@ export class StringUtils {
      *
      * @param query
      */
-    static queryStringToObject(query: string): any {
+    static queryStringToObject<T>(query: string): T {
```
jasonnutter: Yes yes I have taught you well

lib/msal-browser/src/interaction_handler/RedirectHandler.ts
pkanher617: removed the || assignment, added null check for hash

lib/msal-browser/src/interaction_handler/RedirectHandler.ts
pkanher617: done. made handleRedirectCallbacks an async function.

lib/msal-common/src/app/module/AuthorizationCodeModule.ts
pkanher617: done.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -100,8 +105,17 @@ export class PublicClientApplication {
...
      */
-    handleRedirectCallback(authCallback: AuthCallback): void {
-        throw new Error("Method not implemented.");
+    async handleRedirectCallback(authCallback: AuthCallback): Promise<void> {
+        if(!authCallback) {
+            throw BrowserConfigurationAuthError.createInvalidCallbackObjectError(authCallback);
+        }
+
+        this.authCallback = authCallback;
+        const { location: { hash }} = window;
```
DarylThayil: nitpick
even spacing 
```js
 const { location: { hash } } window
```

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -100,8 +105,17 @@ export class PublicClientApplication {
...
+        if (UrlString.hashContainsKnownProperties(hash)) {
+            this.interactionHandler = new RedirectHandler(this.authModule, this.browserStorage, this.authCallback);
```
DarylThayil: Is there a benefit to setting the instance var here? Looks like it is not used again

lib/msal-common/src/error/ServerError.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { AuthError } from "./AuthError";
+
+/**
+ * Error thrown when there is an error with the server code, for example, unavailability.
+ */
+export class ServerError extends AuthError {
```
DarylThayil: expected that this does not have any associated errors?

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: Not sure what you mean by instance var, can you elaborate? 

For some context: This function gets called when we come back from a redirect, so we don't have any context until we understand that the hash has properties we need to handle.

lib/msal-common/src/app/module/AuthorizationCodeModule.ts
pkanher617: This function is implemented in the next PR, so I will leave it for now.

lib/msal-common/src/error/ServerError.ts
pkanher617: Error codes and strings for this class are set from the error sent by the server, either from the hash or in the response of the post request.

lib/msal-browser/src/app/PublicClientApplication.ts
DarylThayil: @jasonnutter  is this null safe ?

lib/msal-browser/src/app/PublicClientApplication.ts
DarylThayil: why do we set this.interactionHandler on PCA ?

lib/msal-common/src/response/TokenResponse.ts
```diff
@@ -22,6 +22,7 @@ import { AuthResponse } from "./AuthResponse";
 export type TokenResponse = AuthResponse & {
```
sameerag: "state" in response , is it not needed in auth_code?

samples/VanillaJSTestApp/index.html
```diff
@@ -25,7 +25,7 @@
     // initialize MSAL
     const msalConfig = {
         auth: {
-            clientId: "Enter_your_client_id",
+            clientId: "16b0c3aa-11cc-4afc-a44a-12f25e5872b6",
```
jmckennon: Two separate `clientId` here, was this meant to be committed?

lib/msal-common/src/app/module/AuthorizationCodeModule.ts
```diff
@@ -64,17 +72,75 @@ export class AuthorizationCodeModule extends AuthModule {
...
+        this.cacheStorage.setItem(TemporaryCacheKeys.REQUEST_PARAMS, this.cryptoObj.base64Encode(JSON.stringify(tokenRequest)));
+
+        return urlNavigate;
     }
 
     async createAcquireTokenUrl(request: AuthenticationParameters): Promise<string> {
         throw new Error("Method not implemented.");
     }
 
+    async acquireTokenAuto(codeResponse: CodeResponse): Promise<TokenResponse> {
```
jmckennon: Question: what's the use case here in comparison to the other `acquireToken` functions?

pkanher617: Good question! I was kicking this around a bit, and I wasn't sure how to best construct this.

Essentially, this function is used when we want to immediately want to exchange a auth code for a token. Otherwise, we return the code response and leave it up to the user to exchange for a token by calling acquireToken themselves. My question: Do we leave this up to the user, or do we just handle both legs of the flow immediately?

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: I set it rather than passing around a handle to separate functions. In the redirect scenario, a new one is created every time, but in the popup scenario, we would probably just end up using the same instance for every call, so I made it a field of the PCA class. Open to suggestions if we feel there is a better way to handle this.

lib/msal-browser/src/app/Configuration.ts
```diff
@@ -77,12 +81,12 @@ export type BrowserLoggerOptions = {
  * - framework: this is where you can configure the running mode of angular. More to come here soon.
  */
 export type Configuration = {
-    auth?: AuthOptions,
+    auth?: BrowserAuthOptions,
```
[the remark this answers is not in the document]
pkanher617: Something like navigateToLoginRequestUrl is something that wouldn't be present on a node app, so i thought it would make sense to do this. Open to suggestions if there is feedback on this.

lib/msal-common/src/response/TokenResponse.ts
pkanher617: Talked about this offline - state is part of the AuthResponse class, so it will be in the response here.

samples/VanillaJSTestApp/index.html
pkanher617: Didn't mean to commit this, will remove it.

PR #1166, opened 2019-12-17
Api sso silent

### LINKED ISSUE #900: Support Login Silent If Feasible
Explore the feasibility of supporting login silent because of the potential uses cases it can help solve for customers 

@sameerag to fill in requirements

Feature / top level api that would allow users who are already logged into to ESTS to silently get their id token.  This is especially useful if the user is logged in but msal is not aware of it.

sameerag:
Introduce an API: `ssoSilent()` as a wrapper for calling `acquireTokenSilent()` with `clientId` to acquire an `id_token`.

Fixes #900


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -560,6 +560,27 @@ export class UserAgentApplication {
...
+     * @param request
+     */
+    ssoSilent(request: AuthenticationParameters): Promise<AuthResponse> {
+        // throw an error on an empty request
+        if (!request) {
+            throw ClientConfigurationError.createEmptyRequestError();
+        }
+
+        // throw an error on no hints passed
+        if(!request.sid || !request.loginHint) {
```
jasonnutter: I thought there were valid use cases for this? e.g. guest accounts

jasonnutter: Reopening...what about apps that want to try to see if there is an existing session, but don't know about the users login hint? If we are going to tell them to call `acquireTokenSilent`, we might as well allow them to call `ssoSilent`, I would argue.

sameerag: We are not allowing them to call `acquireTokenSilent` today, we fail with "User login required". This is the same use case I pitched in the PR review and it seems to be shot down by the team. Even SPO use cases are not supported, they agreed to send `sid` and are in the process of rolling out those changes.

jasonnutter: Ah right, forgot about that.

jasonnutter: I would argue since there is already a check in `acquireTokenSilent`, you don't need this check, especially since its more complicated than just not having a `sid` or `loginHint`.

DarylThayil: The one benefit of having this is specific error is that you get disambiguation between silent and ats errors but I am fine either way

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -560,6 +560,27 @@ export class UserAgentApplication {
         });
     }
 
+    /**
+     * API interfacing idToken request when applications already have a session/hint acquired by authorization client applications
+     * @param request
+     */
+    ssoSilent(request: AuthenticationParameters): Promise<AuthResponse> {
```
DarylThayil: probably want to look at this holistically in the lib moving forward, but have you considered logging here? 

Are there things we can add at a basic level that will help us debug if there are any issues with this moving forward?

lib/msal-core/src/error/ClientConfigurationError.ts
```diff
@@ -81,6 +81,10 @@ export const ClientConfigurationErrorMessage = {
     telemetryConfigError: {
         code: "telemetry_config_error",
         desc: "Telemetry config is not configured with required values"
+    },
+    ssoSilentError: {
+        code: "sso_silent_error",
+        desc: "pass sid or login_hint for ssoSilent"
```
DarylThayil: maybe something like "request must contain either sid of login_hint"

DarylThayil:
overall good, let me know when you have looked at comments and want another review.

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: @DarylThayil That's fair, but since the check in ATS is more complicated, maybe have that factored to a helper function, so the same logic can be used here?

sameerag: Extracting a helper will not be distinguishing between `ssoSilent` and `acquireTokenSilent` calls. Do we want to go that route?

sameerag: I modified the `acquireTokenSilent` error as: `User login is required. For silent calls, request must contain either sid of login_hint`. I still think that separating error messages for these two APIs even for the same root cause is a better user experience. Let us close on this asap.

lib/msal-core/src/error/ClientConfigurationError.ts
```diff
@@ -81,6 +81,10 @@ export const ClientConfigurationErrorMessage = {
     telemetryConfigError: {
         code: "telemetry_config_error",
         desc: "Telemetry config is not configured with required values"
+    },
+    ssoSilentError: {
+        code: "sso_silent_error",
+        desc: "request must contain either sid of login_hint"
```
jasonnutter: "ssoSilent requests must contain either side or login_hint"

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: Okay, lets chat about this tomorrow afternoon.

jasonnutter: Currently, we allow not passing `login_hint` if there is already logged in user (which would silently fetch/renew the existing id token for the currently logged in user). I would argue we should still allow that, thoughts?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -580,6 +580,27 @@ export class UserAgentApplication {
...
+        // throw an error on an empty request
+        if (!request) {
+            throw ClientConfigurationError.createEmptyRequestError();
+        }
+
+        // throw an error on no hints passed
+        if (!request.sid && !request.loginHint) {
```
jasonnutter, line 594: @sameerag The original logic here was wrong, since you are checking negative and then throwing, you need `&&`, not `||`. Also added unit tests.

sameerag: Great. Thanks :)

lib/msal-core/src/error/ClientConfigurationError.ts
```diff
@@ -89,6 +89,10 @@ export const ClientConfigurationErrorMessage = {
...
```
sameerag, line 95: spell check: `or` instead of `of`

sameerag:
LGTM. I cannot approve as I created the PR but looks like you have two reviews already. Please check the minor comments before merging though.

jasonnutter:
> LGTM. I cannot approve as I created the PR but looks like you have two reviews already. Please check the minor comments before merging though.

Done.

PR #1181, opened 2019-12-29
Authorization Code Flow for Single Page Applications: Exchange Code for Token

pkanher617:
This PR adds code to perform the exchange of the authorization code for an access token.

This PR does not cover token caching or response validation, which will be part of a separate PR.

### DIFF
--- a/lib/msal-browser/src/interaction_handler/RedirectHandler.ts
+++ b/lib/msal-browser/src/interaction_handler/RedirectHandler.ts
@@ -69,7 +69,9 @@ export class RedirectHandler extends IInteractionHandler {
 
         try {
             const codeResponse = this.authModule.handleFragmentResponse(locationHash);
-            const tokenResponse: TokenResponse = await this.authModule.acquireToken(null, codeResponse);
+            const tokenResponse: TokenResponse = await this.authModule.acquireToken(null, codeResponse).catch(error => {
+                throw error;
+            });
             this.authCallback(null, tokenResponse);
         } catch (err) {
             this.authCallback(err, null);
--- a/lib/msal-browser/src/interaction_handler/RedirectHandler.ts
+++ b/lib/msal-browser/src/interaction_handler/RedirectHandler.ts
@@ -36,8 +36,6 @@ export class RedirectHandler extends IInteractionHandler {
                 this.authModule.logger.info("Navigate url is empty");
                 throw BrowserAuthError.createEmptyRedirectUriError();
             }
-        }).catch(error => {
-            throw error;
         });
     }
 
@@ -70,9 +68,7 @@ export class RedirectHandler extends IInteractionHandler {
 
         try {
             const codeResponse = this.authModule.handleFragmentResponse(locationHash);
-            const tokenResponse: TokenResponse = await this.authModule.acquireToken(null, codeResponse).catch(error => {
-                throw error;
-            });
+            const tokenResponse: TokenResponse = await this.authModule.acquireToken(null, codeResponse);
             this.authCallback(null, tokenResponse);
         } catch (err) {
             this.authCallback(err, null);
--- a/lib/msal-browser/src/interaction_handler/RedirectHandler.ts
+++ b/lib/msal-browser/src/interaction_handler/RedirectHandler.ts
@@ -32,11 +32,12 @@ export class RedirectHandler extends IInteractionHandler {
             if (urlNavigate && !StringUtils.isEmpty(urlNavigate)) {
                 this.authModule.logger.infoPii("Navigate to:" + urlNavigate);
                 window.location.assign(urlNavigate);
-            }
-            else {
+            } else {
                 this.authModule.logger.info("Navigate url is empty");
                 throw BrowserAuthError.createEmptyRedirectUriError();
             }
+        }).catch(error => {
+            throw error;
         });
     }
 
--- a/lib/msal-browser/src/interaction_handler/RedirectHandler.ts
+++ b/lib/msal-browser/src/interaction_handler/RedirectHandler.ts
@@ -72,7 +72,7 @@ export class RedirectHandler extends IInteractionHandler {
             const tokenResponse: TokenResponse = await this.authModule.acquireToken(null, codeResponse);
             this.authCallback(null, tokenResponse);
         } catch (err) {
-            this.authCallback(err as AuthError, null);
+            this.authCallback(err, null);
         }
     }
 }
--- a/lib/msal-common/src/app/module/AuthorizationCodeModule.ts
+++ b/lib/msal-common/src/app/module/AuthorizationCodeModule.ts
@@ -49,7 +49,11 @@ export class AuthorizationCodeModule extends AuthModule {
     async createLoginUrl(request: AuthenticationParameters): Promise<string> {
         // Initialize authority or use default, and perform discovery endpoint check
         const acquireTokenAuthority = (request && request.authority) ? AuthorityFactory.createInstance(request.authority, this.networkClient) : this.defaultAuthorityInstance;
-        await acquireTokenAuthority.resolveEndpointsAsync();
+        try {
+            await acquireTokenAuthority.resolveEndpointsAsync();
+        } catch (e) {
+            throw ClientAuthError.createEndpointDiscoveryIncompleteError(e);
+        }
 
         // Create and validate request parameters
         const requestParameters = new ServerCodeRequestParameters(
@@ -106,16 +110,20 @@ export class AuthorizationCodeModule extends AuthModule {
                 tokenRequest = JSON.parse(this.cryptoObj.base64Decode(encodedTokenRequest)) as TokenExchangeParameters;
                 this.cacheStorage.removeItem(TemporaryCacheKeys.REQUEST_PARAMS);
             } catch (err) {
-                throw err;
+                throw ClientAuthError.createTokenRequestCacheError(err);
             }
         } else {
             tokenRequest = request;
         }
 
         const acquireTokenAuthority = (request && request.authority) ? AuthorityFactory.createInstance(request.authority, this.networkClient) : this.defaultAuthorityInstance;
-        
+
         if (!acquireTokenAuthority.discoveryComplete()) {
-            await acquireTokenAuthority.resolveEndpointsAsync();
+            try {
+                await acquireTokenAuthority.resolveEndpointsAsync();
+            } catch (e) {
+                throw ClientAuthError.createEndpointDiscoveryIncompleteError(e);
+            }
         }
         const tokenEndpoint = acquireTokenAuthority.tokenEndpoint;
 
--- a/lib/msal-common/src/app/module/AuthorizationCodeModule.ts
+++ b/lib/msal-common/src/app/module/AuthorizationCodeModule.ts
@@ -14,15 +14,15 @@ import { TokenExchangeParameters } from "../../request/TokenExchangeParameters";
 import { TokenResponse } from "../../response/TokenResponse";
 import { ClientConfigurationError } from "../../error/ClientConfigurationError";
 import { AuthorityFactory } from "../../auth/authority/AuthorityFactory";
-import { CodeRequestParameters } from "../../server/CodeRequestParameters";
+import { ServerCodeRequestParameters } from "../../server/ServerCodeRequestParameters";
 import { CodeResponse } from "../../response/CodeResponse";
 import { UrlString } from "../../url/UrlString";
 import { ServerAuthorizationCodeResponse, validateServerAuthorizationCodeResponse } from "../../server/ServerAuthorizationCodeResponse";
-import { ClientAuthError } from "../../error/ClientAuthError";
-import { buildClientInfo } from "../../auth/ClientInfo";
-import { ServerError } from "../../error/ServerError";
+import { ClientAuthError, ClientAuthErrorMessage } from "../../error/ClientAuthError";
 import { ProtocolUtils } from "../../utils/ProtocolUtils";
 import { TemporaryCacheKeys, PersistentCacheKeys } from "../../utils/Constants";
+import { AuthError } from "../../error/AuthError";
+import { ServerTokenRequestParameters } from "../../server/ServerTokenRequestParameters";
 
 /**
  * AuthorizationCodeModule class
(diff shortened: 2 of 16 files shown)


lib/msal-common/src/app/module/AuthorizationCodeModule.ts
```diff
@@ -111,6 +111,36 @@ export class AuthorizationCodeModule extends AuthModule {
     }
 
     async acquireToken(request: TokenExchangeParameters): Promise<TokenResponse> {
+        const acquireTokenAuthority = (request && request.authority) ? AuthorityFactory.createInstance(request.authority, this.networkClient) : this.defaultAuthorityInstance;
+
+        let tokenEndpoint: string;
+        try {
+            tokenEndpoint = acquireTokenAuthority.tokenEndpoint;
```
DarylThayil: if there is a default authority why does this throw?

lib/msal-common/src/app/module/AuthorizationCodeModule.ts
```diff
@@ -111,6 +111,36 @@ export class AuthorizationCodeModule extends AuthModule {
...
+        } catch (e) {
+            const authErr: AuthError = e;
+            if (authErr.errorCode === ClientAuthErrorMessage.endpointResolutionError.code) {
+                await acquireTokenAuthority.resolveEndpointsAsync();
```
DarylThayil: If this is not sucessfull what happens?

lib/msal-common/src/auth/authority/Authority.ts
```diff
@@ -108,10 +108,8 @@ export abstract class Authority {
 
     public abstract async getOpenIdConfigurationAsync(): Promise<string>;
 
-    public async resolveEndpointsAsync(): Promise<ITenantDiscoveryResponse> {
+    public async resolveEndpointsAsync(): Promise<void> {
         const openIdConfigEndpoint = await this.getOpenIdConfigurationAsync();
         this.tenantDiscoveryResponse = await this.discoverEndpoints(openIdConfigEndpoint);
```
DarylThayil: what if these are not sucessful?

lib/msal-common/src/server/ServerRequestParameters.ts
```diff
@@ -0,0 +1,50 @@
...
+        this.xClientSku = Constants.LIBRARY_NAME;
+        this.xClientVer = pkg.version;
+    }
+
+    protected abstract createParamString(): Promise<Array<string>>;
+    
+    /**
+     * Replace client id with the default scopes used for token acquisition.
+     */
+    protected replaceDefaultScopes() {
```
jmckennon: What's the use case for this? Shouldn't our default scopes... just work to begin with? Or does a second call require different scopes?

lib/msal-common/src/app/module/AuthorizationCodeModule.ts
pkanher617: When we come back from a redirect, the default authority does not resolveEndpoints automatically, so we need to check if we still have the endpoint available or we need to resolve again.

lib/msal-common/src/app/module/AuthorizationCodeModule.ts
pkanher617: Same as what would happen if any resolve Endpoints fails, we receive an error response. We should probably handle this ourselves, but I'm not completely clear on all the error cases. I think we should revisit this when we talk through design for optimization of well-known endpoint.

lib/msal-common/src/auth/authority/Authority.ts
pkanher617: Currently will throw an error and fail out. I think this is expected behavior, the only way this would fail is:
1. no network,
2. server unavailable,
3. endpoint configuration object has changed

pkanher617: open for ideas on how to improve this.

lib/msal-common/src/app/module/AuthorizationCodeModule.ts
jasonnutter: So `createInstance` can return `null`?

jasonnutter: Would be simpler in terms of syntax to check the truthiness of `acquireTokenAuthority`, then to do a `try/catch`.

lib/msal-common/src/app/module/AuthorizationCodeModule.ts
```diff
@@ -111,6 +111,36 @@ export class AuthorizationCodeModule extends AuthModule {
...
 
     async acquireToken(request: TokenExchangeParameters): Promise<TokenResponse> {
+        const acquireTokenAuthority = (request && request.authority) ? AuthorityFactory.createInstance(request.authority, this.networkClient) : this.defaultAuthorityInstance;
+
+        let tokenEndpoint: string;
+        try {
+            tokenEndpoint = acquireTokenAuthority.tokenEndpoint;
+        } catch (e) {
+            const authErr: AuthError = e;
+            if (authErr.errorCode === ClientAuthErrorMessage.endpointResolutionError.code) {
```
jasonnutter: How could ` acquireTokenAuthority.tokenEndpoint` throw this error?

lib/msal-common/src/server/ServerRequestParameters.ts
pkanher617: if the client id is present in the list of scopes, we remove it and add the openid and profile scopes to retrieve the id token. If it is only an access token (so no clientid in list of scopes), we don't need to add openid and profile. We ALWAYS need the offline_access scope in order to get the refresh token for SPAs.

lib/msal-common/src/server/ServerTokenRequestParameters.ts
```diff
@@ -0,0 +1,57 @@
...
+        headers.set(HEADER_NAMES.CONTENT_TYPE, Constants.URL_FORM_CONTENT_TYPE);
+        return headers;
+    }
+
+    async createRequestBody(): Promise<string> {
+        const paramString = await this.createParamString();
+        return paramString.join("&");
+    }
+
+    protected async createParamString(): Promise<Array<string>> {
```
jasonnutter: Why is this `async`? Seems synchronous.

lib/msal-common/src/app/module/AuthorizationCodeModule.ts
pkanher617: `createInstance` won't return null. Essentially authority creation has two steps:

1. Create the authority object with no data for oauth endpoints (i.e. /authorize, /token, etc,).
2. Resolve the /.well-known oauth endpoint to retrieve required data for authority.

Checking the truthiness of acquireTokenAuthority doesn't tell us whether we have done (2). there is a private property in the authority class which DOES denote this, so I can change this code to check that parameter, but it essentially accomplishes the same thing.

lib/msal-common/src/app/module/AuthorizationCodeModule.ts
pkanher617: if you review the code, the getter for the tokenEndpoint checks whether endpoint discovery is complete, and throws an error if it isn't. I can change this so that we check whether endpoint discovery is done before calling `acquireTokenAuthority.tokenEndpoint`.

lib/msal-common/src/app/module/AuthorizationCodeModule.ts
jasonnutter: Oh, I see the getter below for `tokenEndpoint`.

lib/msal-common/src/server/ServerTokenRequestParameters.ts
pkanher617: It is synchronous for `ServerTokenRequestParameters` but asynchronous for `ServerCodeRequestParameters`, and since they both extend the same base class, they are both synchronous. I suppose this doesn't need to be defined in the base class, but since both child classes use it I figured it would be better to define it there.

lib/msal-common/src/app/module/AuthorizationCodeModule.ts
jasonnutter: Yeah, missed that initially. Yeah, you would know when calling `createInstance`, correct? That would seem like a more intuitive place to throw.

lib/msal-common/src/auth/authority/Authority.ts
sameerag: How do we distinguish between the above errors? Can we?

lib/msal-common/src/server/ServerTokenRequestParameters.ts
sameerag: Does making the base class async affect the performance of the `ServerTokenRequestParameters` which can be synchronous?

samples/VanillaJSTestApp/index_authCode.html
```diff
@@ -20,7 +20,7 @@
     // initialize MSAL
     const msalConfig = {
         auth: {
-            clientId: "Enter_your_client_id",
+            clientId: "16b0c3aa-11cc-4afc-a44a-12f25e5872b6",
```
sameerag: intentional?

lib/msal-common/src/app/module/AuthorizationCodeModule.ts
```diff
@@ -95,17 +101,48 @@ export class AuthorizationCodeModule extends AuthModule {
...
         if (!request) {
             encodedTokenRequest = this.cacheStorage.getItem(TemporaryCacheKeys.REQUEST_PARAMS);
+            try {
+                tokenRequest = JSON.parse(this.cryptoObj.base64Decode(encodedTokenRequest)) as TokenExchangeParameters;
+                this.cacheStorage.removeItem(TemporaryCacheKeys.REQUEST_PARAMS);
+            } catch (err) {
+                throw ClientAuthError.createTokenRequestCacheError(err);
+            }
+        } else {
+            tokenRequest = request;
```
jasonnutter: Alternative, which doesn't use mutation and a bit easier to read (IMO):

```js
const getCachedRequest = () => {
  try {
    const encodedTokenRequest = this.cacheStorage.getItem(TemporaryCacheKeys.REQUEST_PARAMS);
    const parsedRequest = JSON.parse(this.cryptoObj.base64Decode(encodedTokenRequest)) as TokenExchangeParameters;
    return parseRequest;
  } catch (err) {
     throw ClientAuthError.createTokenRequestCacheError(err);
  }
};
const tokenRequest = request || getCachedRequest();
```

lib/msal-browser/src/interaction_handler/RedirectHandler.ts
```diff
@@ -69,10 +70,12 @@ export class RedirectHandler extends IInteractionHandler {
 
         try {
             const codeResponse = this.authModule.handleFragmentResponse(locationHash);
-            const tokenResponse: TokenResponse = await this.authModule.acquireToken(null, codeResponse);
+            const tokenResponse: TokenResponse = await this.authModule.acquireToken(null, codeResponse).catch(error => {
+                throw error;
```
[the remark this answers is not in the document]
pkanher617: left over from some testing. thanks for catching it.

lib/msal-common/src/app/module/AuthorizationCodeModule.ts
```diff
@@ -95,17 +101,48 @@ export class AuthorizationCodeModule extends AuthModule {
...
+        const acquireTokenAuthority = (request && request.authority) ? AuthorityFactory.createInstance(request.authority, this.networkClient) : this.defaultAuthorityInstance;
+
+        if (!acquireTokenAuthority.discoveryComplete()) {
+            try {
+                await acquireTokenAuthority.resolveEndpointsAsync();
+            } catch (e) {
+                throw ClientAuthError.createEndpointDiscoveryIncompleteError(e);
+            }
         }
+        const tokenEndpoint = acquireTokenAuthority.tokenEndpoint;
```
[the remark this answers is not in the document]
pkanher617: done.

lib/msal-common/src/server/ServerTokenRequestParameters.ts
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    createRequestBody(): string {
+        const paramString = this.createParamString();
+        return paramString.join("&");
+    }
+
+    protected createParamString(): Array<string> {
+        const str: Array<string> = [];
+        this.replaceDefaultScopes();
```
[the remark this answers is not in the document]
pkanher617: I can add this as a function to ScopeSet, but I don't think we want to have it happen immediately, since we want to track whether client id was sent or not.

lib/msal-common/src/server/ServerTokenRequestParameters.ts
pkanher617: made it synchronous

lib/msal-common/src/app/module/AuthorizationCodeModule.ts
pkanher617: made changes

lib/msal-common/src/server/ServerTokenRequestParameters.ts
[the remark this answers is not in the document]
pkanher617: changed

lib/msal-common/src/app/module/AuthorizationCodeModule.ts
pkanher617: Done - made a getCachedRequest helper function

samples/VanillaJSTestApp/index_authCode.html
pkanher617: removed

lib/msal-common/src/auth/authority/Authority.ts
```diff
@@ -56,31 +56,31 @@ export abstract class Authority {
...
     }
 
     public get endSessionEndpoint(): string {
-        if(this.discoveryComplete) {
+        if(this.discoveryComplete()) {
             return this.tenantDiscoveryResponse.end_session_endpoint.replace("{tenant}", this.tenant);
         } else {
             throw ClientAuthError.createEndpointDiscoveryIncompleteError();
         }
     }
```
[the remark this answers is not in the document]
pkanher617: done

lib/msal-common/src/app/module/AuthorizationCodeModule.ts
pkanher617: resolving until we make changes for endpoint discovery

lib/msal-common/src/auth/authority/Authority.ts
pkanher617: resolved offline. We can distinguish by seeing where the error originates:
1. from fetch
2. from server
3. Our library

lib/msal-common/src/app/module/AuthorizationCodeModule.ts
pkanher617: we should resolve the endpoint discovery as a separate issue, this is still an open question in the current library.

PR #1183, opened 2019-12-30
Authorization Code Flow for Single Page Applications: Response Handling Code

pkanher617:
This PR adds the response handling to the authorization code flow for single page applications.

This PR does NOT cover token caching or cache cleanup, those will be addressed in the next immediate PR. Please review PRs #1164 and #1181 before reviewing this PR.

### DIFF
--- a/lib/msal-common/src/app/module/AuthorizationCodeModule.ts
+++ b/lib/msal-common/src/app/module/AuthorizationCodeModule.ts
@@ -183,7 +183,7 @@ export class AuthorizationCodeModule extends AuthModule {
     }
 
     private createTokenResponse(serverTokenResponse: ServerAuthorizationTokenResponse, state: string): TokenResponse {
-        const tokenResponse: TokenResponse = {
+        let tokenResponse: TokenResponse = {
             uniqueId: "",
             tenantId: "",
             tokenType: "",
@@ -202,13 +202,13 @@ export class AuthorizationCodeModule extends AuthModule {
 
         // Retrieve current id token object
         let idTokenObj: IdToken;
-        const cachedIdToken: IdToken = new IdToken(this.cacheStorage.getItem(PersistentCacheKeys.ID_TOKEN), this.cryptoObj);
+        const cachedIdToken: string = this.cacheStorage.getItem(PersistentCacheKeys.ID_TOKEN);
         if (serverTokenResponse.id_token) {
             idTokenObj = new IdToken(serverTokenResponse.id_token, this.cryptoObj);
-            setResponseIdToken(tokenResponse, idTokenObj);
+            tokenResponse = setResponseIdToken(tokenResponse, idTokenObj);
         } else if (cachedIdToken) {
-            idTokenObj = cachedIdToken;
-            setResponseIdToken(tokenResponse, idTokenObj);
+            idTokenObj = new IdToken(cachedIdToken, this.cryptoObj);
+            tokenResponse = setResponseIdToken(tokenResponse, idTokenObj);
         } else {
             // TODO: No account scenario?
         }
@@ -216,7 +216,7 @@ export class AuthorizationCodeModule extends AuthModule {
         // check nonce integrity if idToken has nonce - throw an error if not matched
         const nonce = this.cacheStorage.getItem(`${TemporaryCacheKeys.NONCE_IDTOKEN}|${state}`);
 
-        if (!idTokenObj || idTokenObj.claims.nonce) {
+        if (!idTokenObj || !idTokenObj.claims.nonce) {
             throw ClientAuthError.createInvalidIdTokenError(idTokenObj);
         }
 
--- a/lib/msal-common/src/app/module/AuthorizationCodeModule.ts
+++ b/lib/msal-common/src/app/module/AuthorizationCodeModule.ts
@@ -11,23 +11,18 @@ import { PublicClientSPAConfiguration, buildPublicClientSPAConfiguration } from 
 import { AuthenticationParameters } from "../../request/AuthenticationParameters";
 import { TokenExchangeParameters } from "../../request/TokenExchangeParameters";
 // response
-import { TokenResponse, setResponseIdToken } from "../../response/TokenResponse";
+import { TokenResponse } from "../../response/TokenResponse";
 import { ClientConfigurationError } from "../../error/ClientConfigurationError";
 import { AuthorityFactory } from "../../auth/authority/AuthorityFactory";
 import { ServerCodeRequestParameters } from "../../server/ServerCodeRequestParameters";
 import { CodeResponse } from "../../response/CodeResponse";
 import { UrlString } from "../../url/UrlString";
 import { ServerAuthorizationCodeResponse, validateServerAuthorizationCodeResponse } from "../../server/ServerAuthorizationCodeResponse";
-import { ClientAuthError, ClientAuthErrorMessage } from "../../error/ClientAuthError";
-import { ProtocolUtils } from "../../utils/ProtocolUtils";
+import { ClientAuthError } from "../../error/ClientAuthError";
 import { TemporaryCacheKeys, PersistentCacheKeys } from "../../utils/Constants";
-import { AuthError } from "../../error/AuthError";
 import { ServerTokenRequestParameters } from "../../server/ServerTokenRequestParameters";
 import { ServerAuthorizationTokenResponse, validateServerAuthorizationTokenResponse } from "../../server/ServerAuthorizationTokenResponse";
-import { IdToken } from "../../auth/IdToken";
-import { buildClientInfo } from "../../auth/ClientInfo";
-import { Account } from "../../auth/Account";
-import { ScopeSet } from "../../auth/ScopeSet";
+import { ResponseHandler } from "../../response/ResponseHandler";
 
 /**
  * AuthorizationCodeModule class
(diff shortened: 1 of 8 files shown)


lib/msal-common/src/app/module/AuthorizationCodeModule.ts
```diff
@@ -163,12 +176,91 @@ export class AuthorizationCodeModule extends AuthModule {
...
         const response: CodeResponse = {
             code: hashParams.code,
-            userRequestState: ProtocolUtils.getUserRequestState(hashParams.state)
+            userRequestState: hashParams.state
         };
 
         return response;
     }
 
+    private createTokenResponse(serverTokenResponse: ServerAuthorizationTokenResponse, state: string): TokenResponse {
```
DarylThayil: Should Token response just be a class that takes ServerAuthorizationTokenResponse as a constructor ?

lib/msal-common/src/cache/CacheHelpers.ts
```diff
@@ -76,12 +76,15 @@ export class CacheHelpers {
...
         if (account) {
-            this.setAccountCache(account, serverAuthenticationRequest.state);
+            this.setAccountCache(account);
         }
 
         // Cache the request state
         this.cacheStorage.setItem(TemporaryCacheKeys.REQUEST_STATE, serverAuthenticationRequest.state);
 
+        // Cache the nonce
+        this.cacheStorage.setItem(`${TemporaryCacheKeys.NONCE_IDTOKEN}|${serverAuthenticationRequest.state}`, serverAuthenticationRequest.nonce);
```
DarylThayil: should this key have a generator function as well?

lib/msal-common/src/server/ServerAuthorizationTokenResponse.ts
```diff
@@ -0,0 +1,35 @@
...
+ */
+import { ServerError } from "../error/ServerError";
+
+/**
+ * Deserialized response object from server authorization code request.
+ * - 
+ */
+export type ServerAuthorizationTokenResponse = {
+    // Success
+    token_type?: string;
```
jmckennon: Should these be non-optional? Or at least one/some non-optional?

lib/msal-common/src/app/module/AuthorizationCodeModule.ts
pkanher617: We return the `TokenResponse` type to the user (i.e. what we return today as the `AuthResponse` type), so I can make a separate class that does what you describe, just was a little lost around naming.

lib/msal-common/src/cache/CacheHelpers.ts
pkanher617: That may be useful, I will add that.

lib/msal-common/src/server/ServerAuthorizationTokenResponse.ts
pkanher617: Is there a way to have one or the other?

We will either get:
```
token_type: string;
scope: string;
expires_in: number;
ext_expires_in: number;
access_token: string;
refresh_token: string;
id_token: string;
```
in case of success, or 
```
error: string;
error_description: string;
error_codes: Array<string>;
timestamp: string;
trace_id: string;
correlation_id: string;
```
But I'm not sure how to have typescript set it to one or the other.

jasonnutter: I think you would have to have two separate types for that.

pkanher617: Resolved offline.

lib/msal-common/src/app/module/AuthorizationCodeModule.ts
pkanher617: done, created a response handler class

lib/msal-common/src/cache/CacheHelpers.ts
pkanher617: done.

lib/msal-common/src/error/ClientAuthError.ts
```diff
@@ -48,6 +49,18 @@ export const ClientAuthErrorMessage = {
         code: "state_mismatch",
         desc: "State mismatch error. Please check your network. Continued requests may cause cache overflow."
     },
+    nonceMismatchError: {
+        code: "nonce_mismatch",
+        desc: "Nonce mismatch error. Please check whether concurrent requests are causing this issue."
```
jasonnutter: This doesn't seem as helpful as it could be. We should support concurrent requests without issue, is there some particular  nuance this makes this happen?

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    public setResponseIdToken(originalResponse: TokenResponse, idTokenObj: IdToken) : TokenResponse {
+        if (!originalResponse) {
+            return null;
+        } else if (!idTokenObj) {
+            return originalResponse;
+        }
+    
+        const exp = Number(idTokenObj.claims.exp);
```
jasonnutter: Can this something like `expiresSeconds`? Make it clear that its the expiration and that it is in seconds.

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -0,0 +1,131 @@
...
+        let idTokenObj: IdToken;
+        const cachedIdToken: string = this.cacheStorage.getItem(PersistentCacheKeys.ID_TOKEN);
+        if (serverTokenResponse.id_token) {
+            idTokenObj = new IdToken(serverTokenResponse.id_token, this.cryptoObj);
+            tokenResponse = this.setResponseIdToken(tokenResponse, idTokenObj);
+        } else if (cachedIdToken) {
+            idTokenObj = new IdToken(cachedIdToken, this.cryptoObj);
+            tokenResponse = this.setResponseIdToken(tokenResponse, idTokenObj);
+        } else {
+            // TODO: No account scenario?
```
jasonnutter: Throw exception maybe?

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -0,0 +1,131 @@
...
+        tokenResponse.account = Account.createAccount(idTokenObj, clientInfo, this.cryptoObj);
+    
+        // Set token type
+        tokenResponse.tokenType = serverTokenResponse.token_type;
+    
+        // Save the access token if it exists
+        if (serverTokenResponse.access_token) {
+            const accountKey = this.cacheManager.generateAcquireTokenAccountKey(tokenResponse.account.homeAccountIdentifier);
+            
+            const cachedAccount = JSON.parse(this.cacheStorage.getItem(accountKey)) as Account;
```
jasonnutter: What happens if this throws?

lib/msal-common/src/response/ResponseHandler.ts
pkanher617: Changed.

lib/msal-common/src/response/ResponseHandler.ts
pkanher617: Need to do more investigation, i will tackle this in a separate PR.

lib/msal-common/src/response/ResponseHandler.ts
pkanher617: Fixed in #1185.

lib/msal-common/src/error/ClientAuthError.ts
pkanher617: "May be caused by a race condition concurrent requests"

lib/msal-common/src/app/module/AuthorizationCodeModule.ts
```diff
@@ -123,15 +123,23 @@ export class AuthorizationCodeModule extends AuthModule {
...
             }
         );
 
-        return null;
+        try {
+            validateServerAuthorizationTokenResponse(acquiredTokenResponse);
+            const responseHandler = new ResponseHandler(this.clientConfig.auth.clientId, this.cacheStorage, this.cacheManager, this.cryptoObj);
+            return responseHandler.createTokenResponse(acquiredTokenResponse, tokenReqParams.state);
+        } catch (e) {
+            this.cacheManager.resetTempCacheItems(tokenReqParams.state);
```
sameerag: nitpick: can we merge all error unroll (account = null and cache reset) to a helper API if used at multiple places?

lib/msal-common/src/server/ServerAuthorizationTokenResponse.ts
```diff
@@ -0,0 +1,35 @@
...
+    correlation_id?: string;
+};
+
+export function validateServerAuthorizationTokenResponse(serverResponse: ServerAuthorizationTokenResponse): void {
+    // Check for error
+    if (serverResponse.error || serverResponse.error_description) {
+        const errString = `${serverResponse.error_codes} - [${serverResponse.timestamp}]: ${serverResponse.error_description} - Correlation ID: ${serverResponse.correlation_id} - Trace ID: ${serverResponse.trace_id}`;
+        throw new ServerError(serverResponse.error, errString);
+    }
+}
```
sameerag: Later PR work: Should we have separate helper functions for idToken and accessToken processing for readability?

lib/msal-common/src/app/module/AuthorizationCodeModule.ts
pkanher617: this wouldn't change anything - resetTempCacheItems is the only thing that needs to be called. We can review at the end if this API would add anything, but as it is now it doesn't add any value.

lib/msal-common/src/server/ServerAuthorizationTokenResponse.ts
pkanher617: We process and cache them together, so I don't know that it is required, but I am happy to do this in a later PR.

lib/msal-common/src/error/ClientAuthError.ts
pkanher617: Fixed.

PR #1185, opened 2020-01-03
Authorization Code Flow for Single Page Applications: Token Caching and Cleanup

pkanher617:
This PR add token caching and cache cleanup after requests to the authorization code flow for single page applications. 

Please review PRs #1181 and #1183 before this PR.

### DIFF
--- a/lib/msal-browser/src/cache/BrowserStorage.ts
+++ b/lib/msal-browser/src/cache/BrowserStorage.ts
@@ -132,7 +132,6 @@ export class BrowserStorage implements ICacheStorage {
     }
     
     getKeys(): string[] {
-        console.log("Cache Keys: " + JSON.stringify(Object.keys(this.windowStorage)));
         return Object.keys(this.windowStorage);
     }
 
--- a/lib/msal-browser/src/cache/BrowserStorage.ts
+++ b/lib/msal-browser/src/cache/BrowserStorage.ts
@@ -16,7 +16,6 @@ export class BrowserStorage implements ICacheStorage {
     private windowStorage: Storage;
 
     private clientId: string;
-    private rollbackEnabled: boolean;
 
     constructor(clientId: string, cacheConfig: CacheOptions) {
         this.validateWindowStorage(cacheConfig.cacheLocation);
@@ -24,9 +23,6 @@ export class BrowserStorage implements ICacheStorage {
         this.cacheConfig = cacheConfig;
         this.windowStorage = window[this.cacheConfig.cacheLocation];
         this.clientId = clientId;
-
-        // This is hardcoded to true for now, we will roll this out as a configurable option in the future.
-        this.rollbackEnabled = true;
 
         this.migrateCacheEntries();
     }
@@ -51,7 +47,6 @@ export class BrowserStorage implements ICacheStorage {
      * @param storeAuthStateInCookie
      */
     private migrateCacheEntries() {
-
         const idTokenKey = `${Constants.CACHE_PREFIX}.${PersistentCacheKeys.ID_TOKEN}`;
         const clientInfoKey = `${Constants.CACHE_PREFIX}.${PersistentCacheKeys.CLIENT_INFO}`;
         const errorKey = `${Constants.CACHE_PREFIX}.${ErrorCacheKeys.ERROR}`;
@@ -85,7 +80,7 @@ export class BrowserStorage implements ICacheStorage {
      * @param key
      * @param addInstanceId
      */
-    private generateCacheKey(key: string, addInstanceId: boolean): string {
+    private generateCacheKey(key: string): string {
         try {
             // Defined schemas do not need the key appended
             this.validateObjectKey(key);
@@ -94,7 +89,7 @@ export class BrowserStorage implements ICacheStorage {
             if (key.startsWith(`${Constants.CACHE_PREFIX}`) || key.startsWith(PersistentCacheKeys.ADAL_ID_TOKEN)) {
                 return key;
             }
-            return addInstanceId ? `${Constants.CACHE_PREFIX}.${this.clientId}.${key}` : `${Constants.CACHE_PREFIX}.${key}`;
+            return `${Constants.CACHE_PREFIX}.${this.clientId}.${key}`;
         }
     }
 
@@ -107,18 +102,15 @@ export class BrowserStorage implements ICacheStorage {
     }
 
     setItem(key: string, value: string): void {
-        const msalKey = this.generateCacheKey(key, true);
+        const msalKey = this.generateCacheKey(key);
         this.windowStorage.setItem(msalKey, value);
-        if (this.rollbackEnabled) {
-            this.windowStorage.setItem(this.generateCacheKey(key, false), value);
-        }
         if (this.cacheConfig.storeAuthStateInCookie) {
             this.setItemCookie(msalKey, value);
         }
     }
     
     getItem(key: string): string {
-        const msalKey = this.generateCacheKey(key, true);
+        const msalKey = this.generateCacheKey(key);
         const itemCookie = this.getItemCookie(msalKey);
         if (this.cacheConfig.storeAuthStateInCookie && itemCookie) {
             return itemCookie;
@@ -127,22 +119,20 @@ export class BrowserStorage implements ICacheStorage {
     }
     
     removeItem(key: string): void {
-        const msalKey = this.generateCacheKey(key, true);
+        const msalKey = this.generateCacheKey(key);
         this.windowStorage.removeItem(msalKey);
-        if (this.rollbackEnabled) {
-            this.windowStorage.removeItem(this.generateCacheKey(key, false));
-        }
         if (this.cacheConfig.storeAuthStateInCookie) {
             this.clearItemCookie(msalKey);
         }
     }
     
     containsKey(key: string): boolean {
-        const msalKey = this.generateCacheKey(key, true);
+        const msalKey = this.generateCacheKey(key);
         return this.windowStorage.hasOwnProperty(msalKey) || this.windowStorage.hasOwnProperty(key);
     }
     
     getKeys(): string[] {
+        console.log("Cache Keys: " + JSON.stringify(Object.keys(this.windowStorage)));
         return Object.keys(this.windowStorage);
     }
 
--- a/lib/msal-browser/src/utils/BrowserConstants.ts
+++ b/lib/msal-browser/src/utils/BrowserConstants.ts
@@ -9,7 +9,9 @@
  */
 export const BrowserConstants = {
     CACHE_LOCATION_LOCAL: "localStorage",
-    CACHE_LOCATION_SESSION: "sessionStorage"
+    CACHE_LOCATION_SESSION: "sessionStorage",
+    INTERACTION_STATUS_KEY: "interaction.status",
+    INTERACTION_IN_PROGRESS: "interaction_in_progress"
 };
 
 export enum HTTP_REQUEST_TYPE {
--- a/lib/msal-common/src/cache/AccessTokenKey.ts
+++ b/lib/msal-common/src/cache/AccessTokenKey.ts
@@ -1,0 +1,26 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { ICrypto } from "../crypto/ICrypto";
+import { UrlString } from "../url/UrlString";
+
+/**
+ * @hidden
+ */
+export class AccessTokenKey {
+
+    authority: string;
+    clientId: string;
+    scopes: string;
+    homeAccountIdentifier: string;
+
+    constructor(authority: string, clientId: string, scopes: string, uid: string, utid: string, cryptoObj: ICrypto) {
+        const authorityUri = new UrlString(authority);
+        this.authority = authorityUri.urlString;
+        this.clientId = clientId;
+        this.scopes = scopes;
+        this.homeAccountIdentifier = cryptoObj.base64Encode(uid) + "." + cryptoObj.base64Encode(utid);
+    }
+}
--- a/lib/msal-common/src/cache/AccessTokenKey.ts
+++ b/lib/msal-common/src/cache/AccessTokenKey.ts
@@ -21,6 +21,6 @@ export class AccessTokenKey {
         this.authority = authorityUri.urlString;
         this.clientId = clientId;
         this.scopes = scopes;
-        this.homeAccountIdentifier = cryptoObj.base64Encode(uid) + "." + cryptoObj.base64Encode(utid);
+        this.homeAccountIdentifier = `${cryptoObj.base64Encode(uid)}.${cryptoObj.base64Encode(utid)}`;
     }
 }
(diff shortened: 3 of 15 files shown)


lib/msal-browser/src/cache/BrowserStorage.ts
```diff
@@ -94,7 +89,7 @@ export class BrowserStorage implements ICacheStorage {
             if (key.startsWith(`${Constants.CACHE_PREFIX}`) || key.startsWith(PersistentCacheKeys.ADAL_ID_TOKEN)) {
                 return key;
             }
-            return addInstanceId ? `${Constants.CACHE_PREFIX}.${this.clientId}.${key}` : `${Constants.CACHE_PREFIX}.${key}`;
+            return `${Constants.CACHE_PREFIX}.${this.clientId}.${key}`;
```
sameerag: +1. This is why I like new libraries. No roll overs :-D. Having said that we expect no cache support for implicit_flow to auth_code migration? We can get away with this if `msal@1.2.0` is mandated as the min version for SSO.

lib/msal-common/src/utils/TimeUtils.ts
```diff
@@ -0,0 +1,29 @@
...
+     * Returns time in seconds for expiration based on string value passed in.
+     *
+     * @param expiresIn
+     */
+    static parseExpiresIn(expiresIn: number): number {
+        // if AAD did not send "expires_in" property, use default expiration of 3599 seconds, for some reason AAD sends 3599 as "expires_in" value instead of 3600
+        if (!expiresIn) {
+            expiresIn = 3599;
+        }
+        return expiresIn;
```
jasonnutter: Would prefer not to mutate parameters:

`return expiresIn || 3599;`

lib/msal-common/src/utils/TimeUtils.ts
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    /**
+     * return the current time in Unix time. Date.getTime() returns in milliseconds.
+     */
+    static now(): number {
+        return Math.round(new Date().getTime() / 1000.0);
```
jasonnutter: `getTime()` returns ms already, shouldn't need to do any additional math.

jasonnutter: Also, can the function be named `nowMs` to indicate you are getting ms?

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -51,6 +54,50 @@ export class ResponseHandler {
...
+
+    private saveToken(originalTokenResponse: TokenResponse, authority: string, serverTokenResponse: ServerAuthorizationTokenResponse, clientInfo: ClientInfo): TokenResponse {
+        const tokenResponse = { ...originalTokenResponse };
+        // Set consented scopes in response
+        const requestScopes = ScopeSet.fromString(serverTokenResponse.scope, this.clientId, false);
+        tokenResponse.scopes = requestScopes.asArray();
+
+        const accessTokenCacheItems = this.cacheManager.getAllAccessTokens(this.clientId, authority);
+
+        for (let i = 0; i < accessTokenCacheItems.length; i++) {
```
jasonnutter: What about a `forEach`?

lib/msal-common/src/utils/TimeUtils.ts
jasonnutter: Ah, misread the comments...another reason why labelling the units here makes it clear we are working with seconds.

lib/msal-common/src/utils/TimeUtils.ts
jasonnutter: `parseExpiresInSeconds`

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -51,6 +54,50 @@ export class ResponseHandler {
...
+
+                if(cachedScopes.intersectingScopeSets(tokenResponse.scopes)) {
+                    this.cacheStorage.removeItem(JSON.stringify(accessTokenCacheItem.key));
+                }
+            }
+        }
+
+        // Expiration calculation
+        const expiresIn = TimeUtils.parseExpiresIn(serverTokenResponse.expires_in);
+        const expiration = TimeUtils.now() + expiresIn;
```
jasonnutter: Can we label these as `seconds` to make it clear which units are used?

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -51,6 +54,50 @@ export class ResponseHandler {
...
+        const accessTokenKey = new AccessTokenKey(authority, this.clientId, serverTokenResponse.scope, clientInfo.uid, clientInfo.utid, this.cryptoObj);
+        const accessTokenValue = new AccessTokenValue(serverTokenResponse.access_token, serverTokenResponse.id_token, serverTokenResponse.refresh_token, expiresIn.toString(), tokenResponse.account.homeAccountIdentifier);
+        this.cacheStorage.setItem(JSON.stringify(accessTokenKey), JSON.stringify(accessTokenValue));
+
+        // Save tokens in cache
+        tokenResponse.accessToken = serverTokenResponse.access_token;
+        tokenResponse.idToken = serverTokenResponse.id_token;
+        tokenResponse.refreshToken = serverTokenResponse.refresh_token;
+        tokenResponse.expiresOn = new Date(expiration * 1000);
+        return tokenResponse;
```
jasonnutter: `tokenRespone` doesn't really need to be created until the very end, and I would argue it makes the code harder to follow by creating it (at the top), mutating it, using the updated values, mutating again, then returning.

Recommendation:

```js
const requestScopes = ScopeSet.fromString(serverTokenResponse.scope, this.clientId, false);
const requestScopesArray = requestScopes.asArray();

// ... remove references to `tokenResponse` as noted above

return {
    ...originalTokenResponse,
    scopes: requestScopesArray,
    accessToken: serverTokenResponse.access_token,
    idToken: serverTokenResponse.id_token,
    refreshToken: serverTokenResponse.refresh_token,
    expiresOn: new Date(expirationSeconds * 1000)
}
```

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -51,6 +54,50 @@ export class ResponseHandler {
...
+        tokenResponse.scopes = requestScopes.asArray();
+
+        const accessTokenCacheItems = this.cacheManager.getAllAccessTokens(this.clientId, authority);
+
+        for (let i = 0; i < accessTokenCacheItems.length; i++) {
+            const accessTokenCacheItem = accessTokenCacheItems[i];
+            if (accessTokenCacheItem.key.homeAccountIdentifier === tokenResponse.account.homeAccountIdentifier) {
+                const cachedScopes = ScopeSet.fromString(accessTokenCacheItem.key.scopes, this.clientId, false);
+
+                if(cachedScopes.intersectingScopeSets(tokenResponse.scopes)) {
```
jasonnutter: `requestScopesArray` instead of `tokenResponse.scopes`

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -51,6 +54,50 @@ export class ResponseHandler {
...
+        const tokenResponse = { ...originalTokenResponse };
+        // Set consented scopes in response
+        const requestScopes = ScopeSet.fromString(serverTokenResponse.scope, this.clientId, false);
+        tokenResponse.scopes = requestScopes.asArray();
+
+        const accessTokenCacheItems = this.cacheManager.getAllAccessTokens(this.clientId, authority);
+
+        for (let i = 0; i < accessTokenCacheItems.length; i++) {
+            const accessTokenCacheItem = accessTokenCacheItems[i];
+            if (accessTokenCacheItem.key.homeAccountIdentifier === tokenResponse.account.homeAccountIdentifier) {
```
jasonnutter: Instead of using `tokenResponse`, you can use `originalTokenResponse`.

lib/msal-common/src/cache/AccessTokenKey.ts
```diff
@@ -0,0 +1,26 @@
...
+    clientId: string;
+    scopes: string;
+    homeAccountIdentifier: string;
+
+    constructor(authority: string, clientId: string, scopes: string, uid: string, utid: string, cryptoObj: ICrypto) {
+        const authorityUri = new UrlString(authority);
+        this.authority = authorityUri.urlString;
+        this.clientId = clientId;
+        this.scopes = scopes;
+        this.homeAccountIdentifier = cryptoObj.base64Encode(uid) + "." + cryptoObj.base64Encode(utid);
```
jasonnutter: Prefer string template over manual concatentation.

lib/msal-browser/src/utils/BrowserConstants.ts
```diff
@@ -9,7 +9,9 @@
  */
 export const BrowserConstants = {
     CACHE_LOCATION_LOCAL: "localStorage",
-    CACHE_LOCATION_SESSION: "sessionStorage"
+    CACHE_LOCATION_SESSION: "sessionStorage",
+    INTERACTION_STATUS_KEY: "interaction.status",
+    INTERACTION_IN_PROGRESS: "interaction_in_progress"
```
sameerag: the `_` for only one constant looks out of place while reading :)

lib/msal-common/src/cache/AccessTokenKey.ts
sameerag: +1

lib/msal-common/src/response/ResponseHandler.ts
jmckennon: +1

lib/msal-common/src/response/ResponseHandler.ts
sameerag: +1

jmckennon: +1

lib/msal-common/src/response/ResponseHandler.ts
sameerag: +1. Improves readability.

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -93,39 +137,34 @@ export class ResponseHandler {
...
+        if (!cachedAccount || Account.compareAccounts(cachedAccount, tokenResponse.account)) {
+            this.saveToken(tokenResponse, cachedAuthority, serverTokenResponse, clientInfo);
+        } else if (cachedAccount) {
+            throw ClientAuthError.createAccountMismatchError();
         }
 
         // Return user set state in the response
         tokenResponse.userRequestState = ProtocolUtils.getUserRequestState(state);
+        
+        this.cacheManager.resetTempCacheItems(state);
```
sameerag: Same logic as Jason's comment above, constructing each param and building tokenResponse at the end of the function makes it more readable.

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -51,6 +54,50 @@ export class ResponseHandler {
...
+        tokenResponse.refreshToken = serverTokenResponse.refresh_token;
+        tokenResponse.expiresOn = new Date(expiration * 1000);
+        return tokenResponse;
+    }
+
+    private getCachedAccount(accountKey: string): Account {
+        try {
+            return JSON.parse(this.cacheStorage.getItem(accountKey)) as Account;
+        } catch (e) {
+            return null;
```
jmckennon: do we not want to do anything with this error here if it happens? i get that it's relatively easy to see by looking at this code that it's an error in json parsing, but still might be useful to do something with this besides `return null`

sameerag:
Good work! Over all simplifying tokenProcessing the same way you did for loginRequest() helps in readability IMO.

lib/msal-browser/src/cache/BrowserStorage.ts
pkanher617: I think that we need to have the stipulation that upgrading from msal@1.2.0 is required, but I can revisit this post-alpha/post-beta if we have any issues.

lib/msal-common/src/utils/TimeUtils.ts
pkanher617: I changed the comments in the silent flow PR, it threw me off as well.

lib/msal-common/src/response/ResponseHandler.ts
pkanher617: This is done in a later PR.

lib/msal-common/src/response/ResponseHandler.ts
pkanher617: I have addressed this here, but will address this more in a future PR, as there is more work to be done here for the silent renewal flow.

lib/msal-common/src/response/ResponseHandler.ts

lib/msal-common/src/response/ResponseHandler.ts

lib/msal-common/src/cache/AccessTokenKey.ts
pkanher617: done.

lib/msal-browser/src/utils/BrowserConstants.ts
pkanher617: I understand, but most constants aren't going to be in this class, so I don't foresee this class growing past these variable. Happy to revisit later.

lib/msal-common/src/response/ResponseHandler.ts
pkanher617: Updated the flow, will update more in a future PR since this is used for renewToken flows as well.

lib/msal-common/src/response/ResponseHandler.ts
pkanher617: I'm not sure what we want to do in most cases, but I think it's better to assume that a malformed account is null rather than throw an error and force the application to break. But if you have any alternate suggestions, I'm happy to consider them.

lib/msal-common/src/utils/TimeUtils.ts
pkanher617: done.

lib/msal-common/src/response/ResponseHandler.ts
pkanher617: done

lib/msal-common/src/response/ResponseHandler.ts
jmckennon: I think it's worth to at least do some sort of `logger.error()`, no?

lib/msal-browser/src/cache/BrowserStorage.ts
sameerag: Great. Thanks.

lib/msal-browser/src/utils/BrowserConstants.ts
sameerag: okay!

lib/msal-common/src/response/ResponseHandler.ts
pkanher617: I will add this, good call.

pkanher617: Added in the pre alpha PR.

PR #1193, opened 2020-01-09
Remove isAngular flags for handling redirects

jasonnutter:
In MSAL Angular `0.1.4`, the processing of redirect responses happen in the Angular wrapper itself. For the next version of MSAL Angular library, we want the core library to handle that, so we need to remove these checks.

Cherry-picked from #1160


sameerag (commented):
So what about other places? Are we retaining this flag and taking it off only for `redirect` use cases for now?

jasonnutter:
@sameerag Yeah, not touching any of the other usage, for now.

sameerag (approved):
lgtm. Please track to remove the flag entirely later.

jasonnutter:
It is already removed in the work for v2. I'll make further removals for v1 as needed.

jmckennon (approved):
LGTM

PR #975, opened 2019-09-06
Iframes auth redirect

sameerag:
This PR in conjunction with #939 provides an option for applications living out of an iframe to handle interactive requests using "redirect".

**Problem description:** 
Some identity flows cannot be completed in an iframe. In particular:

- Credential Entry
- Consent

Clickjacking  is an attack whereby a malicious site iframes a trusted site and displays it to the user while simultaneously providing a transparent iframe on top of the trusted site. This transparent iframe is invisible to the user, but is able to record keystrokes (e.g. the password). Alternatively, a malicious site can move the trusted iframes it shows, permitting them to always move the "Consent" button to directly under the users mouse. Without going into detail, the only mitigation here is to forbid iframes on these flows. Identity permits services to use iframes only in combination with prompt=none, which ensures that these flows are not hit.

These limitations placed by Identity Service block any user interaction in an iframe app. This effectively requires all applications to be able to use a top-frame (either the current frame or a pop-up frame) to solve user interaction required prompts.

To mitigate this and have applications perform authentication in iframes, we are adding the below changes for different flows:

- [x] #939 helps with silent and popup calls where additional logic is added to monitor the popup/iframe opened by the MSAL JS to process the token as soon as the service redirects the request. APIs that are supported in iframed applications with this change are:
 - loginPopup()
- acquireTokenPopup()
 - acquireTokenSilent()

- [x] #975 (this PR) supports the "redirect" APIs by establishing a trusted relationship with the parent application (the top frame). As expanded above, the iframed application is prohibited from any interaction by microsoft identity service and hence this can be supported only with a trusted relationship with the parent application (which is a full window and hence can render interactive screens)

Detailed documentation on the protocol applied here will be published with the release of both these PRs.

PR #1201, opened 2020-01-13
Not Review Ready: Redirect API support for apps in iframes

sameerag:
This PR in conjunction with #939 provides an option for applications living out of an iframe to handle interactive requests using "redirect".

**Problem description:**
Some identity flows cannot be completed in an iframe. In particular:

- Credential Entry
- Consent

Clickjacking is an attack whereby a malicious site iframes a trusted site and displays it to the user while simultaneously providing a transparent iframe on top of the trusted site. This transparent iframe is invisible to the user, but is able to record keystrokes (e.g. the password). Alternatively, a malicious site can move the trusted iframes it shows, permitting them to always move the "Consent" button to directly under the users mouse. Without going into detail, the only mitigation here is to forbid iframes on these flows. Identity permits services to use iframes only in combination with prompt=none, which ensures that these flows are not hit.

These limitations placed by Identity Service block any user interaction in an iframe app. This effectively requires all applications to be able to use a top-frame (either the current frame or a pop-up frame) to solve user interaction required prompts.

To mitigate this and have applications perform authentication in iframes, we already added support for  `loginPopup()`, `acquireTokenPopup()` & `acquireTokenSilent()`.

#1201 (this PR) supports the `redirect` APIs: `loginRedirect()` and `acquireTokenRedirect()` by establishing a trusted relationship with the parent application (the top frame). As expanded above, the iframed application is prohibited from any interaction by microsoft identity service and hence this can be supported only with a trusted relationship with the parent application (which is a full window and hence can render interactive screens)

Detailed documentation on the protocol applied [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/files/4054744/RedirectIframes.pdf) will be published with the release of both these PRs.

This replaces #975 as the original merge was delayed in favor of doing this release in stages.

### DIFF
--- a/lib/msal-core/src/Configuration.ts
+++ b/lib/msal-core/src/Configuration.ts
@@ -96,6 +96,17 @@ export type FrameworkOptions = {
 };
 
 /**
+ * Options to specify communication between embedded (iframed) apps and the Top Frame
+ *
+ * - topFrameOrigin             - origin check to restrict messages to the top frame origin only
+ * - embeddedFrameOrigin        - origin check to restrict messages to the top frame origin only
+ */
+export type BrokerOptions = {
+    topFrameOrigin?: string;
+    embeddedFrameOrigin?: string;
+};
+
+/**
  * Use the configuration object to configure MSAL and initialize the UserAgentApplication.
  *
  * This object allows you to configure important elements of MSAL functionality:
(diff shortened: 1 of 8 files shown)


sameerag:
Closing this as the broker work in progress for msal 2.0 will solve this for all scenarios and browsers. cc @jasonnutter

PR #1202, opened 2020-01-13
Update urlContainsHash() API as a public API as documented

### LINKED ISSUE #1151: urlContainsHash removed from public API
## I'm submitting a...

<pre><code>
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Other... Please describe:
</code></pre>

## Browser:
- [x] Chrome version XX
- [x] Firefox version XX
- [ ] IE version XX
- [x] Edge version XX
- [x] Safari version XX

## Library version

<pre><code>
Library version: 1.2.0
</code></pre>

## Current behavior
`urlContainsHash` is missing in the public API

## Expected behavior
1. `urlContainsHash` removal should be mentioned as a breaking change;
1. Informational messages about `isCallback` deprecation should not be logged as public API doesn't contain other options.

In version 1.1.3 calling `isCallback` function resulted in informational message: `isCallback will be deprecated in favor of urlContainsHash in MSAL.js v2.0.`. We've changed the `isCallback` function to `urlContainsHash`. But in version 1.2.0 this function was removed from the UserAgentApplication prototype.

sameerag:
Fixes #1151; Add `urlContainsHash()` to the public interface as documented, instead of `isCallBack()`

### DIFF
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -277,6 +277,14 @@ export class UserAgentApplication {
         }
     }
 
+    /**
+     * Adding public interface for urlContainsHash
+     * @param hash
+     */
+    public urlContainsHash(hash: string) {
+        return UrlUtils.urlContainsHash(hash);
+    }
+
     private authResponseHandler(interactionType: InteractionType, response: AuthResponse, resolve?: any) : void {
         if (interactionType === Constants.interactionTypeRedirect) {
             if (this.errorReceivedCallback) {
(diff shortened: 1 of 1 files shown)


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -277,6 +277,14 @@ export class UserAgentApplication {
         }
     }
 
+    /**
+     * Adding public interface for urlContainsHash
```
jasonnutter: This seems like it could be the commit message, not teh comment :)

jasonnutter: Can use the comment from the `UrlUtils` file: "Check if the url contains a hash with known properties"

sameerag: My mistake, lazy coding. Corrected now. Please check.

PR #1206, opened 2020-01-14
Rewrite React Example: AuthProvider now a hook

isnifer:
There is a #1061, but author didn't go to finish and didn't rewrite HOC to hook. I did it.


samples/react-sample-app/src/useAuthProvider.js
```diff
@@ -0,0 +1,160 @@
...
+    const [account, setAccount] = useState(null);
+    const [error, setError] = useState(null);
+    const [emailMessages, setEmailMessages] = useState(null);
+    const [graphProfile, setGraphProfile] = useState(null);
+
+    useEffect(() => {
+        async function doAuth() {
+            msalApp.handleRedirectCallback(error => {
+                if (error) {
+                    const errorMessage = error.errorMessage
```
benjaminnaesen, line 26: I think `error.errorMessage || "Unable to acquire access token."` would be valid here instead of the ternary.

PR #1209, opened 2020-01-14
1.2.1 release PR

sameerag:
Includes:

- Change log for `msal@1.2.1
- Updated package.json and Constants.ts with the `msal js` version `1.2.1`

### DIFF
--- a/lib/msal-core/changelog.md
+++ b/lib/msal-core/changelog.md
@@ -1,3 +1,12 @@
+# 1.2.1
+
+## Bugs
+* `urlContainsHash()` is restored as a public API (#1202)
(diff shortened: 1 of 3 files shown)


lib/msal-core/changelog.md
```diff
@@ -1,3 +1,12 @@
+# 1.2.1
+
+## Bugs
+* `urlContainsHash()` is restored as a public API (#1202)
+* "allow-forms" added in sandbox properties for the iframes created by `msal js` to support certain B2C scenarios
```
jasonnutter: Back ticks can be used for `allow-forms` as well.

jasonnutter: Also, looks like this and the one below are missing references to PR #s.

PR #1208, opened 2020-01-14
Authorization Code Flow for Single Page Applications: Silent Flow and Logout

pkanher617:
This PR adds the silent flow and logout to the library. The implementation of ADAL SSO and other SSO features is included in a later PR.

This PR also enables CSP in our sample applications. This improves the security of our samples, and should be included in documentation moving forward.

PR #1211, opened 2020-01-15
Authorization Code Flow for Single Page Applications: SSO Fixes and changes

pkanher617:
This PR adds ADAL SSO and fixes other SSO scenarios for the Auth Code library.

Please review PR #1208 before this PR for better context.

### DIFF
--- a/lib/msal-common/src/app/module/AuthorizationCodeModule.ts
+++ b/lib/msal-common/src/app/module/AuthorizationCodeModule.ts
@@ -16,8 +16,6 @@ import { ClientConfigurationError } from "../../error/ClientConfigurationError";
 import { AuthorityFactory } from "../../auth/authority/AuthorityFactory";
 import { ServerCodeRequestParameters } from "../../server/ServerCodeRequestParameters";
 import { CodeResponse } from "../../response/CodeResponse";
-import { UrlString } from "../../url/UrlString";
-import { ServerAuthorizationCodeResponse, validateServerAuthorizationCodeResponse } from "../../server/ServerAuthorizationCodeResponse";
 import { ClientAuthError } from "../../error/ClientAuthError";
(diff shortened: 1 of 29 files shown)


lib/msal-common/src/app/module/AuthorizationCodeModule.ts
```diff
@@ -84,15 +84,20 @@ export class AuthorizationCodeModule extends AuthModule {
             );
 
             // Check for SSO
+            let adalIdToken: IdToken;
```
jmckennon: should we really be naming things with `adal`?

samples/VanillaJSTestApp/index_authCode.html
```diff
@@ -55,7 +55,7 @@
     // signin and acquire a token silently with POPUP flow. Fall back in case of failure with silent acquisition to popup
     function signIn() {
         try {
-            myMSALObj.loginRedirect(loginRequest);
+            myMSALObj.acquireTokenRedirect(loginRequest);
```
jasonnutter: Mean to commit this?

lib/msal-common/src/app/module/AuthorizationCodeModule.ts
jasonnutter: This is for SSOing to an ADAL session, so yes, in this case.

samples/VanillaJSTestApp/index_authCode.html
pkanher617: No, will update this in pre-alpha PR.

lib/msal-common/src/app/module/AuthorizationCodeModule.ts
pkanher617: As Jason said, this is a one-off specifically for ADAL, so doesn't get used otherwise.

PR #1221, opened 2020-01-17
Create SECURITY.md

2191116913:


jasonnutter:
CLA not signed, and I don't think we need this file, as we don't have any security disclosures at this time. Closing.

Issue #1045, opened 2019-10-14
Unable to change account in acquireTokenRedirect


LiamMartens:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Other... Please describe:
</code></pre>

## Browser:
- [x] Chrome version XX
- [x] Firefox version XX
- [x] IE version XX
- [x] Edge version XX
- [x] Safari version XX

## Library version
<pre><code>
Library version: 1.1.3

</code></pre>

## Current behavior

Due to the login hint a user is not able to switch his/her account when using acquireTokenRedirect even though the option is available in the microsoftonline screen. This is only relevant if there is still an account in cache by MSAL but the session has ended (this can be reproduced by fiddling with the cookies and localstorage. The issue is not applicable for all users but there are some enterprise users with certain policies where this issue occurs).
Following error is provided
```
AADSTS50058 A silent sign-in request was sent but none of the currently signed in user(s) match the requested login hint.
```

## Expected behavior
* Either the user should not be able to switch his/her account during the `acquireTokenRedirect` flow
* or the token response should correctly provide a new `preferred_username` and update the local account state so the login can be completed and a token can be retrieved.

## Minimal reproduction of the problem with instructions
* Login as normal using `loginRedirect`
* Remove the cookies from `microsoftonline.com` and remove the `msal.authority` session key or localstorage key (depending on config, issue occurs in both instances)
* Attempt to get a token using acquireTokenRedirect and choose "Log in with another account"
* The auth response will still have the previously logged in preferred_username which means acquireTokenSilent will fail again due to an account mismatch

keystroke:
Thanks for the details! I was hitting this previously, though not sure the circumstances were entirely the same. In my case, I was making a silent token request with different scopes than those I used when I signed-in the user, and that combined with something else was giving my this error. I was able to workaround it by ensuring that I use the same scopes during login that I do with my first silent token request, though I suspect there were other circumstances at play in my case.

LiamMartens:
@jasonnutter thanks for the info, it does appear the library does not allow for not setting the loginHint if there is a account in localStorage as it will attempt to autofill it. I made some changes to be able to hard disable the loginHint

jasonnutter:
@LiamMartens Thanks, we'll take a look at that PR and provide feedback there.

[...]

PR #1234, opened 2020-01-22
Allow disabling loginHint for certain scenarios

LiamMartens:
This is a fix for #1045. It was suggested to disable the loginHint, however this was not possible due to the implementation automatically filling it in.
I have made adjustment to not fill in loginHint (and sid for alignment purposes) when the implementation specifically requests so by setting the parameter to `false`


pkanher617:
Closing per #1045

Issue #1231, opened 2020-01-21
Msal-Browser stale state blocking login?


garrettm:
## Library
`@azure/msal-browser@2.0.0-alpha.0`

## Framework
React, but I don't think that matters

## Description
Sign in state gets stuck on `interaction_in_progress`

See https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/1000#issuecomment-576896230

## Reproduction steps
(I think these work, but not 100%)
1. Open a sign in pop up
2. Close/refresh the web app
3. Close the sign in pop up
4. Try to open a sign in pop up again from the reloaded app
5. Open fails, says it's blocked by interaction

## Expected behavior
I'd expect that a freshly reloaded app wouldn't have login blocked by prior app state

## Browsers
I tried it in Chrome

pkanher617:
Thanks @garrettm, I am reviewing this and debugging the issue.

tnorling:
Closing as this was fixed in the mentioned PR. Let us know if this is still an issue.

randompixel:
![image](https://user-images.githubusercontent.com/2316601/89685694-c2338e80-d8f4-11ea-99d6-e1ddf7648588.png)
I'm still seeing this using MSAL 2.0.1. I had prior cleared out all session and cookies, but coming back to the app this evening encountered this error. Pressing F5 to completely reload app doesn't fix it.

randompixel:
Although it seems by checking if user is authenticated before getting tokens for API request this fixes this. I guess it is trying to get a token as the SPA builds up and then the loginRedirect is being called which fails.

PR #1236, opened 2020-01-23
Authorization Code Flow for Single Page Applications: Cleanup Popup Flows and Window Nav

pkanher617:
This PR adds cleanup for different popup related errors, as well as window navigation actions such as redirect and refresh. This does not affect token acquisition, but does force users to be aware that handleRedirectCallback() should not be used with the popup APIs, as it will cause the window to handle the response in the popup, rather than the main window.

This should fix the issue raised in #1231.

### DIFF
--- a/lib/msal-browser/src/interaction_handler/PopupHandler.ts
+++ b/lib/msal-browser/src/interaction_handler/PopupHandler.ts
@@ -13,9 +13,6 @@ import { BrowserConstants } from "../utils/BrowserConstants";
  */
 export class PopupHandler extends InteractionHandler {
 
-    // Currently opened window handle.
-    private currentWindow: Window;
-
     /**
      * Opens a popup window with given request Url.
      * @param requestUrl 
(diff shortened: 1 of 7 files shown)


samples/VanillaJSTestApp/index_authCode.html
```diff
@@ -53,11 +53,15 @@
...
     // signin and acquire a token silently with POPUP flow. Fall back in case of failure with silent acquisition to popup
     function signIn() {
         try {
-            myMSALObj.loginRedirect(loginRequest);
+            myMSALObj.loginPopup(loginRequest).then(tokenResponse => {
+                console.log("Response: " + JSON.stringify(tokenResponse));
+            }).catch(error => {
+                console.error("Auth Error: " + error);
+            });
         } catch (e) {
```
jmckennon: Did you mean to commit these changes to the sample?

pkanher617: yes, in order to let people run the popup code and test the changes.

lib/msal-browser/src/interaction_handler/PopupHandler.ts
```diff
@@ -152,4 +152,15 @@ export class PopupHandler extends InteractionHandler {
...
         }
     }
+
+    private cleanPopup(popupWindow?: Window): void {
+        if (popupWindow) {
+            // Close window.
+            popupWindow.close();
+        }
+        // Remove window unload function
+        window.onbeforeunload = null;
```
jasonnutter: This will also remove any event listeners added by customers. You should use `window.removeEventListener('unload', callback);`

lib/msal-browser/src/interaction_handler/PopupHandler.ts
```diff
@@ -144,6 +139,11 @@ export class PopupHandler extends InteractionHandler {
             if (popupWindow.focus) {
                 popupWindow.focus();
             }
+            window.onbeforeunload = (): void => {
```
jasonnutter: Use `addEventListener` so that you don't blow away events registered by other scripts.

PR #1249, opened 2020-01-27
MSAL.js Core Library Docs and README update

pkanher617:
This PR updates the docs and README for the core library. The end goal of this PR and the others that follow will be to migrate all information out of the Github wiki and either have the source of truth be in the library /docs folder or the AAD Docs pages.


FAQ.md
```diff
@@ -0,0 +1,46 @@
+# FAQ
+
+## ADAL vs MSAL
+
+MSAL.js integrates with the Azure AD v2.0 endpoint, whereas ADAL.js integrates with the Azure AD v1.0 endpoint. The v1.0 endpoint supports work accounts, but not personal accounts. The v2.0 endpoint is the unification of Microsoft personal accounts and work accounts into a single authentication system. Additionally, with MSAL.js you can also get authentications for Azure AD B2C.
```
jmckennon: Do we want to explicitly state that `MSAL` also supports the v1 endpoint? Or is there already a previous assumption that's made where the v2 endpoint is backwards compatible with the v1 endpoint?

lib/msal-core/docs/account-object.md
```diff
@@ -0,0 +1,28 @@
+
+# Accounts in MSAL
+
+The MSAL SDKs have supported the notion of authenticating multiple user accounts. To align with this model, MSAL.js 1.x uses **Account** object. The object will surface the information about the authenticated user account.
```
jmckennon: Do we want to be bolding **Account** (and other similar references to ideas in the code) or inline coding them like `this`?

lib/msal-core/README.md
```diff
@@ -1,226 +1,104 @@
...
 
     <!-- IE support: add promises polyfill before msal.js  -->
     <script src="https://cdnjs.cloudflare.com/ajax/libs/bluebird/3.3.4/bluebird.min.js"></script>
 
-See here for more details on [supported browsers and known compatability issues](https://github.com/AzureAD/microsoft-authentication-library-for-js/wiki/FAQs#q4-what-browsers-is-msaljs-supported-on).
+See here for more details on [supported browsers and known compatability issues](./FAQ.md#what-browsers-is-msaljs-supported-on).
+
+## Usage
+
+1. [Initialization](./docs/initialization.md)
```
jmckennon: We should make these automatically open as a new tab.

```html
<a href="http://example.com/" target="_blank">Hello, world!</a>
```

FAQ.md
jasonnutter: It supports v1 tokens, but still uses the v2 endpoint, is my understand.

lib/msal-core/docs/account-object.md
pkanher617: I will change it to the inline code format

lib/msal-core/README.md
pkanher617: For external links I agree, internal Github links should stay in the same tab? So we don't have people opening multiple tabs.

FAQ.md
pkanher617: I believe that is correct as well.

jasonnutter: https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-compare-msal-js-and-adal-js#choosing-between-adaljs-and-msaljs

jasonnutter: Can we just link to this instead?

sameerag: Done. Linked to the relevant documentation and added one more section

lib/msal-core/FAQ.md
```diff
@@ -0,0 +1,224 @@
...
+
+In other cases, if navigating to the app's root page requires authentication, it might lead to **nested iframes** or **xframe deny error**.
+
+Since, MSAL.js cannot dismiss the 302 issued by Azure AD and is required to process the returned token, it cannot prevent the `redirect_uri` from getting loaded in the iframe.
+
+To avoid the entire app reloading again or other errors caused due to this, please follow these workarounds:
+
+* **Specify a different html for the iframe:**
+
+    Set `redirect_uri` property on config to a simple page, that does not require authentication. You have to make sure that it matches with the `redirect_uri` registered in Azure portal. This will not affect user's login experience as MSAL saves the start page when user begins the login process and redirects back to the exact location after login is completed.
```
jasonnutter: This needs to be more specific (explicitly mention that it requires `msal@1.2.0` and include a simple example.

lib/msal-core/FAQ.md
```diff
@@ -0,0 +1,224 @@
...
+    })
+    export class MsalComponent {
+      constructor(private Msal: MsalService) {
+      }
+    }
+    ```
+
+## I get this error "Refused to display ... in a frame because it set 'X-Frame-Options' to 'DENY'.
+
+The root cause for this is same as described [above in the answer](#how-to-avoid-page-reloads-when-acquiring-and-renewing-tokens-silently) to avoiding page reloads when acquiring and renewing tokens silently. Please try one of the solutions suggested in the answer.
```
jasonnutter: This isn't the correct answer anymore.

lib/msal-core/docs/account-object.md
jasonnutter: I don't think we need this file at all, this should be reference documentation.

lib/msal-core/docs/acquire-and-renew-tokens.md
```diff
@@ -0,0 +1,105 @@
+# Acquire and Renew a Token to Call an API
+
+In MSAL, you can get access tokens for the APIs your app needs to call using the `acquireTokenSilent` method which makes a silent request(without prompting the user with UI) to Azure AD to obtain an access token. The Azure AD service then returns an [access token](https://docs.microsoft.com/en-us/azure/active-directory/develop/access-tokens) containing the user consented scopes to allow your app to securely call the API.
```
jasonnutter: This seems like it could go on docs.microsoft.com @sameerag

Issue #1213, opened 2020-01-15
MSAL Angular Wrapper is not using protectedResourceMap when building with AOT


ricey-cmd:
## Library
- [x] `msal@1.2.1-beta.1`
- [x] `@azure/msal-angular@1.0.0-alpha.1`

## Framework
Angular v8.1.0

## Description

When building my project with AOT, it seems the protectedResourcesMap is not being picked up correctly by MSAL and is not appending the access token onto the Bearer header in the relevant http calls. I believe this is to do with the new use of adding this as a Map rather than an array  which may not be supported by AOT.

A second issue is an error while attempting to setup the logger in the MSAL configuration in app.module with AOT.
When building the following error is thrown:

  "Function calls are not supported in decorators but 'Logger' was called."

I understand this is the alpha version but I wanted to comment in order to raise awareness of the issue.

Thanks

jasonnutter:
@ricey-cmd Thanks for the feedback, I'll take a look into this and let you know if I need further info.

asanchezfe:
@ricey-cmd 
We're having the logger working on AOT compilation with the same packages that you, try this configuration:
```
/*
import { Logger } from 'msal';
export function MSALConfigFactory(config: ConfigService) {
  return {
    auth: {
     ...
    },
    cache: {
      ...
    },
    framework: {
      ....
    },
    system: {
      logger: new Logger(
        (logLevel, message, piiEnabled) => {
          console.log(message);
        },
        { level: 3 }
      )
    }
  };
}
```

ricey-cmd:
@asanchezfe I tried this configuration and still getting the same issue. Seems it doesn't like the new Logger call. Same Error as before: 'Function calls are not supported in decorators but 'Logger' was called in 'MSALConfigFactory''

asanchezfe:
@ricey-cmd  What webpack, angular... are you using? We use aot, angular 8 and `"@angular-builders/custom-webpack": "^8.4.0"`

Raysapps:
Executing 'ng build ---prod'  with   @azure/msal-angular 1.0.0-beta.1 version is giving an error "Function calls are not supported in decorators but 'Logger' was called in 'environment'". 
I understand that this is a beta version but this is stopping me from publishing on to Azure. If anyone can please suggest a workaround, that'll be great.

ricey-cmd:
[...]

PR #1251, opened 2020-01-28
Add setLogger function to allow dynamically setting logger callback

jasonnutter:
In Angular, function calls are not allowed in decorators when running in `aot` mode. This means if you call `new Logger`, Angular will throw the following error (as reported in #1213):

```
ERROR in src/app/app.module.ts(58,21): Error during template compile of 'AppModule'
  Function calls are not supported in decorators but 'Logger' was called.
```

Adding a `setLogger` function to `UserAgentApplication` mitigates this:

```js
    this.authService.setLogger(new Logger((logLevel, message, piiEnabled) => {
      console.log('MSAL Logging: ', message);
    }, {
      correlationId: CryptoUtils.createNewGuid(),
      piiLoggingEnabled: false
    }));
```

### DIFF
--- a/lib/msal-angular/src/msal.service.ts
+++ b/lib/msal-angular/src/msal.service.ts
@@ -165,10 +165,6 @@ export class MsalService extends UserAgentApplication {
         });
     }
 
-    public getLogger(): Logger {
-        return super.getLogger();
-    }
-
     public getScopesForEndpoint(endpoint: string): string[] {
         return super.getScopesForEndpoint(endpoint);
     }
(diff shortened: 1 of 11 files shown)


pkanher617:
May be better to create a separate PR on dev for the UAA changes and then merge dev into your branch so you can prevent merge conflicts in the future.

pkanher617 (approved):
LGTM

jasonnutter:
@pkanher617 Yeah, I'll likely end up cherry-picking the UAA change into `dev`.

PR #1255, opened 2020-01-29
Move protectedResourceMap and unprotectedResources to Angular-specific config

jasonnutter:
Both `protectedResourceMap` and `unprotectedResources` are only used by Angular, and don't need to be in core. Furthermore, there are issues with `new Map` in aot mode in Angular, so the new config options will accept both an array or a map (and also accept the old config values, which are now deprecated.

This should address #1213.

Should be merged after #1251.


jmckennon (approved):
Looks good to me

PR #1276, opened 2020-02-11
B2C Authority fixes

tnorling:
- Updated B2C Authority to extend the generic Authority class, rather than AADAuthority.
- Removed url slicing which was breaking instantiation of B2CAuthority for certain authorities 
- Added knownAuthorities config parameter which sets the trusted authorities for B2C use cases
- Authority factory will create B2C Authority instance if knownAuthoritites is set, AAD Authority instance otherwise


jasonnutter:
@tnorling What was the reasoning behind changing the base class?

lib/msal-core/src/authority/B2cAuthority.ts
```diff
@@ -39,4 +31,13 @@ export class B2cAuthority extends AadAuthority {
...
     }
+
+    /**
+     * Checks to see if the host is in a list of trusted hosts
+     * @param {string} The host to look up
+     */
+    public IsInTrustedHostList(host: string): boolean {
+        // Change this when we implement Known Authorities
+        return AADTrustedHostList[host.toLowerCase()];
+    }
```
sameerag: Do we need this for B2C? This may morph to the `knownAuthorities()` in the future; For now can we assume `validateAuthority: false` as pre requisite and remove this in `B2CAuthority.ts` all together?

lib/msal-core/src/authority/B2cAuthority.ts
```diff
@@ -3,26 +3,18 @@
...
     public constructor(authority: string, validateAuthority: boolean) {
         super(authority, validateAuthority);
-        const urlComponents = UrlUtils.GetUrlComponents(authority);
-
-        const pathSegments = urlComponents.PathSegments;
-        if (pathSegments.length < 3) {
-            throw ClientConfigurationErrorMessage.b2cAuthorityUriInvalidPath;
-        }
-
-        this.CanonicalAuthority = `https://${urlComponents.HostNameAndPort}/${pathSegments[0]}/${pathSegments[1]}/${pathSegments[2]}/`;
```
jasonnutter: Is this logic elsewhere (e.g. in the base class), or do we not want to run it at all?

tnorling: We do not want to run it at all. It was breaking the b2c sample as the authority url did not contain enough segments. this.CanonicalAuthority is set in the base class, this was only overwriting it

lib/msal-core/src/authority/B2cAuthority.ts
tnorling: Could remove it. My thinking was to use this function to check against knownAuthorities, when implemented, to keep the structure in line with AAD. Can be discussed

tnorling:
> @tnorling What was the reasoning behind changing the base class?

AAD and B2C are distinct instances, it didnt make sense for AADAuthority to extend Authority and then B2C to extend AADAuthority when we are overwriting the AAD functions anyway within B2C

lib/msal-core/src/authority/B2cAuthority.ts
jasonnutter: Gotya. My understanding was that canonical authority helped normalize the trailing slash on the url, which I think we still want? (if so, and there isn't a unit test to catch this, we should add one)

pkanher617: @jasonnutter is correct, we should make sure we using the same mechanisms to normalize the URLs

tnorling: Right, the url is still normalized by the base class. The only thing I removed is the mandate that the url must contain 3 segments (something that is not present for AAD). However, if we want to leave this as is we can change this to only require 2 segments in order to reflect the b2c authority url format: https://<tenant name>.b2clogin.com/<tenant name>.onmicrosoft.com/<policy name>

sameerag: I think we should get rid of it. I see no value add, and this could be an implementation with assumptions that are no longer true (B2C authority formats changed over the years). The base canonical API in UrlUtils should be enough.

lib/msal-core/src/Configuration.ts
```diff
@@ -36,6 +37,7 @@ export type AuthOptions = {
     clientId: string;
     authority?: string;
     validateAuthority?: boolean;
+    knownAuthorities?: Array<string>;
```
sameerag: Can we do surface changes @DarylThayil and mark it `msal@.1.3.0`? I know it isn't exact `semver` but that can help moving the below params to:
```javascript
export type AuthorityOptions = {
    validateAuthority?: boolean;
    knownAuthorities?: Array<string>;
};
```

Or even get rid of `validateAuthority` and have just `knownAuthorities` as the option in this type. That way we can default validation to true for anyone who doesn't pass `knownAuthorities`. Just a thought.

lib/msal-core/src/authority/AuthorityFactory.ts
```diff
@@ -37,15 +22,13 @@ export class AuthorityFactory {
...
         // Depending on above detection, create the right type.
-        switch (type) {
-            case AuthorityType.B2C:
-                return new B2cAuthority(authorityUrl, validateAuthority);
-            case AuthorityType.Aad:
-                return new AadAuthority(authorityUrl, validateAuthority);
-            default:
-                throw ClientConfigurationErrorMessage.invalidAuthorityType;
+        if (Object.keys(B2CTrustedHostList).length) {
+            return new B2cAuthority(authorityUrl, validateAuthority);
```
sameerag: Can we make this ternary check?

tnorling: Good call, will do

sameerag: Also may be consider ADFS in future and makes changes accordingly. ADFS should be a different type of authority.

sameerag: You can default it to a generic case and not B2C if you decide on instantiating all three authorities as different instances.

tnorling: I think ultimately it would make sense to collapse AAD and B2C classes into a single generic class as there is very little difference between them. Not sure how ADFS would differ at that point. But that's probably a longer topic of discussion

sameerag: Yes. If we decide no `type`: All collapse to one class; otherwise they stay separate. Deciding it now and making the changes will make it easy?

lib/msal-core/src/utils/Constants.ts
```diff
@@ -107,6 +107,8 @@ export const AADTrustedHostList =  {
     "login.microsoftonline.us": "login.microsoftonline.us"
 };
 
+export const B2CTrustedHostList = {};
```
pkanher617: I would consider moving this to AuthorityFactory or B2CAuthority since it isn't actually a constant.

pkanher617: I'm also not a huge fan of having a global variable like this, I think it should be a property or field of an object or class

tnorling: Moved this to AuthorityFactory but not sure how to avoid making it global since it needs to be read from the static detectAuthorityFromUrl function.

lib/msal-core/src/authority/B2cAuthority.ts
```diff
@@ -3,26 +3,18 @@
  * Licensed under the MIT License.
  */
 
-import { AadAuthority } from "./AadAuthority";
+import { Authority } from "./Authority";
 import { AuthorityType } from "./Authority";
-import { ClientConfigurationErrorMessage } from "../error/ClientConfigurationError";
-import { UrlUtils } from "../utils/UrlUtils";
+import { ClientConfigurationError } from "../error/ClientConfigurationError";
+import { AuthorityFactory, B2CTrustedHostList } from "./AuthorityFactory";
```
jasonnutter, line 9: This creates a circular dependency :( sorry didn't catch this before.

lib/msal-core/src/authority/AuthorityFactory.ts
```diff
@@ -9,25 +9,42 @@
 import { AadAuthority } from "./AadAuthority";
 import { B2cAuthority } from "./B2cAuthority";
```
jasonnutter, line 10: Other end of the circle

mmacy: Per @tnorling this circular dep was resolved in https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/1479

PR #1232, opened 2020-01-22
MSAL Angular beta w/ Angular 6+ support

jasonnutter:
Initial upgrade of MSAL Angular to be built with Angular 7 CLI, with compatibility for Angular 6+. 

Requires:
- `msal@1.2.1`
- `rxjs@6`

Install: `npm install msal@latest @azure/msal-angular@beta rxjs@6`

Also includes three new sample apps for Angular 6, 7, 8: https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev-angular-1.0-beta/samples

Full changelog: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev-angular-1.0-beta/lib/msal-angular/changelog.md

PR #1277, opened 2020-02-12
MSAL Angular 1.0 beta

jasonnutter:
PR to merge my work so far for MSAL Angular. Future PRs will be done to `dev`.

Includes work done in:
* #1255 
* #1251 
* #1232 
* #1291


DarylThayil:
Very large pr, can you point out the pieces you need review on?

lib/msal-angular/changelog.md
```diff
@@ -1,5 +1,42 @@
 # Changelog
```
jasonnutter: Please review.

lib/msal-angular/docs/0.x-1.x-upgrade-guide.md
```diff
@@ -0,0 +1,58 @@
+# Upgrading from MSAL Angular 0.x to 1.x
```
jasonnutter: Please review.

lib/msal-angular/changelog.md
```diff
@@ -1,5 +1,42 @@
...
+Fixes two issues related to `aot` mode.
+
+* Requires `msal@1.2.2-beta.0`, which adds `setLogger` function to dynamically set the logger callback when running in `aot` mode. (#1213).
+* Moves `protectedResourceMap` and `unprotectedResources` to MSAL Angular-specific configuration object. `protectedResourceMap` can now be `[string, string[]][]` or a `Map`. This is also to mitigate issues with `aot` mode. (#1213).
+
+## 1.0.0-beta.1
+
+Initial upgrade to be compatible with new version of Angular (6+).
+
+* Requires `msal@1.2.1`, `rxjs@6`, `@angular/core@>=6`, `@angular/common@>=6` as a peer dependencies.
```
jmckennon: Grammar here; `as a peer dependencies -> as peer dependencies`

lib/msal-angular/src/msal-guard.service.ts
```diff
@@ -1,96 +1,60 @@
...
-                    }
-                }
+        if (!this.authService.getAccount()) {
+            if (this.msalAngularConfig.popUp) {
+                return this.authService.loginPopup({
+                    scopes: this.msalAngularConfig.consentScopes,
+                    extraQueryParameters: this.msalAngularConfig.extraQueryParameters
+                })
+                    .then(() => true)
+                    .catch(() => false);
```
jmckennon: more questions about guards i guess... do guards not inherently have something like this implemented in them? from what we talked about earlier, it seemed like this was a part of their purpose and may be built in.

if that isn't the case, this is fine, just curious

lib/msal-angular/docs/0.x-1.x-upgrade-guide.md
```diff
@@ -0,0 +1,58 @@
...
+* Broadcast events now emit objects, instead of just strings.
+* Applications using `Redirect` methods must implement the `handleRedirectCallback` method (and have it run on every page load), which will capture the result of redirect operations. See the [Angular sample](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/samples/angular6-sample-app/src/app/app.component.ts) for an example of how to implement.
+
+## Angular 6+ and rxjs@6
+
+MSAL Angular now expects that your application is built with `@angular/core@>=6`, `@angular/core@>=6`, `rxjs@6`, and `rxjs-compat` is no longer required.
+
+Steps:
+1. Install newer versions of Angular and rxjs: `npm install @angular/core @angular/common rxjs`
+2. Uninstall `rxjs-compat`: `npm uninstall rxjs-compat`
```
jmckennon: in the changelog, you mention that "`rxjs-compat` is no longer required." as though you still had the option to use it, yet this suggests that we actually recommend to not use it at all. we should probably make this idea clearer, as it isn't even clear to me which it is by reading this

lib/msal-angular/src/msal-guard.service.ts
jasonnutter: It's on you to implement the desired business logic for your guard, but ultimately, you need to return a boolean indicating whether or not the user has met the criteria for accessing the desired page.

lib/msal-angular/docs/0.x-1.x-upgrade-guide.md
jasonnutter: Good point. I don't believe it will cause any harm to have it installed, and I suppose it might be possible that an app needs it for compatibility with other libraries. I'll make sure its clear that it is no longer needed for MSAL Angular itself.

lib/msal-angular/src/msal.service.ts
```diff
@@ -66,324 +61,151 @@ export class MsalService extends UserAgentApplication {
...
-    private isUnprotectedResource(url: string) {
-        if (this.config && this.config.unprotectedResources) {
-            for (var i = 0; i < this.config.unprotectedResources.length; i++) {
-                if (url.indexOf(this.config.unprotectedResources[i]) > -1) {
-                    return true;
-                }
-            }
-        }
-        return false;
+        return unprotectedResources.some(resource => url.indexOf(resource) > -1);
```
asiffermann: I hope you won't mind if I add a comment to your PR as a simple beta tester 😉
If you want that I open an issue instead, please let me know!

I have two issues with this single line.

- I had no unprotected resources in my Angular application, and I first didn't specify it in the `MsalAngularConfiguration` object given to configure the module. So it was undefined, and throws an error here ("Cannot call method some of undefined"). A simple workaround is to explicitly specify it as an empty array, but the code could be more protective, or the object given as the second argument of `MsalModule.forRoot` could be merged with the default one.

-  The second issue is more problematic, and concerns the URL matching. I had collisions between protected API URLs, and Angular route's paths, which are now automatically pushed in the `unprotectedResources` array. For example, I have protected calls to an API hosted at "https://api.com/contracts" and an Angular route "contracts".
As the matching is done with a "contains" (`url.indexOf(resource) > -1`), the interceptor considered my API calls as unprotected, and didn't include a token for it. Wouldn't it be safer to match URLs with a `startsWith`?

jasonnutter: @asiffermann Of course, appreciate the feedback!

1. `unprotectedResources`: This will be fixed by #1291 
2. URL matching: Yeah, this seems reasonable. I'll investigate and follow up.

jasonnutter: Fix for unprotectedResources has been included in this PR, interceptor will be fixed incrementally.

lib/msal-angular/changelog.md
```diff
@@ -1,5 +1,42 @@
 # Changelog
 
+## 1.0.0-beta.2
+
+Fixes two issues related to `aot` mode.
```
DarylThayil: can you either link the issues and  / or describe issue

lib/msal-angular/changelog.md
```diff
@@ -1,5 +1,42 @@
...
+## 1.0.0-beta.1
+
+Initial upgrade to be compatible with new version of Angular (6+).
+
+* Requires `msal@1.2.1`, `rxjs@6`, `@angular/core@>=6`, `@angular/common@>=6` as peer dependencies.
+* `rxjs-compat` is no longer required by MSAL Angular.
+
+### Known issues
+
+* `aot` compiling will throw errors for `new Logger()` and `new Map()` ("Function calls are not supported in decorators but 'Logger/Map' was called."). This will be addressed in a follow up release.
```
DarylThayil: Was this resolved in 1.0.0 -2

DarylThayil (approved):

DarylThayil:
LGTM<

jmckennon (approved):
Looks good to me now

sameerag (approved):

PR #1278, opened 2020-02-13
render iframe synchronously if the user sets zero navigateToFrame time

sameerag:
Eliminate setTimeOut() calls when a user sets `navigateFrameWait` in `msal` config  to `0` to improve performance.

Currently MSAL JS while loading iframes for silent calls always makes an async call to `loadFrame` and slowing down performance intensive apps. This is introduced for IE which requires delay in loading a frame and the modern browsers seem to do alright with no time out (feedback from users who have this in production).

This PR would thus allow for better perf by eliminating the asynchronous call for the modern browsers when the timeout is set to `0`

### DIFF
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -793,8 +793,8 @@ export class UserAgentApplication {
 
         // render the iframe synchronously if app chooses no timeout, else wait for the set timer to expire
         const iframe: HTMLIFrameElement = this.config.system.navigateFrameWait ?
-            WindowUtils.loadFrameSync(urlNavigate, frameName, this.logger):
-            await WindowUtils.loadFrame(urlNavigate, frameName, this.config.system.navigateFrameWait, this.logger);
+            await WindowUtils.loadFrame(urlNavigate, frameName, this.config.system.navigateFrameWait, this.logger):
+            WindowUtils.loadFrameSync(urlNavigate, frameName, this.logger);
 
         try {
             const hash = await WindowUtils.monitorWindowForHash(iframe.contentWindow, this.config.system.loadFrameTimeout, urlNavigate);
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -792,13 +792,9 @@ export class UserAgentApplication {
         this.cacheStorage.setItem(`${TemporaryCacheKeys.RENEW_STATUS}${Constants.resourceDelimiter}${expectedState}`, Constants.inProgress);
 
         // render the iframe synchronously if app chooses no timeout, else wait for the set timer to expire
-        let iframe: HTMLIFrameElement;
-        if(this.config.system.navigateFrameWait == 0) {
-            iframe = WindowUtils.syncLoadFrame(urlNavigate, frameName, this.logger);
-        }
-        else {
-            iframe = await WindowUtils.loadFrame(urlNavigate, frameName, this.config.system.navigateFrameWait, this.logger);
-        }
+        const iframe: HTMLIFrameElement = this.config.system.navigateFrameWait ?
+            WindowUtils.loadFrameSync(urlNavigate, frameName, this.logger):
+            await WindowUtils.loadFrame(urlNavigate, frameName, this.config.system.navigateFrameWait, this.logger);
 
         try {
             const hash = await WindowUtils.monitorWindowForHash(iframe.contentWindow, this.config.system.loadFrameTimeout, urlNavigate);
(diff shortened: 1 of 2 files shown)


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -791,7 +791,14 @@ export class UserAgentApplication {
...
 
-        const iframe = await WindowUtils.loadFrame(urlNavigate, frameName, this.config.system.navigateFrameWait, this.logger);
+        // render the iframe synchronously if app chooses no timeout, else wait for the set timer to expire
+        let iframe: HTMLIFrameElement;
+        if(this.config.system.navigateFrameWait == 0) {
+            iframe = WindowUtils.syncLoadFrame(urlNavigate, frameName, this.logger);
+        }
+        else {
+            iframe = await WindowUtils.loadFrame(urlNavigate, frameName, this.config.system.navigateFrameWait, this.logger);
+        }
```
jasonnutter: A couple things:

- You can simplify the check to test truthiness of `navigateFrameWait`, since `0` is falsey.
- Can you make this a ternary?

```js
const iframe = this.config.system.navigateFrameWait ? 
  await WindowUtils.loadFrame(urlNavigate, frameName, this.config.system.navigateFrameWait, this.logger) : 
  WindowUtils.syncLoadFrame(urlNavigate, frameName, this.logger);
```

lib/msal-core/src/utils/WindowUtils.ts
```diff
@@ -107,6 +107,23 @@ export class WindowUtils {
...
     }
 
+    /**
+     * @hidden
+     * Loads the iframe synchronously when the navigateTimeFrame is set to `0`
+     * @param urlNavigate
+     * @param frameName
+     * @param logger
+     */
+    static syncLoadFrame(urlNavigate: string, frameName: string, logger: Logger): HTMLIFrameElement{
```
jasonnutter: Could this be `loadFrameSync`, to keep the convention consistent?

lib/msal-core/src/utils/WindowUtils.ts
```diff
@@ -107,6 +107,23 @@ export class WindowUtils {
...
+        const frameHandle = WindowUtils.addHiddenIFrame(frameName, logger);
+        if (frameHandle.src === "" || frameHandle.src === "about:blank") {
+            frameHandle.src = urlNavigate;
+            logger.infoPii("Frame Name : " + frameName + " Navigated to: " + urlNavigate);
```
jasonnutter: And you could have the async version call the sync version inside the `setTimeout`, to make it so the logic of setting the `src` isn't duplicated.

lib/msal-core/src/UserAgentApplication.ts
sameerag: I thought of it but wanted to test it before making it ternary as numbers and strings behave differently in TS :-D. Will do.

lib/msal-core/src/utils/WindowUtils.ts
sameerag: 👍

lib/msal-core/src/UserAgentApplication.ts
jmckennon: +1 to ternary

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -791,7 +791,10 @@ export class UserAgentApplication {
         this.logger.verbose("Set loading state to pending for: " + scope + ":" + expectedState);
         this.cacheStorage.setItem(`${TemporaryCacheKeys.RENEW_STATUS}${Constants.resourceDelimiter}${expectedState}`, Constants.inProgress);
 
-        const iframe = await WindowUtils.loadFrame(urlNavigate, frameName, this.config.system.navigateFrameWait, this.logger);
+        // render the iframe synchronously if app chooses no timeout, else wait for the set timer to expire
+        const iframe: HTMLIFrameElement = this.config.system.navigateFrameWait ?
```
AndreiZe: Do I read this right: sync version will run if navigateFrameWait is truthy? Don't we want the opposite?

jasonnutter: Ah yes, good catch.

sameerag: :-D Sorry. Last minute changes, I need to sleep more :)

sameerag: Done. Thanks @AndreiZe

lib/msal-core/src/UserAgentApplication.ts
sameerag: Done!

DarylThayil:
LGTM, can merge

PR #1267, opened 2020-02-06
Update the framename to reflect authority and scopes

sameerag:
This PR fixes #1225 by making the frame name unique for scopes+authority when making silent calls

Issue #1225, opened 2020-01-20
MSAL fails to create separate hidden iframes for each token request based on target authority, resulting in returning the wrong token or error to registered callbacks


keystroke:
## Library
- [x] `msal@1.2.0` or `@azure/msal@1.x.x`
- [ ] `@azure/msal-angular@0.x.x`
- [ ] `@azure/msal-angular@1.x.x`
- [ ] `@azure/msal-angularjs@1.x.x`

## Framework

None.

## Description

[UPDATE]

See my latest comment; I've renamed the issue according to what I believe is the root cause. It seems that MSAL is creating hidden iframes identified only by target scopes in the token request, but failing to include target authority. This results in multiple concurrent token requests with distinct authorities but identical scopes to be joined together and all of them receive the same result from that single hidden iframe!

I have also found a hacky workaround, the mechanism for which might be usable in more scoped scenarios if people hit this issue, though creating a generic version of the workaround would be difficult (easier to just fix in MSAL directly to scope target hidden iframes by authority and scopes).

[/UPDATE]

When requesting tokens for multiple different authorities at the same time, the same response (error or result) will be return to all the .then or .catch handlers.

Details described in this post are primarily for above issue, but I notice many issues trying to sign-in a user and make API calls to multiple different directories, all concurrently. For example, only one popup may be active at a time. This is perhaps fine, but while a popup is active, any silent token requests will terminate the popup... I would expect that silent token requests should not affect a running popup request, and they should be isolated somehow.

In general, please advise on how to setup using MSAL if I want to sign a user into their home directory, and then proceed to acquire many different tokens concurrently, including handling popups for each necessary consent (e.g. in below issue example you will see 9 directories I am attempting to authenticate with, and some of them require consent; I have built a wrapper to allow the user to proceed with granting consent via popups one at a time, and once consent is granted, I acquire other tokens in each directory and make other API calls, and all of this means I have build an insanely complicated wrapper around MSAL that ensures token requests are processed serially as necessary which is obviously terrible). For example do I need to be creating separate MSAL client objects for each target directory? What about the caching behavior? etc.

## Security
Is this issue security related?

I would say yes, because this error isn't obvious at first and when testing my web app I was taking actions against the wrong tenant! Especially since with the new graph APIs, the target tenant is only in the token, not in the URI like it used to be!

## Regression

Don't recall. I've been waiting for MSAL to mature fully to do my basic operations:

1. sign-in user
[...]

PR #1279, opened 2020-02-13
Example refactor for proper concurrency support

keystroke:
Token requests are not properly disambiguated for target authority or other semantics, this change purposes modification to some of the code that was only considering 'scope' values to instead consider a 'request uniqueness identifier' that is resolved based on the target request and account, instead of just target scopes. It seems like previous change to support setting authority per-token request was only tested when making single token calls, so the lack of proper disambiguation was not noticed.

### DIFF
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -466,7 +466,6 @@ export class UserAgentApplication {
     private acquireTokenHelper(account: Account, interactionType: InteractionType, isLoginCall: boolean, request?: AuthenticationParameters, resolve?: any, reject?: any): void {
         // Track the acquireToken progress
         this.cacheStorage.setItem(TemporaryCacheKeys.INTERACTION_STATUS, Constants.inProgress);
-        const scope = request.scopes ? request.scopes.join(" ").toLowerCase() : this.clientId.toLowerCase();
 
         let serverAuthenticationRequest: ServerRequestParameters;
         const acquireTokenAuthority = (request && request.authority) ? AuthorityFactory.CreateInstance(request.authority, this.config.auth.validateAuthority) : this.authorityInstance;
@@ -537,7 +536,8 @@ export class UserAgentApplication {
                 window.requestType = isLoginCall ? Constants.login : Constants.renewToken;
 
                 // Register callback to capture results from server
-                this.registerCallback(serverAuthenticationRequest.state, scope, resolve, reject);
+                const requestUid = RequestUtils.computeTokenRequestUniquenessIdentifier(serverAuthenticationRequest, account);
+                this.registerCallback(serverAuthenticationRequest.state, requestUid, resolve, reject);
             } else {
                 throw ClientAuthError.createInvalidInteractionTypeError();
             }
@@ -691,12 +691,13 @@ export class UserAgentApplication {
                     .then(() => {
                         /*
                          * refresh attempt with iframe
-                         * Already renewing for this scope, callback when we get the token.
+                         * Already renewing for this request, callback when we get the token.
                          */
-                        if (window.activeRenewals[scope]) {
-                            this.logger.verbose("Renew token for scope: " + scope + " is in progress. Registering callback");
+                        const requestUid = RequestUtils.computeTokenRequestUniquenessIdentifier(serverAuthenticationRequest, account);
+                        if (window.activeRenewals[requestUid]) {
+                            this.logger.verbose("Renew token for request: '" + requestUid + "' is in progress. Registering callback");
                             // Active renewals contains the state for each renewal.
-                            this.registerCallback(window.activeRenewals[scope], scope, resolve, reject);
+                            this.registerCallback(window.activeRenewals[requestUid], requestUid, resolve, reject);
                         }
                         else {
                             if (request.scopes && request.scopes.indexOf(this.clientId) > -1 && request.scopes.length === 1) {
@@ -706,11 +707,11 @@ export class UserAgentApplication {
                                  */
                                 this.logger.verbose("renewing idToken");
                                 this.silentLogin = true;
-                                this.renewIdToken(request.scopes, resolve, reject, account, serverAuthenticationRequest);
+                                this.renewIdToken(resolve, reject, account, serverAuthenticationRequest);
                             } else {
                                 // renew access token
                                 this.logger.verbose("renewing accesstoken");
-                                this.renewToken(request.scopes, resolve, reject, account, serverAuthenticationRequest);
+                                this.renewToken(resolve, reject, account, serverAuthenticationRequest);
                             }
                         }
                     }).catch((err) => {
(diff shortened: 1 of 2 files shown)


keystroke:
@sameerag @jasonnutter I prepared this change which should showcase the kind of modifications necessary to properly support concurrent token requests in MSAL. While working with code I thought to suggest additional abstractions, as the current code in the public client class seems to have several methods with internal logic duplicated (as you can see from this change) that should probably be encapsulated into a separate function. But this change does not address that. This changes extends our discussions from #1225 and also from PR #1267.

lib/msal-core/src/utils/RequestUtils.ts
```diff
@@ -156,4 +158,24 @@ export class RequestUtils {
...
+    static computeTokenRequestUniquenessIdentifier(request: ServerRequestParameters, account: Account): string {
+        // TODO - resolve exact information that should disambiguate a request
+        let parts = [
+            account.accountIdentifier,
+            request.authority,
+            request.scopes.sort().join(Constants.resourceDelimiter),
+            request.promptValue // a token request to prompt consent should be distinct from one that does not require consent
+        ];
+        return parts.join(Constants.resourceDelimiter).toLowerCase();
+    }
```
keystroke: this function is added to request utils, though it could be somewhere else, and it represents the computation of an identifier for a target request and account that is used to disambiguate the request.

The current implementation of MSAL is essentially the behavior when this function returns only the target scopes from the request, as the library previously only considered scope values as the thing that differentiated different potential requests. However, authority is easily identified as another component that MUST be taken into consideration, or the library will inappropriately return the wrong token to the caller for a different authority, if such an alternate request is ongoing.

Above, I have included a few elements to consider which I suspect to be relevant in considering a request as unique or if it is logically equivalent to some other request that may be ongoing. I'm not sure if things like extra query parameters are enough to warrant a semantic difference and thus be considered in above, but in general, this PR creates the necessary abstraction for you folks to curate the logic over time.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1330,11 +1331,11 @@ export class UserAgentApplication {
...
         }
 
-        // note: scope here is clientId
-        this.registerCallback(serverAuthenticationRequest.state, this.clientId, resolve, reject);
+        const requestUid = RequestUtils.computeTokenRequestUniquenessIdentifier(serverAuthenticationRequest, account);
+        this.registerCallback(serverAuthenticationRequest.state, requestUid, resolve, reject);
         this.logger.infoPii("Navigate to:" + urlNavigate);
         frameHandle.src = "about:blank";
-        this.loadIframeTimeout(urlNavigate, frameName, this.clientId).catch(error => reject(error));
+        this.loadIframeTimeout(urlNavigate, frameName, requestUid).catch(error => reject(error));
```
keystroke: this previous logic around using the clientId as a scope seems a bit hacky. I noticed that the ctor for ServerRequestParameters will initialize the scopes array to the clientId if the request does not include it. Curious, why wasn't a scope of "openid" used for this flow? this is probably something to revist. In any case, this PR should continue to accommodate that logic.

keystroke:
Please note, I simply studied the code, made the changes, then produced this PR. I did not run tests or look at modifying them based on my changes, etc.; this PR is primarily meant as a suggestion to the kind of fix / abstraction / modification that should be considered.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -466,7 +466,6 @@ export class UserAgentApplication {
     private acquireTokenHelper(account: Account, interactionType: InteractionType, isLoginCall: boolean, request?: AuthenticationParameters, resolve?: any, reject?: any): void {
         // Track the acquireToken progress
         this.cacheStorage.setItem(TemporaryCacheKeys.INTERACTION_STATUS, Constants.inProgress);
-        const scope = request.scopes ? request.scopes.join(" ").toLowerCase() : this.clientId.toLowerCase();
```
jasonnutter: Setting `clientId` as the only scope of a request is how you tell STS that you are requesting an ID token, as opposed to an access token. So the logic here is that if you don't provide `request.scopes`, request an ID token by setting `clientId` as the only scope.

jasonnutter: (there should probably be a comment above this line explaining that)

lib/msal-core/src/utils/RequestUtils.ts
jasonnutter: We are in the planning/design phase to do work in our libraries (all MSALs, not just MSAL.js) to implement proper throttling mechanisms, and one of the things we will be doing for that is computing a "fingerprint" for each request, and using that to prevent the same request from being made repeatedly and/or unnecessarily. What you have above is actually very close to what has been proposed. 

We'll need to look at this holistically for both implicit and auth code flows (which have some different considerations), and also for the different actions in the library that behavior will control (linking multiple of the "same" concurrent requests together, which is what you have here, and whether or not to make a given request at all, which is the main driver of the throttling initiative). 

We'd love to get your feedback on that as we start building that out. cc: @jmckennon who is leading that for us.

lib/msal-core/src/UserAgentApplication.ts
keystroke: Ah yes I forgot about the ID-token only flow, I had presumed that was just a normal request with scope of "openid" but I guess that returns an access token as well. But above change actually still preserves the logic you've implicated; the lack of any scopes in request.scopes is actually initialized to [this.clientId] within the ctor of the request object itself, so all downstream code can just directly interact with the request object and reference request.scopes, without having to track that in a separate variable. In this change, that scope variable was unused, since it was only pre-computed pass to the registerCallback function, which I refactored to accept "requestUid" instead. I think in general the request object could have a function like GetScopeString() that returns 'this.scopes.join(" ").toLowerCase()' for when that format is necessary.

lib/msal-core/src/utils/RequestUtils.ts
keystroke: Thanks for taking a look so quickly! I haven't worked with typescript before so I was a bit uncertain if I was locating all the necessary areas to change. I actually haven't worked with JavaScript at all before trying-out this MSAL client for a side project; I thought the jokes about supporting IE were just memes but it is actually soul-crushing every time I scroll down to the "supported browsers" section of [some feature](https://developer.mozilla.org/en-US/docs/Web/API/URL/searchParams) or [another](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties) and find everyone EXCEPT IE supports it! haha. The TS compilation of async / await looks to be super useful so I will likely abandon my promise chaining callback hell of a sample for a compiled version ;)

It is interesting to see your team focus / concern on throttling and helping end-user-code not make too many "duplicate" requests in a short window of time. I'm more concerned on the opposite end, what if users make several distinct, concurrent requests, do they all resolve correctly? I think I had previously-mentioned, I actually "don't care so much" whether there is any optimization handled at the MSAL layer, and I would be perfectly fine to have each request resolved on its own without any de-duplication, other than the logic around returning an existing token from the cache if one is available which I obviously would prefer. But I can also appreciate how easy it is to end-up writing such code that falls into redirect-loops and whatnot so having safeguards against that can be helpful.

That bit about the cache is interesting too; I would presume that the logic for finding an existing token in the token cache should accommodate all the necessary considerations for whether a request is duplicate or not, though I guess it wound't take into account "meta-request" properties (like prompt) for such token resolution, which may be necessary for de-duping live requests, so the logic isn't exactly 1:1, but very similar I expect.

I would be interested to see what your request fingerprinting looks like; on what timeline (roughly) are you pursuing those changes?

jasonnutter:
@keystroke Yes, see #1338

PR #1291, opened 2020-02-19
Default unprotectedResources to empty array to fix error when it is undefined/null

jasonnutter:
If `unprotectedResources` is not provided, `unprotectedResources.some` throws an exception. This will make sure it defaults to an empty array if null/undefined.

### DIFF
--- a/lib/msal-angular/tests/MSALSpec.ts
+++ b/lib/msal-angular/tests/MSALSpec.ts
@@ -51,7 +51,6 @@ describe('Msal Angular Pubic API tests', function () {
                     useValue: {
                         popUp: false,
                         consentScopes: ["user.read", "mail.send"],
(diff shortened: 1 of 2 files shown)


lib/msal-angular/tests/MSALSpec.ts
```diff
@@ -51,7 +51,6 @@ describe('Msal Angular Pubic API tests', function () {
                     useValue: {
                         popUp: false,
                         consentScopes: ["user.read", "mail.send"],
-                        unprotectedResources: ["https://google.com"],
```
jasonnutter: Removing this line before the fix reproduced the bug, so keeping it removed to prove it is fixed.

jmckennon (approved):
LGTM

PR #1299, opened 2020-02-24
domain-hint changes

sameerag:
This PR changes `domain_hint` behavior for `msal js`:

- Remove all `domain_hint` code as it exists => `organizations` and `consumers` as `domain_hint` will go away
- User can pass `domain_hint` as a part of `extraQueryParameters` in the request.
  - `msal js` will pass the `domain_hint` as is for `interactive` calls
  -  For all `silent` calls, `domain_hint` is ignored
  - If the `request` already has `sid` populated, `domain_hint` is ignored (this is a `service` requirement, `service`  fails if both are passed)

Side note: `sid` is added as a hint only for `prompt=none` use cases.


lib/msal-core/src/ServerRequestParameters.ts
```diff
@@ -294,11 +245,16 @@ export class ServerRequestParameters {
...
      * @param extraQueryParameters
      */
-    static generateQueryParametersString(queryParameters: StringDict): string {
+    static generateQueryParametersString(queryParameters: StringDict, silentCall?: boolean): string {
         let paramsString: string = null;
 
         if (queryParameters) {
             Object.keys(queryParameters).forEach((key: string) => {
+                // sid cannot be passed along with login_hint or domain_hint
+                if(key == "domain_hint" && (silentCall || paramsString.match("sid"))) {
```
tnorling: Is this going to work? paramsString is null the first time through the loop. Maybe check if the "sid" key exists in queryParameters instead?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -635,9 +635,9 @@ export class UserAgentApplication {
                 request.correlationId,
             );
 
-            // populate QueryParameters (sid/login_hint/domain_hint) and any other extraQueryParameters set by the developer
+            // populate QueryParameters (sid/login_hint) and any other extraQueryParameters set by the developer
             if (ServerRequestParameters.isSSOParam(request) || account) {
-                serverAuthenticationRequest.populateQueryParams(account, request);
+                serverAuthenticationRequest.populateQueryParams(account, request, true);
```
tnorling: Missing parameter. The true here would be mapped to adalIdTokenObject instead of silentCall

lib/msal-core/src/ServerRequestParameters.ts
```diff
@@ -82,7 +82,7 @@ export class ServerRequestParameters {
      * @param request
      * @param serverAuthenticationRequest
      */
-    populateQueryParams(account: Account, request: AuthenticationParameters, adalIdTokenObject?: any): void {
+    populateQueryParams(account: Account, request: AuthenticationParameters, adalIdTokenObject?: any, silentCall?: boolean): void {
```
jasonnutter: Could `adalIdTokenObject` be changed to type `object`? That would help catch mistakenly passing `silentCall` as the third parameter.

lib/msal-core/src/ServerRequestParameters.ts
jasonnutter: Is there an existing constant we can use here instead of the magic string `"domain_hint"`? Same with `"sid"`

jasonnutter: Also, should use ===

lib/msal-core/test/ServerRequestParameters.spec.ts
```diff
@@ -128,4 +128,22 @@ describe("ServerRequestParameters.ts Class", function () {
...
+
+        it("test hints populated using queryParameters", function () {
+            const eQParams = {domain_hint: "Coke.com", locale: "en-us"};
+            const extraQueryParameters = ServerRequestParameters.generateQueryParametersString(eQParams);
+            expect(extraQueryParameters).to.include("domain_hint");
+            expect(extraQueryParameters).to.include("locale");
+        });
+
+        it("test hints populated using queryParameters", function () {
+            const eQParams = {domain_hint: "Coke.com", locale: "en-us"};
```
jasonnutter: Not sure how the `coke.com` folks would feel but might be better to use `microsoft.com` as a test case.

lib/msal-core/src/ServerRequestParameters.ts
sameerag: Thanks for the alert, this is a typo while checking in, will do the needful changes.

lib/msal-core/src/UserAgentApplication.ts
sameerag: Same as above

lib/msal-core/src/ServerRequestParameters.ts
sameerag: Yes. We have constants and we can use them. These changes are first cut.

lib/msal-core/test/ServerRequestParameters.spec.ts
sameerag: :-D I think "Coke.com" raised this issue and playing in my head. Will change it.

gabriel-kohen-jdas: Thank you for taking this on @sameerag . Any idea when it will be published?

sameerag: We stalled the release as we want this a part of 1.3.x release. We now released 1.2.2, so this should be soon merged to dev and released.

lib/msal-core/src/ServerRequestParameters.ts
sameerag: Done

lib/msal-core/test/ServerRequestParameters.spec.ts
sameerag: Addressed "Coke.com" comment

sameerag:
Addressed all comments. @jasonnutter Can you please check and approve?

lib/msal-core/src/ServerRequestParameters.ts
alex-mason-jdas: @sameerag I have tried this PR with my B2C custom policy and have a need to pass both domain_hint and login_hint. Would you explain why you are not at least allowing the developer to pass domain_hint in extraQueryParameters if they so desired. I have policy logic that is using the domain_hint. As discussed elsewhere I do also need to pass login_hint

sameerag: @alex-mason-jdas I have forwarded this requirement to B2C service team and will get back once I hear from them. As stated in the PR description, we made these changes as per AAD service recommendation and may be B2C has a different set of requirements. Will reach out once we have an answer.

PR #1320, opened 2020-02-27
Fixing navigateToLoginRequestUrl in Auth Code Library

pkanher617:
This PR fixes an issue with looping redirects during the navigateToLoginRequestUrl flow in the 2.0 version of the MSAL library.

It also changes the name of the msal-common export from `msal` to `msal-common`.


lib/msal-browser/src/interaction_handler/RedirectHandler.ts
```diff
@@ -51,29 +56,12 @@ export class RedirectHandler extends InteractionHandler {
...
-        } else {
-            window.location.hash = "";
-        }
-
         // Interaction is completed - remove interaction status.
         this.browserStorage.removeItem(BrowserConstants.INTERACTION_STATUS_KEY);
         // Handle code response.
         const codeResponse = this.authModule.handleFragmentResponse(locationHash);
+        // Hash was processed successfully - remove from cache
+        this.browserStorage.removeItem(TemporaryCacheKeys.URL_HASH);
```
jasonnutter: @pkanher617 In v1 we think there might be a race case with this and `handleRedirectCallback`. fyi @tnorling

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -114,8 +114,19 @@ export class PublicClientApplication {
         const cachedHash = this.browserStorage.getItem(TemporaryCacheKeys.URL_HASH);
         try {
             // If hash exists, handle in window. Otherwise, cancel any current requests and continue.
-            const interactionHandler = new RedirectHandler(this.authModule, this.browserStorage, this.config.auth.navigateToLoginRequestUrl);
-            const responseHash = UrlString.hashContainsKnownProperties(hash) ? hash : cachedHash;
+            const interactionHandler = new RedirectHandler(this.authModule, this.browserStorage);
+            let responseHash: string;
+            if (this.config.auth.navigateToLoginRequestUrl && UrlString.hashContainsKnownProperties(hash)) {
```
jmckennon: Add code comments

jmckennon: UrlString.hashContainsKnownProperties should be done before ifs

lib/msal-browser/src/interaction_handler/RedirectHandler.ts
```diff
@@ -41,6 +32,20 @@ export class RedirectHandler extends InteractionHandler {
         return window;
     }
 
+    navigateToRequestUrl(): void {
+        if (window.parent === window) {
+            const loginRequestUrl = this.browserStorage.getItem(TemporaryCacheKeys.ORIGIN_URI);
```
jasonnutter: I think this should be passed in as a parameter. That way this functions becomes "pure"/idempotent, which makes it easier to test and understand.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -114,8 +114,19 @@ export class PublicClientApplication {
...
+                this.browserStorage.setItem(TemporaryCacheKeys.URL_HASH, hash);
+                interactionHandler.navigateToRequestUrl();
+                return;
+            } else if (!this.config.auth.navigateToLoginRequestUrl) {
+                responseHash = UrlString.hashContainsKnownProperties(hash) ? hash : cachedHash;
+                window.location.hash = "";
+            } else {
+                responseHash = cachedHash;
+            }
```
jasonnutter: This block of code is confusing to understand, given the `if/else if/else` and mutation. From our discussion, each of these is for a different scenario, but it is hard to deduce that just from reading the code. My suggestions:

1. Regardless of changes to syntax, add comments making clear what is happening.
2. Avoid mutation if possible (not all mutation is bad, but I personally treat it as the exception, not the rule).
3. Similar to redirect, I think there should be a helper function to modifying `window.location.hash`.
4. Refactor code as follows:

```js
const validHash = UrlString.hashContainsKnownProperties(hash);
// Scenario 1
if (this.config.auth.navigateToLoginRequestUrl && validHash) {
  this.browserStorage.setItem(TemporaryCacheKeys.URL_HASH, hash);
  const loginRequestUrl = this.browserStorage.getItem(TemporaryCacheKeys.ORIGIN_URI);
  interactionHandler.navigateToRequestUrl(loginRequestUrl);
  return;
}

// Scenario 2
if (!this.config.auth.navigateToLoginRequestUrl) {
   window.location.hash = ""; // factor out into helper
}

// Scenario 2 / 3
const responseHash = validHash ? hash : cachedHash;
```

lib/msal-browser/src/interaction_handler/RedirectHandler.ts
```diff
@@ -41,6 +32,20 @@ export class RedirectHandler extends InteractionHandler {
...
+
+            // Redirect to home page if login request url is null (real null or the string null)
+            if (!loginRequestUrl || loginRequestUrl === "null") {
+                this.authModule.logger.error("Unable to get valid login request url from cache, redirecting to home page");
```
jasonnutter: This could probably be a warning (since nothing is actually breaking).

jasonnutter: (And yes I am aware I originally wrote this logic :) )

lib/msal-common/rollup.config.js
```diff
@@ -26,7 +26,7 @@ export default [
             {
                 file: "./lib/msal-common.js",
                 format: "umd",
-                name: "msal",
+                name: "msal-common",
```
jasonnutter: Should these match the npm package name `@azure/msal-common`?

lib/msal-common/src/cache/CacheHelpers.ts
```diff
@@ -165,9 +165,17 @@ export class CacheHelpers {
...
+     * @param atKey 
+     * @param clientId 
+     * @param authority 
+     * @param resource 
+     * @param homeAccountIdentifier 
+     */
     private checkForExactKeyMatch(atKey: AccessTokenKey, clientId: string, authority: string, resource?: string, homeAccountIdentifier?: string): boolean {
         const hasClientId = (atKey.clientId === clientId);
-        const hasAuthorityUri = (atKey.authority === authority);
+        const hasAuthorityUri = !StringUtils.isEmpty(authority) ? (atKey.authority === authority) : true;
```
jasonnutter: This needs a comment to explain why this should be true if `authority` is an empty string.

jasonnutter:
```suggestion
        const hasAuthorityUri = StringUtils.isEmpty(authority) || (atKey.authority === authority);
```

pkanher617: This is checking for an exact key match, if there is no authority passed then we don't want to fail the check as we will get false negatives in our cache lookups. I will add a comment.

lib/msal-browser/src/app/PublicClientApplication.ts
jasonnutter:
```suggestion
            const validHash = UrlString.hashContainsKnownProperties(hash);

            if (this.config.auth.navigateToLoginRequestUrl && validHash) {
            this.browserStorage.setItem(TemporaryCacheKeys.URL_HASH, hash);
            const loginRequestUrl = this.browserStorage.getItem(TemporaryCacheKeys.ORIGIN_URI);
            interactionHandler.navigateToRequestUrl(loginRequestUrl);
            return;
            }

            if (!this.config.auth.navigateToLoginRequestUrl) {
            window.location.hash = "";
            }

            const responseHash = validHash ? hash : cachedHash;
```

lib/msal-common/rollup.config.js
pkanher617: this is the name of the output from rollup (i.e. in the sample when you call `import as xyz`), I wanted to change it to differentiate between msal-browser output.

lib/msal-common/src/cache/CacheHelpers.ts
jasonnutter: Okay, maybe renaming the variable might help also, since its not truthful in that scenario.

lib/msal-browser/src/app/PublicClientApplication.ts
sameerag: Are we back porting (I believe only the syntax) to `msal-core`?

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: fixed both of these.

lib/msal-browser/src/interaction_handler/RedirectHandler.ts
pkanher617: done, please review the updates.

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: I've removed the if/else mutation and added some helpers instead. Please take a look.

lib/msal-browser/src/interaction_handler/RedirectHandler.ts
pkanher617: done.

lib/msal-common/src/cache/CacheHelpers.ts
pkanher617: Added a comment and your suggestion, I can't think of a good var name though, so left it unchanged. feel free to suggest something. I changed this for all lines, so that's why i didn't commit  your suggestion.

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: I've done something similar to this - please take a look

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -92,38 +92,84 @@ export class PublicClientApplication {
...
+            // Loaded page with no valid hash - pass in the value retrieved from cache, or null/empty string
+            return this.handleHash(cachedHash);
+        }
+
+        if (!this.config.auth.navigateToLoginRequestUrl) {
+            // We don't need to navigate - check for hash and prepare to process
+            BrowserUtils.clearHash();
+            return this.handleHash(hash);
+        }
+    }
```
sameerag: nitpick: comments before the `if` statements may make this more readable.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -92,38 +92,84 @@ export class PublicClientApplication {
...
+            this.authCallback(err);
+        }
+    }
+
+    /**
+     * Checks if navigateToLoginRequestUrl is set, and:
+     * - if true, performs logic to cache and navigate 
+     * - if false, handles hash string and parses response
+     */
+    private async handleRedirectResponse(): Promise<void> {
```
sameerag: general comment: Are we back porting these to `msal-core`?

lib/msal-browser/src/utils/BrowserUtils.ts
```diff
@@ -5,20 +5,37 @@
...
-    // #region Window Navigation
+    // #region Window Navigation and URL management
 
     /**
      * Used to redirect the browser to the STS authorization endpoint
      * @param {string} urlNavigate - URL of the authorization endpoint
      */
-    static navigateWindow(urlNavigate: string): void {
-        window.location.assign(urlNavigate);
+    static navigateWindow(urlNavigate: string, noHistory?: boolean): void {
```
jasonnutter: This can be done incrementally, but I think we should have an interface for navigating a browser window, since it will be different in Node, for example.

pkanher617: I think it's better to leave that entirely in the browser/node packages rather than exposing an interface in common. There is too much that goes into managing the UI that is platform specific.

pkanher617: Although I may just be overthinking the node vs browser implementation differences.

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: Adding @tnorling

lib/msal-common/rollup.config.js
jasonnutter: Right...we saw issue in `msal` where there was a mismatch between the `umd` name and the package name.

lib/msal-browser/src/utils/BrowserUtils.ts
jasonnutter: I was thinking generally about an interface for "I want to send the user to this url" and then have each platform implement how that actually happens.

jasonnutter: Although the `noHistory` nuance does make that a little more complicated, I suppose.

lib/msal-browser/src/utils/BrowserUtils.ts
```diff
@@ -5,20 +5,37 @@
...
  * Utility class for browser specific functions
  */
 export class BrowserUtils {
 
-    // #region Window Navigation
+    // #region Window Navigation and URL management
 
     /**
      * Used to redirect the browser to the STS authorization endpoint
      * @param {string} urlNavigate - URL of the authorization endpoint
```
jasonnutter: Update this comment block with new parameter.

pkanher617: done.

lib/msal-browser/src/app/PublicClientApplication.ts
tnorling: Yes, I'm looking into a related issue in msal-core so some of these changes may be ported over but I'm not able to say which ones yet

lib/msal-browser/src/utils/BrowserUtils.ts
```diff
@@ -5,20 +5,38 @@
...
+     * @param {boolean} noHistory - boolean flag, uses .replace() instead of .assign() if true
      */
-    static navigateWindow(urlNavigate: string): void {
-        window.location.assign(urlNavigate);
+    static navigateWindow(urlNavigate: string, noHistory?: boolean): void {
+        const windowParentCheck = (window.parent !== window);
+        if (windowParentCheck) {
+            // If we are not in top frame, we shouldn't redirect. This is also handled by the service.
```
jasonnutter: If this true 100% of the time? We don't want to allow `loginRedirect` and `acquireTokenRedirect` actions inside an iframe, but do we want to always block all redirects inside iframes?

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -92,38 +92,84 @@ export class PublicClientApplication {
...
+     * IMPORTANT: Please do not use this function when using the popup APIs, as it may break the response handling
      * in the main window.
      * 
      * @param {@link (AuthCallback:type)} authCallback - Callback which contains
      * an AuthError object, containing error data from either the server
      * or the library, depending on the origin of the error, or the AuthResponse object 
      * containing data from the server (returned with a null or non-blocking error).
      */
-    async handleRedirectCallback(authCallback: AuthCallback): Promise<void> {
+    async handleRedirectWithCallback(authCallback: AuthCallback): Promise<void> {
```
jasonnutter: Seems like an unrelated breaking change? We know there are improvements we want to make to this, might be good to examine holistically (including the name). If we want to change it, I would argue for something more idiomatic to JS callbacks, e.g. `onAuthRedirect` or something.

pkanher617: I meant to revert this change, I have a proposal to rename these but I will put it in a separate PR.

lib/msal-browser/src/utils/BrowserUtils.ts
pkanher617: We have this code in core for navigateToLoginRequestUrl actions, and the server will also block redirect code requests inside an iframe. We can pass this for the server to throw, or we can block requests. I'm choosing to throw an error here, but we can remove this check if needed.

jasonnutter: Given this function would be used for legitimate redirects inside an iframe, I would argue the business logic for preventing redirects in certain scenarios should go one level up.

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: reverted

lib/msal-browser/src/utils/BrowserUtils.ts
pkanher617: done.

PR #1325, opened 2020-02-28
[msal-common] Configuration and Client 

sameerag:
This PR is the first for `msal-node` and does the following:

- Move around `config` and `app` modules
- Rename them to be more generic

This will be followed up with a `PublicClientApplication` for `msal-node` and `PublicClientSPAApplication` for `msal-browser` working. Currently this branch is compilable only with `msal-common`.


lib/msal-common/src/client/Client.ts
```diff
@@ -31,18 +31,18 @@ export type ResponseStateInfo = {
...
 /**
  * AuthModule class
- * 
+ *
  * Parent object instance which will construct requests to send to and handle responses from the Microsoft STS using the authorization code flow.
- * 
+ *
  */
-export abstract class AuthModule {
+export abstract class Client {
```
jmckennon: We mentioned this already, but this name should be more specific. `InternalClient`, `PrivateClient`, `IPublicClient`? Just some suggestions

lib/msal-common/src/client/Client.ts
```diff
@@ -31,18 +31,18 @@ export type ResponseStateInfo = {
 
 /**
  * AuthModule class
```
jmckennon: Also naming inconsistency in this comment

lib/msal-common/src/client/Client.ts
pkanher617: AuthClient is what I would go with - this could describe either a public or conf client

lib/msal-common/src/client/Client.ts
sameerag: Thanks you. I am adding tests and correcting names, will update in a few hours

lib/msal-common/src/client/Client.ts
sameerag: Yes and please note that this is initial cut w.r.t folder structure. @sangonzal  will be making more changes as needed.

sameerag: Cannot be Internal or Public or Private: as all of them are misrepresentative of this base class. `AuthClient` sounds more fitting. :-) @jasonnutter and @DarylThayil ?

jasonnutter: `AuthClient` is fine. Although it would be great if we were consistent about "Auth" vs "Authentication".

sameerag: We can do `AuthenticationClient` and avoid all `auth` references. Also there will be potentially an `OAuthClient` coming as a base class - @sangonzal feel free to make the changes or let me know today and I can regarding `naming` only changes.

pkanher617: > `AuthClient` is fine. Although it would be great if we were consistent about "Auth" vs "Authentication".

This has always been a point of contention since Auth could mean "Authentication" (AuthN) or "Authorization" (AuthZ) :) Although in this case, since the client is doing both, I'd argue it is OK to name it Auth.

jasonnutter: That's fine, it's just that sometimes we name things "Auth" and sometimes its "Authentication"  (e.g. AuthenticationParameters), and it would be great it we were consistent.

pkanher617: Ah I see, yes I think we will be taking a look at naming as part of the updates to common.

DarylThayil:
/lib/.DSStore should be ignored

pkanher617:
> /lib/.DSStore should be ignored

+1

lerna.json
```diff
@@ -2,7 +2,7 @@
   "lerna": "3.16.4",
   "command": {
     "run": {
-      "scope": ["msal", "@azure/msal-common", "@azure/msal-browser"]
+      "scope": ["msal", "@azure/msal-common", "@azure/msal-browser", "@azure/msal-node"]
```
pkanher617: Did this bootstrap the packages correctly? Please test and make sure.

lib/msal-common/src/client/Client.ts
sameerag: I agree. I believe the logic behind is that we do both `Authorization` and `Authentication` and `Auth` is a good compromise for both. 

My proposal is to:
- Try referencing `Authentication` or `Authorization` as appropriate if the distinction is clear
- Avoid `Auth` as much as possible when not needed; eg: For config etc.
- `OAuth` to be used for generic protocol references as that is less confusing; mostly will be limited to the `client` folder; cc @sangonzal

lib/msal-common/src/config/Configuration.ts
```diff
@@ -20,7 +20,7 @@ const DEFAULT_TOKEN_RENEWAL_OFFSET_SEC = 300;
  * - network: this is where you can configure network implementation.
  * - crypto: implementation of crypto functions
  */
-export type ModuleConfiguration = {
+export type Configuration = {
```
pkanher617: I would call this ClientConfiguration so we can use Configuration namespace in the platform packages.

lerna.json
sameerag: `msal-browser` will not compile; but `msal-common` is compilable and tests pass. Do you want me to make another PR reflecting `msal-browser`? I can once we know this branch is merge ready.

lib/msal-common/src/config/Configuration.ts
sameerag: `ClientConfiguration` is for `client` provided params like `ClientId` etc and this is `BaseConfiguration` initializing system level modules. I am open to feedback, we can talk offline.

lerna.json
pkanher617: yeah it isn't good that one package isn't compiling, we need to make sure everything compiles before we merge this (at least to dev).

lib/msal-common/src/config/Configuration.ts
pkanher617: Then use something like BaseConfiguration or ModuleConfiguration, I want to keep changes from 1.0 to 2.0 as minimal as possible, so we should keep the name of the Configuration class to reflect as such in msal-browser. Happy to talk offline.

sameerag: It is intentional to not touch msal-browser in the same PR. I plan to follow up and group the PRs before merge and ensure linkage. The PRs are kept simple for review ease. I already am making the needed changes.

sameerag: Sorry comment is misplaced, meant for the previous section :-)

lerna.json
sameerag: It is intentional to not touch msal-browser in the same PR. I plan to follow up and group the PRs before merge and ensure linkage. The PRs are kept simple for review ease. I already am making the needed changes.

sameerag:
@DarylThayil  The ignore rules for `.gitignore` currently are:

# MSAL specific ignore files
node_modules/
typings/
lib/

Doesn't this include `/lib/.DSStore`?

jasonnutter:
A pattern for .DS_Store files should be added to the root gitignore (doesn't appear to be there), and the ds_store file in this commit should be deleted.

sameerag:
Ah! I was looking at the wrong place!

lerna.json
pkanher617: Just FYI lerna is now running in our CI, so adding this will run node as part of CI build and test. Make sure this is something you want before committing.

lib/msal-common/src/config/Configuration.ts
sameerag: This is addressed in a future PR

PR #1328, opened 2020-03-01
[DO NOT MERGE - Base branch] [msal-node] scaffolding - tsdx

sameerag:
This PR adds the node build/lint/test environment.

We are choosing the `msal-node` environment based on typescript's recommendation for node projects [here](https://www.typescriptlang.org/docs/home.html). This is to simplify  environment for `msal-node` and to leverage the support [tsdx](https://github.com/jaredpalmer/tsdx#readme) has already inbuilt.

We did slight modifications to add test paths as `jest` does not work out of the box.


lib/msal-node/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2020 sameerag
```
jasonnutter: This should be Microsoft

lib/msal-node/README.md
```diff
@@ -0,0 +1,33 @@
+# TSDX Bootstrap
+
+This project was bootstrapped with [TSDX](https://github.com/jaredpalmer/tsdx).
```
jasonnutter: Can you update this top part to mention MSAL Node?

lib/msal-node/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "version": "0.1.0",
```
jasonnutter: I would start even lower, `0.0.1`

lib/msal-node/.gitignore
```diff
@@ -0,0 +1,264 @@
+## Ignore Visual Studio temporary files, build results, and
+## files generated by popular Visual Studio add-ons.
+
+# User-specific files
```
jmckennon: This file addition appears in both this PR and #1325, potential MCs

lib/msal-node/README.md
```diff
@@ -0,0 +1,33 @@
+# TSDX Bootstrap
```
pkanher617: You should copy the file from MSAL browser and make updates to it here. You can even remove all the text, but just keep the formatting to make it easier. That way we have the same templates across all platforms.

lib/msal-node/package.json
pkanher617: +1

lib/msal-node/LICENSE
sameerag: Addressed. Thanks.

lib/msal-node/README.md
sameerag: https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/1332 is raised to track this for accuracy. Can we please limit this PR for node scaffolding only?

lib/msal-node/.gitignore
sameerag: Thanks @jmckennon. This branch will be the base for #1325 so we are covered.

lib/msal-node/package.json
sameerag: Done

lib/msal-node/package.json
```diff
@@ -0,0 +1,56 @@
...
+  "typings": "dist/index.d.ts",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "start": "tsdx watch",
+    "build": "tsdx build",
+    "test": "tsdx test .*.spec.*",
+    "test:watch": "tsdx test .*.spec.* --watch",
+    "test:coverage": "tsdx test .*.spec.* --coverage",
```
pkanher617: I'm going to defer to you on what we use here, but we need to make sure this can be integrated with the coverage of the other libraries. Otherwise we will have to make exceptions for node, which defeats the purpose of this work.

sameerag: Rasied #1372 to track this.

sameerag:
Rebasing this to a new branch `msal-node-base`

PR #1338, opened 2020-03-03
[WIP] msal-common supports throttling based on thumbprints

jmckennon:
WIP PR for throttling. I'll be leaving comments throughout with questions I have for the team, responses/comments appreciated!

4/3/2020 EDIT:
* #1436 needs to be merged before continuing to work on this. We want to be able to differentiate between interactive and silent requests so that we can properly deal with `interaction_required` errors for the scope of throttling. 
* This code will need a few fixes and needs to be manually and unit tested in accordance with 2.0 being in dev now.
* `sendGetRequestAsync` should also be wrapped similarly to `sendPostRequestAsync`
* `XhrClient` needs to accommodate the interface changes just as `FetchClient` does in browser.
* @sameerag and @sangonzal have also requested specific conversations happen before any merges into dev based on their large ongoing changes in order to work out merge conflicts.


lib/msal-common/src/network/NetworkManager.ts
```diff
@@ -0,0 +1,80 @@
...
+import { AccessTokenKey } from "../cache/AccessTokenKey";
+import { RequestThumbprint } from "./RequestThumbprint";
+import { Constants } from "../utils/Constants";
+import { ICrypto } from "../crypto/ICrypto";
+import { TokenExchangeParameters } from "../request/TokenExchangeParameters";
+import { ClientInfo } from "../auth/ClientInfo";
+import { ErrorValue } from "../cache/ErrorValue";
+
+export type NetworkResponse = {
+    headers: Headers;
```
jmckennon: I haven't tested this, but should `headers` instead be represented by an something like `Map<string, string>` to accomodate XHR/IE?

lib/msal-common/src/network/NetworkManager.ts
```diff
@@ -0,0 +1,80 @@
...
+    body: string;
+    status: number;
+};
+
+export class NetworkManager {
+    cacheStorage: ICacheStorage;
```
jmckennon: would it be better to instead pass in a `CacheHelpers` and implement the related functions in that class, rather than directly writing to the cache `cacheStorage` here?

lib/msal-common/src/network/RequestThumbprint.ts
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+import { ClientInfo } from "../auth/ClientInfo";
+import { ICrypto } from "../crypto/ICrypto";
+import { StringUtils } from "../utils/StringUtils";
+
+export class RequestThumbprint {
```
jmckennon: this feels odd to have in addition to `AccessTokenKey` since they are basically the same fields--though they represent two different ideas. suggestions?

lib/msal-common/src/network/NetworkManager.ts
jasonnutter: Yeah, I would not rely on classes that are specific to one of the platforms.

lib/msal-common/src/network/INetworkModule.ts
```diff
@@ -31,5 +33,5 @@ export interface INetworkModule {
      * @param requestParams 
      * @param enableCaching 
      */
-    sendPostRequestAsync<T>(url: string, options?: NetworkRequestOptions): Promise<T>;
+    sendPostRequestAsync(url: string, options?: NetworkRequestOptions): Promise<NetworkResponse>;
```
jasonnutter:
```js
sendPostRequestAsync<T>(url: string, options?: NetworkRequestOptions): Promise<NetworkResponse<T>>;
```

lib/msal-common/src/network/NetworkManager.ts
```diff
@@ -0,0 +1,80 @@
...
+import { RequestThumbprint } from "./RequestThumbprint";
+import { Constants } from "../utils/Constants";
+import { ICrypto } from "../crypto/ICrypto";
+import { TokenExchangeParameters } from "../request/TokenExchangeParameters";
+import { ClientInfo } from "../auth/ClientInfo";
+import { ErrorValue } from "../cache/ErrorValue";
+
+export type NetworkResponse = {
+    headers: Headers;
+    body: string;
```
jasonnutter:
```js
export type NetworkResponse<T> = {
  body: T
}
```

lib/msal-common/src/network/NetworkManager.ts
```diff
@@ -0,0 +1,80 @@
...
+        const networkResponse = await this.networkClient.sendPostRequestAsync(tokenEndpoint, options);
+
+        this.postProcess(networkResponse, thumbprint, clientInfo);
+
+        return networkResponse.body as ServerAuthorizationTokenResponse;
+    }
+
+    private postProcess(response: NetworkResponse, thumbprint: RequestThumbprint, clientInfo: ClientInfo): void {
+        if (response.status >= 500 || response.status == 429 || response.status >= 300 && response.headers.has("Retry-After")) {
+            const throttleTime = response.headers.get("Retry-After") || Date.now() + Constants.DEFAULT_THROTTLE_TIME_MS;
```
jasonnutter: Will the `Rety-After` header be a number or a string?

lib/msal-common/src/network/NetworkManager.ts
```diff
@@ -0,0 +1,80 @@
...
+            const responseBody = JSON.parse(response.body);
```
jasonnutter: This is already parsed.

lib/msal-browser/src/network/FetchClient.ts
```diff
@@ -30,15 +30,19 @@ export class FetchClient implements INetworkModule {
      * @param headers 
      * @param body 
      */
-    async sendPostRequestAsync<T>(url: string, options?: NetworkRequestOptions): Promise<T> {
+    async sendPostRequestAsync(url: string, options?: NetworkRequestOptions): Promise<NetworkResponse> {
```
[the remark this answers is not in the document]
jmckennon: oh, fancy--will add

lib/msal-common/src/network/NetworkManager.ts
jmckennon: not sure yet, im aware of this though. need to find a way to repro that header coming through or just get the actual info from the ests team

lib/msal-common/src/network/NetworkManager.ts
jmckennon: whoops, will fix

lib/msal-common/src/network/NetworkManager.ts
sameerag: We are changing this for `msal-node`

lib/msal-browser/src/network/FetchClient.ts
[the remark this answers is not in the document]
jmckennon: So I made this series of changes--however, including the changes in this file was making it not compile.. the rest of it seems to be now compiling and I think it still gets the same point across--using a generic on the `body`. Changes are pushed in most recent commit

lib/msal-common/src/network/NetworkManager.ts
jmckennon: @sameerag can you elaborate?

lib/msal-browser/src/network/FetchClient.ts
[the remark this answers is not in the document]
jmckennon: nevermind, VSCode wasn't updating the imported `NetworkResponse` type. fixing again

lib/msal-common/src/network/NetworkManager.ts
```diff
@@ -0,0 +1,76 @@
...
+            const throttleTime = response.headers.get("Retry-After") || Date.now() + Constants.DEFAULT_THROTTLE_TIME_MS;
+
+            const errorValue = new ErrorValue(
+                response.body.error,
+                response.body.error_description,
+                response.body.error_codes.join(" "),
+                throttleTime
+            );
+
+            this.cacheStorage.setItem(JSON.stringify(thumbprint), JSON.stringify(errorValue));
```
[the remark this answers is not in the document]
jmckennon: Resolved offline, pushed changes for this. Value will optionally contain error/throttle info and will rewrite the correct value info with it into the cache if needed.

lib/msal-common/src/network/NetworkManager.ts
```diff
@@ -0,0 +1,86 @@
...
+        this.cacheStorage = cacheStorage;
+        this.networkClient = networkClient;
+        this.cryptoObj = cryptoObj;
+    }
+
+    async sendPostRequest(tokenEndpoint: string, tokenReqParams: ServerTokenRequestParameters, tokenRequest: TokenExchangeParameters, clientInfo: ClientInfo, clientId: string): Promise<ServerAuthorizationTokenResponse> {
+        const thumbprint = new AccessTokenKey(
+            tokenRequest.authority, 
+            clientId, 
+            tokenRequest.scopes.join(" "),
```
keystroke: given you will JSON stringify this to act as the cache key, should these string all be normalized with .toLower, and also for scopes, they should be sorted? Or perhaps the AccessTokenKey class should have a method / property for getting or computing the fingerprint (and thus, these details could be handled internally to that class)?

lib/msal-common/src/network/NetworkManager.ts
```diff
@@ -0,0 +1,86 @@
...
+        const networkResponse = await this.networkClient.sendPostRequestAsync(tokenEndpoint, options);
+
+        this.postProcess(networkResponse, thumbprint);
+
+        return networkResponse.body as ServerAuthorizationTokenResponse;
+    }
+
+    private postProcess(response: NetworkResponse<ServerAuthorizationTokenResponse>, thumbprint: AccessTokenKey): void {
+        if (response.status >= 500 || response.status == 429 || (response.status < 200 || response.status >= 300) && response.headers.has("Retry-After")) {
+            const throttleTime = parseInt(response.headers.get("Retry-After")) || Date.now() + Constants.DEFAULT_THROTTLE_TIME_MS;
```
keystroke: Don't you need to save the time at which the error response was received for this to have meaning? if I pull this item out of the cache, and it has a throttleTime property on it with related errors, what does that mean? how would that context "know" whether the throttle time has elapsed or not at the time the cache item is resolved? Or is the signal then for that cache item "observer" to wait for the specified number of seconds on top of however long the item has been cached? Or perhaps the acting of caching the item adds a cache dateTime property to it?

lib/msal-common/src/network/NetworkManager.ts
```diff
@@ -0,0 +1,86 @@
...
+                parsedValue.refreshToken,
+                parsedValue.expiresOn,
+                parsedValue.extExpiresOn,
+                response.body.error,
+                response.body.error_description,
+                response.body.error_codes.join(" "),
+                throttleTime
+            );
+
+            this.cacheStorage.setItem(cacheKey, JSON.stringify(accessTokenValue));
```
keystroke: Apologies if my understanding of the design and code is way-off here, just trying to wrap my head around these changes.

So is the idea here is that if multiple areas of my are all concurrently generating "overlapping" token requests, and the first one fails with a retry-after type error, this code will cache the result of that token request with the error details, and thus, when another overlapping token request is generated, it will find that representation of the failure in the cache for that request, and can deduce from the time the error was cached, and the retry seconds, how long it should wait before attempting another token request? Am I understanding this correctly? What about that first request that receives a retry-after error, where is the code / engine that is retrying this?

My initial thought thinking about this feature would be to have a "singleton promise" for each unique (based on thumbprint calculation)  ongoing token request. That promise internally handles the logic around making the request, including retrying under the necessary circumstances, and finally resolving with success or error to end-user callbacks. Then, if multiple token requests come-in concurrently which overlap (based on fingerprint calculations), they would each be returned the same "singleton promise" instance to await (promise branching). However, is throttling data separate for each token / authority? My presumption would be that the server is throttling the specific user / client, meaning the throttling would need to apply to all token requests throughout the app, including across multiple MSAL instances? Or is the throttling scoped to a specific authority and set of scopes, or a specific client?

I understand this is WIP and the whole concurrent token requests may not be explicitly focused on in this iteration, but that's all I can think about in the overall flow, and it isn't clear to me why implementing throttling would be anything other than adding an internal retry to the request promise that is making the call to AAD (which could internally check the error conditions and set a timeout to make the request again in X number of seconds).

Anyway sorry for long comment.

lib/msal-common/src/network/NetworkManager.ts
jmckennon: These are all valid points. @pkanher617 do you have insight here based on your testing/knowledge so far?

lib/msal-common/src/network/NetworkManager.ts
jmckennon: I haven't explicitly tested this yet because I need to get some information about the field from the server team as I don't have a way of directly reproing these errors at this moment. That being said, my assumption was that they'd be passing us a value in unix time; so when we go to retrieve this from the cache, we just compare the current time (with a `Date.now()`) to the the retrieved value.

If the case is that we have NOT passed that retrieved unix time, then we will return the given errors, which are the same as the ones given for the original throttling from the server. If we have passed the retrieved unix time, we will send through the request like normal and clear the error/throttlingTime content from the cache.

It is a bit janky to not just have a separate key-value pair for erroneous cases, but we wanted to prioritize space in local storage. I'm willing to be convinced, though.

lib/msal-common/src/network/NetworkManager.ts
jmckennon: So your first paragraph of assumptions is almost correct. There's no mechanism in place that will automatically retry after x amount of seconds. This is on the user of msal to figure out. Getting a `Retry-After` will only happen in extreme cases (like if the server is down, or if a user is totally misusing these endpoints) or when user interaction is required to alleviate a problem. The whole point of this set of PRs I'm working on is to alleviate server strain, and all msals will be implementing the same spec. There will be a `preProcess` function that implements some of this idea where we will be looking in the cache for throttling information related to a thumbprint on every request to the server.

The throttling is scoped to the information in `AccessTokenKey`. This is an idea of a "strict request thumbprint", or a unique thumbprint. That means if you have a request that gets throttled, only requests of the exact same values of the 5 fields on `AccessTokenKey` will be throttled. Any change in these fields will still allow requests through.

The idea is not to throttle every request, even from multiple instances/apps, but to attempt to pin down the exact request that's happening, and throttle that particular request. I think I can share the spec with you if you're interested.

Also, for clarity: This set of PRs is meant to implement this idea of throttling. As a byproduct of that, we get this idea of a "unique request thumbprint", which, to my knowledge, is basically what your previous PR was for. We'll be iterating on this and using the idea of a unique thumbprint for a few things moving forward, specifically telemetry.

Hopefully that clears things up--we appreciate the feedback, and hopefully I got the gist of what you're getting at here. Happy to address further questions/concerns.

lib/msal-common/src/network/NetworkManager.ts
keystroke: Ah, I misread the code the first time. You fallback to Date.now() + const so as long as the retry-after header is the absolute unix time as well then it should be fine.

lib/msal-common/src/network/NetworkManager.ts
keystroke: Ah okay thanks for clarifying; let me see if I understood that correctly.

If my application is making many requests ~concurrently from different areas of my app (all for the same token), let's say to acquire token silent, and the situation is they all fail with "consent required". In today's version of MSAL, the MSAL client will call the server each time to receive the error (as no existing token would be found in the cache). I think there is a little optimization here, as there already exists some code to "merge" ongoing concurrent token requests together based on target scopes. But if the incoming token requests were all staggered slightly (e.g. made serially), MSAL client would call the server each time.

Assuming above is correct (just one example of a failure condition that we are interested in here) you are looking to implement some throttling within MSAL. If MSAL sees many "overlapping" token requests in short time periods, the success case isn't a problem, as all the callers will just get the token from the cache. But for error cases, the current code doesn't "known" that these requests are "doomed" to fail. So you are saving something in a networking cache, to indicate that a particular token request has been made repeatedly recently, and has failed for whatever reason, and so you want to throttle any additional token requests attempts that overlap with the original failed request. Is this actually throttling though? It looks like you aren't simply adding a delay before making the request, but that you would literally resolve with "new" request with the same error that the "previous" token request received?

Basically, extending the concept of the token cache to include recent failure results, instead of just recent success results (the access tokens themselves).

Is that correct? One challenge I see with that is around cache invalidation; how would MSAL know if user proceeded to grant consent via an immediate interactive request? Would subsequent silent token requests continue failing with consent required failures, even though user has granted the consent? Or maybe consent type errors aren't a good example for when you want to throttle? You'd have to know for how "long" an interaction-required error should last, or would have to integrate somehow with the acquire token redirect / popup flows and detect when consent was granted to invalidate the silent token errors in the cache? Or perhaps those consent grants return access tokens that you place in the cache, and this fact "prevents" "stale" cached error data from being activated, as an existing token would be found in that case?

I guess the prime question is resolving precisely which error conditions are relevant here. Is the idea that these conditions would be exposed to end-user code? E.g. when I call acquireTokenSilent, will my catch handler need to handle scenarios where my request is being throttled via a special exception type that I can check for? Or would the goal be to "hide" this fact from end-user code and just delay the promise resolution?

One of the things I am trying to build is a multi-cloud web client, that can sign the user into all their azure directories and Azure Stack directories in one web page. So they can see all their subscriptions and resources. This is what has been motivating me to get all these concurrent token request fixes. But I also need to heavily rely on the dynamic consent capabilities, as many areas of my app will not work initially as user would need to grant consent for each cloud and each directory along the way. In some cases, I may have read access, but not write access. etc. The results in my app generating a lot of acquiretokensilent requests in a short period of time, and I "rely" on them failing with interaction_required errors so I can render the appropriate error view for that cloud, with a button to grant consent for that cloud. It looks like the fix you are making would make it so that if the user was opening and closing an area of my app that made this check repeatedly, you wouldn't need to call the server each time to receive the missing permission error, and instead could return that error result from the cache itself?

Something like retry-after sounds like it could be handled completely internally to MSAL, and just "waiting" to resolve my acquire token promise until that retry-after duration has expired, and the request can be attempted again. Or is the plan that you would throw this exception to me, and I would have to deal with waiting? I suppose we'd need to clarify the expectations of the acquire token methods. Is it reasonable that the promise could take > 2 minutes to resolve in a particular throttling case? How would the token request parameter for "forceRefresh" interact with the throttling mechanism? Would the token request parameters be extended for user to set a maximum timeout duration on the token request? etc.

Anyway I'm starting to ramble now, lol. Thanks for taking the time to clarify! Don't let me distract you too much though.

jasonnutter: The scope of this PR (which is part of an effort across all the MSAL versions, not just JS) is to prevent sending silent requests to the server that are either guaranteed to fail (because user interaction is required) or because the specific user is making too many of the same request. This is to reduce COGS on our services, and help prevent improper usage of the library.

And yes, if a request with a given fingerprint is resolved, silent requests would then be allowed.

We will build on this logic to properly handle concurrent requests (i.e. using the same fingerprint).

keystroke: @jasonnutter I interpret your statement as:

Failed silent token requests will not be retried for a minimum duration of "Constants.DEFAULT_THROTTLE_TIME_MS" (120 seconds), unless a redirect/popup request of the same fingerprint completes successfully during that time window. Such successful token requests would not-only cache the resulting access token, but would also invalidate the cached error details for that request.

Is this being scoped specifically to token request failures like interaction required where you positively identify the failure scenario as one that should be throttled? Or will ALL failed silent token requests, regardless of failure case (outside of checking for special scenarios like retry-after), be cached and induce the 2-minute lock-out?

I'm not sure I see the place in the PR where the existing errors are "found" in the cache, and then the request is not made; is that still being worked-on in a subsequent commit, or did I just not see it here?

jasonnutter: > Is this being scoped specifically to token request failures like interaction required where you positively identify the failure scenario as one that should be throttled?

I believe this is correct, @jmckennon can confirm.

And yeah, this is a WIP PR, so assume that part isn't implemented yet (as Jake noted in an earlier response, there will be a `preprocess` function that will do this, which hasn't been implemented).

jmckennon: @keystroke yes, only certain failures will cause a throttle. Everything @jasonnutter said is true.

As you've identified, there are some specific challenges around `interaction_required` which I haven't quite worked out yet. I'll need some more details from another team for that particular case.

> Something like retry-after sounds like it could be handled completely internally to MSAL, and just "waiting" to resolve my acquire token promise until that retry-after duration has expired, and the request can be attempted again. Or is the plan that you would throw this exception to me, and I would have to deal with waiting?

An exception will be thrown to you. This let's the user control how an app reacts to the failure cases that come through. The point of this is not to make requests wait 120s before trying again, it's to disallow subsequent requests when there is a problem that needs to be solved to reduce load on the server. Again, this isn't quite there in this PR yet. I'll be adding more to this PR + maybe a few other branches off this PR this week.

lib/msal-common/src/app/module/AuthorizationCodeModule.ts
```diff
@@ -402,14 +403,10 @@ export class AuthorizationCodeModule extends AuthModule {
...
     private async getTokenResponse(tokenEndpoint: string, tokenReqParams: ServerTokenRequestParameters, tokenRequest: TokenExchangeParameters, codeResponse?: CodeResponse): Promise<TokenResponse> {
-        // Perform token request.
-        const acquiredTokenResponse = await this.networkClient.sendPostRequestAsync<ServerAuthorizationTokenResponse>(
-            tokenEndpoint,
-            {
-                body: tokenReqParams.createRequestBody(),
-                headers: tokenReqParams.createRequestHeaders()
-            }
-        );
+        const clientInfo = buildClientInfo(this.cacheStorage.getItem(PersistentCacheKeys.CLIENT_INFO), this.cryptoObj);
```
sameerag: I am assuming `CLIENT_INFO` is already cached from the response here? Please throw if there is an error here, there should be no case of `CLIENT_INFO` being null as per eSTS instructions.

sameerag: Rather throw basically wherever you think you are reading `CLIENT_INFO` from the NW for the first response if not already done.

jmckennon: what is NW in this statement?

lib/msal-common/src/cache/AccessTokenValue.ts
```diff
@@ -7,20 +7,30 @@
...
+    expiresOnSec?: string;
+    extExpiresOnSec?: string;
 
-    tokenType: string;
-    accessToken: string;
-    idToken: string;
-    refreshToken: string;
-    expiresOnSec: string;
-    extExpiresOnSec: string;
+    // These fields exist when a thumbprint is being throttled
```
sameerag: Can we not have the same object for success and error for toke successful token response? can we have a common key but register token and errors as different Objects? I can sync up with you offline @jmckennon to discuss this.

lib/msal-common/src/cache/AccessTokenValue.ts
```diff
@@ -7,20 +7,30 @@
...
         this.tokenType = tokenType;
         this.accessToken = accessToken;
         this.idToken = idToken;
         this.refreshToken = refreshToken;
         this.expiresOnSec = expiresOn;
         this.extExpiresOnSec = extExpiresOn;
+        this.error = error;
+        this.errorDescription = errorDescription;
+        this.errorCodes = errorCodes;
+        this.throttleTime = throttleTime;
```
sameerag: Same as above!

lib/msal-common/src/cache/AccessTokenValue.ts
jmckennon: So I talked about this with Prithvi, and we can't have a common key because they will just overwrite each other, i.e. only an access token or only an error can be in the cache at any given time for a particular key, whereas we want to be able to have both. We'd need to key them differently--which also ends up taking up a lot of space. That's why I went with this implementation.

lib/msal-browser/src/network/FetchClient.ts
```diff
@@ -30,15 +31,19 @@ export class FetchClient implements INetworkModule {
      * @param headers 
      * @param body 
      */
-    async sendPostRequestAsync<T>(url: string, options?: NetworkRequestOptions): Promise<T> {
+    async sendPostRequestAsync<T>(url: string, options?: NetworkRequestOptions): Promise<NetworkResponse<T>> {
```
sangonzal: Should we also be wrapping `sendGetRequestAsync` response in NetworkResponse?

jasonnutter: Yep

jmckennon:
#1907 is the new solution for this. Closing

PR #1339, opened 2020-03-04
Split the state correctly for temp cache clearance

### LINKED ISSUE #1333: Temporary local storage entries are not cleared when token request has "state" value
## Library
- [msal@1.2.1] 

## Framework
If you are using a framework, please provide the framework and version (e.g. Angular v8, React v16, etc).
React 16.8.1

## Description
Local storage use grows uncontrollably when silent token fetch request has "state" value passed as part of AuthenticationParameters object. Eventually it results in localStorage exhaustion and failure to fetch tokens in case localStorage is used to cache tokens locally. There is no simple recovery except for calling logout().

## Security
No

## Regression
Not sure

## Configuration
Please provide your MSAL configuration options.

```js
      cache: {
        storeAuthStateInCookie: false
      },
```

## Expected behavior
MSAL should automatically clean up any temporary local storage entries which it uses to facilitate silent token fetch. it fails to do so when "state" value passed as part of AuthenticationParameters object. 

When token request has state then MSAL use it as key which is embedded into local storage keys which are used to store temporary state. The key is generated in the following format {guid}|{state} where guid is generated by MSAL.

The problem is in implementation of resetTempCacheItems method:

image

More specifically, line 138 determine key state based on last occurrence of '|' symbol but does not account for the fact that state itself has this symbol inside it.

## Browsers
Is this issue browser-specific? If so, please detail which browsers are impacted (e.g. IE 11, Safari).
Not browser specific

sameerag:
This PR fixes #1333 where the `msal-core` lib assumes that "state" is always a GUID and does not have a delimiter within itself.

We fixed this by making sure that the `guid` the `msal` attaches before sending a request to `server` is always appended at the beginning and we rectify the split logic to reflect this.

Note: tests will be added before merging this.

### DIFF
--- a/lib/msal-core/src/cache/AuthCache.ts
+++ b/lib/msal-core/src/cache/AuthCache.ts
@@ -133,14 +133,10 @@ export class AuthCache extends BrowserStorage {// Singleton
         let key: string;
         // check state and remove associated cache
         for (key in storage) {
-            if (!state || key.indexOf(state) !== -1) {
-                const splitKey = key.split(Constants.resourceDelimiter);
-                const keyState = splitKey.length > 1 ?  splitKey.slice(1).join("|"): null;
-                if (keyState === state && !this.tokenRenewalInProgress(keyState)) {
-                    this.removeItem(key);
-                    this.setItemCookie(key, "", -1);
-                    this.clearMsalCookie(state);
-                }
+            if ((!state || key.indexOf(state) !== -1) && !this.tokenRenewalInProgress(state)) {
+                this.removeItem(key);
+                this.setItemCookie(key, "", -1);
+                this.clearMsalCookie(state);
             }
         }
         // delete the interaction status cache
--- a/lib/msal-core/src/cache/AuthCache.ts
+++ b/lib/msal-core/src/cache/AuthCache.ts
@@ -135,7 +135,7 @@ export class AuthCache extends BrowserStorage {// Singleton
         for (key in storage) {
             if (!state || key.indexOf(state) !== -1) {
                 const splitKey = key.split(Constants.resourceDelimiter);
-                const keyState = splitKey.length > 1 ? splitKey[splitKey.length-1]: null;
+                const keyState = splitKey.length > 1 ?  splitKey.slice(1).join("|"): null;
                 if (keyState === state && !this.tokenRenewalInProgress(keyState)) {
                     this.removeItem(key);
                     this.setItemCookie(key, "", -1);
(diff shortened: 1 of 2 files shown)


lib/msal-core/src/cache/AuthCache.ts
```diff
@@ -135,7 +135,7 @@ export class AuthCache extends BrowserStorage {// Singleton
         for (key in storage) {
             if (!state || key.indexOf(state) !== -1) {
                 const splitKey = key.split(Constants.resourceDelimiter);
```
AndreiZe: Would it be simpler to just do key.endsWith(state) check instead of doing splits? state always has Guid which should prevent any accidental matches. If you are concerned with older browsers then you already have UrlUtils.endsWith() method

sameerag: Good suggestion. May be that works too .. we wanted to add more params for state and hence did it this way, @jasonnutter  can you please follow up?

sameerag: Okay, we thought about it and I do not want to use key.endsWith(state) as this will only validate if the param we are checking `against` has state but not the exact structure we expect it to have. That would be an incomplete check - necessary but not sufficient (even if it is simple). 

Hence I am leaning towards the original solution for now.

jasonnutter: Turns out, we should just check if the state string is present in the string, as there shouldn't be a scenario where the state string is in the key but not at the end. Making this logic not dependent on the resource delimiter both solves this bug, but is also future proof if we want to change the format of the state string (which we are considering).

jasonnutter:
To summarize: we can take the state string as-is, and remove any local storage values whose key includes the string.

lib/msal-core/src/cache/AuthCache.ts
```diff
@@ -133,14 +133,10 @@ export class AuthCache extends BrowserStorage {// Singleton
...
         for (key in storage) {
-            if (!state || key.indexOf(state) !== -1) {
-                const splitKey = key.split(Constants.resourceDelimiter);
-                const keyState = splitKey.length > 1 ? splitKey[splitKey.length-1]: null;
-                if (keyState === state && !this.tokenRenewalInProgress(keyState)) {
-                    this.removeItem(key);
-                    this.setItemCookie(key, "", -1);
-                    this.clearMsalCookie(state);
-                }
+            if ((!state || key.indexOf(state) !== -1) && !this.tokenRenewalInProgress(state)) {
```
sameerag: Did we test end-to-end? Will this pass for `concurrent` silent calls failing (at the msal-core execution) for any reason?

jasonnutter: Yes, I tested with the React sample. Can you clarify the concern with failing concurrent requests?

jasonnutter:
Approving on behalf of @sameerag, who can't approve as the original author.

lib/msal-core/src/cache/AuthCache.ts
sameerag: This isn't a concern, I was thinking if there will be any cache based on state we do not want to get rid off, that should not be the case.

PR #1343, opened 2020-03-04
Add redirectStartPage property to AuthenticationParameters to allow apps to indicate which page triggered the redirect. 

jasonnutter:
Fixes Angular guard bug where using redirect would send the user to the redirect URI, instead of the resource they were trying to reach.

### DIFF
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -497,17 +497,8 @@ export class UserAgentApplication {
         acquireTokenAuthority.resolveEndpointsAsync().then(async () => {
             // On Fulfillment
             const responseType: string = isLoginCall ? ResponseTypes.id_token : this.getTokenType(account, request.scopes, false);
-            let loginStartPage: string;
 
-            if (isLoginCall) {
-                // if the user sets the login start page - angular only??
-                loginStartPage = this.cacheStorage.getItem(`${TemporaryCacheKeys.ANGULAR_LOGIN_REQUEST}${Constants.resourceDelimiter}${request.state}`);
-                if (!loginStartPage || loginStartPage === "") {
-                    loginStartPage = window.location.href;
-                } else {
-                    this.cacheStorage.setItem(`${TemporaryCacheKeys.ANGULAR_LOGIN_REQUEST}${Constants.resourceDelimiter}${request.state}`, "");
-                }
-            }
+            const loginStartPage = request.redirectStartPage || window.location.href;
 
             serverAuthenticationRequest = new ServerRequestParameters(
                 acquireTokenAuthority,
--- a/lib/msal-angular/src/msal-guard.service.ts
+++ b/lib/msal-angular/src/msal-guard.service.ts
@@ -38,7 +38,10 @@ export class MsalGuard implements CanActivate {
                     .catch(() => false);
             }
 
+            const routePath = `${window.location.origin}${state.url}`;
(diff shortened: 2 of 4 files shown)


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -497,17 +497,8 @@ export class UserAgentApplication {
         acquireTokenAuthority.resolveEndpointsAsync().then(async () => {
             // On Fulfillment
             const responseType: string = isLoginCall ? ResponseTypes.id_token : this.getTokenType(account, request.scopes, false);
-            let loginStartPage: string;
 
-            if (isLoginCall) {
```
jasonnutter: One thing I noticed: this logic didn't run for `acquireTokenRedirect`, which doesn't seem correct. Anyone know why this was only happening for `loginRedirect`?

jmckennon: This may be exactly what you mean, but `isLoginCall` is only set to true when coming in from the login APIs and not the redirect APIs, which makes sense based on the name... was maybe just being misused here

jmckennon (approved):
Looks good, makes sense to me, though I don't think I have the answer to your question

lib/msal-core/src/UserAgentApplication.ts
pkanher617: I actually think this may be a bug, ideally we should redirect whether or not we are making a login call.

jasonnutter: One other thing I realized: for Angular, I don't think we would not want the redirect to occur if there is an error (so instead redirect back to the page that initiated the redirect, not the page they are trying to access). And maybe that logic should live in the Angular lib.

jasonnutter: Or at least the MsalGuard should be smart enough not to trigger loginRedirect if the page received an error response in the hash.

jasonnutter: (which would cause a login loop)

pkanher617: Shouldn't that be up to the user of the lib, as it is in core? (i.e. handleResponse(), catch if error and don't call loginRedirect). The error will come back to the redirectUri same as the token response, so the logic should be the same and branch wherever the actual response handling occurs for error vs success.

pkanher617 (approved):
Looks fine, tests appear to work. I would consider adding some documentation somewhere for these params.

lib/msal-core/src/UserAgentApplication.ts
sameerag: AFAIK the LOGIN_REQUEST  is cached separately <I forget the context> and since <old> angular wrapper is written a certain way, the below cache item for ANGULAR_LOGIN_REQUEST is created. It makes sense to see holistically if we need to identify login calls at all vs token acquisition calls in the cache (I remember me and Prithvi discussing this long back).

This basically narrows down to keeping `Constants.TemporaryCacheKeys.LOGIN_REQUEST` in the cache.

sameerag (approved):
Approving it on the basis that no matter what the choice is for LOGIN_REQUEST in cache, this change still cleans up the interfacing in this use case between `msal-angular` and `msal-core`

jasonnutter:
@pkanher617 Will definitely be added to typedocs. I would consider it advanced usage, so not sure it needs to be included in other samples.

PR #1351, opened 2020-03-05
Always send back the accessToken and scopes if the response includes them

### LINKED ISSUE #1293: AcquireTokenPopup for someone other than logged on user
## Library
- [X] `msal@1.x.x` or `@azure/msal@1.x.x`

## Description
This question relates to `msal@1.2.x` even though we have SPA which uses `@azure/msal-angular@0.1.4-beta.2` and which we're trying to upgrade to `@azure/msal-angular@1.0.0-beta.2` (therefore bringing in the `msal@1.2.x` peer dependency).

Previously when we were using `@azure/msal-angular@0.1.4-beta.2`, we took a dependency of `@azure/msal-angular/msal@0.2.4`. Our use case worked with `@azure/msal-angular/msal@0.2.4` and it no longer works with `msal@1.2.x`.

Our use case is as follows. There is a main logged on user in our app who performs various activities in the app and at some point needs to perform a privileged operation that requires elevated permissions. The user does not have those permissions and needs to call over a manager to type in his or her credentials. Barring the inherent risk of the manager typing his or her credentials on a user workstation (which we plan on curing by enabling 2FA for the manager), this scenario worked fine when we used `acquireTokenPopup` in `@azure/msal-angular/msal@0.2.4` using the following code:
```
const user = this.authService.getUser();
return from(this.authService.acquireTokenPopup(
  ['api://[our_api_uri]/user_impersonation`],
  'https://login.microsoftonline.com/[tenantId],
   user,
   'prompt=login&forceRefresh=true'
));
```
The manager would sign in using his or her own credentials and this would give us an access token which we then used to call our API to execute the privileged operation as the manager. This did not affect the currently logged on user and after the privileged operation was executed, the user would continue performing functions in the app as himself/herself. The manager therefore provided an access token which was used one time, just for the privileged operation.

After upgrading to `msal@1.2.2-beta.0` which is a peer dependency of `@azure/msal-angular@1.0.0-beta.2`, we needed to change the above code to the following:
```
const account = this.authService.getAccount();
return from(this.authService.acquireTokenPopup({
  scopes: [environment.api.audience + '/user_impersonation'],
  account: account,
  prompt: 'login',
  forceRefresh: true
})).pipe(
  map(authResponse => authResponse.accessToken)
);
```

Seems like a simple enough upgrade, right? Unfortunately not. The `acquireTokenPopup` returns a `null` `accessToken` in the updated code. After digging around and debugging the code inside `UserAgentApplication`, I found that even though Azure AD was returning an `accessToken` for the correct account (the manager) in the URL hash, it was being discarded in the [following code](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/1212c640aa619718328a5f6931f14ecf27ea0b36/lib/msal-core/src/UserAgentApplication.ts#L1493-L1504) within the `saveTokenFromHash` method in `UserAgentApplication`:
```
// Process access_token
if (hashParams.hasOwnProperty(ServerHashParamKeys.ACCESS_TOKEN)) {
    this.logger.info("Fragment has access token");

    // retrieve the id_token from response if present
    if (hashParams.hasOwnProperty(ServerHashParamKeys.ID_TOKEN)) {
        idTokenObj = new IdToken(hashParams[ServerHashParamKeys.ID_TOKEN]);
        response.idToken = idTokenObj;
        response.idTokenClaims = idTokenObj.claims;
    } else {
        idTokenObj = new IdToken(this.cacheStorage.getItem(PersistentCacheKeys.IDTOKEN));
        response = ResponseUtils.setResponseIdToken(response, idTokenObj);
    }
```
So in the above, `hashParams[ServerHashParamKeys.ACCESS_TOKEN]` is not used at all in the resultant response. What we get back is actually a response with the `idToken` set for the currently logged on user instead of a response with an `accessToken` for the manager.

The questions are therefore:

1. Is this a bug, and if not,
2. Is there a way that we can get an access token for the manager using `msal@1.2.x` without affecting the currently logged on user and also without affecting any associated session or local storage for the user?

jasonnutter:
Per #1293, there is a regression in `msal@1` where we aren't always including the accessToken in the response if one is included. See ticket for complete context.

Open question: Is there any reason not to do that? Later on in the processing, `saveAccessToken` does [include the token](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/9ba8124ecaed5a5a5897c45ae92e2d02448fd56a/lib/msal-core/src/UserAgentApplication.ts#L1390), however, `saveAccessToken` is not called in this scenario.

### DIFF
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -1495,6 +1495,7 @@ export class UserAgentApplication {
                 // Process access_token
                 if (hashParams.hasOwnProperty(ServerHashParamKeys.ACCESS_TOKEN)) {
                     this.logger.info("Fragment has access token");
+                    response.accessToken = hashParams[ServerHashParamKeys.ACCESS_TOKEN];
 
                     // retrieve the id_token from response if present
                     if (hashParams.hasOwnProperty(ServerHashParamKeys.ID_TOKEN)) {
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -1497,6 +1497,10 @@ export class UserAgentApplication {
                     this.logger.info("Fragment has access token");
                     response.accessToken = hashParams[ServerHashParamKeys.ACCESS_TOKEN];
 
+                    if (hashParams.hasOwnProperty(ServerHashParamKeys.SCOPE)) {
+                        response.scopes = hashParams[ServerHashParamKeys.SCOPE].split(" ");
+                    }
+
                     // retrieve the id_token from response if present
                     if (hashParams.hasOwnProperty(ServerHashParamKeys.ID_TOKEN)) {
                         idTokenObj = new IdToken(hashParams[ServerHashParamKeys.ID_TOKEN]);


jmckennon (approved):

Ivan-L:
Hey @jasonnutter thanks for this PR!

It just occurred to me that it may also be good to have the `scopes` field for the response set as well, since it would be consistent with the rest of the places where an access token is set on an `AuthResponse`.

So I would add the following line as well:
```
if (hashParams.hasOwnProperty(ServerHashParamKeys.SCOPE)) {
    response.scopes = hashParams[ServerHashParamKeys.SCOPE].split(" ");
}
```

The total change would then be:
```
response.accessToken = hashParams[ServerHashParamKeys.ACCESS_TOKEN];
if (hashParams.hasOwnProperty(ServerHashParamKeys.SCOPE)) {
    response.scopes = hashParams[ServerHashParamKeys.SCOPE].split(" ");
}
```

For reference, this is what the `hashParams` look like in our case:
image

pkanher617 (approved):

sameerag:
@Ivan-L I don't believe we want to send scopes in response, I will check internally to clarify if there is any alternative if this cannot be supported.

sameerag:
Let me clarify my question, I see some merit in adding scopes in response in some cases and curious about your use case.

Ivan-L:
Hi @sameerag 

I only made the suggestion to include the scopes so that the fix in this PR is consistent with the rest of the codebase. In other parts of `UserAgentApplication` such as [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/9ba8124ecaed5a5a5897c45ae92e2d02448fd56a/lib/msal-core/src/UserAgentApplication.ts#L1391), [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/9ba8124ecaed5a5a5897c45ae92e2d02448fd56a/lib/msal-core/src/UserAgentApplication.ts#L1402) and [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/9ba8124ecaed5a5a5897c45ae92e2d02448fd56a/lib/msal-core/src/UserAgentApplication.ts#L1237), the `scopes` field is always set alongside the `accessToken` field in `AuthResponse` instances.

In addition, the `MsalService` of msal-angular extends `UserAgentApplication` and broadcasts when tokens are acquired successfully ([here](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/9ba8124ecaed5a5a5897c45ae92e2d02448fd56a/lib/msal-angular/src/msal.service.ts#L102), [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/9ba8124ecaed5a5a5897c45ae92e2d02448fd56a/lib/msal-angular/src/msal.service.ts#L116) and [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/9ba8124ecaed5a5a5897c45ae92e2d02448fd56a/lib/msal-angular/src/msal.service.ts#L134)). If an app does subscribe to the `msal:acquireTokenSuccess` event, it would be consistent to have the scopes as part of the `authResponse` in all cases. That way, subscribers can tell which downstream API the successfully acquired access token is for, by looking at the scopes.

Nevertheless, this PR in its current form is sufficient to solve #1293 and scopes are not needed.

sameerag:
@Ivan-L You are right! Sorry I missed that the response is already sending scopes in other cases. :-). @jasonnutter the suggestion makes sense, please review and add accordingly.

jasonnutter:
@sameerag @Ivan-L @jmckennon Updated, please review, thanks!

Ivan-L:
Hi @jasonnutter 

LGTM.

I cloned this repo, checked out this particular branch and tested our app against it. Issue #1293 is resolved in this branch for us, thanks for the help!

Ivan-L (approved):

sameerag:
LGTM. Approved.

sameerag (approved):

Issue #1306, opened 2020-02-25
No timeout error if authority is not reachable


franzk:
## Library
- [x] `msal@1.2.1`

## Description
There's no timeout error if the authority is not reachable. `WindowUtils::monitorWindowForHash` loops forever, because https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/c7c5d1e2c88f7bcbf6f108099f8fc06350773223/lib/msal-core/src/utils/WindowUtils.ts#L61 always matches. I'd consider a timeout to be a hard limit, whatever happens. I came to this issue because Safari 13's ITP blocks the iframe's request and the same thing happens, without any timeout.

## Security
No.

## Regression
Dunno.

## Configuration

```js
{
  auth: {
    authority: 'https://login.microsoftonline.com/' +
      process.env.REACT_APP_TENANT + '/',
    clientId: process.env.REACT_APP_CLIENT_ID,
    navigateToLoginRequestUrl: false,
    redirectUri: window.location.origin
  },
  cache: {
    cacheLocation: 'localStorage'
  },
  system: {
    loadFrameTimeout: 15000
  }
};
```

## Reproduction steps
1. Add this entry to your /etc/hosts file:
```
127.0.0.1 login.microsoftonline.com
```
2. Call `acquireTokenSilent`

## Expected behavior
I expect a timeout error after the configured period (default 6000 ms).

## Browsers
Happens in my Chromium 80.0.3987.122 (Official Build) Arch Linux (64-bit). Firefox (73.0.1) at least throws "ClientAuthError: Error: could not resolve endpoints. Please check network and try again. Details: 0" earlier in the process.

franzk:
BTW: My workaround is to have my own timeout around `acquireTokenSilent`, but this is silly, as the loop keeps running and the iframe remains.

AndreiZe:
We ran into this as well. 

We are seeing this issue when silent token fetch is called and there is network issue. In addition, there has to be cached response for openid-configuration endpoint; otherwise, endpoints_resolution_error is returned. I believe that the issue is due to the fact that with network error hidden iframe that is used to get token response from login.microsoftonline.com never gets redirected back to reply url. The code that Franz references is the exact spot where issue happens: getting contentWindow.location.href will always fail due to CORS. This results into internal MSAL timeout logic being bypassed. 

This is very serious issue. 

First, this creates infinite loop so token fetch never completes.  
[...]

Issue #1230, opened 2020-01-21
msal@v1.2 - When acquireTokenSilent fails, it never times out, and promise never resolves


Tim91084:
## Library
- [x ] `msal@1.2.x`

## Framework
React@latest

## Description
acquireTokenSilent doesn't fail/time out properly when 3rd party cookies are disabled.

## Security
No

## Regression
Yes, works in 1.1.x

## Reproduction steps
You can reproduce this by disabling 3rd party cookies and then trying to acquire a token silently using the library.  You'll see that the library tries to make a request here:  https://login.microsoftonline.com/[tenent]/oauth2/v2.0/undefined

This won't work and should fail immediately, but for some reason the msal library is not handling this properly.  My app depends on this failure/timeout to then issue an acquireTokenPopup.

## Expected behavior
When acquire token silent fails, it should properly resolve/reject the promise

## Browsers
Tested in Chrome

Draccoz:
Encountered this issue as well.

Chryogenic:
This happens to me as well. Does anyone have a workaround?

jasonnutter:
@Tim91084 Apologies for the delay in responding. I have some theories about what is happening, I'll investigate and follow up.

Chryogenic:
Hi @jasonnutter, were you able to look into this issue? Thanks!

jasonnutter:
@Chryogenic Not yet, I'll follow up when I have.

mbrevda:
This seems to be happening for as well since updating to 1.2.1

jasonnutter:
The fix for this is available in `msal@1.2.2-beta.2`, please try that version and confirm that it has been fixed, thanks!

jasonnutter:
And to be clear, the new version above ensures MSAL will at worst timeout. The specific issue of `https://login.microsoftonline.com/[tenent]/oauth2/v2.0/undefined` requests we believe is an issue with the service, not the library, and the service team is investigating this now.

PR #1354, opened 2020-03-06
Ensure silent operations timeout if the iframe never returns to the app domain

jasonnutter:
There are scenarios where the iframe used for `acquireTokenSilent` would not return to the app domain. `monitorWindowForHash` is naively written to assume the iframe will always return to the app domain at some point, which is when the timeout clock would start. For silent calls, the time spent on STS should be short, and we need to ensure that at worst, `acquireTokenSilent` resolves with a timeout error.

This PR changes `monitorWindowForHash` to ensure it will timeout for silent calls if the iframe never returns to the app domain.

I believe this is the cause for #1306 and #1230.

### DIFF
--- a/lib/msal-core/src/utils/WindowUtils.ts
+++ b/lib/msal-core/src/utils/WindowUtils.ts
@@ -35,7 +35,7 @@ export class WindowUtils {
      * Monitors a window until it loads a url with a hash
      * @ignore
      */
-    static monitorWindowForHash(contentWindow: Window, timeout: number, urlNavigate: string): Promise<string> {
+    static monitorWindowForHash(contentWindow: Window, timeout: number, urlNavigate: string, isSilentCall?: boolean): Promise<string> {
         return new Promise((resolve, reject) => {
             const maxTicks = timeout / WindowUtils.POLLING_INTERVAL_MS;
             let ticks = 0;
@@ -57,15 +57,27 @@ export class WindowUtils {
                     href = contentWindow.location.href;
                 } catch (e) {}
 
-                // Don't process blank pages or cross domain
-                if (!href || href === "about:blank") {
-                    return;
+                if (isSilentCall) {
+                    /*
+                     * Always run clock for silent calls
+                     * as silent operations should be short,
+                     * and to ensure they always at worst timeout.
+                     */
+                    ticks++;
+                } else {
+                    // Don't process blank pages or cross domain
+                    if (!href || href === "about:blank") {
+                        return;
+                    }
+
+                    /*
+                     * Only run clock when we are on same domain for popups
+                     * as popup operations can take a long time.
+                     */
+                    ticks++;
                 }
 
-                // Only run clock when we are on same domain
-                ticks++;
-
-                if (UrlUtils.urlContainsHash(href)) {
+                if (href && UrlUtils.urlContainsHash(href)) {
                     clearInterval(intervalId);
                     resolve(contentWindow.location.hash);
                 } else if (ticks > maxTicks) {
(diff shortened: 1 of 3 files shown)


lib/msal-core/src/utils/WindowUtils.ts
```diff
@@ -57,15 +57,27 @@ export class WindowUtils {
...
+                     * Always run clock for silent calls
+                     * as silent operations should be short,
+                     * and to ensure they always at worst timeout.
+                     */
+                    ticks++;
+                } else {
+                    // Don't process blank pages or cross domain
+                    if (!href || href === "about:blank") {
+                        return;
+                    }
```
jmckennon: To clarify: was the reason that silent operations weren't timing out was because they'd always return on this statement here and not increment `ticks`?

jmckennon (commented):
Looks like the build is failing, but I'm happy to approve once that's fixed!

lib/msal-core/src/utils/WindowUtils.ts
sameerag: I think we never return from the server domain.

lib/msal-core/src/utils/WindowUtils.ts
```diff
@@ -57,15 +57,27 @@ export class WindowUtils {
...
-                ticks++;
+                    /*
+                     * Only run clock when we are on same domain for popups
+                     * as popup operations can take a long time.
+                     */
+                    ticks++;
+                }
 
-                if (UrlUtils.urlContainsHash(href)) {
+                if (href && UrlUtils.urlContainsHash(href)) {
```
sameerag: Are we still error-ing out with the URL messaging type where the user can check why the silent calls failed?

sameerag: What happens if `redirectUri` is "about:blank"? I assume`href` remains valid.

sameerag (approved):
Left some comments but approving.

lib/msal-core/src/utils/WindowUtils.ts
jasonnutter: @jmckennon Correct.

lib/msal-core/src/utils/WindowUtils.ts
jasonnutter: Yep.

jasonnutter: Yeah, `href` is valid but the `urlContainsHash` will return false.

pkanher617 (approved):

lib/msal-core/src/utils/WindowUtils.ts
jmckennon: Cool, thanks for the info

arsnyder16:
@jasonnutter Is there a tentative schedule for when there is going to be a new release cut that contains this fix?  I am using msal in a few different contexts and in one i am hosting inside a CEF based browser and i see this on each `acquireTokenSilent` if the there is no auth cookie

arsnyder16:
Just going to add some details on what i am seeing, in case you are interesting in any underlying issues. Here seems to be the sequence of requests 
```
Request URL:https://login.microsoftonline.com/common/oauth2/v2.0/authorize?...
Request Method:GET
Status Code:200 OK

Request URL:https://login.live.com/Me.htm?v=2
Request Method:GET
Status Code:200 

Request URL:https://login.microsoftonline.com/common/instrumentation/reportstaticmecontroltelemetry?...
Request Method:POST
Status Code:200 OK

Request URL:https://login.microsoftonline.com/common/oauth2/v2.0/undefined
Request Method:GET
Status Code:404 Not Found
```

PR #1355, opened 2020-03-06
If framework.protectedResourceMap/unprotectedResources is empty, use msalAngularConfig.protectedResourceMap/unprotectedResources

### LINKED ISSUE #1341: MsalInterceptor not attaching tokens
## Library
- [x] `@azure/msal-angular@1.0.0-beta.2`

## Framework
Angular 7.2

## Description
The MsalInterceptor is not attaching tokens to API calls. 
I have added the logger callback with console logging and log level set to Verbose to identify where its failing. The log shows that its unable to find the url in the protectedResourceMap array even though it has been passed.

Note: I am passing the protectedResourceMap and unprotectedResources to the MsalAngularConfiguration and not the framework part of MSAL's Configuration.

## Configuration

```
MsalModule.forRoot({
      auth: {
        authority: environment.auth.authority,
        clientId: environment.auth.clientId
      },
      cache: {
        cacheLocation: environment.auth.cacheLocation,
        storeAuthStateInCookie: true,
      },
      system: {
        loadFrameTimeout: environment.auth.loadFrameTimeout
      }
    }, {
      consentScopes: environment.auth.initialConsentScopes,
      protectedResourceMap: environment.auth.autoProtectedResources,
      unprotectedResources: environment.auth.manuallyProtectedResource.prodBotApi
    })
```

jasonnutter:
Ensures that is `msalConfig.framework.protectedResourceMap` is an empty map, to still fallback to `msalAngularConfig.protectedResourceMap`. 

Fixes #1341

### DIFF
--- a/lib/msal-angular/src/msal.service.ts
+++ b/lib/msal-angular/src/msal.service.ts
@@ -173,11 +173,12 @@ export class MsalService extends UserAgentApplication {
             return null;
         }
 
-        if (this.msalConfig.framework && this.msalConfig.framework.protectedResourceMap) {
+        const frameworkProtectedResourceMap = this.msalConfig.framework && this.msalConfig.framework.protectedResourceMap;
+        if (frameworkProtectedResourceMap) {
             this.getLogger().info("msalConfig.framework.protectedResourceMap is deprecated, use msalAngularConfig.protectedResourceMap");
         }
 
-        const protectedResourceMap = (this.msalConfig.framework && this.msalConfig.framework.protectedResourceMap) || new Map(this.msalAngularConfig.protectedResourceMap);
+        const protectedResourceMap = frameworkProtectedResourceMap && frameworkProtectedResourceMap.size ? frameworkProtectedResourceMap : new Map(this.msalAngularConfig.protectedResourceMap);
 
         // process all protected resources and send the matched one
         const keyForEndpoint = Array.from(protectedResourceMap.keys()).find(key => endpoint.indexOf(key) > -1);
--- a/lib/msal-angular/src/msal.service.ts
+++ b/lib/msal-angular/src/msal.service.ts
@@ -70,7 +70,10 @@ export class MsalService extends UserAgentApplication {
     }
 
     private isUnprotectedResource(url: string): boolean {
-        const unprotectedResources = (this.msalConfig.framework && this.msalConfig.framework.unprotectedResources) || this.msalAngularConfig.unprotectedResources || [];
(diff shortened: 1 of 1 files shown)


jmckennon (approved):
LGTM!

lib/msal-angular/src/msal.service.ts
```diff
@@ -173,11 +173,12 @@ export class MsalService extends UserAgentApplication {
...
         }
 
-        if (this.msalConfig.framework && this.msalConfig.framework.protectedResourceMap) {
+        const frameworkProtectedResourceMap = this.msalConfig.framework && this.msalConfig.framework.protectedResourceMap;
+        if (frameworkProtectedResourceMap) {
             this.getLogger().info("msalConfig.framework.protectedResourceMap is deprecated, use msalAngularConfig.protectedResourceMap");
         }
 
-        const protectedResourceMap = (this.msalConfig.framework && this.msalConfig.framework.protectedResourceMap) || new Map(this.msalAngularConfig.protectedResourceMap);
+        const protectedResourceMap = frameworkProtectedResourceMap && frameworkProtectedResourceMap.size ? frameworkProtectedResourceMap : new Map(this.msalAngularConfig.protectedResourceMap);
```
sameerag: Do we even need to have `protectedResourceMap` in core? Isn't the point to have this only in `Angular` or is it maintained for surface compatibility and non breaking changes?

sameerag (changes requested):
Please check the comment and let me know.

lib/msal-angular/src/msal.service.ts
jasonnutter: Right, keeping it to avoid breaking changes.

pkanher617 (approved):

lib/msal-angular/src/msal.service.ts
sameerag: okay. Thanks.

sameerag (approved):

lib/msal-angular/src/msal.service.ts
```diff
@@ -173,11 +173,12 @@ export class MsalService extends UserAgentApplication {
             return null;
         }
 
-        if (this.msalConfig.framework && this.msalConfig.framework.protectedResourceMap) {
+        const frameworkProtectedResourceMap = this.msalConfig.framework && this.msalConfig.framework.protectedResourceMap;
```
abishekreddy: We should add a length check for this.msalConfig.framework.unprotectedResources too (Line 173 of this file).

jasonnutter: Ah yes, good catch, thanks!

PR #1358, opened 2020-03-09
handleRedirectCallback updates

tnorling:
- Moved processing of the hash into constructor
- handleRedirectCallback is no longer required. If called it will just return response saved during processCallback step in constructor
- If second redirect is needed, no processing will be done after first redirect. Hash will be passed to second redirect and processed then
- Fixed unnecessary redirect if window is already on the page that initiated the request


jasonnutter:
I'll look at the code in detail shortly, but you can also make sure to update this part of the README: https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev/lib/msal-core#usage

I think we'll want to call out the specific version (1.3.x) when this became optional.

jasonnutter:
And these two pages: https://docs.microsoft.com/en-us/search/index?search=handleredirectcallback

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1043,31 +1039,24 @@ export class UserAgentApplication {
...
                     const loginRequestUrl = this.cacheStorage.getItem(`${TemporaryCacheKeys.LOGIN_REQUEST}${Constants.resourceDelimiter}${stateInfo.state}`, this.inCookie);
+                    const currentUrl = this.getCurrentUrl();
 
                     // Redirect to home page if login request url is null (real null or the string null)
                     if (!loginRequestUrl || loginRequestUrl === "null") {
                         this.logger.error("Unable to get valid login request url from cache, redirecting to home page");
                         window.location.href = "/";
-                    } else {
-                        window.location.href = loginRequestUrl;
+                    } else if (currentUrl != loginRequestUrl) {
```
jasonnutter:
```suggestion
                    } else if (currentUrl !== loginRequestUrl) {
```

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1043,31 +1039,24 @@ export class UserAgentApplication {
...
+                        window.location.href = loginRequestUrl + locationHash;
```
jasonnutter:
```suggestion
                        window.location.href = `${loginRequestUrl}${locationHash}`;
```

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: Triple (not) equals is preferred over double equals, as it doesn't perform type coercion.

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: String interpolation is preferred over concatenation, as it can be easier to read and less error prone.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1922,6 +1912,16 @@ export class UserAgentApplication {
         return this.config.auth.redirectUri;
     }
 
+    /**
+     * Use to get the current url without the hash.
+     * This will be compared to request url to determine if redirect is needed
+     *
+     * @returns {string} current URL
+     */
+    public getCurrentUrl(): string {
```
jasonnutter: Can this go somewhere else other than `UserAgentApplication`? e.g. `UrlUtils`. We would like to have a few functions as possible here. And then add tests :)

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1043,30 +1039,24 @@ export class UserAgentApplication {
         } else {
             // Redirect cases
             tokenResponseCallback = null;
+            window.location.hash = "";
             // if set to navigate to loginRequest page post login
             if (this.config.auth.navigateToLoginRequestUrl) {
```
pkanher617: I think it might be easier to pull this check and handling code out of the handleAuthenticationResponse function and perform these checks in the constructor or a separate function.

lib/msal-core/src/UserAgentApplication.ts
tnorling: There was already an existing function in `UrlUtils` that does the same thing so I just renamed the existing function to be more usage agnostic and removed this function from `UserAgentApplication`

lib/msal-core/src/UserAgentApplication.ts
tnorling: @pkanher617 I separated handleAuthenticationResponse into 2 functions. One for the iframe/popup use cases and the other for redirect use cases. I think the flow is easier to follow and understand. Thanks for the suggestion

lib/msal-core/README.md
```diff
@@ -110,7 +110,7 @@ Before using MSAL.js you will need to [register an application in Azure AD](http
...
 
-After instantiating your instance, if you plan on using a redirect flow (`loginRedirect` and `acquireTokenRedirect`), you must register a callback handlers  using `handleRedirectCallback(authCallback)` where `authCallback = function(AuthError, AuthResponse)`. The callback function is called after the authentication request is completed either successfully or with a failure. This is not required for the popup flows since they return promises.
+After instantiating your instance, if you plan on using a redirect flow in MSAL 1.2.x or earlier (`loginRedirect` and `acquireTokenRedirect`), you must register a callback handler using `handleRedirectCallback(authCallback)` where `authCallback = function(AuthError, AuthResponse)`. As of MSAL 1.3.0 this is optional. The callback function is called after the authentication request is completed either successfully or with a failure. This is not required for the popup flows since they return promises.
```
pkanher617: have we decided on versioning for this already?

tnorling: I think the consensus was this would be included in 1.3.x but this can be updated as appropriate before we merge

pkanher617: sounds good.

keystroke:
sweet, this behavior always confused me

PR #1368, opened 2020-03-12
[msal-node] crypto module

sameerag:
This PR adds 'crypto' module for `msal-node`. 

Libraries in use:
- [npm uuid](https://www.npmjs.com/package/uuidlibrary) for GUID generation 
- [crypto](https://nodejs.org/api/crypto.html) module of `Node.js` for `PKCECodes` generation and `sha256` hash generation
- [buffer](https://nodejs.org/api/buffer.html) module of `Node.js` for `Base64` encode and decode

Tests:
- All unit tests added
- 100% coverage
- Please test with `npm run test` and `npm run test:c` for coverage

crypto-test-coverage


lib/msal-node/src/utils/NodeConstants.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+/**
+ * http methods
+ */
+export enum method {
+    GET = 'get',
+    POST = 'post',
+}
+
```
jasonnutter: I don't think this needs to be named `NodeConstants`, `Constants` is probably fine.

lib/msal-node/src/utils/NodeConstants.ts
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+/**
+ * http methods
+ */
+export enum method {
```
jasonnutter: Can this be `HttpMethod`, to be more specific?

lib/msal-node/package.json
```diff
@@ -28,9 +30,10 @@
     "start": "tsdx watch",
     "build": "tsdx build",
     "test": "tsdx test .*.spec.*",
-    "test:watch": "tsdx test .*.spec.* --watch",
-    "test:coverage": "tsdx test .*.spec.* --coverage",
-    "lint": "tsdx lint",
+    "test:n": "tsdx test",
+    "test:w": "tsdx test .*.spec.* --watch",
+    "test:c": "tsdx test .*.spec.* --coverage",
+    "lint": "tsdx lint --fix",
```
jasonnutter: I would argue to lint be its own script: `"lint:fix": "npm run lint -- --fix"`

lib/msal-node/package.json
```diff
@@ -28,9 +30,10 @@
     "start": "tsdx watch",
     "build": "tsdx build",
     "test": "tsdx test .*.spec.*",
-    "test:watch": "tsdx test .*.spec.* --watch",
-    "test:coverage": "tsdx test .*.spec.* --coverage",
-    "lint": "tsdx lint",
+    "test:n": "tsdx test",
+    "test:w": "tsdx test .*.spec.* --watch",
+    "test:c": "tsdx test .*.spec.* --coverage",
```
jasonnutter: I would argue to spell out the words, e.g. `watch`, `coverage`, and then have the main `npm test` command run the default test command.

lib/msal-node/src/index.ts
```diff
@@ -1,2 +1,11 @@
...
 export { ConfidentialClientApplication } from './client/ConfidentialClientApplication';
+
+// crypto
+export { CryptoOps } from './crypto/CryptoOps';
+export { GuidGenerator } from './crypto/GuidGenerator';
+export { PkceGenerator } from './crypto/PkceGenerator';
+
+// encode
+export { Base64Encode } from './encode/Base64Encode';
+export { Base64Decode } from './encode/Base64Decode';
```
jasonnutter: Do these need to be available on the public API?

lib/msal-node/src/encode/Base64Decode.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export class Base64Decode {
+    /**
+     * 'utf8': Multibyte encoded Unicode characters. Many web pages and other document formats use UTF-8.
+     * 'base64': Base64 encoding.
+     *
+     * @param base64Str Base64 encoded text
+     */
+    static decode(base64Str: string): string {
+        return Buffer.from(base64Str, 'base64').toString('utf8');
```
jasonnutter: I think it would be good to have constants for both of these strings.

lib/msal-node/src/crypto/CryptoOps.ts
```diff
@@ -0,0 +1,54 @@
...
+import { GuidGenerator } from './GUidGenerator';
+import { Base64Encode } from './../encode/Base64Encode';
+import { Base64Decode } from './../encode/Base64Decode';
+import { PkceGenerator } from './PkceGenerator';
+
+/**
+ * This class implements MSAL node's crypto interface, which allows it to perform base64 encoding and decoding, generating cryptographically random GUIDs and
+ * implementing Proof Key for Code Exchange specs for the OAuth Authorization Code Flow using PKCE (rfc here: https://tools.ietf.org/html/rfc7636).
+ */
+export class CryptoOps implements ICrypto {
```
DarylThayil: What does the name `CryptoOps` mean?

lib/msal-node/test/TestConstants.ts
```diff
@@ -0,0 +1 @@
+export const TESTS = 100;
```
jasonnutter: I feel like this isn't specific enough.

lib/msal-node/src/crypto/PkceGenerator.ts
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * https://tools.ietf.org/html/rfc7636#page-8
+ */
+export class PkceGenerator {
+    /**
+     * generates the codeVerfier and the challenge from the codeVerfier
+     * reference: https://tools.ietf.org/html/rfc7636#section-4.1 and https://tools.ietf.org/html/rfc7636#section-4.2
+     */
+    async generatePkceCodes(): Promise<PkceCodes> {
```
jasonnutter: Can these be static (to be consistent with other classes)?

lib/msal-node/src/crypto/GuidGenerator.ts
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+import { v4 as uuidv4 } from 'uuid';
+
+export class GuidGenerator {
```
DarylThayil: Should this class be in common? we use guids a bit in browser lib

lib/msal-node/src/crypto/PkceGenerator.ts
```diff
@@ -0,0 +1,67 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { PkceCodes } from '@azure/msal-common';
+import { PKCEConstants } from './../utils/NodeConstants';
+import { Base64Encode } from './../encode/Base64Encode';
+const crypto = require('crypto');
```
DarylThayil:
```js
import crypto from `crypto`;
```

lib/msal-node/src/crypto/PkceGenerator.ts
```diff
@@ -0,0 +1,67 @@
...
+ */
+export class PkceGenerator {
+    /**
+     * generates the codeVerfier and the challenge from the codeVerfier
+     * reference: https://tools.ietf.org/html/rfc7636#section-4.1 and https://tools.ietf.org/html/rfc7636#section-4.2
+     */
+    async generatePkceCodes(): Promise<PkceCodes> {
+        const v: string = this.generateCodeVerifier();
+        const c: string = this.generateCodeChallengeFromVerifier(v);
+        return { verifier: v, challenge: c };
```
DarylThayil:
```js
const verifier: string = this.generateCodeVerifier();
const challenge: string = this.generateCodeChallengeFromVerifier(verifier);
return { verifier, challenge };
```

lib/msal-node/src/crypto/PkceGenerator.ts
```diff
@@ -0,0 +1,67 @@
...
+
+import { PkceCodes } from '@azure/msal-common';
+import { PKCEConstants } from './../utils/NodeConstants';
+import { Base64Encode } from './../encode/Base64Encode';
+const crypto = require('crypto');
+
+/**
+ * https://tools.ietf.org/html/rfc7636#page-8
+ */
+export class PkceGenerator {
```
DarylThayil: same question about common since we do auth and pcke in browser

lib/msal-node/src/crypto/CryptoOps.ts
jasonnutter: I would probably argue for a different name ("crypt-oops", Freudian slip?), but its consistent with MSAL browser: https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/1107/files#diff-33a6cab8b15c1d5ebe0810440a955ca9

lib/msal-node/src/encode/Base64Decode.ts
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+export class Base64Decode {
```
DarylThayil: common?

lib/msal-node/src/crypto/PkceGenerator.ts
jasonnutter: I don't think the `: string` parts are needed either

lib/msal-node/src/encode/Base64Encode.ts
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+export class Base64Encode {
```
DarylThayil: Why have two seperate classes ? why not just have a Base64Utils or EncodingUtils

lib/msal-node/src/index.ts
DarylThayil: Why does node export these?

DarylThayil: aren;t they internal

lib/msal-node/src/encode/Base64Encode.ts
jasonnutter: +1 for `EncodingUtils`

lib/msal-node/test/TestConstants.ts
DarylThayil: This probably doesn't need to be constant, can put this in test file because the number or times you want to validate things may change depending on the test.  Also the naming could be clearer

lib/msal-node/src/crypto/CryptoOps.ts
sameerag: The naming, did not have a choice since the interface defined it already. I am fine with anything the team decides.

lib/msal-node/src/utils/NodeConstants.ts
sameerag: Will we have namespace issues as msal-common also has a Constants file? May be not, I can change this in that case

lib/msal-node/src/utils/NodeConstants.ts
sameerag: +1

lib/msal-node/package.json
sameerag: yes, did this as it was annoying me, will do.

lib/msal-node/package.json
sameerag: okay! No preference, this was for typing convenience. I wonder if we can create aliases for commands.

lib/msal-node/src/encode/Base64Decode.ts
sameerag: +1

lib/msal-node/src/index.ts
sameerag: Yes for `encodeUrl()` and `decodeUrl()` as we do not have that interface in `CryptoOps` for some reason. I wanted to check this with @pkanher617 .

lib/msal-node/test/TestConstants.ts
sameerag: Can change the name, it is mostly stress testing codes we generate.

lib/msal-node/src/crypto/PkceGenerator.ts
sameerag: well, honestly node doesn't need it `async` but my hands are tied as the interface defines this as `async`. I can take up a separate PR to change the interface which would include changes in `msal-browser`

lib/msal-node/src/crypto/GuidGenerator.ts
sameerag: well it can be, if @pkanher617 agrees. I can move this to common and have both libraries use the same.

lib/msal-node/src/crypto/PkceGenerator.ts
sameerag: This cannot be, as node.crypto is not available in browsers. These need to stay separate.

lib/msal-node/src/encode/Base64Decode.ts
sameerag: Same, I am using node internal libraries. I am not sure how `browser` can be different. @jasonnutter @pkanher617 to comment.

lib/msal-node/src/encode/Base64Encode.ts
sameerag: I wondered the same, I was constrained by the interface provided by `msal-common` and the pattern in `msal-browser`, this would need to change in all three components if we want it this way.

lib/msal-node/src/index.ts
sameerag: Replied above for Jason's comment: "Yes for encodeUrl() and decodeUrl() as we do not have that interface in CryptoOps for some reason. I wanted to check this with @pkanher617"

lib/msal-node/test/TestConstants.ts
sameerag: yeah, agree. Will move.

lib/msal-node/package.json
sameerag: Done.

lib/msal-node/package.json
sameerag: Done

lib/msal-node/src/encode/Base64Decode.ts
sameerag: Buffer does not accept a 'string', only a set of predetermined strings. 'base64' and 'uint8' being a part of that set.  Hence cannot add them in constants. Will revisit if we can import the buffer constants.

lib/msal-node/src/index.ts
sameerag: Removed now. Initially added as tsdx has trouble building files not included in index.ts and I did not have all of them built yet into the main export file.

sameerag: Removed now. Initially added as tsdx has trouble building files not included in index.ts and I did not have all of them built yet into the main export file.

lib/msal-node/src/utils/NodeConstants.ts
sameerag: Done

lib/msal-node/src/utils/NodeConstants.ts
sameerag: Done

lib/msal-node/test/TestConstants.ts
sameerag: Removed

sameerag: Done

lib/msal-node/src/crypto/PkceGenerator.ts
sameerag: Done

lib/msal-node/src/encode/Base64Decode.ts
jasonnutter: Can we at least have an interface in common?

jasonnutter:
Quick note: You can test the `index.ts` file by having the test files import the relevant modules from the index, instead of reaching into the package. This helps ensures the high-level public API isn't broken.

lib/msal-node/src/crypto/PkceGenerator.ts
sameerag: Done

lib/msal-node/src/encode/Base64Encode.ts
sameerag: Done. We have a class:`/utils/EncodeUtils.ts` which does the Base64 encoding and decoding.

lib/msal-node/src/crypto/CryptoOps.ts
pkanher617: Interface only defines function names, not class name. happy to rename, please provide some alternatives as CryptoUtils didn't sound appropriate.

lib/msal-node/src/crypto/PkceGenerator.ts
pkanher617: You can define these synchronously here if you want and return a Promise-wrapped value in the interface implementation above. making these static isn't affected by the interface.

lib/msal-node/src/index.ts
```diff
@@ -1,2 +1,5 @@
 export { PublicClientApplication } from './client/PublicClientApplication';
 export { ConfidentialClientApplication } from './client/ConfidentialClientApplication';
+
+// crypto
+export { CryptoOps } from './crypto/CryptoOps';
```
pkanher617: do we need to export these?

sameerag: This is placeholder as I cannot compile it otherwise, removed in the following PR since `PublicClientApplication` is fleshed out there.

lib/msal-node/src/crypto/PkceGenerator.ts
sameerag: `static` and `async` cannot be used together. This implementation looked cleaner than wrapping promise return later.

jasonnutter: > static and async cannot be used together

Are you sure? Pretty sure they can.

sameerag: Did not compile for me. I can check in a non vscode environment.

lib/msal-node/package.json
```diff
@@ -44,7 +45,22 @@
...
+    "moduleFileExtensions": [
+      "ts",
+      "tsx",
+      "js",
+      "json",
+      "jsx",
+      "node"
+    ],
+    "testMatch": [
+      "<rootDir>/test/**/?(*.)(spec|test).(ts|js)?(x)"
```
jasonnutter: Looks like you need to pull in the branch that has the fix for this.

sameerag: Yes. I will into the base branch once I fold all the sub branches.

lib/msal-node/src/crypto/PkceGenerator.ts
pkanher617: you can make them private static methods and call them the same way inside your public method.

sameerag: yeah! Well it is either that just not have static. Either way there is one unneeded qualifier on some functions. Not changing for now. Let us revisit, will make a note. Thanks for the feedback.

PR #1371, opened 2020-03-13
[msal-node] network module

sameerag:
This PR adds 'network' module for `msal-node`. 

**Libraries in use:**
- [npm axios](https://www.npmjs.com/package/axios) for making `http` calls in `Node.js`

**Tests:**
- All unit tests added, jest [mocking](https://jestjs.io/docs/en/mock-functions.html) used for mocking [axios](https://www.npmjs.com/package/axios)
- 100% coverage
- Please test with `npm run test` and `npm run test:c` for coverage

network-test-coverage


lib/msal-node/src/network/HttpClient.ts
```diff
@@ -0,0 +1,65 @@
...
+export class HttpClient implements INetworkModule {
+    /**
+     * Axios library for REST endpoints - Get request
+     * @param url
+     * @param headers
+     * @param body
+     */
+    async sendGetRequestAsync<T>(url: string, options?: NetworkRequestOptions): Promise<T> {
+        // setHeaders
+        this.setHeaders(options);
```
DarylThayil: what does this.setHeaders do

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -0,0 +1,65 @@
...
+
+        // axios config
+        const config = {
+            method: HttpMethod.GET,
+            url: url
+        };
+
+        // GET call
+        const response = await axios(config);
+        return await response as unknown as T;
```
DarylThayil: this line confuses me since we are awaiting response already.

should it be return response as T? 
and what is unknown for?

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -0,0 +1,65 @@
...
+            method: HttpMethod.POST,
+            url: url,
+            body: (options && options.body) || ''
+        };
+
+        const response = await axios(config);
+        return await response as unknown as T;
+    }
+
+    private setHeaders(options?: NetworkRequestOptions): void {
```
DarylThayil: I think that this is not what we want to do here, This is for setting defaults on requests, but each request is unique and I would worry we are not cleaning up headers. 

Axios config accepts headers as a key 

```js
const request = {
    method: HttpMethod.POST,
    url, 
    headers: options.headers
}

```

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -0,0 +1,65 @@
...
+     * @param url
+     * @param headers
+     * @param body
+     */
+    async sendPostRequestAsync<T>(url: string, options?: NetworkRequestOptions): Promise<T> {
+        // set Headers
+        this.setHeaders(options);
+
+        // axios config
+        const config = {
```
DarylThayil: rename to something like `request`

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -0,0 +1,65 @@
...
+     * @param url
+     * @param headers
+     * @param body
+     */
+    async sendGetRequestAsync<T>(url: string, options?: NetworkRequestOptions): Promise<T> {
+        // setHeaders
+        this.setHeaders(options);
+
+        // axios config
+        const config = {
```
DarylThayil: rename to something like `request`

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -0,0 +1,65 @@
...
+            method: HttpMethod.POST,
+            url: url,
+            body: (options && options.body) || ''
+        };
+
+        const response = await axios(config);
+        return await response as unknown as T;
```
DarylThayil: same questions here

lib/msal-node/test/network/HttpClient.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+    test('sendGetRequestAsync', async () => {
+
+        //Mocking axios function rather than a method
+        mocked(axios).mockResolvedValue(axiosResponse);
+
+        // Call
+        const result = await httpClient.sendGetRequestAsync('url');
+
+        //Assert
+        expect(result).toStrictEqual({
```
DarylThayil: this object is duplicated, can you define it one place and then expect it to match?

lib/msal-node/test/network/HttpClient.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+
+        //Mocking axios function rather than a method
+        mocked(axios).mockResolvedValue(axiosResponse);
+
+        // Call
+        const result = await httpClient.sendPostRequestAsync('url');
+
+        //Assert
+        expect(result).toStrictEqual({
+            data: {
```
DarylThayil: duplicate

lib/msal-node/test/network/HttpClient.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+        // Call
+        try {
+            await httpClient.sendPostRequestAsync('url', {
+                body: 'randomData',
+                headers: new Map<string, string>().set('key', 'value'),
+            });
+        } catch (e) {
+            //Assert
+            expect(e).toStrictEqual({
+                message: 'axios request failed',
```
DarylThayil: duplicate

DarylThayil:
few things that I think should be fixed but otherwise looks good

lib/msal-node/src/network/HttpClient.ts
sameerag: okay. I was being close to axios name space, but guess it makes sense. Can do.

lib/msal-node/src/network/HttpClient.ts
sameerag: okay here is the dilemma; msal-browser is doing `return await response.json() as T` which fails for msal-node compilation. I tried `return await response as T` : same issue. I want to investigate further to see what is ideal here, but currently this is the only way I could get it to work.

lib/msal-node/src/network/HttpClient.ts
sameerag: answered above

lib/msal-node/src/network/HttpClient.ts
sameerag: yes I am aware. But options.Headers is a `Map<string, string>` - defined by msal-common already. I agree defaults may not be a good idea, considering we change telemetry option, I think the best approach here is to change it while populating headers in msal-common, which I plan to in a follow up PR.

lib/msal-node/test/network/HttpClient.spec.ts
sameerag: Serialization of the response is causing mismatches. I should find more elegant way to do this. Will address it soon.

lib/msal-node/test/network/HttpClient.spec.ts
sameerag: Serialization of the response is causing mismatches. I should find more elegant way to do this. Will address it soon

lib/msal-node/test/network/HttpClient.spec.ts
sameerag: Serialization of the response is causing mismatches. I should find more elegant way to do this. Will address it soon

lib/msal-node/src/network/HttpClient.ts
sangonzal: Should be able to do return `response.data as T`.

lib/msal-node/src/network/HttpClient.ts
sangonzal: We should not be setting defaults at every request. What's the problem with options.Headers being a Map<string, string> ?

This is currently working for me in the device code implementation: 

```
 private async executePostRequest<T>(url: string, headers: Map<string, string>, body: string): Promise<T>{

        const response =  await axios({
            method: 'post',
            url: url,
            data: body,
            headers: headers,
          });

          return response.data as T;
```

sameerag: Awesome! Interesting,  this did not compile for me, may be my vscode is throwing me.  Will clean and do it again, I found issues with linking for me.  Will compile and check.

lib/msal-node/src/network/HttpClient.ts
sameerag: That is the next thought I had :)

sameerag: I want to execute this once for error responses and then update!

sameerag: Done. Please review.

lib/msal-node/src/network/HttpClient.ts
sameerag: Done.

lib/msal-node/src/network/HttpClient.ts
sameerag: Done

lib/msal-node/src/network/HttpClient.ts
sameerag: Done. Please check below.

lib/msal-node/test/network/HttpClient.spec.ts
sameerag: toMatchObject() works. Simplified the code now.

lib/msal-node/test/network/HttpClient.spec.ts
sameerag: Same as above

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -0,0 +1,53 @@
...
+import { INetworkModule, NetworkRequestOptions } from '@azure/msal-common';
+import { HttpMethod } from './../utils/Constants';
+import axios from 'axios';
+
+/**
+ * This class implements the Fetch API for GET and POST requests. See more here: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API
+ */
+export class HttpClient implements INetworkModule {
+    /**
+     * Axios library for REST endpoints - Get request
```
sangonzal: nit: Axios client*

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -0,0 +1,53 @@
...
+        // axios config
+        const request = {
+            method: HttpMethod.GET,
+            url: url,
+            headers: (options && options.headers)
+        };
+
+        // GET call
+        const response = await axios(request);
+        return await response.data as T;
```
sangonzal: Don't think you need to `await` here. Same thing below.

lib/msal-node/src/network/HttpClient.ts
sameerag: Done! It is my local environment, I was able to compile and test now.

lib/msal-node/test/network/HttpClient.spec.ts
sameerag: Done. `toMatchObj` works.

lib/msal-node/src/network/HttpClient.ts
sameerag: Done, sorry carry over from the template code in `msal-browser`

lib/msal-node/src/network/HttpClient.ts
sameerag: Done.

lib/msal-node/test/network/HttpClient.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+            headers: {},
+            config: {},
+        },
+    };
+
+    // test GET
+    test('sendGetRequestAsync', async () => {
+
+        //Mocking axios function rather than a method
+        mocked(axios).mockResolvedValue(axiosResponse);
```
jasonnutter: This should be done in `before` inside each test.

lib/msal-node/test/network/HttpClient.spec.ts
```diff
@@ -0,0 +1,74 @@
+import { HttpClient } from '../../src/network/HttpClient';
+import axios, { AxiosResponse } from 'axios';
+import { mocked } from "ts-jest/dist/util/testing"; //<-- This allows to mock results
```
jasonnutter: Possible to import from the main export instead of reaching into the package?

`import { mocked } from "test-jes"`

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -0,0 +1,53 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { INetworkModule, NetworkRequestOptions } from '@azure/msal-common';
+import { HttpMethod } from './../utils/Constants';
+import axios from 'axios';
+
+/**
+ * This class implements the Fetch API for GET and POST requests. See more here: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API
```
jasonnutter: This comment is incorrect, as fetch is not used in this. I would say something more along the lines of "implements methods for making network requests"

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -0,0 +1,53 @@
...
+     * @param url
+     * @param headers
+     * @param body
+     */
+    async sendGetRequestAsync<T>(url: string, options?: NetworkRequestOptions): Promise<T> {
+
+        // axios config
+        const request = {
+            method: HttpMethod.GET,
+            url: url,
```
jasonnutter: Nit: can just be `url`, instead of `url: url,`

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -0,0 +1,53 @@
...
+import { INetworkModule, NetworkRequestOptions } from '@azure/msal-common';
+import { HttpMethod } from './../utils/Constants';
+import axios from 'axios';
+
+/**
+ * This class implements the Fetch API for GET and POST requests. See more here: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API
+ */
+export class HttpClient implements INetworkModule {
+    /**
+     * Axios CLient library for REST endpoints - Get request
```
jasonnutter: Axios is a implementation detail, I wouldn't mention it here.

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Axios Client for REST endpoints - Post request
+     * @param url
+     * @param headers
+     * @param body
+     */
+    async sendPostRequestAsync<T>(url: string, options?: NetworkRequestOptions): Promise<T> {
+
+        // axios config
```
jasonnutter: Superfluous comment.

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -0,0 +1,53 @@
...
+    async sendGetRequestAsync<T>(url: string, options?: NetworkRequestOptions): Promise<T> {
+
+        // axios config
+        const request = {
+            method: HttpMethod.GET,
+            url: url,
+            headers: (options && options.headers)
+        };
+
+        // GET call
```
jasonnutter: Also superfluous

lib/msal-node/src/network/HttpClient.ts
sameerag: Done

lib/msal-node/src/network/HttpClient.ts
sameerag: Thanks. Sorry missed this :). Corrected now.

lib/msal-node/src/network/HttpClient.ts
sameerag: Done

lib/msal-node/src/network/HttpClient.ts
sameerag: Removed

lib/msal-node/src/network/HttpClient.ts
sameerag: removed.

lib/msal-node/test/network/HttpClient.spec.ts
sameerag: beforeEach() and mock do not seem to work together, it fails. I did some digging and have seen only examples resetting the mock in before it, not add the mock. I will get back to this when I have more time. Keeping it as is for now.

lib/msal-node/test/network/HttpClient.spec.ts
sameerag: Done!

lib/msal-node/test/network/HttpClient.spec.ts
jasonnutter: Ah interesting, good to know.

jasonnutter:
Guessing this also needs test fix merged in.

PR #1373, opened 2020-03-13
Move msal-core references docs to /ref folder

jasonnutter:
Since we want to have markdown files in the `docs` folder, we need to move the reference docs to a different folder. In our Friday session when we made this change, we agreed on the `ref` folder. The new path for the reference docs is now:

https://azuread.github.io/microsoft-authentication-library-for-js/ref/msal-core/
https://azuread.github.io/microsoft-authentication-library-for-js/ref/msal-angular/


sameerag (approved):
👍

pkanher617 (approved):

PR #1379, opened 2020-03-16
[DO NOT REVIEW][msal-common][msal-node] End-to-end Authorization Code Flow

sameerag:
NOTE: Please ignore this in favor of #1409. 

This PR supports end-to-end auth-code flow for `msal-node`
- Public API definitions
- Request/Error are complete 
- Generic framing added to support other flows

## Pending

- [ ]  RespondHandling to be added soon - needed a little clarification with `msal-browser` 
- [ ]  Add numeric_codes to `Error` objects
- [ ]  A web sample demonstrating usage of this API is checked in but would need improvement
- [ ]  Unit tests are in progress


lib/msal-browser/src/config/Configuration.ts
```diff
@@ -2,14 +2,14 @@
...
-import { AuthOptions, SystemOptions, LoggerOptions, INetworkModule, LogLevel } from "@azure/msal-common";
+import { SPAAuthOptions, SystemOptions, LoggerOptions, INetworkModule, LogLevel } from "@azure/msal-common";
 import { BrowserUtils } from "../utils/BrowserUtils";
 import { BrowserConstants } from "../utils/BrowserConstants";
 
 // Default timeout for popup windows in milliseconds
 const DEFAULT_POPUP_TIMEOUT_MS = 60000;
 
-export type BrowserAuthOptions = AuthOptions & {
+export type BrowserAuthOptions = SPAAuthOptions & {
```
DarylThayil: What is the difference between SPA Auth and Browser Auth?

pkanher617: SPA Auth is from common, browser auth is from msal-browser. @sameerag does this need to be separate from node auth config?

sameerag: For now yes, it isn't generic enough to reuse for node. I will work on cleaning and integrating SPA code in msal-common once we have node working end to end

lib/msal-common/src/request/ScopeSet.ts
```diff
@@ -199,4 +254,15 @@ export class ScopeSet {
         }
         return "";
     }
+
+    /**
+     * Prints scopes into a space-delimited string
+     * @param scopes
+     */
+    static finalScopes(scopes: Array<string>): string {
```
pkanher617: printScopes() already does this, not sure why we need another function.

lib/msal-common/src/request/ScopeSet.ts
```diff
@@ -167,11 +223,10 @@ export class ScopeSet {
...
     isLoginScopeSet(): boolean {
-        const hasLoginScopes = (
-            this.originalScopes.has(this.clientId) ||
-            this.originalScopes.has(Constants.OPENID_SCOPE) ||
-            this.originalScopes.has(Constants.PROFILE_SCOPE)
-        );
+        const hasLoginScopes =
+                   this.originalScopes.has(this.clientId) ||
+                   this.originalScopes.has(Constants.OPENID_SCOPE) ||
+                   this.originalScopes.has(Constants.PROFILE_SCOPE);
```
pkanher617: nitpick but removing parentheses lessens readability IMO

lib/msal-common/src/request/ScopeSet.ts
```diff
@@ -125,35 +169,47 @@ export class ScopeSet {
...
-            throw ClientAuthError.createEmptyInputScopeSetError(otherScopes);
+            throw ClientAuthError.createEmptyInputScopeSetError(
+                otherScopes
+            );
         }
-        return this.unionScopeSets(otherScopes).size < (this.scopes.size + otherScopes.getScopeCount());
+        return (
+            this.unionScopeSets(otherScopes).size <
+                   this.scopes.size + otherScopes.getScopeCount()
+        );
```
pkanher617: Revert this change, we don't want OOO messing up here. I know '+' is higher order than '<' but this change doesn't help us ensure that.

lib/msal-common/src/request/ScopeSet.ts
```diff
@@ -21,29 +21,49 @@ export class ScopeSet {
     // Boolean denoting whether scopes are required. Usually used for validation.
     private scopesRequired: boolean;
 
-    constructor(inputScopes: Array<string>, appClientId: string, scopesRequired: boolean) {
+    constructor(
+        inputScopes: Array<string>,
+        appClientId: string,
+        scopesRequired: boolean,
+        nonSPA?: boolean
```
pkanher617: This is not reflective of how it is used for the msal browser use case, I'm not sure I like the name change here.

lib/msal-common/src/request/ScopeSet.ts
```diff
@@ -21,29 +21,49 @@ export class ScopeSet {
...
+
         // Validate and filter scopes (validate function throws if validation fails)
         this.validateInputScopes(filteredInput);
-        const scopeArr = filteredInput ? StringUtils.trimAndConvertArrayEntriesToLowerCase([...filteredInput]) : [];
+
+        const scopeArr = filteredInput
+            ? StringUtils.trimAndConvertArrayEntriesToLowerCase([
+                ...filteredInput
+            ])
+            : [];
```
pkanher617: This is arguably less readable when split up over multiple lines.

lib/msal-common/src/request/ScopeSet.ts
```diff
@@ -21,29 +21,49 @@ export class ScopeSet {
...
         this.scopes = new Set<string>(scopeArr);
         if (!this.scopesRequired) {
             this.appendScope(this.clientId);
         }
         this.originalScopes = new Set<string>(this.scopes);
-        this.replaceDefaultScopes();
+
+        // add default scopes
+        nonSPA ? this.replaceDefaultScopes() : this.addDefaultScopes();
```
pkanher617: Why do we have two separate APIs essentially doing the same thing?

lib/msal-common/src/request/ScopeSet.ts
```diff
@@ -61,62 +81,86 @@ export class ScopeSet {
...
     /**
      * Check if a given scope is present in this set of scopes.
-     * @param scope 
+     * @param scope
      */
     containsScope(scope: string): boolean {
-        return !StringUtils.isEmpty(scope) ? this.scopes.has(scope) : false;
+        return !StringUtils.isEmpty(scope)
+            ? this.scopes.has(scope)
+            : false;
```
pkanher617: same argument about readability

lib/msal-common/src/request/AuthorizationCodeParameters.ts
```diff
@@ -0,0 +1,93 @@
...
+    codeVerifier?: string;
+    clientSecret?: string;
+    correlationId?: string;
+
+    /**
+     * Generates a map for all the params to be sent to the service
+     * @param request
+     * @param config
+     */
+    static generateAuthCodeParams(
```
pkanher617: not sure I understand the point of having a class with public members and a static function. You could define the public variables in a type or you can have the generateAuthCodeParams be a direct API on the object.

lib/msal-common/src/request/AuthorizationCodeUrlParameters.ts
```diff
@@ -0,0 +1,146 @@
...
+    claimsRequest?: string;
+    nonce?: string;
+    correlationId?: string;
+
+    /**
+     * This API validates the `AuthorizationCodeUrlParameters` and creates a URL to be sent to the server
+     * @param request
+     * @param config
+     */
+    static generateAuthCodeUrlParams(
```
pkanher617: same comment about static functions and types.

lib/msal-common/src/request/RequestValidator.ts
```diff
@@ -0,0 +1,84 @@
...
+
+import { ScopeSet } from "./ScopeSet";
+import { StringUtils } from "./../utils/StringUtils";
+import { ClientConfigurationError } from "./../error/ClientConfigurationError";
+import { PromptValue, CodeChallengeMethodValues } from "./../utils/Constants";
+
+/**
+ * Validates server consumable params from the "request" objects
+ */
+export class RequestValidator {
```
pkanher617: If we have classes for the request objects we should put the validation in those classes, the only reason to have a separate class for validation is if we are using types.

lib/msal-common/src/server/ServerParamsGenerator.ts
```diff
@@ -0,0 +1,257 @@
...
+ * Licensed under the MIT License.
+ */
+
+import { AADServerParamKeys, SSOTypes } from "../utils/Constants";
+import { Constants, HeaderNames } from "../utils/Constants";
+import { ScopeSet } from "../request/ScopeSet";
+import { Authority } from "../authority/Authority";
+import { ClientConfigurationError } from "../error/ClientConfigurationError";
+
+export class ServerParamsGenerator {
```
pkanher617: Not sure I understand the point of having a class with static members here. This is difficult to maintain and would rather have a single function that does all of this as we had before.

lib/msal-node/test/client/PublicClientApplication.spec.ts
```diff
@@ -1,8 +1,33 @@
...
+        const request: AuthorizationCodeUrlParameters = {
+            scopes: ['user.read'],
+            authority:
+                'https://login.microsoftonline.com/72f988bf-86f1-41af-91ab-2d7cd011db47',
+            redirectUri: 'http://localhost:3000',
+            codeChallenge: '31943XseS-Ae9f0qGR0p1fruKNjZIcyVu6d-lGP3xf0',
+            codeChallengeMethod: 'sha256',
+            state: '2400cdfb-6a0b-4200-a368-8f2bf6d60500',
+            correlationId: '43b735d3-07b5-400f-b0e0-eaeb7cae1bfe',
+        };
```
pkanher617: You can set up a string constants file so you can reuse values in different tests.

lib/msal-common/src/request/AuthorizationCodeParameters.ts
sameerag: In this case yeah, can make it a directAPI on the object. Will make the changes. I made validator and adding param functions static as they can be called independently by various objects.

lib/msal-common/src/request/AuthorizationCodeUrlParameters.ts
sameerag: Agreed.

lib/msal-common/src/request/RequestValidator.ts
sameerag: nope. The intent is to have all/any request objects be able to call each validator as it applies. Adds flexibility for various requests that will come as we keep adding support for different flows.

lib/msal-common/src/request/ScopeSet.ts
sameerag: This class has to be rewritten to fit a more generic rhetoric. I added this as a temporary mitigation. Can create a new PR to close this in a cleaner fashion.

lib/msal-common/src/request/ScopeSet.ts
sameerag: auto correct by "prettify". We can change the linter rules as we see it before final merge.

lib/msal-common/src/request/ScopeSet.ts
sameerag: auto correct by "prettify". We can change the linter rules as we see it before final merge.

lib/msal-common/src/request/ScopeSet.ts
sameerag: auto correct by "prettify". We can change the linter rules as we see it before final merge.

lib/msal-common/src/request/ScopeSet.ts
sameerag: I did not change this, original code had it. But can remove the parantheses.

lib/msal-common/src/request/ScopeSet.ts
sameerag: There was some issue using it initially. I would like to build more generic interface for the class ScopeSet.ts to fit node, whatever is added is temporary mitigation.

lib/msal-common/src/server/ServerParamsGenerator.ts
sameerag: This is like utils, and need all functions separate as device code and auth code may have some common, some different params. We want to be able to call them independently. Do not want them in a single function for flexibility.

lib/msal-node/test/client/PublicClientApplication.spec.ts
sameerag: Yes. This is more of a dummy test. Will add.

lib/msal-common/src/request/AuthorizationCodeUrlParameters.ts
sameerag: done

lib/msal-common/src/request/AuthorizationCodeParameters.ts
sameerag: done

lib/msal-common/src/request/ScopeSet.ts
sameerag: so for SPA you remove the clientId and replace the scopes, for other flows, clientId as a scope is  not supported.  Raised #1390 to address this.

lib/msal-common/src/request/ScopeSet.ts
sameerag: Changed and committed

lib/msal-common/src/request/ScopeSet.ts
sameerag: #1390 to address this

lib/msal-common/src/request/ScopeSet.ts
sameerag: Edited.

lib/msal-common/src/request/ScopeSet.ts
sameerag: Edited

sameerag:
Thanks for the feedback @pkanher617. This PR is just the base approach for #1409 which @sangonzal is working on. I agree with some of your comments and we tried to partisan the request from the generateParams in that PR. Please take a look when you can.

PR #1383, opened 2020-03-16
[msal-node] lerna bootstrap and lerna builds

sameerag:
This PR adds support  for `lerna bootstrap` and `lerna run build`. This helps to build symlinks in the mono-repo between `@azure/msal-common` and `@azure/msal-node`.


lib/msal-node/package.json
```diff
@@ -65,12 +66,13 @@
...
     "@types/jest": "^25.1.2",
+    "@types/uuid": "^7.0.0",
     "husky": "^4.2.3",
     "tsdx": "^0.12.3",
     "tslib": "^1.10.0",
     "typescript": "^3.7.5"
   },
   "dependencies": {
-    "@types/uuid": "^7.0.0"
+    "@azure/msal-common": "1.0.0-alpha.0"
```
jasonnutter: `axios` needs to also be listed here.

sameerag: Yep, just added and committed.

sameerag: tests are failing now. Will fix and update.

lib/msal-node/package.json
```diff
@@ -65,12 +66,14 @@
   "module": "dist/msal-node.esm.js",
   "devDependencies": {
     "@types/jest": "^25.1.2",
+    "@types/uuid": "^7.0.0",
+    "axios": "^0.19.2",
```
jasonnutter: Needs to be a regular dependency.

sameerag: Done.

PR #1388, opened 2020-03-17
wire up telemetry for acuqire token silent

DarylThayil:
Add instrumentation for Acquire token silent.
This will be to help debug perf, but future instrumentation will be added


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -40,6 +41,7 @@ import { Constants,
     PersistentCacheKeys,
     ErrorCacheKeys,
 } from "./utils/Constants";
+import uuid from "uuid";
```
pkanher617: Any reason we don't use the in house guid creator?

jasonnutter: Use `CryptoUtils` helper function to create a guid instead.

lib/msal-core/test/UserAgentApplication.spec.ts
```diff
@@ -187,15 +187,15 @@ describe("UserAgentApplication.ts Class", function () {
...
-        });
+        // it("telemetry manager doesn't exis in UAA when not configured", () => {
+        //     msal = new UserAgentApplication({
+        //         auth: {
+        //             clientId: TEST_CONFIG.MSAL_CLIENT_ID
+        //         }
+        //     });
+        //     // @ts-ignore
+        //     expect(msal.telemetryManager).to.be.null;
+        // });
```
jasonnutter: Remove instead.

samples/VanillaJSTestApp/JavascriptSPA/authConfig.js
```diff
@@ -2,12 +2,22 @@
...
     cache: {
         cacheLocation: "localStorage", // This configures where your cache will be stored
         storeAuthStateInCookie: false, // Set this to "true" if you are having issues on IE11 or Edge
+    },
+    system: {
+        telemetry: {
+            applicationName: 'msalVanillaTestApp',
+            applicationVersion: 'test1.0',
+            telemetryEmitter: (events) => {
+                console.log(events);
```
jasonnutter: I would doing something like `console.log("Telemetry Event: ", event)` to differentiate from other logs.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -589,7 +591,11 @@ export class UserAgentApplication {
      *
      */
     acquireTokenSilent(userRequest: AuthenticationParameters): Promise<AuthResponse> {
-
+        const requestCorrelationId = userRequest.correlationId || uuid.v4();
+        const apiEvent: ApiEvent = new ApiEvent(requestCorrelationId, this.logger);
+        apiEvent.apiEventIdentifier = API_EVENT_IDENTIFIER.AcquireTokenSilent;
+        apiEvent.apiCode = API_CODE.AcquireTokenSilent;
```
pkanher617: Is it possible to flatten this to a single constructor call? Having public members set without getters or setters is a bad code smell IMO.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -710,10 +716,21 @@ export class UserAgentApplication {
...
-        });
+        })
+            .then(res => {
+                apiEvent.wasSuccessful = true;
+                return res;
+            })
+            .catch((error: AuthError) => {
+                this.cacheStorage.resetTempCacheItems(request.state);
+                apiEvent.apiErrorCode = error.errorCode;
+                apiEvent.wasSuccessful = false;
```
pkanher617: make this a function on apiEvent instead

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -710,10 +716,21 @@ export class UserAgentApplication {
                         return null;
                     });
             }
-        }).catch((error: AuthError) => {
-            this.cacheStorage.resetTempCacheItems(request.state);
-            throw error;
-        });
+        })
+            .then(res => {
+                apiEvent.wasSuccessful = true;
```
pkanher617: make this a function on apiEvent instead

jasonnutter (changes requested):

lib/msal-core/src/UserAgentApplication.ts
DarylThayil: I have thought about that, I am happy to revisit the design when we instrument more, but probably not in this pr

samples/VanillaJSTestApp/JavascriptSPA/authConfig.js
DarylThayil: addiong

lib/msal-core/src/UserAgentApplication.ts
DarylThayil: great call, fixing

DarylThayil: updated

lib/msal-core/src/UserAgentApplication.ts
DarylThayil: same comment on changing design in another pr

lib/msal-core/src/UserAgentApplication.ts
DarylThayil: same comment on changing design in another pr

lib/msal-core/test/UserAgentApplication.spec.ts
DarylThayil: Updated tests to replace these

pkanher617 (approved):
Might be helpful to see some docs for this as well. Approving for now.

lib/msal-core/src/UserAgentApplication.ts
sameerag: On a side note, we also worked very hard to simplify this API :-), so it helps instrumenting the code in independent helpers.

jasonnutter (approved):

PR #1394, opened 2020-03-18
add connection speed to telemetry

DarylThayil:
add connection speed to telemetry

### DIFF
--- a/lib/msal-core/src/telemetry/TelemetryManager.ts
+++ b/lib/msal-core/src/telemetry/TelemetryManager.ts
@@ -9,7 +9,7 @@ import {
     TelemetryEmitter
 } from "./TelemetryTypes";
 import DefaultEvent from "./DefaultEvent";
-import { libraryVersion } from "../utils/Constants";
+import { libraryVersion, Constants } from "../utils/Constants";
 
 // for use in cache events
 const MSAL_CACHE_EVENT_VALUE_PREFIX = "msal.token";
@@ -36,7 +36,15 @@ export default class TelemetryManager {
 
     constructor(config: TelemetryConfig, telemetryEmitter: TelemetryEmitter) {
         // TODO THROW if bad options
-        this.telemetryPlatform = config.platform;
+        this.telemetryPlatform = {
+            sdk: Constants.libraryName,
+            sdkVersion: libraryVersion(),
+            networkInformation: {
+                // @ts-ignore
+                connectionSpeed: navigator && navigator.connection && navigator.connection.effectiveType
+            },
+            ...config.platform
+        };
         this.clientId = config.clientId;
         this.onlySendFailureTelemetry = config.onlySendFailureTelemetry;
         /*
@@ -52,8 +60,6 @@ export default class TelemetryManager {
         const applicationVersion = "0.0";
         const telemetryEmitter = () => {};
         const telemetryPlatform: TelemetryPlatform = {
-            sdk: "msal.js",
-            sdkVersion: libraryVersion(),
             applicationName,
             applicationVersion
         };
--- a/lib/msal-core/src/telemetry/TelemetryTypes.ts
+++ b/lib/msal-core/src/telemetry/TelemetryTypes.ts
@@ -9,14 +9,20 @@ export type TelemetryEmitter = (events: Array<object>) => void;
 
 // SDK SHOULD BE DEFAULTED and Pulled from Package
 export interface TelemetryPlatform {
-    sdk: string;
-    sdkVersion: string;
+    sdk?: string;
+    sdkVersion?: string;
     applicationName: string;
     applicationVersion: string;
+    networkInformation?: NetworkInformation
 }
 
 export interface TelemetryConfig {
     platform: TelemetryPlatform;
     onlySendFailureTelemetry?: boolean;
     clientId: string;
+}
+
+// In the browser this commes from navigator.connection
+export interface NetworkInformation {
+    connectionSpeed: string;
 }
--- a/lib/msal-core/test/telemetry/DefaultEvent.spec.ts
+++ b/lib/msal-core/test/telemetry/DefaultEvent.spec.ts
@@ -15,7 +15,10 @@ describe("DefaultEvent", () => {
             sdk: "javascript",
             sdkVersion: "1.0.0",
             applicationName: "mochaTest",
-            applicationVersion: "1.22"
+            applicationVersion: "1.22",
+            networkInformation: {
+                connectionSpeed: "4g"
+            }
         }
         const defaultEvent: DefaultEvent = new DefaultEvent(
             platformConfig,
@@ -31,5 +34,6 @@ describe("DefaultEvent", () => {
         expect(event["msal.event_name"]).to.eq("msal.default_event");
         expect(event["msal.application_name"]).to.eq("mochaTest");
         expect(event["msal.elapsed_time"]).to.eq(-1);
+        expect(event["msal.effective_connection_speed"]).to.eq("4g");
     });
 });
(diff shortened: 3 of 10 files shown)


samples/VanillaJSTestApp/JavascriptSPA/authConfig.js
```diff
@@ -1,7 +1,7 @@
 // Config object to be passed to Msal on creation
 const msalConfig = {
     auth: {
-        clientId: "4b0db8c2-9f26-4417-8bde-3f0e3656f8e0",
+        clientId: "6226576d-37e9-49eb-b201-ec1eeb0029b6",
```
jasonnutter: Why did this change?

lib/msal-core/src/telemetry/TelemetryManager.ts
```diff
@@ -55,7 +55,11 @@ export default class TelemetryManager {
             sdk: "msal.js",
             sdkVersion: libraryVersion(),
             applicationName,
-            applicationVersion
+            applicationVersion,
+            networkInformation: {
+                // @ts-ignore
+                effectiveType: navigator && navigator.connection && navigator.connection.effectiveType
```
jasonnutter: I would maybe call our property something else, like `connectionSpeed`.

jasonnutter: Also, I would consider adding an enum, as there is a known set of expected values:  'slow-2g', '2g', '3g', or '4g'

lib/msal-core/src/telemetry/TelemetryTypes.ts
```diff
@@ -13,10 +13,16 @@ export interface TelemetryPlatform {
...
 
 export interface TelemetryConfig {
     platform: TelemetryPlatform;
     onlySendFailureTelemetry?: boolean;
     clientId: string;
 }
+
+// In the browser this commes from navigator.connection
+export interface NetworkInformation {
+    effectiveType: string;
```
jasonnutter: Can this be an enum, since there is a known set of expected value?

samples/VanillaJSTestApp/JavascriptSPA/authConfig.js
DarylThayil: The other client id is not configured for non Microsoft accounts , I needed to test against msa as well

lib/msal-core/src/telemetry/TelemetryManager.ts
DarylThayil: good stuggestion

lib/msal-core/src/telemetry/TelemetryTypes.ts
DarylThayil: I would prefer not, since we don't do anything with the value programatically. we just want to dump whatever the browser holds here

lib/msal-core/src/telemetry/TelemetryManager.ts
```diff
@@ -55,7 +55,11 @@ export default class TelemetryManager {
             sdk: "msal.js",
             sdkVersion: libraryVersion(),
             applicationName,
-            applicationVersion
+            applicationVersion,
+            networkInformation: {
+                // @ts-ignore
+                connectionSpeed: navigator && navigator.connection && navigator.connection.effectiveType
```
jasonnutter: I see this is in two places...could it just be in one place, e.g. the constructor?

DarylThayil: Yeah the doubling there is unfortunate. TelemetryManager has a static 'get me a dummy instance' which would be used internally when devs don't set their own. 
I guess you could argue that it is a static value that should just be set by the Telemetry Manager constructor.... 
Infact I think I am convinced.
Will update

DarylThayil: ahhh wait, now I know why. its because this Telemetry platform object is passed to default event and so should the Telemetry Manager take an object and set another value on it ... I am not sure  that is ideal

DarylThayil: There is also wanting Telemetry manager to be something that can be in common ... maybe that is too soon?
Thoughts?

DarylThayil: resolved

samples/VanillaJSTestApp/JavascriptSPA/authConfig.js
jasonnutter: @pkanher617 I think changed this for testing, interested what he thinks.

lib/msal-core/src/telemetry/TelemetryManager.ts
```diff
@@ -36,7 +36,15 @@ export default class TelemetryManager {
 
     constructor(config: TelemetryConfig, telemetryEmitter: TelemetryEmitter) {
         // TODO THROW if bad options
-        this.telemetryPlatform = config.platform;
+        this.telemetryPlatform = {
+            sdk: "Msal.js",
```
sameerag: Can we please use a constant instead ( esp to port this for other libraries as needed)?

lib/msal-core/src/telemetry/TelemetryTypes.ts
```diff
@@ -9,14 +9,20 @@ export type TelemetryEmitter = (events: Array<object>) => void;
 
 // SDK SHOULD BE DEFAULTED and Pulled from Package
 export interface TelemetryPlatform {
-    sdk: string;
-    sdkVersion: string;
+    sdk?: string;
+    sdkVersion?: string;
     applicationName: string;
     applicationVersion: string;
+    networkInformation?: NetworkInformation
```
sameerag: Does JS work with mandated params in between and optionals spread out? In function signatures they need to come after required parameters, I am not sure about object signatures.

lib/msal-core/test/telemetry/DefaultEvent.spec.ts
```diff
@@ -15,7 +15,10 @@ describe("DefaultEvent", () => {
             sdk: "javascript",
             sdkVersion: "1.0.0",
             applicationName: "mochaTest",
-            applicationVersion: "1.22"
+            applicationVersion: "1.22",
+            networkInformation: {
+                connectionSpeed: "4g"
```
sameerag: Can "4g" be an enum?

lib/msal-core/src/telemetry/TelemetryTypes.ts
DarylThayil: not clear to me what is being asked here :)

lib/msal-core/test/telemetry/DefaultEvent.spec.ts
DarylThayil: Since this is a string coming from the browser and we do nothing programatically with it other than dump it in telemetry I opted not to.

lib/msal-core/src/telemetry/TelemetryManager.ts
DarylThayil: Good call

DarylThayil: updated

lib/msal-core/src/telemetry/TelemetryTypes.ts
tnorling: I think @sameerag makes a good point. Usually optional parameters come after the required ones. Not sure if its strictly required for interfaces like it is for functions but it's probably good form to do so.

jasonnutter: Yeah that matters for function signatures, but doesn't matter here, as these values are key-based, not positioned based.

PR #1395, opened 2020-03-19
Turn library state into encoded string that contains guid and timestamp

### LINKED ISSUE #1259: localStorage `expiresIn` out of sync with JWT `exp`
## Library
- [X] `msal@1.2.1`

## Framework
No framework.

## Description
We’re seeing instances where MSAL's `expiresIn` value stored in `localStorage` differs quite a bit from the `exp` value in the JWT token (sometimes in the minutes, sometimes in the hours/days range).

If a user gets into a state where `exp` is in the past, and `expiresIn` is in the future, calling `acquireTokenSilent` will consistently give you an expired token (cause it checks the value in `expiresIn`, and not the value in `exp`), so there's no way for a user to get out of that state without manually deleting `localStorage`.

The root cause is that `exp` value is calculated on the server, whereas `expiresIn` is calculated after-the-fact on the client, using their system clock.

We've seen the biggest difference between the two values when users put our tab in the background, or their computers to sleep. Chrome seems to throttle/pause code execution and/or network requests, and if that happens between the time you get a token back from the server, and the time when `expiresIn` value is calculated, you can get huge drift. We had a user whose JWT `exp` value was a Friday, and his `localStorage` `expiresIn` value was a Monday.

This is where MSAL.js relies on the system clock: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/1212c640aa619718328a5f6931f14ecf27ea0b36/lib/msal-core/src/UserAgentApplication.ts#L1381

## Security
No.

## Regression
No. This problem is even present in ADAL.js. Here's where ADAL.js relies on the system clock: https://github.com/AzureAD/azure-activedirectory-library-for-js/blob/de375da0f438018f01c6902f6dcfaedb24f91320/lib/adal.js#L1661

## Configuration

```js
{
  auth: {
    clientId: "e25ff63c-2374-438a-9fbe-91cd5ee67bae",
    authority: "https://login.microsoftonline.com/organizations",
    redirectUri: "http://localhost:8080/"
  },
  cache: {
    storeAuthStateInCookie: true
  }
}
```

## Reproduction steps
1. Tweak the quickstart example to acquire tokens on a `setInterval`
2. Clear `localStorage` periodically, so that you force the server to send you a new token
3. Mess around with network download latencies, and see the drift
4. Mess around with the system clock, and see the drift
5. Put your computer to sleep at just the right time (kinda hard), and see the huge drift

## Expected behavior
No drift, or a way to get out of this scenario.

- No drift: This would require dropping `expiresIn`, and just relying on JWT's `exp` value. I hear you're working on encrypting tokens, so this may not be possible.
- A way to get out of this scenario: If MSAL thinks a token should be valid, but our code is getting 401s consistently from our APIs, we should have a way of skipping MSAL's cache, and having it give us a fresh token.

## Browsers
We see it happen relatively frequently with Chrome, but I see no reason why it wouldn't affect other browsers.

jasonnutter:
Proof-of-concept for fixing #1259. Per our design discussion, this would turn the value we pass as state into a base64 encoded object containing a guid and original timestamp of the request (and eventually it will contain the request method to address `loginRedirect` bugs). 

Based on this PR to ADAL: https://github.com/AzureAD/azure-activedirectory-library-for-js/pull/898/files

TODO:
- [x] Confirm backwards compat is needed.
- [x] Finalize keys
- [x] Tests
- [x] Error handling
- [x] Test with tab suspending

### DIFF
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -1023,7 +1023,7 @@ export class UserAgentApplication {
 
     /**
      * @hidden
-     * This method must be called for processing the response received from the STS if using popups or iframes. It extracts the hash, processes the token or error 
+     * This method must be called for processing the response received from the STS if using popups or iframes. It extracts the hash, processes the token or error
      * information and saves it in the cache. It then resolves the promises with the result.
      * @param {string} [hash=window.location.hash] - Hash fragment of Url.
      */
@@ -1043,13 +1043,13 @@ export class UserAgentApplication {
 
     /**
      * @hidden
-     * This method must be called for processing the response received from the STS when using redirect flows. It extracts the hash, processes the token or error 
+     * This method must be called for processing the response received from the STS when using redirect flows. It extracts the hash, processes the token or error
      * information and saves it in the cache. The result can then be accessed by user registered callbacks.
      * @param {string} [hash=window.location.hash] - Hash fragment of Url.
      */
     private handleRedirectAuthenticationResponse(hash: string): void {
         this.logger.info("Returned from redirect url");
-        
+
         // clear hash from window
         window.location.hash = "";
 
@@ -1088,7 +1088,7 @@ export class UserAgentApplication {
         if (!parameters) {
             throw AuthError.createUnexpectedError("Hash was not parsed correctly.");
         }
-        if (parameters.hasOwnProperty("state")) {
+        if (parameters.hasOwnProperty(ServerHashParamKeys.STATE)) {
             const parsedState = RequestUtils.parseLibraryState(parameters.state);
 
             stateResponse = {
@@ -1106,13 +1106,13 @@ export class UserAgentApplication {
          */
 
         // loginRedirect
-        if (stateResponse.state === this.cacheStorage.getItem(`${TemporaryCacheKeys.STATE_LOGIN}${Constants.resourceDelimiter}${stateResponse.state}`, this.inCookie) || stateResponse.state === this.silentAuthenticationState) { // loginRedirect
+        if (stateResponse.state === this.cacheStorage.getItem(`${TemporaryCacheKeys.STATE_LOGIN}${Constants.resourceDelimiter}${stateResponse.state}`, this.inCookie) || stateResponse.state === this.silentAuthenticationState) {
             stateResponse.requestType = Constants.login;
             stateResponse.stateMatch = true;
             return stateResponse;
         }
         // acquireTokenRedirect
-        else if (stateResponse.state === this.cacheStorage.getItem(`${TemporaryCacheKeys.STATE_ACQ_TOKEN}${Constants.resourceDelimiter}${stateResponse.state}`, this.inCookie)) { // acquireTokenRedirect
+        else if (stateResponse.state === this.cacheStorage.getItem(`${TemporaryCacheKeys.STATE_ACQ_TOKEN}${Constants.resourceDelimiter}${stateResponse.state}`, this.inCookie)) {
             stateResponse.requestType = Constants.renewToken;
             stateResponse.stateMatch = true;
             return stateResponse;
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -151,7 +151,8 @@ export class UserAgentApplication {
     // state variables
     private silentAuthenticationState: string;
     private silentLogin: boolean;
-    private redirectCallbacksSet: boolean;
+    private redirectResponse: AuthResponse;
+    private redirectError: AuthError;
 
     // Authority Functionality
     protected authorityInstance: Authority;
@@ -209,9 +210,6 @@ export class UserAgentApplication {
         // Set the Configuration
         this.config = buildConfiguration(configuration);
 
-        // Set the callback boolean
-        this.redirectCallbacksSet = false;
-
         this.logger = this.config.system.logger;
         this.clientId = this.config.auth.clientId;
         this.inCookie = this.config.cache.storeAuthStateInCookie;
@@ -239,7 +237,7 @@ export class UserAgentApplication {
 
         // On the server 302 - Redirect, handle this
         if (urlContainsHash && !WindowUtils.isInIframe() && !WindowUtils.isInPopup()) {
-            this.handleAuthenticationResponse(urlHash);
+            this.handleRedirectAuthenticationResponse(urlHash);
         }
     }
 
@@ -256,7 +254,6 @@ export class UserAgentApplication {
     handleRedirectCallback(authCallback: authResponseCallback): void;
     handleRedirectCallback(authOrTokenCallback: authResponseCallback | tokenReceivedCallback, errorReceivedCallback?: errorReceivedCallback): void {
         if (!authOrTokenCallback) {
-            this.redirectCallbacksSet = false;
             throw ClientConfigurationError.createInvalidCallbackObjectError(authOrTokenCallback);
         }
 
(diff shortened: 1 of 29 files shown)


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1103,9 +1104,12 @@ export class UserAgentApplication {
             throw AuthError.createUnexpectedError("Hash was not parsed correctly.");
         }
         if (parameters.hasOwnProperty("state")) {
+            const parsedState = RequestUtils.parseLibraryState(parameters.state);
+
             stateResponse = {
                 requestType: Constants.unknown,
-                state: parameters.state,
+                state: parsedState.state,
+                iat: parsedState.iat,
```
DarylThayil: what is iat?

jasonnutter: Issued At, inspired by the `iat` claim in JWTs.

lib/msal-core/src/utils/RequestUtils.ts
```diff
@@ -140,7 +146,33 @@ export class RequestUtils {
...
+            iat: TimeUtils.now()
+        };
+
+        const stateString = JSON.stringify(stateObject);
+
+        return CryptoUtils.base64Encode(stateString);
+    }
+
+    static parseLibraryState(state: string): StateObject {
+        if (CryptoUtils.isGuid(state)) {
```
DarylThayil: if we release this there should be no need for backwards compat right? any state read should be set by the same lib?

DarylThayil:
what can be tested?

lib/msal-core/src/utils/RequestUtils.ts
jasonnutter: This would namely be necessary for a scenario where an older version of the library generates the state value and then the new version of the library tries to read that value.

lib/msal-core/src/UserAgentApplication.ts
DarylThayil: I wonder if it is better to not resemble a token claim here to avoid confusion.
This represents a requested at time right?

lib/msal-core/src/utils/RequestUtils.ts
jasonnutter: Which may not be possible in the tab suspension scenario? Not sure yet.

DarylThayil: interesting, that is a valid use case.
Any problem with the reverse? 
old version reading new state?

jasonnutter:
> what can be tested?

Unit tests for RequestUtils (and other tests that rely on state), and manual testing of the tab suspension scenario.

DarylThayil:
Any concerns on char length with state?

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: Yeah, thats fair. Could be something like `ts` (probably should be kept short)

lib/msal-core/src/utils/RequestUtils.ts
jasonnutter: That would not be supported.

lib/msal-core/src/UserAgentApplication.ts
DarylThayil: its encoded so length should be ok right?

lib/msal-core/src/utils/RequestUtils.ts
DarylThayil: It should be supported right? it might just view it as a string and not decode it but the state would match?

jasonnutter:
> Any concerns on char length with state?

Until we add more values to the object, I think it should be a consistent 88 characters. That is definitely a bit longer than before, but I don't think its too long. We'll want to double check that as well.

DarylThayil:
Question, the part we are validating is still the full string we are sending right? we are not cracking it open when it comes back and saying does state match stateobj.state?

jasonnutter:
> Question, the part we are validating is still the full string we are sending right? we are not cracking it open when it comes back and saying does state match stateobj.state?

No, this would still decode the state object (which we need to do anyway) and then does checks with the guid.

lib/msal-core/src/utils/RequestUtils.ts
jasonnutter: I suppose so? Regardless, I don't think thats a use case that needs to be supported.

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: Base64 encoded length still grows as the string itself grows. Main benefit encoding the special characters and whitespace.

lib/msal-core/src/utils/RequestUtils.ts
sameerag: I don't think that is a use case to be supported. I would rather prefer to kill the request than process it when the library version changes.

sameerag:
Can we add the 'request` type to the `state` and calculate the encode decode length? we need that to fix popup scenarios in teams anyway.

lib/msal-core/src/UserAgentApplication.ts
DarylThayil: +1 makes sense, I prefer explicit naming but if we need to balance it with size thats fine

DarylThayil (commented):
> > Question, the part we are validating is still the full string we are sending right? we are not cracking it open when it comes back and saying does state match stateobj.state?
> 
> No, this would still decode the state object (which we need to do anyway) and then does checks with the guid.

Seems like whatever we send to state (an encoded string) is what we should validate comes back.  For instance if I create an object 
```js
const stateToSend  = {
   state: "a",
  otherItem: "b"
}

encode(stateToSend) = abcdefg
```

and then another state comes back with the same state parameter but maybe different other parameters 

```js
stateThatComesBack = hijklmnop

decode(statethatComesBack) = {
     state: "a",
     otherItem: "c"  
}
```

This decoded State would evaluate to true, but the Other Item would not which could be a potential surface area for attack.  I would imagine the check should be `abcdefg === abcdefg` and that anything we do when we decode it from there on can be trusted

jasonnutter:
@DarylThayil Yeah, that's fair.

sangonzal (approved):
Looks good. It might be useful to have some logging around what ts is added and returned in state object for debugging.

lib/msal-core/src/utils/RequestUtils.ts
```diff
@@ -140,7 +147,50 @@ export class RequestUtils {
...
+     */
+    static parseLibraryState(state: string): StateObject {
+        const libraryState = state.split("|")[0];
+
+        if (CryptoUtils.isGuid(libraryState)) {
+            return {
+                state,
+                ts: TimeUtils.now()
+            };
+        }
```
jmckennon: did we ever think of the potential use case for this? it feels like we should figure out if this is necessary or not, as it does affect how things work downstream

lib/msal-core/src/utils/RequestUtils.ts
```diff
@@ -10,6 +10,13 @@ import { ScopeSet } from "../ScopeSet";
 import { StringDict } from "../MsalTypes";
 import { StringUtils } from "../utils/StringUtils";
 import { CryptoUtils } from "../utils/CryptoUtils";
+import { TimeUtils } from "./TimeUtils";
+import { ClientAuthError } from "../error/ClientAuthError";
+
+export type LibraryStateObject = {
+    state: string,
```
jasonnutter: Per @pkanher617: Change this to something like `id`.

lib/msal-core/src/utils/RequestUtils.ts
jasonnutter: Yes, a potential use case is a redirect flow: A user initiates a redirect action, and while they are away from the application, the application is updated to use the new format. When they return to the app, their state value would be a guid, and if this code isn't included, the request would fail. This would only happen once, but I think should be accounted for. This would require #1413 also having similar fallback logic to properly handle this use case.

pkanher617 (approved):
+1 for naming changes. Looks good now. I'm assuming you tested this with the react-sample-app but it may be good to add an e2e test or two for this.

jasonnutter:
@pkanher617 I'm not sure how to test something like tab suspending in an automated environment, unfortunately. I think we'll have to rely on unit tests and manual tests for that specific scenario.

pkanher617:
> @pkanher617 I'm not sure how to test something like tab suspending in an automated environment, unfortunately. I think we'll have to rely on unit tests and manual tests for that specific scenario.

Can't we just add a test in the basic use case to test that the state changes are working as expected? Or do you think unit tests are enough to test this? We don't necessarily need the specific case for #1259 in end-to-end, but something to test with the service would be good.

jasonnutter:
> Can't we just add a test in the basic use case to test that the state changes are working as expected? Or do you think unit tests are enough to test this? We don't necessarily need the specific case for #1259 in end-to-end, but something to test with the service would be good.

I would expect the existing e2e tests to fail if the state changes weren't correct, yeah? As the library should throw if state is invalid.

jasonnutter:
Add a test to mimic tab suspending, which I verified fails if the old logic is used.

PR #1399, opened 2020-03-19
Http event

DarylThayil:
add all http and api events

* Changes to Xhr Client to enable instrumentation
* wired all top level apis through


lib/msal-core/src/XHRClient.ts
```diff
@@ -31,7 +32,10 @@ export class XhrClient {
                     reject(this.handleError(xhr.responseText));
                 }
 
-                resolve(jsonResponse);
+                resolve({
+                    client: xhr,
```
jasonnutter: We had talked (in the context of the throttling work) that this would return a few known properties, such as headers, status, and body. I would prefer that over returning the entire `xhr` object.

jasonnutter: And you shuold add a new type to represent this object, instead of `any`.

jasonnutter:
So I'm not the biggest fan of having the telemetry logic living right next to business logic like this, which is why when I proposed a "middleware" type approach to preprocessing and postprocessing when we were talking about throttling.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -335,6 +344,15 @@ export class UserAgentApplication {
...
         // validate request
         const request: AuthenticationParameters = RequestUtils.validateRequest(userRequest, false, this.clientId, Constants.interactionTypeRedirect, this.redirectCallbacksSet);
+
+        const requestCorrelationId = userRequest.correlationId || CryptoUtils.createNewGuid();
+        const apiEvent: ApiEvent = new ApiEvent(requestCorrelationId, this.logger);
+        apiEvent.apiEventIdentifier = API_EVENT_IDENTIFIER.AcquireTokenRedirect;
+        apiEvent.apiCode = API_CODE.AcquireTokenRedirect;
+        this.telemetryManager.startEvent(apiEvent);
+        this.telemetryManager.stopEvent(apiEvent);
+        this.telemetryManager.flush(requestCorrelationId);
```
jasonnutter: Can you pull out any blocks like this into helper functions somewhere?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -368,12 +405,29 @@ export class UserAgentApplication {
...
-            throw error;
-        });
+        })
+            .then((resp) => {
+                apiEvent.wasSuccessful = true;
+                return resp;
+            })
+            .catch((error: AuthError) => {
+                this.cacheStorage.resetTempCacheItems(request.state);
+                apiEvent.apiErrorCode = error.errorCode;
```
jasonnutter: I would prefer to see this more cleanly separated from business logic, if possible.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -323,6 +324,14 @@ export class UserAgentApplication {
     loginRedirect(userRequest?: AuthenticationParameters): void {
         // validate request
         const request: AuthenticationParameters = RequestUtils.validateRequest(userRequest, true, this.clientId, Constants.interactionTypeRedirect, this.redirectCallbacksSet);
+
+        const requestCorrelationId = userRequest.correlationId || CryptoUtils.createNewGuid();
+        const apiEvent: ApiEvent = new ApiEvent(requestCorrelationId, this.logger);
+        apiEvent.apiEventIdentifier = API_EVENT_IDENTIFIER.LoginRedirect;
+        apiEvent.apiCode = API_CODE.LoginRedirect;
```
tnorling: Should apiEventIdentifier and apiCode be part of the constructor? Looks like they are always set right after initialization?

lib/msal-core/src/authority/AadAuthority.ts
```diff
@@ -29,17 +31,30 @@ export class AadAuthority extends Authority {
...
         }
 
         // for custom domains in AAD where we query the service for the Instance discovery
         const client: XhrClient = new XhrClient();
 
-        return client.sendRequestAsync(this.AadInstanceDiscoveryEndpointUrl, "GET", true)
+        const httpMethod = "GET";
+        const httpEvent = new HttpEvent(correlationId);
+        httpEvent.url = this.AadInstanceDiscoveryEndpointUrl;
+        httpEvent.httpMethod = httpMethod;
```
tnorling: Same question as above

lib/msal-core/src/telemetry/TelemetryUtils.ts
```diff
@@ -17,7 +17,7 @@ export const scrubTenantFromUri = (uri: string): String => {
     const pathParams = url.PathSegments;
 
     if (pathParams && pathParams.length >= 2) {
-        const tenantPosition = pathParams[1] ===  B2cAuthority.B2C_PREFIX ? 2 : 1;
+        const tenantPosition = pathParams[1] ===  "tfp" ? 2 : 1;
```
tnorling: Are we still using tfp for anything?

DarylThayil: Not sure, have we stopped doing this? We do use it for taking PII out of urls (tenants)

DarylThayil:
@jasonnutter I sort of agree with the suggestions here about how interwoven things are... However it is not clear to me the best path forward. 
The middleware approach is interesting but doesn't really fit inline with what we do because we need to be able to instrument at any point in the code and have the "top level" flush at the right time. 
Very open to suggestions

lib/msal-core/src/telemetry/TelemetryUtils.ts
tnorling: My understanding was that all the b2c urls moved to something.b2clogin.com/something.onmicrosoft.com/policy and something.b2clogin.com/tenant_id/policy
@sameerag Was that your understanding as well?

lib/msal-core/src/UserAgentApplication.ts
sameerag: +1. We worked to simplify all public APIs and it helps to declutter and manage helper functions.

lib/msal-core/src/UserAgentApplication.ts
sameerag: May be add a helper that:
- Cleans the cache
- Adds telemetry instrumentation
?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -907,19 +966,35 @@ export class UserAgentApplication {
...
+                this.telemetryManager.stopEvent(apiEvent);
+                this.telemetryManager.flush(requestCorrelationId);
+                this.navigateWindow(urlNavigate);
+            })
+            .catch((error: AuthError) => {
+                apiEvent.wasSuccessful = false;
+                apiEvent.apiErrorCode = error.errorCode;
+                this.telemetryManager.stopEvent(apiEvent);
+                this.telemetryManager.flush(requestCorrelationId);
+            });
```
sameerag: Same comment here, please separate this from the business logic.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -2101,7 +2176,8 @@ export class UserAgentApplication {
             scopes: [this.clientId],
             authority: this.authority,
             account: this.getAccount(),
-            extraQueryParameters: request.extraQueryParameters
+            extraQueryParameters: request.extraQueryParameters,
+            correlationId: request.correlationId
```
sameerag: CorrelationId is not always set by the user. I presume we already are validating and adding this to the request by the time we execute this?

lib/msal-core/src/telemetry/TelemetryUtils.ts
sameerag: yes. The more I think, the more I just want to have a type and be done with it. :-). Most MSALs still have this logic + indication from the client that this is 'B2C'

tnorling: So does the distinction between AAD and B2C matter in this example? I think the tenant info will always be in position 1 now right?

sameerag: But, we need it for legacy apps compatibility, the migration may not be yet complete.

lib/msal-core/src/authority/AadAuthority.ts
sameerag: Helper please!

jasonnutter:
@DarylThayil Looks like there are some feedback that needs addressing?

lib/msal-core/src/XHRClient.ts
DarylThayil: Added  XhrResponse type with responseBody, and statusCode can add other valuable attributes as needed

lib/msal-core/src/UserAgentApplication.ts
DarylThayil: @jasonnutter  we paired on this and resolved

lib/msal-core/src/UserAgentApplication.ts
DarylThayil: resolved

lib/msal-core/src/UserAgentApplication.ts
DarylThayil: Updated (ish) to be a little cleaner

lib/msal-core/src/authority/AadAuthority.ts
DarylThayil: Done!  added a helper for this

lib/msal-core/src/telemetry/TelemetryUtils.ts
DarylThayil: @tnorling  can you share where we landed on this?

lib/msal-core/src/UserAgentApplication.ts
DarylThayil: updated and cleaned up

DarylThayil:
Addressed all feedback, waiting to hear on the TFP

lib/msal-core/src/telemetry/TelemetryUtils.ts
tnorling: @DarylThayil There are still cases where b2c authority urls will contain "tfp", so while this check was no longer a good check for the AuthorityFactory it is still needed for this use case here. This can remain as written.

lib/msal-core/src/XHRClient.ts
```diff
@@ -30,8 +33,11 @@ export class XhrClient {
                 } catch (e) {
                     reject(this.handleError(xhr.responseText));
                 }
-
-                resolve(jsonResponse);
+                const response: XhrResponse = {
+                    statusCode: xhr.status,
+                    responseBody: jsonResponse
```
jasonnutter, line 38: I think this can just be `body`, so that it is `response.body` instead of `response.responseBody`.

lib/msal-core/src/telemetry/TelemetryUtils.ts
jasonnutter: @tnorling @DarylThayil Which version should be used here? `B2cAuthority.B2C_PREFIX` or `tfp`?

jasonnutter:
One other thought: we should indicate whether or not a given event was the result of a cache miss or cache hit.

lib/msal-core/src/telemetry/TelemetryUtils.ts
tnorling: I'd probably just go with the ```tfp``` since this will likely be the only place we reference it. But functionally either works.

DarylThayil: Keeping this for now

lib/msal-core/src/XHRClient.ts
DarylThayil: Updated

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -350,13 +351,20 @@ export class UserAgentApplication {
...
+        })
+            .then((resp) => {
+                this.telemetryManager.stopAndFlushApiEvent(request.correlationId, apiEvent, true);
+                return resp;
+            })
+            .catch((error: AuthError) => {
+                this.cacheStorage.resetTempCacheItems(request.state);
+                this.telemetryManager.stopAndFlushApiEvent(request.correlationId, apiEvent, false, error.errorCode);
+                throw error;
+            });
```
sameerag, line 367: nit: In future, we should have a OnSuccess and OnFailure handlers for all Public APIs may be.

DarylThayil: Can you go into more detail what you mean here?

sameerag: For all public APIs, we now clear cache, update telemetry on both success and failure. I am suggesting we can do something like this:

```javascript
loginPopup().then( () => {
    onSuccess();
}).catch( () => {
    onFailure();
});

function onSuccess() {
    clearCache();
    updateTelemetry();
};
...
...
```

PR #1409, opened 2020-03-23
 [msal-node][msal-common] Client applications, authorization code and device code implementation

sangonzal:
Adds implementation public APIs and implementation for authorization code flow and device code flow. Will add a separate PR for unit tests- won't merge this until that is out and reviewed. 

Adds to msal-common: 
- AuthorizationCodeClient - creating authorization endpoint URI and then making token request
- DeviceCodeClient - device code flow 
- BaseClient - will be extended by all oauth clients (refresh token and silent coming next)
- Configuration - configuration of for oauth clients 

response: 
- DeviceCodeResponse - response from first step of device code flow. Returned to user via callback. 
- AuthenticationResult - value returned to the user after a successful token acquisition. Have not finalized if this will be added. 

request:  
- AuthorizationCodeRequest - ClientApp.acquireTokenByAuthCode(AuthorizationCodeRequest)
- AuthorizationCodeUrlRequest - ClientApp.getAuthCodeUrl(AuthorizationCodeUrlRequest)
- DeviceCodeRequest - PublicClientApp.acquireTokenByDeviceCode(DeviceCodeRequest)
- RequestValidator - validation utils for requests 
- RequestUtils - utils for constructing requests 
- CancellationToken - for cancelling polling in device code flow  

Add to msal-node: 
- ClientApplication - abstract base class for PublicClientApp and ConfidentialClientApp. Contains APIs shared by both. Not exported. 
- PublicClientApplication - public client flows. 
- ConfidentialClientApplication - confidential client flows.
- ClientConfiguration - configuration for PublicClientApplication and ConfidentialClientApplication. 
- HeaderUtils - Injecting node default headers into msal-common

Adds to samples: 
- msal-node-auth-code: express app showing authorization code flow
- msal-node-device-code: console app showing device code flow


lib/msal-node/src/cache/Storage.ts
```diff
@@ -0,0 +1,231 @@
+/*
```
sangonzal: Please ignore this file. It's here as placeholder, and @sameerag plans to update this with her cache work.

lib/msal-common/src/request/RequestValidator.ts
```diff
@@ -0,0 +1,82 @@
+/*
```
sangonzal: Considering moving these methods to RequestUtils. Thoughts?

lib/msal-common/src/request/CancellationToken.ts
```diff
@@ -0,0 +1,18 @@
...
+ * Licensed under the MIT License.
+ */
+
+/**
+* Cancellation token used for cancelling the device code request. 
+* 
+* While the user authenticates on a separate device, MSAL polls the the token endpoint of security token service for the interval 
+* specified in the device code resonse (usually 15 minutes). To stop polling and cancel the request, set cancellationToken.cancel = true. 
+* */
+export class CancellationToken {
```
jmckennon: I may just be ignorant here, but is the plan to add to this class? what's the purpose of having a class with just one field and nothing else?

lib/msal-common/src/request/RequestValidator.ts
jmckennon: IMO this file here is actually more representative of a `RequestUtils` file and the other (current) `RequestUtils` file is really just a bunch of helper/setter functions that should probably live where that request it being built

lib/msal-common/src/request/AuthorizationCodeRequest.ts
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+/**
+ * @type AuthorizationCodeRequest: Request object passed by user to acquire a token from the server exchanging a valid authorization code
+ *  (second leg of OAuth2.0 Authorization Code flow)
+ *
+ * scopes:                  A space-separated array of scopes for the same resource.
```
jmckennon: Just wanted to say that I appreciate the effort that gets put into the readability of many of these comments, well done

lib/msal-common/src/utils/RequestUtils.ts
```diff
@@ -0,0 +1,239 @@
...
+            encodeURIComponent("0.0.1")
+        );
+    }
+
+    /**
+     * Utility to create a URL from the params map
+     * @param paramsMap
+     * @param authority
+     */
+    static createQueryString(paramsMap: Map<string, string>): string {
```
jmckennon: Similar to the comment on `RequestValidator`, I think every function in this file probably belongs somewhere in the `request` folder else except for this one as `createQueryString`

lib/msal-common/src/request/CancellationToken.ts
jasonnutter: CancellationToken is a pattern used to be able to short-circuit a long running operation.

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -0,0 +1,185 @@
...
+import { Configuration } from "../config/Configuration";
+
+/**
+ * Oauth2.0 Authorization Code client
+ */
+export class AuthorizationCodeClient extends BaseClient {
+
+    constructor(configuration: Configuration) {
+
+        super({
```
DarylThayil: is there a reason we would not just pass the whole configuration object here? looks to have the same structure

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -0,0 +1,185 @@
...
+     * sent in the request and should contain an authorization code, which can then be used to acquire tokens via
+     * acquireToken(AuthorizationCodeRequest)
+     * @param request
+     */
+    async getAuthCodeUrl(request: AuthorizationCodeUrlRequest): Promise<string> {
+
+        const authority: Authority = await this.createAuthority(request && request.authority);
+        const queryParams: Map<string, string> = this.generateAuthCodeUrlParams(request);
+        const queryString = RequestUtils.createQueryString(queryParams);
+        return authority.authorizationEndpoint + "?" + queryString;
```
DarylThayil: can use string interpolation
```js
return `${authority.authorizationEndpoint}?${queryString}`;
```

lib/msal-common/src/client/BaseClient.ts
```diff
@@ -79,85 +75,60 @@ export abstract class BaseClient {
...
-            const idToken = new IdToken(rawIdToken, this.cryptoObj);
-            const clientInfo = buildClientInfo(rawClientInfo, this.cryptoObj);
-
-            this.account = Account.createAccount(idToken, clientInfo, this.cryptoObj);
-            return this.account;
-        }
-
-        // if login is not yet done, return null
-        return null;
+    protected async executePostToTokenEndpoint(
```
DarylThayil: can we add return type here to make it clear we return a promise?

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -0,0 +1,185 @@
...
+     */
+    private async executeTokenRequest(authority: Authority, request: AuthorizationCodeRequest): Promise<string> {
+
+        const tokenParameters: Map<string, string> = this.generateAuthCodeParams(request);
+        const requestBody = RequestUtils.createQueryString(tokenParameters);
+        const headers: Map<string, string> = this.createDefaultTokenRequestHeaders();
+
+        let acquiredTokenResponse;
+        try {
+            acquiredTokenResponse = this.executePostToTokenEndpoint(
```
DarylThayil: I think you want to make this as await ?

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -0,0 +1,185 @@
...
+     * @param authority
+     * @param request
+     */
+    private async executeTokenRequest(authority: Authority, request: AuthorizationCodeRequest): Promise<string> {
+
+        const tokenParameters: Map<string, string> = this.generateAuthCodeParams(request);
+        const requestBody = RequestUtils.createQueryString(tokenParameters);
+        const headers: Map<string, string> = this.createDefaultTokenRequestHeaders();
+
+        let acquiredTokenResponse;
```
DarylThayil: this block can be written 

```js 
   
return this.executePostToTokenEndpoint( ... )
        .catch((err) => {
                 //do your logging then
                throw err;
       });
```

lib/msal-common/src/client/AuthorizationCodeClient.ts
DarylThayil: actually since this returns a promise never mind

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -0,0 +1,185 @@
...
+
+        // add code: user set, not validated
+        RequestUtils.addAuthorizationCode(paramsMap, request.code);
+
+        // add code_verifier if passed
+        if (request.codeVerifier) {
+            RequestUtils.addCodeVerifier(paramsMap, request.codeVerifier);
+        }
+
+        RequestUtils.addGrantType(paramsMap, GrantType.AUTHORIZATION_CODE_GRANT);
```
DarylThayil: instead of side effects is there a way we can do something like this 
```js
const paramsMap: Map<string, string> RequestUtils.generateParamsMapFromRequest):
```

then in  the utils you can do all your logic around setting int, but writing functions that take maps and just "alter in place" can create bugs and is harder to test.

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -0,0 +1,185 @@
...
+        }
+
+        if (request.nonce) {
+            RequestUtils.addNonce(paramsMap, request.nonce);
+        }
+
+        if(request.claims) {
+            RequestUtils.addClaims(paramsMap, request.claims);
+        }
+
```
DarylThayil: Similar feedback here around side effects.

lib/msal-common/src/client/BaseClient.ts
```diff
@@ -79,85 +75,60 @@ export abstract class BaseClient {
...
 
-    /**
-     * Given an authorization code, it will perform a token exchange using cached values from a previous call to
-     * createLoginUrl() or createAcquireTokenUrl(). You must call this AFTER using one of those APIs first. You should
-     * also use the handleFragmentResponse() API to pass the codeResponse to this function afterwards.
-     * @param codeResponse
-     */
-    abstract async acquireToken(codeResponse: CodeResponse): Promise<TokenResponse>;
+        try {
+            await authority.resolveEndpointsAsync();
```
DarylThayil: does resolveEndpointAsync throw an error on its own?

lib/msal-common/src/client/BaseClient.ts
```diff
@@ -79,85 +75,60 @@ export abstract class BaseClient {
...
+        }
 
-    /**
-     * Retrieves a token from cache if it is still valid, or uses the cached refresh token to renew
-     * the given token and returns the renewed token. Will throw an error if login is not completed (unless
-     * id tokens are not being renewed).
-     * @param request
-     */
-    abstract async renewToken(request: TokenRenewParameters): Promise<TokenResponse>;
+        return authority;
```
DarylThayil: authority is not a promise right ? you want to return authority.resolveEndpointsAsync

lib/msal-common/src/client/BaseClient.ts
```diff
@@ -79,85 +75,60 @@ export abstract class BaseClient {
...
-    abstract async logout(authorityUri?: string): Promise<string>;
-
-    // #endregion
-
-    // #region Getters and Setters
+    protected createDefaultTokenRequestHeaders(): Map<string, string> {
+        const headers = new Map<string, string>();
+        RequestUtils.addContentTypeHeader(headers);
+        RequestUtils.addLibrarydataHeaders(headers);
+        return headers;
```
DarylThayil: same comment about sideeffects

lib/msal-common/src/client/DeviceCodeClient.ts
```diff
@@ -0,0 +1,190 @@
...
+
+/**
+ * OAuth2.0 Device code client
+ */
+export class DeviceCodeClient extends BaseClient {
+
+    private authority: Authority;
+
+    constructor(configuration: Configuration){
+        super({
```
DarylThayil: can we just pass through config?

lib/msal-common/src/client/DeviceCodeClient.ts
```diff
@@ -0,0 +1,190 @@
...
+
+    /**
+     * Creates device code request and executes http GET
+     * @param request
+     */
+    private async getDeviceCode(request: DeviceCodeRequest): Promise<DeviceCodeResponse> {
+
+        const deviceCodeUrl = this.createDeviceCodeUrl(request);
+        const headers: Map<string, string> = new Map<string, string>();
+        RequestUtils.addLibrarydataHeaders(headers);
```
DarylThayil: same comment on side effects

lib/msal-common/src/client/DeviceCodeClient.ts
```diff
@@ -0,0 +1,190 @@
...
+
+        return this.executeGetRequest(deviceCodeUrl, headers);
+    }
+
+    /**
+     * Executes GET request to device code endpoint
+     * @param deviceCodeUrl
+     * @param headers
+     */
+    private async executeGetRequest(deviceCodeUrl: string, headers: Map<string, string>): Promise<DeviceCodeResponse>{
```
DarylThayil: this could be named more descriptively

lib/msal-common/src/client/DeviceCodeClient.ts
```diff
@@ -0,0 +1,190 @@
...
+
+        let deviceCodeResponse;
```
DarylThayil: return this.networkClient.sendGetRequestAsync(...)
    .catch(err => {
          //log error and then
          throw err;    
})

lib/msal-common/src/client/DeviceCodeClient.ts
```diff
@@ -0,0 +1,190 @@
...
+    /**
+     * Create device code endpoint url
+     * @param request
+     */
+    private createDeviceCodeUrl(request: DeviceCodeRequest) : string {
+        const params: Map<string, string> = this.createQueryParameters(request);
+        const queryString: string = RequestUtils.createQueryString(params);
+
+        // TODO add device code endpoint to authority class
+        return this.authority.canonicalAuthority + "oauth2/v2.0/devicecode"  + "?" + queryString;
```
DarylThayil: use string interpolation

DarylThayil: The device code path should be stored in a constant somewhere

lib/msal-common/src/client/DeviceCodeClient.ts
```diff
@@ -0,0 +1,190 @@
...
+    private createQueryParameters(request: DeviceCodeRequest): Map<string, string>{
+
+        const params: Map<string, string> = new Map<string, string>();
+        RequestUtils.addClientId(params, this.config.authOptions.clientId);
+
+        const scopes = RequestValidator.validateAndGenerateScopes(
+            request.scopes,
+            this.config.authOptions.clientId
+        );
+        RequestUtils.addScopes(params, scopes);
```
DarylThayil: same comment on side effects

lib/msal-common/src/client/DeviceCodeClient.ts
```diff
@@ -0,0 +1,190 @@
...
+
+        return params;
+    }
+
+    /**
+     * Creates token request with device code response and polls token endpoint
+     * @param request
+     * @param deviceCodeResponse
+     */
+    private async acquireTokenWithDeviceCode(
```
DarylThayil: I wonder if this function could be named to denote it is starting a poll for device code

lib/msal-common/src/client/DeviceCodeClient.ts
```diff
@@ -0,0 +1,190 @@
...
+        const scopes = RequestValidator.validateAndGenerateScopes(
+            request.scopes,
+            this.config.authOptions.clientId
+        );
+
+        const params: Map<string, string> = new Map<string, string>();
+        RequestUtils.addScopes(params, scopes);
+        RequestUtils.addClientId(params, this.config.authOptions.clientId);
+        RequestUtils.addGrantType(params, GrantType.DEVICE_CODE_GRANT);
+        RequestUtils.addDeviceCode(params, deviceCodeResponse.device_code);
```
DarylThayil: same comment on side effects

lib/msal-common/src/utils/RequestUtils.ts
```diff
@@ -0,0 +1,248 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { AADServerParamKeys, SSOTypes } from "./Constants";
+import { Constants, HeaderNames } from "./Constants";
+import { ScopeSet } from "../request/ScopeSet";
+import { ClientConfigurationError } from "../error/ClientConfigurationError";
+
+export class RequestUtils {
```
DarylThayil: This can be trimmed to a non side effect function  like getparamsfromrequest(request)
map = new map()
if request.something
    map.addSomething
if map.somethingElse
   mapp.addSomethingElse


the benefit is that this function will be idempotent. Given the same input the output will always be the same, when passing in an object to alter, its too easy to end up in hard to trace conditions.

DarylThayil:
made lots of comments, but overall looking good.  It is definitely a challenge to review in totality with the number of changes.  In the future I would encourage thinking through how to do this in say 10 smaller prs.

Maybe something like 
1. stub apis
2. implement request utils
3. implement http client for device code and auth code
4. implement pol for device code
5. etc....

That way you can add tests as you go for each stage and the mental over head of each pr is fairly small. Even if you end up altering previous pieces as you go its no big deal because the number of changes is fairly low.

 I also find that it helps me high level plan out my design better to do smaller prs.

lib/msal-common/src/client/AuthorizationCodeClient.ts
sangonzal: Yep you're right, should be passing configuration object instead - updated both here and DeviceCodeClient

lib/msal-common/src/client/BaseClient.ts
sangonzal: Return type is actually a couple of lines down.

lib/msal-common/src/client/AuthorizationCodeClient.ts
sangonzal: updated

lib/msal-common/src/request/RequestValidator.ts
sangonzal: I've refactored away from using the static methods and just created a "ParametersBuilder" that contains these methods, as they will be shared by all the clients in msal-common

lib/msal-common/src/client/BaseClient.ts
sangonzal: It's an async function, so authority get's wrapped in a promise.

Issue #1301, opened 2020-02-24
Redirect loop when linking to the site with target="_blank"


burns47:
>> Please follow the issue template below. Failure to do so will result in a delay in answering your question.

## Library
- [x] `msal@1.x.x` or `@azure/msal@1.x.x`
- [ ] `@azure/msal-browser@2.x.x`
- [ ] `@azure/msal-angular@0.x.x`
- [ ] `@azure/msal-angular@1.x.x`
- [ ] `@azure/msal-angularjs@1.x.x`

**Important**: Please fill in your exact version number above, e.g. `msal@1.1.3`.

## Framework
Angular 9

## Description
A link to an angular app using target="_blank" results in an infinite loop since the MSAL library thinks that the real window is a popup.  It uses the following line of code from `WindowUtils` to check if we're in a popup context:

```
    static isInPopup(): boolean {
        return !!(window.opener && window.opener !== window);
    }
```

in Chrome at least, when linking to the app using target="_blank" results in window.opener being non-null *and* pointing to the old page.  So the MSAL library thinks we're in a popup, neglects to handle the hash and (later down the line) my app thinks the login didn't occur and redirects again;  infinite loop.

## Security
No

## Regression
Unknown

## Configuration

```js
 MsalModule.forRoot({
      auth: {
        clientId: environment.msalConfig.clientId,
        authority: environment.msalConfig.authority,
        redirectUri: window.location.origin,
        validateAuthority: false,
        navigateToLoginRequestUrl: false
      },
      cache: {
        cacheLocation: 'sessionStorage',
        storeAuthStateInCookie: isIE
      },
      system: {
      }
    }
      , {
        protectedResourceMap: environment.msalConfig.resourceMap,
        unprotectedResourc
```

## Reproduction steps
1. 
2. 
3. 

## Expected behavior
Links opened in new tabs should function correctly. 

## Browsers
I bleieve all browsers expose `window.opener` to pages opened with `target="_blank"`, so they should all have this problem.

jasonnutter:
@burns47 This is a known issue, we plan to have a fix soon. cc: @sameerag

sameerag:
Status: Working on a POC and will release a beta to get this rolling, potentially later this week or early next week.

jasonnutter:
`msal@1.3.0-beta.1` is available with a fix for this issue, please let us know if it is still a problem. Thanks!

turculaurentiu91:
[...]

Issue #1198, opened 2020-01-11
Redirect callback is not called when loginRedirect() is run from a popup window


ehmPlankje:
Currently MSAL.js uses the `WindowUtils.isInPopup()` method to determine if the redirect or popup flow should be applied. However, in my use case, I want to be able to use the login redirect flow from a popup window. Perhaps it would be possible to create a configuration option where the developer can declare explicitly that MSAL.js should use the redirect or popup flow, instead of trying to deduct it from properties of the Window object.

## Library
- [x] `msal@1.2.0`

## Framework
No framework required.

## Description
When MSAL.js is run from a popup window that was created by its calling application, the redirect callback is not called when the `loginRedirect()` method is used. This is caused by the `WindowUtils.isInPopup()` method of returning true, despite MSAL not using the `loginPopup()` flow.

## Security
No security issue.

## Regression
No regression.

## Configuration
Please provide your MSAL configuration options.

```js
auth: {
    clientId: clientId,
    authority: process.env.AUTH_AUTHORITY,
    redirectUri: process.env.AUTH_URL,
    postLogoutRedirectUri: process.env.AUTH_URL,
    navigateToLoginRequestUrl: false
},
cache: {
    cacheLocation: 'localStorage',
    storeAuthStateInCookie: true
}
```

## Reproduction steps
1. Implement a simple page to login a user via MSAL.js with the **redirect flow** `UserAgentApplication.loginRedirect()`. 
2. Create a new page that opens the page with the implementation of step 1 in a **popup window** with `window.open()`.
3. Complete the login flow, and verify that the redirect callback that was declared in `UserAgentApplication.handleRedirectCallback()` is not called when the user returns from the authentication provider.

## Expected behavior
The callback method that was declared with `UserAgentApplication.handleRedirectCallback()` is called when the user returns from the authentication provider.

## Browsers
Tested in Edge and Chrome, latest versions.

sameerag:
@ehmPlankje We are aware of this limitation and this issue is being tagged with #1095 (Teams attempts to do exactly the same - calls `loginRedirect` from pop up); The solution needs #1169 to be supported first and hence we delayed fixing this right away.

We plan to address this by the end of the month. Hope this clarifies.

ehmPlankje:
@sameerag Thank you for the feedback. I already solved the issue for my stack by applying a patch on the `WindowUtils.isInPopup()` method, so that it always returns `false`. 

[...]

Issue #1095, opened 2019-11-05
Login Flow not working for MS Teams App with 1.2.0-Beta.3


davelosert:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Other... Please describe:
</code></pre>

## Browser:
- [X] Chrome version 78.0
- [X] Firefox version 70.0
- [ ] IE version XX
- [ ] Edge version XX
- [ ] Safari version XX

## Library version

<pre><code>
Library version: 1.2.0-beta.3

</code></pre>

## Current behavior

I am trying to make the authentication work in a MS Teams Tab App using MSAL.JS

MS Teams requires me to open a popup for authentication with it's own method (`   microsoftTeams.authentication.authenticate`) as normal popups are blocked by the MS Teams client. So I can not use the `loginPopup()`-Method of MSAL but have to use `loginRedirect()` from within the MS Teams Popup.

The authentication itself works and redirects back with a valid id_token in the url-hash. However, the hash is not handled and the redirect callback nevers gets called.

The reason is [Line227](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/160efdc23963af376b242328de91e51ff9cb6eff/lib/msal-core/src/UserAgentApplication.ts#L227) in the `constructor` of the `UserAgentApplication`:
```JavaScript
if (!this.config.framework.isAngular && urlContainsHash && !WindowUtils.isInIframe() && !WindowUtils.isInPopup()) {
            this.handleAuthenticationResponse(urlHash);
}
```

The last condition `WindowUtils.isInPopup()` is `true` in this case, and therefore `handleAuthenticationResponse()` is never called.

I don't know the reason for this check in this particular case - but is there a way to work around this issue?

Falling back to an older version of MSAL is also no option as I also need the new IFrame-Capapbility to use `acquireTokenSilent()`. I also tried adal.js, but that is currently not working on mobile due to another IFrame issue.

## Expected behavior

The callback of `loginRedirect()` is also called within a popup-window.

## Minimal reproduction of the problem with instructions

Use `loginRedirect()` from within a popup window.

jasonnutter:
[...]

PR #1413, opened 2020-03-24
Ensure responses from redirect requests are always processed

jasonnutter:
When `UserAgentApplication` is instantiated, it will check to see if there is a hash in the url and that we are not in a popup or iframe. If both are true, it infers that the request was a redirect, and thus the hash should be processed. The problem is that there are scenarios where the logic to determine whether or not we are in a popup is incorrect, such as when the window is opened in a new tab, or we are calling a redirect method inside a popup. This results in these checks failing, and the hash not getting processed.

To address this, we will now include a value in the encoded state value that will indicate what type of request method was used (popup, redirect, silent). When we receive a response we will check to see what the request method is, and if it was a redirect, always process it (regardless of whether or not we are in a popup, for example).

This PR builds on:
- #1358 
- #1395 

This PR should fix:
- #1198 
- #1095 
- #1301 

TODO (@tnorling):
- [x] Fix existing tests (@jasonnutter)
- [x] Unit tests
- [x] Verify scenarios listed above are fixed
- [x] Additional code cleanup


lib/msal-core/src/utils/RequestUtils.ts
```diff
@@ -129,11 +137,55 @@ export class RequestUtils {
...
-        // append GUID to user set state  or set one for the user if null
-        return !StringUtils.isEmpty(state) ? CryptoUtils.createNewGuid() + "|" + state : CryptoUtils.createNewGuid();
+    static parseLibraryState(state: string): LibraryStateObject {
+        const libraryState = state.split(Constants.resourceDelimiter)[0];
+
+        if (CryptoUtils.isGuid(libraryState)) {
+            return {
+                id: libraryState,
+                ts: TimeUtils.now()
+            };
```
jasonnutter: In this scenario, I think we may want to have `method` default to redirect, to error on teh side of the response getting processed (as I believe redirect is the most likely scenario where this could happen).

Weschk:
We had the same issue when using redirectLogin so we are currently using popup. However i just tested this branch and it resolves the issue.
Now we are just waiting for it to be released and we can go back to our original auth flow.
God job @jasonnutter !

lib/msal-core/test/UserAgentApplication.spec.ts
```diff
@@ -580,17 +610,21 @@ describe("UserAgentApplication.ts Class", function () {
...
             window.location = oldWindowLocation;
+            let testsExecuted = false;
             const checkErrorFromLibrary = function (authErr: AuthError) {
                 expect(authErr instanceof ClientAuthError).to.be.true;
                 expect(authErr.errorCode).to.equal(ClientAuthErrorMessage.loginProgressError.code);
                 expect(authErr.errorMessage).to.equal(ClientAuthErrorMessage.loginProgressError.desc);
                 expect(authErr.message).to.equal(ClientAuthErrorMessage.loginProgressError.desc);
                 expect(authErr.name).to.equal("ClientAuthError");
                 expect(authErr.stack).to.include("UserAgentApplication.spec.ts");
+                testsExecuted = true;
```
jasonnutter: The `done` function should be used for this, why is a boolean needed?

tnorling: This is to ensure the callback is actually hit and the tests are run. I was running into an issue where all the tests were passing when they should have been failing. Figured out the callback was never being run and thus no tests to pass or fail.

jasonnutter: Interesting, I thought if you passed `done` the test would timeout if `done` wasn't called...

tnorling: Oh you know what, you're totally right. This was happening on tests that were not marked as async and I just applied this to everything calling ```handleRedirectCallback``` without realizing the difference. So I guess the solution here maybe is to make those other tests async instead of doing this

jasonnutter: @tnorling Ah, yeah using `done` for those tests should fix.

jasonnutter:
@tnorling Updates look good!

PR #1416, opened 2020-03-24
[msal-common] Add knownAuthorities parameter to support B2C validation

sameerag:
This PR adds B2C support for `msal-common`; This also ports the changes needed for knownAuthorities() support from #1276.

cc @tnorling

PR #1424, opened 2020-03-26
[msal-common][msal-node] Authority changes

sameerag:
This PR fixes the authority type resolution runtime and adds `B2CAuthority.ts`:

-  B2C Authority extends the generic Authority class, rather than AADAuthority.
- Added `knownAuthorities` config parameter which sets the trusted authorities for B2C use cases
- Authority factory will create B2C Authority instance if `knownAuthorities` is set, AAD Authority instance otherwise; For `ADFS` we look for /adfs as the first path segment in the URL.

PRs #1276 and #1416 fix this for `msal-core` and `msal-common` (msal-browser based version)


lib/msal-common/src/authority/B2cAuthority.ts
```diff
@@ -0,0 +1,61 @@
...
+ */
+import { Authority } from "./Authority";
+import { OpenIdConfigResponse } from "./OpenIdConfigResponse";
+import { AuthorityType } from "./AuthorityType";
+import { B2CTrustedHostList, Constants } from "../utils/Constants";
+import { INetworkModule } from "../network/INetworkModule";
+import { ClientConfigurationError } from "../error/ClientConfigurationError";
+
+/**
+ * The AadAuthority class extends the Authority class and adds functionality specific to the Azure AD OAuth Authority.
```
sangonzal: All comments in this class mention AAD, but should instead be switched over to B2C.

lib/msal-common/src/client/BaseClient.ts
```diff
@@ -76,6 +78,8 @@ export abstract class BaseClient {
         // Set the network interface
         this.networkClient = this.config.networkInterface;
 
+        this.setKnownAuthorities(this.config.authOptions.knownAuthorities);
```
sangonzal: Should we be checking that validateAuthority is set to true before setting known authorities?

lib/msal-common/src/authority/AuthorityFactory.ts
```diff
@@ -20,14 +22,14 @@ export class AuthorityFactory {
...
-            case "tfp":
-                // tfp denotes a b2c url
-                return AuthorityType.B2C;
-            default:
-                // default authority is always AAD
-                return AuthorityType.Aad;
-        }
+
+        if (pathSegments[0] === "adfs")
+            return AuthorityType.Adfs;
```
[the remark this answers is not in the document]
sameerag: Yes. There is an email, will forward it soon to the team.

lib/msal-common/src/authority/B2cAuthority.ts
```diff
@@ -0,0 +1,61 @@
...
+ */
+export class B2cAuthority extends Authority {
+    // Set authority type to AAD
+    public get authorityType(): AuthorityType {
+        return AuthorityType.Aad;
```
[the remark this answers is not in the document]
sameerag: sorry yes. Correcting.

lib/msal-common/src/client/BaseClient.ts
sameerag: There is no longer a "validateAuthority" field. We validate by default. Users can opt out, we will be adding discovery endpoint for AAD and then provide an option to opt out.

sangonzal: So for B2C people have to pass in knownAuthorities by default? How will users be able to opt out if there is no validateAuthority field?

lib/msal-common/src/authority/B2cAuthority.ts
sameerag: Addressed. Sorry missed a push. Now B2C file is upto date and accurate.

lib/msal-common/src/client/BaseClient.ts
sameerag: yes. They will have to set it by default. That will be the future guidance.

sangonzal: How will users be able to opt out authority validation?

lib/msal-common/src/utils/Constants.ts
```diff
@@ -204,3 +204,8 @@ export enum GrantType {
     DEVICE_CODE_GRANT = "device_code"
 };
 
+/**
+ * Initialize B2CTrustedHostList
+ */
+export const B2CTrustedHostList = {};
```
tnorling: Remove this since we moved it to AuthorityFactory

lib/msal-common/src/config/Configuration.ts
```diff
@@ -39,6 +39,7 @@ export type Configuration = {
 export type AuthOptions = {
     clientId: string;
     authority?: string;
+    knownAuthorities?: Array<string>
```
tnorling: Wherever you set your defaults, this should be defaulted to an empty array

lib/msal-common/src/client/BaseClient.ts
sameerag: By setting `KnownAuthorities()`. That will be the guidance, either send us a list we need to look up to or we will validate as per the standard route.

lib/msal-common/src/utils/Constants.ts
sameerag: Done. Thanks, I missed this. :)

lib/msal-common/src/config/Configuration.ts
sameerag: Done.

lib/msal-common/src/authority/AuthorityFactory.ts
```diff
@@ -20,13 +26,31 @@ export class AuthorityFactory {
...
+    /**
+     * @hidden
+     * @ignore
+     * Use when Authority is B2C is set to True to provide list of allowed domains.
+     * @param knownAuthorities
+     */
+    public static setKnownAuthorities(knownAuthorities: Array<string>): void {
+        if (!Object.keys(B2CTrustedHostList).length){
+            if (!knownAuthorities.length) {
+                throw ClientConfigurationError.createKnownAuthoritiesNotSetError();
```
tnorling: Need to remove this since we got rid of the authorityType parameter. This will throw for AAD use cases, which we dont want

lib/msal-common/src/authority/AuthorityFactory.ts
```diff
@@ -20,13 +26,28 @@ export class AuthorityFactory {
...
+    }
+
+    /**
+     * @hidden
+     * @ignore
+     * Use when Authority is B2C is set to True to provide list of allowed domains.
+     * @param knownAuthorities
+     */
+    public static setKnownAuthorities(knownAuthorities: Array<string>): void {
+        if (!Object.keys(B2CTrustedHostList).length && knownAuthorities) {
```
tnorling, line 47:
```suggestion
        if (!B2CTrustedHostList.length) {
```

lib/msal-common/src/authority/AuthorityFactory.ts
```diff
@@ -45,8 +66,13 @@ export class AuthorityFactory {
         switch (type) {
             case AuthorityType.Aad:
                 return new AadAuthority(authorityUrl, networkInterface);
+            case AuthorityType.B2C:
+                return new B2cAuthority(authorityUrl, networkInterface);
+            // TODO: Support ADFS here in a later PR
             default:
-                throw ClientAuthError.createInvalidAuthorityTypeError(`Given Url: ${authorityUrl}`);
+                throw ClientAuthError.createInvalidAuthorityTypeError(
+                    `Given Url: ${authorityUrl}`
```
tnorling, line 74:
```suggestion
                    `${authorityUrl}`
```

tnorling: "Given Url:" Is already part of the InvalidAuthorityTypeError

lib/msal-common/src/authority/B2cAuthority.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * Returns a promise with the TenantDiscoveryEndpoint
+     */
+    public async getOpenIdConfigurationEndpointAsync(): Promise<string> {
+        if (this.isInTrustedHostList(this.canonicalAuthorityUrlComponents.HostNameAndPort)) {
+            return this.defaultOpenIdConfigurationEndpoint;
+        }
+
+        throw ClientConfigurationError.createUnsupportedAuthorityValidationError();
```
tnorling, line 33:
```suggestion
        throw ClientConfigurationError.createUntrustedAuthorityError();
```

lib/msal-common/src/authority/B2cAuthority.ts
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    /**
+     * Checks to see if the host is in a list of trusted hosts
+     * @param {string} The host to look up
+     */
+    public isInTrustedHostList(host: string): boolean {
+        if (!Object.keys(B2CTrustedHostList).length) {
```
tnorling, line 41:
```suggestion
        if (!B2CTrustedHostList.length) {
```

tnorling: You could leave this in or take it out completely. Given the logic to create B2C instance is that this is set, this will never be called unless B2C instance is explicitly created without using AuthorityFactory

lib/msal-common/src/authority/B2cAuthority.ts
```diff
@@ -0,0 +1,47 @@
...
+            throw ClientConfigurationError.createKnownAuthoritiesNotSetError();
+        }
+
+        return B2CTrustedHostList[host.toLowerCase()];
```
tnorling, line 45:
```suggestion
        return B2CTrustedHostList.includes(host.toLowerCase());
```

lib/msal-common/src/authority/AuthorityFactory.ts
```diff
@@ -30,27 +35,14 @@ export class AuthorityFactory {
         if (pathSegments[0] === "adfs") {
             return AuthorityType.Adfs;
         }
-        else if (Object.keys(B2CTrustedHostList).length) {
+        else if (!B2CTrustedHostList.length) {
```
tnorling, line 38:
```suggestion
        else if (B2CTrustedHostList.length) {
```

lib/msal-common/src/authority/AuthorityFactory.ts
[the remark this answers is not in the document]
pkanher617: What are we doing in the case of ADFS?

PR #1434, opened 2020-03-30
[msal-node][msal-common] client applications, authorization code and device code implementation

sangonzal:
Adds implementation public APIs and implementation for authorization code flow and device code flow. Will add a separate PR for unit tests- won't merge this until that is out and reviewed. 

Adds to msal-common:
- AuthorizationCodeClient - creating authorization endpoint URI and then making token request
- DeviceCodeClient - device code flow 
- BaseClient - will be extended by all oauth clients (refresh token and silent coming next)
- Configuration - configuration of for oauth clients 

response: 
- DeviceCodeResponse - response from first step of device code flow. Returned to user via callback. 
- AuthenticationResult - value returned to the user after a successful token acquisition. Have not finalized if this will be added. 

request:  
- AuthorizationCodeRequest - ClientApp.acquireTokenByAuthCode(AuthorizationCodeRequest)
- AuthorizationCodeUrlRequest - ClientApp.getAuthCodeUrl(AuthorizationCodeUrlRequest)
- DeviceCodeRequest - PublicClientApp.acquireTokenByDeviceCode(DeviceCodeRequest)
- RequestValidator - validation utils for requests 
- ParameterBuilder - util for building oauth requests
- CancellationToken - for cancelling polling in device code flow  

Add to msal-node: 
- ClientApplication - abstract base class for PublicClientApp and ConfidentialClientApp. Contains APIs shared by both. Not exported. 
- PublicClientApplication - public client flows. 
- ConfidentialClientApplication - confidential client flows.
- ClientConfiguration - configuration for PublicClientApplication and ConfidentialClientApplication. 
- HeaderUtils - Injecting node default headers into msal-common

Adds to samples: 
- msal-node-auth-code: express app showing authorization code flow
- msal-node-device-code: console app showing device code flow


lib/msal-browser/src/config/Configuration.ts
```diff
@@ -2,14 +2,14 @@
...
-import { AuthOptions, SystemOptions, LoggerOptions, INetworkModule, LogLevel } from "@azure/msal-common";
+import { SPAAuthOptions, SystemOptions, LoggerOptions, INetworkModule, LogLevel } from "@azure/msal-common";
 import { BrowserUtils } from "../utils/BrowserUtils";
 import { BrowserConstants } from "../utils/BrowserConstants";
 
 // Default timeout for popup windows in milliseconds
 const DEFAULT_POPUP_TIMEOUT_MS = 60000;
 
-export type BrowserAuthOptions = AuthOptions & {
+export type BrowserAuthOptions = SPAAuthOptions & {
```
sameerag: Are the 'AuthOptions` from the base class no longer applicable for `msal-node`?

lib/msal-common/src/config/Configuration.ts
```diff
@@ -21,13 +21,26 @@ const DEFAULT_TOKEN_RENEWAL_OFFSET_SEC = 300;
...
 };
 
+/**
+ * @type AuthOptions: Use this to configure the auth options in the Configuration object
+ *
+ *  - clientId                    - Client ID of your app registered with our Application registration portal : https://portal.azure.com/#blade/Microsoft_AAD_IAM/ActiveDirectoryMenuBlade/RegisteredAppsPreview in Microsoft Identity Platform
+ *  - authority                   - You can configure a specific authority, defaults to " " or "https://login.microsoftonline.com/common"
+ *  - redirectUri                 - The redirect URI of the application, this should be same as the value in the application registration portal.Defaults to `window.location.href`.
+ */
+export type AuthOptions = {
```
tnorling, line 39: Are we not allowing setting redirectUri anymore? navigateToLoginRequestUrl?

lib/msal-common/src/config/PublicClientSPAConfiguration.ts
```diff
@@ -5,14 +5,14 @@
...
- * @type AuthOptions: Use this to configure the auth options in the Configuration object
+ * @type SPAAuthOptions: Use this to configure the auth options in the Configuration object
  *
  *  - clientId                    - Client ID of your app registered with our Application registration portal : https://portal.azure.com/#blade/Microsoft_AAD_IAM/ActiveDirectoryMenuBlade/RegisteredAppsPreview in Microsoft Identity Platform
  *  - authority                   - You can configure a specific authority, defaults to " " or "https://login.microsoftonline.com/common"
  *  - redirectUri                 - The redirect URI of the application, this should be same as the value in the application registration portal.Defaults to `window.location.href`.
  *  - postLogoutRedirectUri       - Used to redirect the user to this location after logout. Defaults to `window.location.href`.
  */
-export type AuthOptions = {
+export type SPAAuthOptions = {
```
tnorling, line 15: postLogoutRedirectUri? It's in the block comment above but not in the options below

lib/msal-common/src/config/PublicClientSPAConfiguration.ts
```diff
@@ -5,14 +5,14 @@
...
     clientId: string;
     authority?: string;
     redirectUri?: string | (() => string);
```
pkanher617, line 18: Does Node not use redirectURI or postLogoutRedirectUri?

lib/msal-common/src/response/AuthenticationResult.ts
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Result returned from the authority's token endpoint.
+ */
+export type AuthenticationResult = {
+    // TODO this is temp class, it will be updated.
+    accessToken: string;
+    refreshToken: string;
+    idToken: string;
+    expiresOn: string;
```
tnorling, line 14: Is string the right type for this?

lib/msal-common/src/client/BaseClient.ts
```diff
@@ -79,85 +63,72 @@ export abstract class BaseClient {
...
-    /**
-     * Handles the hash fragment response from public client code request. Returns a code response used by
-     * the client to exchange for a token in acquireToken.
-     * @param hashFragment
-     */
-    abstract handleFragmentResponse(hashFragment: string): CodeResponse;
+        // TODO expensive to resolve authority endpoints every time.
+        const authority: Authority = authorityString
+            ? AuthorityFactory.createInstance(authorityString, this.networkClient)
+            : this.defaultAuthorityInstance;
```
tnorling, line 83: Isn't Authority an abstract class? I don't think you can just instantiate Authority like this. Maybe throw an error instead?

lib/msal-common/src/config/PublicClientSPAConfiguration.ts
sameerag: It does. We made a choice of making `redirectUri` a mandatory param in the `request`, hence removed it from `configuration`. `msal-node` does not need `postLogoutRedirectUri`

lib/msal-common/src/config/Configuration.ts
sameerag: t does. We made a choice of making `redirectUri a mandatory param in the `request`, hence removed it from `configuration`. msal-node does not need `navigateToLoginRequestUrl` as it doesn't need a browser

tnorling: ```redirectUri``` is also in the block comment above. So if we're removing it from the ```AuthOptions``` we should remove it from the comment as well.

lib/msal-common/src/config/PublicClientSPAConfiguration.ts
pkanher617: I think those kinds of things need to be discussed as major API changes. I don't think making a separate type for something is a good strategy if the alternative is simply including a configuration option for default. It is going to be more maintenance. Could we try to add these options back and consolidating AuthOptions and SpaAuthOptions?

lib/msal-common/src/request/AuthorizationCodeRequest.ts
```diff
@@ -0,0 +1,33 @@
...
+ * code:                    The authorization_code that the user acquired in the first leg of the flow.
+ *
+ * codeVerifier:            The same code_verifier that was used to obtain the authorization_code.
+ *                          Required if PKCE was used in the authorization code grant request.
+ *                          For more information, see the PKCE RFC: https://tools.ietf.org/html/rfc7636
+ *
+ * correlationId:           Unique GUID set per request to trace a request end-to-end for telemetry purposes
+ *
+ */
+export type AuthorizationCodeRequest = {
```
pkanher617, line 26: I would consider naming this `AuthorizationTokenRequest` instead of `AuthorizationCodeRequest`, I was a bit confused by this.

lib/msal-common/src/request/CancellationToken.ts
```diff
@@ -0,0 +1,18 @@
...
+ * Licensed under the MIT License.
+ */
+
+/**
+* Cancellation token used for cancelling the device code request. 
+* 
+* While the user authenticates on a separate device, MSAL polls the the token endpoint of security token service for the interval 
+* specified in the device code resonse (usually 15 minutes). To stop polling and cancel the request, set cancellationToken.cancel = true. 
+* */
+export class CancellationToken {
```
pkanher617, line 12: Why do we need a class for this? Couldn't the same be achieved with a boolean field in DeviceCodeRequest?

lib/msal-common/src/request/RequestValidator.ts
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * Validates server consumable params from the "request" objects
+ */
+export class RequestValidator {
+
+    /**
+     * Utility to validate scopes passed by the user in the request
+     */
+    static validateAndGenerateScopes(scopes: Array<string>, clientId: string): Array<string> {
```
pkanher617, line 19: This seems like a pointless function. You should use the ScopeSet object throughout the code until the request is made.

lib/msal-common/src/request/RequestValidator.ts
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    /**
+     * Utility to check if the `redirectUri` in the request is a non-null value
+     * @param redirectUri
+     */
+    static validateRedirectUri(redirectUri: string) : void {
+        // check if the redirectUri is null, set to default
+        if (StringUtils.isEmpty(redirectUri)) {
+            throw ClientConfigurationError.createUrlEmptyError();
```
pkanher617, line 32: I would argue this function isn't need either, but if it is being used we should have a better error message thrown here.

lib/msal-common/src/request/RequestValidator.ts
```diff
@@ -0,0 +1,82 @@
...
+    /**
+     * Utility to validate code_challenge_method
+     * @param codeChallengeMethod
+     */
+    static validateCodeChallengeMethod(codeChallengeMethod: string) : void {
+        // validate prompt parameter
+        if (
+            [
+                CodeChallengeMethodValues.PLAIN,
+                CodeChallengeMethodValues.S256
```
pkanher617, line 76: I would put this array in a constants file.

lib/msal-common/src/request/ScopeSet.ts
```diff
@@ -61,62 +77,84 @@ export class ScopeSet {
         this.appendScope(Constants.OFFLINE_ACCESS_SCOPE);
     }
 
+    /**
+     * Replace client id with the default scopes used for token acquisition.
+     */
+    private addDefaultScopes(): void {
```
pkanher617, line 83: Will the node library get an id token by default?

lib/msal-common/src/request/ScopeSet.ts
```diff
@@ -21,29 +21,45 @@ export class ScopeSet {
...
+        const scopeArr = filteredInput? StringUtils.trimAndConvertArrayEntriesToLowerCase([...filteredInput]): [];
         this.scopes = new Set<string>(scopeArr);
         if (!this.scopesRequired) {
             this.appendScope(this.clientId);
         }
         this.originalScopes = new Set<string>(this.scopes);
-        this.replaceDefaultScopes();
+
+        // add default scopes
+        nonSPA ? this.addDefaultScopes() : this.replaceDefaultScopes();
```
pkanher617, line 49: I don't think this is needed. We should revisit the changes in this class.

lib/msal-common/src/request/ScopeSet.ts
```diff
@@ -199,4 +247,15 @@ export class ScopeSet {
         }
         return "";
     }
+
+    /**
+     * Prints scopes into a space-delimited string
+     * @param scopes
+     */
+    static finalScopes(scopes: Array<string>): string {
```
pkanher617, line 255: this is a duplicate of printScopes

pkanher617: Not needed if you are using ScopeSet correctly.

lib/msal-common/src/response/AuthenticationResult.ts
pkanher617: +1. We are returning a Javascript Date object normally. I'm also not sure why node needs a different response object from browser.

lib/msal-common/src/authority/AadAuthority.ts
```diff
@@ -37,7 +37,7 @@ export class AadAuthority extends Authority {
 
         // for custom domains in AAD where we query the service for the Instance discovery
         const response = await this.networkInterface.sendGetRequestAsync<OpenIdConfigResponse>(this.aadInstanceDiscoveryEndpointUrl);
-        return response.tenant_discovery_endpoint;
+        return response.body.tenant_discovery_endpoint;
```
pkanher617, line 40: Doesn't this break fetch for msal-browser?

lib/msal-common/src/authority/Authority.ts
```diff
@@ -152,6 +153,7 @@ export abstract class Authority {
      */
     public async resolveEndpointsAsync(): Promise<void> {
         const openIdConfigEndpoint = await this.getOpenIdConfigurationEndpointAsync();
-        this.tenantDiscoveryResponse = await this.discoverEndpoints(openIdConfigEndpoint);
+        const response = await this.discoverEndpoints(openIdConfigEndpoint);
+        this.tenantDiscoveryResponse = response.body;
```
pkanher617, line 157: Same comment about fetch in msal browser

lib/msal-common/src/config/Configuration.ts
pkanher617: Either that or make NodeAuthOptions and SPAAuthOptions separate types with no base type.

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -0,0 +1,165 @@
...
+            parameterBuilder.addNonce(request.nonce);
+        }
+
+        if(request.claims) {
+            parameterBuilder.addClaims(request.claims);
+        }
+
+        return parameterBuilder.createQueryString();
+    }
+}
```
pkanher617, line 165: These should be abstracted into a separate class. In fact, I would argue that the RequestParameterBuilder class (as it is) is an unnecessary layer of abstraction. You could pull these functions into that class and get rid of the current functions that are there.

lib/msal-common/src/server/RequestParameterBuilder.ts
```diff
@@ -0,0 +1,210 @@
...
+* Copyright (c) Microsoft Corporation. All rights reserved.
+* Licensed under the MIT License.
+*/
+
+import { AADServerParamKeys, SSOTypes } from "../utils/Constants";
+import { Constants } from "../utils/Constants";
+import { ScopeSet } from "../request/ScopeSet";
+import { ClientConfigurationError } from "../error/ClientConfigurationError";
+
+export class RequestParameterBuilder {
```
pkanher617, line 11: This class as it is doesn't add any value. I think you should abstract the entire construction of the request into this class and get rid of the helpers that are here. See the above comment in `AuthorizationCodeClient`.

lib/msal-common/src/client/DeviceCodeClient.ts
```diff
@@ -0,0 +1,181 @@
...
+        parameterBuilder.addClientId(this.config.authOptions.clientId);
+
+        const scopes = RequestValidator.validateAndGenerateScopes(
+            request.scopes,
+            this.config.authOptions.clientId
+        );
+        parameterBuilder.addScopes(scopes);
+
+        return parameterBuilder.createQueryString();
+    }
```
pkanher617, line 99: You could find an abstraction for this the same way you can for AuthorizationCodeClient.

lib/msal-common/src/client/DeviceCodeClient.ts
```diff
@@ -0,0 +1,181 @@
...
+
+        const requestParameters: RequestParameterBuilder = new RequestParameterBuilder();
+        requestParameters.addScopes( scopes);
+        requestParameters.addClientId(this.config.authOptions.clientId);
+        requestParameters.addGrantType(GrantType.DEVICE_CODE_GRANT);
+        requestParameters.addDeviceCode(deviceCodeResponse.device_code);
+
+        return requestParameters.createQueryString();
+    }
+}
```
pkanher617, line 181: Same comment about abstracting this class into RequestParameterBuilder.

lib/msal-node/src/utils/NetworkUtils.ts
```diff
@@ -0,0 +1,16 @@
...
+
+import { INetworkModule } from '@azure/msal-common';
+import { HttpClient } from './../network/HttpClient';
+
+export class NetworkUtils {
+    /**
+     * Returns best compatible network client object.
+     */
+    static getNetworkClient(): INetworkModule {
+        return new HttpClient();
```
pkanher617, line 14: Do we need this util? It was added in browser because we have a possibility of fetch not being available so we fall back to Xhr. Are we going to have multiple network clients in node?

samples/msal-node-auth-code/index.js
```diff
@@ -0,0 +1,62 @@
+/*
```
pkanher617, line 1: Is this sample runnable? Add a quick readme to run it?

lib/msal-common/src/config/Configuration.ts
sangonzal: Sure, I think making NodeAuthOptions and SpaAuthOptions into separate types makes sense. Having redirectUri in the base class pollutes the public API for node, as it's only used in one (out of 6+ flows).

lib/msal-common/src/config/PublicClientSPAConfiguration.ts
sangonzal: It's on line 19.

lib/msal-common/src/config/PublicClientSPAConfiguration.ts
sangonzal: Having redirectUri in the base class pollutes the public API for node, as it's only used in one (out of 6+ flows). We can get rid of AuthOptions altogether, and have separate AuthOptions for browser and for Node

lib/msal-common/src/request/AuthorizationCodeRequest.ts
sangonzal: AuthorizationCode is the name set by Oauth2.0. When we add the API that has both steps, it'll be something along the lines of InteractiveRequest. This is aligned with .Net, Python, and Java.

lib/msal-common/src/config/PublicClientSPAConfiguration.ts
tnorling: My b. Didn't see it as github collapsed a bunch of lines :)

lib/msal-common/src/request/CancellationToken.ts
sangonzal: Sure, we can use a bool instead of this class.

lib/msal-common/src/request/ScopeSet.ts
sangonzal: Yes, needed for the cache.

lib/msal-common/src/authority/AadAuthority.ts
sangonzal: The response from fetch is wrapped in a NetworkResponse object, defined in msal-common. This is part of the throttling PR

lib/msal-node/src/utils/NetworkUtils.ts
sangonzal: Yes -  we'd like to let customers inject their own http client if they choose to.

lib/msal-common/src/authority/Authority.ts
sangonzal: The response from fetch is wrapped in a NetworkResponse object, defined in msal-common. This is part of the throttling PR

lib/msal-common/src/client/AuthorizationCodeClient.ts
sangonzal: Not sure I follow, could you please expand? 

Are you suggesting that we move  `createTokenRequestBody` and `createQueryString` into  `RequestParameterBuilder` and abstract them so that work for every single type of request (AuthCode, DeviceCode, etc)?

lib/msal-common/src/response/AuthenticationResult.ts
sangonzal: This object is a placeholder - not 100% sure if it will be added. The difference between this and TokenResponse is that we will need to add some Account info. 

Agreed that expiresOn should probably be a Date.

samples/msal-node-auth-code/index.js
sangonzal: Yep, will add a README with instructions on how to run before we release.

lib/msal-common/src/request/RequestValidator.ts
sangonzal: The values are already in constants file. Makes it easier to see what values we are checking for by having them here. Same pattern is followed in validatePrompt

lib/msal-common/src/request/RequestValidator.ts
sangonzal: Switched over to using ScopeSet

lib/msal-common/src/request/RequestValidator.ts
sangonzal: Updated the error message.

lib/msal-common/src/request/ScopeSet.ts
sangonzal: Yep, moved over to using printScopes

lib/msal-common/src/request/ScopeSet.ts
sangonzal: I've updated to get rid of nonSpa. Ultimately I think scopesRequired should be refactored as well, passing in boolean as parameter is a code smell, but I'll leave that for a separate PR.

lib/msal-browser/src/config/Configuration.ts
sangonzal: Not sure what you're asking here, could you please expand? This is browser, msal-node still use AuthOptions. Btw I haven't touched any of the configuration code, these were all changes that you made.

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -0,0 +1,163 @@
...
+        parameterBuilder.addScopes(scopeSet);
+
+        // validate the redirectUri (to be a non null value)
+        RequestValidator.validateRedirectUri(request.redirectUri);
+        parameterBuilder.addRedirectUri(request.redirectUri);
+
+        // generate the correlationId if not set by the user and add
+        const correlationId = request.correlationId
+            ? request.correlationId
+            : this.config.cryptoInterface.createNewGuid();
```
jasonnutter, line 122: `const correlationId = request.correlationId ||  this.config.cryptoInterface.createNewGuid()`

lib/msal-common/src/client/BaseClient.ts
```diff
@@ -79,85 +63,72 @@ export abstract class BaseClient {
...
-            return this.account;
-        }
-
-        // if login is not yet done, return null
-        return null;
+    createDefaultLibraryHeaders(): Map<string, string> {
+        const headers = new Map<string, string>();
+        // library version
+        headers.set(`${AADServerParamKeys.X_CLIENT_SKU}`, Constants.LIBRARY_NAME);
+        headers.set(`${AADServerParamKeys.X_CLIENT_VER}`, "0.0.1");
```
jasonnutter, line 112: Instead of hardcoding, I would pull this dynamically from the `package.json`

lib/msal-common/src/request/ScopeSet.ts
```diff
@@ -21,29 +21,43 @@ export class ScopeSet {
...
+        scopesRequired: boolean,
+    ) {
+        this.clientId = clientId;
         this.scopesRequired = scopesRequired;
+
         // Filter empty string and null/undefined array items
-        const filteredInput = inputScopes ? StringUtils.removeEmptyStringsFromArray(inputScopes) : inputScopes;
+        const filteredInput = (inputScopes)
+            ? StringUtils.removeEmptyStringsFromArray(inputScopes)
+            : inputScopes;
```
jasonnutter, line 35: `cost filteredInput = inputScopes && StringUtils.removeEmptyStringsFromArray(inputScopes)`

lib/msal-common/src/server/RequestParameterBuilder.ts
```diff
@@ -0,0 +1,208 @@
...
+    addAuthorizationCode(code: string): void {
+        this.parameters.set(`${AADServerParamKeys.CODE}`, encodeURIComponent(code));
+    }
+
+    /**
+     * add the `authorization_code` passed by the user to exchange for a token
+     * @param code
+     */
+    addDeviceCode(code: string): void {
+        this.parameters.set(`${AADServerParamKeys.DEVICE_CODE}`, encodeURIComponent(code));
```
jasonnutter, line 171: Backticks not needed.

lib/msal-common/src/server/RequestParameterBuilder.ts
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    /**
+     * Utility to create a URL from the params map
+     */
+    createQueryString(): string {
+        const queryParameterArray: Array<string> = new Array<string>();
+
+        this.parameters.forEach((value, key) => {
+            const keyValuePair = key + "=" + value;
```
jasonnutter, line 202: Can use string interpolation: 

```js
const keyValuePair = `${key}=${value}`;
```

lib/msal-node/src/cache/Storage.ts
```diff
@@ -0,0 +1,231 @@
...
+import {
+    ICacheStorage,
+    Constants,
+    PersistentCacheKeys,
+    TemporaryCacheKeys,
+} from '@azure/msal-common';
+import { CacheOptions } from '../config/ClientConfiguration';
+import { CACHE } from './../utils/Constants';
+
+const fs = require('fs');
```
jasonnutter, line 14: You should be able to use `import`

jasonnutter: And I'd recommend using the `promise` API: https://nodejs.org/api/fs.html#fs_fs_promises_api

I think this should work:
```js
import { promises as fs } from 'fs'
```

lib/msal-node/src/cache/Storage.ts
```diff
@@ -0,0 +1,231 @@
...
+
+// Cookie life calculation (hours * minutes * seconds * ms)
+const COOKIE_LIFE_MULTIPLIER = 24 * 60 * 60 * 1000;
```
jasonnutter, line 17: Can you add the units to the name? e.g. `COOKIE_LIFE_MS`

lib/msal-node/src/cache/Storage.ts
```diff
@@ -0,0 +1,231 @@
...
+        cookieValue: string,
+        expires?: number
+    ): void {
+        let cookieStr = `${cookieName}=${cookieValue};path=/;`;
+        if (expires) {
+            const expireTime = this.getCookieExpirationTime(expires);
+            cookieStr += `expires=${expireTime};`;
+        }
+
+        document.cookie = cookieStr;
```
jasonnutter, line 174: `document` doesn't exist in node, so I don't think you need the cookie logic.

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -3,51 +3,61 @@
...
-            headers: (options && options.headers)
+    async sendGetRequestAsync<T>(
+        url: string,
+        options?: NetworkRequestOptions
+    ): Promise<NetworkResponse<T>> {
+     const request: AxiosRequestConfig = {
+         method: HttpMethod.GET,
+         url: url,
+         headers: options && options.headers,
+         validateStatus: () => { return true }
```
jasonnutter, line 28: I believe you should be able to set this for all requests via the defaults object, which you can do in the constructor of this class.

lib/msal-node/tsconfig.json
```diff
@@ -10,9 +10,9 @@
     "rootDir": "./",
     "strict": true,
     "noImplicitAny": true,
-    "strictNullChecks": true,
+    "strictNullChecks": false,
```
jasonnutter, line 13: The sooner this is set to `true`, the better.

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -0,0 +1,165 @@
...
+
+    /**
+     * API to acquire a token in exchange of 'authorization_code` acquired by the user in the first leg of the authorization_code_grant
+     * @param request
+     */
+    async acquireToken(request: AuthorizationCodeRequest): Promise<string> {
+
+        const authority: Authority = await this.createAuthority(request && request.authority);
+        const response = await this.executeTokenRequest(authority, request);
+        return JSON.stringify(response.body);
```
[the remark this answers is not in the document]
sangonzal: Just returning temporary value. Upcoming PR will handle response and return an AuthenticationResult or TokenResponse object.

lib/msal-common/src/client/DeviceCodeClient.ts
```diff
@@ -0,0 +1,181 @@
...
+
+                        const response = await this.executePostToTokenEndpoint(
+                            this.authority.tokenEndpoint,
+                            requestBody,
+                            headers);
+
+                        if(response.body && response.body.error == Constants.AUTHORIZATION_PENDING){
+                            // user authorization is pending. Will sleep for polling interval and try again
+                            // TODO use logger here
+                            console.log(JSON.stringify(response.body));
```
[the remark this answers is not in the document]
sangonzal: Yep, I've left a TODO in all of the places where we are console.logging to update to using the updated Logger.

lib/msal-node/src/cache/Storage.ts
sangonzal: Sorry should have left a comment on this file (I did so on the previous PR that was accidently closed). Storage.ts not pertinent to us, I think it's here as a placeholder while Sameera works on the Cache.

lib/msal-node/src/network/HttpClient.ts
jasonnutter:
```js
axios.defaults.validateStatus = function () {
    return true;
};
```

lib/msal-common/src/client/DeviceCodeClient.ts
```diff
@@ -0,0 +1,181 @@
...
+
+        const requestBody = this.createTokenRequestBody(request, deviceCodeResponse);
+        const headers: Map<string, string> = this.createDefaultTokenRequestHeaders();
+
+        const deviceCodeExpirationTime = TimeUtils.nowSeconds() + deviceCodeResponse.expires_in;
+        const pollingIntervalMilli = deviceCodeResponse.interval * 1000;
+
+        // Poll token endpoint while (device code is not expired AND operation has not been cancelled by
+        // setting CancellationToken.cancel = true). POST request is sent at interval set by pollingIntervalMilli
+        let intervalId: ReturnType<typeof setTimeout>;
```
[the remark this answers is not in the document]
sangonzal: updated

lib/msal-common/src/client/AuthorizationCodeClient.ts
sangonzal: updated

lib/msal-node/tsconfig.json
sangonzal: Agreed. I'll probably tackle this one next.

lib/msal-node/src/network/HttpClient.ts
sangonzal: Yep, this works. Updated.

lib/msal-common/src/server/RequestParameterBuilder.ts
sangonzal: Updated

lib/msal-common/src/request/ScopeSet.ts
sangonzal: updated

lib/msal-common/src/server/RequestParameterBuilder.ts
sangonzal: Updated

PR #1435, opened 2020-03-31
Don't broadcast msal:login events from Angular Guard

### LINKED ISSUE #1392: Why guard on acquireTokenSilent broadcasts msal:loginSuccess and not just msal:acquireTokenSuccess?
## Library
- [X] `@azure/msal-angular@0.x.x`
- [X] `@azure/msal-angular@1.x.x`

## Description
In my application I need to know when user simply reloads page (by typing url into the browser, in my specific case), and when he logins. On [npm page](https://www.npmjs.com/package/@azure/msal-angular#4-subscribe-to-event-callbacks) it states that `msal:loginSuccess` should be broadcasted in case of `loginPopup()/loginRedirect`. But msal-guard.service.ts broadcasts `msal:loginSuccess` after `acquireTokenSilent` function.

In my case, user types url, and angular navigate him to the page. `msal-guard.service.ts` does his check, get token silently, broadcasts `msal:loginSuccess`, and I dont know if user just was redirected from login page, or he simply pasted url.

Why it has such behavior, and is there a way to know when user simply reloads page, and when he comes from login page?

### LINKED ISSUE #1404: angular-msal 1.0.0 with angular 8 , msal-loginSuccess broadcast event fires twice and page is not redirect back to login is called?
I implemented @azure/angular-msal 1.0.0 with msal 1.2.2 in my angular 8 application , (that uses web api of .NET Core 3.1). I'm not understanding how logout works. My application requirement is like, on page load of any page it checks if login otherwise redirect to login page. But application not having logout functionality. So when in browser new tab, from azure portal user signed out, then application page changes, shouldn't it ask for login again? and how refresh token automatically works?

jasonnutter:
Currently, if the user is logged in, the MSAL guard will call `acquireTokenSilent` and then emit `msal:loginSuccess`. This results in broadcasting two events, `msal:acquireTokenSuccess` (for the ATS call) and `msal:loginSuccess`. This is confusing and counterintuitive, so this simplifies the guard to not broadcast related events when the user is already logged in.

Fixes:
#1404 
#1392

### DIFF
--- a/lib/msal-angular/src/msal-guard.service.ts
+++ b/lib/msal-angular/src/msal-guard.service.ts
@@ -58,14 +58,8 @@ export class MsalGuard implements CanActivate {
             return this.authService.acquireTokenSilent({
                 scopes: [this.msalConfig.auth.clientId]
             })
-                .then((result: AuthResponse) => {
-                    this.broadcastService.broadcast("msal:loginSuccess",  result);
-                    return true;
-                })
-                .catch((error: AuthError) => {
-                    this.broadcastService.broadcast("msal:loginFailure", error);
-                    return false;
-                });
+                .then(() => true)
+                .catch(() => false);
         }
 
     }


tnorling (approved):

sameerag (approved):
LGTM. For my info, so we are broadcasting only for first time login calls?

jasonnutter:
> LGTM. For my info, so we are broadcasting only for first time login calls?

Correct.

llakhani:
Hi @jasonnutter  : with which version of msal , I can get this with angular 8?

llakhani:
Hi, I'm using msal 1.2.2-beta.0 version. But still on auth guard of routing (acquireTokenSilent broadcasts) login success event is also called. I'm having some functionality if user is presented with login screen then only I want to execute few lines of code. On aquireTokenSilent I don't want to run that code.

PR #1436, opened 2020-03-31
[WIP] Support for method enumerations in 2.0

jmckennon:
Looking for feedback on this the idea of this. Tests may be failing as I was having a problem with them on my machine, will work them out later.

I'm mostly concerned about how this makes `msal-common` less headless and also introduces a variable in the `state` parameter that shouldn't theoretically be there--though I'm unsure how else to solve this.

For reference: we want to be able to know if a call being made was originally an interactive or silent call, which `msal-common` does not currently support.

EDIT 4/3/2020:
This PR just needs any requested changes from comments and tests added. Should be merged before #1338


lib/msal-common/src/utils/Constants.ts
```diff
@@ -32,6 +32,15 @@ export const Constants = {
     URL_FORM_CONTENT_TYPE: "application/x-www-form-urlencoded"
 };
 
+/**
+ * Denotes which way MSAL is being used per request
+ */
+export enum MsalMethod {
```
jasonnutter: I think we should not have "Msal" in names of things. This is called "InteractionType" in v1, probably good to agree on "method" vs "interaction type" vs something else.

lib/msal-common/src/utils/ProtocolUtils.ts
```diff
@@ -34,4 +34,8 @@ export class ProtocolUtils {
         }
         return "";
     }
+
+    static setMethodInUserRequestState(userRequestState: string, method: MsalMethod): string {
+        return !StringUtils.isEmpty(userRequestState) ? `${userRequestState}${Constants.RESOURCE_DELIM}${method}` : method;
```
jasonnutter: If you wanted common to be use case agnostic, you could have this method in common take in library state (which each library will compute with its needed business logic) and user state, and just return a string that concatenates them together.

lib/msal-common/src/utils/Constants.ts
```diff
@@ -32,6 +32,15 @@ export const Constants = {
...
+    SILENT = "silent",
+    POPUP = "popup",
+    REDIRECT = "redirect"
```
jasonnutter: And then have this is browser and be part of the computed state (which we know will be a more complicated base64 encoded string)

lib/msal-common/src/utils/Constants.ts
sameerag: yes. It is something we have been following since the library initial days, no MSAL references within the library nomenclature.

lib/msal-common/src/utils/Constants.ts
sameerag: so do we want to keep this or simplify?

lib/msal-common/src/app/module/AuthorizationCodeModule.ts
```diff
@@ -237,7 +237,7 @@ export class AuthorizationCodeModule extends AuthModule {
                     refreshToken: cachedTokenItem.value.refreshToken,
                     expiresOn: new Date(expirationSec * 1000),
                     account: account,
-                    userRequestState: ""
+                    userRequestState: MsalMethod.SILENT
```
sameerag: Isn't this supposed to be set in request validation? That is coming in msal-node, not sure about the msal-browser.

lib/msal-common/src/utils/Constants.ts
jmckennon: sounds good, will change

lib/msal-common/src/app/module/AuthorizationCodeModule.ts
jmckennon: i will double check, you may be correct. will get back to this comment

lib/msal-common/src/utils/ProtocolUtils.ts
jmckennon: i see. that makes sense. more extendable too. will edit

lib/msal-common/src/utils/Constants.ts
jmckennon: @sameerag i think the idea is to keep this, but move it to browser and make the code in msal common (on jason's other comment) be more generic

PR #1440, opened 2020-03-31
Bump version and add changelog for 1.3.0-beta.0

jasonnutter:
# 1.3.0-beta.0

## Enhacements
* Turn library state into encoded string that contains guid and timestamp. (#1395)
* Fix behavior of `handleRedirectCallback`, and make it no longer required. (#1358)
* `domain_hint` is no longer supported in silent calls or when `sid` or `login_hint` is passed. (#1299)

### DIFF
--- a/lib/msal-core/changelog.md
+++ b/lib/msal-core/changelog.md
@@ -1,3 +1,10 @@
+# 1.3.0
+
+## Enhancements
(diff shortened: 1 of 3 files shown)


tnorling (approved):

sameerag (approved):
LGTM

PR #1449, opened 2020-04-02
[msal-node][msal-common] unit tests for msal-node changes

sangonzal:
Corresponding tests for #1434


lib/msal-common/src/client/BaseClient.ts
```diff
@@ -105,7 +103,7 @@ export abstract class BaseClient {
     /**
      * addLibraryData
      */
-    createDefaultLibraryHeaders(): Map<string, string> {
+    protected createDefaultLibraryHeaders(): Map<string, string> {
```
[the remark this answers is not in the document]
sangonzal: Yep, I've opened up an issue.

.gitignore
```diff
@@ -255,6 +255,9 @@ paket-files/
 .DS_Store
 .DS_Store?
 
+# Visual Studio code
+.vscode/
```
jasonnutter, line 259: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/.vscode/settings.json

I like having this in the repo because it we can use to setup things like linting, etc in the repo.

sangonzal: Ok cool I'll remove this from .gitignore

PR #1451, opened 2020-04-02
Adding Silent Iframe Flow in Authorization Code Flow

pkanher617:
This PR adds iframe handling code as well as a new API to acquire authorization codes silently in an iframe when session cookies are available on the service.


lib/msal-browser/src/app/Configuration.ts
```diff
@@ -6,8 +6,9 @@ import { AuthOptions, SystemOptions, LoggerOptions, INetworkModule, LogLevel } f
 import { BrowserUtils } from "../utils/BrowserUtils";
 import { BrowserConstants } from "../utils/BrowserConstants";
 
-// Default timeout for popup windows in milliseconds
+// Default timeout for popup windows and iframes in milliseconds
 const DEFAULT_POPUP_TIMEOUT_MS = 60000;
+const DEFAULT_IFRAME_TIMEOUT_MS = 6000;
```
pkanher617, line 11: We should revisit the values for these defaults

pkanher617:
Still need to add unit tests

lib/msal-browser/src/app/Configuration.ts
```diff
@@ -96,7 +99,9 @@ const DEFAULT_LOGGER_OPTIONS: LoggerOptions = {
 const DEFAULT_SYSTEM_OPTIONS: BrowserSystemOptions = {
     loggerOptions: DEFAULT_LOGGER_OPTIONS,
     networkClient: BrowserUtils.getBrowserNetworkClient(),
-    windowHashTimeout: DEFAULT_POPUP_TIMEOUT_MS
+    windowHashTimeout: DEFAULT_POPUP_TIMEOUT_MS,
+    iframeHashTimeout: DEFAULT_IFRAME_TIMEOUT_MS,
+    loadFrameTimeout: 0
```
jasonnutter, line 104: Could we dynamically set this to `500` if we are in IE/Edge?

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -306,6 +319,32 @@ export class PublicClientApplication {
...
+            // Create silent handler
+            const silentHandler = new SilentHandler(this.authModule, this.browserStorage, this.config.system.loadFrameTimeout);
+            // Get the frame handle for the silent request
+            const msalFrame = await silentHandler.initiateAuthRequest(navigateUrl);
+            // Monitor the window for the hash. Return the string value and close the popup when the hash is received. Default timeout is 60 seconds.
+            const hash = await silentHandler.monitorFrameForHash(msalFrame, this.config.system.iframeHashTimeout, navigateUrl);
+            // Handle response from hash string.
+            return await silentHandler.handleCodeResponse(hash);
+        } catch(e) {
+            throw e;
```
jasonnutter, line 344: I imagine we'll want logging, etc here in addition to throwing?

lib/msal-browser/src/interaction_handler/SilentHandler.ts
```diff
@@ -0,0 +1,175 @@
...
+                    this.removeHiddenIframe(iframe);
+                    clearInterval(intervalId);
+                    resolve(contentHash);
+                    return;
+                } else if (ticks > maxTicks) {
+                    // Timeout error
+                    this.removeHiddenIframe(iframe);
+                    clearInterval(intervalId);
+                    reject(BrowserAuthError.createMonitorWindowTimeoutError(urlNavigate));
+                    return;
```
jasonnutter, line 78: Nitpick: `return` isnt needed

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: I will add logging once we have a good design for it. Working on it this week.

lib/msal-browser/src/app/Configuration.ts
pkanher617: done.

lib/msal-browser/src/interaction_handler/SilentHandler.ts
pkanher617: I'd like to keep it here to make sure we are ending the function after rejecting, this has caused problems in the past.

lib/msal-common/src/error/AuthenticationRequiredError.ts
```diff
@@ -0,0 +1,44 @@
...
+    },
+    consentRequired: {
+        code: "consent_required"
+    },
+    loginRequired: {
+        code: "login_required"
+    }
+};
+
+export class AuthenticationRequiredError extends ClientAuthError {
```
pkanher617, line 23: Rename to `InteractionRequiredAuthError`

lib/msal-browser/src/interaction_handler/SilentHandler.ts
```diff
@@ -0,0 +1,175 @@
...
+     */
+    private loadFrame(urlNavigate: string, frameName: string): Promise<HTMLIFrameElement> {
+        /*
+         * This trick overcomes iframe navigation in IE
+         * IE does not load the page consistently in iframe
+         */
+
+        return new Promise((resolve, reject) => {
+            setTimeout(() => {
+                const frameHandle = this.loadFrameSync(urlNavigate, frameName);
```
sameerag, line 97: I thought the idea of `loadFrameSync` is to have the code outside of `setTimeOut()` if the user sets it to `0`

pkanher617: That happens on line 25

pkanher617: loadFrame just calls loadFrameSync internally

lib/msal-browser/src/error/BrowserAuthError.ts
```diff
@@ -48,13 +48,21 @@ export const BrowserAuthErrorMessage = {
         code: "user_cancelled",
         desc: "User cancelled the flow."
     },
-    popupWindowTimeoutError: {
-        code: "popup_window_timeout",
-        desc: "Popup window token acquisition operation failed due to timeout."
+    monitorWindowTimeoutError: {
+        code: "monitor_window_timeout",
+        desc: "Token acquisition operation failed - monitorWindowForHash timed out."
```
jasonnutter, line 53: `monitorWindowForHash` is an implementation detail and will be meaningless to consumers, in my opinion.

pkanher617: Added it here to differentiate between monitorWindowForHash vs monitorFrameForHash for MSAL devs. If you have a better error message please suggest.

jasonnutter: "Token acquisition in popup failed due to timeout."

lib/msal-common/src/error/AuthenticationRequiredError.ts
pkanher617: done.

lib/msal-browser/src/error/BrowserAuthError.ts
pkanher617: done.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -306,6 +319,32 @@ export class PublicClientApplication {
 
     // #region Silent Flow
 
+    async ssoSilent(request: AuthenticationParameters): Promise<TokenResponse> {
+        // block the reload if it occurred inside a hidden iframe
+        BrowserUtils.blockReloadInHiddenIframes();
+
+        const silentRequest: AuthenticationParameters = {
+            ...request,
+            prompt: PromptValue.NONE
```
jasonnutter, line 328: `clientId` should be set as the only scope, right?

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -306,6 +319,32 @@ export class PublicClientApplication {
 
     // #region Silent Flow
```
jasonnutter, line 321: Can you add a typedoc block here?

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: No the point of this function is to acquire a code silently, regardless of scope. That is why I believe ssoSilent isn't a great name and we should change it here.

jasonnutter: Oh, I was under the impression the only scenario where iframe would be needed for token acquisition was the SSO scenario. When else would an iframe be needed to acquire a token in Auth Code?

jasonnutter: Regardless, I would argue the APIs should be as follows:

- `acquireTokenSilent`: Uses network client for most use cases, iframe handler if needed.
- `ssoSilent`: Uses iframe handler to acquire/renew ID token (similar to v1)

jasonnutter: Point being, you call `acquireTokenSilent` whenever you need an access token, and it figures out which handler to use. You call `ssoSilent` when you want to SSO to an existing session. The fact both might use an iframe is an implementation detail and apps shouldn't be required to know the difference.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -318,31 +318,47 @@ export class PublicClientApplication {
...
+     * 
+     * To renew idToken, please pass clientId as the only scope in the Authentication Parameters.
+     * @returns {Promise.<TokenResponse>} - a promise that is fulfilled when this function has completed, or rejected if an error was raised. Returns the {@link AuthResponse} object
+     */
     async ssoSilent(request: AuthenticationParameters): Promise<TokenResponse> {
         // block the reload if it occurred inside a hidden iframe
         BrowserUtils.blockReloadInHiddenIframes();
 
+        // Check that we have some SSO data
+        if (StringUtils.isEmpty(request.loginHint) && StringUtils.isEmpty(request.sid) && !request.account) {
```
jasonnutter, line 343: @sameerag and I talked about this for #1166 and `loginHint` or `sid` should be required, even if there is an account.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -318,31 +318,47 @@ export class PublicClientApplication {
...
+            throw BrowserAuthError.createSilentSSOInsufficientInfoError();
+        }
+
+        // Check that prompt is set to none, log message if it is set to anything else.
+        if (request.prompt && request.prompt !== PromptValue.NONE) {
```
jasonnutter, line 348: This would mean `prompt` must be passed even though we hard-code it below to `NONE`, which doesn't make sense.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -361,8 +377,52 @@ export class PublicClientApplication {
...
+                    scopes: tokenRequest.scopes,
+                    resource: tokenRequest.resource,
+                    extraQueryParameters: tokenRequest.extraQueryParameters,
+                    authority: tokenRequest.authority,
+                    correlationId: tokenRequest.correlationId,
+                    prompt: PromptValue.NONE,
+                    account: tokenRequest.account,
+                    sid: tokenRequest.sid,
+                    loginHint: tokenRequest.loginHint
+                };
```
jasonnutter, line 397:
```js
const silentRequest = {
  ...tokenRequest,
  prompt: PromptValue.NONE
};
```

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -318,31 +318,47 @@ export class PublicClientApplication {
...
         BrowserUtils.blockReloadInHiddenIframes();
 
+        // Check that we have some SSO data
+        if (StringUtils.isEmpty(request.loginHint) && StringUtils.isEmpty(request.sid) && !request.account) {
+            throw BrowserAuthError.createSilentSSOInsufficientInfoError();
+        }
+
+        // Check that prompt is set to none, log message if it is set to anything else.
+        if (request.prompt && request.prompt !== PromptValue.NONE) {
+            console.log(`Prompt must be set to none for silent requests. The given value was ${request.prompt}`);
```
jasonnutter, line 349: I think this should be output to `logger.warn` instead of `console.log`.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -361,8 +377,52 @@ export class PublicClientApplication {
...
 
-        // Send request to renew token. Auth module will throw errors if token cannot be renewed.
-        return await this.authModule.getValidToken(tokenRequest);
+        try {
+            // Send request to renew token. Auth module will throw errors if token cannot be renewed.
+            return await this.authModule.getValidToken(tokenRequest);
+        } catch (e) {
+            const isServerError = e instanceof ServerError;
+            const isInvalidGrantError = (e.errorCode === BrowserConstants.INVALID_GRANT_ERROR);
+            if (isServerError && isInvalidGrantError) {
```
jasonnutter, line 386: Could there be a `invalidGrant` error that isn't a server error?

lib/msal-browser/src/app/PublicClientApplication.ts
jasonnutter: Or do you only want these specific properties?

lib/msal-browser/src/app/PublicClientApplication.ts
jasonnutter: If not, I might consider doing a `switch` statement here, in case we need to special case handle other error types in the future.

lib/msal-browser/src/app/PublicClientApplication.ts
jasonnutter: Oh nevermind, misread this also.

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: If forceRefresh exists on `tokenRequest` does it get copied to the `silentRequest`?

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: There is no logger object in the browser library currently, leaning on system console for now. I need to get Logger added soon.

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: It is possible, but I don't think it should be a case we consider. By using the class distinction we can short-circuit the check so that if an error other than a ServerError is thrown (i.e. configuration or ClientAuthError), we immediately throw rather than checking the auth code.

tnorling: It's my understanding that the ```invalid_grant``` error is equivalent to Interaction Required. Would this not be a case we would want to throw?

pkanher617: For now we will use silent handler to check if there is still a valid session. It should still throw an interaction required error. We will update this when we start receiving a bad_token error.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -305,6 +318,52 @@ export class PublicClientApplication {
...
+     * If your refresh token has expired, you can use this function to fetch a new set of tokens silently as long as 
+     * you session on the server still exists.
+     * @param {@link AuthenticationParameters} 
+     * 
+     * To renew idToken, please pass clientId as the only scope in the Authentication Parameters.
+     * @returns {Promise.<TokenResponse>} - a promise that is fulfilled when this function has completed, or rejected if an error was raised. Returns the {@link AuthResponse} object
+     */
+    async ssoSilent(request: AuthenticationParameters): Promise<TokenResponse> {
+        // block the reload if it occurred inside a hidden iframe
+        BrowserUtils.blockReloadInHiddenIframes();
```
DarylThayil, line 340: What if the app is in an iframe but is trying to do SSO silent, what will happen?

lib/msal-browser/src/app/PublicClientApplication.ts
DarylThayil: Should this resolve as an error?

DarylThayil:
This looks really well tested thanks!@

lib/msal-browser/src/app/PublicClientApplication.ts
DarylThayil: should this be an error?

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: It will still occur, this blocks reloads inside an iframe (i.e. if we have a hash and making a token call). This prevents nested iframes from occurring.

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: Never mind, I have made this change.

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: Initially I wanted to replace it for them, since calling ssoSilent would imply that they want a prompt=none. I think it may be better to throw an error here however. I have updated the code.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -318,9 +376,54 @@ export class PublicClientApplication {
...
+        try {
+            // Send request to renew token. Auth module will throw errors if token cannot be renewed.
+            return await this.authModule.getValidToken(silentRequest);
+        } catch (e) {
+            const isServerError = e instanceof ServerError;
+            const isInvalidGrantError = (e.errorCode === BrowserConstants.INVALID_GRANT_ERROR);
+            if (isServerError && isInvalidGrantError) {
+                const tokenRequest: AuthenticationParameters = {
+                    ...silentRequest,
+                    prompt: PromptValue.NONE
```
jasonnutter, line 392: Double checking from our discussion, do we need to hard-code any specific scopes here to ensure we are getting an ID token?

pkanher617: no, since we are creating a login url from the authModule, we will by default add openid and profile.

Issue #1422, opened 2020-03-26
Process Hash Removes Entire Hash From URL Instead Of Just MSAL Properties


mindingdata:
- [ ] `msal@1.x.x` or `@azure/msal@1.x.x`
- [x] `@azure/msal-browser@2.x.x`
- [ ] `@azure/msal-angular@0.x.x`
- [ ] `@azure/msal-angular@1.x.x`
- [ ] `@azure/msal-angularjs@1.x.x`

## Description

I have a particular scenario where I am using a Redirect Login, after login the user is redirected to an intermediary page where I check if we are in an IFrame (To stop the infinite loop), and if I am not, I then redirect back to my app using a URL I store in localstorage to redirect the user back to where they started with the id_token appended on the end. 

I set the `navigateToLoginRequestUrl` to false so this all works fine. 

However if I have a situation where a user may be emailed a link with an existing fragment/hash in the URL that jumps them to a specific spot on the page. For example : 

www.mysite.com/mypage#line103

If a user is sent this and they are not authenticated, I first store which page was requested in their local storage then push them to login. They will be redirected to my intermediary page after logging in looking like so : 

www.mysite.com/loginredirect.html#id_token=abc123

I then push the user back to the site they requested like so : 

www.mysite.com/mypage#line103&id_token=abc123

This logs them in fine but the main issue is that this line here : https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/37913c9129caca2e71d05a85f100263c92454c53/lib/msal-core/src/UserAgentApplication.ts#L1053

It removes various hash components of MSAL, but also deletes any other hash components. In my case not such a massive deal since it's only an auto scroll but I assume others who do more with Hash Routing may have larger issues. 

Is there a way in the future to use enum of ServerHashParamKeys and remove only the hash params that belong to MSAL and not those belonging to the app?

## Expected behavior
Would be great to not have MSAL remove hash components not belonging to the library.

mvrak:
Hello Mindingdata,

We had direction to remove hash information as the reply urls do not support hash data.  So you are describing the intended behavior.  I'm not sure we would know what is strippable and what is not strippable.

There is a very similar concern among some app developers that reply URLs do not provide better support for wildcards.  That direction all comes from our security review.

Do you have any suggestions for how to tell what "#..." information should be allowed in an URL?

Thanks

mindingdata:
Hi @mvrak, 

[...]

PR #1452, opened 2020-04-02
MSAL Guard should properly support hash routing and non-root base urls

### LINKED ISSUE #1447: MsalGuard constructs incorrect redirectStartPage when HashLocationStrategy is used
## Library
- [X] `@azure/msal-angular@1.0.0-beta.4`

## Framework
Angular v7.

## Description
PR #1343 introduced the `redirectStartPage` field in `AuthenticationParameters` and the `MsalGuard` started using it in order to preserve the Angular route being hit when redirecting. This was great!

Unfortunately this does not work when an Angular app such as ours uses `HashLocationStrategy` by setting the `useHash` field to `true` in the root router module. For completeness, this is our router module configuration:
```ts
RouterModule.forRoot(
  routes,
  {
    enableTracing: false,
    paramsInheritanceStrategy: 'always',
    useHash: true
  }
)
```

When the `MsalGuard` constructs the `routePath` on [this line](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/06e879cf0b4eb07c6d4408934dceadf08352f296/lib/msal-angular/src/msal-guard.service.ts#L41), the hash is not included.
For example, when the route is meant to be the following:
```
http://localhost:4200/#/path/to/secure/route
```
What actually gets constructed is:
```
http://localhost:4200/path/to/secure/route
```
This is because `window.location.origin` is `http://localhost:4200` and `state.url` is `/path/to/secure/route`. The `/#` in between the two is missing.

I am trying to find out what the best way is to safely construct the full URL to the current route irrespective of whether `PathLocationStrategy` or `HashLocationStrategy` is used in an Angular app. I thought I would post this issue in the meantime in case you already know the answer @jasonnutter.

## Configuration
This issue is not dependent on the MSAL configuration but is instead dependent on setting `useHash` to `true` on the root Angular routing module as above.

## Expected behavior
When using `HashLocationStrategy`, deep-linking to a route protected by the `MsalGuard` should cause the correct `redirectStartPage` to be constructed, with the hash in the URL.

jasonnutter:
This PR addresses the issues raised in #1447: 
- Properly construct the `redirectStartUrl` if hash routing is used.
- Properly construct the `redirectStartUri` if the application uses a different root url than `/`

It also fixes the behavior in `msal` when `redirectStartUri` itself contains a hash. If this occurs, `msal` will process the response hash immediately, then redirect to the `redirectStartUri`. This is to prevent both hash values from being concatenated together, which confuses both msal and the Angular router.

Unfortunately, this may lead to `redirectStartUri` not receiving the `handleRedirectCallback` response (this would happen if `redirectUri` and `redirectStartUri` are not the same base url, e.g. `/redirect` and `/profile#edit`). I wasn't able to find a graceful way to handle this scenario, given the changes that were made in #1358.

Fixes:
- #1447 
- #1422

### DIFF
--- a/lib/msal-angular/src/msal-guard.service.ts
+++ b/lib/msal-angular/src/msal-guard.service.ts
@@ -33,8 +33,9 @@ export class MsalGuard implements CanActivate {
      * @returns Full destination url
      */
     getDestinationUrl(path: string): string {
-        // Absolute base url for the application
-        const baseUrl = this.location.normalize(document.getElementsByTagName("base")[0].href);
+        // Absolute base url for the application (default to origin if base element not present)
+        const baseElements = document.getElementsByTagName("base");
+        const baseUrl = this.location.normalize(baseElements.length ? baseElements[0].href : window.location.origin);
 
         // Path of page (including hash, if using hash routing)
         const pathUrl = this.location.prepareExternalUrl(path);
--- a/lib/msal-angular/src/msal-guard.service.ts
+++ b/lib/msal-angular/src/msal-guard.service.ts
@@ -27,6 +27,28 @@ export class MsalGuard implements CanActivate {
         private broadcastService: BroadcastService
     ) {}
 
+    /**
+     * Builds the absolute url for the destination page
+     * @param path Relative path of requested page
+     * @returns Full destination url
+     */
+    getDestinationUrl(path: string): string {
+        // Absolute base url for the application
+        const baseUrl = this.location.normalize(document.getElementsByTagName("base")[0].href);
+
+        // Path of page (including hash, if using hash routing)
+        const pathUrl = this.location.prepareExternalUrl(path);
+
+        // Hash location strategy
+        if (pathUrl.startsWith("#")) {
+            return `${baseUrl}/${pathUrl}`;
+        }
+
+        // If using path location strategy, pathUrl will include the relative portion of the base path (e.g. /base/page).
+        // Since baseUrl also includes /base, can just concatentate baseUrl + path
+        return `${baseUrl}${path}`;
+    }
+
     canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): boolean | Promise<boolean> {
         this.authService.getLogger().verbose("location change event from old url to new url");
 
@@ -47,10 +69,10 @@ export class MsalGuard implements CanActivate {
                     .catch(() => false);
             }
 
-            const routePath = `${window.location.origin}${state.url}`;
+            const redirectStartPage = this.getDestinationUrl(state.url);
 
             this.authService.loginRedirect({
-                redirectStartPage: routePath,
+                redirectStartPage,
                 scopes: this.msalAngularConfig.consentScopes,
                 extraQueryParameters: this.msalAngularConfig.extraQueryParameters
             });
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -1067,7 +1067,14 @@ export class UserAgentApplication {
                 window.location.href = "/";
                 return;
             } else if (currentUrl !== loginRequestUrl) {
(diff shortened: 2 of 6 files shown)


tnorling:
Another thing to consider; In the case where a 2nd redirect is needed and a hash routing strategy is used, the 2nd redirect will throw an error due to no ```state``` being present in the hash. This is something we will want to also address in conjunction with #1413 where ```state``` is especially important in determining whether to process the hash at all

jasonnutter:
> Another thing to consider; In the case where a 2nd redirect is needed and a hash routing strategy is used, the 2nd redirect will throw an error due to no state being present in the hash

I don't think that is true here, as we are clearing the response hash, processing it, then navigating to the original url (which has the Angular specific hash values, which won't pass MSAL's checks, which look for known properties in the hash).

tnorling:
> > Another thing to consider; In the case where a 2nd redirect is needed and a hash routing strategy is used, the 2nd redirect will throw an error due to no state being present in the hash
> 
> I don't think that is true here, as we are clearing the response hash, processing it, then navigating to the original url (which has the Angular specific hash values, which won't pass MSAL's checks, which look for known properties in the hash).

I stand corrected. I didn't dig into the ```urlContainsHash``` function and assumed it was just a basic check whether or not the url contains a hash but it is in fact inspecting the hash for known properties. So this should be fine. I take back my original concern

tnorling (approved):

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1067,7 +1067,14 @@ export class UserAgentApplication {
                 window.location.href = "/";
                 return;
             } else if (currentUrl !== loginRequestUrl) {
-                window.location.href = `${loginRequestUrl}${hash}`;
+                // If loginRequestUrl contains a hash (e.g. Angular routing), process the hash now then redirect to prevent both hashes in url
+                if (loginRequestUrl.indexOf("#") > -1) {
+                    this.logger.info("loginRequestUrl contains hash, processing response hash immediately then redirecting");
+                    this.processCallBack(hash, stateInfo, null);
```
sameerag, line 1073: If there are two hashes does it always imply angular routing?

jasonnutter: No, #1422 describes another scenario with a non-msal hash value.

jasonnutter: @pkanher617 No, that was the problematic behavior that @tnorling fixed.

lib/msal-angular/src/msal-guard.service.ts
```diff
@@ -27,6 +27,28 @@ export class MsalGuard implements CanActivate {
...
     ) {}
 
+    /**
+     * Builds the absolute url for the destination page
+     * @param path Relative path of requested page
+     * @returns Full destination url
+     */
+    getDestinationUrl(path: string): string {
+        // Absolute base url for the application
+        const baseUrl = this.location.normalize(document.getElementsByTagName("base")[0].href);
```
DarylThayil, line 37: Is this element always available?

DarylThayil (approved):
asked one question but looks good

lib/msal-angular/src/msal-guard.service.ts
jasonnutter: The Angular CLI adds it be default, but yeah, there isn't a guarantee. I'll add a default if it isn't there.

jasonnutter: So Angular apps must either provide this base element, or provide the base path via the `APP_BASE_HREF` injection token. I'm not sure how to account for the `APP_BASE_HREF`, given if the Guard expects it to be injected, Angular will throw an error if it is not provided. So for now, I'll default to `window.location.origin`, and incrementally handle `APP_BASE_HREF`.

PR #1455, opened 2020-04-03
When MSAL Guard fails to silent SSO, prompt for interaction

### LINKED ISSUE #1295: @azure/msal-angular on .NET Core Project. Token Timeout prevents navigation [MsalGuard]
## Library
- [X ] `@azure/msal@1.2.2-beta.0`

## Framework
Angular 8

## Description
This is similar to [this issue](https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/1139#issuecomment-580788098). That issue caused an error on accessing the API, which has now been fixed. 

This issue: An old token allows access to resources, but renders the angular application unable to navigate `MsalGuard`-ed resources.

## Security
No

## Regression
No

## Configuration
Standard - as with the Angular 8 sample. Accessing a local API protected resource. 

```js
export const protectedResourceMap: [string, string[]][] = [
	['http://localhost:8973/api/Values', ['https://[redacted].onmicrosoft.com/api/demo.read']],
	['https://graph.microsoft.com/v1.0/me', ['user.read']]];

const isIE = window.navigator.userAgent.indexOf("MSIE ") > -1 || window.navigator.userAgent.indexOf("Trident/") > -1;


@NgModule({
	declarations: [
		AppComponent,
		HomeComponent,
		LogoutComponent,
		ValuesComponent
	],
	imports: [
		BrowserModule,
		AppRoutingModule,
		HttpClientModule,
    MsalModule.forRoot({
      auth: {
        clientId: 'XXXXXXXX-XXXX-XXX-XXXX-XXXXXXXXX',
        authority: "https://[redacted].b2clogin.com/[redacted].onmicrosoft.com/B2C_1_signupsignin1",
        validateAuthority: false,
        redirectUri: "http://localhost:8744/app/dist/app/home",
        postLogoutRedirectUri: "http://localhost:8744/app/dist/app/home",
      },
      cache: {
        storeAuthStateInCookie: isIE, // set to true for IE 11
      },
    },
      {
        popUp: false,
        consentScopes: [
          "https://[redacted].onmicrosoft.com/api/demo.read"
        ],
        unprotectedResources: ["https://www.microsoft.com/en-us/"],
        protectedResourceMap,
        extraQueryParameters: {}
      })
	],
	providers: [
		{
			provide: HTTP_INTERCEPTORS,
			useClass: MsalInterceptor,
			multi: true
		},
		{
			provide: HTTP_INTERCEPTORS,
			useClass: HttpErrorInterceptorService,
			multi: true,
		}
	],
	bootstrap: [AppComponent]
})
export class AppModule { }
```

## Reproduction steps
1. Create a MsalAngular app (in line with Angular 8 sample) with a local API to access
2. Login, call the api, then leave long enough for the token to expire
3. Call your api (optional, but I did this)
4. Try to navigate around the application.

## Expected behavior
The application bounces to an unguarded resource and is logged out, or preferably, a new token continues to allow navigation. 

NOTE: After the wait, you can access your API, but you can't navigate around the app.

## Browsers
No

jasonnutter:
When MSAL Guard sees that `getAccount` is populated, it will then attempt to silent SSO to verify the existing session. The problem is that if that fails, it just returns false, which prevent navigating to that page, instead of giving the user a chance to interactively renew their session. This ensures we prompt for interaction in this scenario.

Builds on:
- #1452 

Fixes:
- #1295

### DIFF
--- a/lib/msal-angular/src/msal-guard.service.ts
+++ b/lib/msal-angular/src/msal-guard.service.ts
@@ -29,15 +29,15 @@ export class MsalGuard implements CanActivate {
 
     /**
      * Builds the absolute url for the destination page
-     * @param path Relative path of requested page
+     * @param url Relative path of requested page
      * @returns Full destination url
      */
-    getDestinationUrl(path: string): string {
+    getDestinationUrl(url: string): string {
         // Absolute base url for the application
         const baseUrl = this.location.normalize(document.getElementsByTagName("base")[0].href);
 
         // Path of page (including hash, if using hash routing)
-        const pathUrl = this.location.prepareExternalUrl(path);
+        const pathUrl = this.location.prepareExternalUrl(url);
 
         // Hash location strategy
         if (pathUrl.startsWith("#")) {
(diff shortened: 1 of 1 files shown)


lib/msal-angular/src/msal-guard.service.ts
```diff
@@ -38,30 +83,14 @@ export class MsalGuard implements CanActivate {
...
-                .then(() => true)
-                .catch(() => false);
+            return this.loginInteractively(state.url);
         }
 
+        return this.authService.acquireTokenSilent({
+            scopes: [this.msalConfig.auth.clientId]
+        })
+            .then(() => true)
+            .catch(() => this.loginInteractively(state.url));
```
jasonnutter, line 93: This probably should check if the error requires interaction, and only invoke `loginInteractively` if so.

tnorling (approved):
LGTM

lib/msal-angular/src/msal-guard.service.ts
pkanher617: +1

pkanher617 (approved):
+1 for the comment about handling only in cases of "interaction required"

lib/msal-angular/src/msal-guard.service.ts
jasonnutter: Done.

PR #1456, opened 2020-04-03
Ensure state is decoded before it is processed

jasonnutter:
The library will encode the state value, and in the scenario were there is an error, it appears that the service will encode the state parameter again when it is sent back to the library. This means the library will need to decode it twice before it is in the right format to be parsed. Ensure that the deserialize function decodes twice, and that the parseLibraryState also decodes just in case.

### DIFF
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -1093,12 +1093,11 @@ export class UserAgentApplication {
             throw AuthError.createUnexpectedError("Hash was not parsed correctly.");
         }
         if (parameters.hasOwnProperty(ServerHashParamKeys.STATE)) {
-            const decodedState = decodeURIComponent(parameters.state);
-            const parsedState = RequestUtils.parseLibraryState(decodedState);
+            const parsedState = RequestUtils.parseLibraryState(parameters.state);
 
             stateResponse = {
                 requestType: Constants.unknown,
-                state: decodedState,
+                state: parameters.state,
                 timestamp: parsedState.ts,
                 method: parsedState.method,
                 stateMatch: false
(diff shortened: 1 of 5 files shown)


sameerag (approved):
LGTM. Was this identified in Unit tests?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1093,11 +1093,12 @@ export class UserAgentApplication {
...
         }
         if (parameters.hasOwnProperty(ServerHashParamKeys.STATE)) {
-            const parsedState = RequestUtils.parseLibraryState(parameters.state);
+            const decodedState = decodeURIComponent(parameters.state);
+            const parsedState = RequestUtils.parseLibraryState(decodedState);
 
             stateResponse = {
                 requestType: Constants.unknown,
-                state: parameters.state,
+                state: decodedState,
```
pkanher617, line 1101: Any tests we can add for this?

DarylThayil: +1 on tests otherwise lgtm

jasonnutter:
Was able to properly root cause this: we encode the state value on the way out the door, and it appears the service encodes it again when sending back to us. This results in a double encoded state value, so I updated the deserialize function to decode twice to make sure things are in the right format. Added tests.

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: Done, please re-review!

jmckennon (approved):
LGTM

pkanher617 (approved):

jasonnutter:
Hmm I seem to be having trouble reproducing this behavior now, maybe it was fixed server side.

PR #1471, opened 2020-04-07
[msal-node] Cache-2: Msal node cache serialization and helpers

sameerag:
This is the second PR for `Cache` supporting `UnifiedSchema` for `msal-node`, consists of:

- Serialization and Deserialization of all cache entities
- Helper functions to support the same
- Unit Tests added with 100% coverage


lib/msal-common/src/uCache/serialize/CacheInterface.ts
```diff
@@ -0,0 +1,95 @@
...
+     */
+    static serializeJSONBlob(data: string): string {
+        return JSON.stringify(data);
+    }
+
+    /**
+     * Parse the JSON blob in memory and deserialize the content
+     * @param cachedJson
+     */
+    static deserializeJSONBlob(jsonFile?: any): JSONContent {
```
jasonnutter, line 31: Seems like this should be `JSONContent`, not `any`?

lib/msal-common/src/uCache/serialize/CacheInterface.ts
```diff
@@ -0,0 +1,95 @@
...
+            idTokens: {},
+            refreshTokens: {},
+            accounts: {},
+            appMetadata: {}
+        };
+
+        const ParsedJSON = jsonFile ? jsonFile : null;
+        if (!ParsedJSON) {
+            return jsonContent;
+        }
```
jasonnutter, line 43:
```js
if (!jsonFile) {
  return jsonContent;
}
```

jasonnutter: Also, this seems weird to return that default object?

lib/msal-common/src/uCache/serialize/Deserializer.ts
```diff
@@ -0,0 +1,116 @@
...
+            const mappedAT = CacheHelper.renameKeys(
+                accessTokens[key],
+                AccessTokenCacheMaps.fromCacheMap
+            );
+            const accessToken: AccessTokenEntity = new AccessTokenEntity();
+            accessToken.toObject(mappedAT);
+            atObjects[key] = accessToken;
+        });
+
+        return atObjects;
```
jasonnutter, line 39: This can also be done with `reduce`:

```js
return Object.keys(accessTokens).reduce((accessTokens, key) => {
   const mappedAT = CacheHelper.renameKeys(accessTokens[key], AccessTokenCacheMaps.fromCacheMap);
   const accessToken = new AccessTokenEntity();
   accessToken.toObject(mappedAT);
   return {
      ...accessTokens,
      [key]: accessToken
   };
}, {});

Comment (jasonnutter) [reply lib/msal-common/src/uCache/serialize/Deserializer.ts:39]: The methods below also can use `reduce`.

File: lib/msal-common/src/uCache/serialize/Serializable.ts [hunk 4]
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+export class Serializable {
+    toObject(obj: any) {
```
jasonnutter, line 7: This would potentially benefit from generics, which would preferably over `any`.

lib/msal-common/src/uCache/serialize/Serializer.ts
```diff
@@ -0,0 +1,100 @@
...
+import { CacheHelper } from "../utils/CacheHelper";
+import {
+    AccessTokenCacheMaps,
+    IdTokenCacheMaps,
+    RefreshTokenCacheMaps,
+    AccountCacheMaps,
+    AppMetadataCacheMaps
+} from "./JsonKeys";
+
+export class Serializer {
```
jasonnutter, line 20: This seems like an overly generic class name for what the class is doing, is there a more specific name?

lib/msal-common/src/uCache/serialize/Serializer.ts
```diff
@@ -0,0 +1,100 @@
...
+    static serializeIdTokenCacheEntity(idToken: IdTokenEntity) {
+        const mappedIdToken: string = CacheHelper.renameKeys(
+            idToken,
+            IdTokenCacheMaps.toCacheMap
+        );
+        const idTokenKey: string = idToken.generateIdTokenEntityKey();
+
+        const idTokenCacheEntity: any = {};
+        idTokenCacheEntity[idTokenKey] = mappedIdToken;
+        return idTokenCacheEntity;
```
jasonnutter, line 50: Would prefer not to see `any` (hopefully will forbid this via the tsc compiler or lint).

```js
return {
  [idTokenKey]: mappedIdToken
}
```

lib/msal-common/src/uCache/serialize/Serializer.ts
```diff
@@ -0,0 +1,100 @@
...
+        const atCacheEntity: any = {};
+        atCacheEntity[atKey] = mappedAT;
+        return atCacheEntity;
+    }
+
+    /**
+     * generate an id token CacheEntity
+     * @param idToken
+     */
+    static serializeIdTokenCacheEntity(idToken: IdTokenEntity) {
```
jasonnutter, line 41: This needs a return type. If not clear at runtime, consider generics.

lib/msal-common/src/uCache/serialize/CacheInterface.ts
jasonnutter: Not sure I see the value of that, and I would prefer to not return default-style objects (which are somewhat meaningless, in my opinion).  e.g. returning `null`

jasonnutter:
General comment: avoid `any` if possible. If the type of something is dynamic based on the use case, that is a good indicator to look at use generics so that you get type safety.

jasonnutter:
Also, not a fan of the `uCache` folder name. I assume this is unified cache? I would argue it should be named as such if so.

lib/msal-common/src/uCache/serialize/CacheInterface.ts
sameerag: This is mostly glue code to append to in memory list, I had the same objection. probably a better idea is to invoke this API only if the json is read from external memory. Will do the needed changes.

lib/msal-common/src/uCache/serialize/Serializable.ts
sameerag: yep. And I believe will also help with the above file. Will make the changes.

lib/msal-common/src/uCache/serialize/Serializer.ts
sameerag: yep. I added types wherever I can, will revisit this.

sameerag:
I agree with not having. `any` references in the code. I was experimenting with different data types and making changes to remove non-typed references. Also adding generics.

lib/msal-common/src/uCache/serialize/Serializable.ts
sameerag: I had couple of implementations to arrive here, I now think it makes sense to make this a utility instead of a base class and works well with generics. Checked in the modified code.

lib/msal-common/src/uCache/serialize/Serializer.ts
sameerag: Cool. I didn't know we can do that. Modified now.

lib/msal-common/src/uCache/serialize/Deserializer.ts
sameerag: I tried reduce and ran into some issues. Can check offline with you and change it. Is there a preference between map and reduce in general?

lib/msal-common/src/unifiedCache/serialize/JsonKeys.ts
```diff
@@ -0,0 +1,93 @@
...
+    };
+
+    static fromCacheMap = CacheHelper.swap(IdTokenCacheMaps.toCacheMap);
+}
+
+/**
+ * Map for RefreshTokenCache entity serialization and deserialization
+ */
+export class RefreshTokenCacheMaps {
+    static toCacheMap = {
```
pkanher617, line 51: not sure I like having static variables. Should make these constants.

lib/msal-common/src/unifiedCache/serialize/JsonKeys.ts
```diff
@@ -0,0 +1,93 @@
+/*
```
pkanher617, line 1: This file is weird to me. not sure why we are making classes with static vars.

lib/msal-common/src/unifiedCache/serialize/Deserializer.ts
```diff
@@ -0,0 +1,116 @@
...
+} from "../serialize/JsonKeys";
+import { AccessTokenCache, IdTokenCache, RefreshTokenCache, AccountCache, AppMetadataCache, StringDict } from "../../utils/MsalTypes";
+
+// TODO: Can we write this with Generics?
+export class Deserializer {
+    /**
+     * Deserializes access tokens to AccessTokenEntity objects
+     * @param accessTokens
+     */
+    static deSerializeAccessTokens(accessTokens: StringDict): AccessTokenCache {
```
pkanher617, line 27: nitpick: deserialize, not deSerialize

sameerag:
I have simplified/optimized code for Serialize, Deserialize classes. @pkanher617 for your point, could be done with constants, I will handle that in a later PR.

lib/msal-common/src/unifiedCache/serialize/JsonKeys.ts
sameerag: Cannot call functions, in this space specifically `swap` cannot be called unless it is `static`. Ideas welcome for a more efficient way to achieve this mapping.

lib/msal-common/src/uCache/serialize/CacheInterface.ts
sameerag: added types and changed this up a bit.

lib/msal-common/src/uCache/serialize/CacheInterface.ts
sameerag: Done

lib/msal-common/src/uCache/serialize/Serializable.ts
sameerag: Got rid of the base class and added a helper.

lib/msal-common/src/uCache/serialize/Deserializer.ts
sameerag: Sticking to map.

lib/msal-common/src/uCache/serialize/Serializer.ts
sameerag: This class no longer exists

lib/msal-common/src/uCache/serialize/Serializer.ts
sameerag: Done.

lib/msal-common/src/uCache/serialize/Serializer.ts
sameerag: Done.

lib/msal-common/src/unifiedCache/serialize/Deserializer.ts
sameerag: done

lib/msal-common/src/unifiedCache/serialize/JsonKeys.ts
sameerag: Above comment addresses this!

PR #1472, opened 2020-04-07
Update FAQ.md

hamiltonha:


hamiltonha:
Not sure if this is exactly the right spot but think we need to document the server error somewhere here. @hpsin @pkanher617

hpsin:
> Not sure if this is exactly the right spot but think we need to document the server error somewhere here. @hpsin @pkanher617

Yes, that's a good call out.  Will add to the error section in the protocol part.  Need to get the specifics from Pavel.

pkanher617 (approved):

tnorling (approved):

jotatoledo:
If an explicit example for the configuration or a link to a howto would be added, it would be great!

PR #1478, opened 2020-04-08
[msal-node] Switch strictNullChecks: true for msal-node

sangonzal:
Set strictNullChecks = true for MsalNode. Using null assertion operator, as we ensure that all configuration values are assigned default values when we build the configuration, so we can assume that they will not be null.

### DIFF
--- a/lib/msal-node/src/client/ClientApplication.ts
+++ b/lib/msal-node/src/client/ClientApplication.ts
@@ -18,13 +18,6 @@ export abstract class ClientApplication {
     // Input configuration by developer/user
     protected config: ClientConfiguration;
 
-    // Crypto interface implementation
-    protected crypto: CryptoProvider;
-
-    // Storage interface implementation
-    protected storage: Storage;
-
-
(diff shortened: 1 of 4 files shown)


lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -94,15 +77,19 @@ export abstract class ClientApplication {
     }
 
     protected buildOauthClientConfiguration(): Configuration {
+        // using null assertion operator as we ensure that all config values have default values in buildConfiguration()
         return {
             authOptions: this.config.auth,
```
tnorling, line 82: Forgot one?

sangonzal: Don't need one here as we are not accessing any of the properties in `this.config.auth`. 

If we we accessing`this.config.auth.clientId` for example, we would then have to assert that `this.config.auth` is not null by using `this.config.auth!.clientId`.

PR #1479, opened 2020-04-08
[msal-core] Fix circular dependency for B2C Authority

tnorling:
- Moves B2CTrustedHostList to B2cAuthority class to fix circular dependency with AuthorityFactory

### DIFF
--- a/lib/msal-core/src/authority/B2cAuthority.ts
+++ b/lib/msal-core/src/authority/B2cAuthority.ts
@@ -6,7 +6,8 @@
 import { Authority } from "./Authority";
 import { AuthorityType } from "./Authority";
 import { ClientConfigurationError } from "../error/ClientConfigurationError";
-import { AuthorityFactory, B2CTrustedHostList } from "./AuthorityFactory";
+
+export const B2CTrustedHostList: object = {};
 
 /**
  * @hidden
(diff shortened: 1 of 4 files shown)


lib/msal-core/src/authority/B2cAuthority.ts
```diff
@@ -6,7 +6,8 @@
 import { Authority } from "./Authority";
 import { AuthorityType } from "./Authority";
 import { ClientConfigurationError } from "../error/ClientConfigurationError";
-import { AuthorityFactory, B2CTrustedHostList } from "./AuthorityFactory";
+
+export const B2CTrustedHostList: object = {};
```
sameerag, line 10: So we want this to be a list or an array?? How are we parsing user i/p? You may have mentioned it, but I want to make sure we don't get type compat issues later.

tnorling: So in core I made this a dictionary to remain consistent with the AADTrustedHostList and to allow for flexibility to combine the AAD and B2C Authorities if that becomes the direction in the future. In msal-common this will be a list. The user facing ```knownAuthorities``` is a list in both core and common and setKnownAuthorities will populate the trusted host list slightly differently in both to account for the differences in type

tnorling: I agree this is a little confusing and I dont think dict is really the right type for either AAD or B2C but I think consistency is more important. If we want to change this for core we can discuss offline and open a separate PR

PR #1482, opened 2020-04-09
[msal-node][msal-common] Update generation of client info headers

sangonzal:
Moving the generation of the client info headers from msal-common into msal-node. If these values are not passed in from msal-node (or from whoever is calling into msal-common), then we fall back to generating them in common. 

Client info headers: 
- client-sku
- client-version
- client-os
- client-cpu


lib/msal-common/src/utils/Constants.ts
```diff
@@ -4,7 +4,8 @@
  */
 
 export const Constants = {
-    LIBRARY_NAME: "MSAL.JS",
+    LIBRARY_NAME: "MSAL.JS", // TODO JS should be broken up into common, browser, node. What dore core/angular use?
+    SKU: "MSAL.JS.common",
```
sangonzal, line 8: What should this SKU be? SKU should be different for msal-node, msal-browser, and msal-common. 

For msal-node it's "msal.node"

lib/msal-node/tsconfig.json
```diff
@@ -1,30 +1,31 @@
 {
```
sangonzal, line 1: Only relevant change here is the addition of  "resolveJsonModule": true. (I also fixed the indentation, and thus the big diff)

lib/msal-common/src/config/Configuration.ts
```diff
@@ -72,18 +76,26 @@ export type LoggerOptions = {
     logLevel?: LogLevel
 };
 
+/**
+ * Telemetry info about library
+ */
+export type ClientInfo = {
```
sameerag, line 82: Client_Info is used for homeAccountId generation as we get `client_info` as a param in ServerResponse. I would recommend changing this name.

lib/msal-common/src/config/Configuration.ts
```diff
@@ -153,23 +162,38 @@ const DEFAULT_CRYPTO_IMPLEMENTATION: ICrypto = {
     }
 };
 
+const DEFAULT_CLIENT_INFO: ClientInfo = {
```
sameerag, line 165: Same comment as above on the name

lib/msal-common/src/utils/Constants.ts
sameerag: msal.common should not have a SKU IMO, only implementations (node and browser) should update the SKU.

jasonnutter: > msal.common should not have a SKU IMO, only implementations (node and browser) should update the SKU.

Agreed.

sangonzal: @sameerag @jasonnutter why wouldn't msal-common have a SKU? We're making msal-common available to be consumed directly, why wouldn't we want to be able to track who is using it directly? 

When we are running telemetry queries, I think it would be valuable to see how many people are using msal-node, msal-browswer, or just using msal-common directly. SKU has no other effect than letting us split this data out, not sure why we wouldn't just add it.

jasonnutter: I think if you are using common, you should have to pass in a SKU to identify yourself uniquely.

sangonzal: The problem is that I'm pretty sure 95%+ of people wouldn't pass that value in, and so we would never get that data. I think when these headers were originally added they were meant to be used internally only anyways.  

We also have applicationName and applicationVersion in TelemetryOptions which serves the purpose letting people pass in this info if so they desire. IMO MSAL SKU shouldn't be optional.

lib/msal-common/src/config/Configuration.ts
sangonzal: Updated to LibraryInfo

lib/msal-common/src/utils/Constants.ts
jasonnutter: I agree that SKU should be required when using common.

sameerag: Well, okay. Only concern I have is will we able to trace all `msal-common` data to `msal-browser` vs `msal-ndode`? All service requests originate from `msal-common`, so are we overwriting the SKU in that case? Will this be triggered only if someone manages to write their wrapper and calls into `msal-common`?

jasonnutter: Requiring any library that is using `msal-common` to provide a SKU would specifically allow us to be able to differentiate between different libraries, if I'm understanding your concern. If requests have `msal-common` as the SKU, we may not be able to directly know which library actually generated the request.

pkanher617: I agree `msal-common` doesn't need it's own SKU, and we can make it a required config param for `msal-common`. We should be careful with this value because the service uses it to make decisions on which errors to throw back. We need to check with the service to see how the behavior will differ based on which SKU value is given back.

lib/msal-common/src/config/Configuration.ts
```diff
@@ -153,23 +162,38 @@ const DEFAULT_CRYPTO_IMPLEMENTATION: ICrypto = {
     }
 };
 
+const DEFAULT_LIBRARY_INFO: LibraryInfo = {
```
pkanher617, line 165: Remove the defaults here, move them to `msal-node` and `msal-browser`.

lib/msal-common/src/config/Configuration.ts
```diff
@@ -26,7 +29,8 @@ export type Configuration = {
     loggerOptions?: LoggerOptions,
     storageInterface?: ICacheStorage,
     networkInterface?: INetworkModule,
-    cryptoInterface?: ICrypto
+    cryptoInterface?: ICrypto,
+    libraryInfo?: LibraryInfo
```
pkanher617, line 33: Make this parameter mandatory.

lib/msal-common/src/utils/Constants.ts
sangonzal: Discussed in design meeting, agreed that it makes sense to have SKU for msal-common, and default to this SKU if none is passed in. Resolving.

PR #1485, opened 2020-04-09
Update README.md

hamiltonha:
warn against b2c use

### DIFF
--- a/lib/msal-browser/README.md
+++ b/lib/msal-browser/README.md
@@ -22,6 +22,9 @@
 
 ## About
 
+>[!NOTE]
+>MSAL.js 2.0 with Authorization Code Flow **is not yet available for B2C tennant** (comming soon) 
+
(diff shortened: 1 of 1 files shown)


lib/msal-browser/README.md
```diff
@@ -22,6 +22,9 @@
 
 ## About
 
+>[!NOTE]
+>MSAL.js 2.0 with Authorization Code Flow **is not yet available for B2C tennant** (comming soon)
```
jasonnutter, line 26: This syntax won't work on Github, unfortunately.  I would maybe do something like this:

```md
**Important:** MSAL.js 2.0 with Authorization Code Flow is not yet available for B2C tenants (coming soon).
```

pkanher617:
Could we also update the prerequisites section and remove the "Treat application as public client" instructions?

tnorling (approved):

jasonnutter (approved):

Issue #1469, opened 2020-04-06
`@azure/msal-common` is not exporting all type declarations


unindented:
>> Please follow the issue template below. Failure to do so will result in a delay in answering your question.

## Library
- [ ] `msal@1.x.x` or `@azure/msal@1.x.x`
- [x] `@azure/msal-browser@2.0.0-beta.0`  + `@azure/msal-common@1.0.0-beta.0`
- [ ] `@azure/msal-angular@0.x.x`
- [ ] `@azure/msal-angular@1.x.x`
- [ ] `@azure/msal-angularjs@1.x.x`

**Important**: Please fill in your exact version number above, e.g. `msal@1.1.3`.

## Framework
If you are using a framework, please provide the framework and version (e.g. Angular v8, React v16, etc).

No framework

## Description
A clear and concise description of the bug. If you receive specific error codes, please include them.

I'm trying to use `@azure/msal-browser`, but I can't import the `Logger` class from `@azure/msal-common`.

## Security
Is this issue security related?

No.

## Regression
Did this behavior work before?

Yes, `msal@1.2.0` exported that class and type correctly.

## Configuration
Please provide your MSAL configuration options.

Not relevant.

## Reproduction steps
1. Try to import `Logger` from either `@azure/msal-browser` or `@azure/msal-common`.
2. Fail.

## Expected behavior

I expected that class to be exported.

## Browsers
Is this issue browser-specific? If so, please detail which browsers are impacted (e.g. IE 11, Safari).

No.

unindented:
Ah, it looks like there could be a problem with the `tsconfig` settings (`rootDir` maybe?). It generated all the types under `dist/src`:

<img width="190" alt="Screen Shot 2020-04-06 at 4 34 24 PM" src="https://user-images.githubusercontent.com/69485/78614830-97410d00-7824-11ea-90d5-1bda20886f7b.png">

tnorling:
@unindented A fix for this was merged and will be included in the next release of msal-browser/msal-common

unindented:
Cool, thank you!

PR #1486, opened 2020-04-09
Export Logger class from common and browser packages

pkanher617:
This PR is a fix for the issue raised in #1469. 

This adds the Logger object to the exported objects in the `index.ts` file.


lib/msal-browser/src/index.ts
```diff
@@ -12,5 +12,7 @@ export {
     // Error
     AuthError,
     AuthErrorMessage,
-    INetworkModule
+    INetworkModule,
+    // Logger Object
+    Logger
```
jasonnutter, line 17: I think `LogLevel` should also be exported, in case someone wants to write logic based on that for each message.

pkanher617: Done

sameerag:
@sangonzal We should port this to msal-node-base I believe?

PR #1490, opened 2020-04-10
handleRedirectCallback Updates for MSAL 2.0

pkanher617:
This PR ports the changes made in msal-core to the new version of the library in #1358. 

- Moved processing of the hash into constructor
- handleRedirectCallback is no longer required. If called it will just await the promise saved during handleRedirectResponse step in constructor

navigateToLoginRequestUrl changes are already part of this codebase.


lib/msal-browser/package.json
```diff
@@ -40,7 +40,6 @@
         "doc": "npm run doc:generate && npm run doc:deploy",
         "doc:generate": "typedoc --mode modules --excludePrivate --excludeProtected --out ./ref ./src/ --gitRevision dev",
         "doc:deploy": "gh-pages -d ref -a -e ref/msal-browser",
-        "pretest": "npm link @azure/msal-common",
```
sameerag, line 43: Why did we remove this?

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -141,7 +143,7 @@ export class PublicClientApplication {
             } else {
                 BrowserUtils.navigateWindow(loginRequestUrl, true);
```
tnorling, line 144: Will this trigger a refresh if you are already on the loginRequestUrl?

tnorling:
Are we also planning to port the state changes Jason made in core? As those affect these pieces of code as well.

lib/msal-browser/package.json
pkanher617: Not needed anymore since lerna bootstrap should take care of linking

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: Thanks for the callout. I have added a check for this the same way as in #1358.

tnorling: We also want to make sure the hash is processed if the currentUrl is the loginRequestUrl. Currently will return null.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -319,6 +306,22 @@ export class PublicClientApplication {
...
+     * If your refresh token has expired, you can use this function to fetch a new set of tokens silently as long as 
+     * you session on the server still exists.
+     * @param {@link AuthenticationParameters} 
+     * 
+     * To renew idToken, please pass clientId as the only scope in the Authentication Parameters.
+     * @returns {Promise.<TokenResponse>} - a promise that is fulfilled when this function has completed, or rejected if an error was raised. Returns the {@link AuthResponse} object
+     */
     async ssoSilent(request: AuthenticationParameters): Promise<TokenResponse> {
         // block the reload if it occurred inside a hidden iframe
         BrowserUtils.blockReloadInHiddenIframes();
```
sameerag, line 327: Why not add `preFlightRequest()` call here for uniformity? Even if interaction is false, it is in a conditional, so we should be good.

PR #1496, opened 2020-04-14
[msal-node][msal-commmon] Add support for acquiring tokens with a refresh token to msal-node

sangonzal:
Adds support for using a refresh token to get a new batch of tokens from AAD. 

Acquire token by refresh token will be used for:
 
- Migrating applications using ADAL to MSAL (using refresh token the first time, then acquire token silent thereafter)
- Internally by acquire token silent to refresh tokens.


lib/msal-common/src/request/RefreshTokenRequest.ts
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * @type RefreshTokenRequest
+ *
+ * scopes:                  A space-separated array of scopes for the same resource.
+ * authority:               URL of the authority, the security token service (STS) from which MSAL will acquire tokens.
+ * refreshToken:            A refresh token returned from a previous request to the Identity provider.
+ */
+export class RefreshTokenRequest {
+    scopes: Array<string>;
+    refreshToken: string;
```
pkanher617, line 15: I thought the refreshToken wasn't going to be returned to users?

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -0,0 +1,56 @@
...
+import {Authority, NetworkResponse} from "..";
+import {ServerAuthorizationTokenResponse} from "../server/ServerAuthorizationTokenResponse";
+import {RequestParameterBuilder} from "../server/RequestParameterBuilder";
+import {ScopeSet} from "../request/ScopeSet";
+import {GrantType} from "../utils/Constants";
+
+/**
+ * OAuth2.0 refresh token client
+ */
+export class RefreshTokenClient extends BaseClient {
```
pkanher617, line 18: Why are we making a separate client for this? Shouldn't token renewal just be another API on the DeviceCodeClient or AuthCodeClient?

lib/msal-common/src/request/RefreshTokenRequest.ts
sangonzal: Will not be returned. But people using ADAL will have refresh tokens, and this will be used for migrating them to MSAL. (They will only use this API once, and then cache will be populated, so they can continue using acquire token silent thereafter). This is documented in the comment above the method in the code, and will also be part of some migration document that will have to be made available to people.

lib/msal-common/src/client/RefreshTokenClient.ts
pkanher617: I actually think this may be a good idea, but if we are doing this I'd like to see it extended to all flows (including the browser flow). We should also try to take more advantage of subclassing and protected functions.

lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -75,6 +76,19 @@ export abstract class ClientApplication {
...
 
+    /**
+     * Acquires a token by exchanging the refresh token provided for a new set of tokens.
+     *
+     * This API is provided only for scenarios where you would like to migrate from ADAL to MSAL. Instead, it is
+     * recommended that you use acquireTokenSilent() for silent scenarios. When using acquireTokenSilent, MSAL will
+     * handle the caching and refreshing of tokens automatically.
+     * @param request
+     */
+    async acquireTokenByRefreshToken(request: RefreshTokenRequest): Promise<string>{
```
pkanher617, line 87: Shouldn't this be acquireTokenSilent?

lib/msal-common/src/client/RefreshTokenClient.ts
sangonzal: Refresh token it's own Oauth2.0 grant, separate from the others, and therefore it makes sense to have it's own client.

lib/msal-node/src/client/ClientApplication.ts
sangonzal: acquireTokenSilent is different than acquireTokenByRefreshToken. 

acquireTokenSilent will check the cache, and 
1) finds an AT, return it
2) find an expired AT by valid RT, do acquireTokenByRefreshToken
3) if forceRefresh = true, cache is skipped and acquireTokenByRefreshToken

You can think of acquireTokenSilent as wrapping acquireTokenByRefreshToken. acquireTokenByRefreshToken is provided separately for migration scenarios

lib/msal-common/src/client/RefreshTokenClient.ts
pkanher617: Agreed but that's not how we have designed up until now. As I said, it is a good idea since refresh_token grant is pretty much the same across all libs but we should apply it to the SPA lib as well.

lib/msal-node/src/client/ClientApplication.ts
pkanher617: So historically this has been wrapped in the acquireTokenSilent functionality. basically an application will call acquireTokenSilent and acquireTokenByRefreshToken will happen without the caller calling it explicitly. I haven't found any precedence in other MSAL libs for a separate API for this either, but if it is something Node specific I may be missing something.

sangonzal: The three libraries that node is closest too (.NET, Java, Python) all implement it. We can chat about this at the next design meeting. 

For reference: 
 - [.NET](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/442227e1899ba45511840cbb1f59cfbd571b1546/src/client/Microsoft.Identity.Client/ConfidentialClientApplication.cs#L139)
- [Java](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/442227e1899ba45511840cbb1f59cfbd571b1546/src/client/Microsoft.Identity.Client/ConfidentialClientApplication.cs#L139) 
- [Python](https://docs.microsoft.com/en-us/azure/active-directory/develop/migrate-python-adal-msal#migrate-existing-refresh-tokens-for-msal-python)

Note that: 
- Python doesn't expose this in their public API as people can just call directly into internal APIs, due to how Python access modifiers (or lack of) work.

samples/msal-node-refresh-token/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "msal-node-refresh-token",
+  "version": "1.0.0",
+  "description": "Command line app showing how to use a refresh token to acquire new tokens from Azure AD",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
```
jasonnutter, line 7: I would add a start script:

```json
{
  "scripts: {
      "start": "node ./index.js"
   }
}
```

lib/msal-node/test/client/PublicClientApplication.spec.ts
```diff
@@ -71,6 +75,27 @@ describe('PublicClientApplication', () => {
...
+            }
+        };
+
+        const request: RefreshTokenRequest = {
+            scopes: TEST_CONSTANTS.DEFAULT_GRAPH_SCOPE,
+            refreshToken: TEST_CONSTANTS.REFRESH_TOKEN
+        };
+
+        const authApp = new PublicClientApplication(msalConfig);
+        authApp.acquireTokenByRefreshToken(request)
```
jasonnutter, line 92: You need to return this call, so that a promise is returned from the test. I'm guessing this test is ending before the assertions on line 95 are actually hit.

lib/msal-node/test/client/PublicClientApplication.spec.ts
```diff
@@ -1,8 +1,12 @@
...
 import {AuthorizationCodeRequest, ClientConfiguration} from './../../src/index';
-import { DeviceCodeRequest } from "@azure/msal-common/dist/src/request/DeviceCodeRequest";
-import {AuthorizationCodeClient, AuthorizationCodeUrlRequest, DeviceCodeClient} from "@azure/msal-common";
 import {AUTHENTICATION_RESULT, TEST_CONSTANTS} from "../utils/TestConstants";
+import {
+    AuthorizationCodeClient,
+    AuthorizationCodeUrlRequest,
+    DeviceCodeClient,
+    DeviceCodeRequest,
+    RefreshTokenRequest} from "@azure/msal-common";
```
jasonnutter, line 9: Can put the "from" part on the next line.

lib/msal-common/test/client/RefreshTokenClient.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+    AUTHENTICATION_RESULT,
+    DEFAULT_OPENID_CONFIG_RESPONSE,
+    TEST_CONFIG,
+    TEST_TOKENS,
+} from "../utils/StringConstants";
+import {BaseClient} from "../../src/client/BaseClient";
+import {AADServerParamKeys, GrantType} from "../../src/utils/Constants";
+import {ClientTestUtils} from "./ClientTestUtils";
+
+const expect = chai.expect;
```
jasonnutter, line 20: I believe this can be `import chai, { expect } from "chai";`

lib/msal-common/test/client/RefreshTokenClient.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+chai.use(chaiAsPromised);
```
jasonnutter, line 21: Can you put this in a central test setup file that is run for all tests?

samples/msal-node-refresh-token/package.json
sangonzal: Added

lib/msal-node/test/client/PublicClientApplication.spec.ts
sangonzal: You are right, the promises were not being executed. I updated the tests.

lib/msal-node/test/client/PublicClientApplication.spec.ts
sangonzal: Done

lib/msal-common/test/client/RefreshTokenClient.spec.ts
sangonzal: Updated

lib/msal-common/test/client/RefreshTokenClient.spec.ts
sangonzal: Split it into it's own PR: https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/1527

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+import { Configuration } from "../config/Configuration";
+import {BaseClient} from "./BaseClient";
+import {RefreshTokenRequest} from "../request/RefreshTokenRequest";
+import {Authority, NetworkResponse} from "..";
+import {ServerAuthorizationTokenResponse} from "../server/ServerAuthorizationTokenResponse";
+import {RequestParameterBuilder} from "../server/RequestParameterBuilder";
+import {ScopeSet} from "../request/ScopeSet";
+import {GrantType} from "../utils/Constants";
```
sameerag, line 13: nit: Can we maintain the spaces in import statements: `import { BaseClient } ... ;` instead of `import {BaseClient} ... ;`

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    public async acquireToken(request: RefreshTokenRequest): Promise<string>{
+        const authority = await this.createAuthority(request.authority);
+        const response = await this.executeTokenRequest(request, authority);
+        // TODO add response_handler here to send the response
+        return JSON.stringify(response.body);
+    }
+
+    private async executeTokenRequest(request: RefreshTokenRequest, authority: Authority)
+        : Promise<NetworkResponse<ServerAuthorizationTokenResponse>> {
```
sameerag, line 32: Update: Can we have the mapping changes added with this PR?

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -0,0 +1,56 @@
...
+    private createTokenRequestBody(request: RefreshTokenRequest): string {
+        const parameterBuilder = new RequestParameterBuilder();
+
+        parameterBuilder.addClientId(this.config.authOptions.clientId);
+
+        const scopeSet = new ScopeSet(request.scopes || [],
+            this.config.authOptions.clientId,
+            true);
+        parameterBuilder.addScopes(scopeSet);
+
```
sameerag, line 49: Do we need line space?

lib/msal-common/src/client/RefreshTokenClient.ts
sangonzal: @sameerag What mapping changes are you referring to?

lib/msal-common/src/client/RefreshTokenClient.ts
sangonzal: Updated.

lib/msal-common/src/client/RefreshTokenClient.ts
pkanher617: I think it is ok to have those changes in a separate PR.

PR #1499, opened 2020-04-15
Export WindowUtils, UrlUtils, and redirect callback types for MSAL Angular

jasonnutter:
Currently, MSAL Angular reaches into the package of msal to get `UrlUtils`, `WindowUtils`, and others. Instead, we should be importing them from the public API. This removes warnings from Angular that advise that importing like that can be problematic.


tnorling (approved):
LGTM

sangonzal (approved):
:shipit:

Issue #1475, opened 2020-04-08
MsalInterceptor renews token but Authorization header is missing from intercepted call


jukvaa:
## Library
- [x] `msal@1.2.2`
- [x] `@azure/msal-angular@1.0.0-beta.4`
- [x] `@angular@9.0.1`

## Framework
Angular 9

## Description
We have angular front and are using @azure/msal-angular@1.0.0-beta.4. After an hour after login when token expires, next api calls fails, because it has null in Authorization header:

`Authorization: Bearer null`

Renewal itself seems to work because api calls after this failed one work ok. Shouldn't MsalInterceptor renew token and apply it to the intercepted request?

This is msal log from the renewal:

```
main-es2015.f4c8c8ef197a06fb7d31.js:1 MSAL:  Tue, 07 Apr 2020 08:11:43 GMT:1.2.2-Info renewidToken is called
main-es2015.f4c8c8ef197a06fb7d31.js:1 MSAL:  Tue, 07 Apr 2020 08:11:43 GMT:1.2.2-Info Add msal frame to document:msalIdTokenFrame
main-es2015.f4c8c8ef197a06fb7d31.js:1 MSAL:  Tue, 07 Apr 2020 08:11:43 GMT:1.2.2-Info LoadFrame: msalIdTokenFrame
main-es2015.f4c8c8ef197a06fb7d31.js:1 MSAL:  Tue, 07 Apr 2020 08:11:44 GMT:1.2.2-Info Add msal frame to document:msalIdTokenFrame
main-es2015.f4c8c8ef197a06fb7d31.js:1 MSAL:  Tue, 07 Apr 2020 08:11:44 GMT:1.2.2-Info Returned from redirect url
main-es2015.f4c8c8ef197a06fb7d31.js:1 MSAL:  Tue, 07 Apr 2020 08:11:44 GMT:1.2.2-Info Processing the callback from redirect response
main-es2015.f4c8c8ef197a06fb7d31.js:1 MSAL:  Tue, 07 Apr 2020 08:11:44 GMT:1.2.2-Info State status:true; Request type:LOGIN
main-es2015.f4c8c8ef197a06fb7d31.js:1 MSAL:  Tue, 07 Apr 2020 08:11:44 GMT:1.2.2-Info State is right
main-es2015.f4c8c8ef197a06fb7d31.js:1 MSAL:  Tue, 07 Apr 2020 08:11:44 GMT:1.2.2-Info Fragment has id token
[...]

PR #1512, opened 2020-04-16
Mapped correct property to take token from

oleksii-kostin:
https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/1475


jasonnutter:
@oleksii-kostin The MSAL Interceptor is specifically built to retrieve the access token, not the id token (which should not be used to protect APIs). We're aware that it will sometimes retrieve the id token instead, and will have a fix soon (but this is not the correct fix, which will be to ensure that an access token is retrieved every time).

jasonnutter:
Closing, will have a proper fix for this soon.

PR #1514, opened 2020-04-17
Remove .finally from discoverendpoint promise chain as it isnt available in all promise polyfills

jasonnutter:
Not all promise polyfills (such as the one that ships with Create React App, which is used in the React sample) come with a `.finally`, so removing it to prevent breakages.


DarylThayil:
good catch lgtm

DarylThayil:
I wonder if we can add a lint rule to catch this

Issue #1505, opened 2020-04-15
Infinite redirects after updating to MSAL 1.2.2


anton-gorbikov:
## Library
- [x] `msal@1.2.2`

**Important**: Please fill in your exact version number above, e.g. `msal@1.1.3`.

## Framework
Angular

## Description
After updating to version 1.2.2 from version 1.2.1 acquiring token via redirects flow is broken (this flow is the only possible for Safari/iOS).

## Security
No

## Regression
Yes

I've investigated further and found that this bug can be related to this changes:
https://github.com/AzureAD/microsoft-authentication-library-for-js/commit/d492c3569b69525ca6e820472ddd7b68554e23a6

I see that after each redirect this code deletes `msal.acquireTokenAccount|...` key from the store which causes further problems.

## Configuration
Please provide your MSAL configuration options.

```js
{
	auth: {
		clientId: '{{my-app-id}}',
		authority: 'https://login.microsoftonline.com/{{tenant-id}}.onmicrosoft.com',
		redirectUri: `{{application-login-page}}`,
		navigateToLoginRequestUrl: false
	},
	cache: {
		cacheLocation: 'localStorage',
		storeAuthStateInCookie: true
	},
		system: {
			loadFrameTimeout: 15000
		}
	}
}
```

## Reproduction steps
1. Call `acquireTokenRedirect`
2. Receive token (I can see it in the hash)
3. Try to call `acquireTokenSiliently` to get value from cache
4. Token is not in cache. Acquire token silent fails in browsers which disallow third-party cookies
5. Fallback to `acquireTokenRedirect`, getting infinite loop.

## Expected behavior
A token should be returned after redirects.

## Browsers
Safari or any other browsers with `block third-party cookies` flag enabled.

tnorling:
@anton-gorbikov Can you first try upgrading to ```msal@1.3.0-beta.1```? We made some changes to the way msal handles redirects and this update fixed redirect loops other users were experiencing. 

If this doesn't fix your issue, can you clarify in step 2 are you saying that the hash is not being stripped from the URL?

sergey-tihon:
We've tried `msal@1.3.0-beta.1` and the issues is still there.

It looks like an issue in new `resetTempCacheItems` function.
- First `acquireTokenRedirect` works correctly, authenticate current user and set `msal|AquireTokenAccount` in local storage
[...]

Issue #1509, opened 2020-04-16
acquireTokenPopup does not save access token in local cache


barisbikmaz:
#607  Library
- `msal@1.2.2` 

## Framework
-  `React v16`

## Description
Seems like UserAgentApplication.acquireTokenPopup does not save the accessToken in local cache. Is this desired behaviour?

When I call acquireTokenPopup and call directly afterwards acquireTokenSilent with the same scopes a network call happens.

## Regression
Did this behavior work before? Don't know

## Configuration
```js
{
	auth: {
		clientId,
		authority: 'https://login.microsoftonline.com/organizations',
		redirectUri,
	},
};
```

## Reproduction steps
1.  call acquireTokenPopup with some scopes which were not granted
2.  call after success acquireTokenSilent with the same scopes
3.  Observe network call to the auth endpoint

## Expected behavior
acquireTokenPopup should save the acquired access token in the local cache.

riccardos77:
I have encountered the same problem. From a first analysis the problem seems to be in the updateCacheEntries method when invoked with the loginStartPage argument. In this case, setAccountCache is not invoked and, without putting the account in the cacheStorage, the saveTokenFromHash method does not persist the new token.
By moving the invocation of the setAccountCache method out of the else I get the correct behavior

tnorling:
@barisbikmaz @riccardos77 Thanks for bringing this to our attention. I was able to recreate this on my end. I will discuss with the team and see if we can get a fix out soon.

PR #1516, opened 2020-04-17
[msal-core] Fix token caching for acquireToken calls

tnorling:
Using loginStartPage as the condition for setting the account cache prevented access tokens from being cached in interactive flows since loginStartPage is always set. This PR changes the conditional check to more accurately differentiate between login calls and access token calls.

This PR fixes issues: 
#1505 
#1509


jasonnutter:
Looks good, thanks for leading the investigation!

PR #1520, opened 2020-04-20
[msal-node] Cache-4: ResponseHandler added to `msal-common`

sameerag:
This PR adds `ResponseHandler` to `msal-common` with the below functionality:

- Parses the `response` from the server
- Adds `account` and `credentials` (`access_token, id_token, refresh_token`) to cache if not present


lib/msal-common/src/response/AuthenticationResult.ts
```diff
@@ -17,6 +17,7 @@ export class AuthenticationResult {
     tokenType: string; // TODO: get rid of this if we can
     idToken: string;
     idTokenClaims: StringDict;
+    refreshToken: string;
```
sameerag, line 20: This is temporary and will change soon as we should not send the `refreshToken` to the client. I will merge a new PR defining an `external response` and an internal handler where we need this data to be cached.

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -45,13 +52,38 @@ export class AuthorizationCodeClient extends BaseClient {
...
+        return tokenResponse;
+    }
+
+     /**
+     * Set the cache post acquireToken call
+     */
+    private setCache() {
+        const inMemCache = this.unifiedCacheManager.getCacheInMemory();
+        const cache = this.unifiedCacheManager.generateJsonCache(inMemCache);
+        this.cacheStorage.setSerializedCache(Serializer.serializeJSONBlob(cache));
```
sangonzal, line 86: I think all of this should be in BaseClient, as it will be shared among all clients.

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -14,14 +14,21 @@ import { Configuration } from "../config/Configuration";
...
  * Oauth2.0 Authorization Code client
  */
 export class AuthorizationCodeClient extends BaseClient {
 
+    // config
+    private clientConfig: Configuration;
+
     constructor(configuration: Configuration) {
         super(configuration);
+        this.clientConfig = configuration;
```
sangonzal, line 31: This is already in BaseClient, you should be able to access it from here.

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -0,0 +1,300 @@
...
+        if (serverResponseHash.client_info) {
+            buildClientInfo(serverResponseHash.client_info, cryptoObj);
+        }
+    }
+
+    /**
+     * Function which validates server authorization token response.
+     * @param serverResponse
+     */
+    validateServerAuthorizationTokenResponse(
```
sangonzal, line 79: I think think this should `validateTokenResponse`, as it applies to token calls for all clients.

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -0,0 +1,300 @@
...
+import { Authority } from "../authority/Authority";
+import { AuthorityType } from "../authority/AuthorityType";
+
+/**
+ * Class that handles response parsing.
+ */
+export class ResponseHandler {
+    private clientId: string;
+    private uCacheManager: UnifiedCacheManager;
+    private cryptoObj: ICrypto;
```
sangonzal, line 30: nit: I think we could be more precise with naming. Maybe cryptoUtils?

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -0,0 +1,300 @@
...
+            environment,
+            idToken.claims.tid
+        );
+        if (!cachedAccount) {
+            accountEntity = this.generateAccountEntity(
+                serverTokenResponse,
+                idToken,
+                authority
+            );
+            this.uCacheManager.addAccountEntity(accountEntity);
```
sangonzal, line 221: accountEntity is not initialized if there is a cachedAccount. Is this expected behavior?

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -0,0 +1,300 @@
...
+            idToken: null,
+            idTokenClaims: null,
+            accessToken: "",
+            refreshToken: "",
+            scopes: [],
+            expiresOn: null,
+            familyId: null,
+        };
+
+        // IdToken
```
sangonzal, line 154: nit: Unnecessary comment here (and a couple of other places)

lib/msal-common/src/client/AuthorizationCodeClient.ts
sameerag: Yep, changed.

lib/msal-common/src/response/AuthenticationResult.ts
sameerag: Removed `refreshToken` now.

lib/msal-common/src/client/AuthorizationCodeClient.ts
sameerag: Done

lib/msal-common/src/response/ResponseHandler.ts
sameerag: Done.

lib/msal-common/src/response/ResponseHandler.ts
sameerag: no. Will add the logic in `silent` flow support.

lib/msal-common/src/response/ResponseHandler.ts
sameerag: deleted some. Kept some. Lets rehaul comments later.

lib/msal-common/src/response/ResponseHandler.ts
sameerag: Will remember this.

PR #1522, opened 2020-04-20
[msal-node] Cache-5: Msal node cache additional changes for end-to-end cache to work

sameerag:
This is the final PR in the series of cache PRs and works end-to-end, reads a `json` file from the disk, retrieves tokens from the server and appends the newly minted ones to the `json` file in the disk.

Notable changes:
* `NodeStorage` update for read/write from the cache
* Sample update
* Some tests

_**Note:**_ Most of the diff for this PR is `linter` diff; will be trying to minimize this soon by updating the `tsdx` lint rules to not cause so much disruption for line length which accounts to most changes.


lib/msal-common/src/cache/ICacheStorage.ts
```diff
@@ -8,12 +8,14 @@
  */
 export interface ICacheStorage {
     /**
-     * Function to read serialized cache from disk
+     * Function to read serialized Cache from disk
+     * @param key
+     * @param value
```
sangonzal, line 13: nit: getSerializedCache() doesn't have parameters.

lib/msal-common/src/unifiedCache/serialize/Deserializer.ts
```diff
@@ -9,26 +9,13 @@ import { AccessTokenEntity } from "../entities/AccessTokenEntity";
...
-import {
-    AccountCache,
-    IdTokenCache,
-    AccessTokenCache,
-    RefreshTokenCache,
-    AppMetadataCache,
-    InMemoryCache,
-    JsonCache,
-} from "../utils/CacheTypes";
+import { AccountCacheMaps, IdTokenCacheMaps, AccessTokenCacheMaps, RefreshTokenCacheMaps, AppMetadataCacheMaps } from "../serialize/JsonKeys";
```
sangonzal, line 12: Let's agree on a format for imports. I think in most other places we are importing each class in a different line.

lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -16,10 +16,14 @@ import {
 } from '../config/ClientConfiguration';
 import { CryptoProvider } from '../crypto/CryptoProvider';
 import { NodeStorage } from '../cache/NodeStorage';
+import { NodeCacheManager } from '../cache/NodeCacheManager';
 
 export abstract class ClientApplication {
     // Input configuration by developer/user
     protected config: ClientConfiguration;
+    protected nodeStorage: NodeStorage;
```
sangonzal, line 24: We should consider renaming these to get rid of "Node". Maybe cacheStorage would be more precise.

lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -16,10 +16,14 @@ import {
...
+    protected nodeCacheManager: NodeCacheManager;
+    protected cachePath: string;
```
sangonzal, line 26: These are not being used anywhere here.

lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -84,6 +89,20 @@ export abstract class ClientApplication {
...
+     * API to retrieve application cache
+     */
+    public getCache(): string {
+        return this.nodeStorage.getSerializedCache();
+    }
+
+    /**
+     * API to write cache to a file
+     */
+    public setCache(): void {
```
sangonzal, line 102: Is this temporary? I think we shouldn't be providing an implementation of setCache(). We should provide an interface and the developer can provide the implementation.

lib/msal-common/src/cache/ICacheStorage.ts
sameerag: Intentional. It is a getter to avoid access to private variables.

lib/msal-common/src/unifiedCache/serialize/Deserializer.ts
sameerag: yep. Will be in a different line.

lib/msal-node/src/client/ClientApplication.ts
sameerag: Done

lib/msal-node/src/client/ClientApplication.ts
sameerag: Well yes, until we finalize our extension library interface.

lib/msal-node/src/client/ClientApplication.ts
sameerag: Right! Good catch. Updated the code.

lib/msal-common/src/cache/ICacheStorage.ts
pkanher617: We should remove the comments here.

PR #1525, opened 2020-04-20
Switch to file-based sourcemaps for msal.js and msal.min.js

jasonnutter:
Currently, we only generate source maps for non-minified CDN build, which is done inline, greatly increasing the size of `msal.js` for the CDN. This changes source maps to be a file and generates them for both the minified and non-minified CDN builds.

Devtools docs: https://webpack.js.org/configuration/devtool/
Uglify docs: https://webpack.js.org/plugins/uglifyjs-webpack-plugin/#sourcemap

Before:

Annotation 2020-04-20 140432

After:

Annotation 2020-04-20 140600

These source map files will be pulled in when you open the devtools to debug an application using the CDN version of MSAL.

### DIFF
--- a/lib/msal-core/webpack.config.js
+++ b/lib/msal-core/webpack.config.js
@@ -26,7 +26,7 @@ module.exports = {
     extensions: ['.ts', '.tsx', '.js']
   },
 
-  devtool: 'inline-source-map',
+  devtool: 'source-map',
   plugins: [
       new ForkTsCheckerWebpackPlugin(),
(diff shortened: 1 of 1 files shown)


tnorling (approved):

DarylThayil (approved):

sameerag (approved):
lgtm. There was some history on removing source maps vs cdn uploads. @pkanher617 do you remember?

pkanher617 (approved):

PR #1528, opened 2020-04-21
Ensure interceptor uses ID token property if response is of type id_token

### LINKED ISSUE #1475: MsalInterceptor renews token but Authorization header is missing from intercepted call

jasonnutter:
Since it is technically possible to use an ID token to protect a route, the interceptor needs to properly differentiate between access tokens and id tokens and use the correct properties. 

Currently, in situations where an ID token is used to protect a route, the interceptor will attach a `Bearer null` header to outgoing requests, because it naively assumes that the token will always be present on the `accessToken` property. This is because when we retrieve a cached ID token, the response sets the ID token as `accessToken`, however, we retrieving a new ID token from the network, the ID token is only set as `idToken`.

This PR ensures we inspect the `tokenType` value first, and properly use the correct field instead of relying on `accessToken` always being set.

Fixes:
- #1475

### DIFF
--- a/lib/msal-angular/src/msal.interceptor.ts
+++ b/lib/msal-angular/src/msal.interceptor.ts
@@ -12,7 +12,7 @@ import { mergeMap, tap } from 'rxjs/operators';
 
 import {MsalService} from "./msal.service";
 import { BroadcastService } from "./broadcast.service";
-import { AuthResponse } from 'msal';
+import { AuthResponse, ServerHashParamKeys } from 'msal';
 
 @Injectable()
 export class MsalInterceptor implements HttpInterceptor {
@@ -27,14 +27,14 @@ export class MsalInterceptor implements HttpInterceptor {
             return next.handle(req);
         }
 
-        let accessToken: string;
+        let token: string;
 
         // Acquire a token for this request, and attach as proper auth header.
         return from(
             this.auth.acquireTokenSilent({ scopes })
                 .then((response: AuthResponse) => {
-                    accessToken = response.accessToken;
-                    const authHeader = `Bearer ${response.accessToken}`;
+                    token = response.tokenType === ServerHashParamKeys.ID_TOKEN ? response.idToken.rawIdToken : response.accessToken;
+                    const authHeader = `Bearer ${token}`;
                     return req.clone({
                         setHeaders: {
                             Authorization: authHeader,
@@ -48,7 +48,7 @@ export class MsalInterceptor implements HttpInterceptor {
                 event => {},
                 err => {
                     if (err instanceof HttpErrorResponse && err.status === 401) {
-                        this.auth.clearCacheForScope(accessToken);
+                        this.auth.clearCacheForScope(token);
                         this.broadcastService.broadcast('msal:notAuthorized', err.message);
                     }
                 }
--- a/lib/msal-angular/karma.conf.js
+++ b/lib/msal-angular/karma.conf.js
@@ -29,6 +29,7 @@ module.exports = function (config) {
 
             { pattern: 'node_modules/babel-polyfill/browser.js', instrument: false},
             { pattern: 'tests/MSALSpec.ts', watched: false },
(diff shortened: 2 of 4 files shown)


tnorling (approved):

lib/msal-angular/src/msal.interceptor.ts
```diff
@@ -27,14 +27,14 @@ export class MsalInterceptor implements HttpInterceptor {
...
-        let accessToken: string;
+        let token: string;
 
         // Acquire a token for this request, and attach as proper auth header.
         return from(
             this.auth.acquireTokenSilent({ scopes })
                 .then((response: AuthResponse) => {
-                    accessToken = response.accessToken;
-                    const authHeader = `Bearer ${response.accessToken}`;
+                    token = response.tokenType === ServerHashParamKeys.ID_TOKEN ? response.idToken.rawIdToken : response.accessToken;
```
DarylThayil, line 36: Can you add a test for this ?

jasonnutter: Done.

DarylThayil:
Thanks for adding tests, approved

DarylThayil (approved):

Issue #1411, opened 2020-03-24
Null reference exception at ServerRequestParameters.ts in populateQueryParams


JasonPaape:
## Library
- [x] `msal@1.2.2-beta.3`

## Framework
Found while using Angular 9 with @azure/msal-angular 1.0.0-beta.4 but that's irrelevant, the bug is in msal-core.

## Description
There is a null reference bug at msal-core/src/ServerRequestParameters.ts in function 'populateQueryParams' at line 116:
        // sanity check for developer passed extraQueryParameters
        const eQParams: StringDict = request.extraQueryParameters;
Where 'request' can be null. There is a check above on line 88 'if (request)' but line 116 is outside of that.
This happens when the 'populateQueryParams' is called from msal-core/src/UserAgentApplication.ts 
in function 'acquireTokenSilent' at line 644:
        serverAuthenticationRequest.populateQueryParams(account, null, adalIdTokenObject);
Note the second param passed is null.
The code in ServerRequestParameters.ts line 116 should have a null check or be moved into the above 'if (request)' block. 

## Security
No

## Regression
Unknown

## Configuration
Irrelevant.

## Reproduction steps
1. Call acquireTokenSilent with AuthenticationParameters containing values set to flow into this block on line 640 of UserAgentApplication.ts:
      // if user didn't pass login_hint/sid and adal's idtoken is present, extract the login_hint from the adalIdToken
      else if (!account && !StringUtils.isEmpty(adalIdToken)) {
2. See that null gets passed on line 644
3. And that in line 116 of ServerRequestParameters.ts a null exception occurs causing a runtime script exception visible in the console.

## Expected behavior
Expect no script errors, and for aquireTokenSilent to function correctly.

## Browsers
Not browser specific.

Porges:
This is the commit which introduces the bug: https://github.com/AzureAD/microsoft-authentication-library-for-js/commit/bf86b898d120760e4a1c62126becd088deb78b55#diff-01b40b5f69b6021ac7031ce85ef0d452L120-L123

jasonnutter:
@Porges It is not getting typechecked because `strictNullChecks` is disabled (which we are aware of and plan to fix, #1097).

I'll take a look at that PR, thanks!

tnorling:
@JasonPaape This was included in `msal@1.3.0` which was just released. Let us know if this is still an issue. Thanks!

Issue #1254, opened 2020-01-29
TypeError when acquiring token for scopes


ranjanmicrosoft:
>> Please follow the issue template below. Failure to do so will result in a delay in answering your question.

## Library
- [x] `msal@1.2.1`
- [x] `@azure/msal-angular@1.0.0-beta.1`

## Description
Received the following error when trying to get token for graph API.

Error when acquiring token for scopes: user.read TypeError: Cannot read property 'extraQueryParameters' of null

## Regression

I haven't successfully used MSAL yet. 

## Configuration

        MsalModule.forRoot({
            auth: {
                clientId: this.config.clientId,
                authority: 'https://login.microsoftonline.com/' + this.config.tenantId,
                redirectUri: window.location.origin,
                validateAuthority : true,
                postLogoutRedirectUri: window.location.origin,
                navigateToLoginRequestUrl : true
            },
            cache: {
                cacheLocation : 'sessionStorage',
                storeAuthStateInCookie: false
            },
            system: {
                logger: new Logger(
                    // tslint:disable-next-line:no-console
                    (level: LogLevel, error: string) => { console.log('MSAL LOGGING: ' + level + ' ' + error); } , 
                    {
                        level: LogLevel.Verbose,
                        piiLoggingEnabled: false,
                        correlationId: '1234'
                    })
            },
            framework: {
                isAngular: true,
                protectedResourceMap: new Map([['https://graph.microsoft.com/v1.0/me', ['user.read']]])
            }
        }, {
            consentScopes: [ 'user.read', 'user_impersonation' ],
            extraQueryParameters: null,
            popUp: true // todo: configure redirect
        }),

## Expected behavior

Current behavior is as such from verbose logs: 
MSAL LOGGING: 3 Wed, 29 Jan 2020 18:03:54 GMT:1234-1.2.1-Verbose Url: https://graph.microsoft.com/v1.0/me/photo/$value maps to scopes: user.read
MSAL LOGGING: 3 Wed, 29 Jan 2020 18:03:54 GMT:1234-1.2.1-Verbose ADAL's idToken exists. Extracting login information from ADAL's idToken 
MSAL LOGGING: 0 Wed, 29 Jan 2020 18:03:54 GMT:1234-1.2.1-Error Error when acquiring token for scopes: user.read TypeError: Cannot read property 'extraQueryParameters' of null

[...]

Issue #1237, opened 2020-01-23
TypeError: Cannot read property 'extraQueryParameters' of null


davallen:
>> Please follow the issue template below. Failure to do so will result in a delay in answering your question.

## Library
- [X] `msal@1.2.0`

## Framework
React v16.8.4

## Description
We are calling acquireTokenSilent with only scopes in AuthenticationParameters and receiving the "TypeError: Cannot read property 'extraQueryParameters' of null" error.

davallen:
More information on what we are passing into the AuthenticationParameters. We are stumped on how the AuthenticationParameters object is null when we are passing the Authenticationparameters object in the acquireSilentToken call. 

If we have the userName we provide it in the AuthenticationParameters object and it is successful. 

tokenRequest = {
                scopes: scopes,
                loginHint: userName
            };

let token = await userAgentApplication.acquireTokenSilent(tokenRequest);

But if we dont have the userName we provide the following and it appears to be failing quite often.

 tokenRequest = {
                scopes: scopes,
                redirectUri: window.location.origin + '/msalCallback.html'
            };
let token = await userAgentApplication.acquireTokenSilent(tokenRequest).catch(async error => {

sameerag:
@davallen what happens if you just pass `scopes`? I am curious if there is an overwrite somewhere, we check this the first thing in our `request` APIs and did not encounter this before. 

Can you also try with chrome dev tools to check what is the value passed to the msal in the debug window? I suspect a local runtime issue. 

One other option is to share us a sample app with steps of repro, if your debug isn't helping you.

tnorling:
Closing due to inactivity. Please let us know if this is still an issue

NortonIce:
> Closing due to inactivity. Please let us know if this is still an issue

I have similar problem(

jasonnutter:
@NortonIce We have another issue already open for this bug: #1411

PR #1531, opened 2020-04-22
Check if request is null in populateQueryParams

Porges:
Restore a null test removed in https://github.com/AzureAD/microsoft-authentication-library-for-js/commit/bf86b898d120760e4a1c62126becd088deb78b55#diff-01b40b5f69b6021ac7031ce85ef0d452L120-L123

This addresses #1411, #1237, and #1254. I don’t know why some of these type errors are passing compilation as they should be breaking it?

Untested, edited code in browser window.

There should probably be some tests added to CI for this as the last version that works seems to be 1.1.3.

### DIFF
--- a/lib/msal-core/src/ServerRequestParameters.ts
+++ b/lib/msal-core/src/ServerRequestParameters.ts
@@ -113,7 +113,7 @@ export class ServerRequestParameters {
         queryParameters = this.addHintParameters(account, queryParameters);
 
         // sanity check for developer passed extraQueryParameters
-        const eQParams: StringDict = request ? request.extraQueryParameters : null;
+        const eQParams: StringDict|null = request?.extraQueryParameters;
 
         // Populate the extraQueryParameters to be sent to the server
         this.queryParameters = ServerRequestParameters.generateQueryParametersString(queryParameters);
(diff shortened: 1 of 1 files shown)


lib/msal-core/src/ServerRequestParameters.ts
```diff
@@ -113,7 +113,7 @@ export class ServerRequestParameters {
         queryParameters = this.addHintParameters(account, queryParameters);
 
         // sanity check for developer passed extraQueryParameters
-        const eQParams: StringDict = request.extraQueryParameters;
+        const eQParams: StringDict|null = request?.extraQueryParameters;
```
jasonnutter, line 116: This could be a one line fix, correct? Something like:

```js
const eQParams: StringDict = request && request.extraQueryParameters;
```

We definitely want to turn on strict null checks and fix the types, but isn't need to fix this bug, correct?

Porges: Correct.

Porges: Looks like the TS version here is too old for `?.`.

jasonnutter:
Tested this out and it looks good. If you could add a couple tests in `ServerRequestParameters.spec.ts` that prove the bug and the fix, that would be appreciated!

jasonnutter:
Thanks for the fix!

PR #1533, opened 2020-04-22
[msal-common][msal-browser] Add support for sub error codes

### LINKED ISSUE #1348: msal-common supports sub-error codes
Previous issue: #903. We found this was not applicable to msal-core and the implicit flow, but will be implemented in msal-common and the auth-code flow. These are detailed further on #903. They should be passed through to users of the msal client with errors when applicable.

The .NET team has [an implementation](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1254/files) that can be used as prior art.

This is the mapping of [errors from ESTS](https://msazure.visualstudio.com/One/_git/ESTS-Main?path=%2Fsrc%2FProduct%2FESTSFE%2FServiceExceptionMapping.config&version=GBmaster). You can ctrl+f to find relevant ones.

To reproduce example suberrors, use the conditional access [labs accounts](aka.ms/msidlabs). For example, these can reproduce a `consent_required` error that comes with a sub-error.

tnorling:
This PR builds on #1451 and will close #1348 

It adds support for sub-error codes returned from the /token endpoint indicating that interaction is required


lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -372,8 +372,9 @@ export class PublicClientApplication {
             return await this.authModule.getValidToken(silentRequest);
         } catch (e) {
             const isServerError = e instanceof ServerError;
+            const isInteractionRequiredError = e instanceof InteractionRequiredAuthError;
```
jasonnutter, line 375: I think we should be consistent in this logic, given there is a static method on `InteractionRequiredAuthError` that also answers this question.

lib/msal-common/src/error/InteractionRequiredAuthError.ts
```diff
@@ -10,41 +10,48 @@ import { ServerError } from "./ServerError";
...
+
+        return "";
+    }
 
-        const isInteractionRequiredErrorCode = !StringUtils.isEmpty(errorCode) && interactionRequiredCodes.indexOf(errorCode) > -1;
-        const isInteractionRequiredErrorDesc = !StringUtils.isEmpty(errorString) && interactionRequiredCodes.some((irErrorCode) => {
+    static isInteractionRequiredError(errorCode: string, errorString: string, subError?: string) : boolean {
+        const isInteractionRequiredErrorCode = !StringUtils.isEmpty(errorCode) && Object.values(InteractionRequiredAuthErrorMessage).includes(errorCode);
+        const isInteractionRequiredSubError = !StringUtils.isEmpty(subError) && Object.values(InteractionRequiredAuthSubErrorMessage).includes(subError);
+        const isInteractionRequiredErrorDesc = !StringUtils.isEmpty(errorString) && Object.values(InteractionRequiredAuthErrorMessage).some((irErrorCode) => {
```
jasonnutter, line 51: `Object.values` is not supported in IE: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values

I'm not sure if Typescript provides a polyfill, can you check to see if this works in IE11?

lib/msal-common/src/error/InteractionRequiredAuthError.ts
```diff
@@ -10,41 +10,48 @@ import { ServerError } from "./ServerError";
...
  * Error thrown when user interaction is required at the auth server.
  */
 export class InteractionRequiredAuthError extends ServerError {
+    subError: string;
 
-    constructor(errorCode: string, errorMessage?: string) {
+    constructor(errorCode: string, errorMessage?: string, subError?: string) {
         super(errorCode, errorMessage);
         this.name = "InteractionRequiredAuthError";
+        this.subError = InteractionRequiredAuthError.validateSubError(subError);
```
jasonnutter, line 35: Curious why validation is needed, are there suberrors that will be returned from the service that we don't want to support?

lib/msal-browser/src/app/PublicClientApplication.ts
tnorling: The static method determines whether or not we should create an `InteractionRequiredAuthError`. I think this answers the question of whether or not it actually is one. Similar but slight difference and I would personally err towards checking what it is rather than what it should be.

lib/msal-common/src/error/InteractionRequiredAuthError.ts
tnorling: Good to know. Will look into it.

lib/msal-common/src/error/InteractionRequiredAuthError.ts
tnorling: This can be discussed with the team. I can't think of any we wouldn't want to surface. I was going off the previous research Jake had done [here](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=%2FMSALJS_APIChanges%2FSubErrorCodes.md&version=GBsub-error-codes&_a=preview), thinking it had already been agreed upon. He mentioned there might be scenarios where we don't want to surface certain sub-errors to the end-user.

jasonnutter: Okay, let's chat about it tomorrow. I don't remember this point specifically, might be forgetting.

lib/msal-browser/src/app/PublicClientApplication.ts
jasonnutter: Ah, good point.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -372,8 +372,9 @@ export class PublicClientApplication {
...
             const isInvalidGrantError = (e.errorCode === BrowserConstants.INVALID_GRANT_ERROR);
-            if (isServerError && isInvalidGrantError) {
+            if (isServerError && isInvalidGrantError && !isInteractionRequiredError) {
```
tnorling, line 377: I'm thinking we should make this just `!isInteractionRequiredError`. Do we only want to run through the whole flow of getting a code and exchanging it for a token if the access token is expired? Or anytime it fails for a reason other than needing interaction?

jasonnutter: The silent request that proceeds this if statement is specifically to handle the expired RT scenario (which manifests as the invalid grant error).

tnorling: Right but is that the only scenario we want to cover with this? An example I ran into when testing was when there was no Auth Code or RT available. It would fail to get a token above, but then would not enter this if block and the silent request would fail with a ClientAuthError. Is that desired?

jasonnutter: I believe so, @pkanher617 can confirm.

tnorling: The scenario I mentioned above will be handled in a different place in a different PR

lib/msal-common/src/error/InteractionRequiredAuthError.ts
tnorling: Done

PR #1537, opened 2020-04-23
[msal-common][msal-node] Move authority generation from common to node

sangonzal:
This PR moves authority creation from msal-common to msal-node. Instead of creating an authority for every client, we now just create one per application object, and pass this in to msal-common.


lib/msal-common/src/client/BaseClient.ts
```diff
@@ -65,28 +64,7 @@ export abstract class BaseClient {
         this.networkClient = this.config.networkInterface;
 
         // Default authority instance.
-        this.defaultAuthorityInstance = AuthorityFactory.createInstance(
```
sameerag, line 68: Will this not break msal-browser code?

sangonzal: SPAClient was also updated - it's part of this PR.

lib/msal-common/src/client/BaseClient.ts
```diff
@@ -65,28 +64,7 @@ export abstract class BaseClient {
...
-            this.config.authOptions.authority || Constants.DEFAULT_AUTHORITY,
-            this.networkClient
-        );
-    }
-
-    /**
-     * Create authority instance if not set already, resolve well-known-endpoint
-     * @param authorityString
-     */
-    protected async createAuthority(authorityString: string): Promise<Authority> {
```
tnorling, line 78: Is this not needed for browser?

lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -109,4 +121,31 @@ export abstract class ClientApplication {
...
+     */
+    private async createAuthority(authorityString?: string): Promise<Authority> {
+        const authority: Authority = authorityString
+            ? AuthorityFactory.createInstance(authorityString,  this.appConfig.system!.networkClient!)
+            : this.authority;
+
+        await authority.resolveEndpointsAsync().catch(error => {
+            throw ClientAuthError.createEndpointDiscoveryIncompleteError(error);
+        });
+        return authority;
```
jasonnutter, line 138: If you are going to use `await` here, it is idiomatic to use a `try/catch` instead of `.catch`:

```js
try {
  await authority.resolveEndpointsAsync();
  return authority;
} catch (error) {
  throw ClientAuthError.createEndpointDiscoveryIncompleteError(error);
}
```

lib/msal-common/src/client/BaseClient.ts
```diff
@@ -43,7 +42,7 @@ export abstract class BaseClient {
     protected account: Account;
 
     // Default authority object
-    protected defaultAuthorityInstance: Authority;
+    protected authority: Authority;
```
pkanher617, line 45: I don't think we should change the name of this field, it is still the defaultAuthorityInstance. Calling it just authority may lead to confusion down the road and makes the code a little less readable.

lib/msal-node/src/client/ClientApplication.ts
pkanher617: +1

lib/msal-node/src/config/ClientConfiguration.ts
```diff
@@ -3,15 +3,17 @@
...
 import {
-    AuthOptions,
     LoggerOptions,
     INetworkModule,
     LogLevel,
 } from '@azure/msal-common';
 import { NetworkUtils } from '../utils/NetworkUtils';
 import { CACHE } from '../utils/Constants';
 
-export type NodeAuthOptions = AuthOptions;
```
pkanher617, line 14: Shouldn't need this with new config changes in #1575

lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -109,4 +121,31 @@ export abstract class ClientApplication {
             },
         };
     }
+
+    /**
+     * Create authority instance. If authority not passed in request, default to authority set on the application
+     * object. If no authority set in application object, then default to common authority.
+     * @param authorityString
+     */
+    private async createAuthority(authorityString?: string): Promise<Authority> {
```
pkanher617, line 130: Couldn't you accomplish the same thing by exposing this as a public API on the base client? Then this could be used in any application that uses msal-common, and doesn't force msal-browser and msal-node to implement the same function.

lib/msal-node/src/client/ClientApplication.ts
sangonzal: Updated.

lib/msal-common/src/client/BaseClient.ts
sangonzal: This is a bit different than the way it was before. It made sense that this was defaultAuthorityInstance before, as its value was either the authority set on the application object or if none passed in, then the common endpoint. When the user made an acquireToken or Login call, if they didn't "override" the authority in the call, then we defaulted to this value. 

Now that is all done in either node or browser, and this is just the final authority that will be used for that request. We don't differentiate between a default authority and a overriden value.

lib/msal-node/src/config/ClientConfiguration.ts
sangonzal: yep reverted this change.

lib/msal-node/src/client/ClientApplication.ts
sangonzal: Ideally yes but we need to set _authority on the ClientApplication object, which we can't do from   common.

lib/msal-common/src/client/BaseClient.ts
pkanher617: In the context of node + common I agree, but in the context of just msal-common, this is still for all intents and purposes the default authority that is used if there is none passed in the request correct?

lib/msal-node/src/client/ClientApplication.ts
pkanher617: the public API can return an authority object that you set in ClientApplication. basically before you pass the config object you can call `AuthorityFactory.createAuthority()` with the function signature here and then use that authority for all calls. This way I don't need to need to write duplicate code in the browser, I can simply call the same API. See [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/1584/files/06ce0a617f343af9399e9b63bf5b8f7533116bbe#diff-3e92fba27aeb0efe69aaa786b5f9649b)

lib/msal-common/src/client/BaseClient.ts
sangonzal: Yep it is. Originally this PR only contained changes for node. Just added the same for browser. This is now in PublicClientApplication in browser.

lib/msal-common/src/client/BaseClient.ts
sangonzal: Originally this PR only moved authority generation for node. Yesterday I made similar changes for browser now that configuration has been aligned. This distinction is only applicable for SPAClient. I changed it back to defaultAuthority to reflect that difference, although it doesn't really apply for anything else.

lib/msal-node/src/client/ClientApplication.ts
sangonzal: Just ported the changes for browser, and this code is actually not needed there. Can you take a look and see if the changes make sense? Essentially since browser only uses SPAClient, this is not applicable, and thus there is only one default authority, this is not applicable.

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -2,15 +2,20 @@
...
+import { SystemOptions, LoggerOptions, INetworkModule, LogLevel } from "@azure/msal-common";
 import { BrowserUtils } from "../utils/BrowserUtils";
 import { BrowserConstants } from "../utils/BrowserConstants";
 
 // Default timeout for popup windows and iframes in milliseconds
 const DEFAULT_POPUP_TIMEOUT_MS = 60000;
 const DEFAULT_IFRAME_TIMEOUT_MS = 6000;
 
-export type BrowserAuthOptions = AuthOptions & {
+export type BrowserAuthOptions = {
```
sangonzal, line 13: @pkanher617 @sameerag FYI had to make this change to because common clients take `authority: Authority`.

This also now has the benefit that `NodeAuthOptions` doesn't expore `redirectUri`, `postLogoutRedirectUri`, and `NavigateToLoginRequestUrl`, which are not applicable, and thus makes the public API cleaner.

sameerag: I am fine with this.

lib/msal-common/src/client/SPAClient.ts
```diff
@@ -69,7 +70,7 @@ export class SPAClient extends BaseClient {
      */
     private async createUrl(request: AuthenticationParameters, isLoginCall: boolean): Promise<string> {
         // Initialize authority or use default, and perform discovery endpoint check.
-        const acquireTokenAuthority = (request && request.authority) ? AuthorityFactory.createInstance(request.authority, this.networkClient) : this.defaultAuthorityInstance;
+        const acquireTokenAuthority = (request && request.authority) ? AuthorityFactory.createInstance(request.authority, this.networkClient) : this.defaultAuthority;
```
sameerag, line 73: What is the user passes the same `authority` in `Configuration` and `Request`? We are still ending up creating an instance when not needed. Should we do the aliasing check and get away with that?

sangonzal: This is a good point. I have made a note of it in the comments. We should recommend people only pass in authority to client application, and only override in request if it's different. I will leave this as it is, but if we want to further optimize we can add this later on.

PR #1538, opened 2020-04-23
Add tests proving bug with ServerRequestParameters.populateQueryParams when passed null

jasonnutter:
Pulls in the fix from #1531 and adds tests to prove it is fixed.

### DIFF
--- a/lib/msal-core/src/ServerRequestParameters.ts
+++ b/lib/msal-core/src/ServerRequestParameters.ts
@@ -113,7 +113,7 @@ export class ServerRequestParameters {
         queryParameters = this.addHintParameters(account, queryParameters);
 
         // sanity check for developer passed extraQueryParameters
-        const eQParams: StringDict = request ? request.extraQueryParameters : null;
+        const eQParams: StringDict|null = request?.extraQueryParameters;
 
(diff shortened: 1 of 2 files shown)


lib/msal-core/src/ServerRequestParameters.ts
```diff
@@ -82,7 +82,7 @@ export class ServerRequestParameters {
      * @param request
      * @param serverAuthenticationRequest
      */
-    populateQueryParams(account: Account, request: AuthenticationParameters, adalIdTokenObject?: object, silentCall?: boolean): void {
+    populateQueryParams(account: Account, request: AuthenticationParameters|null, adalIdTokenObject?: object, silentCall?: boolean): void {
```
tnorling, line 85: I don't think we need to type things with ```|null``` since we don't have ```strictNullChecks``` enabled right? Seems like something we should get around to since it keeps coming up, but should enabling this and fixing the types be its own PR?

jasonnutter: Yeah, you're right, but I was fine with it since we know want to do it and it doesn't cause any harm.

tnorling (approved):

pkanher617 (approved):

Issue #1497, opened 2020-04-14
How do I automatically sign in on page load using `@azure/msal-browser@2.0.0`?


unindented:
## Library
- [ ] `msal@1.x.x` or `@azure/msal@1.x.x`
- [x] `@azure/msal-browser@2.0.0-beta.0`
- [ ] `@azure/msal-angular@0.x.x`
- [ ] `@azure/msal-angular@1.x.x`
- [ ] `@azure/msal-angularjs@1.x.x`

## Description

I'm modifying the `VanillaJSTestApp2.0` sample to sign in on page load, without user interaction. This is the diff: https://github.com/AzureAD/microsoft-authentication-library-for-js/compare/dev...unindented:dev

The problem is that, with functions like `getLoginInProgress` no longer available, I'm not able to tell whether I need to call `loginRedirect` on page load, so I end up in a loop:
- Page loads, `getAccount` returns `null`, so we call `signIn('loginRedirect')`.
- That redirects to the login flow, and send the user back to our page.
- Page loads, `getAccount` still returns `null`, and we have no way of checking that the login is in progress, so we call `signIn('loginRedirect')`.
- That redirects  to the login flow, etc.

pkanher617:
@unindented Could you explain your model a bit more? Are you attempting to (A) use an access token to acquire a resource on page load, or (B) are you trying to see if the user has an id token available?

If it is (A), then you can use the acquireTokenSilent method to fetch a token from the cache on page load, and if it fails you can use loginRedirect() with the same scope to trigger a login.

If it is (B), you can use the getAccount() method to see if an account object is available, and if no account object is returned (i.e. getAccount() = null) then you trigger a login. 

As I say in the docs, when the page comes back to the redirect URI with the authorization code, it does take a little bit of time to perform the token exchange. You can use handleRedirectCallback to trigger any functionality after the token exchange is done.

unindented:
I'm attempting to trigger sign in as soon as the user loads the page, without the user having to click a button.

I'm doing what you say in (B), but the problem is that, when the login flow redirects back to my page, `getAccount()` is  still `null`, cause `authRedirectCallBack` hasn't fired yet, so I call  `loginRedirect` again, and I get stuck in a loop. 

If you check the code in my fork of your sample app, and/or run it, you may understand my issue, cause it's difficult for me to explain in words any better.

deasems:
I have the same requirement. 

[...]

PR #1543, opened 2020-04-24
[msal-browser] Add pattern for login on page-load

pkanher617:
This PR fixes an issue raised in #1497 where calling loginRedirect on page load will cause the application to go into a redirect loop. 

This PR adds an API which allows applications using the redirect flow to utilize and notify their application that MSAL is done loading.

### DIFF
--- a/lib/msal-browser/src/app/PublicClientApplication.ts
+++ b/lib/msal-browser/src/app/PublicClientApplication.ts
@@ -96,6 +96,7 @@ export class PublicClientApplication {
     // #region Redirect Flow
 
     /**
+     * WARNING: This function will be deprecated soon.
      * Process any redirect-related data and send back the success or error object.
      * IMPORTANT: Please do not use this function when using the popup APIs, as it may break the response handling
      * in the main window.
@@ -106,6 +107,7 @@ export class PublicClientApplication {
      * containing data from the server (returned with a null or non-blocking error).
      */
     async handleRedirectCallback(authCallback: AuthCallback): Promise<void> {
+        console.warn("handleRedirectCallback will be deprecated upon release of msal-browser@v2.0.0. Please transition to using onRedirectAppLoad().");
         // Check whether callback object was passed.
         if (!authCallback) {
             throw BrowserConfigurationAuthError.createInvalidCallbackObjectError(authCallback);
@@ -124,9 +126,10 @@ export class PublicClientApplication {
 
     /**
      * Event handler function which allows users to fire events after the PublicClientApplication object 
-     * has loaded during redirect flows.
+     * has loaded during redirect flows. 
+     * @returns token response or null. If the return value is null, then no auth redirect was detected.
      */
-    async onRedirectAppLoad(): Promise<TokenResponse> {
+    async onRedirectAppLoad(): Promise<TokenResponse|null> {
         return this.tokenExchangePromise;
     }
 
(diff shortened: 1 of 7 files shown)


lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -122,6 +122,14 @@ export class PublicClientApplication {
         }
     }
 
+    /**
+     * Event handler function which allows users to fire events after the PublicClientApplication object 
+     * has loaded during redirect flows.
+     */
+    async onRedirectAppLoad(): Promise<TokenResponse> {
```
jasonnutter, line 129: We should document in the type that this can be null: `Promise<TokenResponse|null>`.

jasonnutter: And the typedoc comment shuold be updated to document why it would return null.

samples/VanillaJSTestApp2.0/app/onPageLoad/auth.js
```diff
@@ -0,0 +1,103 @@
...
+// Register Callbacks for Redirect flow
+myMSALObj.onRedirectAppLoad().then((tokenResponse) => {
+    const accountObj = tokenResponse ? tokenResponse.account : myMSALObj.getAccount();
+    if (accountObj) {
+        console.log('id_token acquired at: ' + new Date().toString());
+        showWelcomeMessage(accountObj);
+        seeProfileRedirect();
+    } else if (tokenResponse && tokenResponse.tokenType === "Bearer") {
+        console.log('access_token acquired at: ' + new Date().toString());
+    } else {
```
jasonnutter, line 26: Can you add comments documenting what each of these scenarios indicate?

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -122,6 +122,14 @@ export class PublicClientApplication {
         }
```
jasonnutter, line 122: Can you add a warning in `handlRedirectCallback` indicating it is deprecated, will be removed in the stable version, and that the new promise should be used instead?

lib/msal-browser/src/app/PublicClientApplication.ts
jasonnutter: And that apps should invoke it on all page loads if they are using the redirect flows and want to know when the redirect has been processed.

jasonnutter, line 129: Name proposal: `handleRedirectPromise` to keep it close to `handleRedirectCallback`

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: done.

samples/VanillaJSTestApp2.0/app/onPageLoad/auth.js
pkanher617: done.

lib/msal-browser/src/app/PublicClientApplication.ts

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -106,6 +107,7 @@ export class PublicClientApplication {
      * containing data from the server (returned with a null or non-blocking error).
      */
     async handleRedirectCallback(authCallback: AuthCallback): Promise<void> {
+        console.warn("handleRedirectCallback will be deprecated upon release of msal-browser@v2.0.0. Please transition to using onRedirectAppLoad().");
```
jasonnutter, line 110: @pkanher617 Forgot to update the warning with the new API name :(

Issue #1097, opened 2019-11-06
Enable `"strictNullChecks": true` everywhere


freshtonic:
## I'm submitting a...
<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Other... Please describe:
</code></pre>

## Browser:
- [ ] Chrome version XX
- [ ] Firefox version XX
- [ ] IE version XX
- [ ] Edge version XX
- [ ] Safari version XX

## Library version

<pre><code>
Library version: 1.2.0-beta.3
</code></pre>

## Current behavior

I discovered a bug where `UserAgentApplication.buildIDTokenRequest` was receiving an `undefined` `AuthenticationParameters` object and failing at runtime. On further investigation I could not comprehend how the calling code could compile because in the scope of the calling code the `request` parameter is optional. I then realised `strictNullChecks` was not enabled on the codebase. After enabling `strictNullChecks` the codebase is littered with compiler errors too numerous to be listed here.

## Expected behavior

The code should compile with `strictNullChecks` enabled.

## Minimal reproduction of the problem with instructions

Enable `strictNullChecks` in the base `tsconfig.base.json`, open the project in VSCode and see the red wiggly lines all over the place.

jasonnutter:
@freshtonic Thanks for the issue, didn't realize that was disabled, I'll definitely take a look.

ab0nent:
Is there any update on this bug? Will it be fixed? Thanks.

jasonnutter:
@ab0nent We haven't had a chance to get to this, but we will at some point soon.

sameerag:
We are working on this in the background, since some of the code needs to be rewritten or needs additional checks, we are taking an incremental approach. This will be updated once we have everything in place. FYI for future libraries, we are enabling this from the start.

tnorling:
Closing as this will no longer be picked up for version 1.x. This work will be done in 2.x, the first part of which has been started in #2602

PR #1544, opened 2020-04-24
Enable strictNullChecks in msal-core

jasonnutter:
Enables `strictNullChecks` in MSAL v1. 

Fixes:
- #1097


lib/msal-core/src/ServerRequestParameters.ts
```diff
@@ -36,7 +36,7 @@ export class ServerRequestParameters {
     promptValue: string;
     claimsValue: string;
 
-    queryParameters: string;
+    queryParameters: string|null;
```
sameerag, line 39: If strict null checks are on, would we need `|null`?

jasonnutter: Yes.

PR #1550, opened 2020-04-27
Update device code client

sangonzal:
- Map server response to expose device code response as camel case instead of snake case.
- Replace `console.log` with logger

### DIFF
--- a/lib/msal-common/src/client/DeviceCodeClient.ts
+++ b/lib/msal-common/src/client/DeviceCodeClient.ts
@@ -3,7 +3,7 @@
  * Licensed under the MIT License.
  */
 
-import { DeviceCodeResponse } from "../response/DeviceCodeResponse";
+import { DeviceCodeResponse, ServerDeviceCodeResponse } from "../response/DeviceCodeResponse";
 import { BaseClient } from "./BaseClient";
 import { DeviceCodeRequest } from "../request/DeviceCodeRequest";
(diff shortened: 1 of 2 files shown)


lib/msal-common/src/client/DeviceCodeClient.ts
```diff
@@ -61,13 +60,22 @@ export class DeviceCodeClient extends BaseClient {
...
             });
+
+        return {
+            userCode: serverDeviceCodeResponse.body.user_code,
+            deviceCode: serverDeviceCodeResponse.body.device_code,
+            verificationUri: serverDeviceCodeResponse.body.verification_uri,
+            expiresIn: serverDeviceCodeResponse.body.expires_in,
+            interval: serverDeviceCodeResponse.body.interval,
+            message: serverDeviceCodeResponse.body.message
+        };
```
jasonnutter, line 78: Nitpick / quick tip: 

ES6 improvements:

```js
const {
  body: {
    user_code: userCode,
    device_code: deviceCode,
    verification_uri: verificationUri,
    expires_in: expiresIn,
    interval,
    message
} = await this.networkClient.sendGetRequestAsync<ServerDeviceCodeResponse>(deviceCodeUrl, { headers });

return {
  userCode,
  deviceCode,
  verificationUrl,
  expiresUri,
  interval,
  message
};
```

Issue #1487, opened 2020-04-09
loginRedirect fails when multiple resources are specified for consent


deasems:
## Library
- [ ] `msal@1.x.x` or `@azure/msal@1.x.x`
- [x] `@azure/msal-browser@^2.0.0-beta.0`
- [ ] `@azure/msal-angular@0.x.x`
- [ ] `@azure/msal-angular@1.x.x`
- [ ] `@azure/msal-angularjs@1.x.x`

## Framework
React 16.12, though the issue occurs prior to loading any React libs or code.

## Description
I specify multiple resources in a `loginRedirect` call and the user is redirected back to the app after successful auth. I then see a 400 network error from a POST request to `https://login.microsoftonline.com/common/oauth2/v2.0/token`. The logged error is:
```
Auth Failed ServerError: 28000 - [2020-04-09 22:47:24Z]: AADSTS28000: Provided value for the input parameter scope is not valid because it contains more than one resource. Scope openid profile email https://graph.microsoft.com/user.readbasic.all https://microsoft.onmicrosoft.com/aria.api/user_impersonation offline_access is not valid.
Trace ID: 01c174e2-664f-42e9-a0f3-8121f8aa0000
Correlation ID: e4b33f3a-d5c4-442a-bd0f-743b84656134
Timestamp: 2020-04-09 22:47:24Z - Correlation ID: e4b33f3a-d5c4-442a-bd0f-743b84656134 - Trace ID: 01c174e2-664f-42e9-a0f3-8121f8aa0000
```

Based on [this](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/request-response-object.md#scopes) I'd expect to be able to specify up to 3 distinct resources in the login scopes array.

If I workaround this initial error by not specifying scopes in the login request, I then get this error when I silently acquire tokens: `ClientAuthError: No tokens were found for the given scopes, and no authorization code was passed to acquireToken. You must retrieve an authorization code before making a call to acquireToken(). Scopes: https://microsoft.onmicrosoft.com/aria.api/user_impersonation offline_access`. Please consider this scenario because in first party apps consent is not required (handled separately by preauthorization) so we should be able to silently get access tokens without pre-consenting.

## Security
Is this issue security related?
Unknown

## Regression
Did this behavior work before?
I've only tried v2 releases because I depend on auth code support. This issue has been present in both the alpha and beta releases.

## Configuration
[...]

PR #1553, opened 2020-04-28
[msal-common] Fix caching for multiple resources

tnorling:
`offline_access` scope was causing all scope sets to intersect, meaning all access tokens were being cached under the same entry with new scopes appended in the cache key. This PR aims to address this by ignoring `offline_access` when attempting to determine if we are adding or updating an access token in the cache.

This will fix: #1487


DarylThayil:
In general looks good can we add a test to validate different scope scenarios ?

lib/msal-common/src/auth/ScopeSet.ts
```diff
@@ -147,13 +147,21 @@ export class ScopeSet {
 
     /**
      * Check if scopes intersect between this set and another.
+     * Do not allow offline_access to be the only intersecting scope
```
jasonnutter, line 150: I think this comment should be below on 159, along with any other nuances you think someone should know.

lib/msal-common/src/auth/ScopeSet.ts
```diff
@@ -153,7 +153,15 @@ export class ScopeSet {
...
-        return this.unionScopeSets(otherScopes).size < (this.scopes.size + otherScopes.getScopeCount());
+
+        const unionScopes = this.unionScopeSets(otherScopes);
+
+        // Do not allow offline_access to be the only intersecting scope
+        const sizeOtherScopes = otherScopes.containsScope(Constants.OFFLINE_ACCESS_SCOPE)? otherScopes.getScopeCount() - 1 : otherScopes.getScopeCount();
+        const sizeThisScopes = this.containsScope(Constants.OFFLINE_ACCESS_SCOPE)? this.getScopeCount() - 1: this.getScopeCount();
+        const sizeUnionScopes = unionScopes.has(Constants.OFFLINE_ACCESS_SCOPE)? unionScopes.size - 1: unionScopes.size;
+
+        return sizeUnionScopes < (sizeThisScopes + sizeOtherScopes);
```
sameerag, line 164: Dumb q: Why are checking count than content? What if count matches and content differs for the set? I may be missing some logic and will check. This is mostly a question I immediately had.

tnorling: `unionScopes` is a set of scope array 1 + scope array 2 containing only unique entries. If the size of this is less than the size of both arrays added together then you know there were shared scopes between the 2 individual arrays

pkanher617: Size of both sets, not arrays. these are all unique values.

sameerag: okay, got it.

Issue #1406, opened 2020-03-22
AuthError object is empty when loginRedirect() fails


BenjaminBorlet:
>> Please follow the issue template below. Failure to do so will result in a delay in answering your question.

## Library
- [x] `msal@1.2.2.beta.3`
- [x] `@azure/msal-angular : 1.0.0-beta.4`

## Framework
Angular 9

## Description
The login workflow work fine using popUp workflow, but does not trigger loging and AuthError object if I use redirect workflow (i.e. loginRedirect())
## Security
No

## Configuration
Please provide your MSAL configuration options.

```js
function MSALConfigFactory(): Configuration {
  return {
    auth: {
      clientId: 'xxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxx',
      authority: 'https://login.microsoftonline.com/xxx.xxx.xxx',
      validateAuthority: true,
      redirectUri: 'http://localhost:4200/',
      postLogoutRedirectUri: 'http://localhost:4200/login',
      navigateToLoginRequestUrl: true
    },
    cache: {
      cacheLocation: 'localStorage',
      storeAuthStateInCookie: false // set to true for IE 11
    }
  };
}

function MSALAngularConfigFactory(): MsalAngularConfiguration {
  return {
    popUp: false,
    unprotectedResources: ['https://www.microsoft.com/en-us/'],
    extraQueryParameters: {}
  };
}
```

## Reproduction steps
1. Clone this repo
2. change popUp option to false
3. use       this.authService.loginRedirect(); instead of popUp in app.component.ts

## Expected behavior
All the log should appear in the same way with popUp or with redirect workflow
## Browsers
I only tested it on chrome last version

jasonnutter:
@BenjaminBorlet Are you referring to the broadcast events?

BenjaminBorlet:
`
this.authenticationService.handleRedirectCallback((authError, response) => {
      if (authError) {
        console.error('Redirect Error: ', authError.errorMessage);
        //authError is emtpy with redirect workflow
        return;
      }

      if (!response.account) {
        console.error(
          this.authenticationService
            .getCacheStorage()
            .getItem('msal.error.description')
//But here i get the correct value in the cache storage
        );
      }
    });
`

tnorling:
[...]

Issue #1554, opened 2020-04-28
"Forgot Password" click results in msal:acquireTokenSuccess Event


Hampfter:
## Library
- `"@azure/msal-angular": "^1.0.0-beta.5"`
- `"msal": "^1.3.0"`
 

## Framework
 `"@angular/core": "8.2.3",`

## Description
When I start a Login UserFlow and the User clicks on `Forgot your password?`, I should get a redirect to my Page with an AuthError like this:
```
ServerError: AADB2C90118: The user has forgotten their password.
Correlation ID: xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
```

The Redirect happens, but if i subscribe to the msal:loginFailure or msal:acquireTokenFailure events in the broadcastService, they are never triggered. Instread I get a msal:acquireTokenSuccess Event.
I guess the error has to be in the handleRedirectCallback method of the MsalService

## Regression
I'm not sure why this is happening. It worked with beta4 when I implemented it. Now it isn't even working with beta4.

## Configuration
```
MsalModule.forRoot(
{
	auth: {
		clientId: "**hidden**",
		authority: "**hidden**",
		validateAuthority: false,
		redirectUri: "**hidden**",
		postLogoutRedirectUri: "**hidden**"
	},
	cache: {
		cacheLocation: "localStorage",
		storeAuthStateInCookie: isIE // set to true for IE 11
	}
},
{
	unprotectedResources: [**hidden**],
	popUp: !isIE,
	consentScopes: [
	]
});

```

## Expected behavior
Getting the right event.

tnorling:
@Hampfter Msal just passes back what the service returns. If the service doesn't throw an error, neither will msal. Are you getting a token in the response? What happens when it triggers your acquireTokenSuccess event? Also, want to confirm, this is a b2c scenario?

jasonnutter:
@Hampfter Can you try reproducing this bug with [our Angular B2C sample](https://github.com/Azure-Samples/active-directory-b2c-javascript-angular-spa)? And then share the link to the fork here? Thanks!

Hampfter:
You only need to change the login method in the app.component.ts and use loginredirect.
```
  login() {
    if (isIE) {
      this.authService.loginRedirect(loginRequest);
    } else {
      this.authService.loginRedirect(loginRequest);
    }
  }
```

[...]

PR #1562, opened 2020-04-29
[msal-angular] Fix error broadcast in handleRedirectCallback

tnorling:
The order in which the angular handleRedirectCallback determines whether to return a success or failure is backwards. A response object will always be populated with at least state, whereas error will only be populated if there is an error. Therefore we need to check for error to be truthy first and if not, only then do we return a success.

This PR fixes: #1406 and #1554

### DIFF
--- a/lib/msal-angular/src/msal.service.ts
+++ b/lib/msal-angular/src/msal.service.ts
@@ -125,7 +125,7 @@ export class MsalService extends UserAgentApplication {
     handleRedirectCallback(authOrTokenCallback: authResponseCallback | tokenReceivedCallback, errorReceivedCallback?: errorReceivedCallback): void {
         super.handleRedirectCallback((authError: AuthError, authResponse: AuthResponse) => {
             if (authError) {
-                if (authResponse.tokenType === "id_token") {
+                if (!this.getAccount()) {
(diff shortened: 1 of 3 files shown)


jasonnutter:
Great find!

lib/msal-angular/src/msal.service.ts
```diff
@@ -124,31 +124,31 @@ export class MsalService extends UserAgentApplication {
     handleRedirectCallback(authCallback: authResponseCallback): void;
     handleRedirectCallback(authOrTokenCallback: authResponseCallback | tokenReceivedCallback, errorReceivedCallback?: errorReceivedCallback): void {
         super.handleRedirectCallback((authError: AuthError, authResponse: AuthResponse) => {
-            if (authResponse) {
+            if (authError) {
                 if (authResponse.tokenType === "id_token") {
```
tnorling, line 128: @jasonnutter The `authResponse` will not populate the `tokenType` in the case of an error. Should we collapse these 2 broadcasts into a single generic `msal:Error` or find a way to expose the type in core?

jasonnutter: AH, good. I think for now, we can infer it via `getAccount` (if it returns null, assume we were making a login call).

Issue #1310, opened 2020-02-25
Safari History with acquireTokenSilent


alex-volfson-jdas:
>> Please follow the issue template below. Failure to do so will result in a delay in answering your question.

## Library
- [x] `msal@1.2.1`
- [ ] `@azure/msal-browser@2.x.x`
- [ ] `@azure/msal-angular@0.x.x`
- [ ] `@azure/msal-angular@1.x.x`
- [ ] `@azure/msal-angularjs@1.x.x`

**Important**: Please fill in your exact version number above, e.g. `msal@1.1.3`.

## Framework
If you are using a framework, please provide the framework and version (e.g. Angular v8, React v16, etc).
Angular v8
## Description

Using acquireTokenSilent will add to browser history in Safari browser
## Security
Is this issue security related?
No
## Regression
Did this behavior work before?
Not sure
## Configuration
Please provide your MSAL configuration options.

```js
// Provide configuration values here.
// For Azure B2C issues, please include your policies.
```

## Reproduction steps
1. Login using a popup
2. In your app, go to a different view which requires a new scope and calls acquireTokenSilent
   a. The view should change the URL in the browser
3. Press the back button

## Expected behavior
Browser should take you back to previous view. Instead, the injected iframe adds to safari browser history.
## Browsers
Safari

widget31too:
This is also happening in Chrome, at least version 79.0.3945.130

jasonnutter:
@alex-volfson-jdas Apologies for the delay. We'll investigate this and follow up.

vikramdadwal:
Any updates on this issue? I am also facing the same issue

PR #1577, opened 2020-04-30
[msal-core] mitigates issue where iframe redirects adds to browser history

jo-arroyo:
```acquireTokenSilent``` was adding to browser history when using Safari. This PR addresses this by not assigning a ```src``` to the hidden iframe created in ```renewToken``` and ```renewIdToken```, which removes extra history steps in the Safari browser, while still allowing IE to use the hidden iframes created.

This addresses issue #1310


jo-arroyo:
Checked changes made to renewToken and renewIdToken in Safari, Chrome, and Firefox. Fix works. No other issues.

PR #1578, opened 2020-04-30
Allow applications to bypass network request for OpenID configuration

jasonnutter:
The allows applications to bypass the network request to retrieve the OpenID configuration that is normally fetched on every request. This data must be retrieved from the openid configuration endpoint and passed wholesale as a JSON string.

Steps:
1. Go to `https://login.microsoftonline.com/common/discovery/instance?api-version=1.0&authorization_endpoint={authorizeEndpoint}` for your authority: https://login.microsoftonline.com/common/discovery/instance?api-version=1.0&authorization_endpoint=https://login.microsoftonline.com/common/oauth2/v2.0/authorize
2. Go to the location set as the `tenant_discovery_endpoint`: https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration

A side-effect of this change is that the library will cache the response and use that for every call for the current session, regardless if it was provided in the configuration or retrieved via a network call.

Before:

Annotation 2020-05-18 141615

After: 

Annotation 2020-05-18 141850

TODO:
- [x] Additional tests
- [x] Documentation


lib/msal-core/src/Configuration.ts
```diff
@@ -117,6 +118,7 @@ const DEFAULT_AUTH_OPTIONS: AuthOptions = {
     clientId: "",
     authority: null,
     validateAuthority: true,
+    authorityMetadata: "",
```
DarylThayil, line 121: Do we want this to be null? If not passed in we dont want this to evaluate to truthy

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -220,6 +220,7 @@ export class UserAgentApplication {
         this.telemetryManager = this.getTelemetryManagerFromConfig(this.config.system.telemetry, this.clientId);
 
         AuthorityFactory.setKnownAuthorities(this.config.auth.validateAuthority, this.config.auth.knownAuthorities);
+        AuthorityFactory.parseAuthorityMetadata(this.config.auth.authority, this.config.auth.authorityMetadata);
```
DarylThayil, line 223: name? maybe setAuthorityMetadataFromJsonString

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -514,7 +515,11 @@ export class UserAgentApplication {
             }
         }
 
-        acquireTokenAuthority.resolveEndpointsAsync(this.telemetryManager, request.correlationId).then(async () => {
+        try {
+            if (!acquireTokenAuthority.validateTenantDiscoveryResponse()) {
+                await AuthorityFactory.resolveAuthorityAsync(serverAuthenticationRequest.authorityInstance);
+            }
```
DarylThayil, line 521: Is there an else here?

lib/msal-core/src/authority/Authority.ts
```diff
@@ -147,11 +148,21 @@ export abstract class Authority {
      * Discover endpoints via openid-configuration
      * If successful, caches the endpoint for later use in OIDC
      */
-    public async resolveEndpointsAsync(telemetryManager?: TelemetryManager, correlationId?: string): Promise<Authority> {
+    public async resolveEndpointsAsync(telemetryManager?: TelemetryManager, correlationId?: string): Promise<ITenantDiscoveryResponse> {
```
DarylThayil, line 151: Does this work for any authority? Like B2c?

lib/msal-core/src/authority/Authority.ts
```diff
@@ -147,11 +148,21 @@ export abstract class Authority {
...
         this.tenantDiscoveryResponse = await this.DiscoverEndpoints(openIdConfigurationEndpointResponse, telemetryManager, correlationId);
 
-        return this;
+        return this.tenantDiscoveryResponse;
+    }
+
+    /**
+     * Checks if there is a cached tenant discovery response with required fields.
+     */
+    public validateTenantDiscoveryResponse(): boolean {
```
DarylThayil, line 161: since this returns a boolean, it might be better named `hasValidTenantDiscoveryResponse`

lib/msal-core/src/authority/AuthorityFactory.ts
```diff
@@ -25,6 +28,27 @@ export class AuthorityFactory {
         }
     }
 
+    public static async resolveAuthorityAsync(authorityInstance: Authority): Promise<ITenantDiscoveryResponse> {
+        const metadata = await authorityInstance.resolveEndpointsAsync(); // todo: errors?
```
DarylThayil, line 32: oh this is where i was curious if this works for all sup authority types

lib/msal-core/src/authority/AuthorityFactory.ts
```diff
@@ -25,6 +28,27 @@ export class AuthorityFactory {
...
+        this.metadataMap.set(authorityInstance.CanonicalAuthority, metadata);
+        return metadata;
+    }
+
+    public static parseAuthorityMetadata(authorityUrl: string, authorityMetadataJson: string) {
```
DarylThayil, line 37: offered name change above

DarylThayil:
made some comments, if we can get those resolved and add error handling + tests lets ship it

lib/msal-core/src/Configuration.ts
jasonnutter: Empty string is falsey.

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: Nothing functional, probably good to add logger calls.

lib/msal-core/src/authority/Authority.ts
jasonnutter: Yes.

lib/msal-core/src/authority/Authority.ts
jasonnutter: Or maybe `hasCachedTenantMetadata`?

lib/msal-core/src/UserAgentApplication.ts
sameerag, line 223: Should we write a helper for all authority initiation code? Looks like we are doing one too many things when we initiate authority instance.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -483,13 +484,13 @@ export class UserAgentApplication {
      * Helper function to acquireToken
      *
      */
-    private acquireTokenHelper(account: Account, interactionType: InteractionType, isLoginCall: boolean, request: AuthenticationParameters, resolve?: any, reject?: any): void {
+    private async acquireTokenHelper(account: Account, interactionType: InteractionType, isLoginCall: boolean, request: AuthenticationParameters, resolve?: any, reject?: any): Promise<void> {
```
sameerag, line 487: Does this effect performance for calls that don't need this async?

lib/msal-core/src/authority/Authority.ts
sameerag: I like `hasCachedMetadata`

lib/msal-core/src/authority/AuthorityFactory.ts
```diff
@@ -25,6 +28,27 @@ export class AuthorityFactory {
...
+        try {
+            if (authorityMetadataJson) {
```
sameerag, line 39: Just clarifying: The end user is expected to send a json and we parse out the needed values. And this isn't limited to the three values we are setting here and expandable in the future. Is my understanding correct?

sameerag (commented):
Over all looks good. One concern I had is the bubbling up `async` for the interactive helper function `acquireTokenHelper` and the impact of performance for operations that do not need `async`.

lib/msal-core/src/authority/AuthorityFactory.ts
jasonnutter: Correct, this API expects a JSON string.

lib/msal-core/src/UserAgentApplication.ts
tnorling: I can add this idea to the cloud instance PR after this PR gets merged as I'll be adding a few more steps on initiation in that PR.

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: Potentially, but should be offset when not making the openid-config call at all. Will verify.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -718,42 +726,46 @@ export class UserAgentApplication {
...
-                                this.silentLogin = true;
-                                this.renewIdToken(requestSignature, resolve, reject, account, serverAuthenticationRequest);
-                            } else {
-                                // renew access token
-                                this.logger.verbose("renewing accesstoken");
-                                this.renewToken(requestSignature, resolve, reject, account, serverAuthenticationRequest);
-                            }
+
+                try {
+                    if (!serverAuthenticationRequest.authorityInstance.hasCachedMetadata()) {
```
tnorling, line 733: Can we put this if/else block in a helper function? We're going to need to repeat this in a number of places, no?

lib/msal-core/src/authority/AuthorityFactory.ts
```diff
@@ -25,6 +29,32 @@ export class AuthorityFactory {
         }
     }
 
+    public static async resolveAuthorityAsync(authorityInstance: Authority, telemetryManager: TelemetryManager, correlationId: string): Promise<ITenantDiscoveryResponse> {
```
tnorling, line 32: Can we brainstorm other names for this and `parseAuthorityMetadata` to more clearly describe what they are doing and how they are related? Something along the lines of `saveMetadataFromNetwork` and `saveMetadataFromConfig`?

tnorling: This is a little too similar to `resolveEndpointsAsync` and it took me a minute to understand how these differ and why this needs to be in AuthorityFactory

lib/msal-core/src/UserAgentApplication.ts
pkanher617: That's a bit wordy, I think parseAuthorityMetadata is fine.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -514,7 +515,14 @@ export class UserAgentApplication {
             }
         }
 
-        acquireTokenAuthority.resolveEndpointsAsync(this.telemetryManager, request.correlationId).then(async () => {
+        try {
+            if (!acquireTokenAuthority.hasCachedMetadata()) {
```
pkanher617, line 519: Instead of checking for this here, why don't we just do this inside resolveAuthorityAsync()? If the cached metadata is available, just return it, otherwise make a network request (similar to acquireTokenSilent)

lib/msal-core/src/UserAgentApplication.ts
pkanher617: See above comment about adding this as part of resolveAuthorityAsync.

lib/msal-core/src/authority/AuthorityFactory.ts
pkanher617: +1 for naming.

lib/msal-core/src/error/ClientConfigurationError.ts
```diff
@@ -7,7 +7,12 @@ import { Constants } from "../utils/Constants";
 import { ClientAuthError } from "./ClientAuthError";
 import { TelemetryOptions } from "../Configuration";
 
-export const ClientConfigurationErrorMessage = {
+interface IClientConfigurationErrorMessage {
```
pkanher617, line 10: Why are we making this change for error message classes, and should it be done for all error message classes?

lib/msal-core/src/UserAgentApplication.ts
tnorling: I like this too. Though this would prevent intentional refreshes without an additional `forceRefresh` parameter. Not sure if there's a valid argument for supporting that use case but something to consider.

pkanher617: We can't do intentional refreshes anyway with the current code.

jasonnutter: One of the goals is to minimize the number of async code paths, which is why the `if` is done here, so that way if the check fails, we don't enter an async function. If this was factored out, it would entered the async every time.

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: Same comment about wanting to minimize async calls.

lib/msal-core/src/error/ClientConfigurationError.ts
jasonnutter: I felt we should codify the pattern of `code/desc` in the object below. I would actually complete refactor this into a different pattern, but felt this was an incremental improvement.

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: Also, I did not consider if `forceRefresh` should force a network call for openid configuration, in addition to forcing a network call for a new token. Initially, I would think not. @DarylThayil any opinion here?

lib/msal-core/src/authority/AuthorityFactory.ts
jasonnutter: Renamed based on feedback.

tnorling (approved):

DarylThayil (approved):

Issue #1381, opened 2020-03-16
Hidden iFrame accessibility error


Sl1MBoy:
I have noticed that the application is loaded to the hidden iframe while using msal. Unfortunately this causing A11Y voice over features to focus and read out elements inside the iframe. This can be prevented by adding aria-hidden="true" attribute to the iframe. 

I have experienced this with msal 1.2.1 and Safari browser.

jasonnutter:
@Sl1MBoy Thanks for the information, we definitely want to make sure this is fixed.

DarylThayil:
Note: This may be solved in 1.2.2 by removing hidden iframes, but we should add correct aria tags regardless

jo-arroyo:
@Sl1MBoy We believe we have a potential fix for this issue in #1581. However, we are having trouble reproducing the behavior you described. Would it be possible for you to provide a video of demonstrating this behavior? Thanks!

PR #1581, opened 2020-05-01
Add hidden iframe accessibility support

jo-arroyo:
Voiceover features were reading out elements inside hidden iframes. This PR addresses this by adding aria-hidden="true" attribute to creation of iframes in msal-core.

This will fix: #1381


DarylThayil:
great job!

PR #1582, opened 2020-05-01
Use instance discovery for trusted hosts and combine authority classes

tnorling:
This PR does several things in regards to Authorities:

- Removes the hardcoded trusted host list for AAD hosts and replaces it with a network call to discover hosts known to MS. This is done the first time `resolveEndpointsAsync` is called.
- Combines B2C and AAD authority classes into a single generic class

Developers can opt out of the network call in one of 3 ways:
1. Set `validateAuthority: false`
2. Provide all trusted authorities they intend to use in `knownAuthorities` config option
3. Provide instance metadata via `authorityMetadata` config option

Note: This does not include `preferred_cache` or `preferred_network` fields and does not include a scheduled refresh. These can be added in separate PRs if needed for core.


lib/msal-core/src/telemetry/TelemetryUtils.ts
```diff
@@ -1,15 +1,14 @@
...
 import { UrlUtils } from "../utils/UrlUtils";
+import { TrustedAuthority } from "../authority/TrustedAuthority";
 
 export const scrubTenantFromUri = (uri: string): String => {
 
     const url = UrlUtils.GetUrlComponents(uri);
 
     // validate trusted host
-    if (!AADTrustedHostList[url.HostNameAndPort.toLocaleLowerCase()]) {
+    if (!TrustedAuthority.IsInTrustedHostList(url.HostNameAndPort.toLocaleLowerCase())) {
```
tnorling, line 11: Since we no longer distinguish between AAD/B2C can we change this to just check if the authority is ADFS since ADFS does not have the concept of a tenant and this code will never be executed if the host is not trusted? It looks like the logic below will still work for B2C

tnorling:
Per discussion: Will look at kicking off Instance Discovery in constructor and having the `resolveEndpointAsync` call await the response

jasonnutter:
> Per discussion: Will look at kicking off Instance Discovery in constructor and having the `resolveEndpointAsync` call await the response

As we saw when doing the React POC, having the constructor perform side effects like that I believe is a bad idea.

tnorling:
> > Per discussion: Will look at kicking off Instance Discovery in constructor and having the `resolveEndpointAsync` call await the response
> 
> As we saw when doing the React POC, having the constructor perform side effects like that I believe is a bad idea.
> 
> I would personally argue that we should provide mechanisms to avoid the call at all, instead of potentially over-optimizing by invoking the network call before any token requests are made.
> 
> cc: @DarylThayil

Tested and it seems to work but open to discussing this further. I don't feel strongly one way or the other. Mechanisms to avoid the network call are 1. Provide your authority in `knownAuthorities` or 2. Provide your metadata

tnorling:
Per discussion: First invocation of `resolveEndpointsAsync` will call Discovery Endpoint if `TrustedHostList` is not already set.

lib/msal-core/src/authority/TrustedAuthority.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * 
+     * @param telemetryManager 
+     * @param correlationId 
+     */
+    private static async getAliases(telemetryManager: TelemetryManager, correlationId?: string): Promise<Array<any>> {
+        const client: XhrClient = new XhrClient();
+
+        const httpMethod = "GET";
```
technical-boy, line 30: Do we have a constant for this?

pkanher617: +1 for using a constant here.

lib/msal-core/src/telemetry/TelemetryUtils.ts
```diff
@@ -1,15 +1,14 @@
-import { B2cAuthority } from "../authority/B2cAuthority";
-import { AADTrustedHostList } from "../utils/Constants";
 import { TENANT_PLACEHOLDER, EVENT_NAME_PREFIX } from "./TelemetryConstants";
 import { CryptoUtils } from "../utils/CryptoUtils";
 import { UrlUtils } from "../utils/UrlUtils";
+import { TrustedAuthority } from "../authority/TrustedAuthority";
 
 export const scrubTenantFromUri = (uri: string): String => {
```
pkanher617, line 6: Can we create a utility class with static functions instead of exporting function constants?

lib/msal-core/src/telemetry/TelemetryUtils.ts
pkanher617: +1 to just checking for ADFS

lib/msal-core/src/authority/TrustedAuthority.ts
tnorling: Done

lib/msal-core/src/telemetry/TelemetryUtils.ts
tnorling: Done

lib/msal-core/src/telemetry/TelemetryUtils.ts
tnorling: I think that's reasonable but since this PR isn't really changing any telemetry files other than to fix compilation, I think that can be taken separately.

PR #1551, opened 2020-04-27
[msal-browser] Export TokenResponse at top level

HarryGifford:
`TokenResponse` is a return type and argument type from a few `PublicClientApplication` methods.

PR #1584, opened 2020-05-02
[msal-browser][msal-common] Instance Aware Authentication

pkanher617:
This PR implements the instance_aware feature for msal-browser. Specifically, it adds the authority object to the response and handles all parameters which come back in the fragment.

It also adds TokenResponse to the browser exports (as per PR #1551).


lib/msal-common/src/client/SPAClient.ts
```diff
@@ -69,12 +70,14 @@ export class SPAClient extends BaseClient {
...
-            await acquireTokenAuthority.resolveEndpointsAsync();
-        } catch (e) {
-            throw ClientAuthError.createEndpointDiscoveryIncompleteError(e);
-        }
+        // const acquireTokenAuthority = (request && request.authority) ? AuthorityFactory.createInstance(request.authority, this.networkClient) : this.defaultAuthorityInstance;
+        // try {
+        //     await acquireTokenAuthority.resolveEndpointsAsync();
+        // } catch (e) {
+        //     throw ClientAuthError.createEndpointDiscoveryIncompleteError(e);
+        // }
```
jasonnutter, line 78: Can these commented blocks be deleted?

lib/msal-common/src/server/ServerAuthorizationCodeResponse.ts
```diff
@@ -8,13 +8,23 @@
...
  */
 export type ServerAuthorizationCodeResponse = {
+    // Success case
     code?: string;
     client_info?: string;
     state?: string;
+    cloud_instance_name?: string;
+    cloud_instance_host_name?: string;
+    cloud_graph_host_name?: string;
+    msgraph_host?: string;
```
tnorling, line 26: Why does graph need to be included in this?

lib/msal-common/src/client/SPAClient.ts
```diff
@@ -318,18 +329,56 @@ export class SPAClient extends BaseClient {
         this.cacheManager.resetTempCacheItems(cachedState || "");
     }
 
+    /**
+     * Gets a valid authority object and resolves endpoints if needed.
+     * @param authorityUri 
+     */
+    private async getDiscoveredAuthority(authorityUri: string): Promise<Authority> {
```
tnorling, line 336: Maybe this along with `getAuthorityObject` belongs in Authority or AuthorityFactory?

pkanher617: Not a bad idea, I will see if I can make that change.

lib/msal-common/src/server/ServerAuthorizationCodeResponse.ts
pkanher617: These are the parameters sent back when instance_aware=true, I believe it is because the graph host is different for different clouds. We may need to add this to the response in the future, for now I am just adding it as part of the params  sent back in the hash.

tnorling: I feel like we shouldn't be coding against specific resources. I'm not totally clear on what "instance aware" means in this context but are you saying that the developer would not know which graph host they need to hit without this in the response? Shouldn't the developer be responsible for knowing their resource endpoints?

sameerag:
@pkanher617 Lets do this review in person. Just to check we adhere to user expectations as other MSALs in this case. Thanks for initiating this.

PR #1600, opened 2020-05-05
Add scope trimming for msal-core

pkanher617:
This PR fixes an issue raised in #845 where a leading space in the scopes was causing an issue in caching. This PR ensure that scopes are always trimmed and made lower case before being used in any authentication calls or caching.


lib/msal-core/test/UserAgentApplication.spec.ts
```diff
@@ -1008,7 +1009,7 @@ describe("UserAgentApplication.ts Class", function () {
             cacheStorage.setItem(JSON.stringify(accessTokenKey), JSON.stringify(accessTokenValue));
 
             msal.acquireTokenSilent(tokenRequest).then(function(response) {
-                expect(response.scopes).to.be.deep.eq(tokenRequest.scopes);
+                expect(response.scopes).to.be.deep.eq(ScopeSet.trimAndConvertToLowerCase(tokenRequest.scopes));
```
jasonnutter, line 1012: It would be better to hard code the expected response, instead of calling the same implementation detail again.

pkanher617: done

lib/msal-core/src/ScopeSet.ts
```diff
@@ -16,9 +16,10 @@ export class ScopeSet {
      */
     // TODO: Rename this, intersecting scopes isn't a great name for duplicate checker
     static isIntersectingScopes(cachedScopes: Array<string>, scopes: Array<string>): boolean {
-        cachedScopes = this.convertToLowerCase(cachedScopes);
-        for (let i = 0; i < scopes.length; i++) {
-            if (cachedScopes.indexOf(scopes[i].toLowerCase()) > -1) {
+        const convertedCachedScopes = this.trimAndConvertToLowerCase([...cachedScopes]);
+        const requestScopes = this.trimAndConvertToLowerCase([...scopes]);
+        for (let i = 0; i < requestScopes.length; i++) {
```
DarylThayil, line 21: not a huge deal here, but calling out this is n^2

lib/msal-core/src/ScopeSet.ts
```diff
@@ -32,8 +33,9 @@ export class ScopeSet {
      * @param scopes
      */
     static containsScope(cachedScopes: Array<string>, scopes: Array<string>): boolean {
-        cachedScopes = this.convertToLowerCase(cachedScopes);
-        return scopes.every((value: any): boolean => cachedScopes.indexOf(value.toString().toLowerCase()) >= 0);
+        const convertedCachedScopes = this.trimAndConvertToLowerCase([...cachedScopes]);
+        const requestScopes = this.trimAndConvertToLowerCase([...scopes]);
+        return requestScopes.every((value: any): boolean => convertedCachedScopes.indexOf(value.toString().toLowerCase()) >= 0);
```
DarylThayil, line 38: is every supported in all browsers @jasonnutter

lib/msal-core/src/ScopeSet.ts
```diff
@@ -54,7 +56,8 @@ export class ScopeSet {
      */
     // TODO: Rename this, too generic name for a function that only deals with scopes
     static removeElement(scopes: Array<string>, scope: string): Array<string> {
-        return scopes.filter(value => value !== scope);
+        const scopeVal = scope.trim().toLowerCase();
```
DarylThayil, line 59: is there a way to have this be encapsulated by trimAndConvertToLowerCase so we are not duplicating this logic?

lib/msal-core/src/ServerRequestParameters.ts
```diff
@@ -58,7 +59,8 @@ export class ServerRequestParameters {
         this.nonce = CryptoUtils.createNewGuid();
 
         // set scope to clientId if null
-        this.scopes = scopes? [ ...scopes] : [clientId];
+        this.scopes = scopes ? [ ...scopes] : [clientId];
```
DarylThayil, line 62: if scopes is already an array why spread it into a new one?

lib/msal-core/src/ScopeSet.ts
jasonnutter: Yes.

lib/msal-core/src/ServerRequestParameters.ts
pkanher617: To prevent mutation of the original object.

pkanher617: Before if we directly assigned the value it would end up mutating the array passed in by the user.

DarylThayil: okay awesome!

lib/msal-core/test/UserAgentApplication.spec.ts
jasonnutter: Still seeing the same code?

lib/msal-core/src/ScopeSet.ts
pkanher617: Making this linear complexity is going to require implementing a hash table, because the scope arrays are unsorted and there is no way to order scopes.

lib/msal-core/src/ScopeSet.ts
pkanher617: done.

lib/msal-core/test/UserAgentApplication.spec.ts
pkanher617: Changes got overwritten for some reason. Updated.

lib/msal-core/src/ScopeSet.ts
DarylThayil: Yeah I think its fine if its exponential, there are limited scopes so its not a huge deal

PR #1601, opened 2020-05-05
[msal-common][msal-browser] Remove RT from acquire token entries and make it top-level in cache 

pkanher617:
This PR removes the refresh token that is cached with access token entries today and creates a top-level Persistent Cache Item (similar to ID token) that is used to obtain new tokens from the service. This item has a 24 hour lifetime, and a new value is received on receipt of a new token.

We have also removed the refresh token from the response object, as it should not be returned to applications.

This PR also include a small change to the RefreshTokenClient that will add default scopes to the call.

This PR is an intermediate step until the full changes for the cache are ready.

### DIFF
--- a/lib/msal-common/src/client/SPAClient.ts
+++ b/lib/msal-common/src/client/SPAClient.ts
@@ -20,13 +20,14 @@ import { AccessTokenCacheItem } from "../cache/AccessTokenCacheItem";
 import { AuthorityFactory } from "../authority/AuthorityFactory";
 import { IdToken } from "../account/IdToken";
 import { ScopeSet } from "../request/ScopeSet";
-import { TemporaryCacheKeys, PersistentCacheKeys, AADServerParamKeys } from "../utils/Constants";
+import { TemporaryCacheKeys, PersistentCacheKeys, AADServerParamKeys, Constants } from "../utils/Constants";
 import { TimeUtils } from "../utils/TimeUtils";
 import { StringUtils } from "../utils/StringUtils";
 import { UrlString } from "../url/UrlString";
 import { Account } from "../account/Account";
 import { buildClientInfo } from "../account/ClientInfo";
 import { B2cAuthority } from "../authority/B2cAuthority";
+import { ServerError } from "../error/ServerError";
 
 /**
  * SPAClient class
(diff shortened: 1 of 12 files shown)


lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -154,13 +154,18 @@ export class ResponseHandler {
...
 
-        // Get id token
+        // Get and save id token
         if (!StringUtils.isEmpty(originalTokenResponse.idToken)) {
             this.cacheStorage.setItem(PersistentCacheKeys.ID_TOKEN, originalTokenResponse.idToken);
         }
 
+        // Get and save refresh token
+        if (serverTokenResponse && !StringUtils.isEmpty(serverTokenResponse.refresh_token)) {
+            this.cacheStorage.setItem(PersistentCacheKeys.REFRESH_TOKEN, serverTokenResponse.refresh_token);
```
pkanher617, line 164: Need to cache using oid

lib/msal-common/src/client/SPAClient.ts
```diff
@@ -235,11 +235,17 @@ export class SPAClient extends BaseClient {
                 return StringUtils.isEmpty(cachedTokenItem.value.idToken) ? defaultTokenResponse :
                     ResponseHandler.setResponseIdToken(defaultTokenResponse, new IdToken(cachedTokenItem.value.idToken, this.cryptoUtils));
             } else {
-                // Renew the tokens.
+                // Get the current refresh token
+                const refreshToken = this.cacheStorage.getItem(PersistentCacheKeys.REFRESH_TOKEN);
+                if (StringUtils.isEmpty(refreshToken)) {
+                    // TODO: Figure out what to throw in this case
+                    throw new ServerError(Constants.INVALID_GRANT_ERROR, "No refresh token found in cache.");
```
pkanher617, line 242: Update this to throw a library error instead

deasems:
Is this planned to be part of the next release? I'm running into the issue of not being able to silently get tokens for new resources after initial auth which I understand this PR addresses.

pkanher617:
Closing as this is addressed by UnifiedCache

PR #1603, opened 2020-05-05
@azure/msal-browser and @azure/msal-common beta.1 release

pkanher617:
This PR adds the final changes for the beta.1 release for 1.0.0 version of @azure/msal-common and the 2.0.0 version of @azure/msal-browser.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -7,7 +7,6 @@
         "@azure/msal-common", 
         "@azure/msal-browser", 
         "@azure/msal-angular",
-        "@azure/msal-node",
         "vanilla-js-test-app",
         "vanilla-js-test-app-2.0"
       ]
--- a/lerna.json
+++ b/lerna.json
@@ -7,6 +7,7 @@
         "@azure/msal-common", 
         "@azure/msal-browser", 
         "@azure/msal-angular",
+        "@azure/msal-node",
         "vanilla-js-test-app",
         "vanilla-js-test-app-2.0"
       ]
--- a/lib/msal-browser/docs/initialization.md
+++ b/lib/msal-browser/docs/initialization.md
@@ -83,13 +83,17 @@ The popup APIs use ES6 Promises that resolve when the authentication flow in the
 - `loginRedirect`
 - `acquireTokenRedirect`
 
-The redirect APIs do not use Promises, but are `void` functions which redirect the browser window after caching some basic info. If you choose to use the redirect APIs, you **MUST** call the following function **IMMEDIATELY** after the constructor in order to process the returning fragment containing the code and response and obtain an access token:
+The redirect APIs do not use Promises, but are `void` functions which redirect the browser window after caching some basic info. If you choose to use the redirect APIs, be aware that there is a small delay after loading the MSAL object where the exchanging of the auth code for the token is occurring. You can use the following function to perform an action when this token exchange is completed:
 ```javascript
-msalInstance.handleRedirectCallback((error, response) => {
-    // handle redirect response or error
-});
+msalInstance.handleRedirectPromise()
+.then((tokenResponse) => {
+    // Check if the tokenResponse is null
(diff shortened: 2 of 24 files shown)


lib/msal-browser/docs/initialization.md
```diff
@@ -83,13 +83,17 @@ The popup APIs use ES6 Promises that resolve when the authentication flow in the
...
-msalInstance.handleRedirectCallback((error, response) => {
-    // handle redirect response or error
-});
+msalInstance.handleRedirectPromise()
+.then((tokenResponse) => {
+    // Check if the tokenResponse is null
+    // If the tokenResponse !== null, then you have successfully authenticated. 
+    // If the tokenResponse === null, you are not authenticated.
+})
+.catch();
```
jasonnutter, line 94: I would expand this to like `.then` to show that it will be invoked if there is an error.

lib/msal-browser/docs/login-user.md
```diff
@@ -62,6 +66,23 @@ When a login call has succeeded, you can use the `getAccount()` function to retr
...
+```javascript
+const silentRequest = {
+    scopes: ["User.Read", "Mail.Read"],
+    loginHint: "user@contoso.com"
+};
+
+try {
+    const loginResponse = await msalInstance.ssoSilent(silentRequest);
+} catch (err) {
+    // handle error
```
jasonnutter, line 82: Should show invoking loginPopup.

lib/msal-browser/docs/initialization.md
```diff
@@ -83,13 +83,17 @@ The popup APIs use ES6 Promises that resolve when the authentication flow in the
...
+The redirect APIs do not use Promises, but are `void` functions which redirect the browser window after caching some basic info. If you choose to use the redirect APIs, be aware that there is a small delay after loading the MSAL object where the exchanging of the auth code for the token is occurring. You can use the following function to perform an action when this token exchange is completed:
 ```javascript
-msalInstance.handleRedirectCallback((error, response) => {
-    // handle redirect response or error
-});
+msalInstance.handleRedirectPromise()
+.then((tokenResponse) => {
+    // Check if the tokenResponse is null
+    // If the tokenResponse !== null, then you have successfully authenticated. 
+    // If the tokenResponse === null, you are not authenticated.
```
jasonnutter, line 92: Hmm this isn't quite true, right? If `tokenResponse` is `null` that means the page isn't returning from a redirect operation, right?

jasonnutter: i.e. if you are not returning from a redirect operation in `null` is returned, it doesn't mean you aren't authenticated.

pkanher617: Good point, will update

lib/msal-browser/docs/initialization.md
pkanher617: done.

lib/msal-browser/docs/login-user.md
pkanher617: done.

lib/msal-browser/package.json
```diff
@@ -10,7 +10,7 @@
         "type": "git",
         "url": "https://github.com/AzureAD/microsoft-authentication-library-for-js.git"
     },
-    "version": "2.0.0-beta.0",
+    "version": "2.0.0-beta.1",
```
jasonnutter, line 13: The lockfile also has a version that needs to be bumped.

pkanher617: done.

lerna.json
```diff
@@ -7,7 +7,6 @@
         "@azure/msal-common", 
         "@azure/msal-browser", 
         "@azure/msal-angular",
-        "@azure/msal-node",
```
sangonzal, line 10: Will this break lerna bootstrap for node?

PR #1604, opened 2020-05-06
Fix redirection to urls containing query strings

### LINKED ISSUE #1602: Uris with query params do not work as redirect uris
## Library
- [x] `msal@1.3.0`

## Framework
none

## Description
When using a query parameter in a redirect uri, msal js fails to sign in the user. For example:

Using this uri: `http://localhost:3000/index.html?someparam=bla`

When going through the redirect flow and the user is signed in, the user is redirected to `http://localhost:3000/index.html?someparam=bla#id_token=...`, but the msal library does not get the id token and finish the authentication.

When using `http://localhost:3000/index.html` (no query params), the redirect works as expected and the user is signed in.

## Regression
Yes, this worked in 1.1.3

## Configuration
Please provide your MSAL configuration options.

```js
{
    auth: {
        clientId: [id]
    },
    cache: {
        storeAuthStateInCookie: true
    }
}
```

## Expected behavior
The msal library should read the token from the hash and sign in the user.

## Browsers
I've tested on Edge Canary.

tnorling:
This PR fixes #1602 by ensuring query strings don't disrupt the redirect flow. Msal compares the `currentUrl` to the `loginRequestUrl` to determine if another redirect is required. 

**Previous Behavior:** The `currentUrl` was being stripped of both hash and query string whereas the `loginRequestUrl` was left as is. This was fine for hash routing strategies because we had another check to see if a hash was present in the `loginRequestUrl`. However, with a query string present the Url comparison check will always fail and the token response hash will not be processed.

**New Behavior:** Both `currentUrl` and `loginRequestUrl` will be normalized and compared using the full url minus any non-msal hash to determine if a redirect is needed. If they are equal (final page has been reached), add any non-msal hash back onto the url and then process the msal hash.

TODO: 
- [x] Unit Tests
- [x] Sanity Checks

### DIFF
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -1113,16 +1113,19 @@ export class UserAgentApplication {
                 this.logger.error("Unable to get valid login request url from cache, redirecting to home page");
                 window.location.href = "/";
                 return;
-            } else if (currentUrl !== loginRequestUrl) {
-                // If loginRequestUrl contains a hash (e.g. Angular routing), process the hash now then redirect to prevent both hashes in url
-                if (loginRequestUrl.indexOf("#") > -1) {
-                    this.logger.info("loginRequestUrl contains hash, processing response hash immediately then redirecting");
-                    this.processCallBack(hash, stateInfo, null);
-                    window.location.href = loginRequestUrl;
-                } else {
+            } else {
+                if (!UrlUtils.isSamePage(currentUrl, loginRequestUrl)) {
                     window.location.href = `${loginRequestUrl}${hash}`;
+                    return;
+                } else {
+                    let loginRequestUrlComponents = UrlUtils.GetUrlComponents(loginRequestUrl);
+                    if (loginRequestUrlComponents.Hash){
+                        window.location.hash = loginRequestUrlComponents.Hash;
+                    }
+                    if (loginRequestUrlComponents.Search){
+                        window.location.search = loginRequestUrlComponents.Search;
+                    }
                 }
-                return;
             }
         }
 
--- a/lib/msal-core/src/UserAgentApplication.ts
+++ b/lib/msal-core/src/UserAgentApplication.ts
@@ -1106,7 +1106,6 @@ export class UserAgentApplication {
         // if set to navigate to loginRequest page post login
         if (this.config.auth.navigateToLoginRequestUrl && window.parent === window) {
             const loginRequestUrl = this.cacheStorage.getItem(`${TemporaryCacheKeys.LOGIN_REQUEST}${Constants.resourceDelimiter}${stateInfo.state}`, this.inCookie);
-            const currentUrl = UrlUtils.getCurrentUrl();
 
             // Redirect to home page if login request url is null (real null or the string null)
             if (!loginRequestUrl || loginRequestUrl === "null") {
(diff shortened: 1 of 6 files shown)


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1111,18 +1111,21 @@ export class UserAgentApplication {
...
             if (!loginRequestUrl || loginRequestUrl === "null") {
                 this.logger.error("Unable to get valid login request url from cache, redirecting to home page");
-                window.location.href = "/";
+                window.location.assign("/");
                 return;
-            } else if (currentUrl !== loginRequestUrl) {
-                // If loginRequestUrl contains a hash (e.g. Angular routing), process the hash now then redirect to prevent both hashes in url
-                if (loginRequestUrl.indexOf("#") > -1) {
-                    this.logger.info("loginRequestUrl contains hash, processing response hash immediately then redirecting");
-                    this.processCallBack(hash, stateInfo, null);
```
jasonnutter, line 1120: I'm interested as to why we don't need to call `processCallback` here anymore, can you provide more detail here?

tnorling: Yeah so before we processed the token hash and then redirected to the `loginStartPage` with, what we'll call, the user hash. Now instead redirection only occurs when host/path doesn't match. If it does we will only change the hash of the page to the user hash instead of doing a full redirect, and then process the token hash (which was extracted at the beginning of the function in both scenarios)

tnorling: Essentially I changed the logic to pull apart the `loginStartPage` into the different parts and make sure the final page matches, while also making sure we only redirect when we need to and the token hash gets processed on the final page

jasonnutter: Okay, so the process for this scenario (redirect start page includes a non-msal hash) is:

1. Redirect page loads with MSAL hash, navigateToLoginRequestUrl is true, loginRequestUrl is set with non-msal hash
2. MSAL hash is parsed and stored
3. Check to see if current url is different from loginRequest url, which would be false in this scenario since redirect uris cant contain hashes.
4. Since it is false, redirect to loginRequestUrl, which is where the stored MSAL hash is processed.

Correct?

tnorling: Not exactly. The flow would go more like this: 

1. Redirect page loads with MSAL hash, navigateToLoginRequestUrl is true, loginRequestUrl is set with non-msal hash
2. Strip msal hash, hold in local variable
3. Check if current url is different from loginRequest url, not taking into account any hash or query string param in either url
4. If the previous check is false, redirect to the loginRequestUrl (without non-msal hash/query string) + msal hash
5. Now step 3 is true, no more redirect required
6. Check cached loginRequestUrl for non-msal hash or query string. If present, add these onto the current window url
7. Process and store msal hash from step 2

Happy to go into this in more detail on Wednesday if needed.

jasonnutter: Ah okay. Didn't see the `processCallback` at the bottom of the function. Tested this with the Angular sample with hash routing and redirects, and worked as expected.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1111,18 +1111,22 @@ export class UserAgentApplication {
...
+                    window.location.assign(`${finalRedirectUrl}${hash}`);
+                    return;
                 } else {
-                    window.location.href = `${loginRequestUrl}${hash}`;
+                    const loginRequestUrlComponents = UrlUtils.GetUrlComponents(loginRequestUrl);
+                    if (loginRequestUrlComponents.Hash){
+                        window.location.hash = loginRequestUrlComponents.Hash;
+                    }
+                    if (loginRequestUrlComponents.Search){
+                        window.location.search = loginRequestUrlComponents.Search;
```
jasonnutter, line 1127: Changing `window.location.search` actually initiates a browser navigation, while changing `window.location.hash` does not. I think you may want to put `processCallback` before you change `search`, to prevent a race condition?

jasonnutter: In other words, the order in which you call process hash versus changing the `window.location.hash` does not matter, but it technically matters if you are changing `window.location.search`, since it could navigate away from the page completely. I think there should be separate behavior depending on if there is a query parameter, instead of relying on `this.processCallback` at the end of the function.

tnorling: Oh interesting. I didn't experience another navigation in my testing. Is this only in some scenarios? I could just move the url manipulation block underneath the `processCallback` call. That makes sense to me.

PR #1609, opened 2020-05-06
[msal-node][msal-common] Cache Lookup 1: PR defining lookup interface

sameerag:
This PR moves around all cache code to a common folder `cache` and:
* Adds look up interface
* Adds `Credential` APIs to optimize entity based key generation
* All current `cache` code is moved to `spacache` under `cache` folder. This will eventually go away as I submit more PRs on `cache lookup`


sameerag:
Changes ported to other PRs. Closing this.

PR #1617, opened 2020-05-07
[msal-common] ADFS 2019 support

sangonzal:
Notes from taken from [python implementation](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/77) and email thread with AzureStack team: 
- ADFS does not have concept of tenant id. Instead, path will always contain "adfs", and should contain nothing aftewards. We should consider validating url to either throw error or format authority to delete everything after "adfs". 
- ADFS supports OpenId configuration endpoint, but not the instance discovery endpoint. OpenId configuration endpoint is formatted differently than AAD and B2C. (there is no 'v2' in path)
- Only ADFS 2019 works with MSALs scope format. Earlier versions will not. This behavior must be enabled manually by customer. It's common that customers think that AFDS is on server 2019, but it might not actually have internally upgraded it's runitime to support the newer flow. To check if ADFS supports MSAL, check for "nbf" claim in OpenId configuration endpoint. We should consider checking for this claim and throwing an error if not found. 

Cache: 
- Tenant id is null, and therefore HomeAccountId should just use `oid` (which is subject claim from IdToken)
- Id token from ADFS 2019 contains "upn", but not preferred_username
- Client info is not available from ADFS 2019 response


sangonzal:
Two open issues for msal-browser: 
- Need to get ADFS 2019 to support CORS 
- browser's cache relies on ClientInfo returned in the response. Instead of updating current cache design, we will wait until it is moved to new cache.

PR #1622, opened 2020-05-10
[msal-node]Cache-6: Read and write to disk, merge cache

sameerag:
This PR has the following functionality added to the cache:

* Reads and writes asynchronously from the disk
* Introduces `CacheContext` which manages reading and writing into a cache file and merging the current `cache` with stored `cacheBlob` while doing so.
### Todo:
Will have to make sure during file write/read, the accessed file is locked. This logic should be mostly in extension library, will update once we reach a consensus.

CacheSync


lib/msal-common/test/client/AuthorizationCodeClient.spec.ts
```diff
@@ -126,7 +127,7 @@ describe("AuthorizationCodeClient unit tests", () => {
 
     describe("Acquire a token", () => {
 
-        it("Acquires a token successfully", async () => {
+        xit("Acquires a token successfully", async () => {
```
pkanher617, line 130: typo: xit

sameerag: `xit` skips a test case. I need to modify this to get it to working.

pkanher617: my mistake.

PR #1624, opened 2020-05-11
[msal-node] Cache Lookup - 1: Logical keys for cache entities

sameerag:
This PR generates `AccountId` and `CredentialId` as a part of the `CredentialEntity` and `AccountEntity` which will be later used in looking up cache elements as needed.

#### Schema
`Account` and `Credential` cache has the below schema for keys:

| Entity type | Key components                                               |
| ----------- | ------------------------------------------------------------ |
| Account     | <home_account_id>-< environment>-<realm*>                     |
| Credential  | <home_account_id*>-< environment>-<credential_type>-<client_id>-<realm*>-<target*> |

But to tie up a `Credential` with `Account`, the keys are logically divided as follows:

#### Logical key components
| Field         | Format                                | Applicable to          |
| ------------- | ------------------------------------- | ---------------------- |
| account_id    | <home_account_id>-< environment>       | Account <br>Credential |
| realm             | < realm>                               | Account                |
| credential_id | <credential_type>-<client_id>-< realm> | Credential             |
| target        | < target>                              | Credential             |


lib/msal-common/src/unifiedCache/entities/AccountEntity.ts
```diff
@@ -25,17 +25,31 @@ export class AccountEntity {
...
+    generateAccountId(): string {
+        const accountId: Array<string> = [this.homeAccountId, this.environment];
+        return accountId.join(Separators.CACHE_KEY_SEPARATOR).toLowerCase();
+    }
+
+    /**
+     * Generate Account Id key component as per the schema: <home_account_id>-<environment>
+     */
+    generateRealm(): string {
+        return (this.realm || "").toLowerCase();
```
sangonzal, line 40: I think empty space is used in multiple places, consider adding it to the constants file

pkanher617: +1 for EMPTY_STRING constant

lib/msal-common/src/unifiedCache/entities/Credential.ts
```diff
@@ -3,13 +3,62 @@
...
     secret: string;
-};
+    familyId?: string;
+    realm?: string;
+    target?: string;
+
+    /**
+     * Generate Account Id key component as per the schema: <home_account_id>-<environment>
+     */
+    generateAccountId(): string {
```
pkanher617, line 24: Can these be used as utils for cache lookup (i.e. generate keys for looking up a specific cred)?

sameerag: So generating accountId for an account object instance should be separate from generating it from a set of values. I will try coming up with common code as needed, for now I added static helpers in `CacheHelper.ts` in a later PR to address this. Point noted though.

Issue #1241, opened 2020-01-24
acquireTokenSilent() doesn't get new ID token


navzam:
## Library
- [x] `msal@1.2.0`
- [ ] `@azure/msal-browser@2.x.x`
- [ ] `@azure/msal-angular@0.x.x`
- [ ] `@azure/msal-angular@1.x.x`
- [ ] `@azure/msal-angularjs@1.x.x`

## Description
If I call `acquireTokenSilent()` with the `openid` scope after the cached token has expired, the result contains a new access token and the expired ID token. I don't see a new ID token anywhere in the result.

The request to the `/authorize` endpoint has `response_type=token` (not `response_type=id_token token`) so I assume AAD isn't giving back an ID token.

## Regression
Not tested on other versions of MSAL.

## Configuration
MSAL configuration options:

```js
{
  auth: {
    clientId: AAD_APP_ID,
    redirectUri: AAD_REDIRECT_URI,
  },
  cache: {
    storeAuthStateInCookie: true,
  },
}
```

## Reproduction steps
1. Call `acquireTokenSilent()` for the first time:

```js
let authTokenResult = await myMSALObj.acquireTokenSilent({
  scopes: ['user.read', 'openid'],
  loginHint: USER_UPN,
});

// Take note of authTokenResult.idToken.rawIdToken and authTokenResult.idToken.expiration
```

2. Wait for the cached access and ID tokens to expire (usually > 1 hr).
3. Call `acquireTokenSilent()` again with the same parameters.
4. Observe that `authTokenResult.idToken.rawIdToken` and `authTokenResult.idToken.expiration` are the same as before, and there is no new ID token.

## Expected behavior
I expected to have a valid ID token after calling `acquireTokenSilent()` with the `openid` scope.

## Browsers
Tested on Edge (81.0.396.0) and Chrome (79.0.3945.130)

arsnyder16:
Seeing this issue as well. As a workaround for now i am always passing ` forceRefresh: true` when i need an `idToken`

arunprasathv:
How do you call acquireTokenSilent when token expired. Do you call it from failure() ?

GuillaumeBAECHLER:
I was having the same issue and using ```forceRefresh:true``` didn't seem to work. I ended up only refreshing the id_token by calling ```acquireTokenSilent``` and passing the clientId as the only scope in the Authentication Parameters.

arunprasathv:
I did the same too.

[...]

PR #1628, opened 2020-05-11
[msal-core] Scopes-based response type

technical-boy:
## Changes

This PR addresses  and redefines the logic through which `acquireToken` and `login` APIs determine what `response_type` to set in a server request.

- The main change this PR makes is in the update to `UserAgentApplication`'s `getTokenType()` instance method.
    - The method signature was updated to remove the `silentCall` boolean attribute, given that it is no longer necessary.
    - The method contents were replaced by a simplified conditional flow that directly implements the Response Types table in the `docs/response-types.md` documentation file.

- Changes were also made to references/calls to `getTokenType` within `UserAgentApplication.ts` to remove the `silentCall` parameter from the calls.

- Another significant addition is the documentation of the behavior and configuration of the `response_type` parameter in the server requests.

These changes are contingent on the general update to the scopes validation process that is proposed in #1681.

Fixes: #1241 
Base PR: #1681


tnorling:
I think the solution needs to be option 2 for the simple reason that it's possible for people to include openid along with other scopes. We should probably do a hybrid approach where if openid and profile are the only scopes requested we return an id_token but if other scopes are also requested we return id_token_token. Returning only an id_token in cases where openid is included in the requested scopes may lead to unexpected behavior. Another thing to note, when you include the client_id as a scope it needs to be the only scope (msal will throw an error otherwise), in this case we know that we are requesting an id_token and only an id_token.

technical-boy:
@tnorling Yes, I agree the second alternative is ideal and seems the most intuitive. We'd just have to extend the logic to consider more than two cases (like checking for the 'openid' 'profile' combination.

lib/msal-core/src/utils/RequestUtils.ts
```diff
@@ -75,13 +77,40 @@ export class RequestUtils {
...
+        if (!request.scopes) {
+            request.scopes = [];
+        }
+
+        // Append clientId to scopes by default for login calls
+        request.scopes = ScopeSet.appendScopes(request.scopes, [clientId]);
+
+        // validate request
+        const userRequest: AuthenticationParameters = RequestUtils.validateRequest(request, clientId, interactionType);
+        userRequest.scopes = ScopeSet.appendScopes(userRequest.scopes, userRequest.extraScopesToConsent);
```
pkanher617, line 102: This was being called before validateInputScopes before, is now being called afterwards - is that ok? Would have failed the client id check otherwise, but since that was removed might be better to move it before `validateRequest`.

lib/docs/token-return-types.md
```diff
@@ -0,0 +1,10 @@
+# Token Return Types
```
sameerag, line 1: Can you please add some text explaining why the table is the way it is. Add what sending `openid`, `profile` would mean and what sending `client_id` as `scope` was; how both these concepts are now addressed with this new PR.

technical-boy: Yes, should get that done today or tomorrow at the latest!

lib/docs/token-return-types.md
```diff
@@ -0,0 +1,71 @@
...
+| **id_token** | Indicates a request will be made for an ID Token |
+| **id_token token** | Indicates that both an ID Token and Access Token will be requested |
+
+The following table describes how MSAL JS v1 determines which token request type is set and therefore what kind of token is returned by the API in question. 
+
+| Scenarios | ClientId as Only Scope | No Scopes | Access Token Scopes with No ClientId | Access Token Scopes with ClientId
+| ------- | ------- | -------- | --------- | -------- |
+| **loginRedirect/Popup** | id_token | id_token |id_token token | id_token token
+| **acquireTokenInteractive** | id_token | Error | token | id_token token 
+| **acquireTokenSilent** | id_token | Error | token | id_token token
```
sameerag, line 17: cc @hamiltonha Once `error` doc is ready, please link to the exact error thrown here. Can you still update if the errors are distinct? If not leave it be until @hamiltonha's doc is ready.

lib/docs/token-return-types.md
```diff
@@ -0,0 +1,71 @@
...
+| **loginRedirect/Popup with accounts not matching** | id_token | id_token | id_token token | id_token token
+| **acquireTokenInteractive with accounts not matching** | id_token | Error | id_token token | id_token token
+| **acquireTokenSilent with accounts not matching** | id_token | Error | id_token token | id_token token
+
+As described in the table above, the Token Request Type is determined based on specific combinations of the following attributes:
+
++ `Scenario`: Each row is a combination of the public API that is called in the client application and whether or not the account passed in matches the account in the MSAL cache.
```
sameerag, line 24: The `account` is not always passed. Since we do not support multiple accounts in `msal@1.x.x`, we do look up the current `account` in the cache, when nothing is passed. Please verify this and update the statement accordingly.

lib/docs/token-return-types.md
```diff
@@ -0,0 +1,71 @@
...
+
+```js
+    const request = { scopes: ['user.read', 'APP_CLIENT_ID_VALUE'] };
+```
+
+### Login Scopes Considerations
+
+Within MSAL JS v1, `openid` and `profile` are referred to as "login scopes" or OIDC (OpenID Connect) scopes. From the library's perspective, setting the client ID value as a scope in the request configuration object is functionally equivalent to adding 'openid', 'profile', or both scopes. 
+
+This is because during request validation, if the client ID value is present in the scopes, it is removed from the array and replaced with both 'openid' and 'profile'.
```
sameerag, line 57: `is replaced by`

lib/docs/token-return-types.md
```diff
@@ -0,0 +1,71 @@
...
+
+Similarly, whenever 'openid' or 'profile' is present without the other in the scopes array, the missing login scope will be added during request validation and before the request is actually sent to the server.
+
+This means that:
+
++ `scopes: ['APP_CLIENT_ID_VALUE']`
++ `scopes: ['openid']`
++ `scopes: ['profile']`
+
+Are all functionaly equivalent to `scopes: ['openid', 'profile']` when set in the request configuration object before passing it in as a parameter to any of UserAgentApplication's public APIs.
```
sameerag, line 67: simplify: All the above will be treated as same request `scopes: ['openid', 'profile']`

lib/docs/token-response-types.md
```diff
@@ -0,0 +1,74 @@
...
+
+    Similarly, whenever `openid` or `profile` is present without the other in the scopes array, the missing login scope will be added during request validation and before the request is actually sent to the server.
+
+    This means that:
+
+    + `scopes: ['APP_CLIENT_ID_VALUE']`
+    + `scopes: ['openid']`
+    + `scopes: ['profile']`
+
+    Are all functionaly equivalent to `scopes: ['openid', 'profile']` when set in the request configuration object before passing it in as a parameter to any of UserAgentApplication's public APIs.
```
jo-arroyo, line 70: nit: functionaly -> functionally

lib/docs/token-response-types.md
```diff
@@ -0,0 +1,74 @@
...
+
++ On a related note, `login` APIs (i.e. `loginPopup` and `loginRedirect`) will append 'openid' and 'profile' (without duplication) to the scopes array in every call, no matter what other scopes are present. This means that `login` APIs may be called with an empty scopes array in the request configuration object.
```
jo-arroyo, line 72: nit: `openid` and `profile` instead of 'openid' and 'profile', per rest of doc?

lib/docs/token-response-types.md
technical-boy: Good catch, corrected!

lib/docs/token-response-types.md
```diff
@@ -0,0 +1,74 @@
+# Token Response Types
```
tnorling, line 1: Instead of creating a docs folder in /lib can we put this in /lib/msal-core if this is going to be core specific or in /docs if we want to have a top level folder for docs that apply to all libraries

jasonnutter: Agree, I would put this in `lib/msal-core/docs` to start.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -2149,39 +2149,21 @@ export class UserAgentApplication {
...
+        // Returns response type "token" if request has no login scopes and the account passed matches the current account object
+        if (!ScopeSet.isLoginScopes(scopes, this.clientId) && Account.compareAccounts(accountObject, this.getAccount())) {
+            return ResponseTypes.token;
         }
 
+        /**
+         * Returns response type "id_token token" if both login scopes and access token scopes are present
+         * Returns response type "id_token" if only login scopes are present
+         */
+        return (scopes.length > 2) ? ResponseTypes.id_token_token : ResponseTypes.id_token;
```
jasonnutter, line 2166: This assumes `openid` and `profile` are always included, correct? I would add a comment making that clear (or calculating the length without those two scopes).

lib/docs/token-response-types.md
technical-boy: Yes, good call, I'll move it!

lib/msal-core/src/UserAgentApplication.ts
technical-boy: Correct, I've made sure any login scopes will already be in the array by the time this method even gets called. I'll add the comment.

lib/docs/token-response-types.md
```diff
@@ -0,0 +1,74 @@
...
+
+| Response Type | Definition | Reference | Request Example |
+| ------- | -------- | -------- | -------- |
+| **token** | Indicates a request will be made for an Access Token | [OAuth 2.0 Implicit Flow Specification](https://tools.ietf.org/html/rfc6749#section-4.2.1) | [Request example](https://tools.ietf.org/html/rfc6749#section-4.2.1)
+| **id_token** | Indicates a request will be made for an ID Token | [OpenID Connect Specification](https://openid.net/specs/openid-connect-core-1_0.html#Authentication) | [Request Example](https://openid.net/specs/openid-connect-core-1_0.html#id_tokenExample)
+| **id_token token** | Indicates that both an ID Token and Access Token will be requested | [OpenID Connect Specification](https://openid.net/specs/openid-connect-core-1_0.h:tml#Authentication) | [Request Example](https://openid.net/specs/openid-connect-core-1_0.html#id_token-tokenExample)
+
+The table below describes the mapping of `msal@1.x.x` APIs in combination with the user-provided `scopes` to the resulting `response_type` set in the server request.
+
+| Scenarios | Client ID as Only Scope | No Scopes | Access Token Scopes with No client ID | Access Token Scopes with client ID
```
tnorling, line 13: Instead of singling out Client ID here can we define what a "Login Scope" is above and then replace any reference to Client ID with "Login Scopes"?

lib/docs/token-response-types.md
```diff
@@ -0,0 +1,74 @@
...
+| ------- | ------- | -------- | --------- | -------- |
+| **loginRedirect/Popup** | id_token | id_token |id_token token | id_token token
+| **acquireTokenInteractive** | id_token | Error | token | id_token token 
+| **acquireTokenSilent** | id_token | Error | token | id_token token
+| **loginRedirect/Popup with accounts not matching** | id_token | id_token | id_token token | id_token token
+| **acquireTokenInteractive with accounts not matching** | id_token | Error | id_token token | id_token token
```
tnorling, line 19: Can we call this `acquireTokenRedirect/Popup` for more clarity

lib/docs/token-response-types.md
```diff
@@ -0,0 +1,74 @@
...
+
+    This means that:
+
+    + `scopes: ['APP_CLIENT_ID_VALUE']`
+    + `scopes: ['openid']`
+    + `scopes: ['profile']`
+
+    Are all functionally equivalent to `scopes: ['openid', 'profile']` from the library's perspective.
+
++ On a related note, `login` APIs (i.e. `loginPopup` and `loginRedirect`) will append `openid` and `profile` (without duplication) to the scopes array in every call, no matter what other scopes are present. This means that `login` APIs may be called with an empty scopes array in the request configuration object.
```
tnorling, line 72: Update this based on discussion to specify all calls will include openid and profile but that return type will be based on what was provided except for loginPopup and loginRedirect which will always be of type `id_token`

lib/msal-core/src/UserAgentApplication.ts
technical-boy: Nevermind, I'm unsure about all cases being covered here. I'll spend some more time making sure this is right and fixing it if not. Thanks @jasonnutter!

lib/docs/token-response-types.md
technical-boy: Good call

lib/docs/token-response-types.md
technical-boy: Sure

lib/msal-core/src/UserAgentApplication.ts
technical-boy: @jasonnutter I did a lot of work to summarize the response types table in the doc, there was some repetition. After that, I made sure every single case was covered by the code (it wasn't). The new code, although a little more verbose, covers every case in what I believe to be the least amount of checks possible. I also added comments explaining the logic to make sure the intention isn't lost in the future (making sure to answer "why" over "what". Thanks again for catching this, I think I'll need to add some tests for this behavior at the public APIs level.

lib/docs/token-response-types.md
technical-boy: @tnorling done, I did some major changes to try to make the concepts more clear and better organized, making sure this equivalence was emphasized. Hopefully the way I wrote it won't add to the confusion.

lib/docs/token-response-types.md
technical-boy: @tnorling I removed the acquireTokenSilent rows because they matched popup and redirect in every case, so I went for 'acquireTokenX' in the new version. What do you think?

lib/docs/token-response-types.md
technical-boy: Done. I think I've made that distinction clear in the **Important Note**. Please let me know if it's too verbose or too ambiguous!

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -2149,39 +2149,31 @@ export class UserAgentApplication {
...
-        if (silentCall) {
-            if (Account.compareAccounts(accountObject, this.getAccount())) {
-                tokenType = (scopes.indexOf(this.config.auth.clientId) > -1) ? ResponseTypes.id_token : ResponseTypes.token;
-            }
-            else {
-                tokenType  = (scopes.indexOf(this.config.auth.clientId) > -1) ? ResponseTypes.id_token : ResponseTypes.id_token_token;
-            }
-            return tokenType;
+        // response_type is id_token because only login scopes are present (["openid", "profile"].length == 2);
+        if (containsLoginScopes && (scopes.length == 2)) {
```
tnorling, line 2163: I would rather explicitly check that the scopes array only contains `openid` and `profile` to avoid a buggy scenario where scopes list is ["User.Read", "openid"]

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -541,7 +541,7 @@ export class UserAgentApplication {
             }
 
             // On Fulfillment
-            const responseType: string = isLoginCall ? ResponseTypes.id_token : this.getTokenType(account, request.scopes, false);
+            const responseType: string = this.getTokenType(account, request.scopes);
```
tnorling, line 544: Why did we get rid of the `isLoginCall` logic here?

lib/msal-core/docs/token-response-types.md
```diff
@@ -0,0 +1,90 @@
...
+
+***Important Note: `msal@1.x.x` will append `openid` and `profile` to the `scopes` parameter of every server request URL, whether or not any of these two specific scopes are passed by the developer into the API call. The presence of login scopes in the configuration object has an effect on which `response_type` value is set in the actual server request `msal1.x.x` makes, but it does not make a difference on the value of the `scopes` parameter in said server request.***
+
+The Scopes-based Response Types table below describes the mapping of `msal@1.x.x` APIs in combination with the user-provided `scopes` to the resulting `response_type` set in the server request.
+
+## Scopes-based Response Types
+
+|| Account Object in Request Configuration | Empty Scopes Array | Login Scopes Only | Access Token Scopes Only | Access Token and Login Scopes
+| -------- | ------- | ------- | -------- | --------- | -------- |
+| **loginRedirect/Popup** | **None or Matching Account in MSAL Cache** | id_token | id_token |id_token token | id_token token
```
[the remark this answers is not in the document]
technical-boy: @pkanher617 It's been a while since we made the original table but I believe there was a reason why we thought login calls should allow access token scopes and set response type to id_token_token. Is there a need for a discussion here or should this be changed as Thomas suggests?

lib/msal-core/src/UserAgentApplication.ts
technical-boy: Given the way #1681 validates scopes, there is no way this code is reached with either login scope without appending the other. Adding that explicit check feels like repeating validation, which I believe should be a responsibility of the API rather than this method. The bug in question would only occur if the validation behavior changes. Having said this, if you still think having the explicit check is necessary, I have no other reason to disagree.

lib/msal-core/src/UserAgentApplication.ts
technical-boy: The idea was to make the entire decision based on the scopes and not which API was called. We couldn't remove the account factor from the decision because we need to support the specific use case of the account passed in being different from the one in the cache. We were able to remove isLoginCall by having the API validate and manage the scopes set before this point is reached, which shifts the responsibility from the `isLoginCall` boolean to the state of the scopes.

lib/msal-core/docs/token-response-types.md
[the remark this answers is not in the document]
tnorling: To follow up: This would be a breaking change, because today you can request multiple resources in an `id_token` request in order to pre-consent to the scopes. Having this return an access token as well would result in an error if scopes spanning more than 1 resource are requested.

@pkanher617 Can correct me if I'm wrong, but I believe Auth Code is allowing `id_token token` requests on a login call. But I don't think this change should be made in core

lib/msal-core/src/UserAgentApplication.ts
tnorling: Since this is a utils function I think it's a good idea to not assume anything has been done before it. If you think the logic is duplicated I would suggest writing a helper function you can call in both places.

lib/msal-core/docs/token-response-types.md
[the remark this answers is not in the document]
pkanher617: We have always allowed access tokens to be returned on login calls. We do allow users to request multiple resources, but generally we have restricted that through the use of extraScopesToConsent. 

@tnorling `response_type` is always `=code` in 2.0, which allows for multiple resources (doesn't matter if it is a login call or not). 

id_token token will still accept multiple resources as well, but also return an access token for the first resource (I'm not sure how this determination is made). For example, if I send `response_type=id_token token` and `scope=openid profile User.Read https://resource2/mail.read` I will get an id_token and an access token for Graph and still see a consent screen for the additional scope.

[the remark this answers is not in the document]
tnorling: > We have always allowed access tokens to be returned on login calls. 

I don't think this is true. Currently we set `response_type` to `id_token` on all login calls, regardless of what scopes are requested. We can _consent_ to scopes, for up to 3 resources, via a login call but it will not return an access_token. You can verify this by looking at the diff on this PR. Our messaging has been consistent that you need to login to get an id_token and then call an acquireToken method to get an access token. I don't think we should change this behavior.

I don't disagree with the statement that `id_token token` will return an access token while still allowing you to consent to other scopes, but I think we should discourage this outside of using `extraScopesToConsent` because it could lead to some confusing behaviors if someone is: A. getting only a single access token and B. Not understanding which resource you got the access token for and why

lib/msal-core/docs/token-response-types.md
```diff
@@ -0,0 +1,90 @@
...
+
+    This means that:
+
+    + `scopes: ['APP_CLIENT_ID_VALUE']`
+    + `scopes: ['openid']`
+    + `scopes: ['profile']`
+
+    Are all functionally equivalent to `scopes: ['openid', 'profile']` from the library's perspective. From this point on, when this document makes a reference to "login scopes", it refers to any combination or permutation of `openid`, `profile` and the value of `client ID`. The term "access token scopes", on the other hand, refers to any resource-related scope set that an access token is being requested for (i.e. 'user.read', 'mail.read', etc.). 
+
++ On a related note, `login` APIs (i.e. `loginPopup` and `loginRedirect`) will append `openid` and `profile` (without duplication) to the scopes array every time they are called, no matter what other scopes are present. This means that `login` APIs can be called with an empty, non-null scopes array in the request configuration object (as opposed to acquireToken APIs which will throw an error with an empty or null `scopes` array in the request).
```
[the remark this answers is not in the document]
technical-boy: Should be clearer and less redundant in the updated version. I'm finding it hard to make this point overwhelmingly clear without being too repetitive.

lib/msal-core/docs/token-response-types.md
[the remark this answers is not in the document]
technical-boy: I believe the most recent changes to the code and documentation have resolved this conversation based on the consensus reached in our last design discussion.

lib/msal-core/src/UserAgentApplication.ts
technical-boy: I re-added the ternary check for `isLoginCall`. There's no way to tell if it's a login call based on scopes only, seems like this was a moot effort, but I still think the new version of this code is an improvement. Thanks!

lib/msal-core/src/UserAgentApplication.ts
technical-boy: Added a new static method to ScopeSet called `containsOIDCScopes` that does a hard check for both `openid` and `profile`, replaced the call to `isLoginScopes` here with a call to `containsOIDCScopes`. Now the possible bug you mentioned should be mitigated. Thanks!

lib/msal-core/docs/response-types.md
```diff
@@ -0,0 +1,152 @@
...
+
+A URL encoded string that `msal@1.x.x` generates based on the server request **configuration object** and uses to obtain your access or ID token from the server:
+
+Example: 
+
+`https://login.microsoftonline.com/common/oauth2/v2.0/authorize?response_type=token&scope=s1%20openid%20profile&client_id=YOUR_CLIENT_ID&redirect_uri=YOUR_REDIRECT_URI&state=1234&nonce=NONCE_VALUE&client_info=1&x-client-SKU=MSAL.JS&x-client-Ver=1.3.2&client-request-id=CLIENT_REQUEST_ID_VALUE`
+
+### Implications
+
+In summary, the server request **configuration object** is what `msal@1.x.x` uses (in the Implicit Flow) to determine the contents of the actual server request **URL** string that makes up the `GET` request sent to the server. In relation to scopes and response types, this has the following implications:
```
sameerag, line 55: `Regarding` instead of `In Relation`

technical-boy: Changed.

PR #1629, opened 2020-05-11
Extend acquireTokenSilent Instrumentation

jo-arroyo:
This PR adds instrumentation through logger messages to acquireTokenSilent for msal-core.


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -757,12 +779,15 @@ export class UserAgentApplication {
             }
         })
             .then(res => {
+                this.logger.verbose("Successfully acquired token");
                 this.telemetryManager.stopAndFlushApiEvent(request.correlationId, apiEvent, true);
+                this.logger.info("Stopping telemetryManager and flushing API event");
```
jasonnutter, line 784: I'm working in the telemetry manager now, and will add these sorts of logs there, so you can skip these here.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -708,27 +724,33 @@ export class UserAgentApplication {
...
 
-                // start http event
+                // start http event TODO: add why
                 return serverAuthenticationRequest.authorityInstance.resolveEndpointsAsync(this.telemetryManager, request.correlationId)
                     .then(() => {
                         /*
                          * refresh attempt with iframe
                          * Already renewing for this scope, callback when we get the token.
                          */
+                        this.logger.verbose(`Updated authority: ${JSON.stringify(serverAuthenticationRequest.authorityInstance)}`);
```
jasonnutter, line 751: How big is this object? A more succinct might be better.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -708,27 +724,33 @@ export class UserAgentApplication {
...
                 if (!serverAuthenticationRequest.authorityInstance) {
                     serverAuthenticationRequest.authorityInstance = request.authority ? AuthorityFactory.CreateInstance(request.authority, this.config.auth.validateAuthority) : this.authorityInstance;
+                    if (!request.authority){
+                        this.logger.verbose("Using default authority since no authority was passed to API");
+                    }
                 }
                 // cache miss
 
-                // start http event
+                // start http event TODO: add why
```
jasonnutter, line 744: I'm working on this and can explain provide the context.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -668,23 +678,29 @@ export class UserAgentApplication {
                 request.correlationId,
             );
 
+            this.logger.verbose(`ServerAuthenticationRequest: ${JSON.stringify(serverAuthenticationRequest)}`);
```
jasonnutter, line 681: I think providing succinct, specific messages will be more helpful than stringify entire objects. What sort of specific pieces of information might helpful to know here?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -630,32 +630,42 @@ export class UserAgentApplication {
...
 
             const scope = request.scopes.join(" ").toLowerCase();
+            this.logger.verbose(`Serialized scopes: ${scope}`);
 
             // if the developer passes an account, give that account the priority
             const account: Account = request.account || this.getAccount();
+            this.logger.verbosePii(`Set account to: ${JSON.stringify(account)}`);
 
-            // extract if there is an adalIdToken stashed in the cache
+            // extract if there is an adalIdToken stashed in the cache TODO: Ask about why
```
jasonnutter, line 652: This allows MSAL to SSO to an existing session that was started with ADAL and upgrade it to an MSAL session. This allows for seemless SSO when upgrading an app from ADAL -> MSAL.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -630,32 +630,42 @@ export class UserAgentApplication {
...
         const requestSignature = RequestUtils.createRequestSignature(request);
 
         return new Promise<AuthResponse>((resolve, reject) => {
 
-            // block the request if made from the hidden iframe
+            // block the request if made from the hidden iframe TODO: Ask about why to extend comment
             WindowUtils.blockReloadInHiddenIframes();
 
             const scope = request.scopes.join(" ").toLowerCase();
+            this.logger.verbose(`Serialized scopes: ${scope}`);
```
jasonnutter, line 646: I believe this should be PII since it could contain client ID.

lib/msal-core/src/UserAgentApplication.ts
jo-arroyo: Removed.

lib/msal-core/src/UserAgentApplication.ts
jo-arroyo: Updated.

lib/msal-core/src/UserAgentApplication.ts
jo-arroyo: Changed logger to no longer print object but simply state that authority has been updated

lib/msal-core/src/UserAgentApplication.ts
jo-arroyo: Agreed. Changed logger to no longer print out object, but simply state that a server authentication request was built.

lib/msal-core/src/UserAgentApplication.ts
jo-arroyo: Updated comment to reflect the above.

lib/msal-core/src/UserAgentApplication.ts
jo-arroyo: Removed TODOs for now. Will revisit when working on code comments.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -630,6 +630,8 @@ export class UserAgentApplication {
      *
      */
     acquireTokenSilent(userRequest: AuthenticationParameters): Promise<AuthResponse> {
+        this.logger.info("AcquireTokenSilent has been called");
+        this.logger.verbose(`AcquireTokenSilent params:\n${JSON.stringify(userRequest)}`);
```
jo-arroyo, line 634:
```suggestion
```

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -680,11 +697,14 @@ export class UserAgentApplication {
...
             }
 
+            this.logger.verbose("Query parameters populated");
+
             const userContainedClaims = request.claimsRequest || serverAuthenticationRequest.claimsValue;
 
             let authErr: AuthError;
             let cacheResultResponse;
 
+            // If request.forceRefresh is set to true, we want to force a request for a new token instead of getting it from the cache
```
jo-arroyo, line 707:
```suggestion
            // If request.forceRefresh is set to true, force a request for a new token instead of getting it from the cache
```

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -641,21 +642,34 @@ export class UserAgentApplication {
...
+             * In the event of no account being passed in the config, no session id, and no pre-existing adalIdToken, user will need to log in
+             */
             if (!account && !(request.sid  || request.loginHint) && StringUtils.isEmpty(adalIdToken) ) {
                 this.logger.info("User login is required");
+                // The promise rejects with a UserLoginRequiredError, which should be caught and user should be prompted to log in interactively
                 return reject(ClientAuthError.createUserLoginRequiredError());
             }
 
             // set the response type based on the current cache status / scopes set
             const responseType = this.getTokenType(account, request.scopes, true);
```
pkanher617, line 671: May be worth adding logging inside this function as well.

jasonnutter: They are just focusing on the `acquireTokenSilent` function for this PR. They will incrementally add logs for rest of UAA in future PRs.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -630,6 +630,7 @@ export class UserAgentApplication {
      *
      */
     acquireTokenSilent(userRequest: AuthenticationParameters): Promise<AuthResponse> {
+        this.logger.info("AcquireTokenSilent has been called");
```
jasonnutter, line 633: I think logs like this at the start of APIs should be verbose.

sameerag: Or at-least highlight what is called, `request` specifics if it helps with log analysis, not just the API name. Depending on PII settings ofcourse.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -641,21 +642,34 @@ export class UserAgentApplication {
...
 
             // if the developer passes an account, give that account the priority
-            const account: Account = request.account || this.getAccount();
+            let account: Account;
+            if (request.account) {
+                account = request.account;
+                this.logger.verbose("Account set from request");
+            } else {
+                account = this.getAccount();
+                this.logger.verbose("Account set from MSAL configuration");
```
sameerag, line 654: Can we please rephrase this as `Account set from MSAL Cache`?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -708,17 +727,20 @@ export class UserAgentApplication {
             else {
                 let logMessage;
                 if (userContainedClaims) {
-                    logMessage = "Skipped cache lookup since claims were given.";
+                    logMessage = "Skipped cache lookup since claims were given";
                 } else if (request.forceRefresh) {
                     logMessage = "Skipped cache lookup since request.forceRefresh option was set to true";
                 } else {
-                    logMessage = "Token is not in cache for scope:" + scope;
+                    logMessage = "Token is not in cache for scope: " + scope;
```
sameerag, line 734: Why is the `if` case `verbose` and `else` case `info` for logs here?

jo-arroyo: @sameerag This is unclear. These logs were already here, we just changed the punctuation. Do you feel that info or verbose is more appropriate?

lib/msal-core/src/UserAgentApplication.ts
jo-arroyo: Noted, will add logs when we get to those functions.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -641,21 +642,34 @@ export class UserAgentApplication {
...
+                this.logger.verbose("Account set from MSAL configuration");
+            }
 
-            // extract if there is an adalIdToken stashed in the cache
+            // Extract adalIdToken if stashed in the cache to allow for seamless ADAL to MSAL migration
             const adalIdToken = this.cacheStorage.getItem(Constants.adalIdToken);
 
-            // if there is no account logged in and no login_hint/sid is passed in the request
+            /**
+             * if there is no account logged in and no login_hint/sid is passed in the request
```
tnorling, line 661: Can we remove this line since it's explained again below?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -708,17 +727,20 @@ export class UserAgentApplication {
...
+                    logMessage = "Token is not in cache for scope: " + scope;
                 }
                 this.logger.verbose(logMessage);
 
-                // Cache result can return null if cache is empty. In that case, set authority to default value if no authority is passed to the api.
+                // Cache result can return null if cache is empty. In that case, set authority to default value if no authority is passed to the API.
                 if (!serverAuthenticationRequest.authorityInstance) {
                     serverAuthenticationRequest.authorityInstance = request.authority ? AuthorityFactory.CreateInstance(request.authority, this.config.auth.validateAuthority) : this.authorityInstance;
+                    if (!request.authority){
+                        this.logger.verbose("Using default authority since no authority was passed to API");
```
tnorling, line 742: Could we instead log the authority being used regardless of default or not? Can grab it from the instance created.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -745,7 +769,7 @@ export class UserAgentApplication {
                                 this.renewIdToken(requestSignature, resolve, reject, account, serverAuthenticationRequest);
                             } else {
                                 // renew access token
-                                this.logger.verbose("renewing accesstoken");
+                                this.logger.verbose("renewing access token");
```
tnorling, line 772: nit:
```suggestion
                                this.logger.verbose("Renewing access token");
```

lib/msal-core/src/UserAgentApplication.ts
jo-arroyo: Sure. Change made.

lib/msal-core/src/UserAgentApplication.ts
jo-arroyo: Good catch! Removed.

lib/msal-core/src/UserAgentApplication.ts
jo-arroyo: Thanks! Changed.

lib/msal-core/src/UserAgentApplication.ts
jo-arroyo: Chose to change to verbose for now.

lib/msal-core/src/UserAgentApplication.ts
jo-arroyo: Sure! Changed.

lib/msal-core/src/UserAgentApplication.ts
jo-arroyo: Noted. Changed in commit.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -630,6 +630,8 @@ export class UserAgentApplication {
      *
      */
     acquireTokenSilent(userRequest: AuthenticationParameters): Promise<AuthResponse> {
+        this.logger.verbose("AcquireTokenSilent has been called");
+        this.logger.verbose(`AcquireTokenSilent params:\n${JSON.stringify(userRequest)}`);
```
jasonnutter, line 634: This could include Pii. I would argue it can be removed, since relevant parts of the request are logged downstream.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -708,18 +725,20 @@ export class UserAgentApplication {
...
+                    logMessage = "Token is not in cache for scope: " + scope;
                 }
                 this.logger.verbose(logMessage);
 
-                // Cache result can return null if cache is empty. In that case, set authority to default value if no authority is passed to the api.
+                // Cache result can return null if cache is empty. In that case, set authority to default value if no authority is passed to the API.
                 if (!serverAuthenticationRequest.authorityInstance) {
                     serverAuthenticationRequest.authorityInstance = request.authority ? AuthorityFactory.CreateInstance(request.authority, this.config.auth.validateAuthority) : this.authorityInstance;
                 }
+                this.logger.verbose(`Authority instance: ${serverAuthenticationRequest.authority}`);
```
jasonnutter, line 740: This could include the client ID, so should be `verbosePii`.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -668,6 +680,8 @@ export class UserAgentApplication {
                 request.correlationId,
             );
 
+            this.logger.verbose("Finished building server authentication request");
+
             // populate QueryParameters (sid/login_hint) and any other extraQueryParameters set by the developer
             if (ServerRequestParameters.isSSOParam(request) || account) {
                 serverAuthenticationRequest.populateQueryParams(account, request, null, true);
```
jasonnutter, line 687: Can we add a log statement inside this `if`?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -680,11 +694,14 @@ export class UserAgentApplication {
                 serverAuthenticationRequest.populateQueryParams(account, null, adalIdTokenObject, true);
             }
 
+            this.logger.verbose("Query parameters populated");
```
jasonnutter, line 697: Instead of this, I would add an `else` to the `if/else if` directly above then add a statement indicating no additional query parameters added.

lib/msal-core/src/UserAgentApplication.ts
jo-arroyo: Agreed. Removed.

lib/msal-core/src/UserAgentApplication.ts
jo-arroyo: Makes sense. Changed.

lib/msal-core/src/UserAgentApplication.ts
jo-arroyo: Noted. Changes made.

lib/msal-core/src/UserAgentApplication.ts
jo-arroyo: Added log statement. Let me know if you want this statement changed.

PR #1631, opened 2020-05-11
[msal-common][msal-browser] Fix issue where config was being overwritten by undefined values

pkanher617:
This PR fixes an issue where the default system configuration in `msal-common` was being overwritten by the undefined values in `msal-browser`. This PR adds the default system values to `msal-common` exports and uses them in the `msal-browser` package.

This should fix #1623.

PR #1633, opened 2020-05-11
Release @azure/msal-browser v2.0.0-beta.2

pkanher617:
Please review #1613, #1631 and #1632 before reviewing this PR.


lib/msal-common/src/client/SPAClient.ts
```diff
@@ -213,7 +213,13 @@ export class SPAClient extends BaseClient {
             // Get current cached tokens
```
sangonzal, line 213: Some unecessary console.log statements in this file.

sangonzal:
We should consider adding tests in both browser and node to catch these types of regressions.

lib/msal-common/src/client/SPAClient.ts
pkanher617: > We should consider adding tests in both browser and node to catch these types of regressions.

Planning on reviewing this tomorrow morning, need to improve unit testing in common.

PR #1634, opened 2020-05-12
[msal-browser] E2E tests for multiple resource scenarios

tnorling:
This PR adds a sample for requesting access tokens to multiple resources and an e2e test to ensure both tokens are being cached separately


pkanher617:
This is great, thanks for putting this out!

samples/VanillaJSTestApp2.0/app/multiple_resources/test/multiple_resources.spec.ts
```diff
@@ -0,0 +1,115 @@
+import "mocha";
+import puppeteer from "puppeteer";
+import { expect } from "chai";
+import fs from "fs";
```
jasonnutter, line 4: I would recommend using the promise fs APIs: https://nodejs.org/api/fs.html#fs_fs_promises_api

```js
import fs from "fs/promises";
```

PR #1646, opened 2020-05-12
[msal-browser][msal-node][msal-common] Fix initialization of B2cTrustedHostList

tnorling:
With some of the recent changes we are no longer calling `setKnownAuthorities` in the correct place. This needs to be called before we create the authority instance. This PR ensures that happens for browser and node use cases.


lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -91,6 +92,9 @@ export class PublicClientApplication {
         // Initialize the browser storage class.
         this.browserStorage = new BrowserStorage(this.config.auth.clientId, this.config.cache);
 
+        // Initialize default authority instance
+        B2cAuthority.setKnownAuthorities(this.config.auth.knownAuthorities);
```
sangonzal, line 96: Thoughts on moving this into `AuthorityFactory.createInstace()`? It's a pre-requisite for B2C to work, and right now we have to remember to do this every time we create an instance (that's not obvious from the code). Instead we can pass `this.config.auth.knownAuthorities` as a parameter and that way ensure that it's always called if set on the configuration.

sangonzal: Chatted offline, doesn't make sense.

PR #1655, opened 2020-05-14
[msal-node] Cache Lookup - 2: Save, lookup and delete entities

sameerag:
This PR adds the functionality to save, lookup and delete cache entities. It defines the below helpers:
- `saveAccount()`
- `saveCredential()`
- `getAccount()`
- `getCredential()`
- `getAccountsFilteredBy()`
- `getCredentialsFilteredBy()`
- `removeAccount()`
- `removeCredential()`

Unit tests added.

Note: These helpers help support `silent` and `refreshToken` flows in the `msal-node` library


lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -172,11 +172,14 @@ export class ResponseHandler {
      * @param authority
      */
     addAccountToCache(serverTokenResponse: ServerAuthorizationTokenResponse, idToken: IdToken, authority: Authority): void {
-        const environment = authority.canonicalAuthorityUrlComponents.HostNameAndPort;
-        let accountEntity: AccountEntity;
-        const cachedAccount: AccountEntity = this.uCacheManager.getAccount(this.homeAccountIdentifier, environment, idToken.claims.tid);
+        const accountEntity: AccountEntity = this.generateAccountEntity(
```
sameerag, line 175: This change is temporary. Next PR will change `ResponseHandler.ts` to handle silent flows. Please ignore this file in this review.

lib/msal-common/src/unifiedCache/UnifiedCacheManager.ts
```diff
@@ -61,47 +111,307 @@ export class UnifiedCacheManager {
...
+            ]
+        ) {
+            case "idtoken":
+                return cache.idTokens[key] || null;
+            case "accesstoken":
+                return cache.accessTokens[key] || null;
+            case "refreshtoken":
+                return cache.refreshTokens[key] || null;
+            default:
+                console.log("Cache entity type mismatch");
```
sangonzal, line 179: Logger instead of console.log?

lib/msal-common/src/unifiedCache/UnifiedCacheManager.ts
```diff
@@ -61,47 +111,307 @@ export class UnifiedCacheManager {
...
-        this.inMemoryCache.refreshTokens[refreshToken.generateCredentialKey()] = refreshToken;
+    private getCredentialsFilteredByInternal(
+        cacheCredentials: object,
+        homeAccountId?: string,
+        environment?: string,
+        credentialType?: string,
+        clientId?: string,
+        realm?: string,
+        target?: string
+    ): Object {
```
sangonzal, line 300: You can also do early return here as well. If any return false, you should return there instead of evaluating the rest. This is applicable in multiple where you are filtering.

lib/msal-common/src/unifiedCache/UnifiedCacheManager.ts
```diff
@@ -61,47 +111,307 @@ export class UnifiedCacheManager {
...
-    addAccountEntity(account: AccountEntity): void {
-        const accKey = account.generateAccountKey();
-        if (!this.inMemoryCache.accounts[accKey]) {
-            this.inMemoryCache.accounts[accKey] = account;
+    removeAccount(account: AccountEntity): boolean {
+        const cache = this.getCacheInMemory();
+        const accountKey = account.generateAccountKey();
+
+        delete cache.accounts[accountKey];
+        return true;
```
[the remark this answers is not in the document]
sameerag: You are right! Will make the changes. I thought I will get back to this and forgot :-D

lib/msal-common/src/unifiedCache/utils/CacheHelper.ts
```diff
@@ -43,4 +45,76 @@ export class CacheHelper {
...
+     * @param target
+     * @param credentialTarget
+     */
+    static targetsIntersect(credentialTarget: string, target: string): boolean {
+        const targetSet = new Set(target.split(" "));
+        const credentialTargetSet = new Set(credentialTarget.split(" "));
+
+        let isSubset = true;
+        targetSet.forEach((key) => {
+            isSubset = isSubset && credentialTargetSet.has(key);
```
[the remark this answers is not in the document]
sameerag: MDN docs say: `There is no way to stop or break a forEach() loop other than by throwing an exception. If you need such behavior, the forEach() method is the wrong tool.`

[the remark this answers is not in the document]
jasonnutter: Do these needs to be Sets? You would have more tools at your disposal (e.g. Array.some).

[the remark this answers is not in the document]
jasonnutter: Manually ending early also seems like an over optimization, what is the order of magnitude we are expecting these sets/arrays to be?

lib/msal-common/src/utils/Constants.ts
```diff
@@ -259,14 +259,27 @@ export enum CacheEntity {
...
+ * accountId: <home_account_id>-<environment>
+ * credentialId: <credential_type>-<client-id>-<realm>
+ */
+export enum CacheKeyPosition {
+    HOME_ACCOUNT_ID = 0,
+    ENVIRONMENT,
+    CREDENTIAL_TYPE,
+    CLIENT_ID,
+    REALM,
+    TARGET
```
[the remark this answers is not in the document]
jasonnutter: They get int values by default (order based).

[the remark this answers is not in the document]
jasonnutter: But I agree we should be consistent on setting them explicitly or not.

lib/msal-common/src/unifiedCache/utils/CacheHelper.ts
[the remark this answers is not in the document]
sameerag: Shouldn't be much. That is one reason why I let it be.

lib/msal-common/src/unifiedCache/UnifiedCacheManager.ts
sameerag: logger will be added as a separate PR as we need to pass the logger object

lib/msal-common/src/unifiedCache/UnifiedCacheManager.ts
pkanher617: +1 for early return here, would make these operations less expensive in the long-run

lib/msal-common/src/unifiedCache/utils/CacheHelper.ts
```diff
@@ -43,4 +45,86 @@ export class CacheHelper {
...
+    static matchTarget(key: string, target: string): boolean {
+        return CacheHelper.targetsIntersect(key.split(Separators.CACHE_KEY_SEPARATOR)[CacheKeyPosition.TARGET], target);
+    }
+
+    /**
+     * returns a boolean if the sets of scopes intersect (scopes are stored as "target" in cache)
+     * @param target
+     * @param credentialTarget
+     */
+    static targetsIntersect(credentialTarget: string, target: string): boolean {
```
pkanher617, line 109: This behavior is already present in  ScopeSet, not sure why we are duplicating it here.

lib/msal-common/src/unifiedCache/utils/CacheHelper.ts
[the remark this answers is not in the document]
pkanher617: This operation for intersection is more expensive than it needs to be. You can accomplish the same by comparing set lengths as we do in ScopeSet. Again, I don't know why this logic is duplicated here.

lib/msal-common/src/utils/Constants.ts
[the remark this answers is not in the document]
pkanher617: I think we should set them explicitly, it is more readable.

lib/msal-common/src/unifiedCache/utils/CacheHelper.ts
```diff
@@ -43,4 +45,86 @@ export class CacheHelper {
...
+        const targetSet = new Set(target.split(" "));
```
pkanher617, line 110:
```suggestion
        const targetSet = ScopeSet.fromString(target);
        const credentialTargetSet = ScopeSet.fromString(credentialTarget);
        return targetSet.intersectingScopeSets(credentialTargetSet);
```

lib/msal-common/src/unifiedCache/utils/CacheHelper.ts
```diff
@@ -43,4 +45,86 @@ export class CacheHelper {
         });
         return Object.assign({}, ...keyValues);
     }
+
```
pkanher617, line 48: Make sure we have tests cases for this file, especially null/empty inputs

lib/msal-common/src/response/ResponseHandler.ts
jasonnutter: Can you add a TODO here?

lib/msal-common/src/unifiedCache/interface/ICacheManager.ts
```diff
@@ -0,0 +1,69 @@
...
+     * @param homeAccountId
+     * @param environment
+     * @param realm
+     */
+    getAccountsFilteredBy(
+        filter: AccountFilter
+    ): AccountCache;
+
+    /**
+     * retrieve credentails matching all provided filters; if no filter is set, get all credentials
```
jasonnutter, line 46: Spelling "credentials"

lib/msal-common/src/unifiedCache/UnifiedCacheManager.ts
```diff
@@ -61,47 +111,341 @@ export class UnifiedCacheManager {
...
+        switch (
+            key.split(Separators.CACHE_KEY_SEPARATOR)[
+                CacheKeyPosition.CREDENTIAL_TYPE
+            ]
+        ) {
+            case "idtoken":
+                return cache.idTokens[key] || null;
+            case "accesstoken":
+                return cache.accessTokens[key] || null;
+            case "refreshtoken":
```
jasonnutter, line 176: Can you put these in an enum?

jasonnutter: Looks like it might already exist?

lib/msal-common/src/unifiedCache/UnifiedCacheManager.ts
```diff
@@ -26,33 +38,71 @@ export class UnifiedCacheManager {
...
+     * @param refreshToken
+     */
+    addCredentialCache(
+        accessToken: AccessTokenEntity,
+        idToken: IdTokenEntity,
+        refreshToken: RefreshTokenEntity
+    ): void {
+        this.inMemoryCache.accessTokens[
+            accessToken.generateCredentialKey()
+        ] = accessToken;
```
jasonnutter, line 86: Can you add setter functions to `InMemoryCache`? This is leaking implementation details.

lib/msal-common/src/unifiedCache/UnifiedCacheManager.ts
```diff
@@ -26,33 +38,71 @@ export class UnifiedCacheManager {
...
+        ] = refreshToken;
+    }
+
+    /**
+     * append account to the in memory cache
+     * @param account
+     */
+    addAccountEntity(account: AccountEntity): void {
+        const accKey = account.generateAccountKey();
+        if (!this.inMemoryCache.accounts[accKey]) {
```
jasonnutter, line 101: What if the existing `account` for this key is different than the passed in `account`? Or does that not matter?

lib/msal-common/src/unifiedCache/UnifiedCacheManager.ts
```diff
@@ -61,47 +111,341 @@ export class UnifiedCacheManager {
...
+        const key = credential.generateCredentialKey();
+
+        switch (credential.credentialType) {
+            case CredentialType.ID_TOKEN:
+                cache.idTokens[key] = credential as IdTokenEntity;
+                break;
+            case CredentialType.ACCESS_TOKEN:
+                cache.accessTokens[
+                    key
+                ] = credential as AccessTokenEntity;
```
jasonnutter, line 139: Same comment about setter functions.

lib/msal-common/src/unifiedCache/UnifiedCacheManager.ts
```diff
@@ -61,47 +111,341 @@ export class UnifiedCacheManager {
...
+            }
+
+            if (!!realm) {
+                matches = matches && CacheHelper.matchRealm(key, realm);
+            }
+
+            if (matches) {
+                matchingAccounts[key] = accounts[key];
+            }
+        });
```
jasonnutter, line 234: This could be rewritten using `reduce`:

```js
const matchingAccounts: AccountCache = Object.keys(accounts).reduce((matches, key) => {
  const matchHomeAccoutId = !!homeAccountId && CacheHelper.matchHomeAccountId(key, homeAccountId);
  const matchEnvironment = !!environment && CacheHelper.matchEnvironment(key, environment);
  const matchRealm = !!realm && CacheHelper.matchRealm(key, realm);
 
  if (matchHomeAccountId && matchEnvironment && matchRealm) {
    return {
      ...matches,
      [key]: accounts[key]
    }
  }

  return matches;
}, {});

File: lib/msal-common/src/unifiedCache/UnifiedCacheManager.ts [hunk 16]
```diff
@@ -61,47 +111,341 @@ export class UnifiedCacheManager {
...
+
+            // idTokens do not have "target", target specific refreshTokens do exist for some types of authentication
+            if (!!target && CacheHelper.getCredentialType(key) != CredentialType.ID_TOKEN.toString().toLowerCase()) {
+                matches = matches && CacheHelper.matchTarget(key, target);
+            }
+
+            if (matches) {
+                matchingCredentials[key] = cacheCredentials[key];
+            }
+        });
```
jasonnutter, line 371: Same comment about using reduce instead of `forEach`.

lib/msal-common/src/unifiedCache/UnifiedCacheManager.ts
```diff
@@ -61,47 +111,341 @@ export class UnifiedCacheManager {
...
+        const cache = this.getCacheInMemory();
+
+        const accountId = account.generateAccountId();
+
+        // TODO: Check how this should be done, do we just remove the account or also the associated credentials always?
+        Object.keys(cache.accessTokens).forEach((key) => {
+            if (
+                cache.accessTokens[key].generateAccountId() === accountId
+            )
+                this.removeCredential(cache.accessTokens[key]);
```
jasonnutter, line 406: Can you add curly brackets here (this will be enforced by linting soon).

lib/msal-common/src/unifiedCache/UnifiedCacheManager.ts
```diff
@@ -61,47 +111,341 @@ export class UnifiedCacheManager {
...
+                    credential.generateCredentialKey()
+                ];
+                return true;
+            case CredentialType.ACCESS_TOKEN:
+                delete cache.accessTokens[
+                    credential.generateCredentialKey()
+                ];
+                return true;
+            case CredentialType.REFRESH_TOKEN:
+                delete cache.refreshTokens[
```
jasonnutter, line 442: Helper functions would be nice here, as this is also leaking implementation details.

lib/msal-common/src/response/ResponseHandler.ts
sameerag: The PR immediately after addresses this holistically. Since I pull them in sequence, not adding a `TODO`.

lib/msal-common/src/unifiedCache/utils/CacheHelper.ts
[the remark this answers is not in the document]
sameerag: Set lengths comparison will not work in all cases here. If I want to pick up access_tokens that match for a set of scopes but for different accounts, I will still use this filter and would want to match the scopes content and not the count. We can discuss this further.

I can do an early return if that is a concern.

lib/msal-common/src/unifiedCache/utils/CacheHelper.ts
sameerag: explanation above. Keeping that comment open.

sameerag:
> Approving, but I'm starting to see a pattern of repeated logic, and it is going to bloat our codebase. Please make sure you understand the current codebase before making changes like this.

In this use case I believe `scopes` should be compared by content instead of count. As specified above, If a use case looks up `access_tokens` that match for a set of `scopes` but for different `accounts`, we would have to match the scopes content and not the count. We can discuss this further if I am missing anything.

lib/msal-common/src/unifiedCache/utils/CacheHelper.ts
sameerag: I added some, will append more corner cases soon

lib/msal-common/src/unifiedCache/UnifiedCacheManager.ts
sameerag: @jasonnutter  one question here. The statement if:
```javascript
if (matchHomeAccountId && matchEnvironment && matchRealm) 
```
Does this mandate all three are needed? The intent of the check is to pick any combination of the parameters of the given filter and return what matches. Hence the mutation on `matches` to ensure we skip/add params as passed by the user.

Open to any other way we can get this done.

lib/msal-common/src/unifiedCache/utils/CacheHelper.ts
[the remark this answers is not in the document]
pkanher617: > If I want to pick up access_tokens that match for a set of scopes but for different accounts, I will still use this filter and would want to match the scopes content and not the count

Not sure what you mean by this. I don't see any code comparing accounts here. The content is matched by the set class anyway. Look at the logic of the intersection function. We take the two sets, union them, and compare the length. If the length of the union is equal to the length of the two sets added together, there is no intersection. if the length of the union is less than the length of the two sets added together, it means that one of the values was common between both sets and we can assume there was an intersection.

lib/msal-common/src/unifiedCache/UnifiedCacheManager.ts
pkanher617: I think we overwrite it, the account key is based on homeAccountId so there is little chance of this being overwritten with an incorrect account.

lib/msal-common/src/unifiedCache/UnifiedCacheManager.ts
sameerag: I will add early return for all the cases as needed.

lib/msal-common/src/unifiedCache/interface/ICacheManager.ts
sameerag: Corrected

lib/msal-common/src/unifiedCache/UnifiedCacheManager.ts
sameerag: Need to convert existing ones to lower case. Done now.

lib/msal-common/src/unifiedCache/UnifiedCacheManager.ts
sameerag: Done

lib/msal-common/src/unifiedCache/utils/CacheHelper.ts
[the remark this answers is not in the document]
sameerag: As we discussed offline, I want to confirm scope match/intersection combination and how to handle that. I added this code as a primary requirement, but I do agree that we can move this functionality as needed to `ScopeSet(). Let me clarify the intersection rules and will address this.

lib/msal-common/src/unifiedCache/utils/CacheHelper.ts
sameerag: As mentioned above, will move this to ScopeSet in a follow up PR.

lib/msal-common/src/unifiedCache/UnifiedCacheManager.ts
sameerag: Done

lib/msal-common/src/unifiedCache/UnifiedCacheManager.ts
sameerag: I have added overWrite as "remove" + "save" to be closer to a DB implementation.

lib/msal-common/src/unifiedCache/UnifiedCacheManager.ts
sameerag: Done

lib/msal-common/src/unifiedCache/UnifiedCacheManager.ts
sameerag: @jasonnutter  This is the only comment not addressed for this PR for now. I will add this once end to end testing is verified, as a part of this issue: #1744

lib/msal-common/src/unifiedCache/UnifiedCacheManager.ts
sameerag: Done

PR #1657, opened 2020-05-14
[msal-node] Clean up public API

sangonzal:


lib/msal-common/src/request/AuthorizationCodeUrlRequest.ts
```diff
@@ -3,8 +3,11 @@
  * Licensed under the MIT License.
  */
 
+import { Prompt, ResponseMode } from "../utils/Constants";
+
 /**
- * @type AuthorizationCodeUrlRequest: Request object passed by user to retrieve a Code from the server (first leg of authorization code grant flow)
+ * @type AuthorizationCodeUrlRequest: Request object passed by user to retrieve a Code from the
+ * server (first leg of authorization code grant flow)
```
sameerag, line 10: Thanks for this. Did we verify typedocs format we would need for comments like this?

lib/msal-common/src/request/AuthorizationCodeUrlRequest.ts
```diff
@@ -77,8 +91,8 @@ export type AuthorizationCodeUrlRequest = {
     claims?: string;
 
     /**
-     *  A value included in the request that is also returned in the token response. A randomly
-     *  generated unique value is typically used for preventing cross site request forgery attacks.
+     *  A value included in the request that is returned in the id token. A randomly
+     *  generated unique value is typically used to mitigate replay attacks.
```
sameerag, line 95: General q. For type docs, shouldn't all comments be added before the type definition and not inside? I am not 100% sure, cc @pkanher617 since you deployed these for `msal-browser` can you please weigh in here?

lib/msal-common/src/request/DeviceCodeRequest.ts
```diff
@@ -30,8 +30,9 @@ export type DeviceCodeRequest = {
     cancel?: boolean;
 
     /**
-     * URI of the authority from which MSAL will acquire the tokens from. If this value is not set, MSAL defaults
-     * to the authority used when configuring the PublicClientApplication.
+     * Url of the authority which the application acquires tokens from. Defaults to
+     * https://login.microsoftonline.com/common. If using the same authority for all request, authority should set
+     * on client application object and not request, to avoid resolving authority endpoints multiple times.
```
sameerag, line 35: +1

lib/msal-common/src/utils/Constants.ts
```diff
@@ -34,7 +34,6 @@ export const Constants = {
     CODE_GRANT_TYPE: "authorization_code",
```
sameerag, line 34: You may want to clean some of these as they have their own enums defined in the same file below!

lib/msal-common/src/request/AuthorizationCodeUrlRequest.ts
sangonzal: Have not verified. If we need a different format, I'll update as part of the PR for enabling typedoc.

PR #1666, opened 2020-05-15
Move Msal Common to Async Storage + Make the Updates In Node

DarylThayil:
Update Storage to be async to enable persistent cache operations in Node.js but also handle potential browser cache async operations like IndexDb
Some of this may seem necessary in terms of async method changes for getters and setters but there are unforseen sync operations that may need to happen and I would rather change the surface once 

* Moved Msal Common to use Async Storage 
* Updated Node to use this interface 
* Fixed tests 

Many APIS changed from sync to Async as a consequence


pkanher617:
Can we please include changes to the browser in this PR?

DarylThayil:
@pkanher617 once we are ideologically aligned I will make those changes in a separate pr. It may be worth considering this as a Major bump to common especially since there are already betas out there depending on common

lib/msal-common/src/cache/CacheHelpers.ts
```diff
@@ -56,114 +56,121 @@ export class CacheHelpers {
...
-                    try {
-                        const parseAtKey = JSON.parse(key) as AccessTokenKey;
-                        if (this.checkForExactKeyMatch(parseAtKey, clientId, authority, resource, homeAccountIdentifier)) {
-                            const newAccessTokenCacheItem = new AccessTokenCacheItem(parseAtKey, JSON.parse(value) as AccessTokenValue);
-                            return tokens.concat([ newAccessTokenCacheItem ]);
-                        }
-                    } catch (e) {
-                        throw ClientAuthError.createCacheParseError(key);
+    async getAllAccessTokens(clientId: string, authority: string, resource?: string, homeAccountIdentifier?: string): Promise<Array<AccessTokenCacheItem>> {
+        const cacheKeys: Array<string> = await this.cacheStorage.getKeys();
```
DarylThayil, line 129: Please review this code in depth, I had to make some changes and want to make sure it checks out. 
Tests passed. 

In interesting change is the change from reduce to map for promises

lib/msal-common/src/cache/CacheHelpers.ts
```diff
@@ -56,114 +56,121 @@ export class CacheHelpers {
...
+                    throw ClientAuthError.createCacheParseError(key);
                 }
             }
-            return tokens;
-        }, []);
+            return null;
+        });
 
-        return results;
+        return Promise.all(foundTokens).then(arr => arr.filter(item => item));
```
DarylThayil, line 149: This filter basically takes the resolution of all promises and then only returns Items that are not null resolutions;

lib/msal-common/src/cache/CacheHelpers.ts
```diff
@@ -56,114 +56,121 @@ export class CacheHelpers {
...
-            const keyMatches = key.match(clientId) && key.match(authority) && key.match(resource) && key.match(homeAccountIdentifier);
-            if (keyMatches) {
-                try {
-                    const parseAtKey = JSON.parse(key) as AccessTokenKey;
-                    if (this.checkForExactKeyMatch(parseAtKey, clientId, authority, resource, homeAccountIdentifier)) {
-                        this.cacheStorage.removeItem(key);
-                    }
-                } catch (e) {
-                    throw ClientAuthError.createCacheParseError(key);
+    async removeAllAccessTokens(clientId: string, authority: string, resource?: string, homeAccountIdentifier?: string): Promise<void> {
```
DarylThayil, line 157: Review this as well, no tests failed but want to make sure logic checks out

lib/msal-common/src/cache/CacheHelpers.ts
```diff
@@ -56,114 +56,121 @@ export class CacheHelpers {
...
+                if (this.checkForExactKeyMatch(parsedAccessTokenKey, clientId, authority, resource, homeAccountIdentifier)) {
+                    return await this.cacheStorage.removeItem(key);
                 }
+                return null;
+            } catch (e) {
+                throw ClientAuthError.createCacheParseError(key);
             }
-        });
+        }).filter(resolution => resolution);
+        return Promise.all(removedTokens).then(() => null);
```
DarylThayil, line 173: Since this returns Promise of void, there is no filter needed, just wait for all to resolve then return null

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -68,7 +68,7 @@ export class AuthorizationCodeClient extends BaseClient {
         );
 
         // set the final cache and return the auth response
-        this.updateCache();
+        await this.updateCache();
```
DarylThayil, line 71: @jasonnutter  I would love a TS rule that doesn't allow floating un-awaited promises. Its sorta hard to make sure I got all the async stuff that needs to be awaited. I looked into it but couldnt figure it out

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -103,11 +103,11 @@ export class ResponseHandler {
      * @param resource
      * @param state
      */
-    generateAuthenticationResult(serverTokenResponse: ServerAuthorizationTokenResponse, authority: Authority): AuthenticationResult {
+    async generateAuthenticationResult(serverTokenResponse: ServerAuthorizationTokenResponse, authority: Authority): Promise<AuthenticationResult> {
```
DarylThayil, line 106: Please review / think through consequences of response handling being async

lib/msal-common/src/unifiedCache/UnifiedCacheManager.ts
```diff
@@ -9,54 +9,65 @@ import { AccessTokenEntity } from "./entities/AccessTokenEntity";
...
      * setter  for in cache memory
      */
-    setCacheInMemory(cache: InMemoryCache): void {
+    async setCacheInMemory(cache: InMemoryCache): Promise<void> {
         this.inMemoryCache = cache;
     }
 
+    // if inMemmory cache is not set, set it from caheStorage
+    // this async operation can't be done in constructor
+    async setCacheInMemmoryIfNotSet(): Promise<void> {
```
DarylThayil, line 36: This is interesting and sorta unseltting.  The get cache function being async means that it cant be called in the constructor for the initial population. So in all the member functions I have had to add this set if not set function to make sure the internal state is created before any operations take place

DarylThayil:
Please consider suggesting test coverage that I may not have . @pkanher617  added really great test coverage so I felt it was pretty well covered

lib/msal-common/src/unifiedCache/UnifiedCacheManager.ts
sameerag: yes. This is the initial problem I had with async functions, so I wrapped async only for file access and made other APIs sync. I do see the value in this though, as in if we have (key, value) persistence, we should have a reasonable wrapper for async implementation. Let's brainstorm if there is a better approach.

DarylThayil:
Closing for now, decided to punt async

PR #1668, opened 2020-05-15
ADFS 2019 Support

### LINKED ISSUE #99: ADFS Support?
Any plan to add ADFS support as the.net library seems to be working on it.?

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/143

### LINKED ISSUE #1703: ClientInfo not received in the response from AAD
I'm trying to obtain an access token using `.getAccessToken(args)` from the `react-aad-msal` npm package.  The console is reporting a couple of errors that originate from this repo:

`[ERROR] ClientAuthError: The service did not populate client_info in the response, Please verify with the service team Failed with error: ClientInfo not received in the response from the server`

And

`VM443 1.chunk.js:27570 Uncaught (in promise) ClientAuthError: The service did not populate client_info in the response, Please verify with the service team Failed with error: ClientInfo not received in the response from the server
    at ClientAuthError.AuthError [as constructor] (http://localhost:3000/static/js/1.chunk.js:27570:24)
    at new ClientAuthError (http://localhost:3000/static/js/1.chunk.js:27715:24)
    at Function.ClientAuthError.createClientInfoNotPopulatedError (http://localhost:3000/static/js/1.chunk.js:27798:12)
    at MsalAuthProvider.UserAgentApplication.saveTokenFromHash (http://localhost:3000/static/js/1.chunk.js:25782:93)
    at MsalAuthProvider.UserAgentApplication.processCallBack (http://localhost:3000/static/js/1.chunk.js:25245:23)
    at MsalAuthProvider.UserAgentApplication.handleAuthenticationResponse (http://localhost:3000/static/js/1.chunk.js:25299:10)
    at MsalAuthProvider.<anonymous> (http://localhost:3000/static/js/1.chunk.js:25034:20)
    at step (http://localhost:3000/static/js/1.chunk.js:84333:17)
    at Object.next (http://localhost:3000/static/js/1.chunk.js:84264:14)
    at fulfilled (http://localhost:3000/static/js/1.chunk.js:84217:24)`

Stepping through the source code:

https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/18c3f99fa1a3877c7965963de45db4e65a9d9c43/lib/msal-core/src/UserAgentApplication.ts#L1617

I can see that `hashParams` doesn't contain a `client_info`, which is why I get the error. 

However, the comment:
`// retrieve client_info - if it is not found, generate the uid and utid from idToken`
suggests that it will attempt to generate the client info from the `id_token` which I do have.  Unfortunately, from what I can tell, the code logs the error and throws an exception without trying to create the client info.

What is the intended behaviour?  Should the library attempt to create the object from the `id_token` (as per the comment), or does `client_info` really need to be set correctly up stream?

tnorling:
This PR ports the changes in #1617 to msal-core and builds on #1582 and will close #99

Differences from AAD/B2C Authorities:
- ADFS does not return `ClientInfo`, this PR builds `uid` from the subject claim on idToken instead
- ADFS does not return `preferred_username`, this PR sets this to `upn` instead
- ADFS uses `.well-known/openid-configuration` instead of `2.0/.well-known/openid-configuration`


lib/msal-core/src/ClientInfo.ts
```diff
@@ -30,6 +30,12 @@ export class ClientInfo {
         this._utid = utid;
     }
 
+    static createClientInfoFromIdToken(oid:string): string {
```
technical-boy, line 33: I think the method name may cause a little confusion given that the argument it receives is not really an idToken but a specific claim. Maybe it'd be better to receive the idToken and extract the subject/oid inside the method?

lib/msal-core/src/authority/Authority.ts
```diff
@@ -33,6 +33,25 @@ export class Authority {
         this.tenantDiscoveryResponse = authorityMetadata;
     }
 
+    public static isAdfs(authorityUrl: string): boolean {
+        const components = UrlUtils.GetUrlComponents(authorityUrl);
+        const pathSegments = components.PathSegments;
+
+        if (pathSegments.length && pathSegments[0].toLowerCase() === Constants.ADFS) {
```
technical-boy, line 40: Can we just return the check's result?

```js
        return (pathSegments.length && pathSegments[0].toLowerCase() === Constants.ADFS);
```

Saves us the if and the return true/false.

lib/msal-core/src/authority/Authority.ts
```diff
@@ -33,6 +33,25 @@ export class Authority {
...
+            return true;
+        }
+
+        return false;
+    }
+
+    public get AuthorityType(): AuthorityType {
+        if (Authority.isAdfs(this.canonicalAuthority)) {
```
technical-boy, line 48: I assume this is not a ternary because there will be more Authority Types in the future. If that's the case, can we make this a `switch` statement and have AuthorityType.Default be the `default` case?

lib/msal-core/src/authority/Authority.ts
```diff
@@ -87,6 +106,9 @@ export class Authority {
 
     // http://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata
     protected get DefaultOpenIdConfigurationEndpoint(): string {
+        if (this.AuthorityType === AuthorityType.Adfs){
+            return `${this.CanonicalAuthority}.well-known/openid-configuration`;
```
technical-boy, line 110: Should we make ".well-known/openid-configuration" a constant?

hectormmg (approved):

technical-boy:
LGTM, just a few nitpicks

lib/msal-core/src/authority/Authority.ts
tnorling: No I actually did make this a ternary but must have lost that change when I was moving stuff around. Thanks for catching this!

lib/msal-core/src/ClientInfo.ts
tnorling: I dont disagree. Thanks for the suggestion

lib/msal-core/src/authority/Authority.ts
tnorling: Done

lib/msal-core/src/authority/Authority.ts
tnorling: Done

jmckennon (approved):
Looks great.

lib/msal-core/src/ClientInfo.ts
```diff
@@ -30,6 +31,12 @@ export class ClientInfo {
         this._utid = utid;
     }
 
+    static createClientInfoFromIdToken(idToken:IdToken): string {
+        const clientInfo = {uid: idToken.subject, utid: ""};
```
jasonnutter, line 35: Can you put each property on a new line?

```suggestion
        const clientInfo = {
            uid: idToken.subject, 
            utid: ""
        };
```

jasonnutter (approved):
One nitpick otherwise LGTM, good work!

James-Randal:
So is this resolved or not?

tnorling:
@James-Randal ADFS 2019 is supported in msal@1.4.0. If you have questions or find a bug please open a new issue. Thanks!

sglickman:
Is there an example of the configuration required to get this working with AD FS 2019? I'm trying to follow <a href="https://docs.microsoft.com/en-us/azure/active-directory/develop/tutorial-v2-angular-auth-code">this tutorial</a> but it's attempting to hit login.microsoftonline.com which returns a 400 error.

```
    MsalModule.forRoot(new PublicClientApplication({
      auth: {
        clientId: CLIENT_ID, // This is your client ID
        authority: ADFS_2019_ENDPOINT, // This is your tenant ID
        redirectUri: 'http://localhost:4200/' // This is your redirect URI
      },
      cache: {
        storeAuthStateInCookie: isIE, // Set to true for Internet Explorer 11
      }
    }), null, null)
```

`ADFS_2019_ENDPOINT` looks like 'https://staging.mydomain.com/adfs'

I would have expected that supplying a URL for the `authority` field would have overridden any references to login.microsoftonline.com, but am I doing something wrong?

PR #1670, opened 2020-05-16
Fixed package name in Upgrade Guide

jerone:
And improve flow of sentence.


jasonnutter:
Good catch, thanks for the fix!

PR #1675, opened 2020-05-18
[msal-browser][msal-common] Make library state an object and add timestamp

pkanher617:
This PR converts library state from a GUID to a stringified JSON object with id and timestamp fields. This will allow msal-common and msal-browser to handle situations where responses are handled long after the request was initiated, and setting correct token expiration lifetimes.

Also includes tests.


lib/msal-common/src/utils/ProtocolUtils.ts
```diff
@@ -15,23 +28,56 @@ export class ProtocolUtils {
...
+    /**
+     * Extracts user state value and library state value 
+     * from the state sent with the authentication request.
+     * @param state 
+     * @param cryptoObj 
+     */
+    static parseRequestState(state: string, cryptoObj: ICrypto): RequestStateObject {
+        try {
+            if (StringUtils.isEmpty(state)) {
+                state = "";
```
jasonnutter, line 63: I don't think you should handle this case, I think throwing an error below is preferable.

jasonnutter: Also, we should not be mutating the passed in variable (even if null).

lib/msal-common/src/utils/ProtocolUtils.ts
```diff
@@ -15,23 +28,56 @@ export class ProtocolUtils {
...
             }
+
+            const [
+                libraryState,
+                userState
+            ] = decodeURIComponent(state).split(Constants.RESOURCE_DELIM);
```
jasonnutter, line 69: We should make sure that we are properly handling if `userState` contains the delimiter.

lib/msal-common/src/utils/ProtocolUtils.ts
pkanher617: The problem is if state is `null`, then `decodeURIComponent(state).split(Constants.RESOURCE_DELIM)` returns `["null"]`. If state is empty string `""`, then it correctly throws the error.

jasonnutter: Gotya. How about this:

```js
if (StringUtils.isEmpty(state) {
  throw ClientAuthError.createInvalidStateError(state);
} 

try {
```

pkanher617: done.

lib/msal-common/src/utils/ProtocolUtils.ts
pkanher617: I've updated the code to handle this scenario, and added tests for the use case as well.

lib/msal-common/src/utils/ProtocolUtils.ts
```diff
@@ -15,23 +28,52 @@ export class ProtocolUtils {
...
+     * @param cryptoObj 
+     */
+    static parseRequestState(state: string, cryptoObj: ICrypto): RequestStateObject {
+        if (StringUtils.isEmpty(state)) {
+            throw ClientAuthError.createInvalidStateError(state);
+        }
+
+        try {
+            const splitState = decodeURIComponent(state).split(Constants.RESOURCE_DELIM);
+            const libraryState = splitState.splice(0, 1)[0];
```
tnorling, line 67: Why do we need to splice here?

jasonnutter: Comment might be helpful here @pkanher617

pkanher617: Will add a comment here. @tnorling This is so that we can do the `.join()` on the next line and preserve the user state.

tnorling: Would it not be cleaner to use `splitState[0]` here and then do `splitState.slice(1).join(Constants.RESOURCE_DELIM)` below instead of mutating the original array?

pkanher617: done.

lib/msal-common/src/utils/ProtocolUtils.ts
```diff
@@ -15,23 +28,52 @@ export class ProtocolUtils {
...
+     */
+    static parseRequestState(state: string, cryptoObj: ICrypto): RequestStateObject {
+        if (StringUtils.isEmpty(state)) {
+            throw ClientAuthError.createInvalidStateError(state);
+        }
+
+        try {
+            const splitState = decodeURIComponent(state).split(Constants.RESOURCE_DELIM);
+            const libraryState = splitState[0];
+            const userState = splitState.slice(1).join(Constants.RESOURCE_DELIM);
```
tnorling, line 68: Probably should check that it actually has a second element to avoid unexpected errors:
```suggestion
            const userState = splitState.length > 1 ? splitState.slice(1).join(Constants.RESOURCE_DELIM): "";
```

PR #1680, opened 2020-05-19
[msal-node] Cache Lookup - 3: Msal node cache response

sameerag:
This PR updates `ResponseHandler.ts`, introduces `CacheRecord`.


lib/msal-node/src/cache/CacheContext.ts
```diff
@@ -16,18 +16,17 @@ import { Storage } from '../cache/Storage';
...
         IdToken: {},
         AccessToken: {},
         RefreshToken: {},
-        AppMetadata: {}
+        AppMetadata: {},
     };
 
     constructor() {
-        this.cachePath = "";
+        this.cachePath = '';
```
sangonzal, line 29: Do we need cachePath anymore?

lib/msal-node/src/cache/Storage.ts
```diff
@@ -2,23 +2,20 @@
...
 export class Storage implements ICacheStorage {
     // Cache configuration, either set by user or default values.
-    private cacheConfig: CacheOptions;;
+    private cacheConfig: CacheOptions;
     private inMemoryCache: InMemoryCache;
 
     constructor(cacheConfig: CacheOptions) {
         this.cacheConfig = cacheConfig;
-        if (this.cacheConfig.cacheLocation! === "fileCache")
+        if (this.cacheConfig.cacheLocation! === 'fileCache')
```
sangonzal, line 18: Temporary?

lib/msal-node/src/cache/CacheContext.ts
sameerag: We don't if we have keyTar working as path is calculated. Also I think Daryl's PR makes this unnecessary. I can pull dev and change this.

lib/msal-node/src/cache/Storage.ts
sameerag: Indication we are using application cache by the user: extension or otherwise instead of just `inMemoryCache`. Ideas welcome.

PR #1681, opened 2020-05-19
[msal-core] Update and refactor scope validation

hectormmg:
## Description

This PR updates the way scopes are validated and dynamically generated for login requests as a primer for token request type definition behavior changes (PR #1628).

## Overview

1. Rather than having a single `validateRequest` method in `RequestUtils` that checks whether the request is of "Login" type through a boolean argument, this PR adds a separate `validateLoginRequest` method that is called from our login APIs, and wraps the `validateRequest` method which is called from `acquireToken` APIs.
2. The new `RequestUtils.validateLoginRequest` method is responsible for appending `openid` and `profile` to the request scopes before calling the more general `validateRequest` method.
3. This PR refactors `UrlUtils.translateclientIdUsedInScope` out into `ScopeSet.generateLoginScopes` in order to have that behavior under the right namespace.
4. Updates all automated tests that break because of the new scopes validation and manipulation behavior
5. Add tests for `ScopeSet` and `RequestUtils` classes.

### ScopeSet.ts

1. `validateInputScopes`: Since `RequestUtils.validateRequest` doesn't need to check for `!isLoginCall` anymore and the check for `!scope` is carried out regardless, `ScopeSet.validateInputScopes` can assume request is non-null. This means validateInputScopes now only throws in the following cases:
    1. Null scopes in request
    2. Non-Array scopes in request (i.e. space separeated scope strings)
    3. Empty scopes arrays

2. `generateLoginScopes`: Since `UrlUtils.translateclientIdUsedInScope` deals with scopes, it is moved to `ScopeSet` and refactored as `generateLoginScopes`. This method is always called in `RequestUtils.validateLoginRequest` for login API calls, and is called in `RequestUtils.validateRequest` when any of the three recognized "login scopes" is included in the request scopes, as determined by the result of calling `ScopeSet.isLoginScopes`. This forces login API calls to contain 'openid' and 'profile' scopes, and allows access token requests to also make id token requests (i.e. id_token token response type). This method is written in a way that avoids adding duplicates of login scopes.

3. `isLoginScopes`: A simple new method that returns true if any of the login scopes ('openid', 'profile' or the clientId value) are included in the scopes. This method determines which non-login API acquire token calls should also have 'openid' and 'profile' scopes appended and, therefore, obtain an id token as well as an access token.

### ServerRequestParameters.ts

1. Refactor `scopes` assignment into a single line that trims and converts scopes array to lowercase or sets scopes to empty array if request scopes are null.

### UserAgentApplication:

1. Replace call to `RequestUtils.validateRequest` with `RequestUtils.validateLoginRequest` in login API calls, as well as update parameters in all `validateRequest` calls given the change in the method's signature (`isLoginCall` param no longer required).

2. `acquireTokenHelper`: Remove `scope` assignment for `registerCallback` call at the top of the method and add an updated version that simply joins the request scopes, which have already been trimmed, downcased and validated, within the `this.registerCallback` call arguments.

3. `buildIDTokenRequest` Remove client id from request, replace with a new array build from spreading whatever scopes the request already has.

#### ClientConfigurationError.ts

1. Remove `client_id_input_scopes_error` since we're no longer forcing clientId as only scope.

#### RequestUtils.ts

1. `validateRequest`:
    1. Remove `isLoginCall` boolean parameter from method signature
    2. Remove scopes validation logic (move to new `validateLoginRequest`).
    3. Add a line that checks whether scopes are/should be login scopes and generatesLoginScopes if true (for access token requests that also obtain an idToken).
    4. Remove extra scopes to consent logic and move to `validateLoginRequest`.

2. `validateLoginRequest`: New method in `RequestUtils` that wraps `validateRequest`, only called from login APIs in `UserAgentApplication`.
    1. Appends extra scopes to consent to scopes array
    2. Forces generation of login scopes (i.e. add 'openid' and 'profile' if they're not already in scopes array).
    3. Calls `validateRequest` to continue regular request validation.

#### UrlUtils.ts

1. `createNavigateUrlString`:
    1. Remove line that appends clientId to request scopes (this line mutated the scopes)
    2. Remove login scopes "translation" logic.
        - `openid` and `profile` are always added, regardless of whether `clientId` is present or not
        - `clientId` is not being removed anymore, so it is added as a scope to the request URL string (in accordance to the B2C use case where `clientId` value is a valid scope).
2.`translateClientIdUsedInScope`: Method refactored out into `ScopeSet.generateLoginScopes` and `ScopeSet.isLoginScopes`.


lib/msal-core/src/utils/RequestUtils.ts
```diff
@@ -75,13 +78,40 @@ export class RequestUtils {
...
+        if (!request.scopes) {
+            request.scopes = [];
+        }
+
+        // Append clientId to scopes by default for login calls
+        request.scopes = ScopeSet.appendScopes(request.scopes, [clientId]);
+
+        // validate request
+        const userRequest: AuthenticationParameters = RequestUtils.validateRequest(request, clientId, interactionType);
+        userRequest.scopes = ScopeSet.appendScopes(userRequest.scopes, userRequest.extraScopesToConsent);
```
technical-boy, line 103: Echoing what @pkanher617 noted on the original branch:

This was being called before validateInputScopes before, is now being called afterwards - is that ok? Would have failed the client id check otherwise, but since that was removed might be better to move it before validateRequest.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -320,13 +320,27 @@ export class UserAgentApplication {
...
     loginRedirect(userRequest?: AuthenticationParameters): void {
+        // Check if userRequest is empty
+        if (!userRequest) {
+            userRequest = {};
+        }
+        
+        // Check if request scopes are null
+        if (!userRequest.scopes) {
+            userRequest.scopes = [];
+        }
```
jasonnutter, line 337: Instead of creating this `userRequest`, I would prefer to see a new array created (and then have downstream code properly handle a falsey `userRequest`:

```js
const scopes = (userRequest && userRequest.scopes) || [];
```

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -320,13 +320,27 @@ export class UserAgentApplication {
...
     // #endregion
+
     /**
      * Use when initiating the login process by redirecting the user's browser to the authorization endpoint.
      * @param {@link (AuthenticationParameters:type)}
      */
     loginRedirect(userRequest?: AuthenticationParameters): void {
+        // Check if userRequest is empty
+        if (!userRequest) {
+            userRequest = {};
```
jasonnutter, line 331: We should avoid mutating function parameters (even when they are falsey), as it can cause intended side-effects. We'll have a lint rule in the future to disallow this.

lib/msal-core/src/utils/RequestUtils.ts
```diff
@@ -75,13 +78,40 @@ export class RequestUtils {
...
+        // Check if request scopes are null
+        if (!request.scopes) {
+            request.scopes = [];
+        }
+
+        // Append clientId to scopes by default for login calls
+        request.scopes = ScopeSet.appendScopes(request.scopes, [clientId]);
+
+        // validate request
+        const userRequest: AuthenticationParameters = RequestUtils.validateRequest(request, clientId, interactionType);
```
jasonnutter, line 102: Same comment as above about scopes:

```js
const requestScopes = (request && request.scopes) || [];
const scopes = ScopeSet.appendScopes(requestScopes, [ clientId ]);

const userRequest: AuthenticationParameters = RequestUtils.validateRequest({ ...request, scopes }, clientId, interactionType);

Comment (jasonnutter) [reply lib/msal-core/src/utils/RequestUtils.ts:102]: Key in both these examples, I don't like conflation the falseyness of `request` with the need to mutate the passed in scopes. I think it is simpler to construct the needed scopes array outside the context of the request, and then handle the truthiness of that seperately.

File: lib/msal-core/src/utils/RequestUtils.ts [hunk 5]
```diff
@@ -34,20 +34,24 @@ export class RequestUtils {
...
+    static validateRequest(request: AuthenticationParameters, clientId: string, interactionType: InteractionType): AuthenticationParameters {
         // Throw error if request is empty for acquire * calls
-        if(!isLoginCall && !request) {
+        if (!request) {
             throw ClientConfigurationError.createEmptyRequestError();
         }
+        
+        // Check if sent scopes are Login scopes
+
+        // const isLoginScopes = ScopeSet.isLoginScopes(request.scopes, clientId);
```
jasonnutter, line 45: Can you remove these commented out lines?

lib/msal-core/src/ScopeSet.ts
```diff
@@ -130,13 +120,46 @@ export class ScopeSet {
...
+     */
+    static generateLoginScopes(scopes: Array<string>, clientId: string): Array<string> {
+        const loginScopes = [...scopes];
+        const clientIdIndex: number = loginScopes.indexOf(clientId);
+        if (this.isLoginScopes(loginScopes, clientId)) {
+            if (clientIdIndex >= 0) {
+                loginScopes.splice(clientIdIndex, 1);
+            }
+            if (loginScopes.indexOf("openid") === -1) {
+                loginScopes.push("openid");
```
jasonnutter, line 144: Does "openid" and "profile" exist as constants somewhere? I think they should, given how often they are referenced.

lib/msal-core/src/ScopeSet.ts
```diff
@@ -130,13 +120,46 @@ export class ScopeSet {
...
+    /**
+     * append the required scopes: https://openid.net/specs/openid-connect-basic-1_0.html#Scopes
+     * @param scopes
+     */
+    static generateLoginScopes(scopes: Array<string>, clientId: string): Array<string> {
+        const loginScopes = [...scopes];
+        const clientIdIndex: number = loginScopes.indexOf(clientId);
+        if (this.isLoginScopes(loginScopes, clientId)) {
+            if (clientIdIndex >= 0) {
+                loginScopes.splice(clientIdIndex, 1);
```
jasonnutter, line 141: Can you add a comment about why this is needed?

jasonnutter (changes requested):

lib/msal-core/src/utils/RequestUtils.ts
jasonnutter: And the `...` operator takes care of that for us.

lib/msal-core/src/UserAgentApplication.ts
technical-boy: Would a ternary operator work better here? I assume this would set scopes to a boolean "true" when userRequest and userRequest.scopes are truthy, rather than set scopes to userRequest.scopes.

lib/msal-core/src/ScopeSet.ts
technical-boy: Yes, the constants where already in place, now using them in generateLoginScopes.

lib/msal-core/src/UserAgentApplication.ts
technical-boy: Done.

lib/msal-core/src/UserAgentApplication.ts
technical-boy: Nvm, thanks for the explanation on this JS functionality.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -320,13 +320,27 @@ export class UserAgentApplication {
...
+        }
+        
+        // Check if request scopes are null
+        if (!userRequest.scopes) {
+            userRequest.scopes = [];
+        }
+
+        // Append clientId to scopes by default for login calls
+        userRequest.scopes = ScopeSet.appendScopes(userRequest.scopes, [this.clientId]);
```
tnorling, line 340: Can we change this to openid and profile instead of clientId since we will replace clientId later anyway?

lib/msal-core/src/ScopeSet.ts
technical-boy: Done.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -2032,7 +2045,8 @@ export class UserAgentApplication {
         // acquireTokenSilent
         if (silentCall) {
             if (Account.compareAccounts(accountObject, this.getAccount())) {
-                tokenType = (scopes.indexOf(this.config.auth.clientId) > -1) ? ResponseTypes.id_token : ResponseTypes.token;
+                const idTokenRequested = scopes.indexOf(this.config.auth.clientId) > -1 || scopes.indexOf("openid") > -1; // maybe add profile
```
tnorling, line 2048: If this is done after scopes are validated/set we don't need clientId here and can just stick to openid/profile

lib/msal-core/src/utils/RequestUtils.ts
technical-boy: Done.

lib/msal-core/src/utils/RequestUtils.ts
```diff
@@ -75,13 +78,40 @@ export class RequestUtils {
...
+            request = {};
+        }
+
+        // Check if request scopes are null
+        if (!request.scopes) {
+            request.scopes = [];
+        }
+
+        // Append clientId to scopes by default for login calls
+        request.scopes = ScopeSet.appendScopes(request.scopes, [clientId]);
```
tnorling, line 99: clientId again

lib/msal-core/src/ScopeSet.ts
```diff
@@ -80,13 +80,10 @@ export class ScopeSet {
      * @param {boolean} scopesRequired - Boolean indicating whether the scopes array is required or not
```
jasonnutter, line 80: Can you update the comments here?

lib/msal-core/src/UserAgentApplication.ts
technical-boy: Right, the next PR will be dedicated to updating the getTokenType logic to include all the response type cases we've discussed.

lib/msal-core/src/utils/RequestUtils.ts
technical-boy: Done

lib/msal-core/src/UserAgentApplication.ts
technical-boy: Done!

lib/msal-core/src/utils/RequestUtils.ts
technical-boy: Yeah, this was duplicated code that was removed during the design discussion, I changed it to an array containing openid and profile in the right place.

lib/msal-core/src/ScopeSet.ts
technical-boy: Done!

lib/msal-core/test/ScopeSet.spec.ts
```diff
@@ -1,16 +1,205 @@
...
+    });
+
+    describe("trimAndConvertToLowerCase", () => {
+        it("should downcase and remove all whitespace from the scope passed in", () => {
+            const scope = " S1 ";
+            expect(ScopeSet.trimAndConvertToLowerCase(scope)).to.be.eq("s1");
+        });
+    });
+
+    describe("removeElement", function() {
```
tnorling, line 64: Add a test for output array = input array if `scopeToRemove` is not in the input array

lib/msal-core/test/ScopeSet.spec.ts
```diff
@@ -1,16 +1,205 @@
...
+            const scopes = ["s1"];
+            expect(() => ScopeSet.validateInputScope(scopes)).to.not.throw();
+        });
+
+        it("should not throw any errors when multiple scopes are passed in scopes array", () => {
+            const scopes = ["s1", "s2"];
+            expect(() => ScopeSet.validateInputScope(scopes)).to.not.throw();
+        });
+
+        it("should throw createScopesRequiredError if scopes are empty or null", function() {
```
tnorling, line 85:
```suggestion
        it("should throw createScopesRequiredError if scopes are null", function() {
```

lib/msal-core/test/ScopeSet.spec.ts
```diff
@@ -1,16 +1,205 @@
...
-    it("tests trimAndConvertArrayToLowerCase", () => {
-        const scopeSet = ["S1", " S2", " S3 "];
-        expect(ScopeSet.trimAndConvertArrayToLowerCase(scopeSet)).to.be.deep.eq(["s1", "s2", "s3"]);         
+            expect(clientConfigError instanceof ClientConfigurationError).to.eq(true);
+            expect(clientConfigError.errorCode).to.equal(ClientConfigurationErrorMessage.scopesRequired.code);
+            expect(clientConfigError.message).to.contain(ClientConfigurationErrorMessage.scopesRequired.desc);
+        });
+        
+        it("should throw createScopesNonArrayError if scopes is not an array object", function() {
+            const scopes = {};
```
tnorling, line 101: Can we make this a string of space delimited example scopes as that is a more likely scenario?

lib/msal-core/test/ScopeSet.spec.ts
```diff
@@ -1,16 +1,205 @@
...
+
+    describe("generateLoginScopes", function() {
+        it("should append openid and profile to scopes, remove clientId from scopes if original scopes include clientId", function() {
+            expect(ScopeSet.generateLoginScopes([clientId], clientId)).to.include(openid);
+            expect(ScopeSet.generateLoginScopes([clientId], clientId)).to.include(profile);
+            expect(ScopeSet.generateLoginScopes([clientId], clientId)).to.not.include(clientId);
+        });
+
+        it("should append openid to scopes if original scopes are login scopes and does not include openid", function() {
+            expect(ScopeSet.generateLoginScopes([clientId, profile], clientId)).to.include(openid);
```
tnorling, line 168: Can we also verify profile is still there but clientId is not?

lib/msal-core/test/ScopeSet.spec.ts
```diff
@@ -1,16 +1,205 @@
...
+        });
+        
+        it("should append profile to scopes if original scopes are login scopes and does not include profile", function() {
+            expect(ScopeSet.generateLoginScopes([clientId, openid], clientId)).to.include(profile);
```
tnorling, line 172: Same as above

lib/msal-core/test/UserAgentApplication.spec.ts
```diff
@@ -807,7 +807,7 @@ describe("UserAgentApplication.ts Class", function () {
 
         it("LoginStartPage is cached on acquireTokenRedirect call", (done) => {
             const tokenRequest: AuthenticationParameters = {
-                scopes: ["S1"], 
+                scopes: ["s1"],
```
tnorling, line 810: Are these not getting set to lower case by `trimAndConvertToLowerCase`? Does that usually happen before `AuthenticationParameters` is created?

lib/msal-core/test/utils/RequestUtils.spec.ts
```diff
@@ -4,63 +4,131 @@ import { CryptoUtils } from "../../src/utils/CryptoUtils";
...
+            expect(emptyRequestError.errorCode).to.equal(ClientConfigurationErrorMessage.emptyRequestError.code);
+            expect(emptyRequestError.name).to.equal("ClientConfigurationError");
+            expect(emptyRequestError.stack).to.include("RequestUtils.spec.ts");
+        });
+
+        it("should throw scopesRequiredError if scopes are empty or null", () => {
+            let scopesRequiredError: ClientConfigurationError;
+
+            try {
+                const userRequest: AuthenticationParameters = { };
```
tnorling, line 37: Instead of an empty request can we pass in a real request with just scopes missing? Should apply to any test specifically testing scope validation via `validateRequest` as this would break if the order of validation ever changes.

tnorling (changes requested):

lib/msal-core/src/utils/UrlUtils.ts
```diff
@@ -41,14 +41,11 @@ export class UrlUtils {
...
 
-        if (scopes.indexOf(serverRequestParams.clientId) === -1) {
-            scopes.push(serverRequestParams.clientId);
-        }
         const str: Array<string> = [];
         str.push("response_type=" + serverRequestParams.responseType);
 
-        this.translateclientIdUsedInScope(scopes, serverRequestParams.clientId);
-        str.push("scope=" + encodeURIComponent(ScopeSet.parseScope(scopes)));
+        const loginScopes = ScopeSet.generateLoginScopes(scopes, serverRequestParams.clientId);
```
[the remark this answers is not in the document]
technical-boy: Great catch, that line was always called but the ScopeSet.isLoginScopes method was called inside generateLoginScopes to check and only add the login scopes if it was a login scenario. Now that the login scopes are being added early in validateLoginRequest, we can remove this line. Would you agree?

lib/msal-core/src/ScopeSet.ts
```diff
@@ -141,15 +130,49 @@ export class ScopeSet {
...
+            loginScopes.push(Constants.profileScope);
+        }
+        
+        return loginScopes;
+    }
+
+    /**
+     * 
+     */
+    static isLoginScopes(scopes: Array<string>, clientId: string): boolean {
```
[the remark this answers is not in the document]
technical-boy: This method was used to avoid adding login scopes to non-login scenarios, I removed the only reference to it in last-minute changes. I think, per my previous answer, that we should remove this method, but I'll think it through a bit more and update the PR.

lib/msal-core/test/ScopeSet.spec.ts
```diff
@@ -1,16 +1,205 @@
...
+        });
+
+        it("should return false if cachedScopes and scopes are both empty arrays", () => {
+            const cachedScopes = [];
+            const scopes = [];
+            expect(ScopeSet.isIntersectingScopes(cachedScopes, scopes)).to.eql(false);
+        });
+    });
+
+    describe("trimAndConvertToLowerCase", function() {
```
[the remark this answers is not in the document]
technical-boy: Sure, both the method and tests for these are actually from dev, they only show up as new because I refactored them, but I'll add the test!

lib/msal-core/test/UserAgentApplication.spec.ts
technical-boy: I changed all scopes except the ones in the tests for trimAndConvertToLowerCase to lower case. I'll look over them again and see if there's any other tests where we can capitalize and assume they'll be downcased by trimAndConvert..

lib/msal-core/src/ScopeSet.ts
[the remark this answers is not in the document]
technical-boy: With the latest changes, this method is called within `UserAgentApplication.validateRequest` to determine whether login scopes need to be generated.

lib/msal-core/test/ScopeSet.spec.ts
technical-boy: Done.

lib/msal-core/test/ScopeSet.spec.ts
technical-boy: Done.

lib/msal-core/test/ScopeSet.spec.ts
technical-boy: Done.

lib/msal-core/test/ScopeSet.spec.ts
technical-boy: Done.

lib/msal-core/test/UserAgentApplication.spec.ts
technical-boy: Changed some of these back to confirm, they ARE being trimmed and converted in the scopes validation process before the `AuthenticationParameters are instantiated`.

lib/msal-core/test/utils/RequestUtils.spec.ts
technical-boy: I made the change for this specific test. The non-empty request I added only has a valid authority, which other request elements do you suggest I add? I'm not sure which of those would realistically be added in a token request other than scopes.

tnorling: I would suggest creating a complete and valid request object in a before or beforeEach block and then manipulate the particular field you are testing. e.g. Remove the scope field for this test.

tnorling (approved):

lib/msal-core/src/ScopeSet.ts
```diff
@@ -141,15 +130,49 @@ export class ScopeSet {
...
-        if (reqScopes) {
-            const convertedExtraScopes = reqExtraScopesToConsent ? this.trimAndConvertArrayToLowerCase([...reqExtraScopesToConsent]) : null;
+    static appendScopes(reqScopes: Array<string>, scopesToAppend: Array<string>): Array<string> {
+        if(reqScopes) {
+            const convertedScopesToAppend = scopesToAppend ? this.trimAndConvertArrayToLowerCase([...scopesToAppend]) : null;
             const convertedReqScopes = this.trimAndConvertArrayToLowerCase([...reqScopes]);
-            return convertedExtraScopes ? [...convertedReqScopes, ...convertedExtraScopes] : convertedReqScopes;
+            return convertedScopesToAppend ? [...convertedReqScopes, ...convertedScopesToAppend] : convertedReqScopes;
         }
         return null;
```
jmckennon, line 139: Would it be better to never set `convertedScopesToAppend` to `null` and just leave it empty, and so then we just return the first expression of the ternary? This avoids playing around with `null` more than we need to.

lib/msal-core/src/ScopeSet.ts
```diff
@@ -141,15 +130,49 @@ export class ScopeSet {
...
+        const clientIdIndex: number = loginScopes.indexOf(clientId);
+        
+        if (clientIdIndex >= 0) {
+            // Splice removes clientId from scope set, which will be replaced by openid and profile
+            loginScopes.splice(clientIdIndex, 1);
+        }
+        if (loginScopes.indexOf(Constants.openidScope) === -1) {
+            loginScopes.push(Constants.openidScope);
+        }
+        if (loginScopes.indexOf(Constants.profileScope) === -1) {
```
jmckennon, line 160: I personally think this is more clear/readable, though it has the same functionality: 
`if (!loginScopes.includes(Constants.openidScope))`

jmckennon (changes requested):
Few comments, nice job on the thorough tests.

lib/msal-core/src/ScopeSet.ts
tnorling: I don't think `.includes` works on IE11. I've run into this a few times and .indexOf is the safer method to retain compatibility.

technical-boy: Yes, like @tnorling says, IE11 doesn't support the .includes, so we have to rely on indexOf, unfortunately.

lib/msal-core/src/ScopeSet.ts
technical-boy: @jmckennon if I understood correctly, what you are suggesting is:

```js
    static appendScopes(reqScopes: Array<string>, scopesToAppend: Array<string>): Array<string> {
        if(reqScopes) {
            const convertedReqScopes = this.trimAndConvertArrayToLowerCase([...reqScopes]);
            return [...convertedReqScopes, ...scopesToAppend];
        }
        return null;
    }
```

Assuming my understanding is correct, I see a couple of problems:

1. `scopesToAppend` need to be trimmed and downcased before being appended, which would be skipped in this example
2. Calling the spread operator on null (scopesToAppend) would create an error.

But I feel like I might have misunderstood your suggestion, could you elaborate a bit please?

jmckennon: This is what I meant with my suggestion:
```typescript
    static appendScopes(reqScopes: Array<string>, scopesToAppend: Array<string>): Array<string> {
        if(reqScopes) {
            const convertedScopesToAppend = this.trimAndConvertArrayToLowerCase([...scopesToAppend]);
            const convertedReqScopes = this.trimAndConvertArrayToLowerCase([...reqScopes]);
            return [...convertedReqScopes, ...scopesToAppend];
        }
        return null;
    }
```
Though as you mentioned, if reqScopes is null we'd get an error with the spread operator. I didn't see that previously.

What you've got is fine, approving!

jmckennon (approved):

lib/msal-core/src/ScopeSet.ts
technical-boy: Thanks!

pkanher617 (approved):
Great job here. Could maybe add some comments or docs around scopes, but can be done in a separate PR.

jasonnutter (approved):
Great job @technical-boy and @jo-arroyo!

lib/msal-core/src/utils/RequestUtils.ts
```diff
@@ -75,13 +71,35 @@ export class RequestUtils {
         return validatedRequest;
     }
 
+    /**
+     * 
+     * @param request 
+     * @param clientId 
+     * @param interactionType 
+     */
+    static validateLoginRequest(request: AuthenticationParameters, clientId: string, interactionType: InteractionType): AuthenticationParameters {
```
tnorling, line 80: Do we still need a separate validate function for login vs token requests? Since we're always adding `openid` and `profile` to a request is there a difference?

lib/msal-core/src/ScopeSet.ts
```diff
@@ -141,15 +130,44 @@ export class ScopeSet {
...
     }
 
     // #endregion
 
+    /**
+     * Add openid and profile to scopes array
+     * OIDC required scopes: https://openid.net/specs/openid-connect-basic-1_0.html#Scopes
+     * @param scopes
+     */
+    static generateLoginScopes(scopes: Array<string>): Array<string> {
```
tnorling, line 149: Can we call this something like `addDefaultScopes` since this is no longer Login specific?

technical-boy: How do you feel about `generateOIDCScopes`? I hesitate to call it `default` since they're only default for the request URL and not the request configuration (for response_type definition purposes they're not necessarily included).

lib/msal-core/src/utils/RequestUtils.ts
technical-boy: Good point, as long as we're using `isLoginCall` to determine the `response_type` for login APIs, there's no reason why the OIDC scopes need to be in the array before `createNavigateURL` is called before the request.

lib/msal-core/src/ScopeSet.ts
tnorling: I like that, more specific. To stick to camel case convention: `generateOidcScopes` or maybe `addOpenIdScopes` if you don't like lowercasing the acronym

technical-boy: Yeah, I'll also update the containsOIDCScopes method I added to match the camel case convention (containsOidcScopes)

technical-boy:
Closing in favor of going with the solution implemented in #2022. The approach and original goals of this PR were based on a slight misunderstanding of the problem, so the solution isn't appropriate.

PR #1686, opened 2020-05-21
Remove url from timeout error message, move to errorPii logger

### LINKED ISSUE #1599: Error message contains EUII (login_hint) which compromises Microsoft Privacy Policy

## Library
- [x] `msal@1.2.1`

## Description
We have noticed that in one particular error message, the login_hint is present and that compromises the privacy policy, please take a look into it and see what fix can be done in MSAL.js for that.

Error msg: "URL navigated to is https://login.microsoftonline.com/common/oauth2/v2.0/authorize?response_type=id_token token&scope=394866fc-eedb-4f01-8536-3ff84b16be2a%2F.default%20openid%20profile&client_id=b23dd4db-9142-4734-867f-3577f640ad0c&redirect_uri=https%3A%2F%2Fword-edit.officeapps.live.com%2Fwe%2FWacOAuth.html&state=f382f9de-8979-4d06-9ed8-70b12e2c783d&nonce=b303cd2b-a76d-4484-9e6d-68ee4d2fbe4e&client_info=1&x-client-SKU=MSAL.JS&x-client-Ver=1.2.1&**login_hint=<REDACTED>**&client-request-id=53f30bd6-7f70-48e8-9c1c-00c01f459fe9&prompt=none&response_mode=fragment, Token renewal operation failed due to timeout."

## Security
**This issue is Privacy related, should have high priority.**

## Reproduction steps
1. This error message is tied to timeout.

## Expected behavior
In the error message, login_hint or any other value contains user identifier should be removed.

jasonnutter:
The timeout error returned when polling a popup/window includes the url, which may contain Pii, such as the login_hint. This should be removed, and instead the url will be output to `errorPii` logger.

This also adjusts the log messages to indicate when pii logger is used, and changed a few log messages that could be considered pii.

Fixes: 
- #1599

### DIFF
--- a/lib/msal-core/src/Logger.ts
+++ b/lib/msal-core/src/Logger.ts
@@ -74,10 +74,10 @@ export class Logger {// Singleton Class
         const timestamp = new Date().toUTCString();
         let log: string;
         if (!StringUtils.isEmpty(this.correlationId)) {
-            log = timestamp + ":" + this.correlationId + "-" + libraryVersion() + "-" + LogLevel[logLevel] + " " + logMessage;
+            log = timestamp + ":" + this.correlationId + "-" + libraryVersion() + "-" + LogLevel[logLevel] + (containsPii ? "-pii" : "") + " " + logMessage;
         }
         else {
-            log = timestamp + ":" + libraryVersion() + "-" + LogLevel[logLevel] + " " + logMessage;
+            log = timestamp + ":" + libraryVersion() + "-" + LogLevel[logLevel] + (containsPii ? "-pii" : "") + " " + logMessage;
         }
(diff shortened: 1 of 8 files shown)


tnorling (approved):
All looks reasonable

sameerag (approved):
Lgtm. General comment: is it possible to facilitate `logger` access across the library instead of passing it to every class that needs it?

jasonnutter:
> Lgtm. General comment: is it possible to facilitate `logger` access across the library instead of passing it to every class that needs it?

In theory yes, e.g. attaching it to the `window` object (which isn't great, imo). I would personally prefer to see any approach where the utility classes are instantiate with a logger instance, instead of being static classes/methods.

PR #1687, opened 2020-05-21
[msal-node] hone surface for msal extension library

DarylThayil:
## Plan 
Begin with developer called cache apis

```js
import { PublicClientApplication } from "msal-node"
import { PersistenceCachePlugin } from "msal-extension"


const publicApplication = new PublicClientApplication({
    auth: {...},
    cache: {
        cachePlugin: PersistenceCachePlugin
    }
});
const cacheManager = publicApplication.getCacheManager();

await cacheManager.deserialize(); // read into memmory

// do a bunch of stuff that results in memmory cache changes

await cacheManager.serialize(); //write to persistence

```

Eventually we will instrument public apis / storage with these calls but for the sake of end to end, and still enabling customers who may not want cache calls everytime they call an api, this should be a good starting point.

Msal Extension will expose two functions that `provideWithPersistence` will provide

```js

export interface ICachePlugin {
    readFromStorage: () => Promise<string>;
    writeToStorage: (getMergedState: (oldState: string) => string) => Promise<void>;
}

```

`readFromStorage` will implement reading the persistent cache file and providing it as a return to the lib
`writetoStorage` will do fun stuff, accepting a callback called `getMergedState(stateFromDisk)` which will return a merged state of in memmory and disk.  This mainly entails makingsure removals are respsected and updates and new additions happen.
When the function is envoked a file lock will begin, a read will happen and pass to getMergedState, which will pass back the state to be written, and then the lock will be lifted.

## Msal Node
- readFromPersistence / setCache should over write internal memmory
- writeToPersistance should Lock -> read persistance file -> merge updates and removals -> write file -> unlock
- merge logic lives at this layer so that anyone implementing persistence does not need to worry about this logic.

## Extension
Implents readFromStorage: () => Promise<string>;, writeToStorage: (getMergedState: (oldState: string) => string) => Promise<void>;


lib/msal-node/src/config/Configuration.ts
```diff
@@ -20,12 +19,12 @@ export type NodeAuthOptions = AuthOptions;
...
  * - cacheLocation            - Used to specify the cacheLocation user wants to set. Valid values are "localStorage" and "sessionStorage"
  * - storeAuthStateInCookie   - If set, MSAL store's the auth request state required for validation of the auth flows in the browser cookies. By default this flag is set to false.
+ * - cachePlugin for persistence provided to library
  */
 // TODO Temporary placeholder - this will be rewritten by cache PR.
 export type CacheOptions = {
-    cacheLocation?: string;
     storeAuthStateInCookie?: boolean;
-    cacheInMemory?: InMemoryCache;
+    cacheInMemory?: JsonCache;
```
DarylThayil, line 27: change this to string JsonCache should not be exported

lib/msal-node/src/cache/Storage.ts
```diff
@@ -13,13 +14,23 @@ import { CacheOptions } from '../config/Configuration';
...
     // Cache configuration, either set by user or default values.
-    private cacheConfig: CacheOptions;;
     private inMemoryCache: InMemoryCache;
+    private changeEmitters: Array<Function> = [];
 
     constructor(cacheConfig: CacheOptions) {
-        this.cacheConfig = cacheConfig;
-        if (this.cacheConfig.cacheLocation! === "fileCache")
-            this.inMemoryCache = this.cacheConfig.cacheInMemory!;
+        const instanceConfiguredCache = cacheConfig.cacheInMemory;
```
DarylThayil, line 21: this should be string need to parse somewhere

lib/msal-node/src/cache/CacheManager.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    setPersistence(persistence: ICachePlugin): void {
+        this.persistence = persistence;
+    }
+
+    handleChangeEvent() {
+        this.hasChanged = true;
+    }
+
+    async serialize(): Promise<string> {
```
DarylThayil, line 35: I dont love serialize and deserialize, can have better namings open to suggestions

lib/msal-node/src/cache/CacheManager.ts
```diff
@@ -0,0 +1,73 @@
...
+const defaultSerializedCache: JsonCache = {
+    Account: {},
+    IdToken: {},
+    AccessToken: {},
+    RefreshToken: {},
+    AppMetadata: {}
+};
+
+
+export class CacheManager {
```
DarylThayil, line 14: need to make more stuff private in here

jasonnutter:
Your pseudo code above doesn't align with teh sample code. Are we providing the implementation to MSAL Node and then calling an API on MSAL, or are we providing MSAL Node to the extension library and then calling an API on the extension library?

I would personally advocate strongly for providing the implementation to MSAL Node and calling an API on MSAL node (which invokes the provided implementations from the extension library).

lib/msal-node/src/cache/CacheManager.ts
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    handleChangeEvent() {
+        this.hasChanged = true;
+    }
+
+    async serialize(): Promise<string> {
+        const cache: string = JSON.stringify(Serializer.serializeAllCache(this.storage.getCache()));
+        if (this.persistence) {
+            await this.persistence.writeToStorage(cache);
```
sangonzal, line 38: After thinking about this for a bit, I realized that `await this.persistence.writeToStorage(cache);` needs to be triggered from within `Storage`. The library needs the latest state from disk when doing any operation that modifies the cache, not when serializing or deserializing. 

Essentially, if we are going to follow this model, where msal triggers a `await writeToStorage` or `await readFromStorage()`, Storage has to be made async. I'm not sure we can get around that. The first PR you put out, where it's up to the application to read and write to disk after any MSAL operation that modify the cache is the only way that we can keep synchronous. That PR is missing a couple of things (like hasStateChanged) but I think it could work, although it's not nearly as friendly for customers as this model (it relies on people knowing what MSAL operations alter the cache, because they have to wrap those call in disk reads and writes).

samples/msal-node-auth-code/index.js
```diff
@@ -18,29 +16,55 @@ const publicClientConfig = {
         redirectUri: "http://localhost:3000/redirect",
     },
     cache: {
-        cacheLocation: "fileCache", // This configures where your cache will be stored
-        storeAuthStateInCookie: false, // Set this to "true" if you are having issues on IE11 or Edge
+        storeAuthStateInCookie: false // Set this to "true" if you are having issues on IE11 or Edge
```
sangonzal, line 19: Although not added in this PR, I think we can get rid of `inMemory` and `storeAuthStateInCookie` in the cache configurations.

DarylThayil:
@jasonnutter  originally I thought about doing the passing extension to node model, but after chatting with @sangonzal , It sort of became clear that the write to Storage and readFrom storage were somewhat duplicative of the get cache and set cache functions.  Cache Manager is an attempt to say, there is one way to get cache state from the lib (Cache Manager), so if you hand roll your own persistence, you can use that.  If you use our persistence model, just use the provider from it so that CacheManager apis write and read from it. Seems like a reasonable way to keep the surface low, but the extensibility high.  

it also lets us stay away from altering storage to async

DarylThayil:
@sangonzal  agree that the wrapping is a bit unfortunate, but in async await code it looks much cleaner

```js
await cacheManager.deserialize();
// do stuff
await cacheManager.serialize(); 

```

samples/msal-node-auth-code/index.js
```diff
@@ -54,12 +78,13 @@ app.get('/redirect', (req, res) => {
     pca.acquireTokenByCode(tokenRequest).then((response) => {
         console.log("\nResponse: \n:", response);
         res.send(200);
-        // uncomment this to show writing of cache, dont commit real tokens.
-        // fs.writeFileSync("./data/cache.json", JSON.stringify(pca.readCache()), null, 4);
+        return msalCacheManager.serialize();
```
DarylThayil, line 81: add check for cacheHasChanged

lib/msal-node/src/cache/CacheManager.ts
```diff
@@ -0,0 +1,73 @@
...
+        }
+        this.hasChanged = false;
+        return cache;
+    }
+
+    async deserialize(cache?: string): Promise<void> {
```
DarylThayil, line 44: @jasonnutter  thoughts on better names for deserialize and serialize?

sangonzal:
> @sangonzal agree that the wrapping is a bit unfortunate, but in async await code it looks much cleaner
> 
> ```js
> await cacheManager.deserialize();
> // do stuff
> await cacheManager.serialize(); 
> ```

Understood, my concern is that it leaks out some details that customers don't necessarily have to be aware of. With this model, callers have to always wrap msal code in `deserialize()` and `serialize()`. 

This brings the question, if we are asking people to always do this, why can't we just do this in MSAL?  Essentially the only thing that's keeping us from incorporating this into MSAL is keeping Storage synchronous. 

All to say, we are making a tradeoff here that we should be aware of- Keeping storage synchronous (because we want to share cache code with msal- browser) at the expense of a more complicated Node API.

lib/msal-node/src/cache/CacheManager.ts
DarylThayil: leaving for now with the goal of possibly separating into two separate apis

DarylThayil:
@sangonzal  and I spent a ton of time detailing what needs to happen and think we have a good understanding

lib/msal-node/src/config/Configuration.ts
DarylThayil: removed

lib/msal-node/src/cache/CacheManager.ts
```diff
@@ -0,0 +1,94 @@
...
+        this.hasChanged = true;
+    }
+
+    mergeState(oldState: JsonCache, currentState: JsonCache) {
+        // TODO
+        // mergeUpdates(old, new)
+        // mergeRemovals(old, new)
+        return {
+            ...oldState,
+            ...currentState
```
jasonnutter, line 44: This does a shallow merge, is that okay?

lib/msal-node/src/cache/CacheManager.ts
```diff
@@ -0,0 +1,94 @@
...
+        }
+        this.hasChanged = false;
+        return cache;
+    }
+
+    async deserialize(cache?: string): Promise<void> {
+        let deserializedCache: InMemoryCache;
+        if (this.persistence) {
+            const stringCacheFromStorage = await this.persistence.readFromStorage();
+            deserializedCache = Deserializer.deserializeAllCache(this.overlayDefaults(JSON.parse(stringCacheFromStorage)));
```
jasonnutter, line 71: Why is this code path ignoring `cache`?

lib/msal-node/src/cache/CacheManager.ts
```diff
@@ -0,0 +1,94 @@
...
+    RefreshToken: {},
+    AppMetadata: {}
+};
+
+
+export class CacheManager {
+
+    hasChanged: boolean;
+    storage: Storage;
+    persistence: ICachePlugin;
```
jasonnutter, line 18: `private`?

samples/msal-node-auth-code/index.js
```diff
@@ -4,12 +4,42 @@
...
-const myLocalCache = require("./data/cache");
 const fs = require("fs");
 
 const SERVER_PORT = process.env.PORT || 3000;
 
-// initialize msal public client application
+
+const readFromStorage = () => {
+    return new Promise((resolve, reject) => {
+        fs.readFile("./data/cache.json", (err, data) => {
```
jasonnutter, line 14: Can these use the fs promise APIs instead?

samples/msal-node-auth-code/index.js
```diff
@@ -4,12 +4,42 @@
...
+                }
+            });
+        });
+    });
+};
+
+const cachePlugin = {
+    readFromStorage: readFromStorage,
+    writeToStorage: writeToStorage
+};
```
jasonnutter, line 41:
```suggestion
const cachePlugin = {
    readFromStorage,
    writeToStorage
};
```

samples/msal-node-auth-code/index.js
```diff
@@ -18,29 +48,32 @@ const publicClientConfig = {
         redirectUri: "http://localhost:3000/redirect",
     },
     cache: {
-        cacheLocation: "fileCache", // This configures where your cache will be stored
         storeAuthStateInCookie: false, // Set this to "true" if you are having issues on IE11 or Edge
+        cachePlugin: cachePlugin
```
jasonnutter, line 52:
```suggestion
        cachePlugin
```

samples/msal-node-auth-code/index.js
```diff
@@ -18,29 +48,32 @@ const publicClientConfig = {
...
 };
 const pca = new msal.PublicClientApplication(publicClientConfig);
-pca.initializeCache(myLocalCache);
+const msalCacheManager = pca.getCacheManager();
+
 
 // Create Express App and Routes
 const app = express();
 
 app.get('/',  (req, res) => {
```
jasonnutter, line 62: You could mark this as `async` and then use `async/await` below.

samples/msal-node-auth-code/index.js
jasonnutter:
```js
import { promises as fs } from "fs";

const readFromStorage = () => fs.readFile("./data/cache.json");
```

samples/msal-node-auth-code/index.js
```diff
@@ -4,12 +4,42 @@
...
+                resolve(data);
+            }
+        });
+    });
+};
+const writeToStorage = (getMergedState) => {
+    return new Promise((resolve, reject) => {
+        readFromStorage().then(oldFile => {
+            const mergedState = getMergedState(oldFile);
+            fs.writeFile("./data/cache.json", mergedState, (err, data) => {
```
jasonnutter, line 27:
```js
import { promises as fs } from "fs";

const writeToStorage = async (getMergedState) => {
  const oldFile = await readFromStorage();
  const mergedState = getMergedState(oldFile);
  return fs.writeFile("./data/cache.json");
}

File: lib/msal-node/src/cache/Storage.ts [hunk 17]
```diff
@@ -66,7 +79,11 @@ export class Storage implements ICacheStorage {
      * TODO: implement after the lookup implementation
      */
     removeItem(key: string): void {
-        if (!key) return;
+        if (!key) {
+            this.emitChange();
+            return;
+        }
+        this.emitChange();
```
jasonnutter, line 86: The `if` seems useless, as both code paths are doing the same thing?

lib/msal-node/src/cache/Storage.ts
```diff
@@ -45,8 +56,10 @@ export class Storage implements ICacheStorage {
      */
     setItem(key: string, value: string): void {
         if (key && value) {
+            this.emitChange();
             return;
         }
+        this.emitChange();
```
jasonnutter, line 62: Same thing here?

lib/msal-node/src/cache/Storage.ts
```diff
@@ -6,20 +6,30 @@ import {
...
+        idTokens: {}
+    };
+    private changeEmitters: Array<Function> = [];
 
-    constructor(cacheConfig: CacheOptions) {
-        this.cacheConfig = cacheConfig;
-        if (this.cacheConfig.cacheLocation! === "fileCache")
-            this.inMemoryCache = this.cacheConfig.cacheInMemory!;
+    constructor() {
+    }
```
jasonnutter, line 25: Not needed if empty.

lib/msal-node/src/cache/CacheManager.ts
```diff
@@ -0,0 +1,94 @@
...
+        let deserializedCache: InMemoryCache;
+        if (this.persistence) {
+            const stringCacheFromStorage = await this.persistence.readFromStorage();
+            deserializedCache = Deserializer.deserializeAllCache(this.overlayDefaults(JSON.parse(stringCacheFromStorage)));
+        } else if (cache) {
+            deserializedCache = Deserializer.deserializeAllCache(this.overlayDefaults(JSON.parse(cache)));
+        } else {
+            throw Error("Cache Must be passed in or configured");
+        }
+        this.storage.setCache(deserializedCache)
```
jasonnutter, line 77: Nit pick: not a big fan of mutation here, as it doesn't really add much value.

```js
if (this.persistence) {
  const stringCacheFromStorage = await this.persistence.readFromStorage();
  const deserlizedCache = Deserializer.deserializeAllCache(this.overlayDefaults(JSON.parse(stringCacheFromStorage)));
  this.storage.setCache(deserializedCache);
  return;
}

if (cache) {
  const deserializedCache = Deserializer.deserializeAllCache(this.overlayDefaults(JSON.parse(cache)));
  this.storage.setCache(deserializedCache);
  return;
}

throw Error("Cache Must be passed in or configured");

File: lib/msal-node/src/cache/CacheManager.ts (hunk 10 above)
Comment (jasonnutter) [reply lib/msal-node/src/cache/CacheManager.ts:71]: And do we need local error handling for `JSON.parse`?

File: lib/msal-node/src/cache/CacheManager.ts [hunk 21]
```diff
@@ -0,0 +1,94 @@
...
+            ...currentState
+        };
+    }
+
+    //TODO think about separating serialize / deserialize from writeToPersistance and readFromPersistance
+
+    async serialize(): Promise<string> {
+        const cache: string = JSON.stringify(Serializer.serializeAllCache(this.storage.getCache()));
+        if (this.persistence) {
+            await this.persistence.writeToStorage((stateFromDisk) => {
```
jasonnutter, line 53: I would argue not to put the callback inline, makes it difficult to read:

```js
if (this.persistence) {
  const getMergedState = (stateFromDisk) => {
    const jsonFromDisk = JSON.parse(stateFromDisk);
    const allCachedSerliazed = Serializer.serializeAllCache(this.storage.getCache());
    const mergedState = this.mergeState(jsonFromDisk, allCachedSerliazed);
    return JSON.stringify(mergedState, null, 4);
  }
  await this.persistence.writeToStorage(getMergedState);
}

Comment (jasonnutter) [reply lib/msal-node/src/cache/CacheManager.ts:53]: Also, combining callbacks and promises like this makes for a weird API, and I'm not sure I'm a fan of the ergonomics.

File: lib/msal-node/src/cache/CacheManager.ts (hunk 8 above)
Comment (jasonnutter) [reply lib/msal-node/src/cache/CacheManager.ts:67]: `deserializeCache` maybe to be more specific? Or is that redundant?

File: samples/msal-node-auth-code/index.js (hunk 6 above)
Comment (jasonnutter) [reply samples/msal-node-auth-code/index.js]: `storeAuthStateInCookie` is a browser concept, and can be removed for Node.

File: lib/msal-node/src/cache/CacheManager.ts (hunk 9 above)
Comment (DarylThayil) [reply lib/msal-node/src/cache/CacheManager.ts:44]: Yeah in the long run this is going to be deep when mergeRemovals and Updates is added.
I may add that in this pr to just get in in

File: lib/msal-node/src/cache/CacheManager.ts (hunk 10 above)
Comment (DarylThayil) [reply lib/msal-node/src/cache/CacheManager.ts:71]: Yeah I should add this.  I probably need to do an error handling pass period

File: samples/msal-node-auth-code/index.js (hunk 13 above)
Comment (DarylThayil) [reply samples/msal-node-auth-code/index.js:41]: does this work in the sample without and transpilation?

Comment (jasonnutter) [reply samples/msal-node-auth-code/index.js:41]: Latest versions of Node support this out-of-the-box.

File: lib/msal-node/src/cache/CacheManager.ts (hunk 11 above)
Comment (sameerag) [reply lib/msal-node/src/cache/CacheManager.ts:18]: Will `hasChanged` be marked for any update in `inMemoryCache`?

File: lib/msal-node/src/cache/CacheManager.ts [hunk 22]
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+    }
+
+    setPersistence(persistence: ICachePlugin): void {
+        this.persistence = persistence;
+    }
+
+    handleChangeEvent() {
+        this.hasChanged = true;
```
sameerag, line 35: Where are we resetting this? After write?

lib/msal-node/src/cache/CacheManager.ts
sameerag: `serialize` and `deserialize` are used across MSAL libraries. The name is used mostly for users who are familiar with serialized persistence. Just FYI.

lib/msal-node/src/cache/CacheManager.ts
```diff
@@ -0,0 +1,94 @@
...
+    //TODO think about separating serialize / deserialize from writeToPersistance and readFromPersistance
+
+    async serialize(): Promise<string> {
+        const cache: string = JSON.stringify(Serializer.serializeAllCache(this.storage.getCache()));
+        if (this.persistence) {
+            await this.persistence.writeToStorage((stateFromDisk) => {
+                const jsonFromDisk = JSON.parse(stateFromDisk);
+                return JSON.stringify(
+                    this.mergeState(
+                        jsonFromDisk,
```
sameerag, line 57: is jsonFromDisk of the format `JsonCache`? I think so but would like to test this once. Also if the disk cache is empty JSON.parse() would not return `{}`, it would return `null`; we may need to handle it.

samples/msal-node-auth-code/index.js
DarylThayil: removing

lib/msal-node/src/cache/CacheManager.ts
DarylThayil: leaving for now

lib/msal-node/src/cache/Storage.ts
DarylThayil: removing

lib/msal-node/src/cache/CacheManager.ts
DarylThayil: resolving

lib/msal-node/src/cache/CacheManager.ts
DarylThayil: fixed

DarylThayil: Not sure there is a better ergonomic solution here, and its an internal implementation a customer never really deals with

lib/msal-node/src/cache/CacheManager.ts
DarylThayil: In the serialize function after the cache has been written

lib/msal-node/src/cache/CacheManager.ts
DarylThayil: it is json cache but that type is not exposed to dev 
, I think extension portion should handle what to pass here since it knows the state of the file. 
I changed to initialize to empty object and throw on bad parse

lib/msal-node/src/cache/CacheManager.ts
DarylThayil: added error handling to json parse

lib/msal-node/src/cache/CacheManager.ts
DarylThayil: updated to private, and yes it will be. Storage will trigger changes when it makes changes in memmory

samples/msal-node-auth-code/index.js
DarylThayil: updated

lib/msal-node/src/config/Configuration.ts
```diff
@@ -20,12 +19,11 @@ export type NodeAuthOptions = AuthOptions;
  *
  * - cacheLocation            - Used to specify the cacheLocation user wants to set. Valid values are "localStorage" and "sessionStorage"
  * - storeAuthStateInCookie   - If set, MSAL store's the auth request state required for validation of the auth flows in the browser cookies. By default this flag is set to false.
+ * - cachePlugin for persistence provided to library
  */
 // TODO Temporary placeholder - this will be rewritten by cache PR.
 export type CacheOptions = {
-    cacheLocation?: string;
-    storeAuthStateInCookie?: boolean;
```
sangonzal, line 27: nit: delete comments for the options that we are removing (line 21-22)

lib/msal-node/src/cache/Storage.ts
sameerag: Code changed. Closing this comment.

lib/msal-node/src/cache/Storage.ts
sameerag: Edited.

lib/msal-common/src/unifiedCache/utils/CacheTypes.ts
```diff
@@ -18,6 +18,7 @@ export type RefreshTokenCache = { [key: string]: RefreshTokenEntity };
 export type AppMetadataCache = { [key: string]: AppMetadataEntity };
 
 export type JsonCache = {
+    [key: string]: StringDict;
```
jasonnutter, line 21: What is this? Is this allow artbitrary nested StrictDict object?

lib/msal-node/src/cache/CacheManager.ts
```diff
@@ -0,0 +1,221 @@
...
+
+                if (newValueNotNull && newValueIsObject && newValueIsNotArray) {
+                    this.mergeUpdates(oldState[newKey], newValue);
+                } else {
+                    oldState[newKey] = newValue;
+                }
+            }
+        });
+
+        return oldState;
```
jasonnutter, line 168: Instead of mutating the passed in parameter (which lint rules will not allow), can you construct a new object?

sangonzal: Yep, will update.

lib/msal-node/src/cache/CacheManager.ts
```diff
@@ -0,0 +1,221 @@
...
+            // if entity is in oldState but not in newState remove it
+            if (oldEntries != null) {
+                Object.keys(oldEntries).forEach((oldKey) => {
+                    if (!newEntries || !(newEntries.hasOwnProperty(oldKey))) {
+                        delete oldEntries[oldKey];
+                    }
+                })
+            }
+        });
+        return oldState;
```
jasonnutter, line 194: Same comment about mutating passed in value.

sameerag:
Closing as this merged with #1801

PR #1691, opened 2020-05-22
Add onRedirectNavigate callback to stop navigatation and get redirect url

### LINKED ISSUE #1644: Feature Request: API for retrieving the redirect url for MSAL.AcquireTokenRedirect
Scenario:
1. Silent Authentication fails with Multi Factor Authentication Required
2. SPA prompts the end user to provide credentials (rather than redirecting immediately)

This scenario is difficult to support, as we cannot regenerate the original token request (via AcquireTokenRedirect) when the user clicks the hyperlink. If we had an API to build the redirect url from a request, we could pass it to our UX element.

jasonnutter:
This adds a new callback for `loginRedirect` and `acquireTokenRedirect` to allow applications to get the URL that would be navigated to, and to stop navigation. This is to support scenarios where applications want to perform the redirect themselves, instead of relying on MSAL.

Usage:

```js
msal.loginRedirect({
  onRedirectNavigate: url => {
    // url that would be navigated to
    console.log(url);
   
     // Applications must explicitly return false to stop navigation
     return false;
  }
})
```

### DIFF
--- a/lib/msal-core/src/AuthenticationParameters.ts
+++ b/lib/msal-core/src/AuthenticationParameters.ts
@@ -25,7 +25,8 @@ export type AuthenticationParameters = {
     forceRefresh?: boolean;
     redirectUri?: string;
     redirectStartPage?: string;
-    authorityMetadata?: string
+    authorityMetadata?: string;
+    onRedirectNavigate?: ((url: string) => void | boolean)
 };
 
 export function validateClaimsRequest(request: AuthenticationParameters) {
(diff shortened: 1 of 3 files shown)


DarylThayil (approved):

tnorling (approved):
LGTM

Inframatic:
How could one transform that received redirect URL and pass that in as the new redirect url?

My purpose would be editing the redirect url to remove the `login_hint`

PR #1699, opened 2020-05-22
Stale bot

tnorling:
Auto label issues and PRs with a stale label after 14 days of inactivity.
Runs in debug-only mode for now. Can remove flag later if it works.

https://github.com/actions/stale


jasonnutter:
Big fan!

PR #1701, opened 2020-05-22
Auto PR Labeler

tnorling:
Github Workflow for labeling our PRs based on which files they touch. Could be interesting to try out instead of marking it in the title. 

Labels for:
- Node
- Angular
- Angularjs
- msal-browser
- msal-common
- msal@1.x
- samples

Feel free to share thoughts on this or suggest additional labels

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,12 @@
+node:
+  - lib/msal-node/**/*
+angular:
+  - lib/msal-angular/**/*
+msal@1.x:
+  - lib/msal-core/**/*
+msal-common:
+  - lib/msal-common/**/*
+msal-browser:
+  - lib/msal-browser/**/*
+samples:
+  - samples/**/*
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -2,6 +2,8 @@ node:
   - lib/msal-node/**/*
 angular:
   - lib/msal-angular/**/*
+angularjs:
+  - lib/msal-angularjs/**/*
 msal@1.x:
   - lib/msal-core/**/*
 msal-common:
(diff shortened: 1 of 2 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,12 @@
+node:
+  - lib/msal-node/**/*
+angular:
+  - lib/msal-angular/**/*
```
jasonnutter, line 4: Can you add AngularJS?

.github/labeler.yml
```diff
@@ -0,0 +1,12 @@
+node:
+  - lib/msal-node/**/*
+angular:
```
jasonnutter, line 3: There is already an Angular label, is that okay?

tnorling: Yes. All of these already exist. If it doesn't work it should get logged in the Actions tab

.github/labeler.yml
tnorling: Done

.github/labeler.yml
jasonnutter: Ah right.

.github/workflows/label.yml
```diff
@@ -0,0 +1,19 @@
...
+
+jobs:
+  label:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/labeler@v2
+      with:
+        repo-token: "${{ secrets.GITHUB_TOKEN }}
```
sameerag, line 19: Do we also need individuals to get assigned an issue in this same PR? Or are we scraping that?

sameerag: Never mind! I see a new PR by @jasonnutter on that.

PR #1711, opened 2020-05-27
[msal-node][#1] SilentFlow support in msal-common

sameerag:
This PR adds support for SilentFlow in `msal-common`

This unblocks the below use cases for the end user:
* Ability to call `acquireTokenSilent` from `msal-browser` and `msal-node` when the user has knowledge of the cached `account`
* If the user chooses to `forceRefresh` or if the `accessToken` has expired, acquires a token with `RefreshTokenFlow`. This is achieved by calling the service with the cached `refreshToken` to acquire a new set of tokens.


lib/msal-common/src/client/SilentFlowClient.ts
```diff
@@ -0,0 +1,152 @@
...
+     * @param request
+     */
+    public async acquireToken(request: SilentFlowRequest): Promise<AuthenticationResult>{
+        let cacheRecord: CacheRecord;
+        let idTokenObj: IdToken;
+        const requestScopes = new ScopeSet(request.scopes || [], this.config.authOptions.clientId, true);
+
+        // We currently do not support silent flow for account===null use cases; This will be revisited for confidential flow usecases
+        if (request.account === null) {
+            throw ClientAuthError.createNoAccountInSilentRequestError();
```
pkanher617, line 41: Are we going to ask users to pass the relevant account object, or will that be handled by the platform libraries?

lib/msal-browser/test/cache/BrowserStorage.spec.ts
```diff
@@ -16,7 +16,7 @@ class TestCacheStorage implements ICacheStorage {
     getItem(key: string): string {
         throw new Error("Method not implemented.");
     }
-    removeItem(key: string): void {
+    removeItem(key: string): boolean {
```
jasonnutter, line 19: Why boolean?

lib/msal-browser/test/interaction_handler/PopupHandler.spec.ts
```diff
@@ -88,14 +88,8 @@ describe("PopupHandler.ts Unit Tests", () => {
...
-                getCache: (): InMemoryCache => {
-                    return {
-                        accounts: {},
-                        idTokens: {},
-                        accessTokens: {},
-                        refreshTokens: {},
-                        appMetadata: {},
-                    };
+                getCache: (): object => {{
+                    return {};}
```
jasonnutter, line 92: `getCache: (): object => ({})`

lib/msal-common/src/unifiedCache/utils/CacheHelper.ts
```diff
@@ -159,35 +153,117 @@ export class CacheHelper {
...
+    }
+
+    /**
+     * Generate target key component as per schema: <target>
+     */
+    private static generateTargetForCacheKey(scopes: string): string {
+        return (scopes || "").toLowerCase();
+    }
+
+    static toIAccount(accountObj: AccountEntity): IAccount {
```
jasonnutter, line 261: I would argue methods like this should construct instances of implementations of `IAccount`, but not instances of `IAccount` itself. Is there are a reason this isn't constructing one of the implementatinos of `IAccount`?

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -92,7 +94,7 @@ export class ResponseHandler {
         if (serverResponse.client_info) {
             this.clientInfo = buildClientInfo(serverResponse.client_info, this.cryptoObj);
             if (!StringUtils.isEmpty(this.clientInfo.uid) && !StringUtils.isEmpty(this.clientInfo.utid)) {
-                this.homeAccountIdentifier = this.cryptoObj.base64Encode(this.clientInfo.uid) + "." + this.cryptoObj.base64Encode(this.clientInfo.utid);
+                this.homeAccountIdentifier = this.clientInfo.uid + "." + this.clientInfo.utid;
```
jasonnutter, line 97: Can you use string template instead?

lib/msal-common/src/account/IAccount.ts
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+export interface IAccount {
```
jasonnutter, line 6: This seems be used as a type, instead of an interface (you are constructing instances of it). I would prefer to see a concrete type that implements this interface, or change this to a type (and rename to `Account` or something like that.

lib/msal-common/package.json
```diff
@@ -47,7 +47,7 @@
         "test:coverage": "nyc --reporter=text mocha --exit",
         "test:coverage:only": "npm run clean:coverage && npm run test:coverage",
         "build:modules": "rollup -c",
-        "build:modules:watch": "rollup -cw",
+        "build:modules:watch": "npm run lint && rollup -cw",
```
jasonnutter, line 50: I would like to have the build tasks just perform the build step, and not perform other tasks.

jasonnutter: That way if I just want to build and not worry about lint, I can.

lib/msal-common/test/unifiedCache/mockStorage.ts
```diff
@@ -0,0 +1,286 @@
...
+                        break;
+                    }
+                    case CredentialType.REFRESH_TOKEN: {
+                        credential =
+                            (cache.refreshTokens[key] as RefreshTokenEntity) ||
+                            null;
+                        break;
+                    }
+                }
+                return credential!;
```
jasonnutter, line 147: Instead of breaking and mutating `credential` in each `case`, you can just return in each case.

lib/msal-common/src/client/SilentFlowClient.ts
```diff
@@ -0,0 +1,161 @@
...
+                    return refreshTokenClient.acquireToken(refreshTokenRequest);
+                }
+                // no refresh Token
+                else {
+                    throw ClientAuthError.createNoTokenInCacheError();
+                }
+            }
+
+            if (cacheRecord.accessToken === null) {
+                throw ClientAuthError.createNoTokenInCacheError();
```
jasonnutter, line 77: Why throw here? Shouldn't we try a network request?

lib/msal-common/src/client/SilentFlowClient.ts
```diff
@@ -0,0 +1,161 @@
...
+
+                    return refreshTokenClient.acquireToken(refreshTokenRequest);
+                }
+                // no refresh Token
+                else {
+                    throw ClientAuthError.createNoTokenInCacheError();
+                }
+            }
+
+            if (cacheRecord.accessToken === null) {
```
pkanher617, line 76: Should be handling the null access token with a refresh token call.

lib/msal-common/src/client/SilentFlowClient.ts
```diff
@@ -0,0 +1,161 @@
...
+     */
+    public async acquireToken(request: SilentFlowRequest): Promise<AuthenticationResult>{
+        let cacheRecord: CacheRecord;
+        let idTokenObj: IdToken;
+        const requestScopes = new ScopeSet(request.scopes || [], this.config.authOptions.clientId, true);
+
+        // We currently do not support silent flow for account === null use cases; This will be revisited for confidential flow usecases
+        if (request.account === null) {
+            throw ClientAuthError.createNoAccountInSilentRequestError();
+        } else {
```
jasonnutter, line 43: `else` is not needed. This allows you put the declarations `cacheRecord` and `idTokenObj` below in scope.

lib/msal-common/src/client/SilentFlowClient.ts
```diff
@@ -0,0 +1,161 @@
...
+                        scopes: request.scopes,
+                        refreshToken: cacheRecord.refreshToken.secret,
+                        authority: request.authority
+                    };
+
+                    return refreshTokenClient.acquireToken(refreshTokenRequest);
+                }
+                // no refresh Token
+                else {
+                    throw ClientAuthError.createNoTokenInCacheError();
```
jasonnutter, line 72: I think this should be a unique error relating to not having a valid refresh token.

pkanher617, line 72: Create a new type of error for no valid refresh tokens.

lib/msal-common/src/client/SilentFlowClient.ts
```diff
@@ -0,0 +1,161 @@
...
+        );
+        return this.unifiedCacheManager.getCredential(idTokenKey) as IdTokenEntity;
+    }
+
+    /**
+     * fetches accessToken from cache if present
+     * @param request
+     * @param scopes
+     */
+    private fetchAccessToken(homeAccountId: string, environment: string, scopes: ScopeSet, inputRealm: string): AccessTokenEntity {
```
jasonnutter, line 116: This doesn't feel like the right name, as it implies that a fetch request is involved. What about `readAccessTokenFromCache`?

lib/msal-common/src/client/SilentFlowClient.ts
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    /**
+     * Retrieves a token from cache if it is still valid, or uses the cached refresh token to renew
+     * the given token and returns the renewed token
+     * @param request
+     */
+    public async acquireToken(request: SilentFlowRequest): Promise<AuthenticationResult>{
+        let cacheRecord: CacheRecord;
+        let idTokenObj: IdToken;
```
jasonnutter, line 37: These can be defined below as `const`

lib/msal-common/src/client/SilentFlowClient.ts
```diff
@@ -0,0 +1,161 @@
...
+        const requestScopes = new ScopeSet(request.scopes || [], this.config.authOptions.clientId, true);
+
+        // We currently do not support silent flow for account === null use cases; This will be revisited for confidential flow usecases
+        if (request.account === null) {
```
jasonnutter, line 41: Can this be `if (!request.account)`? Do you need to explicitly check for `null`?

lib/msal-common/src/client/SilentFlowClient.ts
jasonnutter: Also, you should mark this as potentially returning `null`.

jasonnutter: We should turn on `strictNullChecks` ASAP.

lib/msal-common/src/response/SPAResponseHandler.ts
```diff
@@ -229,7 +229,7 @@ export class SPAResponseHandler {
 
         // Retrieve current id token object
         let idTokenObj: IdToken;
-        const cachedIdToken: string = this.cacheStorage.getItem(PersistentCacheKeys.ID_TOKEN);
+        const cachedIdToken: string = this.cacheStorage.getItem(PersistentCacheKeys.ID_TOKEN) as string;
```
jasonnutter, line 232: Is `as string` needed? Looks like `getItem` returns a string.

jasonnutter: If `getItem` can return multiple types, I would argue for generics instead of making consumers cast.

lib/msal-common/src/response/AuthenticationResult.ts
```diff
@@ -12,6 +13,7 @@ export class AuthenticationResult {
     uniqueId: string;
     tenantId: string;
     scopes: Array<string>;
+    account: IAccount;
     idToken: string;
     idTokenClaims: StringDict;
```
jasonnutter, line 18: FYI We found in v1 that this is not the type we should be using for this. See #1766

lib/msal-common/src/client/SilentFlowClient.ts
jasonnutter: Also, can you move checks for error conditions to the top? That way you don't need `else`.

lib/msal-common/src/client/SilentFlowClient.ts
```diff
@@ -0,0 +1,161 @@
...
+            cacheRecord.account = this.unifiedCacheManager.getAccount(accountKey);
+
+            const homeAccountId = cacheRecord.account.homeAccountId;
+            const environment = cacheRecord.account.environment;
+
+            // fetch idToken, accessToken, refreshToken
+            cacheRecord.idToken = this.fetchIdToken(homeAccountId, environment, cacheRecord.account.realm);
+            idTokenObj = new IdToken(cacheRecord.idToken.secret, this.config.cryptoInterface);
+            cacheRecord.accessToken = this.fetchAccessToken(homeAccountId, environment, requestScopes, cacheRecord.account.realm);
+            cacheRecord.refreshToken = this.fetchRefreshToken(homeAccountId, environment);
```
jasonnutter, line 55: Adding a constructor to `CacheRecord` could make this code cleaner and easier to read/understand:

```js
const account = this.unifiedCacheManager.getAccount(accountKey);

const idToken = this.fetchIdToken(account.homeAccountId, account.environment, cacheRecord.account.realm);
const idTokenObj =  new IdToken(idToken.secret, this.config.cryptoInterface);

const accessToken = this.fetchAccessToken(account.homeAccountId, account.environment, requestScopes, account.realm);
const refreshToken = this.fetchRefreshToken(account.homeAccountId, account.environment);

const cacheRecord = new CacheRecord(idToken, accessToken, refreshToken);
```

lib/msal-common/src/client/SPAClient.ts
```diff
@@ -66,13 +82,24 @@ export class SPAClient extends BaseClient {
...
+                ? AuthorityFactory.createInstance(
+                      request.authority,
+                      this.networkClient
+                  )
+                : this.defaultAuthority;
 
         // This is temporary. Remove when ADFS is supported for browser
-        if(acquireTokenAuthority.authorityType == AuthorityType.Adfs){
-            throw ClientAuthError.createInvalidAuthorityTypeError(acquireTokenAuthority.canonicalAuthority);
+        if (acquireTokenAuthority.authorityType == AuthorityType.Adfs) {
```
jasonnutter, line 99: `===` (there will be a lint rule about this soon)

lib/msal-common/src/client/SilentFlowClient.ts
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    /**
+     * check if an access token is expired
+     * @param expiresOn
+     */
+    private isTokenExpired(expiresOn: string): boolean {
+        // check for access token expiry
+        const expirationSec = Number(expiresOn);
+        const offsetCurrentTimeSec = TimeUtils.nowSeconds() + this.config.systemOptions.tokenRenewalOffsetSeconds;
```
sangonzal, line 156: Should this be `TimeUtils.nowSeconds() - this.config.systemOptions.tokenRenewalOffsetSeconds;`? Otherwise you will return expired tokens.

lib/msal-browser/test/cache/BrowserStorage.spec.ts
jasonnutter: Would it be more helpful if an exception was thrown instead?

lib/msal-common/src/client/SilentFlowClient.ts
pkanher617: The offset is used to renew the tokens at a specified amount of time before the actual expiration. I think the actual boolean check needs to be flipped, but the logic itself is correct.

lib/msal-common/src/client/BaseClient.ts
```diff
@@ -110,22 +124,33 @@ export abstract class BaseClient {
...
 
     /**
-     * TODO: modify this soon
-     * Set the cache post acquireToken call
+     * Get all currently signed in accounts.
      */
-    protected updateCache(): void {
-        const cache = this.unifiedCacheManager.getCacheInMemory();
-        this.cacheStorage.setCache(cache);
+    public getAllAccounts(): IAccount[] {
```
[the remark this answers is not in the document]
pkanher617: Already resolved this, we would like to prevent the UnifiedCacheManager from being exposed as a public interface.

lib/msal-common/src/client/BaseClient.ts
```diff
@@ -110,22 +124,33 @@ export abstract class BaseClient {
...
-        return this.networkClient.sendPostRequestAsync<ServerAuthorizationTokenResponse>(
-            tokenEndpoint,
-            {
-                body: queryString,
-                headers: headers,
-            });
+        headers: Map<string, string>
+    ): Promise<NetworkResponse<ServerAuthorizationTokenResponse>> {
+        return this.networkClient.sendPostRequestAsync<
+            ServerAuthorizationTokenResponse
```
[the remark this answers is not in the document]
pkanher617: These changes are from someone's vscode formatting, I can't keep changing them back.

lib/msal-common/src/client/BaseClient.ts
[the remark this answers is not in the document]
sangonzal: Resolved where?

lib/msal-common/package.json
pkanher617: I don't know why this was added to the watch command, but build:modules and build:modules:watch will let you build without linting, but I would like a command to run during the pipeline that also does lint checks.

lib/msal-common/test/unifiedCache/mockStorage.ts
pkanher617: I will fix this in a later PR.

lib/msal-common/src/client/SilentFlowClient.ts
pkanher617: Agreed, we should only throw if no refresh token is found.

lib/msal-common/src/client/SilentFlowClient.ts
pkanher617: done.

lib/msal-common/src/client/SilentFlowClient.ts
pkanher617: done.

lib/msal-common/src/client/SilentFlowClient.ts
pkanher617: done.

lib/msal-common/src/client/SilentFlowClient.ts
pkanher617: Will turn on strict null checks after next beta release.

pkanher617: renamed function.

lib/msal-common/src/response/ResponseHandler.ts
pkanher617: done.

lib/msal-common/src/client/SilentFlowClient.ts
pkanher617: done, and I removed instances of CacheRecord from these files since we don't really need them.

lib/msal-common/src/client/SilentFlowClient.ts
pkanher617: changed.

lib/msal-browser/test/interaction_handler/PopupHandler.spec.ts
pkanher617: done.

lib/msal-common/src/response/AuthenticationResult.ts
pkanher617: We can resolve this as a separate PR.

lib/msal-common/src/response/SPAResponseHandler.ts
pkanher617: This will be resolved in a future PR when we add functionality for consumer driven storage.

lib/msal-common/src/account/IAccount.ts
pkanher617: Changed in a future PR, will be up shortly.

lib/msal-common/src/client/SilentFlowClient.ts
pkanher617: Future PR.

lib/msal-browser/test/cache/BrowserStorage.spec.ts
pkanher617: We can revisit this at another date.

lib/msal-common/src/unifiedCache/utils/CacheHelper.ts
pkanher617: No longer user interface, IAccount is now a type called AccountInfo. See a future PR.

lib/msal-common/package.json
jasonnutter: I think lint should be added as a separate task (which I started doing for #1700).

PR #1751, opened 2020-06-05
Update CacheManager. Update auth code sample.

sangonzal:


lib/msal-node/src/cache/CacheManager.ts
```diff
@@ -15,88 +16,158 @@ const defaultSerializedCache: JsonCache = {
...
+                if (newValueNotNull && newValueIsObject && newValueIsNotArray) {
+                    this.mergeUpdates(oldState[newKey], newValue);
+                } else {
+                    oldState[newKey] = newValue;
+                }
+            }
+        });
+
+        return oldState;
+    }
```
sameerag, line 145: Looks good. Can we add comments for readability?

lib/msal-node/src/cache/CacheManager.ts
```diff
@@ -1,8 +1,15 @@
+/*
```
sangonzal, line 1: I've split serialize/deserialize from writeToPersistence and readToPersistance, as that seemed much more transparent to me, but I'm happy to discuss/ change it back if want to do that.

DarylThayil:
LGTM!

PR #1752, opened 2020-06-05
Extend acquireTokenHelper instrumentation

jo-arroyo:
This PR adds instrumentation through logger messages to acquireTokenHelper for msal-core.


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -543,10 +549,12 @@ export class UserAgentApplication {
             if (interactionType === Constants.interactionTypeRedirect) {
                 if (!isLoginCall) {
                     this.cacheStorage.setItem(`${TemporaryCacheKeys.STATE_ACQ_TOKEN}${Constants.resourceDelimiter}${request.state}`, serverAuthenticationRequest.state, this.inCookie);
+                    this.logger.verbose("State cached for redirect");
                 }
```
jasonnutter, line 553: Can we add an `else` and a logger statement for when `isLoginCall` is true?

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -555,6 +563,7 @@ export class UserAgentApplication {
             }
```
jasonnutter, line 563: I think a log statement would be valuable inside the `else` here.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -499,9 +499,12 @@ export class UserAgentApplication {
      *
      */
     private async acquireTokenHelper(account: Account, interactionType: InteractionType, isLoginCall: boolean, request: AuthenticationParameters, resolve?: any, reject?: any): Promise<void> {
+        this.logger.verbose("AcquireTokenHelper has been called");
+
```
jasonnutter, line 503: I think a log message indicating the value of `isLoginCall` (which is not PII) would be helpful.

jasonnutter: Along with `interactionType`

pkanher617: +1

lib/msal-core/src/UserAgentApplication.ts
pkanher617: +1

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -543,10 +549,12 @@ export class UserAgentApplication {
             if (interactionType === Constants.interactionTypeRedirect) {
                 if (!isLoginCall) {
                     this.cacheStorage.setItem(`${TemporaryCacheKeys.STATE_ACQ_TOKEN}${Constants.resourceDelimiter}${request.state}`, serverAuthenticationRequest.state, this.inCookie);
+                    this.logger.verbose("State cached for redirect");
```
pkanher617, line 552: Did we choose to not log state as PII?

jo-arroyo: For some reason, I thought that if we weren't actually printing out the state that Pii wasn't necessary, but I can change it so that any reference to state is Pii.

jasonnutter: > For some reason, I thought that if we weren't actually printing out the state that Pii wasn't necessary

This is correct. I think @pkanher617 is specifically asking about logging the state value itself (in addition to the log message you added).

jo-arroyo: Ah, that makes sense. I chose to just log that message above, as it didn't seem like logging the state as Pii would be helpful in the logs. What do you think, @pkanher617 @jasonnutter?

lib/msal-core/src/UserAgentApplication.ts
jo-arroyo: Done

lib/msal-core/src/UserAgentApplication.ts
jo-arroyo: Added. Let me know if you think the log message needs to be changed

lib/msal-core/src/UserAgentApplication.ts
jo-arroyo: Done.

lib/msal-core/src/UserAgentApplication.ts
jo-arroyo: @pkanher617 @jasonnutter Added Pii logs for state being cached

PR #1756, opened 2020-06-05
Instance Discovery 2.x

tnorling:
Ports #1582 into msal-common
Similar to the previous PR for msal-core this PR does several things in regards to Authorities:

- Removes the hardcoded trusted host list for AAD hosts and replaces it with a network call to discover hosts known to MS. This is done the first time resolveEndpointsAsync is called.
- Combines B2C, AAD and ADFS authority classes into a single generic class
- Introduces `preferred_cache` and `preferred_network` values used to determine which endpoint to hit and what authority to cache under. This is the only piece that differs from the msal-core implementation.

Developers can opt out of the network call in one of 2 ways:
1. Provide all trusted authorities they intend to use in knownAuthorities config option (B2C and ADFS only)
2. Provide instance metadata via instanceMetadata config option (AAD only)

TODO:
- [ ] Add a few more unit tests
- [x] Doc Updates

Scheduled Refresh will be taken in future PR


.vscode/settings.json
```diff
@@ -1,12 +1,12 @@
...
     "editor.formatOnSave": false,
-    "eslint.autoFixOnSave": true,
     "eslint.validate": [
         "javascript",
         "javascriptreact",
-        { "language": "typescript", "autoFix": true }
+        "typescript"
     ],
     "eslint.quiet": true,
+    "eslint.workingDirectories": ["./lib/msal-core", "./lib/msal-common", "./lib/msal-browser", "./lib/msal-node", "./lib/msal-angular"],
```
pkanher617, line 9: Are these changes meant to be part of this PR?

lib/msal-common/src/authority/TrustedAuthority.ts
```diff
@@ -0,0 +1,84 @@
+import { Constants } from "../utils/Constants";
+import { INetworkModule } from "../network/INetworkModule";
+import { OpenIdConfigResponse } from "./OpenIdConfigResponse";
+import { InstanceMetadataType } from "./InstanceMetadataType";
+import { IInstanceDiscoveryMetadata } from "./IInstanceDiscoveryMetadata";
+import { ClientConfigurationError } from "../error/ClientConfigurationError";
+
+export class TrustedAuthority {
+    private static InstanceMetadata: InstanceMetadataType = {};
```
pkanher617, line 9: Is there any reason to make these static? I would rather only make functions and fields static if needed.

pkanher617: Things like knownAuthorities and the network interface can then be passed into the constructor and re-used

lib/msal-common/src/config/ClientConfiguration.ts
```diff
@@ -44,6 +45,9 @@ export type AuthOptions = {
     clientId: string;
     authority?: Authority;
     knownAuthorities?: Array<string>;
+    instanceMetadata?: Array<IInstanceDiscoveryMetadata>;
+    redirectUri?: string | (() => string);
```
pkanher617, line 49: `redirectUri` and `postlLogoutRedirectUri` should no longer be part of the client config

lib/msal-common/src/authority/TrustedAuthority.ts
tnorling: I received feedback when I implemented knownAuthorities originally that it should be static. The idea is that it's a set once variable to prevent trying to switch up your knownAuthorities mid-stream so we don't want it to be tied to an instance where we may need to rebuild it.

.vscode/settings.json
tnorling: I can take it out. I just got tired of putting it in every time I wanted linting to work properly.

tnorling:
> Looks good except for a couple things that we could change. Would also like to see a sample but it can be done later.

Sample showing `knownAuthorities` in PR #1775 
Will create a sample for `instanceMetadata` next week

lib/msal-common/src/config/ClientConfiguration.ts
tnorling: Done

.vscode/settings.json
tnorling: Done

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -14,6 +14,7 @@ export type BrowserAuthOptions = {
     clientId: string;
     authority?: string;
     knownAuthorities?: Array<string>;
+    instanceMetadata?: Array<IInstanceDiscoveryMetadata>;
```
tnorling, line 17: Naming here might get confusing when we implement the other perf changes which introduced `authorityMetadata` in msal-core. Open to suggestions for a better name here.

PR #1757, opened 2020-06-05
B2C Multiple Policy Support

tnorling:
Implements a workaround to support switching b2c policies in msal 1.x
Without additional cache changes this does not support caching multiple id_tokens outside of those bundled with the access_token. 

While not a perfect solution, this will unblock customers wanting to switch policies without logging out.
Can be enhanced in the future once B2C service sends back the user flow as a separate field in the `clientInfo`. Once this happens we can use that value instead of string parsing the authority url.


lib/msal-core/docs/FAQ.md
```diff
@@ -346,25 +347,25 @@ Please see the documentation on [Tenancy in Azure Active Directory](https://docs
...
-<head>
-   <!-- adding pollyfil for promises on IE11  -->
-      <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/js-polyfills/0.1.42/polyfill.min.js"> 
-   </script>
-</head>
-```
+## How can I implement password reset user flow in my B2C application with MSAL.js?
 
-For **arrow functions** and **template literals**, you need to transpile them to old JavaScript. You can use [this tool](https://babeljs.io/repl) to help with the process.
+Please review our sample [here](https://github.com/Azure-Samples/active-directory-b2c-javascript-msal-singlepageapp) to see how to implement the **password reset** user flow.
```
sameerag, line 368: nit: Please `checkout` our ...

tnorling:
```suggestion
Please checkout our sample [here](https://github.com/Azure-Samples/active-directory-b2c-javascript-msal-singlepageapp) to see how to implement the **password reset** user flow.
```

lib/msal-core/src/ClientInfo.ts
```diff
@@ -63,4 +63,28 @@ export class ClientInfo {
             throw ClientAuthError.createClientInfoDecodingError(e);
         }
     }
+
+    static stripPolicyFromUid(uid: string, authority: string): string {
+        const uidSegments = uid.split("-");
+        const urlSegments = authority.split("/").reverse();
+        let policy = ""
```
pkanher617, line 70: nitpick - semicolon

lib/msal-core/src/ClientInfo.ts
```diff
@@ -63,4 +63,28 @@ export class ClientInfo {
...
+
+        if (!StringUtils.isEmpty(urlSegments[0])) {
+            policy = urlSegments[0];
+        } else if (urlSegments.length > 1) {
+            policy = urlSegments[1];
+        }
```
pkanher617, line 76: Could you add some comments here describing what parsing is going on here?

lib/msal-core/src/ClientInfo.ts
```diff
@@ -63,4 +63,28 @@ export class ClientInfo {
...
+
+        if (uidSegments[uidSegments.length - 1] ===  policy) {
+            return uidSegments.slice(0, uidSegments.length - 1).join("-");
+        }
+
+        return uid;
+    }
+
+    public encodeClientInfo() {
```
pkanher617, line 85: Do we really need a helper for this?

lib/msal-core/src/ClientInfo.ts
pkanher617: I also wonder whether it would be better to have a getPolicy on the authority object and pass the policy string into this function.

change/msal-2020-08-12-14-47-04-b2c-multiple-policies.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "minor",
+  "comment": "B2C Multiple Policy Support (#1757)",
+  "packageName": "msal",
+  "email": "thomas.l.norling@gmail.com",
```
jo-arroyo, line 5: Wondering if this should be a microsoft email?

tnorling: Yes! Thanks for catching this!

lib/msal-core/src/ClientInfo.ts
tnorling: Parsing the authority string is not the long term solution and this part of the logic will be replaced by a separate "policy" field on the ClientInfo object once the work is done by the server. Given that this is temporary and we no longer have a B2CAuthority class I'd rather leave the changes confined to ClientInfo for now.

Will add comments however.

lib/msal-core/src/ClientInfo.ts
tnorling: I had to call this in a couple places, I thought it was helpful but open to alternative suggestions.

lib/msal-core/src/ClientInfo.ts
tnorling: Done

lib/msal-core/src/ClientInfo.ts
pkanher617: Nvm, Should be fine.

PR #1762, opened 2020-06-08
[msal-node][#2] Merge cache logic in `msal-common` for `node` and `browser` platforms 

sameerag:
This PR merges the cache operations for `node` and `browser` platforms. It:

* Rewrites storing cache with redefined API`s for browser with the new Cache format
* Updates SPAClient to consume the new APIs
* Remove old cache references for `browser` and `SPA Client`

cc @pkanher617


lib/msal-browser/src/cache/BrowserStorage.ts
```diff
@@ -138,13 +135,48 @@ export class BrowserStorage implements ICacheStorage {
      * Will retrieve frm cookies if storeAuthStateInCookie is set to true.
      * @param key
      */
-    getItem(key: string): string {
-        const msalKey = this.generateCacheKey(key);
-        const itemCookie = this.getItemCookie(msalKey);
-        if (this.cacheConfig.storeAuthStateInCookie && itemCookie) {
-            return itemCookie;
+    getItem(key: string, type: string): string | object {
```
jasonnutter, line 138: Per discussion with @pkanher617, instead of baking this business logic into these functions, you should instead create an abstract class with functions for each of the `case` statements, and then have `BrowserStorage` implement `ICacheStorage` and extend the abstract class.

jasonnutter: That way consumers can use a use-case specific function instead of passing a magic string, and that way if applications provide their own storage implementations, they don't have to know about this logic.

lib/msal-common/src/client/SilentFlowClient.ts
```diff
@@ -152,10 +153,13 @@ export class SilentFlowClient extends BaseClient {
...
     private isTokenExpired(expiresOn: string): boolean {
         // check for access token expiry
-        const expirationSec = Number(expiresOn);
+        let expirationSec = Number(expiresOn);
         const offsetCurrentTimeSec = TimeUtils.nowSeconds() + this.config.systemOptions.tokenRenewalOffsetSeconds;
 
         // Check if refresh is forced, or if tokens are expired. If neither are true, return a token response with the found token entry.
-        return (expirationSec && expirationSec > offsetCurrentTimeSec);
+        if (!expirationSec) {
+            expirationSec = 0;
```
jasonnutter, line 161: This could be done when you create `expirationSec`: `const expirationSec = Number(expiresOn) || 0;`

samples/VanillaJSTestApp2.0/app/default/auth.js
```diff
@@ -9,56 +9,52 @@ const isIE = msie > 0 || msie11 > 0;
...
-if (myMSALObj.getAccount()) {
-    // avoid duplicate code execution on page load in case of iframe and Popup window.
-    showWelcomeMessage(myMSALObj.getAccount());
-}
-
 async function signIn(method) {
     signInType = isIE ? "loginRedirect" : method;
     if (signInType === "loginPopup") {
-        const loginResponse = await myMSALObj.loginPopup(loginRequest).catch(function (error) {
+        myMSALObj.loginPopup(loginRequest).then(handleResponse).catch(function (error) {
```
jasonnutter, line 43: I would return this promise chain.

samples/VanillaJSTestApp2.0/app/default/auth.js
```diff
@@ -9,56 +9,52 @@ const isIE = msie > 0 || msie11 > 0;
...
-        }
     } else if (signInType === "loginRedirect") {
         myMSALObj.loginRedirect(loginRequest)
     }
 }
 
 function signOut() {
-    myMSALObj.logout();
+    const currentAcc = myMSALObj.getAccountByUsername(username);
+    myMSALObj.logout(currentAcc);
```
jasonnutter, line 53: TODO: Make `logout` accept a object (similar to other APIs)

lib/msal-browser/src/cache/BrowserStorage.ts
```diff
@@ -254,19 +284,37 @@ export class BrowserStorage implements ICacheStorage {
...
 
+    /**
+     * Prepend msal.<client-id> to each key; Skip for any JSON object as Key (defined schemas do not need the key appended: AccessToken Keys or the upcoming schema)
+     * @param key
+     * @param addInstanceId
+     */
+    generateCacheKey(key: string): string {
+        try {
+            // Defined schemas do not need the key migrated
+            this.validateObjectKey(key);
```
jasonnutter, line 308: Outside the scope of this PR, but could `validateObjectJey` just return a boolean, since you are catching and handling all errors?

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -555,13 +558,26 @@ export class PublicClientApplication {
...
+
+    /**
+     * Returns the signed in account matching username.
+     * (the account object is created at the time of successful login)
+     * or null when no state is found
+     * @returns {@link IAccount} - the account object stored in MSAL
+     */
+    public getAccountByUsername(userName: string): IAccount {
+        const allAccounts = this.getAllAccounts();
+        return allAccounts.filter((accountObj) => {
```
jasonnutter, line 578: Nit: this can be a one-liner: `return allAccounts.filter(accountObj => accountObj.username === userName)[0];`

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -522,9 +521,13 @@ export class PublicClientApplication {
      * Use to log out the current user, and redirect the user to the postLogoutRedirectUri.
      * Default behaviour is to redirect the user to `window.location.href`.
      */
-    logout(): void {
+    logout(account: IAccount, authorityString?: string): void {
```
jasonnutter, line 524: This can instead request an object, similar to the other APIs? e.g. `LogoutRequest`

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -28,6 +28,7 @@ export type BrowserAuthOptions = {
 export type CacheOptions = {
     cacheLocation?: string;
     storeAuthStateInCookie?: boolean;
+    storeInMemory?: boolean;
```
sangonzal, line 31: Should this be exposed to users?

pkanher617: We remove this in a later PR.

lib/msal-common/src/client/SilentFlowClient.ts
pkanher617: done.

samples/VanillaJSTestApp2.0/app/default/auth.js
pkanher617: done.

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: done.

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: Will tackle this in a future PR

lib/msal-browser/src/cache/BrowserStorage.ts
pkanher617: Done in a future PR.

samples/VanillaJSTestApp2.0/app/default/auth.js
pkanher617: Will be done in a future PR.

lib/msal-browser/src/cache/BrowserStorage.ts
pkanher617: We will revisit this when we add the feature for consumer driven storage.

jasonnutter: We already have consumer driven storage (the extensions library).

jasonnutter: @sameerag @pkanher617 Let's talk about this next design meeting.

jasonnutter: We also have other storage scenarios for the web such as VS code extensions that we will want to support, which would fall under "user-provided storage"

sameerag: The abstract class is not implemented in the follow up PR #1771. Regarding abstracting `getItem` implementation, I agree we can improve it for `custom storage`, however I would like to track it separately and as an enhancement.

sameerag: We can prioritize if we can arrive at a design soon but considering the timelines to get a working branch earlier, I would like to close this PR.

PR #1770, opened 2020-06-11
[msal-common family][#3] Utilize ScopeSet across the library

pkanher617:
This PR does the following:
- Removes scopeRequired from ScopeSet
- Always adds openid and profile to all requests
- Using containsScopeSet and intersectingScopeSets in the UnifiedCacheManager
- Add browser maintenance code
- Adds `fromCache` to the AuthenticationResult
- Adds nonce check to the ResponseHandler.ts


lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -137,20 +131,59 @@ export abstract class ClientApplication {
         };
     }
 
+    /**
+     * Generates a request with the default scopes.
+     * @param request 
+     */
+    protected generateUrlRequest(request: AuthorizationUrlRequest): AuthorizationUrlRequest {
```
sangonzal, line 138: We should not duplicate this code here (ln 134-175) and device code request in PublicClientApplication. 

Consider creating ScopeSet and passing in as part of ClientConfiguration or creating a parent class for requests.

lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -137,20 +132,31 @@ export abstract class ClientApplication {
         };
     }
 
+    /**
+     * Generates a request with the default scopes.
+     * @param authRequest 
+     */
+    protected initializeRequestScopes(authRequest: BaseAuthRequest): BaseAuthRequest {
+        const request: BaseAuthRequest = { ...authRequest };
+        if (!request.scopes) {
```
sangonzal, line 141: Should you also be checking for array length?

pkanher617: resolved offline, don't need to check for length.

PR #1771, opened 2020-06-11
[msal-common][msal-node][#4] Update UnifiedCacheManager.ts

pkanher617:
This PR does the following:
- Update UnifiedCacheManager.ts to be a class of static functions, rather than a constructed class
- Make getAllAccounts a static function
- Move all code dealing with JSON to in-memory serialization to `msal-node`


lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -192,4 +192,8 @@ export abstract class ClientApplication {
             this.storage.getCache() as InMemoryCache
         );
     }
+
+    getAllAccounts(): IAccount[] {
+        return AuthorizationCodeClient.getAllAccounts(this.storage);
```
sangonzal, line 197: This is confusing. What if the user is not doing authorization code flow?

pkanher617: Doesn't matter, you can call it off any client that extends the base client.

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -3,109 +3,137 @@
...
 
     /**
-     * Returns all accounts in memory
+     * Returns all accounts in cache
      */
-    getAllAccounts(): AccountCache {
-        return this.getAccountsFilteredBy();
+    getAllAccounts(): IAccount[] {
+        const currentAccounts: AccountCache = this.getAccountsFilteredBy();
+        const accountValues: AccountEntity[] = Object.values(currentAccounts);
```
tnorling, line 67: Object.values won't work in IE11 afaik. I think there's already a lot of incompatibility in 2.0 so this won't be the thing that breaks it, but its more to clean up later if/when we do work on compatibility.

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -367,12 +366,35 @@ export class UnifiedCacheManager implements ICacheManager {
...
-        return this.cacheStorage.removeItem(
-            key,
-            CacheSchemaType.CREDENTIAL,
-            this.inMemory
-        );
+        return this.removeItem(key, CacheSchemaType.CREDENTIAL);
+    }
+}
+
+export class DefaultStorageClass extends CacheManager {
```
sangonzal, line 375: All default implementation are in ClientConfiguration. Seems like this should also go int there?

lib/msal-common/src/client/BaseClient.ts
```diff
@@ -33,17 +28,11 @@ export abstract class BaseClient {
     protected cryptoUtils: ICrypto;
 
     // Storage Interface
-    protected cacheStorage: ICacheStorage;
+    protected cacheStorage: CacheManager;
```
sangonzal, line 31: We should either rename CacheManager to CacheStorage, or rename the variables to match the class name.

lib/msal-common/src/cache/CacheManager.ts
sameerag: good call. @pkanher617 can you please note this to track when we do a recon on msal-common?

lib/msal-common/src/cache/CacheManager.ts
pkanher617: I put this here because if the interface ever changes, you can make all the changes in a single file instead of having to jump between here and ClientConfiguration.ts

lib/msal-common/src/client/BaseClient.ts
pkanher617: I can rename the variable to cacheManager.

PR #1777, opened 2020-06-12
Extend response and token processing instrumentation

jo-arroyo:
This PR adds instrumentation through logger messages to the following functions in msal-core:
- `openPopup`

Response
- `isCallback`
- `processCallback`
- `handleAuthenticationResponse`
- `handleRedirectAuthenticationResponse`
- `getResponseState`

Token processing
- `getCachedToken`
- `getUniqueAuthority`
- `extractADALIdToken`
- `renewToken`
- `renewIdToken`
- `saveAccessToken`
- `saveTokenFromHash`
- `populateAuthority`


lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1297,23 +1323,27 @@ export class UserAgentApplication {
...
         }
 
         // external api requests may have many renewtoken requests for different resource
         if (!stateResponse.stateMatch) {
+            this.logger.verbose("State does not match cached state, setting requestType to type from window");
             stateResponse.requestType = window.requestType;
             const statesInParentContext = window.renewStates;
             for (let i = 0; i < statesInParentContext.length; i++) {
                 if (statesInParentContext[i] === stateResponse.state) {
+                    this.logger.verbose("StateMatch set to true");
```
tnorling, line 1346: Since this is in a for loop, this may give many of the same message. Consider not having this log statement or finding a way to disambiguate between different messages.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1334,21 +1364,25 @@ export class UserAgentApplication {
      * @param {Account} account - Account for which the scopes were requested
      */
     private getCachedToken(serverAuthenticationRequest: ServerRequestParameters, account: Account): AuthResponse {
+        this.logger.verbose("GetCachedToken has been called");
         let accessTokenCacheItem: AccessTokenCacheItem = null;
         const scopes = serverAuthenticationRequest.scopes;
 
         // filter by clientId and account
         const tokenCacheItems = this.cacheStorage.getAllAccessTokens(this.clientId, account ? account.homeAccountIdentifier : null);
+        this.logger.verbose("Getting all cached tokens");
```
tnorling, line 1373:
```suggestion
        this.logger.verbose("Getting all cached access tokens");
```

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1402,10 +1442,12 @@ export class UserAgentApplication {
         }
 
         if (accessTokenCacheItem != null) {
+            this.logger.verbose("Evaluating access token found");
             const expired = Number(accessTokenCacheItem.value.expiresIn);
             // If expiration is within offset, it will force renew
             const offset = this.config.system.tokenRenewalOffsetSeconds || 300;
             if (expired && (expired > TimeUtils.now() + offset)) {
+                this.logger.verbose("Token is not expired, renewing token");
```
tnorling, line 1450: This may be a little confusing. I wouldnt expect a token to renew if it's not expired. Consider specifying why it's renewing, in this case because, expiration is within the offset

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1771,13 +1847,15 @@ export class UserAgentApplication {
                         }
                         // Save the token
                         else {
+                            this.logger.verbose("Nonce matches, saving idToken to cache, saving idToken as access token for app");
```
tnorling, line 1850: The access token thing we do here may be a little confusing to customers, I suggest we don't call this out specifically. It's enough to say we're caching the id token
```suggestion
                            this.logger.verbose("Nonce matches, saving idToken to cache");
```

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1771,13 +1847,15 @@ export class UserAgentApplication {
...
                             this.cacheStorage.setItem(PersistentCacheKeys.IDTOKEN, hashParams[ServerHashParamKeys.ID_TOKEN]);
                             this.cacheStorage.setItem(PersistentCacheKeys.CLIENT_INFO, clientInfo);
 
                             // Save idToken as access token for app itself
                             this.saveAccessToken(response, authority, hashParams, clientInfo, idTokenObj);
                         }
                     } else {
+                        this.logger.verbose("No idToken or no nonce. Authority set as state");
```
tnorling, line 1858:
```suggestion
                        this.logger.verbose("No idToken or no nonce. Cache key for Authority set as state");
```

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1689,44 +1752,53 @@ export class UserAgentApplication {
...
                         response.idTokenClaims = idTokenObj.claims;
                     } else {
+                        this.logger.verbose("No idToken on fragment, getting idToken from cache");
                         idTokenObj = new IdToken(this.cacheStorage.getItem(PersistentCacheKeys.IDTOKEN));
                         response = ResponseUtils.setResponseIdToken(response, idTokenObj);
                     }
 
                     // set authority
                     const authority: string = this.populateAuthority(stateInfo.state, this.inCookie, this.cacheStorage, idTokenObj);
+                    this.logger.verbose("Authority populated");
```
tnorling, line 1767:
```suggestion
                    this.logger.verbose("Got authority from cache");
```

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -839,12 +839,12 @@ export class UserAgentApplication {
                              * App uses idToken to send to api endpoints
                              * Default scope is tracked as clientId to store this token
                              */
-                            this.logger.verbose("Renewing idToken");
+                            this.logger.verbose("ClientId is the only scope, renewing idToken");
```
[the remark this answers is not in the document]
jo-arroyo: Yes! I had a chat with Hector about this. The logic will have to change here once he's done with all his scope changes, but he suggested that I write the logging so it reflects how it currently works, and they will change the logging once they make the relevant scope changes.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -911,7 +912,7 @@ export class UserAgentApplication {
 
             return popupWindow;
         } catch (e) {
-            this.logger.error("error opening popup " + e.message);
+            this.logger.error("Error opening popup " + e.message);
```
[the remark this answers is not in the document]
jo-arroyo: Perhaps. This was already here. I will double check to see if there's a specific reason why it needs to be logged here as well as throwing an error.

lib/msal-core/src/UserAgentApplication.ts
jasonnutter: This looks to be the success case, and we break out of the loop when this happens. I would maybe say something like "matching state found for request", and then have a log message that is printed at the end of the loop that says "matching state not found for request".

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1154,8 +1156,11 @@ export class UserAgentApplication {
      */
     private processCallBack(hash: string, stateInfo: ResponseStateInfo, parentCallback?: Function): void {
         this.logger.info("Processing the callback from redirect response");
+        this.logger.verbose("ProcessCallBack has been called");
```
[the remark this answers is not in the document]
jo-arroyo: Done.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1177,24 +1182,29 @@ export class UserAgentApplication {
...
                     }
+                    this.logger.verbose(`Response tokenType set to ${ServerHashParamKeys.ACCESS_TOKEN}`);
                     response.tokenType = ServerHashParamKeys.ACCESS_TOKEN;
                 }
                 else if (stateInfo.requestType === Constants.login) {
+                    this.logger.verbose(`Response tokenType set to ${ServerHashParamKeys.ID_TOKEN}`);
                     response.tokenType = ServerHashParamKeys.ID_TOKEN;
                 }
                 if (!parentCallback) {
+                    this.logger.verbose("No callbacks provided, setting redirectResponse to response");
```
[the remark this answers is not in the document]
jo-arroyo: Thanks! Changed.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1177,24 +1182,29 @@ export class UserAgentApplication {
...
                     this.redirectResponse = response;
                     return;
                 }
             } else if (!parentCallback) {
+                this.logger.verbose("Response is null and no callbacks provided, building redirectResponse");
```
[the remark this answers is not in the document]
jo-arroyo: Done.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1230,15 +1243,19 @@ export class UserAgentApplication {
...
         this.logger.info("Returned from redirect url");
+        this.logger.verbose("HandleRedirectAuthenticationResponse has been called");
 
         // clear hash from window
         window.location.hash = "";
+        this.logger.verbose("Window.location.hash cleared");
 
         // if (window.parent !== window), by using self, window.parent becomes equal to window in getResponseState method specifically
         const stateInfo = this.getResponseState(hash);
+        this.logger.verbose("Ensured window.parent is equal to window");
```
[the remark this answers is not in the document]
jo-arroyo: Done.

lib/msal-core/src/UserAgentApplication.ts
jo-arroyo: Changed. Let me know if there's another way you'd prefer this to look.

lib/msal-core/src/UserAgentApplication.ts
jo-arroyo: Sure, makes sense. Changed.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -787,7 +787,7 @@ export class UserAgentApplication {
 
             // resolve/reject based on cacheResult
             if (cacheResultResponse) {
-                this.logger.verbose("Token is already in cache for scope: " + scope);
+                this.logger.verbose("Token found in cache lookup");
```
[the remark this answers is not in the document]
jo-arroyo: Done.

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1247,18 +1262,24 @@ export class UserAgentApplication {
                 window.location.assign("/");
                 return;
             } else {
+                this.logger.verbose("Valid login request url obtained from cache");
                 const currentUrl = UrlUtils.removeHashFromUrl(window.location.href);
                 const finalRedirectUrl = UrlUtils.removeHashFromUrl(loginRequestUrl);
                 if (currentUrl !== finalRedirectUrl) {
+                    this.logger.verbose("Current url is not login request url, navigating");
```
tnorling, line 1269: Can we add a PII log stating what both currentUrl and finalRedirectUrl are? This would help with debugging redirect issues.

jo-arroyo: Sure thing! Added.

PR #1787, opened 2020-06-16
Dev react native poc merge from forked repo

ahnmelis:


ahnmelis:
Merging my branch from my forked repo to the actual repo. 
In order to rename the module, I created a new library and copied my existing methods into it. I also created a new sample app with the existing code, which is now housed in a folder adjacent to MSALModule. Because of issues with npm packages cluttering up the sample app in the past, I will be testing my native module from now on by directly putting the java file in the sample app android folder and adjusting import lines accordingly. Switching the logic to accommodate npm will be done when the library is complete.

lib/msal-react-native-android-poc/SampleApp/index.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * @format
+ */
+
+import {AppRegistry} from 'react-native';
```
technical-boy, line 5: This is just a nitpick, but for the sake of consistency I'd suggest adding spaces after opening and before closing curly braces.

lib/msal-react-native-android-poc/SampleApp/ios/SampleApp-tvOS/Info.plist
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
technical-boy, line 1: It's been a while since I did react-native or mobile development in general, but I think .plist files may be better off gitignored. I don't think it's very important, but I suggest you research whether or not these files should be included in the git repo if you have time.

lib/msal-react-native-android-poc/SampleApp/ios/SampleApp.xcodeproj/xcshareddata/xcschemes/SampleApp-tvOS.xcscheme
```diff
@@ -0,0 +1,88 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
technical-boy, line 1: Same as .plist, may be a good time investment to look into what iOS/Android build files should be .gitignored.

lib/msal-react-native-android-poc/MSALModule/android/src/main/java/com/reactlibrary/MSALModule.java
```diff
@@ -0,0 +1,111 @@
...
+    private ISingleAccountPublicClientApplication IPublicClientApplication;
+    private IAccount mAccount;
+
+    public MSALModule(ReactApplicationContext reactContext) {
+        super(reactContext);
+        this.reactContext = reactContext;
+
+        try{
+            //initialize IPublicClientApplication with config file, located in main/java/res/raw/auth_config_single_account.json
+            IPublicClientApplication = PublicClientApplication.createSingleAccountPublicClientApplication(
```
technical-boy, line 27: I'm a bit out of my depth here, so I may be wrong about this. Given that the "I" prefix in variable names is usually intended to identify interfaces, and this line is assigning an instance of `PublicClientApplication` to `IPublicClientApplication`, maybe renaming the variable to something like `publicClientApplication` would be a good idea.

Also, the variable declaration in line 18 defines the type as `ISingleAccountPublicClientApplication`, which is an interface (to my knowledge, Java interfaces cannot and should not be instantiated). I think you might want to use the `SingleAccountPublicClientApplication` class, which implements `ISingleAccountPublicClientApplication` instead.

Like I said, a little out of my depth here so let me know if I'm off base!

ahnmelis: This makes sense and I will go ahead and rename the variable. 
For `ISingleAccountPublicClientApplication` in general, for some reason, `createSingleAccountPublicClientApplication` isn't a method of `singleAccountPublicClientApplication` (I tried this first a while ago) but of the interface instead. I modeled a lot of MSALmodule off of the MSAL Android sample app, which also instantiates `ISingleAccountPublicClientApplication` in a similar way. This also confuses me and I should have look into a better  explanation earlier and will do so now!

lib/msal-react-native-android-poc/SampleApp/ios/SampleApp-tvOS/Info.plist
ahnmelis: did a check for both plist and xcsheme in a suggested .gitignore list (https://www.toptal.com/developers/gitignore/api/reactnative) and although both aren't there, this was good for me to look up

lib/msal-react-native-android-poc/SampleApp/android/app/src/main/java/com/sampleapp/MainApplication.java
```diff
@@ -0,0 +1,82 @@
...
+import com.facebook.react.ReactApplication;
+import com.facebook.react.ReactInstanceManager;
+import com.facebook.react.ReactNativeHost;
+import com.facebook.react.ReactPackage;
+import com.facebook.soloader.SoLoader;
+import java.lang.reflect.InvocationTargetException;
+import java.util.List;
+
+
+public class MainApplication extends Application implements ReactApplication {
```
jmckennon, line 15: For my own understanding/learning... this is saying `MainApplication` inherits (or is a child of) `Application` and also implements the interface `ReactApplication`?

lib/msal-react-native-android-poc/MSALModule/android/src/main/java/com/reactlibrary/MSALModule.java
technical-boy: To be honest, that is a very good reason for having the code the way you wrote it. If it works and it matches the Android sample you must be on the right track. Maybe someone from the Android team can shed a little light on why the code is organized/works that way.

lib/msal-react-native-android-poc/SampleApp/android/app/src/main/java/com/sampleapp/MainApplication.java
ahnmelis: Yes I believe that's correct. From what I learned in class, I think in java, a class can only extend one class while it can implement many interfaces

PR #1788, opened 2020-06-16
[msal-node][msal-extensions] Add msal extensions 

sangonzal:


jasonnutter:
Can this be broken up into smaller PRs? 16k lines 👀

sameerag:
@jasonnutter  part of it could be because @sangonzal based this on dev and some PRs are yet to be merged. Either way, if breaking it up logically is feasible, I would prefer that too.

PR #1789, opened 2020-06-17
Changing IAccount Interface to AccountInfo

pkanher617:
This PR changes the account object returned to the user from the IAccount interface to the AccountInfo type.

It also refactors the code to remove the CacheHelpers.ts file, and removes the `matches` logic from the CacheManager.ts file.


lib/msal-node/src/cache/serializer/EntitySerializer.ts
```diff
@@ -12,10 +13,11 @@ export class EntitySerializer {
       * @param key
       */
     static mapAccountKeys(accCache: AccountCache, key: string): object {
```
sangonzal, line 15: Some extra stuff in this file that I think is no longer needed (commented out)

lib/msal-common/src/cache/entities/AccountEntity.ts
```diff
@@ -72,6 +72,32 @@ export class AccountEntity {
...
+
+    /**
+     * Generates account key from interface
+     * @param accountInterface
+     */
+    static generateAccountCacheKey(accountInterface: AccountInfo): string {
+        const accountKey = [
+            accountInterface.homeAccountId,
+            accountInterface.environment || "",
+            accountInterface.tenantId || "",
```
tnorling, line 95: Do we need to include username here as well?

lib/msal-node/src/cache/serializer/EntitySerializer.ts
tnorling: +1 for removing unneeded comments

lib/msal-common/src/cache/entities/AccountEntity.ts
pkanher617: username isn't part of the key, it is a part of the value.

tnorling: That's confusing to me because it's passed to this function on line 51 and also returned in `getAccountInfo`

lib/msal-node/src/cache/serializer/EntitySerializer.ts
pkanher617: This is being fixed by @sameerag.

lib/msal-common/src/cache/entities/AccountEntity.ts
tnorling: Resolved offline. Agreed to add comments around Key,Value for Account

lib/msal-browser/src/cache/BrowserStorage.ts
```diff
@@ -144,22 +144,22 @@ export class BrowserStorage extends CacheManager {
         switch (type) {
             case CacheSchemaType.ACCOUNT: {
                 const account = new AccountEntity();
-                return (CacheHelper.toObject(account, JSON.parse(value)) as AccountEntity);
+                return (CacheManager.toObject(account, JSON.parse(value)) as AccountEntity);
```
tnorling, line 147: I am getting an error on this `JSON.parse` call when there's an entry in the cache that is not in JSON format, e.g. a non-msal cache value. We probably need additional logic somewhere to ignore cache entries we don't expect

pkanher617: I've added a fix for this code as discussed. The account filter code will now pass over any non-MSAL cache items.

PR #1790, opened 2020-06-17
[msal-browser][msal-common] Make library state an object and add timestamp (Updated)

pkanher617:
This PR is an update of #1675, made because the cache update drastically changed the surface of the library. The changes made in this PR will be merged, and #1675 will be closed.

This PR converts library state from a GUID to a stringified JSON object with id and timestamp fields. This will allow msal-common and msal-browser to handle situations where responses are handled long after the request was initiated, and setting correct token expiration lifetimes.

Also includes tests.


lib/msal-common/src/error/ClientAuthError.ts
```diff
@@ -42,6 +42,10 @@ export const ClientAuthErrorMessage = {
         code: "blank_guid_generated",
         desc: "The guid generated was blank. Please review the trace to determine the root cause."
     },
+    invalidStateError: {
+        code: "invalid_state",
+        desc: "State was not valid, please check the network trace."
```
jasonnutter, line 47: Is "network trace" the right place to be looking?

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -188,8 +185,15 @@ export class ResponseHandler {
...
         const responseScopes = ScopeSet.fromString(serverTokenResponse.scope);
         // Expiration calculation
-        const expiresInSeconds = TimeUtils.nowSeconds() + serverTokenResponse.expires_in;
-        const extendedExpiresInSeconds = expiresInSeconds + serverTokenResponse.ext_expires_in;
+        const currentTime = TimeUtils.nowSeconds();
+        let tokenExpirationSeconds = serverTokenResponse.expires_in;
+        if (!libraryState) {
+            tokenExpirationSeconds += currentTime;
+        } else {
+            console.log(libraryState.ts);
```
jasonnutter, line 193: `console.log`

jasonnutter: This could be a one-line ternary as well:

```js
const timeStamp = libraryState ? libraryState.ts : currentTime;
const tokenExpirationSeconds = severTokenResponse.expires_in + timeStamp;
```

lib/msal-common/src/utils/ProtocolUtils.ts
```diff
@@ -15,23 +28,44 @@ export class ProtocolUtils {
...
+
+    static parseRequestState(state: string, cryptoObj: ICrypto): RequestStateObject {
+        if (StringUtils.isEmpty(state)) {
+            throw ClientAuthError.createInvalidStateError(state, "Null, undefined or empty state");
+        }
+
+        try {
+            const splitState = decodeURIComponent(state).split(Constants.RESOURCE_DELIM);
+            const libraryState = splitState[0];
+            const userState = splitState.length > 1 ? splitState.slice(1).join(Constants.RESOURCE_DELIM) : "";
```
jasonnutter, line 60: Can you add a comment here to explain why this code is needed?

lib/msal-common/src/error/ClientAuthError.ts
pkanher617: Changed.

lib/msal-common/src/utils/ProtocolUtils.ts
pkanher617: added.

lib/msal-common/src/response/ResponseHandler.ts

jasonnutter:
Can you add a test like this? https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-core/test/utils/RequestUtils.spec.ts#L130-L147

lib/msal-browser/test/app/PublicClientApplication.spec.ts
```diff
@@ -932,7 +959,7 @@ describe("PublicClientApplication.ts Class Unit Tests", () => {
                 "tid": "3338040d-6c67-4c5b-b112-36a304b66dad",
                 "nonce": "123523",
             };
-            const testAccount: IAccount = {
+            const testAccount: AccountInfo = {
```
tnorling, line 962: It looks like we recreate this testAccount in a lot of places. Can we just declare this at the top of the file or in a utils file once?

pkanher617: Will optimize tests in a later PR.

PR #1793, opened 2020-06-17
msal-browser uses AuthCodeClient and SilentFlowClient

pkanher617:
This PR removes the SPAClient.ts file and changes the PublicClientApplication class in msal-browser to
- Create a client on every API call
- Generate authority dynamically
- Instead of using SPAClient, we now use the AuthorizationCodeClient and SilentFlowClient classes

This PR also does the following:
- a request object for logout APIs (Currently only used by the browser)
- some cleanup code
- remove redirectUri from the ClientConfiguration and add it to all request objects
- refactor AuthorizationCodeModule to fit for SPA use cases


lib/msal-common/src/client/BaseClient.ts
```diff
@@ -54,7 +59,24 @@ export abstract class BaseClient {
...
-        this.defaultAuthority = this.config.authOptions.authority;
+        this.authority = this.config.authOptions.authority;
+    }
+
+    /**
+     * Use to log out the current user, and redirect the user to the postLogoutRedirectUri.
+     * Default behaviour is to redirect the user to `window.location.href`.
+     * @param authorityUri
+     */
+    async getLogoutUri(logoutRequest: EndSessionRequest): Promise<string> {
```
sangonzal, line 70: This is only applicable to SPA, so I don't it should live in BaseClient.

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -58,20 +65,36 @@ export class AuthorizationCodeClient extends BaseClient {
...
 
         return tokenResponse;
     }
 
+    /**
+     * Handles the hash fragment response from public client code request. Returns a code response used by
+     * the client to exchange for a token in acquireToken.
+     * @param hashFragment
+     */
+    public handleFragmentResponse(hashFragment: string, cachedState: string): string {
```
sangonzal, line 80: This is only applicable to SPAs, so it might make sense to have a SPAClient that wraps AuthCode Client, and adds handleFragmentResponse and EndSession API.

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -58,20 +65,36 @@ export class AuthorizationCodeClient extends BaseClient {
             this.logger
         );
 
+        // Validate response. This function throws a server error if an error is returned by the server.
         responseHandler.validateTokenResponse(response.body);
-        const tokenResponse = responseHandler.generateAuthenticationResult(response.body, this.defaultAuthority);
+        const tokenResponse = responseHandler.generateAuthenticationResult(response.body, this.authority, cachedNonce, cachedState);
```
sangonzal, line 70: Can be done in a separate PR but at some point we should add this to BaseClient - These are used and duplicate across all client at the moment.

lib/msal-common/src/client/AuthorizationCodeClient.ts
pkanher617: This is a convenience function that can be extended to check for query params as well, so I think it is ok to keep it here.

lib/msal-common/src/client/AuthorizationCodeClient.ts
pkanher617: the nonce and state checks are specific to this flow though.

lib/msal-common/src/client/BaseClient.ts
pkanher617: This could be a convenience function for msal-node in the future, I will keep it here for now.

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -54,11 +54,17 @@ export class RefreshTokenClient extends BaseClient {
     private createTokenRequestBody(request: RefreshTokenRequest): string {
         const parameterBuilder = new RequestParameterBuilder();
 
+        parameterBuilder.addClientId(this.config.authOptions.clientId);
+
+        parameterBuilder.addRedirectUri(request.redirectUri);
```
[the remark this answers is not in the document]
pkanher617: Yes to ensure it is a valid CORS redirect. I don't think it affects server side flows.

lib/msal-common/src/client/BaseClient.ts
sangonzal: Right, although it could be a convenience function, it's still not applicable to 7/8 flows, so I don't why it would be here.

sangonzal: If it's ever used for Node, it would also just be used for authorization code flow

lib/msal-common/src/client/RefreshTokenClient.ts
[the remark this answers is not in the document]
sangonzal: Ok makes sense

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -486,15 +466,13 @@ export class PublicClientApplication {
...
      */
-    logout(account: AccountInfo, authorityString?: string): void {
-        const authorityObj = StringUtils.isEmpty(authorityString) ? this.defaultAuthorityInstance : AuthorityFactory.createInstance(
-            this.config.auth.authority,
-            this.config.system.networkClient
-        );
-
-        // create logout string and navigate user window to logout. Auth module will clear cache.
-        this.authModule.logout(account, authorityObj).then((logoutUri: string) => {
+    logout(logoutRequest: EndSessionRequest): void {
```
jasonnutter, line 471: Do we just want to make `logout` async as well? That way you can use async/await.

jasonnutter: Also, this seems like something that could have been done incrementally after this PR, in teh interest of a smaller PR.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -511,7 +489,15 @@ export class PublicClientApplication {
      *
      */
     public getRedirectUri(): string {
-        return this.authModule.getRedirectUri();
+        if (this.config.auth.redirectUri) {
+            if (typeof this.config.auth.redirectUri === "function") {
+                return this.config.auth.redirectUri();
+            } else if (!StringUtils.isEmpty(this.config.auth.redirectUri)) {
```
jasonnutter, line 495: I would like to better understand the `redirectUri` as a function use case, as it would be nice to remove if its not truly needed. Needing code like this is annoying, frankly.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -521,7 +507,17 @@ export class PublicClientApplication {
      * @returns {string} post logout redirect URL
      */
     public getPostLogoutRedirectUri(): string {
-        return this.authModule.getPostLogoutRedirectUri();
+        if (this.config.auth.postLogoutRedirectUri) {
+            if (typeof this.config.auth.postLogoutRedirectUri === "function") {
+                return this.config.auth.postLogoutRedirectUri();
+            } else if (
+                !StringUtils.isEmpty(this.config.auth.postLogoutRedirectUri)
```
jasonnutter, line 514: Given the truthy check about for `postLogoutRedirectUri`, this statement will never be false, and can be an `else` instead.

lib/msal-browser/src/app/PublicClientApplication.ts
jasonnutter: Also, this statement will never be false, given the truthy check above.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -219,7 +184,7 @@ export class PublicClientApplication {
 
             if (StringUtils.isEmpty(loginRequestUrl) || loginRequestUrl === "null") {
                 // Redirect to home page if login request url is null (real null or the string null)
-                this.authModule.logger.warning("Unable to get valid login request url from cache, redirecting to home page");
+                console.warn("Unable to get valid login request url from cache, redirecting to home page");
```
jasonnutter, line 187: Why not use `logger`?

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -614,17 +651,32 @@ export class PublicClientApplication {
...
+     * @param request 
+     */
+    private generateLoginRequest(request: AuthorizationUrlRequest): AuthorizationUrlRequest {
+        const loginRequest = { ...request };
+        if (!loginRequest.scopes) {
+            loginRequest.scopes = [Constants.OPENID_SCOPE, Constants.PROFILE_SCOPE];
+        } else {
+            loginRequest.scopes.push(Constants.OPENID_SCOPE, Constants.PROFILE_SCOPE);
+        }
+        return loginRequest;
```
jasonnutter, line 677:
```js
return {
   ...loginRequest,
  scopes: [ ...loginRequest.scopes, Constants.OPENID_SCOPE, Constants.PROFILE_SCOPE ]
}
```

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -640,5 +692,17 @@ export class PublicClientApplication {
...
+    private generateLogoutRequest(request: EndSessionRequest): EndSessionRequest {
+        const logoutRequest = {
+            ...request
+        };
+
+        if (StringUtils.isEmpty(logoutRequest.postLogoutRedirectUri)) {
+            logoutRequest.postLogoutRedirectUri = this.getPostLogoutRedirectUri();
+        }
+
+        return logoutRequest;
```
jasonnutter, line 704:
```js
return {
  ...request,
  postLogoutRedirectUri: !StringUtils.isEmpty(request.postLogoutRedirectUri) ? request.postLogoutRedirectUri : this.getPostLogoutRedirectUri()
}
```

lib/msal-browser/src/cache/BrowserStorage.ts
```diff
@@ -343,6 +343,17 @@ export class BrowserStorage extends CacheManager {
         this.setItem(this.generateCacheKey(authorityKey), authority, CacheSchemaType.TEMPORARY);
     }
 
+    /**
+     * Gets the cached authority based on the cached state. Returns empty if no cached state found.
+     */
+    getCachedAuthority(): string {
+        const state = this.getItem(this.generateCacheKey(TemporaryCacheKeys.REQUEST_STATE), CacheSchemaType.TEMPORARY) as string;
+        if (!state) {
+            return "";
```
jasonnutter, line 352: I think `null` is a better semantic representation of not being found that empty string.

lib/msal-common/src/request/RefreshTokenRequest.ts
```diff
@@ -10,8 +10,9 @@
  * authority:               URL of the authority, the security token service (STS) from which MSAL will acquire tokens.
  * refreshToken:            A refresh token returned from a previous request to the Identity provider.
  */
-export class RefreshTokenRequest {
+export type RefreshTokenRequest = {
     scopes: Array<string>;
     refreshToken: string;
+    redirectUri?: string;
```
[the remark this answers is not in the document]
jasonnutter: And should this be `string | function`?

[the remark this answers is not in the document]
jasonnutter: Another reason why I'd like to see if we can remove the function from the redirectUri type.

lib/msal-common/src/authority/AuthorityFactory.ts
```diff
@@ -17,44 +17,71 @@ import { AdfsAuthority } from "./AdfsAuthority";
...
+     * @param networkInterface 
+     */
+    private static detectAuthorityFromUrl(authorityString: string, networkInterface: INetworkModule): Authority {
+        const authorityUrl: UrlString = new UrlString(authorityString);
+        const components = authorityUrl.getUrlComponents();
+        const pathSegments = components.PathSegments;
+
+        if (pathSegments.length && pathSegments[0].toLowerCase() === Constants.ADFS)
+            return new AdfsAuthority(authorityString, networkInterface);
+        else if (B2cAuthority.B2CTrustedHostList.length)
```
jasonnutter, line 81: Can you add brackets please (this will be enforced by lint soon)?

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -558,17 +554,54 @@ export class PublicClientApplication {
...
+
+    /**
+     * Creates a Client Configuration object with the given authority, or the default authority.
+     * @param authorityUri 
+     */
+    private async getClientConfiguration(authorityUri?: string): Promise<ClientConfiguration> {
+        return {
+            authOptions: {
+                clientId: this.config.auth.clientId,
+                authority: authorityUri ? await AuthorityFactory.createDiscoveredInstance(authorityUri, this.config.system.networkClient) : await this.defaultAuthorityPromise,
```
jasonnutter, line 584: Can you pull this into a local variable before the return statement? Would make it easier to read.

lib/msal-common/src/client/BaseClient.ts
pkanher617: OK I will move it to Auth code client

lib/msal-browser/src/app/PublicClientApplication.ts
jasonnutter: I believe we can drop the redirectUri as a function, and just adjust the way to select the current location as the default.

```js
public getRedirectUri(): string {
  return this.config.auth.redirectUri || window.location.href;
}
```

And then make sure we don't set `window.location.href` as the redirectUri when `PublicClientApplication` is instantiated.

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: The authModule no longer exists here, I need to find a way to expose the logger independently of the client.

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: The only async part of login is the authority discovery, so I would rather keep it void.

pkanher617: Moving it out of this PR.

lib/msal-browser/src/interaction_handler/RedirectHandler.ts
```diff
@@ -59,6 +59,7 @@ export class RedirectHandler extends InteractionHandler {
         const requestState = this.browserStorage.getItem(this.browserStorage.generateCacheKey(TemporaryCacheKeys.REQUEST_STATE), CacheSchemaType.TEMPORARY) as string;
         const cachedNonceKey = this.browserStorage.generateNonceKey(requestState);
         const cachedNonce = this.browserStorage.getItem(this.browserStorage.generateCacheKey(cachedNonceKey), CacheSchemaType.TEMPORARY) as string;
+        console.log(cachedNonce);
```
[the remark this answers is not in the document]
pkanher617: done.

lib/msal-common/src/request/RefreshTokenRequest.ts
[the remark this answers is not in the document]
pkanher617: done. Will remove the function in a separate PR

lib/msal-common/src/client/BaseClient.ts
pkanher617: done.

lib/msal-browser/src/app/PublicClientApplication.ts

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: will follow up in the next PR

lib/msal-browser/src/app/PublicClientApplication.ts

lib/msal-browser/src/app/PublicClientApplication.ts

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: removed, will add in a later PR.

lib/msal-browser/src/cache/BrowserStorage.ts
pkanher617: done.

lib/msal-common/src/authority/AuthorityFactory.ts
pkanher617: done.

lib/msal-browser/src/app/PublicClientApplication.ts

sangonzal:
Looks good. 

Only open question is about sending redirect URI as part of refresh token request. We should ensure that:
- Indeed it's necessary
- Including it as part of Node requests does not change or break behavior

lib/msal-common/src/authority/AuthorityFactory.ts
```diff
@@ -17,44 +17,71 @@ import { AdfsAuthority } from "./AdfsAuthority";
...
-    private static detectAuthorityFromUrl(authorityString: string): AuthorityType {
-        const authorityUrl = new UrlString(authorityString);
-        const components = authorityUrl.getUrlComponents();
-        const pathSegments = components.PathSegments;
+    static async createDiscoveredInstance(authorityUri: string, networkClient: INetworkModule, adfsDisabled?: boolean): Promise<Authority> {
+        // Initialize authority and perform discovery endpoint check.
+        const acquireTokenAuthority: Authority = AuthorityFactory.createInstance(authorityUri, networkClient);
+
+        // This is temporary. Remove when ADFS is supported for browser
+        if (adfsDisabled && acquireTokenAuthority.authorityType == AuthorityType.Adfs) {
```
tnorling, line 35: Can we just get rid of this? My understanding is that the only blocker at the moment is knowing what to tell customers when they ask how to configure their ADFS servers to support Auth Code. I believe the library should support it, so I would argue we shouldn't artificially block this.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -429,29 +404,33 @@ export class PublicClientApplication {
...
+            return await silentAuthClient.acquireToken(silentRequest);
         } catch (e) {
             const isServerError = e instanceof ServerError;
             const isInteractionRequiredError = e instanceof InteractionRequiredAuthError;
             const isInvalidGrantError = (e.errorCode === BrowserConstants.INVALID_GRANT_ERROR);
             if (isServerError && isInvalidGrantError && !isInteractionRequiredError) {
                 const silentAuthUrlRequest: AuthorizationUrlRequest = this.initializeRequest({
                     ...silentRequest,
-                    redirectUri: "",
+                    redirectUri: silentRequest.redirectUri,
```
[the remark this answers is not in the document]
pkanher617: In the redirect request, it is used to verify in token requests. I am verifying this behavior

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -511,7 +487,15 @@ export class PublicClientApplication {
      *
      */
     public getRedirectUri(): string {
-        return this.authModule.getRedirectUri();
+        if (this.config.auth.redirectUri) {
+            if (typeof this.config.auth.redirectUri === "function") {
```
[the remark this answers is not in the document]
pkanher617: not yet, we just moved it to msal-browser

lib/msal-common/src/authority/AuthorityFactory.ts
pkanher617: done.

lib/msal-browser/src/app/PublicClientApplication.ts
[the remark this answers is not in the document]
jasonnutter: I'm cool removing it in a follow PR for the next beta.

lib/msal-browser/src/app/PublicClientApplication.ts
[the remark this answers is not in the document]
jasonnutter: This request is for the silent handler, which uses iframes, where redirectUri is needed.

PR #1794, opened 2020-06-17
Retrieve account if already signed in

ahnmelis:
The PR will add functionality for the app to retrieve an existing signed in account (if there is any) by creating the `getAccount(Promise promise)` ReactMethod. This method will return a map with the account information if it exists and a message detailing the status otherwise. This method and two helper functions (`loadAccount()` and `mapAccount(IAccount account)` were created in MSALModule.java.
Because of Metro Server connectivity issues occurring after last merge, I recreated the SampleApp for the third (and hopefully last) time, which is why there are many random files modified again in this PR. I added functionality to App.js to check for an existing account using the getAccount function when it loads. The Sign In button correspondingly disables if an account is signed in.

For next steps, I am currently having trouble with retrieving information from the returned account map (a username, for example). I think it might have something to do with the "map" not being a WritableNativeMap anymore once it goes over the React bridge (maybe a JSON object?). 
If anyone has experience with React or React Native and promises, let me know!

### DIFF
--- a/lib/msal-react-native-android-poc/MSALModule/android/src/main/java/com/reactlibrary/MSALModule.java
+++ b/lib/msal-react-native-android-poc/MSALModule/android/src/main/java/com/reactlibrary/MSALModule.java
@@ -24,7 +24,6 @@ public class MSALModule extends ReactContextBaseJavaModule {
 
     private final ReactApplicationContext reactContext;
     private ISingleAccountPublicClientApplication publicClientApplication;
-    private IAccount mAccount;
 
     public MSALModule(ReactApplicationContext reactContext) {
         super(reactContext);
@@ -77,8 +76,7 @@ public class MSALModule extends ReactContextBaseJavaModule {
 
             //update account
                 Log.d(TAG, "Account: " + authenticationResult.getAccount().getUsername());
-                mAccount = authenticationResult.getAccount();
-                promise.resolve(mapMSALResult(authenticationResult));
+                promise.resolve(mapAccount(authenticationResult.getAccount()));
             }
 
             @Override
@@ -161,7 +159,7 @@ public class MSALModule extends ReactContextBaseJavaModule {
         WritableNativeMap map = new WritableNativeMap();
         //add attributes from account
         map.putString("authority", account.getAuthority());
-        map.putString("accountId", account.getId());
+        map.putString("id", account.getId());
         map.putString("tenantId", account.getTenantId());
         map.putString("username", account.getUsername());
         map.putString("idToken", (String) account.getClaims().get("id_token"));
--- a/lib/msal-react-native-android-poc/MSALModule/android/src/main/java/com/reactlibrary/MSALModule.java
+++ b/lib/msal-react-native-android-poc/MSALModule/android/src/main/java/com/reactlibrary/MSALModule.java
@@ -97,45 +97,44 @@ public class MSALModule extends ReactContextBaseJavaModule {
         };
     }
 
-     /**
+    /**
      * getAccount(): retrieves account currently signed in
      * Parameters: Promise promise (resolve will return the account as a map if it exists; reject will return null for nonexistent map or error)
      */
     @ReactMethod
     public void getAccount(Promise promise) {
-        //first load account and get status string
-        String message = loadAccount();
-        if (mAccount == null) {
-            promise.reject("loadaccountnull", message);
+        //if account is null, either no account is signed in or an error occured
+        IAccount account = loadAccount();
+        if (account == null) {
+            promise.reject("loadaccountnull", "No signed in account, or exception. Check Android log for details.");
         } else {
-            promise.resolve(mapAccount(mAccount));
+            promise.resolve(mapAccount(account));
         }
     }
 
     /**
-     * loadAccount(): will load a currently signed in account. Called in constructor.
-     * Returns a String with a message of the status (possibly an exception)
+     * loadAccount(): will load a currently signed in account. 
+     * Returns an IAccount if an account exists; returns null if no account is signed in or an error occurs
      */
 
-    private String loadAccount() {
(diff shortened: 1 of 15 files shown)


lib/msal-react-native-android-poc/MSALModule/android/src/main/java/com/reactlibrary/MSALModule.java
```diff
@@ -48,15 +56,15 @@ public String getName() {
     @ReactMethod
     public void signIn(String scopesValue, Promise promise) {
         if (!scopesValue.isEmpty()) {
```
tnorling, line 58: Not sure how Android implemented this, but in the js libraries empty scopes are actually valid for login calls since we will always add `openid` and `profile` if they were not provided. If msal android doesn't do this by default, consider adding that logic here

lib/msal-react-native-android-poc/MSALModule/android/src/main/java/com/reactlibrary/MSALModule.java
```diff
@@ -89,6 +97,47 @@ public void onCancel() {
...
 
+     /**
+     * getAccount(): retrieves account currently signed in
+     * Parameters: Promise promise (resolve will return the account as a map if it exists; reject will return null for nonexistent map or error)
+     */
+    @ReactMethod
+    public void getAccount(Promise promise) {
+        //first load account and get status string
+        String message = loadAccount();
+        if (mAccount == null) {
```
tnorling, line 108: I would try to avoid using a global mAccount and rather move your logic from `loadAccount` into this function so you can return it directly.

lib/msal-react-native-android-poc/MSALModule/android/src/main/java/com/reactlibrary/MSALModule.java
```diff
@@ -89,6 +97,47 @@ public void onCancel() {
...
+        }
+        String message = "";
+        try {
+            ICurrentAccountResult result = publicClientApplication.getCurrentAccount();
+            IAccount currAccount = result.getCurrentAccount();
+            if (currAccount == null) {
+                message = "No account currently signed in";
+            } else {
+                mAccount = currAccount;
+                message = "Retrieved signed in account";
```
tnorling, line 132: This message variable looks like it isn't being used unless an error occurs. Is that right? Consider removing this variable and instead add verbose logging here to help with debugging

lib/msal-react-native-android-poc/MSALModule/android/src/main/java/com/reactlibrary/MSALModule.java
ahnmelis: I think this is what Jason was talking about in my first PR and I misunderstood :) According to the docs for the Android sign in function, openid, profile, and offline access are always sent as scopes and it specifically says not to include these as scopes parameters. Problem is, when I try to pass in an empty scopes array, I still get an error. If I can't pass in an empty array nor the default scopes, I'm not really sure how we can just pass the default scopes. How is this implemented in the other JS libraries? 
https://docs.microsoft.com/en-us/java/api/com.microsoft.identity.client.isingleaccountpublicclientapplication.signin?view=azure-java-stable

lib/msal-react-native-android-poc/MSALModule/android/src/main/java/com/reactlibrary/MSALModule.java
ahnmelis: I rewrote loadAccount() to return the account directly. Logging will be handled by loadAccount(), and getAccount() will reject with a message to check the Android console if the returned account is null.

lib/msal-react-native-android-poc/MSALModule/android/src/main/java/com/reactlibrary/MSALModule.java
ahnmelis: Initially the message variable was for logging in the react console (currently there isn't really a way to log to the react console directly from the android module) but my implementation wasn't the best. I'll focus on logging in the android console to be consistent.

lib/msal-react-native-android-poc/MSALModule/android/src/main/java/com/reactlibrary/MSALModule.java
tnorling: I would suggest looking at the android sample to see what they pass to login calls. For the js libraries we do allow you to pass the default scopes and add them if they were not provided. We also do not throw an error if you pass an empty array. If you don't find an answer in the sample, let's follow up with the android team. It should be possible to only request the default scopes for sign ins.

lib/msal-react-native-android-poc/MSALModule/android/src/main/java/com/reactlibrary/MSALModule.java
tnorling: Changes look great!

lib/msal-react-native-android-poc/MSALModule/android/src/main/java/com/reactlibrary/MSALModule.java
ahnmelis: From what I can tell in their sample app, they pass in whatever their scopes textInput contains, even if that is empty. I believe then the auth callback would throw an exception (at least that's what happened in my experience). I'll message Shane to get more information about this.

ahnmelis: Shane says that I should just pass User.Read as a default, which is equivalent to openid profile. The Android team is going to look further into this issue.

ahnmelis:
Screenshot_1592504316
Updated sample app with email and disabled button

PR #1805, opened 2020-06-19
Changing datatype for JsonCache type

sameerag:
This PR:

* Adds the right type to the `JsonCache`, which is an intermediary data structure used to read and write serialized cache to a file.
* Introduces new types on `SerializationCacheTypes.ts` for stricter typechecking
* Modifies `Serializer.ts` and `Deserializer.ts` to respect types and adds inline conversions
* Removes `JSONKeys.ts` and `EntitySerializer.ts` since they are no longer needed with the new design


lib/msal-node/src/cache/CacheManager.ts
```diff
@@ -177,11 +177,11 @@ export class CacheManager {
...
-        const refreshTokens = oldState.RefreshToken != null ? this.mergeRemovalsStringDict(oldState.RefreshToken, newState.RefreshToken) : oldState.RefreshToken;
-        const idTokens = oldState.IdToken != null ? this.mergeRemovalsStringDict(oldState.IdToken, newState.IdToken) : oldState.IdToken;
-        const appMetadata = oldState.AppMetadata != null ? this.mergeRemovalsStringDict(oldState.AppMetadata, newState.AppMetadata) : oldState.AppMetadata;
+        const accounts = oldState.Account != null ? this.mergeRemovalsDict(oldState.Account, newState.Account) : oldState.Account;
+        const accessTokens = oldState.AccessToken != null ? this.mergeRemovalsDict(oldState.AccessToken, newState.AccessToken) : oldState.AccessToken;
+        const refreshTokens = oldState.RefreshToken != null ? this.mergeRemovalsDict(oldState.RefreshToken, newState.RefreshToken) : oldState.RefreshToken;
+        const idTokens = oldState.IdToken != null ? this.mergeRemovalsDict(oldState.IdToken, newState.IdToken) : oldState.IdToken;
+        const appMetadata = oldState.AppMetadata != null ? this.mergeRemovalsDict(oldState.AppMetadata, newState.AppMetadata) : oldState.AppMetadata;
```
jasonnutter, line 184: These could be `const appMetadata = oldState.AppMetadata && this.mergeRemovalsDict(oldState.AppMetadata, newState.AppMetadata)`

lib/msal-node/src/cache/CacheManager.ts
```diff
@@ -5,10 +5,10 @@
 
 import { Storage } from './Storage';
 import {
-    ClientAuthError, StringDict,
+    ClientAuthError,
     StringUtils
 } from '@azure/msal-common';
-import { InMemoryCache, JsonCache } from "cache/serializer/SerializerTypes";
+import { InMemoryCache, JsonCache, Dict } from "cache/serializer/SerializerTypes";
```
jasonnutter, line 11: I'm surprised this works, I would expect this to need a leading `./`: `./cache/serializer/SerializerTypes`

lib/msal-node/src/cache/serializer/SerializerTypes.ts
```diff
@@ -3,14 +3,18 @@
  * Licensed under the MIT License.
  */
 
-import { StringDict, AccountCache, IdTokenCache, AccessTokenCache, RefreshTokenCache, AppMetadataCache } from "@azure/msal-common";
+import {AccountCache, IdTokenCache, AccessTokenCache, RefreshTokenCache, AppMetadataCache } from "@azure/msal-common";
+
+export type Dict = {
```
jasonnutter, line 8: `Dict` is not specific enough, can this be something like StringDictionary?

lib/msal-node/src/cache/serializer/SerializerTypes.ts
```diff
@@ -3,14 +3,18 @@
...
+    [key: string]: object;
```
jasonnutter, line 9: Also, can this be `Record<string, object>`?

lib/msal-node/src/cache/CacheManager.ts
pkanher617: This will throw in the tests, and possibly the build - for some reason the vscode compiler doesn't catch it though

lib/msal-node/src/cache/CacheManager.ts
pkanher617: +1

lib/msal-node/src/cache/serializer/SerializerTypes.ts
pkanher617: +1

lib/msal-node/src/cache/serializer/SerializerTypes.ts
pkanher617: Is that an equivalent statement? Where is `Record` defined?

jasonnutter: Yes, those should be equivalent. It's a built in TS type.

lib/msal-node/src/cache/serializer/SerializerTypes.ts
sameerag: Updated the usage and got rid of `Dict`

lib/msal-node/src/cache/serializer/SerializerTypes.ts
sameerag: `Record<string, object>` is now used as appropriate and `Dict` is removed

lib/msal-node/src/cache/CacheManager.ts
sameerag: Addressed.

lib/msal-node/src/cache/CacheManager.ts
sameerag: To comply with testing of extensions and to keep this PR in scope, I would let @sangonzal address this in his extensions PR.

lib/msal-node/src/cache/CacheManager.ts
```diff
@@ -177,25 +181,25 @@ export class CacheManager {
...
-        const refreshTokens = oldState.RefreshToken != null ? this.mergeRemovalsStringDict(oldState.RefreshToken, newState.RefreshToken) : oldState.RefreshToken;
-        const idTokens = oldState.IdToken != null ? this.mergeRemovalsStringDict(oldState.IdToken, newState.IdToken) : oldState.IdToken;
-        const appMetadata = oldState.AppMetadata != null ? this.mergeRemovalsStringDict(oldState.AppMetadata, newState.AppMetadata) : oldState.AppMetadata;
+        const accounts = oldState.Account != null ? this.mergeRemovalsDict(oldState.Account, newState.Account) : oldState.Account;
+        const accessTokens = oldState.AccessToken != null ? this.mergeRemovalsDict(oldState.AccessToken, newState.AccessToken) : oldState.AccessToken;
+        const refreshTokens = oldState.RefreshToken != null ? this.mergeRemovalsDict(oldState.RefreshToken, newState.RefreshToken) : oldState.RefreshToken;
+        const idTokens = oldState.IdToken != null ? this.mergeRemovalsDict(oldState.IdToken, newState.IdToken) : oldState.IdToken;
+        const appMetadata = oldState.AppMetadata != null ? this.mergeRemovalsDict(oldState.AppMetadata, newState.AppMetadata): oldState.AppMetadata;
```
jasonnutter, line 188: Doesnt look like you are passing the type parameter here? `this.mergeRemovalsDict<Type>(oldState.AppMetadata, newState.AppMetadata)`

lib/msal-node/src/cache/serializer/Deserializer.ts
```diff
@@ -45,14 +54,19 @@ export class Deserializer {
...
+                    homeAccountId: serializedIdT.home_account_id,
+                    environment: serializedIdT.environment,
+                    credentialType: serializedIdT.credential_type,
+                    clientId: serializedIdT.client_id,
+                    secret: serializedIdT.secret,
+                    realm: serializedIdT.realm,
+                };
                 const idToken: IdTokenEntity = new IdTokenEntity();
                 CacheHelper.toObject(idToken, mappedIdT);
                 idObjects[key] = idToken;
```
jasonnutter, line 72: This can be addressed later, but I would use reduce here:

```js
return Object.keys(idTokens).reduce((idObjects, key) => {
  // ...
  idObjects[key] = idToken;
  return idObjects;
}, {});

File: lib/msal-node/src/cache/serializer/Serializer.ts [hunk 7]
```diff
@@ -34,10 +46,20 @@ export class Serializer {
...
      */
-    static serializeIdTokens(idTCache: IdTokenCache): StringDict{
-        const idTokens: StringDict = {};
-        Object.keys(idTCache).map(function (key) {
-            idTokens[key] = JSON.stringify(EntitySerializer.mapIdTokenKeys(idTCache, key));
+    static serializeIdTokens(
+        idTCache: IdTokenCache
+    ): Record<string, SerializedIdTokenEntity> {
+        const idTokens: Record<string, SerializedIdTokenEntity> = {};
+        Object.keys(idTCache).map(function(key) {
```
jasonnutter, line 53: Every place where you instantiate a new object, map over an existing object and add values onto the new object can be written to use reduce instead as shown above.

lib/msal-node/src/cache/CacheManager.ts
sameerag: Ah! Yes. Addressed.

lib/msal-node/src/cache/serializer/Deserializer.ts
sameerag: I remember trying `reduce` and it not working. Can check again and as you said will do it  in a different PR.

lib/msal-node/src/cache/serializer/Serializer.ts
sameerag: Please check the above comment. I will have to test once and will update in a follow up PR.

PR #1809, opened 2020-06-22
[msal-node] SilentFlow node interface

sameerag:
This PR adds support for:

* `acquireTokenSilent` in `msal-node`
* `getAccounts()` and `removeAccount` support in `msal-node/CacheManager.ts`
*  Sample for silent flow

Things to note:

* Unit tests for `Storage.ts` are not added as a follow up PR will change the functionality.
* Unit tests for `TokenCache.ts` are in the pipeline and will be a separate PR


lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -106,8 +107,27 @@ export abstract class ClientApplication {
...
+        );
+        const silentFlowClient = new SilentFlowClient(
+            silentFlowClientConfig
+        );
+        return silentFlowClient.acquireToken(this.initializeRequestScopes(request) as SilentFlowRequest);
+    }
+
     getCacheManager(): TokenCache {
-        return this.tokenCache;
+        return this.tokenCache;;
```
sangonzal, line 130: extra semi-colon

lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -106,8 +107,27 @@ export abstract class ClientApplication {
         return refreshTokenClient.acquireToken(this.initializeRequestScopes(request) as RefreshTokenRequest);
     }
 
+    /**
+     * Acquires a token silently when a user specifies the account the token is requested for.
+     *
+     * This API expects the user to provide an account object and looks into the cache to retrieve the token if present.
+     * There is also an optional "forceRefresh" boolean the user can send, to bypass the cache for access_token and id_token
+     * In case of an expired/unfound "refresh_token", an error is thrown
```
sangonzal, line 115: "In case refresh token is expired or no refresh token is found,"

lib/msal-node/src/cache/Storage.ts
```diff
@@ -245,7 +245,8 @@ export class Storage extends CacheManager {
 
         // read all keys
         Object.keys(cache).forEach(key => {
-            Object.keys(key).forEach(internalKey => {
+            // @ts-ignore
```
sameerag, line 248: The upcoming PR that addresses `storage` will get rid of this ts-ignore. Put this out to proceed with testing today.

sangonzal, line 248: Dead code here and line 266

samples/msal-node-silent-flow/data/cache.json
```diff
@@ -0,0 +1,110 @@
...
+            "secret": "an access token",
+            "realm": "microsoft",
+            "target": "scope4 scope5",
+            "client_id": "mock_client_id",
+            "cached_at": "1000",
+            "home_account_id": "uid.utid",
+            "extended_expires_on": "4600",
+            "expires_on": "4600"
+        },
+        "fb3680ec-7b98-4e1c-8434-5be16c5e78fd.90b8faa8-cc95-460e-a618-ee770bee1759-login.windows.net-accesstoken-99cab759-2aab-420b-91d8-5e3d8d4f063b-90b8faa8-cc95-460e-a618-ee770bee1759-openid profile user.read email": {
```
sangonzal, line 48: We shouldn't be committing tokens to samples. This is similar to committing passwords.

samples/msal-node-silent-flow/index.js
```diff
@@ -0,0 +1,71 @@
+/*
```
sangonzal, line 1: If we are not doing auth code as part of the sample, there is not need to have an express app. You can just show silent flow in a command line app (similar to the device code/refresh flow sample), without all of the extra overhead.

sameerag: I will be adding auth-code soon. Checked in just the happy case.

lib/msal-node/src/client/ClientApplication.ts
sameerag: Addressed.

lib/msal-node/src/client/ClientApplication.ts
sameerag: addressed

lib/msal-node/src/cache/Storage.ts
sameerag: Addressed this.

sameerag: It isn't dead code but rather syntax for `//ts-ignore`. This part of the code has changed and no longer applicable.

samples/msal-node-samples/msal-node-silent-flow/index.js
```diff
@@ -0,0 +1,71 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+const express = require("express");
+const msal = require('@azure/msal-node');
+const { promises: fs } = require("fs");
+
+const SERVER_PORT = process.env.PORT || 3000;
```
pkanher617, line 9: Can we make this default port 30662?

lib/msal-node/src/cache/Storage.ts
```diff
@@ -240,15 +240,14 @@ export class Storage extends CacheManager {
...
-        Object.keys(cache).forEach(key => {
-            Object.keys(key).forEach(internalKey => {
-                cacheKeys.push(internalKey);
-            });
-        });
+        cacheKeys = Object.keys(cache.accounts) as Array<string>;
+        cacheKeys.push(...Object.keys(cache.idTokens) as Array<string>);
+        cacheKeys.push(...Object.keys(cache.accessTokens) as Array<string>);
+        cacheKeys.push(...Object.keys(cache.refreshTokens) as Array<string>);
+        cacheKeys.push(...Object.keys(cache.appMetadata) as Array<string>);
```
jasonnutter, line 250:
```js
return [ 
  ...Object.keys(cache.accounts),
  ...Object.keys(cache.idTokens),
  ...Object.keys(cache.accessTokens),
  ...Object.keys(cache.refreshTokens),
  ...Object.keys(cache.appMetadata)
];

samples/msal-node-samples/msal-node-silent-flow/index.js
sangonzal: Find with changing it, but if we let's make all node samples use the same port.

sameerag: I updated @pkanher617 offline, lets make it uniform in a `samples` only PR

lib/msal-node/src/cache/Storage.ts
sameerag: Addressed this. Thanks, this looks much better.

PR #1813, opened 2020-06-23
Update merge logic and fix file formatting

sangonzal:


lib/msal-node/src/cache/TokenCache.ts
```diff
@@ -146,15 +154,13 @@ export class TokenCache {
      * @param newState
      */
     private mergeUpdates(oldState: any, newState: any): JsonCache {
-
-        let finalState = { ...oldState };
```
sangonzal, line 150: Had to revert this change.

PR #1815, opened 2020-06-23
Get token silently

ahnmelis:
acquireTokenSilent function implemented in MSALModule will return the results from the acquireTokenSilent MSAL Android method, notably the access token. In the sample app, the access token is then used to make a GET request to the Microsoft Graph API with the URL specified in the TextInput. 
The isSharedDevice function has also been added to the module to obtain the Device Mode of the app.

image

### DIFF
--- a/lib/msal-react-native-android-poc/MSALModule/android/src/main/java/com/reactlibrary/MSALModule.java
+++ b/lib/msal-react-native-android-poc/MSALModule/android/src/main/java/com/reactlibrary/MSALModule.java
@@ -22,6 +22,8 @@ import com.microsoft.identity.client.exception.MsalException;
 
 import static com.facebook.react.views.textinput.ReactTextInputManager.TAG;
 
+import java.util.Arrays;
+
 public class MSALModule extends ReactContextBaseJavaModule {
 
     private final ReactApplicationContext reactContext;
@@ -138,6 +140,20 @@ public class MSALModule extends ReactContextBaseJavaModule {
         
     }
 
+     /**
+     * isSharedDevice: returns a boolean on if device is shared or not
+     * Parameters: Promise promise (resolve will return a boolean and reject will return an exception)
+     */
+    @ReactMethod
+    public void isSharedDevice(Promise promise) {
+        try {
+            promise.resolve(publicClientApplication.isSharedDevice());
+        } catch (Exception error) {
+            Log.d(TAG, "Error in 'isSharedDevice': " + error.toString());
+            promise.reject(error);
+        }
+    }
+
     /**
      * signOut(): signs out the user currently signed in
      * Parameters: Promise promise (resolve will return a boolean true if account was successfully signed out and reject will return the exception)
@@ -159,7 +175,48 @@ public class MSALModule extends ReactContextBaseJavaModule {
         });
     }
 
-    /*
+    /**
+     * acquireTokenSilent: attempts to obtain a token silently from the cache
+     * Parameters: string scopesValue (string containing scopes separated by a " "), Promise promise (returned to an async function)
+     */
+    @ReactMethod
+    public void acquireTokenSilent(String scopesValue, Promise promise) {
+        //get account
+        IAccount account = loadAccount();
+        if (account == null) {
+            promise.reject("loadaccountnull", "acquireTokenSilent: No signed in account, or exception. Check Android log for details.");
+        } else if (!scopesValue.isEmpty()) {
+            publicClientApplication.acquireTokenSilentAsync(scopesValue.toLowerCase().split(" "), account.getAuthority(), getAuthSilentCallback(promise));
+        } else {
+            //the default is User.Read
+            String[] array = {"User.Read"};
+            publicClientApplication.acquireTokenSilentAsync(array, account.getAuthority(), getAuthSilentCallback(promise));
+        }
+    }
+
+    /**
+     * SilentAuthenticationCallback is the callback function for acquireTokenSlient.
+     * If successful, it will return a map of the IAuthenticationResult.
+     * If there's an error, the exception will be returned.
+     */
+
+    private SilentAuthenticationCallback getAuthSilentCallback(final Promise promise) {
+        return new SilentAuthenticationCallback() {
+            @Override
+            public void onSuccess(IAuthenticationResult authenticationResult) {
+                Log.d(TAG, "Acquire tokens silently success.");
+                promise.resolve(mapMSALResult(authenticationResult));
+            }
+
+            @Override
+            public void onError(MsalException exception) {
+                Log.d(TAG, "Error while acquiring tokens silently: " + exception.toString());
+                promise.reject(exception);
+            }
+        };
+    }
+
+    /**
     * Helper functions:
     * mapMSALResult(IAuthenticationResult result): used to convert AuthenticationResult into a map
     * so that it can be passed through a promise on the JS side
@@ -170,6 +227,12 @@ public class MSALModule extends ReactContextBaseJavaModule {
         //add attributes from account
         map.putMap("account", mapAccount(result.getAccount()));
         //add attributes from result
+        map.putString("accessToken", result.getAccessToken());
+        map.putString("authenticationScheme", result.getAuthenticationScheme());
+        map.putString("authorizationHeader", result.getAuthorizationHeader());
+        map.putString("expiresOn", result.getExpiresOn().toString());
+        map.putString("scope", Arrays.toString(result.getScope()));
+        map.putString("tenantId", result.getTenantId());
         return map;
     }
 
--- a/lib/msal-react-native-android-poc/SampleApp/App.js
+++ b/lib/msal-react-native-android-poc/SampleApp/App.js
@@ -36,9 +36,13 @@ class App extends Component {
     };
   }
 
-  componentDidMount() {
-    this.getAccount();
-    this.isSharedDevice();
+  async componentDidMount() {
+    try {
+      await this.getAccount();
+      await this.isSharedDevice();
+    } catch (error) {
+      console.log(error);
+    }
   }
 
   getAccount = async () => {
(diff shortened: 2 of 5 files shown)


lib/msal-react-native-android-poc/SampleApp/App.js
```diff
@@ -28,17 +28,30 @@ class App extends Component {
...
       aUsername: null,
       aTenantId: null,
       aIdToken: null,
       aAuthority: null,
       aId: null,
-      resultMap: null
+      aAccessToken: null,
+      aAuthenticationScheme: null,
+      aAuthorizationHeader: null,
+      aExpiresOn: null,
```
ahnmelis, line 42: Is there a good way to store objects (for example, an Account object with the results from getAccount() ) as state variables? I couldn't find good resources online, so I resorted to creating a state variable for all attributes.

lib/msal-react-native-android-poc/MSALModule/android/src/main/java/com/reactlibrary/MSALModule.java
```diff
@@ -159,7 +175,48 @@ public void onError(@NonNull MsalException exception) {
...
-    /*
+    /**
+     * acquireTokenSilent: attempts to obtain a token silently from the cache
+     * Parameters: string scopesValue (string containing scopes separated by a " "), Promise promise (returned to an async function)
+     */
+    @ReactMethod
+    public void acquireTokenSilent(String scopesValue, Promise promise) {
+        //get account
+        IAccount account = loadAccount();
+        if (account == null) {
```
technical-boy, line 186: I think you can just do `if  (!account)` here.

lib/msal-react-native-android-poc/SampleApp/App.js
technical-boy: Can you not add the Account object itself as a state variable? You could then do, for example,  `this.state.account.property`.

Is that what you mean?

lib/msal-react-native-android-poc/SampleApp/App.js
```diff
@@ -93,8 +117,9 @@ class App extends Component {
                 <Text style={styles.textSubHeader}> MSGraph URL: </Text>
                 <TextInput
                 style={styles.textInput}
-                placeholder="Some URL"
-                underlineColorAndroid='#8a8886' />
+                defaultValue='https://graph.microsoft.com/v1.0/me'
```
technical-boy, line 120: I think making this a constant string (maybe at the top of the file if you only use it in App.js, or in a constants directory somewhere else that you can import, may save you a few headaches down the line.

lib/msal-react-native-android-poc/SampleApp/App.js
```diff
@@ -175,14 +201,51 @@ class App extends Component {
                 style={styles.button}>
                   <Button 
                   title="Get Graph Data Silently"
-                  color ='#2b88d8' />
+                  color ='#2b88d8'
+                  onPress = {async () => {
```
technical-boy, line 205: I would suggest moving the logic to a method outside render, then calling it here. Since this is where the main action happens, having it in the middle of the markdown can make it less readable.

lib/msal-react-native-android-poc/SampleApp/App.js
```diff
@@ -175,14 +201,51 @@ class App extends Component {
...
                 </View>
               </View>
             </View>
            <View
            style={styles.viewResult}>
              <Text>{this.state.isSignedIn ?  "Welcome, " + this.state.aUsername : "Welcome! Please sign in."}</Text>
            </View>
+           <View
+           style={styles.viewResult}>
+             <Text>{this.state.graphResult != null ? JSON.stringify(this.state.graphResult, null, 4) : ""}</Text>
```
technical-boy, line 247: I would also recommend moving this to it's own method, mainly because it contains more logic than your average ternary assignment, so it would benefit from being more readable.

lib/msal-react-native-android-poc/SampleApp/App.js
jo-arroyo: +1

lib/msal-react-native-android-poc/MSALModule/android/src/main/java/com/reactlibrary/MSALModule.java
jo-arroyo: +1

lib/msal-react-native-android-poc/SampleApp/App.js
ahnmelis: I've been having trouble with setting a variable directly equal to my module functions, but I tried it again for the first time in a while and it worked... I wonder if my initial buggy signIn function had an impact on this. I was able to created an object for the account and deleted the remaining variables pertaining to the acquireTokenSilent function (since they don't need to be state variables in my sample app). This cleaned up my code considerably!

lib/msal-react-native-android-poc/MSALModule/android/src/main/java/com/reactlibrary/MSALModule.java
ahnmelis: I'm getting a "bad operand type IAccount for unary operator '!'" error with (!account) ....

lib/msal-react-native-android-poc/SampleApp/App.js
ahnmelis: Doing this has definitely crossed my mind, but Ive been having issues with moving the logic to a method within the class. For some reason, when I end up calling the wrapper signIn function inside the 'onPress" prop for my signIn button, the function gets called continuously without me even clicking anything. 
Let me know if you have an idea of what I could be doing wrong... I'll keep trying to debug this.

ahnmelis: it was as simple as dropping the () from my function call. I'm still a bit confused as to why the parentheses would cause it to be continuously called, but I'm just glad it works now

lib/msal-react-native-android-poc/SampleApp/App.js
ahnmelis: done

lib/msal-react-native-android-poc/SampleApp/App.js
ahnmelis: url string coded in one place (in the constructor as URLValue) and now the defaultValue prop references it

lib/msal-react-native-android-poc/MSALModule/android/src/main/java/com/reactlibrary/MSALModule.java
jasonnutter: @sangonzal

lib/msal-react-native-android-poc/SampleApp/App.js
```diff
@@ -28,38 +28,95 @@ class App extends Component {
...
+      this.setState({
+        graphResult: json
+      });
+    } catch (exception) {
+      console.log(exception);
+    }
+  }
+
+  showResult = () => {
+   return (this.state.graphResult != null ? JSON.stringify(this.state.graphResult, null, 4) : "");
```
jasonnutter, line 118: This can be `return this.state.graphResult && JSON.stringify(this.state.graphResult, null, 4)`

lib/msal-react-native-android-poc/SampleApp/App.js
```diff
@@ -28,38 +28,95 @@ class App extends Component {
...
+
+  startGetGraphDataSilently = async () => {
+    try {
+      var result = await MSAL.acquireTokenSilent(this.state.scopesValue);
+      //make api call
+      var response = await fetch (
+        this.state.URLValue, {
+          method: 'GET',
+          headers: {
+            'Authorization': result.accessToken
```
jasonnutter, line 104: I believe this needs to be 

```
Authorization: `Bearer ${result.accessToken}`

jasonnutter:
A coupe things otherwise looks good!

lib/msal-react-native-android-poc/SampleApp/App.js
ahnmelis: done. Out of curiosity, why does my initial statement still work if this is the correct way to code it?

lib/msal-react-native-android-poc/SampleApp/App.js
ahnmelis: Done. && in this case means return the second expression if the first is considered truthy, and if not return the first?

lib/msal-react-native-android-poc/SampleApp/App.js
technical-boy: So, this is a fairly complicated problem that shows up a lot in React and (I assume) React Native. My own understanding of this is pretty limited.

When you set an onPress/Click event handler, your goal is to tell the handler what method or code to execute when the event is triggered.

Sending an arrow function (i.e. () => {}) means that the arrow function contents will be evaluated on every call to the component's render method. Given that the render method is called at many stages of the component's lifecycle, you have to be careful about how you send in the code you want to be executed.

Sending in the method with the parentheses means the function will be executed every time execution reaches that line of code. However, wrapping the method in an arrow function and removing the parentheses from the statement allows you assign the callback/function to the event, rather than actually having it be executed when the line is "read". This way the actual contents of the method will only be executed when the actual onPress event is triggered.

If this is still confusing, let me know. I think someone like @jasonnutter with more React experience may be able to explain it in simpler terms/better than I can.

ahnmelis: This makes a lot more sense. Thank you for the awesome explanation!

lib/msal-react-native-android-poc/SampleApp/App.js
jasonnutter: Good question, not sure. Maybe it defaults to the `Bearer` type if not provided.

PR #1823, opened 2020-06-24
Add monitorIframeForHash to ensure silent requests timeout

jasonnutter:
We have seen a class of silent token requests that appear to take longer than the configured timeout, but still succeed (when we would expect them to throw an error). We believe this is due to the browser tab being suspended, which stops async processing, such as `setInterval`. Since `monitorWindowForHash` was based on ticks, this meant that a request could take longer than the configured timeout (e.g. interval runs -> tab suspended for non-trivial duration -> interval resumes).

To prevent this, I have split `monitorWindowForHash` into two fuctions, `monitorIframeForHash` and `monitorPopupForHash`:
 
- `monitorPopupForHash` is the old `monitorWindowForHash` function, with the code for silent flow removed. This needs to remain tick-based, since we do not want to include interaction time when calculating timeouts.
- `monitorIframeForHash` is a new function that is purely time-based, since we don't need to account for interaction for silent calls.

This should result in requests that take longer than the configured timeout to result in an error, instead of succeeding.


lib/msal-core/src/utils/WindowUtils.ts
```diff
@@ -40,13 +41,53 @@ export class WindowUtils {
...
+                    clearInterval(intervalId);
+                    resolve(contentWindow.location.hash);
+                } 
+            }, WindowUtils.POLLING_INTERVAL_MS);
+        });
+    }
+
     /**
      * @hidden
      * Monitors a window until it loads a url with a hash
```
jasonnutter, line 84: TODO: update comment

lib/msal-core/src/utils/WindowUtils.ts
```diff
@@ -40,13 +41,53 @@ export class WindowUtils {
         return `${prefix}${Constants.resourceDelimiter}${requestSignature}`;
     }
 
+    static monitorIframeForHash(contentWindow: Window, timeout: number, urlNavigate: string, logger: Logger): Promise<string> {
```
jasonnutter, line 44: TODO: add comment

lib/msal-core/src/utils/WindowUtils.ts
```diff
@@ -40,13 +41,53 @@ export class WindowUtils {
...
+                    /*
+                     * Will throw if cross origin,
+                     * which should be caught and ignored
+                     * since we need the interval to keep running while on STS UI.
+                     */
+                    href = contentWindow.location.href;
+                } catch (e) {}
+
+                if (href && UrlUtils.urlContainsHash(href)) {
+                    logger.verbose("monitorIframeForHash found url in hash");
```
jasonnutter, line 74: TODO: fix comment

pkanher617 (approved):
One question but otherwise looks good, would like to see any changes that need to be made in 2.0 in a follow up PR (Happy to help pair program on this).

lib/msal-core/src/utils/TimeUtils.ts
```diff
@@ -26,4 +26,11 @@ export class TimeUtils {
     static now(): number {
         return Math.round(new Date().getTime() / 1000.0);
     }
+
+    /**
+     * Returns the amount of time in milliseconds since the page loaded.
+     */
+    static relativeNowMs(): number {
+        return window.performance.now();
```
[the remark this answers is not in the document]
jasonnutter: `window.performance.now` returns high resolution duration in milliseconds, relative to the start of the page load. We don't need to use `Date`, since we are calculating a relative duration. https://developer.mozilla.org/en-US/docs/Web/API/Performance/now#Example

jasonnutter:
> One question but otherwise looks good, would like to see any changes that need to be made in 2.0 in a follow up PR (Happy to help pair program on this).

Yes, we'll want to do this in v2 for SilentHandler.

tnorling (approved):

abellisai:
Hi, do you have any idea about release date of msal@1.3.3 ?

Thanks

jasonnutter:
> Hi, do you have any idea about release date of msal@1.3.3 ?
> 
> Thanks

Not yet, should be soon though.

PR #1827, opened 2020-06-24
[msal-node][msal-common][msal-browser] Console logs to instrumentation

jo-arroyo:
This PR aims to remove console logs from msal-node and msal-common whenever possible, and replaces them with errors. Unit tests are also added to ensure errors are correctly thrown.

msal-node
- Storage.ts
  - `setItem`
  - `getItem`
  - `removeItem`

msal-common
- AccountEntity.ts
  - `generateType`
- CredentialEntity.ts
  - `generateType` 

msal-browser
- Docs including: acquire-token.md, token-lifetimes.md, v1-migration.md
- BrowserStorage.ts
  - `setItem`
  - `getItem`


lib/msal-node/src/cache/Storage.ts
```diff
@@ -154,8 +154,8 @@ export class Storage extends CacheManager {
                 return (cache.appMetadata[key] as AppMetadataEntity) || null;
             }
             default: {
-                console.log('Invalid Cache Type');
-                return {};
+                throw ClientAuthError.createInvalidCacheTypeError();
+                return {}; //TODO: what about this?
```
sangonzal, line 158: You can delete it. If you throw, the return statement will not be executed.

jo-arroyo: Great, thanks!

lib/msal-common/src/request/RequestValidator.ts
```diff
@@ -80,7 +80,6 @@ export class RequestValidator {
         // Remove any query parameters already included in SSO params
         queryParams.forEach((value, key) => {
             if (eQParams[key]) {
-                console.log("Removed param " + key + " from extraQueryParameters since it was already present in library query parameters.");
```
jo-arroyo, line 83: Didn't seem appropriate for this to throw an error, since it is just noting that extra parameters have been removed, and must continue to go through the rest of the params to remove duplicates. Not able to use a logger in static. Thinking that this log may not be necessary after all.

lib/msal-browser/docs/v1-migration.md
```diff
@@ -111,10 +111,10 @@ async function signIn(method) {
 
 async function getTokenPopup(request) {
     return await myMSALObj.acquireTokenSilent(request).catch(async (error) => {
-        console.log("silent token acquisition fails. acquiring token using popup");
+        this.logger.info("silent token acquisition fails. acquiring token using popup");
         // fallback to interaction when silent call fails
         return await myMSALObj.acquireTokenPopup(request).catch(error => {
-            console.log(error);
+            handleError(error);
```
sameerag, line 117: Are we adding a note what `handleError(error)` should be?

lib/msal-common/src/request/RequestValidator.ts
sameerag: I agree that error isn't appropriate. The log message is needed though, lets figure out how to fix this. Can be in a later PR.

lib/msal-browser/docs/v1-migration.md
jo-arroyo: @pkanher617 Do you have any input? After our discussion on Friday I just changed it to just handleError, but am happy to put something else that is more descriptive

PR #1828, opened 2020-06-24
Complete Single Account Mode MSAL for React Native

ahnmelis:
This PR adds the acquireToken function to MSALModule and utilizes it in the sample app. 
So far, the module should be able to wrap MSAL for Android's single account mode functions for AAD, which is the minimum project goal I set for myself. It would be good to get all this polished before I move on to more features, which could include multiple accounts mode.

### DIFF
--- a/lib/msal-react-native-android-poc/MSALModule/android/src/main/java/com/reactlibrary/MSALModule.java
+++ b/lib/msal-react-native-android-poc/MSALModule/android/src/main/java/com/reactlibrary/MSALModule.java
@@ -22,6 +22,9 @@ import com.microsoft.identity.client.exception.MsalException;
 
 import static com.facebook.react.views.textinput.ReactTextInputManager.TAG;
 
+//import DEFAULT_SCOPES from Constants.java
+import static com.reactlibrary.Constants.DEFAULT_SCOPES;
+
 import java.util.Arrays;
 
 public class MSALModule extends ReactContextBaseJavaModule {
--- a/lib/msal-react-native-android-poc/MSALModule/android/src/main/java/com/reactlibrary/MSALModule.java
+++ b/lib/msal-react-native-android-poc/MSALModule/android/src/main/java/com/reactlibrary/MSALModule.java
@@ -176,6 +176,49 @@ public class MSALModule extends ReactContextBaseJavaModule {
     }
 
     /**
+     * acquireToken: attempts to obtain a token interactively
+     * Parameters: string scopesValue (string containing scopes separated by a " "), Promise promise (returned to an async function)
+     */
+    @ReactMethod
+    public void acquireToken(String scopesValue, Promise promise) {
+        if (!scopesValue.isEmpty()) {
+            publicClientApplication.acquireToken(getCurrentActivity(), scopesValue.toLowerCase().split(" "), getAuthInteractiveCallback(promise));
+        } else {
+            //the default is User.Read
+            String[] array = {"User.Read"};
+            publicClientApplication.acquireToken(getCurrentActivity(), array, getAuthInteractiveCallback(promise));
+        }
+    }
+
+    /**
+     * getAuthInteractiveCallback is the callback function for acquireToken.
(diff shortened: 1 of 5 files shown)


lib/msal-react-native-android-poc/MSALModule/android/src/main/java/com/reactlibrary/MSALModule.java
```diff
@@ -175,6 +175,49 @@ public void onError(@NonNull MsalException exception) {
...
+    /**
+     * getAuthInteractiveCallback is the callback function for acquireToken.
+     * If successful, it will return a map of the IAuthenticationResult.
+     * If there's an error, the exception will be returned.
+     */
+
+    private AuthenticationCallback getAuthInteractiveCallback(final Promise promise) {
+        return new AuthenticationCallback() {
+            @Override
+            public void onCancel() {
```
tnorling, line 202: Is this event raised by MSAL android or the end-user? We have cleanup code for things like cache when a login event is cancelled, you'll want to make sure this is still being done.

lib/msal-react-native-android-poc/SampleApp/android/app/src/main/java/com/sampleapp/MSALModule.java
```diff
@@ -175,6 +175,49 @@ public void onError(@NonNull MsalException exception) {
...
+     * acquireToken: attempts to obtain a token interactively
+     * Parameters: string scopesValue (string containing scopes separated by a " "), Promise promise (returned to an async function)
+     */
+    @ReactMethod
+    public void acquireToken(String scopesValue, Promise promise) {
+        if (!scopesValue.isEmpty()) {
+            publicClientApplication.acquireToken(getCurrentActivity(), scopesValue.toLowerCase().split(" "), getAuthInteractiveCallback(promise));
+        } else {
+            //the default is User.Read
+            String[] array = {"User.Read"};
```
tnorling, line 188: This is another spot you could use the constant suggested above

lib/msal-react-native-android-poc/MSALModule/android/src/main/java/com/reactlibrary/MSALModule.java
```diff
@@ -175,6 +175,49 @@ public void onError(@NonNull MsalException exception) {
...
+     * acquireToken: attempts to obtain a token interactively
+     * Parameters: string scopesValue (string containing scopes separated by a " "), Promise promise (returned to an async function)
+     */
+    @ReactMethod
+    public void acquireToken(String scopesValue, Promise promise) {
+        if (!scopesValue.isEmpty()) {
+            publicClientApplication.acquireToken(getCurrentActivity(), scopesValue.toLowerCase().split(" "), getAuthInteractiveCallback(promise));
+        } else {
+            //the default is User.Read
+            String[] array = {"User.Read"};
```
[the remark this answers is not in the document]
pkanher617: +1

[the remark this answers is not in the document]
ahnmelis: Done! Created Constants.java in the same directory. array is now DEFAULT_SCOPES

lib/msal-react-native-android-poc/MSALModule/android/src/main/java/com/reactlibrary/MSALModule.java
ahnmelis: I'm waiting from a response from the android team regarding cleanup code, but I believe the event is raised by the end user (when they back out of the auth flow).  Looking at their sample app, all they have within their onCancel is a logging statement. So hopefully I won't have to worry about clearing the cache but I'll check to make sure.

ahnmelis: Shane says that MSAL Android doesn't create anything in the token cache until there is an actual token response. So when the user cancels it wouldn't have gotten that far yet

PR #1829, opened 2020-06-25
Fix issues with cookie length in msal-core

### LINKED ISSUE #1188: Prevent MSAL from generating cookies that exceed a set limit
When `storeAuthStateInCookie` is enabled, MSAL will store information in cookies on the application domain. These cookies are meant only for MSAL, but they could get sent with an AJAX request sent to the server, and if these cookies are too long for the server, then the server will likely return a 400 error (as in #1174). When creating cookies, MSAL should first inspect the current set of cookies, and if the new cookie length will exceed some limit (e.g. 4096), then perform logic to ensure the new cookie will not exceed that limit.

This limit should be configurable by an application, including the option to disable the limit completely.

AC:
1. Cookies created by MSAL do not exceed a known limit set in the library. Any truncation that occurs should only remove values set by MSAL.
2. Default limit is 4096.
3. Limit can be disabled.
4. If possible, find root cause of cookies not getting deleted (as in #1174).

jasonnutter:
Includes the following cookie-related fixes/enhancements:
- Helper function added to compute temporary cache keys.
- Temporary cache keys now include just the state id (a guid), instead of the entire encoded state value (which can be quite long). This results in a shorter cache key names, which results in shorter cookie length.
- `getAccount` will attempt to read from cookies, if enabled.
- Items that are stored in cookies will no longer have the rollback cache key enabled, to reduce the number of cookies created.
- Add `clearItemCookie` to `AuthCache`, to ensure we are properly deleting cookies.
- If `state` is passed to `clearMsalCookie`, cookies with that state in the name will be removed, otherwise, all msal cookies will be removed (this is called in logout)

These should have the following outcomes:
- Ensure we are generate the minimum number of cookies.
- Ensure those cookies are as small as possible.
- Ensure we are deleting cookies when they are not needed.


lib/msal-core/test/Storage.sessionStorage.spec.ts
```diff
@@ -291,37 +292,20 @@ describe("CacheStorage.ts Class - Session Storage", function () {
...
-            expect(msalCacheStorage.getItem(`${TemporaryCacheKeys.NONCE_IDTOKEN}|RANDOM_GUID`)).to.be.null;
-            expect(msalCacheStorage.getItem(`${TemporaryCacheKeys.RENEW_STATUS}|RANDOM_GUID`)).to.be.null;
+            expect(msalCacheStorage.getItem(`${AuthCache.generateTemporaryCacheKey(TemporaryCacheKeys.NONCE_IDTOKEN, TEST_STATE)}`)).to.be.null;
+            expect(msalCacheStorage.getItem(`${AuthCache.generateTemporaryCacheKey(TemporaryCacheKeys.RENEW_STATUS, TEST_STATE)}`)).to.be.null;
         });
 
         it.skip("tests that resetCacheItems only deletes instance-specific cache items");
     });
-
-    describe("static key generators", function () {
```
jasonnutter, line 311: These tests were just duplicates of the ones in the localStorage test file.

lib/msal-core/src/cache/AuthCache.ts
```diff
@@ -83,9 +84,11 @@ export class AuthCache extends BrowserStorage {// Singleton
      * @param value
      * @param enableCookieStorage
      */
-    setItem(key: string, value: string, enableCookieStorage?: boolean, state?: string): void {
```
jasonnutter, line 86: `state` was unused.

lib/msal-core/src/cache/AuthCache.ts
```diff
@@ -83,9 +84,11 @@ export class AuthCache extends BrowserStorage {// Singleton
...
+    setItem(key: string, value: string, enableCookieStorage?: boolean): void {
         super.setItem(this.generateCacheKey(key, true), value, enableCookieStorage);
-        if (this.rollbackEnabled) {
+
+        // Values stored in cookies will have rollback disabled to minimize cookie length
+        if (this.rollbackEnabled && !enableCookieStorage) {
             super.setItem(this.generateCacheKey(key, false), value, enableCookieStorage);
         }
```
jasonnutter, line 93: Rollback was generating two cookies for each value, greatly increasing the overall cookie length.

lib/msal-core/src/cache/AuthCache.ts
```diff
@@ -128,16 +131,20 @@ export class AuthCache extends BrowserStorage {// Singleton
...
      */
-    resetTempCacheItems(state: string): void {
+    resetTempCacheItems(state?: string): void {
+        const stateId = state && RequestUtils.parseLibraryState(state).id;
+        const isTokenRenewalInProgress = this.tokenRenewalInProgress(state);
+
         const storage = window[this.cacheLocation];
         let key: string;
         // check state and remove associated cache
-        for (key in storage) {
```
jasonnutter, line 135: This was causing the loop to be run with values from the storage's prototype, e.g. `getItem`, `setItem`. Objects.keys ensure we are just looping the values that are in local storage.

lib/msal-core/src/cache/AuthCache.ts
tnorling: What is rollback used for?

tnorling (approved):
Just one general clarifying question but LGTM

lib/msal-core/src/cache/AuthCache.ts
jasonnutter: Storage keys that did not include the client id.

lib/msal-core/src/cache/AuthCache.ts
```diff
@@ -217,14 +246,26 @@ export class AuthCache extends BrowserStorage {// Singleton
...
-        return `${TemporaryCacheKeys.ACQUIRE_TOKEN_ACCOUNT}${Constants.resourceDelimiter}${accountId}${Constants.resourceDelimiter}${state}`;
+        const stateId = RequestUtils.parseLibraryState(state).id;
+        return `${TemporaryCacheKeys.ACQUIRE_TOKEN_ACCOUNT}${Constants.resourceDelimiter}${accountId}${Constants.resourceDelimiter}${stateId}`;
     }
 
     /**
      * Create authorityKey to cache authority
      * @param state
      */
     public static generateAuthorityKey(state: string): string {
```
jmckennon, line 257: what's the benefit of having both `generateAuthorityKey` and `generateTemporaryCacheKey`?

jasonnutter: It was already there and decided not to remove it. Just a convenience method.

jmckennon (approved):

PR #1830, opened 2020-06-25
[msal-extensions] Extensions #1

sangonzal:
- Adds package.json, README, tsconfig, and sets directory structure
- Adds Windows DPAPI node addon


extensions/.gitignore
```diff
@@ -0,0 +1,77 @@
+# Webstorm
```
pkanher617, line 1: Does the .gitignore need to be separate?

pkanher617 (approved):

extensions/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "msal-node-extensions",
```
jasonnutter, line 2: These files should go in `extensions/msal-node-extensions`, and the name of the package should be `@azure/msal-node-extensions` or similar.

extensions/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+MSAL Node supports an in-memory cache by default and provides the ICacheManager to perform cache serialization. Developers are required to implement their own cache persistance across multiple platforms and Microsoft Authentication Extensions makes this simpler.
+
+The supported platforms are Windows, Mac and Linux.
+
+Windows - Dpapi is used for encryption.
+
+MAC - The MAC KeyChain is used.
+
+Linux - LibSecret is used for encryption.
```
jasonnutter, line 12: I would have these as bullets:

```md
Supported platforms are Windows, macOS, and Linux:

- Windows: DPAPI is used for encryption.
- macOS: The macOS Keychain is used for encryption.
- Linux: Libsecret is used for encryption.

File: extensions/README.md [hunk 4]
```diff
@@ -0,0 +1,42 @@
+# Microsoft Authentication Extensions for Node
+The Microsoft Authentication Extensions for Node offers secure mechanisms for client applications to perform cross-platform token cache serialization and persistence. It gives additional support to the Microsoft Authentication Library for Node (MSAL).
+
+MSAL Node supports an in-memory cache by default and provides the ICacheManager to perform cache serialization. Developers are required to implement their own cache persistance across multiple platforms and Microsoft Authentication Extensions makes this simpler.
+
+The supported platforms are Windows, Mac and Linux.
```
jasonnutter, line 6: We should consider being specific about which versions of Linux we have tested.

extensions/README.md
```diff
@@ -0,0 +1,42 @@
...
+> Note: It is recommended to use this library for cache persistence support for Public client applications such as Desktop apps only. In web applications, this may lead to scale and performance issues. Web applications are recommended to persist the cache in session.
+
+
+## Building
+
+The extensions contain prebuild binaries. To build from source, you will need Python on you path.
+
+- Navigate to `lib/msal-common` and run `npm run build` then `npm link`
+- Navigate to `extensions` and run `npm link @azure/msal-common`
+- Install [node-gyp](https://github.com/nodejs/node-gyp) by running `npm install -g node-gyp`
```
jasonnutter, line 23: I believe we can include this as a dev dependency.

jasonnutter (changes requested):

extensions/README.md
sangonzal: Updated

extensions/package.json
sangonzal: Updated

extensions/README.md
sangonzal: Updated

extensions/README.md
sangonzal: Agreed. Added item to backlog to test on multiple distributions and add to README

extensions/msal-node-extensions/src/dpapi-addon/dpapi_win.cpp
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	if (info[2]->IsNullOrUndefined() || !info[2]->IsString())
+	{
+		isolate->ThrowException(v8::Exception::TypeError(
+			CreateUtf8String(isolate, "Third argument, scope, must be a string")));
+	}
+
+	DWORD flags = 0;
+	if (!info[2]->IsNullOrUndefined())
```
jbzdarkid, line 48: check is redundant? Didn't you already ensure that this was false on L41?

extensions/msal-node-extensions/src/dpapi-addon/dpapi_win.cpp
```diff
@@ -0,0 +1,117 @@
...
+	{
+		isolate->ThrowException(v8::Exception::TypeError(
+			CreateUtf8String(isolate, "First argument, data, must be a valid Uint8Array")));
+	}
+
+	if (!info[1]->IsNull() && !info[1]->IsUint8Array())
+	{
+		isolate->ThrowException(v8::Exception::TypeError(
+			CreateUtf8String(isolate, "Second argument, optionalEntropy, must be null or an ArrayBuffer")));
+	}
```
jbzdarkid, line 39: I'd recommend constructing the correct C++ types ASAP after verifying the input types, that way you can work directly in C++, without having to think about v8 variadic types.

extensions/msal-node-extensions/src/dpapi-addon/dpapi_win.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+	auto buffer = node::Buffer::Data(info[0]);
+	auto len = node::Buffer::Length(info[0]);
+
+	DATA_BLOB entropyBlob;
+	entropyBlob.pbData = nullptr;
+	if (!info[1]->IsNull())
+	{
+		entropyBlob.pbData = reinterpret_cast<BYTE*>(node::Buffer::Data(info[1]));
+		entropyBlob.cbData = node::Buffer::Length(info[1]);
```
jbzdarkid, line 66: Do you have a guarantee that both of these buffers are null-terminated?
Or, does `node::Buffer::Length` have some other form of safety check to prevent an overrun?

extensions/msal-node-extensions/README.md
```diff
@@ -0,0 +1,39 @@
+# Microsoft Authentication Extensions for Node
+The Microsoft Authentication Extensions for Node offers secure mechanisms for client applications to perform cross-platform token cache serialization and persistence. It gives additional support to the Microsoft Authentication Library for Node (MSAL).
+
+MSAL Node supports an in-memory cache by default and provides the ICacheManager to perform cache serialization. Developers are required to implement their own cache persistance across multiple platforms and Microsoft Authentication Extensions makes this simpler.
```
sameerag, line 4: `persistence`

extensions/msal-node-extensions/README.md
```diff
@@ -0,0 +1,39 @@
...
+
```
sameerag, line 5: A little unclear. Shouldn't we say `msal-extensions-for-node` is a default implementation of cache persistence for developers who choose to not implement their own persistence?

extensions/msal-node-extensions/README.md
```diff
@@ -0,0 +1,39 @@
...
+Supported platforms are Windows, Mac and Linux:
+
+- Windows - DPAPI is used for encryption.
+- MAC - The MAC KeyChain is used.
+- Linux - LibSecret is used for encryption.
```
sameerag, line 10: make it a table with `Platform` and `Description` as fields?

extensions/msal-node-extensions/README.md
sangonzal: Updated

extensions/msal-node-extensions/README.md
sangonzal: Updated

extensions/msal-node-extensions/README.md
sangonzal: I don't think theres a need for a table here, bullet points work. Can always add it later if more stuff is added.

extensions/msal-node-extensions/src/dpapi-addon/dpapi_win.cpp
sangonzal: Yep, it's redundant. Removed.

sameerag (approved):
Approving.

extensions/msal-node-extensions/src/dpapi-addon/dpapi_win.cpp
```diff
@@ -0,0 +1,114 @@
...
+			CreateUtf8String(isolate, "Third argument, scope, must be a string")));
+	}
+
+	DWORD flags = 0;
+	v8::String::Utf8Value strData(isolate, info[2]);
+	std::string scope(*strData);
+	if (stricmp(scope.c_str(), "LocalMachine") == 0)
+	{
+		flags = CRYPTPROTECT_LOCAL_MACHINE;
+	}
```
jbzdarkid, line 53: Why are you passing a string parameter if all you want is a DWORD?

extensions/msal-node-extensions/src/dpapi-addon/dpapi_win.cpp
sameerag: The data here (info[1]) is the user data that we are passing in to the windows API `CryptProtectData()`. My theory is that node's `ArrayBufferView()` implementation - used in the internal implementation of `node::Buffer:Data()`, should be adding the null terminator but I asked @sangonzal to run some experiments confirming that. If not, we may need to add the null termination ourselves.

extensions/msal-node-extensions/src/dpapi-addon/dpapi_win.cpp
sangonzal: The Node addon needs to be loaded on Mac and Linux as well, not just Windows, so we have use to platform agnostic parameters (my understanding is that DWORD is a type only defined for Windows)

jasonnutter (approved):

PR #1831, opened 2020-06-25
[msal-extensions] Extensions #2

sangonzal:
- Add cross platform lock
- Add constant file


extensions/src/lock/CrossPlatformLock.ts
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * Locks cache from read or writes by creating file with same path and name as
+     * cache file but with .lockfile extension. If another process has already created
+     * the lockfile, will retry again based on configuration settings set by CrossPlatformLockOptions
+     */
+    public async lock(): Promise<void> {
+        const processId = pid.toString();
+        for (let tryCount = 0; tryCount < this.retryNumber; tryCount++)
+            try {
+                console.log("Pid " + pid + " trying to acquire lock");
```
sangonzal, line 38: Every console.log statement will be switched over to using the msal-common logger in upcoming PR

extensions/src/error/PersistenceError.ts
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+/**
+ * Error thrown when trying to write MSAL cache to persistence.
+ */
+export class PersistenceError extends Error {
```
pkanher617, line 9: Any reason you don't extend AuthError.ts here? That way you don't have to do all the constructor code below again.

extensions/src/error/PersistenceError.ts
```diff
@@ -0,0 +1,49 @@
...
+        const errorString = errorMessage ? `${errorCode}: ${errorMessage}` : errorCode;
+        super(errorString);
+        Object.setPrototypeOf(this, PersistenceError.prototype);
+
+        this.errorCode = errorCode;
+        this.errorMessage = errorMessage;
+        this.name = "PersistenceError";
+    }
+
+    static createFileSystemError(errorCode: string, errorMessage: string): PersistenceError {
```
pkanher617, line 26: nitpick but can we add comments to the errors here?

pkanher617:
Depending on the timeline, would like to see some form of unit testing.

extensions/src/lock/CrossPlatformLock.ts
```diff
@@ -0,0 +1,80 @@
...
+                const openPromise = promisify(open);
```
jasonnutter, line 39: `fs` has its own built in promisified API, `import { promises as fs } from 'fs'`

extensions/src/error/PersistenceError.ts
sangonzal: Sure, will add.

extensions/src/lock/CrossPlatformLock.ts
sangonzal: Ah good to know, so much better. Will update.

extensions/src/error/PersistenceError.ts
sangonzal: Interesting, had not considered - these errors are not auth errors though, they're instead solely related to persistence. I think it could be valuable to make that distinction.

sangonzal: It's a good point, hadn't considered it. I think since these are not really auth errors (they are all related ), it's good to make the differentiation.

sameerag: I think different `error` class makes sense here.

pkanher617: I guess I agree that a different error class makes sense here, but AuthError is meant to be a base class for all MSAL errors, not just auth based ones (I agree the name is misleading). Any way we can merge it so as not to have duplicated code? Things like setPrototypeOf should preferably not be duplicated across multiple files.

sangonzal: Will make this change, as well as other improvements to errors, in follow up PR.

PR #1832, opened 2020-06-25
[msal-extensions] Extensions #3

sangonzal:
- Add persistence on Linux, Windows, and Mac
- Add persistence plugin, which can be set on MSAL


extensions/src/persistence/FilePersistence.ts
```diff
@@ -0,0 +1,116 @@
...
+        await closePromise(await openPromise(this.filePath, "a"))
+    }
+
+    private async createFileDirectory(): Promise<void> {
+        try {
+            const mkdirPromise = promisify(mkdir);
+            await mkdirPromise(dirname(this.filePath), {recursive: true});
+        } catch (err) {
+            if (err.code == Constants.EEXIST_ERROR) {
+                console.log(`Directory ${dirname(this.filePath)} " already exists"`);
```
sangonzal, line 110: every console.log statement will be refactored to use the msal-common logger in an upcoming PR

pkanher617:
Same comment about unit tests

extensions/src/persistence/PersistenceCachePlugin.ts
```diff
@@ -0,0 +1,90 @@
...
+                console.log("Reload necessary.  Last sync time: " + this.lastSync);
+                this.crossPlatformLock = new CrossPlatformLock(this.lockFilePath, this.lockOptions);
+                await this.crossPlatformLock.lock();
+
+                this.currentCache = await this.persistence.load();
+                this.lastSync = new Date().getTime();
+                console.log("Last sync time updated to: ", this.lastSync);
+            } finally {
+                await this.crossPlatformLock.unlock();
+                delete this.crossPlatformLock;
```
jasonnutter, line 59: I don't think this is the best use of `delete`. Setting to `null` would be better, as it wouldn't actually remove the property.

sangonzal: Makes sense. Thinking about it futher, I think the lock object can be reused, so completely removed this.

extensions/msal-node-extensions/src/lock/CrossPlatformLock.ts
```diff
@@ -57,7 +57,7 @@ export class CrossPlatformLock {
         try {
             // delete lock file
```
sameerag, line 58: Should the comment change? As you are `unlinking` and not `deleting`?

sangonzal: We are deleting the file. https://nodejs.org/api/fs.html#fs_fs_unlink_path_callback

PR #1833, opened 2020-06-25
[msal-extensions] Extensions #4

sangonzal:
- Adds extensions sample


samples/msal-node-extensions/index.js
```diff
@@ -0,0 +1,67 @@
...
+                    redirectUri: "http://localhost:3000/redirect",
+                };
+
+                // get url to sign user in and consent to scopes needed for application
+                pca.getAuthCodeUrl(authCodeUrlParameters).then((response) => {
+                    res.redirect(response);
+                }).catch((error) => console.log(JSON.stringify(error)));
+            });
+
+            app.get('/redirect', (req, res) => {
```
sameerag, line 46: We should upgrade the sample to use `async`, `await`, Can be done later.

samples/msal-node-extensions/index.js
```diff
@@ -0,0 +1,67 @@
...
+            app.get('/', (req, res) => {
+                const authCodeUrlParameters = {
+                    scopes: ["openid"],
+                    redirectUri: "http://localhost:3000/redirect",
+                };
+
+                // get url to sign user in and consent to scopes needed for application
+                pca.getAuthCodeUrl(authCodeUrlParameters).then((response) => {
+                    res.redirect(response);
+                }).catch((error) => console.log(JSON.stringify(error)));
```
jasonnutter, line 43: I would send back the error in the response, in addition to printing it to the console.

PR #1835, opened 2020-06-25
[msal-extensions] Extensions #5

sangonzal:
- Add logger
- Update README and .gitignore


extensions/src/persistence/FilePersistence.ts
```diff
@@ -9,14 +9,20 @@ import { dirname } from "path";
...
+import { Logger, LoggerOptions, LogLevel } from "@azure/msal-common";
 
 export class FilePersistence implements IPersistence {
 
     private filePath: string;
+    private logger: Logger;
 
-    public static async create(fileLocation: string): Promise<FilePersistence> {
+    private constructor() {
+    }
```
jasonnutter, line 20: Not needed if not used.

extensions/README.md
```diff
@@ -16,17 +16,15 @@ Linux - LibSecret is used for encryption.
 
 ## Building
 
-The extensions contain prebuild binaries. To build from source, you will need Python on you path.
+The extensions contain prebuild binaries. To build from source, you will need Python on you path,
+as [node-gyp](https://github.com/nodejs/node-gyp) is used to build addons for accessing system APIs.
+
+On linux, the library uses `libsecret` so you may need to install it.
```
jasonnutter, line 22: I would include or link to instructions for how to check that it is installed and installed it if not.

extensions/src/persistence/FilePersistence.ts
sangonzal: Removed

extensions/README.md
sangonzal: Added instructions for a couple of distributions.

sameerag:
One general comment: Dependency on `msal-common` seems to be hardcoded. We need to have a strategy regarding updating `msal-common` across libraries.

PR #1841, opened 2020-06-26
[msal-browser] Beta.3 Samples Update 

pkanher617:
Updates sample folders for msal-browser samples in VanillaJSTestApp2.0.


samples/VanillaJSTestApp2.0/app/onPageLoad/ui.js
```diff
@@ -9,7 +9,8 @@ const profileDiv = document.getElementById("profile-div");
 function showWelcomeMessage(account) {
     // Reconfiguring DOM elements
     cardDiv.style.display = 'initial';
-    welcomeDiv.innerHTML = `Welcome ${account.name}`;
+    username = account.username;
```
tnorling, line 12: Can we move this to where we handle the response in `auth.js` to make it more clear how we identify the user? I'm afraid this will get looked over and people will be confused how to call `getAccountByUsername`

pkanher617: done.

pkanher617:
> One comment but otherwise LGTM. As a more general note, less related to this specific PR, how do we feel about adding top level folders in `/samples` to denote what library they are applicable to? e.g. `msal-browser`, `msal-node`, etc.

+1 for adding package folders to samples, would make it easier to parse through when developing.

PR #1849, opened 2020-06-29
[msal-extensions] Extensions #6

sangonzal:
- Add tests


extensions/msal-node-extensions/src/lock/CrossPlatformLock.ts
```diff
@@ -32,18 +32,17 @@ export class CrossPlatformLock {
...
+        for (let tryCount = 0; tryCount < this.retryNumber; tryCount++) {
             try {
                 this.logger.info(`Pid ${pid} trying to acquire lock`);
                 this.lockFileHandle = await fs.open(this.lockFilePath, "wx+");
 
                 this.logger.info(`Pid ${pid} acquired lock`);
-                await fs.write(this.lockFileHandle, processId);
-                break;
+                await this.lockFileHandle.write(pid.toString());
+                return;
```
jasonnutter, line 45: Nitpick: `return await this.lockFileHandle.write(pid.toString());`

sangonzal: fs.write returns the bytes written (in this case, pid.toString()). We don't want to return this from lock.lock(), so that's why return needs to be in it's own line.

PR #1852, opened 2020-06-29
[msal-browser] Fix encoding issues

pkanher617:
This PR fixes a couple issues
- Cookies were not encoded before being saved, which resulted in cookie key:value errors when setting b64 values with trailing `=` characters. 
- State was being double encoded in some cases on the server-side, adding an additional `decodeURIComponent` should fix those issues.

### DIFF
--- a/lib/msal-browser/src/cache/BrowserStorage.ts
+++ b/lib/msal-browser/src/cache/BrowserStorage.ts
@@ -215,7 +215,7 @@ export class BrowserStorage extends CacheManager {
         let key: string;
         for (key in this.windowStorage) {
             // Check if key contains msal prefix; For now, we are clearing all the cache items created by MSAL.js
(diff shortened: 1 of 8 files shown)


tnorling:
> Looks good, but if state is (arbitrarily?) being double-encoded on the backend... shouldn't we look to get that fixed rather than double decode as well? Seems odd. What if they end up triple encoding something someday?

Agree with this. Should we maybe check for a certain property to know we're done decoding?

jasonnutter:
> Looks good, but if state is (arbitrarily?) being double-encoded on the backend... shouldn't we look to get that fixed rather than double decode as well? Seems odd. What if they end up triple encoding something someday?

Yes, we should get this fixed on the server, but for now, I think it is fine to assume things will at most be encoded twice.

PR #1855, opened 2020-06-29
Update sample folder structure

pkanher617:
This PR is based off of feedback received in #1841 that samples are becoming numerous. Now we have parent folders for each package that will have the respective samples in that folder.


pkanher617:
> Did you confirm lerna still works correctly?

Had to update lerna.json, thanks for catching this.

PR #1860, opened 2020-06-30
[msal-common] Fix expiration calculation for AuthenticationResult object

pkanher617:
This PR fixes an issue where expiration wasn't being calculated correctly.

### DIFF
--- a/lib/msal-common/src/response/ResponseHandler.ts
+++ b/lib/msal-common/src/response/ResponseHandler.ts
@@ -128,8 +128,8 @@ export class ResponseHandler {
             idTokenClaims: idTokenObj.claims,
             accessToken: serverTokenResponse.access_token,
             fromCache: true,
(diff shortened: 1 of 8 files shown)


jasonnutter:
Can you update the destination branch so that this diff is easier to digest?

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -218,4 +184,50 @@ export class ResponseHandler {
...
         return new CacheRecord(cachedAccount, cachedIdToken, cachedAccessToken, cachedRefreshToken);
     }
+
+    /**
+     * Generate Account
+     * @param serverTokenResponse
+     * @param idToken
+     * @param authority
+     */
+    private generateAccountEntity(serverTokenResponse: ServerAuthorizationTokenResponse, idToken: IdToken, authority: Authority): AccountEntity {
```
pkanher617, line 194: Moved without changing

PR #1861, opened 2020-06-30
Update redirectUri and postLogoutRedirectUri

tnorling:
- `redirectUri` and `postLogoutRedirectUri` must be type `string`
- `redirectUri` is optional on request object. If not provided, will take `pca.config.auth.redirectUri` or `window.location.href`


lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -684,6 +660,7 @@ export class PublicClientApplication {
 
         const authCodeRequest: AuthorizationCodeRequest = {
             ...request,
+            redirectUri: request.redirectUri,
```
jasonnutter, line 663: Should this default to the one set in the config?

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -481,19 +481,7 @@ export class PublicClientApplication {
      *
      */
     public getRedirectUri(): string {
```
jasonnutter, line 483: You could also augment this function to take in the redirect uri from a request:

`return requestRedirectUri || this.config.auth.redirectUri || BrowserUtils.getCurrentUri();`

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -503,19 +491,7 @@ export class PublicClientApplication {
...
-                    throw BrowserConfigurationAuthError.createPostLogoutRedirectUriEmptyError();
-                }
-                return postLogoutUri;
-            }
-
-            return this.config.auth.postLogoutRedirectUri;
-        }
-        // This should never throw unless window.location.href is returning empty.
-        throw BrowserConfigurationAuthError.createPostLogoutRedirectUriEmptyError();
+        return this.config.auth.postLogoutRedirectUri ? this.config.auth.postLogoutRedirectUri : BrowserUtils.getCurrentUri();
```
jasonnutter, line 494: These can just be `||`: `return this.config.auth.postLogoutRedirectUri || BrowserUtils.getCurrentUti();`

lib/msal-browser/src/app/PublicClientApplication.ts
tnorling: By the time it get's here it has already been set from `request || config || current window`
This is only here because the compiler was complaining it's an optional parameter on the `request` object but required on `AuthorizationCodeRequest`. Not sure if there's a better way to handle this.

lib/msal-browser/src/app/PublicClientApplication.ts
tnorling: Done

lib/msal-browser/src/app/PublicClientApplication.ts
tnorling: Done

lib/msal-browser/src/app/PublicClientApplication.ts
jasonnutter: Ah, makes sense.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -503,19 +491,7 @@ export class PublicClientApplication {
...
-                    throw BrowserConfigurationAuthError.createPostLogoutRedirectUriEmptyError();
-                }
-                return postLogoutUri;
-            }
-
-            return this.config.auth.postLogoutRedirectUri;
-        }
-        // This should never throw unless window.location.href is returning empty.
-        throw BrowserConfigurationAuthError.createPostLogoutRedirectUriEmptyError();
+        return this.config.auth.postLogoutRedirectUri || BrowserUtils.getCurrentUri();
```
pkanher617, line 494: Make the same change as redirectUri function above.

tnorling: Done

PR #1862, opened 2020-06-30
Fix redirection to pages with custom hashes or query params

tnorling:
- Adds `redirectStartPage` to request object, enabling developers to specify a page other than `redirectUri` that they would like to end on after a `loginRedirect` or an `acquireTokenRedirect`
- Fixes redirection/hash processing logic to enable returning to pages with custom hashes or query params


lib/msal-common/src/request/AuthorizationUrlRequest.ts
```diff
@@ -18,6 +18,13 @@ export type AuthorizationUrlRequest = BaseAuthRequest & {
      */
     redirectUri: string;
 
+    /**
+     * The page that should be returned to after loginRedirect or acquireTokenRedirect. This should only be used
+     * if this is different from the redirectUri and will default to the page that initiates the request.
+     * When the navigateToLoginRequestUrl config option is set to false this parameter will be ignored.
+     */
+    redirectStartPage?: string;
```
pkanher617, line 26: Because we are adding this parameter, and redirectURI is optional in msal-browser but required in msal-node, it may be better to have a wrapped request object that exists in `msal-browser` and use that for all calls:
```javascript
export type BrowserAuthorizationUrlRequest = AuthorizationUrlRequest & {
    redirectUri?: string;
    redirectStartPage?: string;
}
```

lib/msal-common/src/request/AuthorizationCodeRequest.ts
```diff
@@ -34,4 +34,5 @@ export type AuthorizationCodeRequest = BaseAuthRequest & {
     code: string;
     codeVerifier?: string;
     correlationId?: string;
+    redirectStartPage?: string;
```
pkanher617, line 37: Why is this needed in this request obj?

lib/msal-common/src/url/UrlString.ts
```diff
@@ -86,6 +86,10 @@ export class UrlString {
         return this.urlString;
     }
 
+    static removeHashFromUrl(url: string): string {
+        return url.split("#")[0];
```
pkanher617, line 90: this can be 
`return url.split("#").shift();`

lib/msal-browser/src/utils/BrowserUtils.ts
```diff
@@ -34,6 +34,16 @@ export class BrowserUtils {
         window.location.hash = "";
     }
 
+    /**
+     * Replaces current hash with hash from provided url
+     */
+    static replaceHash(url: string): void {
+        const urlParts = url.split("#");
```
pkanher617, line 41: You can use `UrlString.removeHash()` here

tnorling: I'm trying to retrieve the right side of the split here. `UrlString.removeHashFromUrl()` returns the left side

pkanher617: nvm, looks fine.

lib/msal-common/src/request/AuthorizationCodeRequest.ts
tnorling: Was needed because this was the request object being used in `RedirectHandler.initiateAuthRequest()` where the start page is cached. I just moved the caching back one level to `acquireTokenRedirect` so this is no longer needed

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -164,16 +164,12 @@ export class PublicClientApplication {
...
-            } else {
-                // Loaded page with no valid hash - pass in the value retrieved from cache, or null/empty string
-                return this.handleHash(cachedHash);
-            }
+
+        const currentUrlNormalized = UrlString.removeHashFromUrl(window.location.href);
+        const loginRequestUrlNormalized = UrlString.removeHashFromUrl(loginRequestUrl || "");
+        if (loginRequestUrlNormalized === currentUrlNormalized) {
+            this.config.auth.navigateToLoginRequestUrl ? BrowserUtils.replaceHash(loginRequestUrl) : BrowserUtils.clearHash();
+            return isResponseHash ? this.handleHash(hash) : this.handleHash(`${cachedHash}`);
```
jasonnutter, line 172: Nitpick: `return this.handleHash(isResponseHash ? hash : cachedHash);`

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -164,16 +164,12 @@ export class PublicClientApplication {
...
-                return this.handleHash(hash);
-            } else {
-                // Loaded page with no valid hash - pass in the value retrieved from cache, or null/empty string
-                return this.handleHash(cachedHash);
-            }
+
+        const currentUrlNormalized = UrlString.removeHashFromUrl(window.location.href);
+        const loginRequestUrlNormalized = UrlString.removeHashFromUrl(loginRequestUrl || "");
+        if (loginRequestUrlNormalized === currentUrlNormalized) {
+            this.config.auth.navigateToLoginRequestUrl ? BrowserUtils.replaceHash(loginRequestUrl) : BrowserUtils.clearHash();
```
jasonnutter, line 171: I would prefer to see `if/else` instead of bare ternary.

PR #1863, opened 2020-07-01
Remove deprecated handleRedirectCallback

tnorling:


lib/msal-browser/test/app/PublicClientApplication.spec.ts
```diff
@@ -87,19 +76,6 @@ describe("PublicClientApplication.ts Class Unit Tests", () => {
     describe("Redirect Flow Unit tests", () => {
 
         describe("handleRedirectCallback()", () => {
```
pkanher617, line 78: Rename this to handleRedirectPromise, and make sure all tests in this suite are now using `handleRedirectPromise()` instead of `handleRedirectCallback()`

lib/msal-browser/test/app/PublicClientApplication.spec.ts
```diff
@@ -87,19 +76,6 @@ describe("PublicClientApplication.ts Class Unit Tests", () => {
...
-
-            it("throws error if callback is not valid", async () => {
-                await expect(pca.handleRedirectCallback(null)).rejectedWith(BrowserConfigurationAuthErrorMessage.invalidCallbackObject.desc);
-                await expect(pca.handleRedirectCallback(null)).rejectedWith(BrowserConfigurationAuthError);
-            });
-
-            it("does nothing if no hash is detected", (done) => {
```
pkanher617, line 96: This test should probably be converted to use `handleRedirectPromise()`

pkanher617: instead of deleting it

lib/msal-browser/test/app/PublicClientApplication.spec.ts
tnorling: Done

lib/msal-browser/test/app/PublicClientApplication.spec.ts
tnorling: Done

PR #1868, opened 2020-07-01
[msal-browser] Update request initialization code

pkanher617:
This PR updates the PublicClientApplication class in msal-browser to do the following:

- Rename initializer methods to be more descriptive
- Update acquireTokenSilent() to initialize the request authority and correlation ID
- Update all APIs to set default scopes

This PR also adds the following behavior that was unfinished:
- Add correlationID to all request flows.


lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -661,11 +689,23 @@ export class PublicClientApplication {
...
-    private generateLogoutRequest(logoutRequest?: EndSessionRequest): EndSessionRequest {
-        return {
-            ...logoutRequest,
-            postLogoutRedirectUri: this.getPostLogoutRedirectUri(logoutRequest ? logoutRequest.postLogoutRedirectUri : "")
+    /**
+     * Initializer for the logout request.
+     * @param logoutRequest 
+     */
+    private initializeLogoutRequest(logoutRequest?: EndSessionRequest): EndSessionRequest {
+        const validLogoutRequest = {
```
DarylThayil, line 697: what if logout request is null, will `...logoutRequest` throw?

DarylThayil: @jasonnutter  cc

DarylThayil: I think it will throw

jmckennon: yeah, just tested it, spread operator can't be called on null

jasonnutter: Spreading `null` should work inside objects (but not arrays).

PR #1871, opened 2020-07-01
Add documentation for Accounts

sameerag:
Docs for Accounts usage:
* `getAccounts()` and `getAccountByUsername()` usage
* Guidance on multiple accounts use cases

Note:
Part of the doc will be replaced with `typedocs` links when `msal-common` typedocs are available. cc @jo-arroyo

### DIFF
--- a/lib/msal-common/docs/Accounts.md
+++ b/lib/msal-common/docs/Accounts.md
@@ -1,8 +1,6 @@
-# Accounts in MSAL JS
+# Accounts in MSAL Javascript
 
-MSAL JS libraries (`msal-browser` and `msal-node`) support both single account and multiple accounts scenarios in java script applications. An `account` object is standardized across the libraries:
-
-// TODO replace this with typedocs link when available
+MSAL Javascript libraries (`msal-browser` and `msal-node`) support both single account and multiple accounts scenarios in java script applications. An `account` object is standardized across the libraries:
 
 ```javascript
 export type AccountInfo = {
@@ -25,7 +23,7 @@ Sample usage is as below:
 function handleResponse(resp) {
     if (resp !== null) {
         username = resp.account.username;
-        showWelcomeMessage(resp.account);
+        ...
     } else {
         // need to call getAccount here?
         const currentAccounts = myMSALObj.getAllAccounts();
@@ -34,8 +32,9 @@ function handleResponse(resp) {
         } else if (currentAccounts.length > 1) {
             // Add choose account code here
         } else if (currentAccounts.length === 1) {
+            // Single Account usecase
             username = currentAccounts[0].username;
-            showWelcomeMessage(currentAccounts[0]);
+            ...
         }
     }
 }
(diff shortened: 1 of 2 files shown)


lib/msal-common/docs/Accounts.md
```diff
@@ -0,0 +1,63 @@
+# Accounts in MSAL JS
+
+MSAL JS libraries (`msal-browser` and `msal-node`) support both single account and multiple accounts scenarios in java script applications. An `account` object is standardized across the libraries:
```
DarylThayil, line 3: javascript one word.

lib/msal-common/docs/Accounts.md
```diff
@@ -0,0 +1,63 @@
...
+
+* We provide a public API `getAllAccounts()` which lists all the accounts currently in the cache. An application must choose an account to acquire tokens silently.
+* `msal-browser` provides a public API `getAccountByUsername()` for an application to retrieve a specific account based on `username`
+
+Sample usage is as below:
+
+``` javascript
+
+function handleResponse(resp) {
+    if (resp !== null) {
```
DarylThayil, line 26: might want an example that doesn't have `showWelcomeMessage`

lib/msal-common/docs/Accounts.md
```diff
@@ -0,0 +1,63 @@
...
+            console.error(error);
+        }
+    });
+}
+```
+
+## Notes
+
+* The current msal-browser default [sample](../../samples/msal-browser-samples/VanillaJSTestApp2.0) has a working single account scenario.
+* If you have a multiple accounts scenario, please modify the sample (in `handleResponse()`) to list all cached accounts and choose a specific account
```
DarylThayil, line 62: which sample?

sameerag: Updated (added a link to the exact file this change needs to go in the sample)

lib/msal-common/docs/Accounts.md
sameerag: removed all `showWelcomeMessage()` calls

lib/msal-common/docs/Accounts.md
sameerag: modified

lib/msal-common/docs/Accounts.md
```diff
@@ -0,0 +1,62 @@
...
+    if (resp !== null) {
+        username = resp.account.username;
+        ...
+    } else {
+        // need to call getAccount here?
+        const currentAccounts = myMSALObj.getAllAccounts();
+        if (currentAccounts === null) {
+            return;
+        } else if (currentAccounts.length > 1) {
+            // Add choose account code here
```
jmckennon, line 33: Yeah, my only call out that I brought up in the other doc was this here. I think we should provide a baseline multiple accounts case that we can point people to--either here or in a "multiple accounts" sample

pkanher617: Agreed, we will work to add some samples showing multiple account scenarios in future PRs.

PR #1873, opened 2020-07-01
Update msal-browser SilentHandler.monitorIframeForHash to be purely timing-based

jasonnutter:
This PR updates the polling in SilentHandler to be purely timing-based, similar to #1823.

Other changes:
- Renames SilentHandler method to `monitorIframeForHash`
- Renames PopupHandler method to `monitorPopupForHash`
- Removes `urlNavigate` from error message, as it contains PII (will be added back later in `errorPii` log statement).
- Adds tests for both methods.

### DIFF
--- a/lib/msal-browser/src/interaction_handler/SilentHandler.ts
+++ b/lib/msal-browser/src/interaction_handler/SilentHandler.ts
@@ -35,22 +35,22 @@ export class SilentHandler extends InteractionHandler {
 
     /**
      * Monitors an iframe content window until it loads a url with a known hash, or hits a specified timeout.
-     * @param iframeContentWindow 
+     * @param iframe 
      * @param timeout 
-     * @param urlNavigate 
      */
-    monitorFrameForHash(iframe: HTMLIFrameElement, timeout: number, urlNavigate: string): Promise<string> {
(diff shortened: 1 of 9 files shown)


lib/msal-browser/src/interaction_handler/SilentHandler.ts
```diff
@@ -35,22 +35,22 @@ export class SilentHandler extends InteractionHandler {
...
+             * Polling for iframes can be purely timing based,
+             * since we don't need to accout for interaction.
+             */
+            const nowMark = window.performance.now();
+            const timeoutMark = nowMark + timeout;
 
             const intervalId = setInterval(() => {
-                if (contentWindow.closed) {
-                    // Window is closed
-                    this.removeHiddenIframe(iframe);
```
pkanher617, line 51: Why are we no longer removing the iframe?

pkanher617 (approved):
1 question but otherwise looks good

lib/msal-browser/src/interaction_handler/SilentHandler.ts
jasonnutter: Ah, missed that, will fix.

tnorling (approved):

PR #1884, opened 2020-07-02
IE11 Auth Code Compatibility

tnorling:


lib/msal-common/src/utils/StringUtils.ts
```diff
@@ -41,6 +41,10 @@ export class StringUtils {
         return (typeof str === "undefined" || !str || 0 === str.length);
     }
 
+    static startsWith(str: string, search: string): boolean {
+        return str.substring(0, search.length -1) === search;
+    }
+
```
pkanher617, line 47: Tests for this function?

lib/msal-common/src/utils/StringUtils.ts
```diff
@@ -41,6 +41,10 @@ export class StringUtils {
         return (typeof str === "undefined" || !str || 0 === str.length);
     }
 
+    static startsWith(str: string, search: string): boolean {
+        return str.substring(0, search.length -1) === search;
```
jasonnutter, line 45: I would use `indexOf` and `length - 1`

lib/msal-common/src/url/UrlString.ts
```diff
@@ -39,7 +39,7 @@ export class UrlString {
             url = url.toLowerCase();
         }
 
-        if (url && !url.endsWith("/")) {
+        if (url && (url.charAt(url.length - 1) !== "/")) {
```
jasonnutter, line 42: Can we add an `endsWith` string util similar to startsWith?

tnorling: Done

lib/msal-common/src/utils/StringUtils.ts
tnorling: Done

lib/msal-common/src/utils/StringUtils.ts
tnorling: Done

PR #1885, opened 2020-07-02
[msal-browser] Move authority initialization from constructor

pkanher617:
This PR updates the constructor to remove async authority creation and have it occur only on the first API call.


lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -492,6 +471,37 @@ export class PublicClientApplication implements IPublicClientApplication {
...
+
+    /**
+     * Use to get the post logout redirect uri configured in MSAL or null.
+     * Evaluates postLogoutredirectUri if its a function, otherwise simply returns its value.
+     *
+     * @returns {string} post logout redirect URL
+     */
+    private getPostLogoutRedirectUri(requestPostLogoutRedirectUri?: string): string {
+        return requestPostLogoutRedirectUri || this.config.auth.postLogoutRedirectUri || BrowserUtils.getCurrentUri();
+    }
```
pkanher617, line 493: Moved without change

PR #1887, opened 2020-07-02
Add button to call graph in MSAL Node silent sample

jasonnutter:
Adds a button to call MS graph with the token acquired silently.

Annotation 2020-07-02 161833

Annotation 2020-07-02 161108

### DIFF
--- a/samples/msal-node-samples/msal-node-silent-flow/index.js
+++ b/samples/msal-node-samples/msal-node-silent-flow/index.js
@@ -6,6 +6,7 @@ const express = require("express");
 const handlebars = require('express-handlebars');
 const msal = require('@azure/msal-node');
 const { promises: fs } = require("fs");
+const axios = require("axios");
 
 const SERVER_PORT = process.env.PORT || 3000;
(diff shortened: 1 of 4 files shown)


samples/msal-node-samples/msal-node-silent-flow/index.js
```diff
@@ -144,6 +147,21 @@ app.get('/silentFlow', (req, res) => {
         });
 });
 
+// Calls MS Graph with access token
+app.get('/me', (req, res) => {
+    axios.default.get("https://graph.microsoft.com/v1.0/me", {
+        headers: {
+            Authorization: `Bearer ${accessToken}`
```
jasonnutter, line 154: I suppose there may be a better way than just a local variable

jasonnutter: Ah yeah acquireTokenSilent

technical-boy:
We've adapted most of this into #1888 with some minor changes, thanks!

PR #1888, opened 2020-07-02
[msal-node] Update silent flow sample UI

technical-boy:
This PR:

- Adds a user interface to the `msal-node-silent-flow` sample application
- Adds and uses `express-handlebars` to make the sample's UI dynamic
- Adds a Microsoft Graph request in the form of a "See profile" button and flow for parity with other samples
- Renames the `data/cache.json` file to `data/example.cache.json` and updates instructions in `index.js` comments to make it clear that the developer needs to provide their own valid cache file for the sample to work.


pkanher617:
Can we remove the `msal-node` from all the folder names in the `msal-node-samples` folder? It can look like this now:
- `msal-node-samples`
    - `silent-flow`
    - `auth-code`
    - `device-code`
    - `refresh-token`

samples/msal-node-samples/msal-node-silent-flow/index.js
```diff
@@ -34,35 +50,99 @@ const publicClientConfig = {
...
     },
 };
+
+/** Request Configuration */
+
+const scopes = ["user.read"];
+
+const authCodeUrlParameters = {
+    scopes: scopes,
+    redirectUri: ["http://localhost:3000/redirect"],
```
sangonzal, line 60: This should be a string, not an array.

technical-boy: Fixed

PR #1907, opened 2020-07-07
[2.0] msal 2.0 supports throttling

jmckennon:
This PR implements the primary portion of client-side throttling in `msal-common`. There will be a few accompanying PRs associated with tests and `msal-browser`.


lib/msal-common/src/network/ThrottlingManager.ts
```diff
@@ -0,0 +1,95 @@
...
+    public async sendPostRequest<T>(thumbprint: RequestThumbprint, tokenEndpoint: string, options: NetworkRequestOptions): Promise<NetworkResponse<T>> {
+
+        this.preProcess(thumbprint);
+        const response = await this.networkClient.sendPostRequestAsync<T>(tokenEndpoint, options);
+        this.postProcess(thumbprint, response);
+
+        return response;
+    }
+
+    public preProcess(thumbprint: RequestThumbprint): void {
```
jmckennon, line 38: Ideally, we could make `preProcess` and `postProcess` static to allow other libraries to consume them without passing arround a `ThrottlingManager`, but I couldn't figure out a good way of accomplishing that as we also need to call the abstract functions here. Any thoughts?

lib/msal-common/src/network/ThrottlingManager.ts
```diff
@@ -0,0 +1,95 @@
...
+import { RequestThumbprintValue } from "./RequestThumbprintValue";
+import { AuthError } from "../error/AuthError";
+
+export type NetworkResponse<T> = {
+    headers: Map<string, string>;
+    body: T;
+    status: number;
+};
+
+export abstract class ThrottlingManager {
```
pkanher617, line 19: Not sure I like calling this `ThrottlingManager`, would rather keep it as `NetworkManager`. IMO throttling is an implementation detail of our network class.

lib/msal-common/src/network/ThrottlingManager.ts
pkanher617: I don't think you need to make this static. You could make them protected.

lib/msal-common/src/network/ThrottlingManager.ts
tnorling: +1 I think `NetworkManager` should stay as is and call the throttling functions when needed

lib/msal-common/src/network/ThrottlingManager.ts
tnorling: I'm still unclear on why this needs to be abstract. Each library should provide their own `cacheManager` which you can use without needing each library to implement their own `ThrottlingManager`.

lib/msal-common/src/network/ThrottlingManager.ts
jmckennon: @tnorling what do you consider "throttling functions"? the ones that are currently abstract?

lib/msal-common/src/network/RequestThumbprintValue.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    constructor(throttleTime: number, error?: string, errorCodes?: Array<string>, errorMessage?: string, subError?: string) {
+        this.throttleTime = throttleTime;
+        this.error = error;
+        this.errorCodes = errorCodes;
+        this.errorMessage = errorMessage;
+        this.subError = subError;
+    }
+
+    public generateStorageValue(): string {
```
tnorling, line 25: This function should be part of your Throttling class and just define a type in this file

lib/msal-common/src/network/RequestThumbprint.ts
```diff
@@ -0,0 +1,24 @@
...
+    homeAccountIdentifier?: string;
+
+    constructor(clientId: string, authority: string, scopes: Array<string>, homeAccountIdentifier?: string) {
+        this.clientId = clientId;
+        this.authority = authority;
+        this.scopes = scopes;
+        this.homeAccountIdentifier = homeAccountIdentifier;
+    }
+
+    public generateStorageKey(): string {
```
tnorling, line 21: Same comment here

lib/msal-common/src/network/ThrottlingManager.ts
tnorling: Anything that is for the purpose of throttling, e.g. `preProcess`, `postProcess`, `getItem`, `setItem`,etc.

lib/msal-common/src/network/ThrottlingManager.ts
pkanher617: Since they are public or protected sub-classes can implement them and overwrite functionality

lib/msal-common/src/network/ThrottlingManager.ts
jmckennon: So then the point of `NetworkManager` existing at all would be exclusively to hook into pre/post API call? Because at that point, the only method in the class is `sendPostRequest`. I kind of feel like it can be combined with what's in `INetworkModule` at that point, if I'm understanding correctly

lib/msal-common/src/network/ThrottlingManager.ts
jmckennon: I think what Thomas is saying is that we don't necessarily need this to be abstract if we just directly use the provided cacheManager instead of forcing each library to implement it's own storage relative to throttling. cacheManager should be generic. Node just isn't set up to do that today, it would need to be incremental.

lib/msal-common/src/network/ThrottlingManager.ts
tnorling: Yeah, I suppose since the `NetworkManager` isn't implemented today you could just put the calls to `preProcess` and `postProcess` in the library implementations of the `INetworkModule`

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -129,10 +130,16 @@ export class AuthorizationCodeClient extends BaseClient {
      * @param request
      */
     private async executeTokenRequest(authority: Authority, request: AuthorizationCodeRequest): Promise<NetworkResponse<ServerAuthorizationTokenResponse>> {
+        const thumbprint = new RequestThumbprint(
+            this.config.authOptions.clientId,
+            request.authority,
```
pkanher617, line 135: should use the authority passed into the function instead of the request.authority here.

pkanher617: `authority.canonicalAuthority`

lib/msal-common/src/client/BaseClient.ts
```diff
@@ -33,6 +34,9 @@ export abstract class BaseClient {
     // Network Interface
     protected networkClient: INetworkModule;
 
+    // Throttling Interface
+    protected throttlingManager: ThrottlingManager;
```
pkanher617, line 38: change to network manager.

lib/msal-common/src/network/ThrottlingManager.ts
jmckennon: Ended up moving these into a separate file, no longer have any abstract functions and we're directly using `cacheManager`.

lib/msal-common/src/network/ThrottlingManager.ts
jmckennon: I've kept `NetworkManager` separate for now as to not change the INetworkModule interface and to keep the implementation in common.

lib/msal-common/src/network/RequestThumbprint.ts
jmckennon: Done

lib/msal-common/src/client/AuthorizationCodeClient.ts
jmckennon: Done

lib/msal-common/src/client/BaseClient.ts
jmckennon: Done

lib/msal-common/src/network/RequestThumbprintValue.ts
jmckennon: Done

lib/msal-common/src/network/ThrottlingUtils.ts
```diff
@@ -0,0 +1,83 @@
...
+            const thumbprintValue: RequestThumbprintValue = {
+                throttleTime: ThrottlingUtils.calculateThrottleTime(parseInt(response.headers.get(HeaderNames.RETRY_AFTER))),
+                error: response.body.error,
+                errorCodes: response.body.error_codes,
+                errorMessage: response.body.error_description,
+                subError: response.body.suberror
+            };
+            cacheManager.setItem(
+                ThrottlingUtils.generateThrottlingStorageKey(thumbprint),
+                ThrottlingUtils.generateThrottlingStorageKey(thumbprintValue)
```
tnorling, line 63: Should this just be?
```suggestion
                JSON.stringify(thumbprintValue)
```

lib/msal-common/src/utils/StringUtils.ts
```diff
@@ -78,4 +78,10 @@ export class StringUtils {
             return !StringUtils.isEmpty(entry);
         });
     }
+
+    static jsonParseHelper(json: string) {
+        try {
+            return JSON.parse(json);
+        } catch (e) { }
```
tnorling, line 85: Return null here?

lib/msal-common/src/network/ThrottlingUtils.ts
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    static preProcess(cacheManager: CacheManager, thumbprint: RequestThumbprint): void {
+        const key = ThrottlingUtils.generateThrottlingStorageKey(thumbprint);
+        const storageValue = cacheManager.getItem(key) as string;
+
+        if (storageValue) {
+            const parsedValue = StringUtils.jsonParseHelper(storageValue);
+
+            if (parsedValue.throttleTime >= Date.now()) {
```
tnorling, line 42: To account for possibility of `jsonParseHelper` failing:
```suggestion
            if (parsedValue && parsedValue.throttleTime >= Date.now()) {
```

lib/msal-common/src/utils/Constants.ts
```diff
@@ -34,14 +36,26 @@ export const Constants = {
...
-    CONTENT_TYPE = "Content-Type"
+    CONTENT_TYPE = "Content-Type",
+    RETRY_AFTER = "Retry-After",
+    X_MS_LIB_CAPABILITY = "x-ms-lib-capability"
+}
+
+/**
+ * Header values
+ */
+export enum HeaderValues {
```
tnorling, line 57: Does this need to be an enum? Are there going to be more items under this?

lib/msal-common/src/network/ThrottlingUtils.ts
```diff
@@ -0,0 +1,83 @@
...
+ * Licensed under the MIT License.
+ */
+
+import { NetworkResponse } from "./NetworkManager";
+import { ServerAuthorizationTokenResponse } from "../server/ServerAuthorizationTokenResponse";
+import { Constants, HeaderNames } from "../utils/Constants";
+import { CacheManager } from "../cache/CacheManager";
+import { StringUtils } from "../utils/StringUtils";
+
+export type RequestThumbprint = {
```
tnorling, line 12: These types should each be in their own separate files

jasonnutter: Type only exports end up requiring typescript 3.9 in the client application, so we're trying to avoid that (at least for now).

tnorling: Is that relevant for types that aren't considered publicly facing? We already have many Type only exports across both common and browser

jasonnutter: Ah, that may not be the case for types that aren't part of the main export.

jmckennon: When we say "public" here, do we mean being consumed by browser/node, or actually visible to customers? Because at least browser will need to use these to cover a few cases, I think

lib/msal-common/src/utils/StringUtils.ts
jmckennon: Done

lib/msal-common/src/network/ThrottlingUtils.ts
jmckennon: Done

lib/msal-common/src/utils/Constants.ts
jmckennon: I ended up just putting this with the enum above it, HeaderNames. We can extract separate it for readability if needed

lib/msal-common/src/network/ThrottlingUtils.ts
jmckennon: Done

lib/msal-common/src/network/ThrottlingUtils.ts
jmckennon: I ended up putting these in separate files.

lib/msal-browser/src/cache/BrowserStorage.ts
```diff
@@ -23,7 +23,7 @@ export class BrowserStorage extends CacheManager {
     // Window storage object (either local or sessionStorage)
     private windowStorage: Storage;
     // Client id of application. Used in cache keys to partition cache correctly in the case of multiple instances of MSAL.
-    private clientId: string;
+    public clientId: string;
```
jmckennon, line 26: I don't really like this change, but I needed a way to obtain the `clientId` in the interaction handler class after a redirect and before calling into the `acquireToken` APIs. Any thoughts?

jmckennon: This usage can be seen in the other changes to msal-browser in this PR

jasonnutter: This technically makes it a public setter. I would keep it private then add a public getter function (e.g. getClientId)

jasonnutter:
I think we might want to wait for after GA to merge and release this, thoughts @pkanher617? Should add to appropriate milestone.

lib/msal-browser/src/cache/BrowserStorage.ts
tnorling: You should be able to get it from the cached request in the `RedirectHandler`

lib/msal-common/src/client/BaseClient.ts
```diff
@@ -78,6 +85,7 @@ export abstract class BaseClient {
         headers.set(`${AADServerParamKeys.X_CLIENT_VER}`, this.config.libraryInfo.version);
         headers.set(`${AADServerParamKeys.X_CLIENT_OS}`, this.config.libraryInfo.os);
         headers.set(`${AADServerParamKeys.X_CLIENT_CPU}`, this.config.libraryInfo.cpu);
+        headers.set(HeaderNames.X_MS_LIB_CAPABILITY, HeaderNames.X_MS_LIB_CAPABILITY_VALUE);
```
tnorling, line 88: Why is this added here and in `createDefaultTokenRequestHeaders`?

lib/msal-common/src/network/ThrottlingUtils.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    /**
+     * Performs necessary throttling checks before a network request.
+     * @param cacheManager
+     * @param thumbprint 
+     */
+    static preProcess(cacheManager: CacheManager, thumbprint: RequestThumbprint): void {
+        const key = ThrottlingUtils.generateThrottlingStorageKey(thumbprint);
+        const storageValue = cacheManager.getItem(key) as string;
```
tnorling, line 32: What cacheSchema does this fall under? Temporary? Throttling?

lib/msal-common/src/network/ThrottlingUtils.ts
```diff
@@ -0,0 +1,114 @@
...
+                JSON.stringify(thumbprintValue)
+            );
+        }
+    }
+
+    /**
+     * Checks a NetworkResponse object's status codes against 429 or 5xx
+     * @param response
+     */
+    static checkResponseStatus(response: NetworkResponse<ServerAuthorizationTokenResponse>): boolean {
```
tnorling, line 73: FYI I implemented this function slightly differently in Telemetry and included it in the NetworkManager class. Take a look at that PR and see if that might overlap with this here

lib/msal-common/src/utils/Constants.ts
```diff
@@ -34,14 +36,20 @@ export const Constants = {
     S256_CODE_CHALLENGE_METHOD: "S256",
     URL_FORM_CONTENT_TYPE: "application/x-www-form-urlencoded;charset=utf-8",
     AUTHORIZATION_PENDING: "authorization_pending",
-    NOT_DEFINED: "not_defined"
+    NOT_DEFINED: "not_defined",
+    // Default time to throttle RequestThumbprint in milliseconds
+    DEFAULT_THROTTLE_TIME_SECONDS: 60,
```
tnorling, line 41: Should we add a separate Throttling object in this file for your constants?

lib/msal-common/src/client/BaseClient.ts
jmckennon: This is a new header the server wants to receive for each request to basically tell the server that we support throttling. It's outlined in the original spec.

lib/msal-common/src/network/ThrottlingUtils.ts
jmckennon: Thanks for this, had forgotten about the cache schema stuff, it's temporary. Changing

lib/msal-common/src/client/BaseClient.ts
tnorling: Sure but the question is, why is it being added twice? Once as part of the default library headers and again as part of the default token request headers?

lib/msal-common/src/utils/Constants.ts
jmckennon: Done

lib/msal-browser/src/cache/BrowserStorage.ts
jmckennon: I created a `getClientId()` for now. @tnorling It looks like the cached request is an `AuthorizationCodeRequest` which doesn't contain `clientId`, is there something else?

tnorling: Ah you're right. Thought it was part of `BaseAuthRequest`. You could also pass it in from `PublicClientApplication` since it will be part of the config, but not sure what's cleaner

jmckennon: Going to leave the `getClientId()` for now.

lib/msal-common/src/network/ThrottlingUtils.ts
jmckennon: They're slightly different use cases, though both check against `status`. I'm going to leave this as-is for now.

lib/msal-browser/src/cache/BrowserStorage.ts
```diff
@@ -420,4 +420,11 @@ export class BrowserStorage extends CacheManager {
             throw BrowserAuthError.createTokenRequestCacheError(err);
         }
     }
+
+    /**
+     * Gets the clientId
+     */
+    getClientId(): string {
```
tnorling, line 427: I still don't love this being part of the Storage class since it doesnt really have anything to do with storage. Would rather the clientId get passed into the RedirectHandler if thats the only place this is used

lib/msal-common/src/client/BaseClient.ts
tnorling: Question still stands. This headers is added on Line 72 and on 88

jmckennon: Sorry didn't see the reply here--I assumed this was two different places where headers get created, or does the token request headers just get added to the default library headers?

jmckennon: On second though we probably only need them to be on the token requests anyway, as that's the only time we're looking to throttle a new thumbprint. Will change

jmckennon: Changed, it now takes clientId as a parameter and uses that instead

lib/msal-browser/src/cache/BrowserStorage.ts
jmckennon: Done

lib/msal-browser/src/interaction_handler/RedirectHandler.ts
```diff
@@ -45,7 +45,7 @@ export class RedirectHandler extends InteractionHandler {
      * Handle authorization code response in the window.
      * @param hash
      */
-    async handleCodeResponse(locationHash: string, browserCrypto?: ICrypto): Promise<AuthenticationResult> {
+    async handleCodeResponse(locationHash: string, browserCrypto?: ICrypto, clientId?: string): Promise<AuthenticationResult> {
```
tnorling, line 48: Should this just be a required parameter?

jmckennon: I don't think it can be since handleCodeResponse overrides the handleCodeResponse in InteractionHandler which does not specify clientId as a required parameter

jmckennon: We could technically just rename the function here I think and be able to do that, though

tnorling: Ahh that's right. No worries then

pkanher617 (approved):

PR #1917, opened 2020-07-09
Telemetry #1: Enable telemetry in common

tnorling:
This PR enables sending Telemetry data to eSTS via 2 extra headers on /token calls. Libraries consuming `msal-common` will be responsible for passing an instance of the `TelemetryManager` on each request. The `msal-browser` implementation is in PR #1918

`msal-common`:

- Cache errors


lib/msal-browser/src/interaction_handler/InteractionHandler.ts
```diff
@@ -49,7 +49,7 @@ export abstract class InteractionHandler {
         this.authCodeRequest.code = authCode;
 
         // Acquire token with retrieved code.
-        const tokenResponse = await this.authModule.acquireToken(this.authCodeRequest, cachedNonce, requestState);
+        const tokenResponse = await this.authModule.acquireToken(this.authCodeRequest, undefined, cachedNonce, requestState);
```
tnorling, line 52: This will be replaced with `telemetryManager` in #1918

lib/msal-browser/src/interaction_handler/RedirectHandler.ts
```diff
@@ -68,7 +68,7 @@ export class RedirectHandler extends InteractionHandler {
         this.browserStorage.removeItem(this.browserStorage.generateCacheKey(TemporaryCacheKeys.URL_HASH));
 
         // Acquire token with retrieved code.
-        const tokenResponse = await this.authModule.acquireToken(this.authCodeRequest, cachedNonce, requestState);
+        const tokenResponse = await this.authModule.acquireToken(this.authCodeRequest, undefined, cachedNonce, requestState);
```
tnorling, line 71: This will be replaced with `telemetryManager` in #1918

lib/msal-common/src/telemetry/TelemetryManager.ts
```diff
@@ -0,0 +1,106 @@
...
+ * Licensed under the MIT License.
+ */
+
+import { MSER_TELEM_CONSTANTS, HeaderNames, CacheSchemaType } from "../utils/Constants";
+import { CacheManager } from "../cache/CacheManager";
+import { RequestFailures } from "./RequestFailures";
+import { AuthError } from "../error/AuthError";
+
+export class TelemetryManager {
+    private cacheStorage: CacheManager;
```
tnorling, line 12: Change to `cacheManager`

lib/msal-common/src/telemetry/TelemetryManager.ts
```diff
@@ -0,0 +1,106 @@
...
+            }
+        } else {
+            failures = {
+                requests: [this.apiId, this.correlationId],
+                errors: [error.errorCode],
+                count: 1
+            };
+        }
+
+        this.cacheStorage.setItem(MSER_TELEM_CONSTANTS.FAILURES, JSON.stringify(failures), CacheSchemaType.TEMPORARY);
```
tnorling, line 68: CacheSchemaType.TELEMETRY

lib/msal-common/src/network/NetworkManager.ts
```diff
@@ -2,12 +2,47 @@
...
+     * Wraps sendPostRequestAsync with necessary preflight and postflight logic
+     * @param tokenEndpoint
+     * @param options
+     */
+    async sendPostRequest<T>(tokenEndpoint: string, options: NetworkRequestOptions): Promise<NetworkResponse<T>> {
+        const response = await this.networkClient.sendPostRequestAsync<T>(tokenEndpoint, options);
+
+        if (NetworkManager.serverLoggedRequest(response)) {
+            // Request was logged by server, clear telemetry cache
+            TelemetryManager.clearTelemetryCache(this.cacheManager);
```
tnorling, line 35: Make this a function on cacheManager

lib/msal-common/src/telemetry/TelemetryManager.ts
```diff
@@ -0,0 +1,96 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { SERVER_TELEM_CONSTANTS, HeaderNames, CacheSchemaType } from "../utils/Constants";
+import { CacheManager } from "../cache/CacheManager";
+import { RequestFailures } from "./RequestFailures";
+import { AuthError } from "../error/AuthError";
+
+export class TelemetryManager {
```
jasonnutter, line 11: jasonnutter Member Pending
TelemetryManager is teh name of the class in v1 that handle telemetry that was passed back to client applications, so I think we need to chat and come up with more distinct names for both.

jasonnutter: Decided on `ServerTelemetryManager` and `ClientTelemetryManager`

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -52,14 +53,14 @@ export class AuthorizationCodeClient extends BaseClient {
      * authorization_code_grant
      * @param request
      */
-    async acquireToken(request: AuthorizationCodeRequest, cachedNonce?: string, cachedState?: string): Promise<AuthenticationResult> {
+    async acquireToken(request: AuthorizationCodeRequest, telemetryManager?: TelemetryManager, cachedNonce?: string, cachedState?: string): Promise<AuthenticationResult> {
```
sangonzal, line 56: Instead of passing in TelemetryManager into every acquireToken call, would it might make more sense to include as when you create the AuthorizationCodeClient (by passing into the ClientConfiguration). This applies to all the other objects that extend BaseClient. 

You could also potentially move 

```
if (telemetryManager) {
            headers = telemetryManager.addTelemetryHeaders(headers);
        }
```
to executePostToTokenEndpoint in BaseClient, and remove the duplication

lib/msal-common/src/telemetry/TelemetryManager.ts
```diff
@@ -0,0 +1,96 @@
...
+        this.forceRefresh = forceRefresh || false;
+    }
+
+    // API to add MSER Telemetry to request
+    protected currentRequest(): string {
+        const forceRefreshInt = this.forceRefresh ? 1 : 0;
+        const request = `${this.apiId}${SERVER_TELEM_CONSTANTS.VALUE_SEPARATOR}${forceRefreshInt}`;
+        const platformFields = ""; // TODO: Determine what we want to include
+
+        return [SERVER_TELEM_CONSTANTS.SCHEMA_VERSION, request, platformFields].join(SERVER_TELEM_CONSTANTS.CATEGORY_SEPARATOR);
```
sangonzal, line 30: Should you be checking that the headers are not too large? I think the maximum size is recommended in the spec.

lib/msal-common/src/client/SilentFlowClient.ts
```diff
@@ -71,9 +72,11 @@ export class SilentFlowClient extends BaseClient {
                 refreshToken: cachedRefreshToken.secret
             };
 
-            return refreshTokenClient.acquireToken(refreshTokenRequest);
+            return refreshTokenClient.acquireToken(refreshTokenRequest, telemetryManager);
         }
 
+        // Return tokens from cache
+        telemetryManager.incrementCacheHits();
```
sangonzal, line 79: I believe cache hits should only be incremented for access token hits, not when the tokens have to be refreshed.

lib/msal-common/src/telemetry/TelemetryManager.ts
jasonnutter: Also, can you put these files into a `telemetry/server` folder?

lib/msal-common/src/telemetry/TelemetryManager.ts
tnorling: Instead of checking the header size we're limiting ourselves to the last 3 errors which we're reasonably certain will never cause us to reach the limit

lib/msal-common/src/client/SilentFlowClient.ts
tnorling: That's what this is doing, if the token is refreshed it is returned above this line

lib/msal-common/src/telemetry/TelemetryManager.ts
tnorling: Done

lib/msal-common/src/telemetry/TelemetryManager.ts
sangonzal: Ok that should also work.

lib/msal-common/src/client/SilentFlowClient.ts
sangonzal: Oh I see, thanks for clarifying.

lib/msal-common/src/client/BaseClient.ts
```diff
@@ -88,12 +104,18 @@ export abstract class BaseClient {
...
-        return this.networkClient.sendPostRequestAsync<
+    protected async executePostToTokenEndpoint(tokenEndpoint: string, queryString: string, headers: Map<string, string>): Promise<NetworkResponse<ServerAuthorizationTokenResponse>> {
+        const response = await this.networkManager.sendPostRequest<
         ServerAuthorizationTokenResponse
         >(tokenEndpoint, {
             body: queryString,
             headers: headers,
         });
+
+        if (this.config.serverTelemetryManager && response.status < 500 && response.status !== 429) {
```
pkanher617, line 115: Can we add some comments here?

tnorling: Done

lib/msal-common/test/telemetry/ServerTelemetryManager.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+import { TEST_CONFIG } from "../utils/StringConstants";
+
+let store = {};
+class TestCacheManager extends CacheManager {
+    setItem(key: string, value: string | object, type?: string): void {
+        store[key] = JSON.stringify(value) as string;
+    }
+    getItem(key: string, type?: string): string | object {
+        console.log(key)
+        console.log(store[key])
```
jasonnutter, line 15: Not sure if you meant to include these?

tnorling: Ahh thanks! will remove

PR #1918, opened 2020-07-09
Telemetry #2: Implement Telemetry in msal-browser

tnorling:
This PR builds on PR #1917 to enable sending Telemetry data to eSTS via 2 extra headers on all /token calls

`msal-common`:

- Cache errors


lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -183,11 +185,20 @@ export class PublicClientApplication implements IPublicClientApplication {
             return null;
         }
 
-        // Hash contains known properties - handle and return in callback
-        const currentAuthority = this.browserStorage.getCachedAuthority();
-        const authClient = await this.createAuthCodeClient(currentAuthority);
-        const interactionHandler = new RedirectHandler(authClient, this.browserStorage);
-        return interactionHandler.handleCodeResponse(responseHash, this.browserCrypto);
+        const correlationId = this.browserStorage.getItem(this.browserStorage.generateCacheKey(TemporaryCacheKeys.CORRELATION_ID), CacheSchemaType.TEMPORARY) as string;
```
pkanher617, line 188: This item is already available as part of the cached auth code request, might be easier to get it from the request object?

PR #1919, opened 2020-07-09
[msal-common] Improve Cache Lookup logic

pkanher617:
This PR is an improvement on the cache lookup logic in msal-common for accounts and credentials:
- Account lookup will check for `authorityType` in entity, early return if it is undefined
- Both account and credential lookup do not use boolean `matches` anymore, just early return if checks do not pass.


lib/msal-common/test/cache/CacheManager.spec.ts
```diff
@@ -239,6 +245,29 @@ describe("CacheManager.ts test cases", () => {
         const filterTwo: AccountFilter = { homeAccountId: "Wrong Id" };
         accounts = cacheManager.getAccountsFilteredBy(filterTwo);
         expect(Object.keys(accounts).length).to.eql(0);
+
+        sinon.stub(TrustedAuthority, "getCloudDiscoveryMetadata").callsFake(() => {
```
tnorling, line 249: You can call `ClientTestUtils.setCloudDiscoveryMetadataStubs();` instead of setting this yourself both here and on L-289

lib/msal-common/test/cache/CacheManager.spec.ts
```diff
@@ -239,6 +245,29 @@ describe("CacheManager.ts test cases", () => {
...
+            return {
+                preferred_cache: Constants.DEFAULT_AUTHORITY_HOST, 
+                preferred_network: Constants.DEFAULT_AUTHORITY_HOST, 
+                aliases: [Constants.DEFAULT_AUTHORITY_HOST]}
+        });
+        const filterThree: AccountFilter = { environment: "login.microsoftonline.com" };
```
tnorling, line 255: Can we break each of these tests into a separate test and make the current `it` a `describe`?

lib/msal-common/test/cache/CacheManager.spec.ts
```diff
@@ -256,9 +285,108 @@ describe("CacheManager.ts test cases", () => {
...
+        });
+        const filterThree: CredentialFilter = { environment: "login.microsoftonline.com" };
         credentials = cacheManager.getCredentialsFilteredBy(filterThree);
+        expect(Object.keys(credentials.idTokens).length).to.eql(1);
+        expect(Object.keys(credentials.accessTokens).length).to.eql(2);
+        expect(Object.keys(credentials.refreshTokens).length).to.eql(2);
+        sinon.restore();
+
+        // filter by environment
+        const filterFour: CredentialFilter = { environment: "Wrong Env" };
```
tnorling, line 303: Same comment here about breaking into smaller tests

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -297,50 +298,92 @@ export abstract class CacheManager implements ICacheManager {
...
+
+        return true;
+    }
+
+    /**
+     * Removes all accounts and related tokens from cache.
+     */
+    removeAllAccounts(): boolean {
+        const allCacheKeys = this.getKeys();
+        allCacheKeys.forEach((cacheKey) => {
```
tnorling, line 362: Instead of repeating logic here, could you just call `getAccountsFilteredByInternal()`? Alternatively, we could factor out the logic to a helper function, something like `isAccount`, that you could call in both places?

lib/msal-common/test/cache/CacheManager.spec.ts
pkanher617: done.

lib/msal-common/test/cache/CacheManager.spec.ts
pkanher617: done.

lib/msal-common/test/cache/CacheManager.spec.ts
pkanher617: done.

lib/msal-common/src/cache/CacheManager.ts
pkanher617: Done.

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -297,50 +286,78 @@ export abstract class CacheManager implements ICacheManager {
...
-                matches = matches && this.matchClientId(entity, clientId);
+            if (!StringUtils.isEmpty(clientId) && !this.matchClientId(entity, clientId)) {
+                return;
             }
 
             // idTokens do not have "target", target specific refreshTokens do exist for some types of authentication
             // TODO: Add case for target specific refresh tokens
-            if (!StringUtils.isEmpty(target) && credType === CredentialType.ACCESS_TOKEN) {
-                matches = matches && this.matchTarget(entity, target);
+            if (!StringUtils.isEmpty(target) && !this.matchTarget(entity, target)) {
```
sameerag, line 311: why did we remove `credType === CredentialType.ACCESS_TOKEN` check here? Not adding is harmless but does it makes sense to retain it?

sameerag: Never mind, I see you added the check in match Target. I still think the check should be surfaced here and have `matchTarget()` purely do its name sake operation.

pkanher617: I elected to move it because it was easier to understand the logic. We can revisit this when we add target-specific refresh tokens.

PR #1921, opened 2020-07-09
Telemetry #3: Implement Telemetry in msal-node

tnorling:
This PR builds on PR #1917 to enable sending Telemetry data to eSTS via 2 extra headers on all /token calls

`msal-common`:

- Cache errors


lib/msal-node/src/utils/Constants.ts
```diff
@@ -45,3 +45,11 @@ export const CACHE = {
 export const Constants = {
     MSAL_SKU: 'msal.js.node',
 };
+
+// Telemetry Constants
```
tnorling, line 49: Comment DeviceCode, AuthCode, SilentFlow api codes

lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -83,14 +84,21 @@ export abstract class ClientApplication {
...
         this.logger.info("acquireTokenByCode called");
-        const authClientConfig = await this.buildOauthClientConfiguration(
-            request.authority
-        );
-        this.logger.verbose("Auth client config generated");
-        const authorizationCodeClient = new AuthorizationCodeClient(
-            authClientConfig
-        );
-        return authorizationCodeClient.acquireToken(this.initializeRequestScopes(request) as AuthorizationCodeRequest);
+        const validRequest = this.initializeRequest(request) as AuthorizationCodeRequest;
```
sangonzal, line 87: Might make more sense to pass the TelemetryManager into the `buildOauthClientConfiguration` then to each acquireTokenCall.

lib/msal-node/src/utils/Constants.ts
```diff
@@ -45,3 +45,17 @@ export const CACHE = {
...
 };
+
+/**
+ * API Codes for Telemetry purposes. 
+ * Before adding a new code you must claim it in the MSAL Telemetry tracker as these number spaces are shared across all MSALs
+ * 0-99 Silent Flow
+ * 600-699 Device Code Flow
+ * 800-899 Auth Code Flow
+ */
+export enum ApiId {
```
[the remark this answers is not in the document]
tnorling: Yes

lib/msal-node/src/client/ClientApplication.ts
tnorling: Yeah I thought about that. Are we reasonably certain that we will always create a new client for each request? Or could that change in the future? That was the only reason I thought it was better to add it to the acquireTokenCall

sangonzal: Yea a new client will be created for each request. (It would probably make more sense to name them "requests" instead of "clients", but that's a separate issue)

tnorling: Ended up taking both of these suggestions. Code is much cleaner now

lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -106,14 +116,22 @@ export abstract class ClientApplication {
...
+            const refreshTokenClientConfig = await this.buildOauthClientConfiguration(
+                request.authority,
+                serverTelemetryManager
+            );
+            this.logger.verbose("Auth client config generated");
+            const refreshTokenClient = new RefreshTokenClient(
+                refreshTokenClientConfig
+            );
+            return refreshTokenClient.acquireToken(validRequest);
+        } catch (e) {
```
sameerag, line 131: Are we sure the catch is always telemetry initialization error?

lib/msal-node/src/utils/Constants.ts
```diff
@@ -47,6 +47,20 @@ export const Constants = {
...
+    acquireTokenSilent = 62,
+    acquireTokenByCode = 871,
+    acquireTokenByRefreshToken = 872,
+    acquireTokenByDeviceCode = 671
+};
```
sameerag, line 62: Are they no API ids for confidential clients? Are we adding it later?

sameerag (approved):
two comments, approved.

lib/msal-node/src/client/ClientApplication.ts
tnorling: It's not. The catch is for any error that occurs in the flow so that Telemetry can properly cache it

lib/msal-node/src/utils/Constants.ts
tnorling: Thanks for the callout. Confidential Client wasn't built out yet when this PR was first put up. Will add.

PR #1929, opened 2020-07-10
Extensions azure pipelines

sangonzal:
Add azure pipelines build for the extensions library.


build/npm-build-extensions-steps.yml
```diff
@@ -0,0 +1,26 @@
+# Steps to build and test npm project
+
+steps:
+    - task: NodeTool@0
+      inputs:
+          versionSpec: '10.x'
+      displayName: 'Install Node.js'
+
+    - task: Npm@1
+      displayName: 'npm ci'
```
jasonnutter, line 10: Is this running these commands against the entire repo?

sangonzal: Not anymore. It only runs against the extensions (line 13)

jasonnutter: Ah right missed that.

sangonzal: I just added it today

extensions/msal-node-extensions/package.json
```diff
@@ -21,7 +21,9 @@
   },
   "scripts": {
     "start": "tsdx watch",
-    "build": "tsdx build",
+    "build": "npm run build:common && npm run link:common && tsdx build",
+    "build:common": "cd ../../lib/msal-common && npm i && npm run build && npm link",
+    "link:common": "npm link @azure/msal-common",
```
sameerag, line 26: Doesn't lerna do this? Or is this the `lerna` issue you were referring to?

pkanher617: can't use lerna with extensions since it downloads python right? might be good to revisit this in the future though.

sameerag: Makes sense. Thanks.

PR #1946, opened 2020-07-13
Check if tokens are returned before attempting to create token entity

tnorling:
Fixes a bug where an error is thrown if an access token is not returned by the server, such as a scenario where login is called with an empty request object (we only expect id_token in the response)

### DIFF
--- a/lib/msal-common/src/response/ResponseHandler.ts
+++ b/lib/msal-common/src/response/ResponseHandler.ts
@@ -143,44 +143,53 @@ export class ResponseHandler {
         }
 
         // IdToken
-        const cachedIdToken = IdTokenEntity.createIdTokenEntity(
-            this.homeAccountIdentifier,
-            env,
-            serverTokenResponse.id_token,
-            this.clientId,
-            idTokenObj.claims.tid
-        );
+        let cachedIdToken: IdTokenEntity;
+        if (!StringUtils.isEmpty(serverTokenResponse.id_token)) {
+            cachedIdToken = IdTokenEntity.createIdTokenEntity(
+                this.homeAccountIdentifier,
+                env,
+                serverTokenResponse.id_token,
+                this.clientId,
+                idTokenObj.claims.tid
+            );
+        }
 
         // AccessToken
-        const responseScopes = ScopeSet.fromString(serverTokenResponse.scope);
+        let cachedAccessToken: AccessTokenEntity;
+        if (!StringUtils.isEmpty(serverTokenResponse.access_token)) {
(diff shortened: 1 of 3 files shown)


lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -214,19 +223,32 @@ export class ResponseHandler {
      * @param stateString 
      */
     static generateAuthenticationResult(cacheRecord: CacheRecord, idTokenObj: IdToken, fromTokenCache: boolean, stateString?: string): AuthenticationResult {
-        const responseScopes = ScopeSet.fromString(cacheRecord.accessToken.target);
+        let accessToken: string;
```
pkanher617, line 226: set these to default values (i.e. empty string)

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -143,44 +143,53 @@ export class ResponseHandler {
...
 
         // IdToken
-        const cachedIdToken = IdTokenEntity.createIdTokenEntity(
-            this.homeAccountIdentifier,
-            env,
-            serverTokenResponse.id_token,
-            this.clientId,
-            idTokenObj.claims.tid
-        );
+        let cachedIdToken: IdTokenEntity;
```
pkanher617, line 146: set to default null

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -143,44 +143,53 @@ export class ResponseHandler {
...
-            this.homeAccountIdentifier,
-            env,
-            serverTokenResponse.access_token,
-            this.clientId,
-            idTokenObj.claims.tid,
-            responseScopes.printScopes(),
-            tokenExpirationSeconds,
-            extendedTokenExpirationSeconds
-        );
+        let cachedAccessToken: AccessTokenEntity;
```
pkanher617, line 158: set default null

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -143,44 +143,53 @@ export class ResponseHandler {
...
 
         // refreshToken
-        const cachedRefreshToken = RefreshTokenEntity.createRefreshTokenEntity(
-            this.homeAccountIdentifier,
-            env,
-            serverTokenResponse.refresh_token,
-            this.clientId,
-            serverTokenResponse.foci
-        );
+        let cachedRefreshToken: RefreshTokenEntity;
```
pkanher617, line 183: set default null

jasonnutter:
Agree about tests, otherwise looks good!

PR #1962, opened 2020-07-15
[msal-browser] Remove default scopes from silent flow

pkanher617:
This PR removes the addition of `openid` and `profile` scopes when passing a requests to the SilentFlowClient in msal-common. This prevents cache misses when requesting from non-graph resources.


sameerag:
Is this applicable to `msal-node` too?

pkanher617:
> Is this applicable to `msal-node` too?

Yes I think you do not need to add default scopes to `acquireTokenByRefreshToken` and `acquireTokenSilent` in `ClientApplication.ts`

PR #1964, opened 2020-07-15
[msal-browser] Add SilentRequest.ts

pkanher617:
This PR adds a SilentRequest object that is used for the `acquireTokenSilent()` API. Also fixes the comments in for each of the request objects.


lib/msal-browser/src/app/IPublicClientApplication.ts
```diff
@@ -1,19 +1,21 @@
...
+    acquireTokenSilent(silentRequest: SilentRequest): Promise<AuthenticationResult>;
     getAccountByUsername(userName: string): AccountInfo;
     getAllAccounts(): AccountInfo[];
     handleRedirectPromise(): Promise<AuthenticationResult | null>;
-    loginPopup(request: AuthorizationUrlRequest): Promise<AuthenticationResult>;
-    loginRedirect(request: AuthorizationUrlRequest): Promise<void>;
+    loginPopup(request: PopupRequest): Promise<AuthenticationResult>;
+    loginRedirect(request: RedirectRequest): Promise<void>;
     logout(logoutRequest?: EndSessionRequest): Promise<void>;
     ssoSilent(request: AuthorizationUrlRequest): Promise<AuthenticationResult>;
```
jasonnutter, line 20: Should ssoSilent also get the silent request?

pkanher617: No because it is not doing any cache lookups. I would rather just keep it the request from common for now, if we need to add a specific request for ssoSilent we should do it when it is needed.

jasonnutter: Ah right, makes sense.

PR #1974, opened 2020-07-16
Replace /organizations with tenant ID in replaceTenantPath

### LINKED ISSUE #1583: What's the reason of the change in UrlUtils.replaceTenantPath() in fix #1299
## Library
- [X] `msal@1.3.0`

## Description
I noticed change #1299 updated UrlUtils.replaceTenantPath() as well, and I wonder whether it was intended:

`        -if (tenantId && (pathArray.length !== 0 && (pathArray[0] === Constants.common || pathArray[0] === SSOTypes.ORGANIZATIONS))) {
        +if (tenantId && (pathArray.length !== 0 && pathArray[0] === Constants.common)) {
`
The code is used when caching an incoming token, by tweaking the authority.
So in msal 1.2.2:
- tokens acquired using common or organizations authority are cached with a tenant'ed authority in the key
- tokens acquired using consumers authority are cached as is

Whereas in msal 1.3.0:
- tokens acquired using common authority are cached with a tenant'ed authority in the key
- tokens acquired using organizations or consumers authority are cached as is.

None of these behaviors is particularly consistent, but external code may have subtle dependency on how the things work internally.

So I'd like to understand whether the change is due to the unfortunate use of SSOTypes.ORGANIZATIONS in the old code (even though this has nothing to do with SSO), or there's some other thing/development I'm missing.

Thanks

Cc @jasonnutter

jasonnutter:
Partially reverts changes made in #1299, which stopped replacing `/organizations` with the tenant ID. Also added tests. Addresses #1583

### DIFF
--- a/lib/msal-core/src/utils/UrlUtils.ts
+++ b/lib/msal-core/src/utils/UrlUtils.ts
@@ -115,10 +115,10 @@ export class UrlUtils {
      * @param tenantId The tenant id to replace
      */
     static replaceTenantPath(url: string, tenantId: string): string {
-        url = url.toLowerCase();
-        const urlObject = this.GetUrlComponents(url);
+        const lowerCaseUrl = url.toLowerCase();
+        const urlObject = this.GetUrlComponents(lowerCaseUrl);
         const pathArray = urlObject.PathSegments;
-        if (tenantId && (pathArray.length !== 0 && pathArray[0] === Constants.common)) {
+        if (tenantId && (pathArray.length !== 0 && (pathArray[0] === Constants.common || pathArray[0] === SSOTypes.ORGANIZATIONS))) {
             pathArray[0] = tenantId;
         }
         return this.constructAuthorityUriFromObject(urlObject, pathArray);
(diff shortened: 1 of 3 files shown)


lib/msal-core/src/utils/UrlUtils.ts
```diff
@@ -115,10 +115,10 @@ export class UrlUtils {
      * @param tenantId The tenant id to replace
      */
     static replaceTenantPath(url: string, tenantId: string): string {
-        url = url.toLowerCase();
-        const urlObject = this.GetUrlComponents(url);
+        const lowerCaseUrl = url.toLowerCase();
+        const urlObject = this.GetUrlComponents(lowerCaseUrl);
```
sameerag, line 119: can we not just say: 
``` javascript
const urlObject = this.GetUrlComponents(url.toLowerCase());
```

sameerag (approved):
One minor comment but approved. Looks good.

lib/msal-core/src/utils/UrlUtils.ts
jasonnutter: We can, I personally don't mind the verbosity. Easier to read imo.

tnorling (approved):

PR #1981, opened 2020-07-16
[msal-browser] Add negative offset value prevention for popup windows

### LINKED ISSUE #1928: .loginPop() window position within browser window is cut off in Chrome and Edge.

## Library
- [X ] `@azure/msal-browser@2.0.0-beta.4`

## Framework
vanilla JS

## Description
When calling .loginPopup on Chrome and Edge and the browser window is near the top of the screen, the popup window will have the top of it's frame outside the boundry of the screen (cut off and cannot see the title bar to move/resize). Using Firefox, the whole popup window will appear inside the boundry of browser window and displays correctly.

## Error Message
None
## Security

## Regression
Version:

## MSAL Configuration

```
{
            auth: {
                clientId: 'xxxxxxxxxxxxxxxxx',
                redirectUri: 'http://localhost/'

            },
            cache: {
                storeAuthStateInCookie: false
            }
 }    
```

## Reproduction steps

1. Open the browser. Move the top of the browser to be near the top of your screen. 
2. Trigger an event to cause the .loginpopup() method to be called.
3. Note the position of the popup window. Edge and Chrome will show this window above the browser window and, depending on how close to the top to set the browser, the titlebar will not appear. Doing the same step on Firefox, the popup window will appear within the bounds of the browser window. 

## Expected behavior
The entire popup window should be visible initially.

## Browsers/Environment
- [ X] Chrome-> 83 and 86(canary)
- [ X] Firefox=> 78
- [ X] Edge -> 83 and 86 (edge canary)
- [ ] Safari
- [ ] IE

technical-boy:
This PR prevents PopUp window position offsets from being negative by using `Math.max()` in order to make the top and left offsets 0 if they are negative immediately before they are added to the calculated position.

Fixes #1928

### DIFF
--- a/lib/msal-browser/src/interaction_handler/PopupHandler.ts
+++ b/lib/msal-browser/src/interaction_handler/PopupHandler.ts
@@ -126,8 +126,8 @@ export class PopupHandler extends InteractionHandler {
              */
             const width = window.innerWidth || document.documentElement.clientWidth || document.body.clientWidth;
             const height = window.innerHeight || document.documentElement.clientHeight || document.body.clientHeight;
-            const left = ((width / 2) - (popUpWidth / 2)) + winLeft;
-            const top = ((height / 2) - (popUpHeight / 2)) + winTop;
+            const left = ((width / 2) - (popUpWidth / 2)) + Math.max(0, winLeft);
+            const top = ((height / 2) - (popUpHeight / 2)) + Math.max(0, winTop);
 
             // open the window
             const popupWindow = window.open(urlNavigate, title, "width=" + popUpWidth + ", height=" + popUpHeight + ", top=" + top + ", left=" + left);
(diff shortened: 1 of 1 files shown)


lib/msal-browser/src/interaction_handler/PopupHandler.ts
```diff
@@ -126,8 +126,8 @@ export class PopupHandler extends InteractionHandler {
              */
             const width = window.innerWidth || document.documentElement.clientWidth || document.body.clientWidth;
             const height = window.innerHeight || document.documentElement.clientHeight || document.body.clientHeight;
-            const left = ((width / 2) - (popUpWidth / 2)) + winLeft;
-            const top = ((height / 2) - (popUpHeight / 2)) + winTop;
+            const left = ((width / 2) - (popUpWidth / 2)) + Math.max(0, winLeft);
+            const top = ((height / 2) - (popUpHeight / 2)) + Math.max(0, winTop);
```
jasonnutter, line 130: Hmm, what if `height` is 100 and popupHeight is 150? This would still result in a negative number, right? I think you need to wrap the entire calculate in `Math.max`:

`const top = Math.max(0, ((height / 2) - (popUpHeight / 2)) + winTop)`

technical-boy: Right, let me update this. Sadly I can't repro the issue in either Windows 10 or macOS Catalina.

PR #1992, opened 2020-07-17
Move FAQ from Wiki to repo

tnorling:
Copies content from Wiki to msal-core and msal-browser repos and adds a few more. Also reorganizes content


pkanher617:
Isn't this the same thing as the FAQ? Should we just merge the docs/link to them from the same section in the README?

tnorling:
> Isn't this the same thing as the FAQ? Should we just merge the docs/link to them from the same section in the README?

Yes you're totally right. I had no idea the FAQ section has been kept up to date, thought we were trying to move away from the Wiki to more docs within the repo. Is that still the case? Should we move the content from the Wiki to this doc and then update the wiki to link to the doc?

pkanher617:
> Yes you're totally right. I had no idea the FAQ section has been kept up to date, thought we were trying to move away from the Wiki to more docs within the repo. Is that still the case? Should we move the content from the Wiki to this doc and then update the wiki to link to the doc?

Yes let's move the content from the wiki and link to the doc. The doc should have all of the most recent info.

lib/msal-browser/FAQ.md
```diff
@@ -1,17 +1,158 @@
...
+**[Tokens](#Tokens)**
+
+1. [How do I acquire an access token? How do I use it?](#how-do-i-acquire-an-access-token-how-do-i-use-it)
+1. [How do I renew tokens with MSAL.js?](#how-do-i-renew-tokens-with-msaljs)
+1. [How can I acquire tokens faster?](#how-can-i-acquire-tokens-faster)
+1. [I'm seeing scopes openid, profile, email, offline_access in my tokens, even though I haven't requested them. What are they?](#im-seeing-scopes-openid-profile-email-offline_access-and-userread-in-my-tokens-even-though-i-havent-requested-them-what-are-they)
+1. [How long do tokens last? How long are they valid for?](#how-long-do-tokens-last-how-long-are-they-valid-for)
+1. [What are the differences between supported audiences and account types?](#what-are-the-differences-between-supported-audiences-and-account-types)
+
+**[Scopes & Resources](Scopes-&-Resources)**
```
jo-arroyo, line 36:
```suggestion
**[Scopes & Resources](#Scopes-&-Resources)**
```

lib/msal-browser/FAQ.md
```diff
@@ -1,17 +1,158 @@
...
+## Will MSAL 2.x support B2C?
 
 We are currently working with the B2C service team to allow for authorization code flow to work in the browser with B2C tenants. We hope to have this available shortly after release.
 
-# Common Error Handling 
+# Authentication
+
+## I don't understand the redirect flow. How does the `handleRedirectPromise` function work?
+
+The redirect flow can be confusing, as redirecting away from the page means you are creating a whole new instance of the application when you return. This means that calling a redirect method cannot return anything. Rather what happens is, the page is redirected away, you enter your credentials and you are redirected back to your application with the response in the url hash.
```
jo-arroyo, line 76:
```suggestion
The redirect flow can be confusing, as redirecting away from the page means you are creating a whole new instance of the application when you return. This means that calling a redirect method cannot return anything. Rather, what happens is that the page is redirected away, you enter your credentials, and you are redirected back to your application with the response in the url hash.
```

lib/msal-browser/FAQ.md
```diff
@@ -1,17 +1,158 @@
...
+loginPopup().then((response) => {
+    const uniqueID = response.account.homeAccountId;
+})
+```
+
+# Configuration
+
+## What is the difference between `sessionStorage` and `localStorage`?
+
+We offer two methods of storage for Msal, `localStorage` and `sessionStorage`.  Our recommendation is to use `sessionStorage` because it is more secure in storing tokens that are acquired by your users, but `localStorage` will give you Single Sign On accross tabs and user sessions.  We encourage you to explore the options and make the best decision for your application.
```
jo-arroyo, line 112:
```suggestion
We offer two methods of storage for Msal, `localStorage` and `sessionStorage`.  Our recommendation is to use `sessionStorage` because it is more secure in storing tokens that are acquired by your users, but `localStorage` will give you Single Sign On across tabs and user sessions.  We encourage you to explore the options and make the best decision for your application.
```

lib/msal-browser/FAQ.md
```diff
@@ -1,17 +1,158 @@
...
+
+Please refer to token guide [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/acquire-token.md). 
+
+## How do I renew tokens with MSAL.js?
+
+MSAL.js provides the `acquireTokenSilent` method which handles token renewal by making silent token requests without prompting the user. The method first looks for a valid cached token in the browser storage. If it does not find one, the library makes the silent request to Azure AD and if there is an active user session (determined by a cookie set in browser on the Azure AD domain), a fresh token is returned. The library does not automatically invoke the `acquireTokenSilent` method. It is recommended that you call `acquireTokenSilent` in your app before making an API call to get the valid token.
+
+In certain cases, the `acquireTokenSilent` method's attempt to get the token silently fails. Some examples of this are when there is an expired user session with Azure AD or a password change by the user, etc. which requires user interaction. When the `acquireTokenSilent` fails, you need to call one of the interactive acquire token methods.(`acquireTokenPopup` or `acquireTokenRedirect`).
```
jo-arroyo, line 132:
```suggestion
In certain cases, the `acquireTokenSilent` method's attempt to get the token silently fails. Some examples of this are when there is an expired user session with Azure AD or a password change by the user, etc. which requires user interaction. When the `acquireTokenSilent` fails, you need to call one of the interactive acquire token methods (`acquireTokenPopup` or `acquireTokenRedirect`).
```

lib/msal-browser/FAQ.md
```diff
@@ -1,17 +1,158 @@
...
+## What browsers are supported by MSAL.js?
+
+MSAL.js has been tested with the following browsers:
 
-# Does this library work for iframed applications?
+IE 11, Edge, Chrome, Firefox and Safari
+
+Keep these steps in mind when [using MSAL.js with IE](https://github.com/AzureAD/microsoft-authentication-library-for-js/wiki/Using-msal.js-with-Internet-Explorer).
+
+There are certain known issues and mitigations documented for Safari, IE and Edge. Please check out:
```
sameerag, line 56: nit: Very similar to the sentence above.Either move them to the same line or change the language.

lib/msal-browser/FAQ.md
```diff
@@ -1,17 +1,158 @@
...
 
-# Will MSAL 2.x support B2C?
+## Will MSAL 2.x support B2C?
 
 We are currently working with the B2C service team to allow for authorization code flow to work in the browser with B2C tenants. We hope to have this available shortly after release.
 
-# Common Error Handling 
+# Authentication
+
+## I don't understand the redirect flow. How does the `handleRedirectPromise` function work?
```
sameerag, line 74: nit: Do you want to add a visual representation at some point?

lib/msal-browser/FAQ.md
```diff
@@ -1,17 +1,158 @@
...
+
+## How can I acquire tokens faster?
+
+Please refer to our performance guide [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/performance.md)
+
+## I'm seeing scopes `openid`, `profile`, `email`, `offline_access` and `User.Read` in my tokens, even though I haven't requested them. What are they?
+
+The first four (`openid`, `profile`, `email` and `offline_access`) are called **default scopes**. They are added to Azure AD as part of Azure AD - OAuth 2.0/OpenID Connect compliance. They are **not** part of any particular API. You can read more about them [here](https://openid.net/specs/openid-connect-core-1_0.html).
```
sameerag, line 142: We do not append `email` explicitly in `msal` though right? Do we still get it in the response?

lib/msal-browser/FAQ.md
tnorling: This was just copied over from what was already in the Wiki. Edits can be made later on a case by case basis.

lib/msal-browser/FAQ.md
tnorling: Yes, it's on my list but will take it later.

lib/msal-browser/FAQ.md
tnorling: No we do not. Would have to verify if we get it in the response, I believe it adds a claim to idTokens rather than returning an access token. I think this is referring to the app registration, but as this was copied from the Wiki let's make adjustments as needed.

PR #1997, opened 2020-07-20
[msal-browser] B2C docs

derisen:
#docs: working with b2c

This PR adds a documentation on working with B2C covering app configuration, scopes, permissions types, endpoints, common use cases etc.

p.s. I'm afraid it turned into more of a discussion of the service than the library itself. I'll try to revise if that seems out of place.


lib/msal-browser/docs/working-with-b2c.md
```diff
@@ -0,0 +1,103 @@
...
+## B2C App Configuration
+
+The following is a B2C app configuration example:
+
+```javascript
+const msalConfig = {
+  auth: {
+    clientId: "<your-clientID>",
+    authority: "https://<your-tenant>.b2clogin.com/<your-tenant>.onmicrosoft.com/<your-policyID>",
+    validateAuthority: true,
```
tnorling, line 25: Since this is a msal-browser doc we should remove `validateAuthority` as it is no longer a config option in 2.0

lib/msal-browser/docs/working-with-b2c.md
```diff
@@ -0,0 +1,103 @@
...
+
+**Delegated permissions** specify *scope-based access* using *interactive* authorization from the signed-in user. These permissions are presented to the resource (*e.g.* your web API, MS Graph API etc.) at run-time as `scp` claims in the client's **Access Token**.
+
+A user's B2C authentication cannot be used to authorize to AAD protected apps, or **Microsoft APIs** (which are also protected by AAD). As such, when you use **MSAL.js**, you cannot use the `<your-tenant>.b2clogin.com/.../token` endpoint to obtain a token for **MS Graph API**.
+
+### OpenID Connect Permissions
+
+The exception to the rule above comes from a special set of scopes known as **OpenID Connect** (OIDC) permissions, which includes `openid`, `profile`, `email`. These permissions originate from [Azure AD - OpenID Connect compliance](https://docs.microsoft.com/azure/active-directory/develop/v2-protocols-oidc). Another special permission is the `offline_access`, which gives your app access to a resources on behalf of the user for an extended time (using a **Refresh Token**). MSAL.js will supply `openid`, `profile` and `offline_access` by default during `loginPopup()` and `loginRedirect()` requests.
```
tnorling, line 71: We actually no longer provide `offline_access` by default (and the guidance is not to use offline_access at all as it is a long lived refresh token). For B2C it is required for the user to pass this in until #1999 gets resolved

lib/msal-browser/docs/working-with-b2c.md
derisen: Aah, of course -removed

lib/msal-browser/docs/working-with-b2c.md
derisen: Oh? Ok removed, and added this bit: 

> :information_source: At the moment, `offline_access` must be added to scopes when requesting a token from B2C endpoint. If you're reasonably certain that the token is valid and available in the cache then you may omit `offline_access` in an `acquireTokenSilent()` call in order to retrieve the **Access Token** from the cache. However, when the token must be renewed you must make sure to include `offline_access`.

lib/msal-browser/docs/working-with-b2c.md
```diff
@@ -0,0 +1,104 @@
+# Working with MSAL.js and Azure AD B2C
+
+> :warning: Before you start here, make sure you understand [how to initialize an app object](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/initialization.md) and [working with resources and scopes](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/resources-and-scopes.md). We also recommend general familiarity with **Azure AD B2C**. See the [B2C documentation](https://docs.microsoft.com/azure/active-directory-b2c/technical-overview) for more.
+
+**MSAL.js supports** authentication with **social** (Microsoft, Google, Facebook etc.) and **local** (Azure AD work and school) identities using **Azure AD B2C** (*B2C* for short). When developing B2C apps with **MSAL.js**, there are a few important details to keep in mind.
```
mmacy, line 5: @derisen In B2C parlance, a **local** account is a consumer account created using email/password (or username/password).

https://docs.microsoft.com/azure/active-directory-b2c/technical-overview#consumer-accounts

From the doc:

### Consumer accounts

With a *consumer* account, users can sign in to the applications that you've secured with Azure AD B2C. Users with consumer accounts can't, however, access Azure resources, for example the Azure portal.

A consumer account can be associated with these identity types:

* **Local** identity, with the username and password stored locally in the Azure AD B2C directory. We often refer to these identities as "local accounts."
* **Social** or **enterprise** identities, where the identity of the user is managed by a federated identity provider like Facebook, Microsoft, ADFS, or Salesforce.

lib/msal-browser/docs/working-with-b2c.md
```diff
@@ -0,0 +1,104 @@
+# Working with MSAL.js and Azure AD B2C
+
+> :warning: Before you start here, make sure you understand [how to initialize an app object](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/initialization.md) and [working with resources and scopes](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/resources-and-scopes.md). We also recommend general familiarity with **Azure AD B2C**. See the [B2C documentation](https://docs.microsoft.com/azure/active-directory-b2c/technical-overview) for more.
```
mmacy, line 3: Should we not instead link to the docs article for the initialization?

https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-js-initializing-client-applications

lib/msal-browser/docs/working-with-b2c.md
derisen: Ah, my bad -fixed!

lib/msal-browser/docs/working-with-b2c.md
derisen: I did link the initialization docs in the repository just above. I have the impression that its better if these docs refer to each other within the repo.

lib/msal-browser/docs/working-with-b2c.md
```diff
@@ -0,0 +1,104 @@
...
+
+**MSAL.js supports** authentication with **social** (Microsoft, Google, Facebook etc.), **enterprise** (ADFS, Salesforce etc.) and **local** (stored in the Azure AD B2C directory) identities using **Azure AD B2C** (*B2C* for short). When developing B2C apps with **MSAL.js**, there are a few important details to keep in mind.
+
+## Quick Facts
+
+With B2C:
+
+- Users **can** authenticate with their social identities.
+- Users **can** be authorized to access **B2C protected** resources (but not AAD protected resources).
+- Users **cannot** obtain tokens for Microsoft APIs (e.g. MS Graph API) using [delegated permissions](#b2c-and-delegated-permissions).
```
sameerag, line 13: +1

lib/msal-browser/docs/working-with-b2c.md
```diff
@@ -0,0 +1,104 @@
...
+
+## B2C and Delegated Permissions
+
+**Delegated permissions** specify *scope-based access* using *interactive* authorization from the signed-in user. These permissions are presented to the resource (*e.g.* your web API, MS Graph API etc.) at run-time as `scp` claims in the client's **Access Token**.
+
+A user's B2C authentication cannot be used to authorize to AAD protected apps, or **Microsoft APIs** (which are also protected by AAD). As such, when you use **MSAL.js**, you cannot use the `<your-tenant>.b2clogin.com/.../token` endpoint to obtain a token for **MS Graph API**.
+
+### OpenID Connect Permissions
+
+The exception to the rule above comes from a special set of scopes known as **OpenID Connect** (OIDC) permissions, which includes `openid` and `profile`. Another special permission is the `offline_access`, which gives your app access to a resources on behalf of the user for an extended time (using a **Refresh Token**). MSAL.js will supply `openid`, and `profile` by default during `loginPopup()` and `loginRedirect()` requests.
```
sameerag, line 70: Why is this an exception? It is confusing to say OIDC permissions are exception, as they mean different things for B2C endpoint and MS Graph endpoint. Should we rephrase this?

derisen: Hmm, by exception I meant that even though they are listed under MS Graph permissions on the Azure Portal, they're not used for any Graph resource (/userinfo is technically hosted on Graph but...). 

I'm thinking that until the Portal UI updates, people will be confused why these scopes and not other Graph scopes. I think once that's done we can remove this explanation.. -or do you have a suggestion for rephrase?

PR #1998, opened 2020-07-20
Fix empty request

tnorling:
When not providing a request object to loginRedirect, error occurs because it expects `redirectStartPage`. This PR fixes this.


lib/msal-browser/test/app/PublicClientApplication.spec.ts
```diff
@@ -460,6 +460,22 @@ describe("PublicClientApplication.ts Class Unit Tests", () => {
                 pca.loginRedirect(loginRequest);
             });
 
+            it.only("loginRedirect navigates to created login url, with empty request", (done) => {
```
jasonnutter, line 463: `.only`

tnorling: Ah! Thanks for catching this!

Issue #1761, opened 2020-06-07
specifying common authority in acquireTokenSilent results in cache miss


onlyann:
>> Please follow the issue template below. Failure to do so will result in a delay in answering your question.

## Library
- [x] `msal@1.3.1` 

## Description
Specifying an authority of _https://login.microsoftonline.com/common/_ to `acquireTokenSilent` will result in a access token cache miss every time.

## Error Message

No error message.
However, the access token will be fetched from Azure AD every single time the token is silently acquired.

## MSAL Configuration

It doesn't seem to have an impact. Example of configuration:

```js
auth: {
        clientId: config.clientId,
        postLogoutRedirectUri: window.location.origin,
        redirectUri: window.location.origin
      },
cache: {
        cacheLocation: "localStorage",
        storeAuthStateInCookie: true
      }
```

## Reproduction steps
```js
acquireTokenSilent({
  scopes,
  authority: "https://login.microsoftonline.com/common/"
})
```
## Expected behavior

I would expect that the library either warns that _https://login.microsoftonline.com/common/_ is not a valid authority value when calling `acquireTokenSilent` or that it would ignore such authority.

When logging in with _https://login.microsoftonline.com/common/_ authority, the retrieved access token gets cached using the `tid` value of the id token. 
Whenever an authority is passed to `acquireTokenSilent`, it gets used as is to find an access token from the cache, which explains the cache miss.

## Browsers/Environment
Any

jmckennon:
If you don't include the authority as `https://login.microsoftonline.com/common/`, do you still see the same functionality? The common endpoint is the default authority. This should be a workaround for now.

We've been able to reproduce the bug when that default authority is provided and we always get a cache miss. Will move it over to be fixed.

onlyann:
Yes, I am currently working around it by not providing the authority.

lidermanrony:
Hi. Any ETA for a fix?

jmckennon:
@lidermanrony It's in our backlog but we have not dove deeper into this yet, as the workaround seems to be fairly simple, making this low-priority.

Are you seeing other problems or unable to use the workaround for now?

lidermanrony:
I had no problem working around it, but the issue is very common (no pun intended :-)) and on the other hand is not very easy to notice (since it mostly effects performance). It took us a while to understand that our customers are experiencing double the amount of wait time when we moved from ADAL to MSAL

lidermanrony:
[...]

PR #2015, opened 2020-07-21
[msal-core] Fix cache miss when common authority is provided

ahnmelis:
Addresses issue #1761 .
Created a function in `UrlUtils` named `isCommonAuthority` which takes in an authority string and returns a `boolean`. `isCommonAuthority` is called by `getCachedToken`, which if true, will ignore the authority when filtering.


ahnmelis:
New commits add fixes that address a cache miss when organizations authority is provided. Added additional authority domain filter to address the case of a user logged in with two different domains (but the same tenant).

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1404,7 +1404,26 @@ export class UserAgentApplication {
...
-                throw ClientAuthError.createMultipleMatchingTokensInCacheError(scopes.toString());
+                //find an exact match for domain
+                const requestDomain = UrlUtils.GetUrlComponents(serverAuthenticationRequest.authority).HostNameAndPort;
+                const filteredAuthorityItems: Array<AccessTokenCacheItem> = [];
+                for (let i = 0; i < filteredItems.length; i++) {
+                    const domain = UrlUtils.GetUrlComponents(filteredItems[i].key.authority).HostNameAndPort;
+                    if (domain === requestDomain) {
+                        filteredAuthorityItems.push(filteredItems[i]);
+                    }
+                }
```
jasonnutter, line 1415: This can also be written using `filteredItems.filter`:

```js
const filteredAuthorityItems = filteredItems.find(filteredItem => {
  const domain = UrlUtils.GetUrlComponents(filteredItem.key.authority).HostNameAndPort;
  return domain === requestedDomain;
});

ahnmelis: for my own understanding, does using the find() function only return the first element where `domain === requestDomain`?

jasonnutter: Yes, sorry, meant to use `.filter` (.find is not supported in IE). Updated snippet.

ahnmelis: made the changes

PR #2017, opened 2020-07-22
Api change suggestions

AndrewCraswell:
Made some changes to make more use of the Context API and hooks. Cleaned up the samples and adhered to tighter typing conventions. This PR isn't necessarily meant to be merged, but if some of the conventions look useful they can be ported over.

- Moved the `MsalContext` into its own file and configured default values
- Added a `useMsal` hook which just subscribes to the Context values
- Migrated the `MsalProvider` to a Function Component and made use of hooks
- Updated the `withMsal` to improve typing support and adhere closer to best practices
- Renamed the `useIsAuthenticated` hook to `useMsalAuthentication` to make its use more clear. This hook needs more work to better expose the authentication state and any errors. It's possible this hook isn't necessary, and the logic can be moved into the `MsalProvider`
- Added `AuthenticatedTemplate` and `UnauthenticatedTemplate` which renders the contents only if the authentication state is correct
- Renamed the `AuthenticatedComponent` to `MsalAuthentication` for clarity of purpose, and removed `renderProps` in favor of composition using `AuthenticatedTemplate` and `UnauthenticatedTemplate` components
- Made the `MsalAuthentication`, `AuthenticatedTemplate`, and `UnauthenticatedTemplate` components utilize optional FaaC pattern
- Updated the samples to use the new components

I had trouble fully testing all the scenarios -- I think I didn't set up the project correctly and was getting errors thrown in the dependency libraries. And there's definitely some authentication edge cases that need to be built out fully. I think the API surface area can be reduced further to make it simpler, but for a start, it already feels much more flexible than `react-aad-msal`.


samples/msal-react-sample/src/index.tsx
```diff
@@ -12,7 +12,7 @@ import { Configuration } from "@azure/msal-browser";
 const configuration: Configuration = {
```
sameerag, line 12: can we separate the config file as other samples to have more flexibility in testing?

samples/msal-react-sample/src/App.tsx
```diff
@@ -33,60 +59,16 @@ function App() {
...
-                                        e.preventDefault();
-                                        msal?.logout();
-                                    }}
-                                >
-                                    Logout
-                                </button>
-                            </AuthenticatedComponent>
-                        )}
-                    </MsalConsumer>
+                    <AuthenticatedTemplate>
```
jasonnutter, line 62: Why `Template` versus `Component`?

samples/msal-react-sample/src/msal-react/MsalContext.ts
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
+import { AuthenticationParameters, TokenResponse, TokenRenewParameters, Account,  } from "@azure/msal-browser";
+
+// TODO: Move this to msal-browser
+export interface IPublicClientApplication {
```
jasonnutter, line 5: This now exists in msal-browser, FYI.

samples/msal-react-sample/src/msal-react/index.tsx
```diff
@@ -1,82 +1,87 @@
...
-                setError(error);
-            });
-        } else {
-            context?.loginRedirect(request);
-        }
-        
-    }
+    // TODO: This is error prone because it asynchronously sets state, but the component may be unmounted before the process completes
+    //  May be better to move this into the the MsalContext
+    const loginInteractively = useCallback(
```
jasonnutter, line 68: I personally am not a fan of the design where logic around forcing login and interaction is part of configuring MSAL itself, which is why we the first version put that logic into `AuthenticatedComponent`. As an ethos, we try to to favor composition over configuration, and delegating requiring logic, preferred interaction, etc to each component tree that requires authentication more closely aligns with that, I believe.

jasonnutter:
@AndrewCraswell Can you please updated the `README.md` to summarize your proposed changes? Finding it hard to grep all the proposed changes by just looking at the diff, thanks!

samples/msal-react-sample/src/App.tsx
AndrewCraswell: There may be a better alternative than `Template`, but `Component` feels redundant and might imply there is more logic behind this when it's really just mounting a defined `ReactNode` if the authentication state is correct. I was hoping `Template` would elude to it being pretty logic-less

samples/msal-react-sample/src/index.tsx
AndrewCraswell: God point. I've split the `Configuration` into a [`config.ts`](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/156db446b880dae69f9c9f68ebe6e4425af10ded/samples/msal-react-sample/src/config.ts) file and also moved the values into the [.env](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/156db446b880dae69f9c9f68ebe6e4425af10ded/samples/msal-react-sample/.env) file.

samples/msal-react-sample/src/msal-react/MsalContext.ts
AndrewCraswell: Thanks! After referencing the proper interface I see that the `@azure/msal-browser` package has made a lot of changes. I updated the code to work with the latest package version, but I think there's a lot of design decisions that are raised. `msal-browser` takes some key API changes that should probably modify how the `msal-react` component is expected to behave.

PR #2022, opened 2020-07-23
[msal-core] Fix response type configuration by basing it mainly on scopes

### LINKED ISSUE #1241: acquireTokenSilent() doesn't get new ID token

### LINKED ISSUE #1779: acquireTokenSilent - refreshed ID tokens have incomplete scope
My app uses 4 scopes [`profile`, `email`, `openid`, `User.Read`] when retrieving an ID token and everything works fine, e.g. the initial ID token includes the "email" claim. 

When the ID token expires, the app uses `UserAgentApplication.acquireTokenSilent` to fetch a new token, but ID token received is missing the `email` claim and other info related to `User.Read` scope. As stated in the code "To renew idToken, please pass clientId as the only scope in the Authentication Parameters", the app is passing only the clientId in the scopes array.

The library is configured to use session storage, but I've experienced the same issue with localstorage.

After forking the repo and modifying a file (see below), I managed to fix this problem, so it appears to be an issue in the library. I'm looking for suggestions so that I don't have to maintain a fork of the lib.

Looking at the http requests, this is the first request, which uses the 4 scopes configured in my app, and it gets and ID token with all the claims as expected:

```
https://login.microsoftonline.com/consumers/oauth2/v2.0/authorize
  ?response_type=token
  &scope=profile%20email%20openid%20User.Read
  &...
```

This is the request used by MSAL to refresh the ID token, via `acquireTokenSilent`, and it has only 2 scopes (`openid` and `profile`):

```
https://login.microsoftonline.com/consumers/oauth2/v2.0/authorize
  ?response_type=id_token
  &scope=openid%20profile
  &...
```

`UserAgentApplication.acquireTokenSilent` accepts a `userRequest` parameter, which has `scopes` and `extraScopesToConsent` parameters, but no matter how these params are set, custom scopes are ignored, and the request to `login.microsoftonline.com` includes only 2 scopes (`openid` and `profile`), which are hardcoded in `UrlUtils.translateclientIdUsedInScope` https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/b3a76366d899b75764f47edabbdb14fba84f7210/lib/msal-core/src/utils/UrlUtils.ts#L85
`RequestUtils.validateRequest` is also designed to ignore `extraScopesToConsent` in this scenario, and the library causes an infinite loop if `scopes` contains anything other that the client id.

As mentioned, this is not a problem with AAD. After modifying `UrlUtils.translateclientIdUsedInScope` to include my custom scopes, I managed to get the refreshed ID tokens with all the information I needed, i.e. the same claims of the initial ID token.

```typescript
private static translateclientIdUsedInScope(scopes: Array<string>, clientId: string): void {    
    const clientIdIndex: number = scopes.indexOf(clientId);
    if (clientIdIndex >= 0) {
        scopes.splice(clientIdIndex, 1);
        if (scopes.indexOf("openid") === -1) {
            scopes.push("openid");
        }
        if (scopes.indexOf("profile") === -1) {
            scopes.push("profile");
        }

        // Workaround for missing claims in refreshed ID tokens
        if (scopes.indexOf("email") === -1) {
            scopes.push("email");
        }
        
        // Workaround for missing claims in refreshed ID tokens
        if (scopes.indexOf("User.Read") === -1) {
            scopes.push("User.Read");
        }
    }
}
```

Call stack when renewing ID tokens, leading to 2 scopes only, regardless of the initial params

1. UserAgentApplication.acquireTokenSilent (scope contains only `clientId` as documented)
2. -> UserAgentApplication.renewIdToken (this is called only if scope contains `clientId`)
3. -> UrlUtils.createNavigateUrl (this is called passing `serverAuthenticationRequest` which contains a copy of the scopes in `request`)
4. -> UrlUtils.createNavigationUrlString 
5. -> UrlUtils.translateclientIdUsedInScope

hectormmg:
This PR:
- Fixes #1241 by enabling the id_token token response type for acquireToken APIs when scopes configured include `openid` or `profile`
- Removes clientId as only scope restriction, while keeping the behavior sending in clientId as a single scope enabled (i.e. setting response_type to id_token for acquireToken calls)
- Refactors the logic that adds `openid` and `profile` to the request scopes by default if they are missing
- Refactors the logic that removes clientId from final scopes to make sure it is only removed when it is the only scope (otherwise it is now treated as a resource scope)
- Adds through tests for the mapping of scopes and account configuration to the resulting response_type for acquireToken logic.
- Adds docs on `Scopes` and `Response Types` behavior for msal-core


lib/msal-core/docs/response-types.md
```diff
@@ -0,0 +1,72 @@
...
+* *OIDC scopes only: Same as OIDC scopes but with no other scopes in the array*
+
+| Input scopes | Account passed in | Response Type Result |
+| ----------------- | ------------ | -------------------- |
+| ClientId as only scope | Any case | `id_token`|
+| OIDC scopes only | Any Case | `id_token`|
+| ClientId with OIDC scopes | Any case | `id_token token` |
+| Resource scope(s) only | Matches cached account object | `token` |
+| Resource scope(s) and ClientId | Matches cached account object | `token` |
+| Resource scope(s) with OIDC scopes | Matches cached account object | `id_token token` |
```
tnorling, line 66: Would this be "Any Case" as well? As line 68 looks the same?

lib/msal-core/src/ScopeSet.ts
```diff
@@ -160,4 +153,93 @@ export class ScopeSet {
...
+    }
+
+    /**
+     * @ignore
+     * Removes clientId from scopes array if included as only scope. If it's not the only scope, it is treated as a resource scope.
+     * @param scopes Array<string>: Pre-normalized scopes array
+     * @param clientId string: The application's clientId that is searched for in the scopes array
+     */
+    private static spliceClientIdIfSingleScope(scopes: Array<string>, clientId: string): Array<string> {
+        const clientIdIndex: number = scopes.indexOf(clientId);
```
tnorling, line 236: you can simplify this function by doing something like 

`return this.onlyContainsClientId(scopes, clientId)? [] : scopes`

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -2238,41 +2225,14 @@ export class UserAgentApplication {
...
-            else {
-                tokenType = (scopes.indexOf(this.clientId) > -1) ? ResponseTypes.id_token : ResponseTypes.token;
-            }
-
-            return tokenType;
-        }
-
+    private getTokenType(accountObject: Account, scopes: string[]): string {
+        const accountsMatch = Account.compareAccounts(accountObject, this.getAccount());
+        const loginScopesOnly = ScopeSet.onlyContainsClientId(scopes, this.clientId) || ScopeSet.onlyContainsOidcScopes(scopes);
```
tnorling, line 2234: Why not do the clientId translation when validating the request and then you only have to call `onlyContainsOidcScopes` here?

tnorling: Then instead of having the normalize function that translates then appends default scopes you can just append the default scopes. The flow make's a little bit more sense to me.

tnorling (commented):
Overall nice work! Just a few comments about cleaning things up a bit

lib/msal-core/docs/response-types.md
```diff
@@ -0,0 +1,72 @@
...
+
+| Response Type | Specification that defines it | Expected token type from successful request | Action |
+| ------------- | ----------------------------- | ------------------------------------------- | ------ |
+| `token` |[OAuth 2.0](https://tools.ietf.org/html/rfc6749#section-3.1.1) | Access Token | Authorization |
+| `id_token`| [OpenID Connect](https://openid.net/specs/openid-connect-core-1_0.html#Authentication) | ID Token | Authentication |
+|`id_token token`| [OpenID Connect](https://openid.net/specs/openid-connect-core-1_0.html#Authentication) | Access Token and ID token | Authorization & Authentication |
+
+**Note: Given that `msal@1.x` uses the OAuth 2.0 Implicit Flow exclusively, which leverages URL fragments for token reception, it is important to be mindful of URL length limitations. In some cases, getting both an access token and ID token in the same URL may cause unexpected or incorrect behavior.**
```
sameerag, line 20: Please add  `browsers like IE which may impose restrictions on the URL length`

jmckennon (approved):
Looks good!

lib/msal-core/docs/response-types.md
technical-boy: Right!

lib/msal-core/src/ScopeSet.ts
technical-boy: Awesome! I had a feeling there was a way to clean this up that I hadn't though of. This is exactly what I was looking for!

lib/msal-core/src/UserAgentApplication.ts
technical-boy: Yeah, that definitely sounds better, implementing.

lib/msal-core/docs/response-types.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Response Type configuration and behavior
+
+The `response_type` attribute presented above is not a directly configurable element of an authorization server request in `msal@1.x`. However, it is important to understand the way `msal@1.x` determines which response type is set and, therefore, what kind of token the developer can expect for each scenario. The factors that come into consideration when setting the request's `response_type` parameter are the following:
```
sameerag, line 24: I would rather say: `The response_type is not user configurable. However the library sets this parameter based on the below factors`

lib/msal-core/docs/response-types.md
technical-boy: I will, I actually wasn't 100% sure which browser it was until now, thanks!

lib/msal-core/docs/response-types.md
```diff
@@ -0,0 +1,72 @@
...
+
+1. The `msal@1.x` API called
```
sameerag, line 26: Applicable only to `msal@1.x`

lib/msal-core/docs/response-types.md
```diff
@@ -0,0 +1,72 @@
...
+2. Whether the account passed into the request configuration matches the account in the MSAL cache
+3. The contents of the `scopes` array in the Authorization Request Configuration object. For more information on `scopes` configuration, please consult the [Scopes](/docs/scopes.md) document.
+
+**Important note: Login APIs will always set `response_type=id_token`. It doesn not matter whether a matching account is passed in or what scopes are configured.** The fact that there is no extra considerations for `login` APIs to determine the `response_type` make it the most important factor.
```
sameerag, line 30: typo `doesn't or does not`

lib/msal-core/docs/response-types.md
technical-boy: Done.

lib/msal-core/docs/response-types.md
sameerag: You can simplify the note to: `Login APIs by definition are for a user to login (authenticate) and always default to response_type = id_token`

lib/msal-core/docs/response-types.md
technical-boy: Changed.

lib/msal-core/src/ScopeSet.ts
technical-boy: Ok, ended up implementing this in the renamed translateClientIdIfSingleScope, only instead of an empty array we now actually return the OIDC scopes array if clientId is the only scope.

lib/msal-core/src/UserAgentApplication.ts
technical-boy: Implementing this change actually revealed a bunch of other places where clientId was being added as a single scope for requests that had null scopes. In all of the found cases, clientId has been replaced with OIDC scopes array to avoid having clientId being added after validation, to avoid clientId ending up in the final scopes array. In other words, since we now translate during validation and not at the end, if the clientId was added post-validation, it wouldn't be caught.

lib/msal-core/docs/response-types.md
technical-boy: Added another warning at the top of the document.

lib/msal-core/docs/response-types.md
technical-boy: Simplified per your suggestion, typo removed as well.

lib/msal-core/docs/scopes.md
```diff
@@ -0,0 +1,88 @@
...
+
+### Scope Functions
+
+#### Function of scopes in OAuth 2.0
+
+The main function of the `scopes` configuration, per the [OAuth 2.0 Access Token Scope Reference](https://tools.ietf.org/html/rfc6749#section-3.3), is to determine the permissions for which an application requests `authorization` on behalf of the user. Said function is both supported and covered by `msal@1.x` and the Microsoft identity platform in general. For more information on the regular function of authorization scopes, please consult the official [Microsoft identity platform documentation](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-permissions-and-consent).
+
+#### Special use of scopes in msal@1.x
+
+In addition to the global concept and use of `scopes`, it is important to understand that `msal@1.x` gives scopes a special use that adds to the importance of their configuration. In short, `msal@1.x` allows developers to leverage certain scopes in order to determine the `response_type` for the final request. For more information on the way the scopes configuration determines the `response_type` parameter, please refer to the [Response Types Document](/docs/response-types.md).
```
sameerag, line 24: `in short` is not needed

lib/msal-core/docs/scopes.md
```diff
@@ -0,0 +1,88 @@
...
+
+
+## Scopes Behavior
+
+### Default Scopes on Authorization Requests
+
+Understanding how `OIDC scopes` configure the claims included in an authentication response's ID Token is important when using `msal@1.x` to acquire said ID Tokens. However, there is an important note to be made on how the `openid` and `profile` scopes are added by `msal@1.x` to all server requests by default that does not directly relate to the OpenID Connect specification.
+
+Like previously mentioned, the Secure Token Service that `msal@1.x` requests access and ID tokens from also makes use of the `openid` and `profile` scopes. Specifically, the STS expects these two scopes in order to configure and provide the `client_info` parameter in authorization and authentication responses. The `msal@1.x` library depends on the contents of `client_info` in order to successfuly cache tokens and, therefore, provide silent token acquisition as a feature.
```
sameerag, line 59: `client_info` is internal detail. Would we want this in documentation? I prefer not to. cc @tnorling

lib/msal-core/docs/scopes.md
```diff
@@ -0,0 +1,88 @@
...
+
+**For this reason, whether or not the developer adds the `openid` or `profile` scopes to their request configuration, `msal@1.x` will make sure they are included before sending the request to the STS.**
+
+### Special OIDC Scopes behavior cases
+
+The following is a list of practical implications and examples of the default scope behavior described in the previous section.
+
+- If the scopes array does not include either `openid` or `profile`, whichever is missing (could be both) will be added to the scopes array by default before the request is sent out.
```
sameerag, line 67: Adding a custom scope (instead of graph scope) in an example would help

lib/msal-core/src/ServerRequestParameters.ts
```diff
@@ -277,4 +277,29 @@ export class ServerRequestParameters {
...
         return request && (request.account || request.sid || request.loginHint);
     }
+
+    /**
+     * Returns the correct response_type string attribute for an acquireToken request configuration
+     * @param accountsMatch boolean: Determines whether the account in the request matches the cached account
+     * @param scopes Array<string>: AuthenticationRequest scopes configuration
+     * @param loginScopesOnly boolean: True if the scopes array ONLY contains the clientId or any combination of OIDC scopes, without resource scopes
+     */
+    static determineResponseType(accountsMatch: boolean, scopes: Array<string>, loginScopesOnly: boolean) {
```
sameerag: `setResponseType()`?

lib/msal-core/src/UserAgentApplication.ts
sameerag: Remember, you cannot mutate user passed scopes, and if we are returning what is requested (not sure if we do), we need to retain the user passed scopes info. Having said that, I agree with @tnorling.

lib/msal-core/src/utils/Constants.ts
```diff
@@ -72,6 +72,19 @@ export enum ServerHashParamKeys {
...
+ * response_type from OpenIDConnect
+ * References: https://openid.net/specs/oauth-v2-multiple-response-types-1_0.html & https://tools.ietf.org/html/rfc6749#section-4.2.1
+ *
+ */
+export const ResponseTypes = {
+    id_token: "id_token",
+    token: "token",
+    id_token_token: "id_token token"
+};
+
```
sameerag, line 88: Is this is new, can we remove the older reference in `Constants.ts`?

technical-boy: Not new, it was in UserAgentApplication, moved it to Constants.

lib/msal-core/src/ServerRequestParameters.ts
technical-boy: I think calling it setResponseType would be a misnomer, since "set" is usually an attribute mutator and this method returns a value without mutating the actual request params.

lib/msal-core/docs/scopes.md
technical-boy: I tried not to go too into the weeds, but these docs are for developers at the end of the day. I think anyone trying to change this in the future should know why it was done.

tnorling: Others may disagree but, I personally feel that docs should be as much for the customers as they are for us/new members of the team so I don't have an issue with sharing implementation details as long as customers are still easily able to find what they're looking for without wading through stuff they don't care about. Maybe to accomplish this we include a quick TLDR summary at the top of each doc that just contains the important points and then we go into detail below?

lib/msal-core/src/ScopeSet.ts
```diff
@@ -160,4 +153,71 @@ export class ScopeSet {
...
+            oidcScopesFound += 1;
+        }
+
+        if (scopes.indexOf(Constants.profileScope) > -1) {
+            oidcScopesFound += 1;
+        }
+
+        const nonOidcScopes = scopes.length - oidcScopesFound;
+
+        return (nonOidcScopes === 0);
```
tnorling, line 173:
```suggestion
        return (scopes.length === oidcScopesFound);
```

lib/msal-core/src/ScopeSet.ts
```diff
@@ -160,4 +153,71 @@ export class ScopeSet {
...
+
+        return (containsOpenIdScope || containsProfileScope);
+    }
+
+    /**
+     * @ignore
+     * Returns true if the clientId is the only scope in the array
+     */
+    static onlyContainsClientId(scopes: Array<String>, clientId: string): boolean {
+        return (scopes) ? (scopes.indexOf(clientId) > -1 && scopes.length === 1) : false;
```
tnorling, line 193: Doesn't need to be a ternary
```suggestion
        return scopes && (scopes.indexOf(clientId) > -1) && (scopes.length === 1) ;
```

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -528,8 +515,8 @@ export class UserAgentApplication {
 
         // Track the acquireToken progress
         this.cacheStorage.setItem(TemporaryCacheKeys.INTERACTION_STATUS, Constants.inProgress);
-        const scope = request.scopes ? request.scopes.join(" ").toLowerCase() : this.clientId.toLowerCase();
-        this.logger.verbosePii(`Serialized scopes: ${scope}`);
+        const requestSignature = request.scopes ? request.scopes.join(" ").toLowerCase() : this.clientId.toLowerCase();
```
tnorling, line 518: Should the else condition be `Constants.oidcScopes`?

tnorling (approved):
Just a few nitpick suggestions but looks good otherwise

lib/msal-core/src/ScopeSet.ts
technical-boy: I thought so too, but I did some testing and it turns out when scopes is null this would actually return null and not false (unless you double negate it).

lib/msal-core/src/UserAgentApplication.ts
technical-boy: Given this is only used as the request signature I didn't see an absolute need to change it. Mostly because I didn't know the reasoning behind originally using clientId (does it need to match the scopes for any specific reason?), so I didn't want to change this without knowing.

lib/msal-core/src/ScopeSet.ts
tnorling: Can you do `!!scopes` ?

tnorling: Oh didn't see your note about double negation

technical-boy: Do you prefer `!!scopes`? I have no preference, personally.

lib/msal-core/src/UserAgentApplication.ts
tnorling: Since we moved the translation to happen before this I don't think it's wise to use clientId as a single scope or default. Even if it's not used for anything of substance today it may be used in the future when we may forget about this.

lib/msal-core/docs/scopes.md
technical-boy: I actually added it because I don't like seeing `msal@1.x` (which starts with a lowercase 'm' at the beginning of a sentence. For the same reason I think it's better to keep it.

lib/msal-core/src/UserAgentApplication.ts
technical-boy: I did change the variable name for the same reason, to avoid future confusion. If there's no reason too keep the clientId as request signature, I'll change it to use OIDC scopes as default signature itself. Thanks!

lib/msal-core/src/ScopeSet.ts
technical-boy: I removed the unnecessary subtraction per your suggestion, but I also added a non-zero length check for the array (technically empty arrays shouldn't make it this far, but you can never be too careful).

lib/msal-core/src/ScopeSet.ts
tnorling: I don't feel strongly about it but I think I slightly prefer just doing a straight boolean check over a ternary. I'll leave the ultimate decision up to you or if someone else feels more strongly one way or the other.

lib/msal-core/docs/scopes.md
technical-boy: I think the main purpose of this specific set of docs is to bring clarity to ambiguous features/behaviors. Even while trying to be concise, I don't feel comfortable obfuscating the very implementation details that made this such a complex issue in the first place.

lib/msal-core/src/ScopeSet.ts
technical-boy: I made the change, adding a comment about the double negation to make it clearer!

lib/msal-core/docs/scopes.md
tnorling: I totally agree, but I think you can have a summary that distills it down to a few important points without explaining the why if someone only cares about "how do I do x?" For example for the scopes doc the 2 most important things to take away are: 
- Msal.js will always add `openid` and `profile to every request
- ClientId sent as the only scope will be replaced by `openid` and `profile` and return an id_token

For the response_type doc:
- Login requests will always return only an id_token
- AcquireToken requests will always return an id_token if `openid` or `profile` are included in the request
- AcquireToken requests will always return an access_token if a resource scope is requested

Sure there are more nuances and more details and those should stay. But I think these quick points will answer 75% of people's questions without needing to read the whole doc.

technical-boy: Added TLDRs at the top of each doc as a "Quick Reference" section with a small indication that those are the key takeaways and people should read on for more info/clarification. Thanks for the suggestions!

lib/msal-core/docs/scopes.md
technical-boy: Added a custom scope example!

sameerag (approved):
LGTM.

lib/msal-core/docs/response-types.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Response Type configuration and behavior
+
+The `response_type` attribute presented above cannot be configured directly. However, it is important to understand the way `msal@1.x` determines which response type is set and, therefore, what kind of token the developer can expect for each scenario. The factors that come into consideration when setting the request's `response_type` parameter are the following:
+
+1. The `msal@1.x` API called
+2. Whether the account passed into the request configuration matches the account in the MSAL cache
+3. The contents of the `scopes` array in the Authorization Request Configuration object. For more information on `scopes` configuration, please consult the [Scopes](/docs/scopes.md) document.
+
+**Important note: Login APIs will always set `response_type=id_token`, given that they are designed to perform user login (authentication).
```
jo-arroyo, line 44: Wondering if you intended to call this out with ** at the beginning only, or wanting to bold it with ** at the end too?

lib/msal-core/docs/response-types.md
```diff
@@ -0,0 +1,85 @@
...
+
+Applies to: `acquireTokenRedirect`, `acquireTokenPopup`, `acquireTokenSilent`
+
+* *OIDC scopes: any combination of `openid` and/or `profile`*
+* *OIDC scopes only: Same as OIDC scopes but with no other scopes in the array*
+
+| Input scopes | Account passed in | Response Type Result |
+| ----------------- | ------------ | -------------------- |
+| ClientId as only scope | Any case | `id_token`|
+| OIDC scopes only | Any Case | `id_token`|
```
jo-arroyo, line 76:
```suggestion
| OIDC scopes only | Any case | `id_token`|
```
Lowercase to match the rest of the table?

lib/msal-core/docs/scopes.md
```diff
@@ -0,0 +1,105 @@
...
+
+
+## Scopes Behavior
+
+### Default Scopes on Authorization Requests
+
+Understanding how `OIDC scopes` configure the claims included in an authentication response's ID Token is important when using `msal@1.x` to acquire said ID Tokens. However, there is an important note to be made on how the `openid` and `profile` scopes are added by `msal@1.x` to all server requests by default that does not directly relate to the OpenID Connect specification.
+
+Like previously mentioned, the Secure Token Service that `msal@1.x` requests access and ID tokens from also makes use of the `openid` and `profile` scopes. Specifically, the STS expects these two scopes in order to configure and provide the `client_info` parameter in authorization and authentication responses. The `msal@1.x` library depends on the contents of `client_info` in order to successfuly cache tokens and, therefore, provide silent token acquisition as a feature.
```
jo-arroyo, line 74:
```suggestion
Like previously mentioned, the Secure Token Service that `msal@1.x` requests access and ID tokens from also makes use of the `openid` and `profile` scopes. Specifically, the STS expects these two scopes in order to configure and provide the `client_info` parameter in authorization and authentication responses. The `msal@1.x` library depends on the contents of `client_info` in order to successfully cache tokens and, therefore, provide silent token acquisition as a feature.
```

lib/msal-core/docs/scopes.md
```diff
@@ -0,0 +1,105 @@
...
+    ```js
+        { scopes: ['User.Read'] } // becomes { scopes: ['User.Read', 'openid', 'profile'] } before the request is sent
+
+        { scopes: ['User.Read', 'openid'] } // becomes { scopes ['User.Read', 'openid', 'profile']} before the request is sent
+
+        { scopes: ['User.Read', 'profile'] } // becomes { scopes ['User.Read', 'profile', 'openid']} before the request is sent
+
+        { scopes: ['http://contoso.com/scope'] } // becomes { scopes ['http://contoso.com/scope', 'openid', 'profile'] }
+    ```
+- ClientId is removed form the scopes array when it is the only scope in the configuration.If it is not the only scope, it is treated as a resource scope and will be sent in the final server request.
```
jo-arroyo, line 95:
```suggestion
- ClientId is removed form the scopes array when it is the only scope in the configuration. If it is not the only scope, it is treated as a resource scope and will be sent in the final server request.
```

jo-arroyo (approved):
LGTM! A few nits

lib/msal-core/docs/response-types.md
technical-boy: Good catch, thanks!

lib/msal-core/docs/scopes.md
technical-boy: Fixed.

lib/msal-core/docs/scopes.md
technical-boy: Added the space and corrected "form" to "from", thanks!

lib/msal-core/docs/response-types.md
technical-boy: Yes, I meant to close it, thanks!

PR #2023, opened 2020-07-24
[msal node] Adds confidential client

sangonzal:
Adds support for: 
- building confidential client application
- using a client secret 
- using a client certificate
- using client assertion

Coming soon: 
- support for client credentials grant
- support for on-behalf-of flow


lib/msal-node/src/utils/Constants.ts
```diff
@@ -45,3 +45,18 @@ export const CACHE = {
...
+    ALGORITHM: 'alg',
+    RSA_256: 'RS256',
+    X5T: 'x5t', 
+    AUDIENCE: 'aud',
+    EXPIRATION_TIME: 'exp',
+    ISSUER: "iss",
+    SUBJECT: "sub",
+    NOT_BEFORE: "nbf",
+    JWT_ID: "jti",
+}
```
pkanher617, line 62: We should make sure that files are ending with a newline, I get linter errors sometimes because of this.

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -158,6 +158,16 @@ export class AuthorizationCodeClient extends BaseClient {
             parameterBuilder.addCodeVerifier(request.codeVerifier);
         }
 
+        if (this.config.clientCredentials.clientSecret) {
```
sameerag, line 161: Nit: Strict null checks before this API would prevent  this failing? I mean what happens if `config.clientCredentials` is null?

sameerag: Since there is a default to `clientCredentials` it should be fine. Please add a test case though.

lib/msal-common/src/utils/Constants.ts
```diff
@@ -96,7 +96,10 @@ export enum AADServerParamKeys {
     X_CLIENT_OS = "x-client-OS",
     X_CLIENT_CPU = "x-client-CPU",
     POST_LOGOUT_URI = "post_logout_redirect_uri",
-    DEVICE_CODE = "device_code"
+    DEVICE_CODE = "device_code",
```
sameerag, line 99: Should we create a separate Enum? Can be done separately and not just these but the general constants lumped together - should they be divided?

lib/msal-node/src/client/ClientAssertion.ts
```diff
@@ -0,0 +1,88 @@
...
+import { CryptoProvider } from "../crypto/CryptoProvider";
+import { EncodingUtils } from '../utils/EncodingUtils';
+import { JwtConstants } from "../../src/utils/Constants"
+
+/**
+ * Client assertion of type jwt-bearer used in confidential client flows
+ */
+export class ClientAssertion {
+
+    public static readonly ASSERTION_TYPE = "urn:ietf:params:oauth:client-assertion-type:jwt-bearer";
```
sameerag, line 17: Should this be in `Constants.ts`?

sangonzal: Moved it to constants

lib/msal-node/src/client/ConfidentialClientApplication.ts
```diff
@@ -1 +1,61 @@
...
+            return;
+        }
+
+        if (configuration.auth.clientCertificate != null) {
+            const certificate = configuration.auth.clientCertificate;
+            if (StringUtils.isEmpty(certificate.thumbprint) || StringUtils.isEmpty(certificate.privateKey)) {
+                throw ClientAuthError.createInvalidCredentialError();
+            }
+
+            this.clientAssertion = ClientAssertion.fromCertificate(certificate.thumbprint, certificate.privateKey);
```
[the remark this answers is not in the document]
sangonzal: No, it should only be one or the other. Added a check with an useful error message to make this clear.

lib/msal-node/src/utils/Constants.ts
sangonzal: Added new line.

lib/msal-common/src/client/AuthorizationCodeClient.ts
sangonzal: Added test

baywet:
@sangonzal has this been released or does it still need to be released?

sangonzal:
@baywet Yes this has been released. The latest MSAL node version (v1.0.0-alpha.12) supports client credential grant and on-behalf-of flow.

Issue #2000, opened 2020-07-20
getAccountByUsername doesn't work in B2C scenarios


tnorling:
## Library
- [x] `@azure/msal-browser@2.x.x`

## Description
The `preferred_username` claim is not present on `idTokens` coming from B2C, breaking `getAccountByUsername`

## Expected behavior
`idTokens` should contain `preferred_username` claim or some other identifier to disambiguate users

## Workaround
For now when an account is needed, users should call `getAllAccounts()` and select their desired account from the returned list

tnorling:
Addressed in version 2.1.0 by providing `getAccountByHomeId` API as recommended approach.
To use `getAccountByUsername` in B2C scenarios, enable the `emails` claim on the B2C tenant

PR #2028, opened 2020-07-24
Fix getAccountByUsername for B2C Scenarios

tnorling:
Addresses #2000 

B2C does not return the `preferred_username` claim on idTokens which causes calls to `getAccountByUsername` to fail. B2C instead provides an `emails` claim that can be enabled on each user_flow. It returns an array of emails associated with the user. This PR provides the first email in this array as a fallback if `preferred_username` is undefined. 

There may be some future optimizations possible if we decide to support multiple emails returned on the `emails` claim.

TODO: 
- [ ] Currently being discussed with B2C team to find out if this is the recommended approach
- [ ] Add docs or FAQ explaining how to enable the claim on your user_flow

### DIFF
--- a/lib/msal-common/src/account/IdTokenClaims.ts
+++ b/lib/msal-common/src/account/IdTokenClaims.ts
@@ -14,6 +14,7 @@ export type IdTokenClaims = {
     ver?: string,
     upn?: string,
     preferred_username?: string,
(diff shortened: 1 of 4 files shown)


tnorling:
Closing due to decision made to take a different approach

Issue #1966, opened 2020-07-15
azure-msal-angular.js - unprotectedResources Array Issue


svpoon:
>> Please follow the issue template below. Failure to do so will result in a delay in answering your question.

## Library
- [x ] `msal@1.x.x` or `@azure/msal@1.x.x`
- [ ] `@azure/msal-browser@2.x.x`
- [ ] `@azure/msal-angular@0.x.x`
- [ ] `@azure/msal-angular@1.x.x`
- [ ] `@azure/msal-angularjs@1.x.x`

**Important**: Please fill in your exact version number above, e.g. `msal@1.1.3`.

## Description
Inside azure-msal-angular.js, the following code is not quite correct:
```
        _this.router.events.subscribe(function (event) {
            for (var i = 0; i < router.config.length; i++) {
                if (!router.config[i].canActivate) {
                    if (_this.msalAngularConfig.unprotectedResources) {
                        if (!_this.isEmpty(router.config[i].path) && !_this.isUnprotectedResource(router.config[i].path)) {
                            _this.msalAngularConfig.unprotectedResources.push(router.config[i].path);
                        }
                    }
                }
            }
        });
```

Should the condition  !_this.isUnprotectedResource be  _this.isUnprotectedResource instead?

In my scenario, it is adding all the paths from my routes as UnproectedResources.

Please check if it is correct or I am not setting up the configuration improperly.

## Security
- [ ] Is this issue security related? No

## Regression
- [ ] Did this behavior work before? 
Version:

## MSAL Configuration
```
        MsalModule.forRoot({
            auth: {
                clientId: '....',
                authority: '....',
                redirectUri: '....'
            },
        },
            {
                popUp: !isIE,
                consentScopes: [
                    'api://........', "openid", "profile"
                ],
                unprotectedResources: ["https://www.microsoft.com/en-us/"],
                protectedResourceMap: [
                    ['http://localhost:5000/', ['api://......']]
                ],
                extraQueryParameters: {}
            })
```

```js
// Provide configuration values here.
// For Azure B2C issues, please include your policies.
```

## Reproduction steps
```js
// Provide relevant code snippets here.
// For Azure B2C issues, please include your policies.
```

## Browsers/Environment
- [x] Chrome
- [x] Firefox
- [x] Edge
- [ ] Safari
- [ ] IE
- [ ] Other (Please add browser name here)

jo-arroyo:
[...]

Issue #1776, opened 2020-06-12
wildcard for protectedResourceMap ?


gItSpLasHeR:
Please follow the issue template below. Failure to do so will result in a delay in answering your question.

## Library
- [ ] `msal@1.x.x` or `@azure/msal@1.x.x`
- [ ] `@azure/msal-browser@2.x.x`
- [ ] `@azure/msal-angular@0.x.x`
- [x] `@azure/msal-angular@1.x.x`
- [ ] `@azure/msal-angularjs@1.x.x`

## Description
msal interceptor will attach the token to the requests which url & scope listed in "protectedResourceMap" as below format
`      protectedResourceMap: [
        ['https://graph.microsoft.com/v1.0/me', ['user.read']]
      ]`
since in my situation i got some urls with same scope.
so is it possible to use a wild card url in protectedResourceMap? 
which similar like this
`      protectedResourceMap: [
        ['https://*.microsoft.com/*', ['user.read']]
      ]`

or

`      protectedResourceMap: [
        ['*', ['user.read']]
      ]`

jasonnutter:
@gItSpLasHeR This is reasonable ask, we'll track it and follow up when we have a chance to address, thanks!

vitaliidasaev:
For those who looks about msal-angular MsalHttpInterceptor.  
I've made own implementation of MsalInterceptor to overcome issues with protectedResourceMap.  
This is my custom code block:

    // #region own workaround in order not to put every api endpoint url to settings
    if (!scopes && req.url.startsWith(this.settingsService.apiUrl)) {
      scopes = [this.auth.getCurrentConfiguration().auth.clientId];
    }
    // #endregion
    
    // If there are no scopes set for this request, do nothing.
    if (!scopes) {
      return next.handle(req);
    }

Issue #1712, opened 2020-05-27
Authorization missing in request headers for specific words in URL 


mylle:
>> Please follow the issue template below. Failure to do so will result in a delay in answering your question.

## Library
- [ ] `msal@1.x.x` or `@azure/msal@1.x.x`
- [ ] `@azure/msal-browser@2.x.x`
- [ ] `@azure/msal-angular@0.x.x`
- [x ] `@azure/msal-angular@1.0.0-beta-4`
- [ ] `@azure/msal-angularjs@1.x.x`

**Important**: Please fill in your exact version number above, e.g. `msal@1.1.3`.

## Framework
Angular

## Description
I have a weird problem. I created a few new endpoints in my API, but when I have a specific word 'multipress' in the request API the request is send without authorization bearer token.
When I change the word 'multipress' to 'mp' in the API endpoint URL, the request works (and bearer is in the request header.

## Security
- [ ] Is this issue security related?

## Regression
- [ ] Did this behavior work before? 
Version:

## MSAL Configuration
With AD, and API also.
```js
// Provide configuration values here.
// For Azure B2C issues, please include your policies.
```

## Reproduction steps
```js
// Provide relevant code snippets here.
// For Azure B2C issues, please include your policies.
```

## Browsers/Environment
- [x ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari
- [ ] IE
- [ ] Other (Please add browser name here)

Hyperionza:
Ditto, except mine refuses to attach a bearer to a GET  request with endpoint `https://localhost:44323/api/security/profilemenu`
If I change it to 'menu' instead of 'profilemenu' it works. Also in parallel with this request is another GET request to the same '/api/security' controller but targeting endpoint 'navigationmenu' which works fine. To make sure I wasn't imagining it, I switched the endpoint url back and forth and it consistently only failed at 'profilemenu'

using Angular 9 with  MSAL 1.3.1 and MSAL-Angular 1.0.0, browser: chrome

Happy to provide more details if required

pbranforn:
Are there any news on this topic? Sounds scary to me and we are hitting this, too!

jasonnutter:
Apologies for the delay, as I've been busy with some other work items. We're aware of this issue and will be addressing it soon. Thanks!

vitaliidasaev:
I've made own implementation of MsalInterceptor to overcome issues with protectedResourceMap.  
[...]

Issue #1695, opened 2020-05-22
unprotectedResources comming from router unprotect api endpoints


SacyrTIC:
>> Please follow the issue template below. Failure to do so will result in a delay in answering your question.

## Library
- [ ] `msal@1.x.x` or `@azure/msal@1.x.x`
- [ ] `@azure/msal-browser@2.x.x`
- [ ] `@azure/msal-angular@0.x.x`
- [x] `@azure/msal-angular@1.0.0`
- [ ] `@azure/msal-angularjs@1.x.x`

**Important**: Please fill in your exact version number above, e.g. `msal@1.1.3`.

## Framework
Angular 9

## Description
Currently, in class MsalService constructor, if unprotectedResources are given, routes comming from Router are added to unprotectedResources. In most of our proyects we have a correlation between app routes and rest api endpoints, for example:
https://myApp/profile -> https://myApi/api/profile

Taking into account routes were added to unprotected resources, lately, on isUnprotectedResource method, the validation that checks id an endpoint is on the unprotectedResources returns true, althought endpoint wasn't added to them.

![image](https://user-images.githubusercontent.com/65762234/82653832-19905f80-9c20-11ea-8f11-2c976f5c09ea.png)

## Security
Is this issue security related?

## Regression
Did this behavior work before?
Don't Know

## Configuration
Please provide your MSAL configuration options.

```js
{
...
unprotectedResources: ['assets/i18n']
...
}
```

## Reproduction steps
See description above

## Expected behavior
Routes comming from router don't unprotect api endpoints

## Browsers
Is this issue browser-specific? no

MisterPhilip:
Bumping this - is there a good workaround for this issue other than renaming routes?

jasonnutter:
@MisterPhilip Does not providing `unprotectedResources` work as a mitigation?

mbendtsen:
This took me 1½ day to figure this out :( 
I can confirm that setting `unprotectedResources` to `null` fixes my issue.

```
function MSALAngularConfigFactory(config: ConfigService): MsalAngularConfiguration {
  return {
    popUp: true,
    consentScopes: [
      config.getSettings('resourceScope'),
      ...config.getSettings('loginRequest')
    ],
    unprotectedResources: null,
    protectedResourceMap: [
      [
        config.getSettings('baseUrl'),
        [config.getSettings('resourceScope')]
      ]
    ],
    extraQueryParameters: {}
  };
```

[...]

PR #2029, opened 2020-07-24
msal-angular Update protectedResources and unprotectedResources logic

jo-arroyo:
This PR aims to address the following issues with msal-angular:
- Wildcard and glob pattern url matching requested for resource configuration: #1776 and #552 
- `unprotectedResources` causing problems for users: #1695 , #1712 , #1438 , #1966 

Changes made in this PR to address the above issues:
- Wildcard matching has been implemented for `protectedResources` in msal.service. The `*` wildcard can now be used when adding resources to the `protectedResourceMap` in the MSAL Angular Configuration.
- `unprotectedResources` has been deprecated for both the MSAL Configuration and MSAL Angular Configuration. All resources will be unprotected unless added to the `protectedResourceMap`. Logic related to `unprotectedResources` has been removed.
- Unit tests added
- Documentation updated to reflect above changes


lib/msal-angular/src/msal.service.ts
```diff
@@ -189,12 +160,24 @@ export class MsalService extends UserAgentApplication {
         }
 
         const protectedResourceMap = frameworkProtectedResourceMap && frameworkProtectedResourceMap.size ? frameworkProtectedResourceMap : new Map(this.msalAngularConfig.protectedResourceMap);
-
-        // process all protected resources and send the matched one
-        const keyForEndpoint = Array.from(protectedResourceMap.keys()).find(key => endpoint.indexOf(key) > -1);
```
jasonnutter, line 194: Applications can provide just the base url of their API in the protected resource map ([example]( https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/2036)), and any API calls with that base url will get tokens attached (since it previously did endpoint.indexOf(baseUrl)). I think we may want to fallback to indexOf  to make sure this is still support, since I don't think minimatch will match that case without the `*`.

lib/msal-angular/src/msal.service.ts
```diff
@@ -189,12 +160,24 @@ export class MsalService extends UserAgentApplication {
...
-        if (keyForEndpoint) {
-            return protectedResourceMap.get(keyForEndpoint);
-        }
+        
+        const protectedResourcesArray = Array.from(protectedResourceMap.keys());
+        const keyMatchesEndpointArray = protectedResourcesArray.filter(key => {
+            const minimatch = new Minimatch(key);
+            return minimatch.match(endpoint);
```
jasonnutter, line 167:
```suggestion
            return minimatch.match(endpoint) || endpoint.indexOf(key) > -1;
```

jasonnutter: You can test this with a test case of `http://localhost:3000` in the protected resource map and `http://localhost:3000/api` as the endpoint (which should work today).

lib/msal-angular/src/msal.service.ts
jo-arroyo: Done. Test added.

Madhufuture:
Hi, 
```MSAL version: ^1.3.2```
```MSAL angular:  @azure/msal-angular: ^1.0.0```


I have an angular application which acts as a container and imports 10 other angular apps as packages.
From my main angular app I'm calling different API's from different apps as below. I wanted to generate bearer token only to the cloudapi and not for other API's

Eg: 
```1) http://mydomain.com/serviceapp1/handler/proxy/resource/api/endpoint1```
```2) http://mydomain.com/serviceapp2/handler/proxy/resource/api/endpoint2```
```3) http://mydomain.com/serviceapp3/handler/resource/api/endpoint3```
```4) http://mydomain.com/serviceapp4/cloud/resource/api/cloudapi```

```protectedResourceMap``` configured as below
['/serviceapp*/cloud/resource/api/cloudapi',[demo.read]],
['/serviceapp*/handler/*', null]

Based on the ```protectedResourceMap``` above, my expectation was ```['/serviceapp*/handler/*', null]``` should match the pattern of the endpoint, and should not attach bearer token as the scope is null. 
But Minimatch.match is returning matched url's as 0. 

When i change the ```/serviceapp*/handler/*``` to ```/serviceapp*/handler/*/*/*/*``` it is able to match for ```endpoint1``` and ```endpoint2``` only.

Could you please suggest me an approach which actually matches the whole url instead of checking each and every url

jo-arroyo:
@Madhufuture Are you able to open an issue with one of our [pre-configured issue templates](https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/new/choose), so that our livesite engineer can prioritize this against our other efforts and others facing this issue can follow the resolution? Thank you.

Issue #1837, opened 2020-06-26
msal:loginSucces not broadcast unless handleRedirectCallback() is called


ChrisFulstow:
## Library
- `@azure/msal-angular@1.0.0`
- `msal@1.3.2`

## Framework
- `Angular 9.1.11`

## Description
The `BroadcastService` doesn't trigger an `msal:loginSuccess` event after successful login via `loginRedirect()` unless `this.msal.handleRedirectCallback()` has been registered.

## Error Message
N/A

## Security
- Not security related

## Regression
- Unknown

## MSAL Configuration

```js
function MSALConfigFactory(): Configuration {
  return {
    auth: {
      clientId: environment.msal.clientId,
      authority: 'https://login.microsoftonline.com/common/',
      validateAuthority: true,
      redirectUri: environment.msal.redirectUri,
      postLogoutRedirectUri: environment.msal.postLogoutRedirectUri,
      navigateToLoginRequestUrl: true,
    },
    cache: {
      cacheLocation: 'localStorage',
      storeAuthStateInCookie: false,
    }
  };
}

function MSALAngularConfigFactory(): MsalAngularConfiguration {
  return {
    popUp: false,
    consentScopes: environment.msal.consentScopes,
    protectedResourceMap: environment.msal.protectedResourceMap as [string, string[]][]
  };
}
```

## Reproduction steps
Here I'd expect the BroadcastService to push a `msal:loginSuccess` event after a successful login completes.
```js
ngOnInit() {
  this.broadcast.subscribe('msal:loginSuccess', () => {
    console.log('msal:loginSuccess');
  });
}

login(): void {
  this.msal.loginRedirect();
}
```

However, the event isn't pushed unless I also register `handleRedirectCallback()`, then it works fine.
```js
ngOnInit() {
  this.msal.handleRedirectCallback(() => {});
  this.broadcast.subscribe('msal:loginSuccess', () => {
    console.log('msal:loginSuccess');
  });
}
```

## Expected behavior
This is ok, but the [1.0.0 changelog](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/693e7bc442784a69b926fa765bdf8bd26e534e36/lib/msal-angular/changelog.md#100) is perhaps misleading

> Applications using Redirect methods can optionally implement the handleRedirectCallback method

[...]

PR #2032, opened 2020-07-27
Improve angular handleRedirectCallback documentation

jmckennon:
In reference to #1837

### DIFF
--- a/change/@azure-msal-angular-2020-08-11-10-47-44-redirect-docs.json
+++ b/change/@azure-msal-angular-2020-08-11-10-47-44-redirect-docs.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Improved documentation of handleRedirectCallback in msal-angular (#1837)",
+  "comment": "Improved documentation of handleRedirectCallback in msal-angular (#2032)",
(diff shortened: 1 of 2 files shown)


change/@azure-msal-angular-2020-08-11-10-47-44-redirect-docs.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
```
tnorling, line 2: This should be type none, right?

jmckennon: I considered that, should it be? Honestly not sure how we want to use these tags. I went with "patch" since it was addressing a problem from GitHub even though this is only documentation.

tnorling: The type determines how the version will be bumped. Since this doesn't touch the library, this shouldn't bump the version

I believe the types correspond to version bumps of:
<major>.<minor>.<patch>-<Prerelease??>

jmckennon: Makes sense, fixed it now.

PR #2033, opened 2020-07-27
Documentation for throttling

jmckennon:
Documentation for the throttling feature on #1907

### DIFF
--- a/lib/msal-common/docs/Throttling.md
+++ b/lib/msal-common/docs/Throttling.md
@@ -16,9 +16,9 @@ export type RequestThumbprint = {
 * Response contains a `Retry-After` header *and* does not have a 2xx HTTP status.
 * The request requires user interaction, represented in `msaljs` as `UserInteractionRequiredError`.
 
-Before making a silent call, `msaljs` checks to see if a unique `RequestThumbprint` is currently being throttled, and if so, will not make that call and return an error back to the user. Interactive calls bypass this gate as they are made to alleviate the problem. Additionally, a throttle will be cleared upon a successful interactive response from the server.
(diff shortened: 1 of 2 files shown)


lib/msal-common/docs/Throttling.md
```diff
@@ -0,0 +1,28 @@
...
+};
+```
+
+`msaljs` will throttle a specific `RequestThumbprint` if the server returns one of the following cases after making a call:
+
+* Response contains a 429 or 5xx HTTP status.
+* Response contains a `Retry-After` header *and* does not have a 2xx HTTP status.
+* The request requires user interaction, represented in `msaljs` as `UserInteractionRequiredError`.
+
+Before making a silent call, `msaljs` checks to see if a unique `RequestThumbprint` is currently being throttled, and if so, will not make that call and return an error back to the user. Interactive calls bypass this gate as they are made to alleviate the problem. Additionally, a throttle will be cleared upon a successful interactive response from the server.
```
tnorling, line 19:
```suggestion
Before making a silent call, `msaljs` checks to see if a unique `RequestThumbprint` is currently being throttled, and if so, will not make that call and return an error back to the user. Interactive calls bypass this gate as they are made to alleviate the problem. Additionally, a throttled `RequestThumbprint` will be cleared upon a successful interactive response from the server.
```

lib/msal-common/docs/Throttling.md
```diff
@@ -0,0 +1,28 @@
...
+
+The default throttling time is 60 seconds, and maximum possible throttling time is 3600 seconds.
```
tnorling, line 21: Can you elaborate on how these 2 values are used?

lib/msal-common/docs/Throttling.md
jmckennon: Good suggestion, thanks! Done

lib/msal-common/docs/Throttling.md
jmckennon: I updated this to give a bit more explanation here, good callout.

PR #2045, opened 2020-07-28
[msal-browser] Add platform-level state object to state string and read interaction type

pkanher617:
This PR addresses an issue where the handleRedirectPromise() code would be processed inside a popup before monitorWindowForHash can detect the hash value. This may happen if the opening window's execution is paused or suspended.

I have added the following:
- A new object for the browser library to populate parameters sent in the state
- A platform state string in the common library to allow platforms to set state values


lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -132,15 +135,28 @@ export class PublicClientApplication implements IPublicClientApplication {
...
         const isResponseHash = UrlString.hashContainsKnownProperties(hash);
-        const loginRequestUrl = this.browserStorage.getItem(this.browserStorage.generateCacheKey(TemporaryCacheKeys.ORIGIN_URI), CacheSchemaType.TEMPORARY) as string;
 
-        const currentUrlNormalized = UrlString.removeHashFromUrl(window.location.href);
+        // Check for interaction type
+        if (isResponseHash) {
+            // Deserialize hash fragment response parameters.
+            const hashUrlString: UrlString = new UrlString(hash);
+            const serverParams: ServerAuthorizationCodeResponse = hashUrlString.getDeserializedHash<ServerAuthorizationCodeResponse>();
+            const requestStateObj: RequestStateObject = ProtocolUtils.parseRequestState(this.browserCrypto, serverParams.state);
```
tnorling, line 148: Would it make sense to write a helper function that gets the `interactionType` from `state`? Is this something we might need to do in other places as well?

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -383,7 +399,7 @@ export class PublicClientApplication implements IPublicClientApplication {
                     ...silentRequest,
                     redirectUri: request.redirectUri,
                     prompt: PromptValue.NONE
-                });
+                }, InteractionType.SILENT);
```
tnorling, line 402: Would we ever care about differentiating between SilentFlow and SilentHandler?

pkanher617: They are both calling the same helper which creates SilentHandler, so I don't think it is necessary.

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: We can extract this in the future if we need it, but for now this is only needed for redirect flows.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -132,15 +135,28 @@ export class PublicClientApplication implements IPublicClientApplication {
...
         const { location: { hash } } = window;
-        const cachedHash = this.browserStorage.getItem(this.browserStorage.generateCacheKey(TemporaryCacheKeys.URL_HASH), CacheSchemaType.TEMPORARY) as string;
         const isResponseHash = UrlString.hashContainsKnownProperties(hash);
-        const loginRequestUrl = this.browserStorage.getItem(this.browserStorage.generateCacheKey(TemporaryCacheKeys.ORIGIN_URI), CacheSchemaType.TEMPORARY) as string;
 
-        const currentUrlNormalized = UrlString.removeHashFromUrl(window.location.href);
+        // Check for interaction type
+        if (isResponseHash) {
+            // Deserialize hash fragment response parameters.
+            const hashUrlString: UrlString = new UrlString(hash);
```
tnorling, line 146: Wouldn't you need to pass in the whole url here? Why not just call `StringUtils.queryStringToObject` on the hash instead?

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -132,15 +135,28 @@ export class PublicClientApplication implements IPublicClientApplication {
...
+            const serverParams: ServerAuthorizationCodeResponse = hashUrlString.getDeserializedHash<ServerAuthorizationCodeResponse>();
+            const requestStateObj: RequestStateObject = ProtocolUtils.parseRequestState(this.browserCrypto, serverParams.state);
+            const platformStateObj: BrowserStateObject = BrowserProtocolUtils.parseBrowserRequestState(this.browserCrypto, requestStateObj.libraryState.platformState);
+            if (StringUtils.isEmpty(platformStateObj.interactionType) || platformStateObj.interactionType !== InteractionType.REDIRECT) {
```
jasonnutter, line 150: I think `platformStateObj.interactionType !== InteractionType.REDIRECT` is the only check that is needed.

lib/msal-browser/src/utils/BrowserConstants.ts
```diff
@@ -50,3 +50,12 @@ export enum TemporaryCacheKeys {
...
     SCOPES = "scopes"
 }
+
+/**
+ * Interaction type of the API - used for state and telemetry
+ */
+export enum InteractionType {
+    REDIRECT = "redirect",
+    POPUP = "popup",
+    SILENT = "silent"
```
jasonnutter, line 60: Throttling I believe also has this concept. @jmckennon

lib/msal-browser/src/utils/BrowserProtocolUtils.ts
```diff
@@ -0,0 +1,51 @@
...
+            return "";
+        }
+
+        const stateObj: BrowserStateObject = {
+            interactionType: requestInteractionType
+        };
+
+        const stateString = JSON.stringify(stateObj);
+
+        return browserCrypto.base64Encode(stateString);
```
jasonnutter, line 31: I don't think this needs to be encoded here. What about returning the `stateObj` and then adding that as is to the library state object? That way we are only encoding/stringifying/decoding/parsing once.

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: Added in #2078

lib/msal-browser/src/utils/BrowserProtocolUtils.ts
pkanher617: I didn't want common to specify the exact object structure, which is why I made it a string. I felt like adding `any` type for platformState was too generic. We can talk about this offline.

lib/msal-browser/src/utils/BrowserProtocolUtils.ts
```diff
@@ -0,0 +1,52 @@
...
+     * @param state 
+     */
+    static extractBrowserRequestState(browserCrypto: ICrypto, state: string): BrowserStateObject {
+        if (StringUtils.isEmpty(state)) {
+            return null;
+        }
+
+        try {
+            const requestStateObj: RequestStateObject = ProtocolUtils.parseRequestState(browserCrypto, state);
+            debugger;
```
jasonnutter, line 46: `debugger`;

lib/msal-browser/src/utils/BrowserProtocolUtils.ts
```diff
@@ -0,0 +1,52 @@
...
+};
+
+export class BrowserProtocolUtils {
+
+    /**
+     * Generates the state value used by the platform library
+     * @param browserCrypto 
+     * @param requestInteractionType 
+     */
+    static generateBrowserRequestState(browserCrypto: ICrypto, requestInteractionType: InteractionType): string {
```
jasonnutter, line 20: I believe this can be deleted.

lib/msal-browser/src/utils/BrowserUtils.ts
```diff
@@ -60,6 +60,12 @@ export class BrowserUtils {
         return window.location.href.split("?")[0].split("#")[0];
     }
 
+    static getHomepage(): string {
+        const currentUrl = new UrlString(window.location.href);
+        const urlComponents = currentUrl.getUrlComponents();
+        return urlComponents.Protocol + "//" + urlComponents.HostNameAndPort + "/";
```
jasonnutter, line 66: Nitpick: this can be a string template.

lib/msal-browser/src/utils/BrowserProtocolUtils.ts
pkanher617: done.

lib/msal-browser/src/utils/BrowserProtocolUtils.ts
pkanher617: done.

lib/msal-browser/src/utils/BrowserUtils.ts
pkanher617: done.

Issue #2019, opened 2020-07-23
login error 'state_mismatch: State mismatch error' with msal-browser


flytiny:
>> Please follow the issue template below. Failure to do so will result in a delay in answering your question.

## Library
- [ ] `msal@1.x.x` or `@azure/msal@1.x.x`
- [x] `@azure/msal-browser@2.x.x`
- [ ] `@azure/msal-angular@0.x.x`
- [ ] `@azure/msal-angular@1.x.x`
- [ ] `@azure/msal-angularjs@1.x.x`

**Important**: Please fill in your exact version number above, e.g. `msal@1.1.3`.

## Framework
@azure/msal-browser:2.0.0
## Description
We are a first-party app (I don't know if this information is important). After the user logs in and jumps back to the page, the AuthenticationResult that msalObj gets is empty, then the page will automatically refresh again soon, and this error appears in the console.
I saw [this issue](https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/1846) reported by others, but it seems to be fixed already.

## Error Message
![image](https://user-images.githubusercontent.com/6790396/88282512-4ba36800-cd1c-11ea-88bb-a0260a7a7212.png)

## Security
- [ ] Is this issue security related?

## Regression
- [ ] Did this behavior work before? 
Version:

## MSAL Configuration

```js
const msalConfig: Configuration = {
  auth: {
    clientId: "****",
    authority: "https://login.microsoftonline.com/common",
    redirectUri: "****",
  },
  cache: {
    cacheLocation: "localStorage",
    storeAuthStateInCookie: true,
  },
};
```
other code

```js
const myMSALObj = new PublicClientApplication(msalConfig);
let account: any = null;

const loginRedirectRequest = {
  scopes: config.authConfig.loginRequestScopes,
  redirectStartPage: window.location.href
};

myMSALObj.handleRedirectPromise().then((resp: AuthenticationResult | null) => {
  handleResponse(resp);
}).catch(console.error);


function handleResponse(response: AuthenticationResult | null) {
  if (response !== null) {
    account = response.account;
  } else {
    account = getAccount();
  }
}

function getAccount(): AccountInfo | null {
  const currentAccounts = myMSALObj.getAllAccounts();
  if (currentAccounts === null) {
    console.log("No accounts detected");
    return null;
  }
  if (currentAccounts.length > 1) {
    // Add choose account code here
    console.log("Multiple accounts detected, need to add choose account code.");
[...]

PR #2049, opened 2020-07-28
[msal-browser] Fix state mismatch error due to encoding

jo-arroyo:
It appears that the state that is returned from the server is being mutated by their base64 encoding mechanism, which is different to our encoding mechanism, causing a state mismatch error. This PR attempts to provide a temporary solution by decoding the state strings before comparing the response state and cached state. More investigation and communication with the server team may be necessary to figure out why this mutation occurs.

This PR addresses issue #2019 .


mzguimaraes:
Hi all, it appears that this PR will fix an issue my team is currently having with MSAL.  We're not sure what sort of release cadence you are planning for with 2.0.  What sort of timeline should we expect for when this fix will be available through NPM?

Issue #2053, opened 2020-07-29
Fix function signatures for getAllAccounts and getAccountByUsername


jasonnutter:
Please follow the issue template below. 

## Library
- [ ] `msal@1.x.x` or `@azure/msal@1.x.x`
- [x] `@azure/msal-browser@2.x.x`
- [ ] `@azure/msal-angular@0.x.x`
- [ ] `@azure/msal-angular@1.x.x`
- [ ] `@azure/msal-angularjs@1.x.x`

## Upcoming Library
- [ ] `@azure/msal-node@1.x`

## Source
- [x] Internal (MSAL.js team)
- [ ] Customer request

## Description

Today, the signatures for `PublicClientApplication.getAllAccounts` and `PublicClientApplication.getAccountByUsername` do not indicate that these methods can return `null`.

## Priority

Medium

## Acceptance Criteria

1. `PublicClientApplication.getAllAccounts` and `PublicClientApplication.getAccountByUsername` return types are updated to indicate they can return `null`.

aappddeevv:
Wow! I just hit this 1 hour ago but you already had this issue created. Thanks!

PR #2059, opened 2020-07-29
Fix function signatures for getAllAccounts and getAccountByUsername

tnorling:
Addresses #2053


lib/msal-browser/test/app/PublicClientApplication.spec.ts
```diff
@@ -1368,10 +1368,10 @@ describe("PublicClientApplication.ts Class Unit Tests", () => {
             expect(account).to.be.length(2);
         });
 
-        it("getAllAccounts returns null if no accounts signed in", () => {
+        it("getAllAccounts returns empty array if no accounts signed in", () => {
             window.sessionStorage.clear();
-            const account = pca.getAllAccounts();
-            expect(account).to.be.null;
+            const accounts = pca.getAllAccounts();
+            expect(accounts).to.deep.eq([]);
```
sameerag, line 1374: Can we add a test case where user passes `null` for `getAccountByUserName()`? I suppose that returns `null`?

tnorling: Done

aappddeevv:
Thanks for putting this in so quickly!

Issue #2054, opened 2020-07-29
Allow request objects for loginRedirect and loginPopup to be optional


jasonnutter:
Please follow the issue template below. 

## Library
- [ ] `msal@1.x.x` or `@azure/msal@1.x.x`
- [x] `@azure/msal-browser@2.x.x`
- [ ] `@azure/msal-angular@0.x.x`
- [ ] `@azure/msal-angular@1.x.x`
- [ ] `@azure/msal-angularjs@1.x.x`

## Upcoming Library
- [ ] `@azure/msal-node@1.x`

## Source
- [x] Internal (MSAL.js team)
- [ ] Customer request

## Description

The request objects passed to `PublicClientApplication.loginRedirect` and `PublicClientApplication.loginPopup` should be optional since scopes should not be required.

## Priority

Medium

## Acceptance Criteria

1. The request objects passed to `PublicClientApplication.loginRedirect` and `PublicClientApplication.loginPopup` should not be required.

PR #2061, opened 2020-07-29
Make request object optional for login methods

tnorling:
Addresses #2054


lib/msal-browser/src/utils/BrowserConstants.ts
```diff
@@ -50,3 +50,7 @@ export enum TemporaryCacheKeys {
     REQUEST_PARAMS = "request.params",
     SCOPES = "scopes"
 }
+
+export const DEFAULT_REQUEST = {
```
pkanher617, line 54: We should add a typing to this request, can be `AuthorizationUrlRequest`.

lib/msal-browser/src/utils/BrowserConstants.ts
```diff
@@ -50,3 +50,7 @@ export enum TemporaryCacheKeys {
...
+    scopes: ["openid", "profile"]
```
pkanher617, line 55: Use the constant vars instead of strings here.

pkanher617: `Constants.OPENID_SCOPE, Constants.PROFILE_SCOPE`

tnorling: Yep good call! Done.

lib/msal-browser/src/utils/BrowserConstants.ts
```diff
@@ -50,3 +52,7 @@ export enum TemporaryCacheKeys {
     REQUEST_PARAMS = "request.params",
     SCOPES = "scopes"
 }
+
+export const DEFAULT_REQUEST: AuthorizationUrlRequest = {
+    scopes: [Constants.OPENID_SCOPE, Constants.PROFILE_SCOPE]
```
jasonnutter, line 57: If you pass empty array will these scopes get added automatically?

jasonnutter: We could have the logic downstream also use this same constant.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -37,7 +37,7 @@ import { RedirectHandler } from "../interaction_handler/RedirectHandler";
 import { PopupHandler } from "../interaction_handler/PopupHandler";
 import { SilentHandler } from "../interaction_handler/SilentHandler";
 import { BrowserAuthError } from "../error/BrowserAuthError";
-import { BrowserConstants, TemporaryCacheKeys } from "../utils/BrowserConstants";
+import { BrowserConstants, TemporaryCacheKeys, DEFAULT_REQUEST } from "../utils/BrowserConstants";
 import { BrowserUtils } from "../utils/BrowserUtils";
 import { version } from "../../package.json";
 import { IPublicClientApplication } from "./IPublicClientApplication";
```
jasonnutter, line 43: Can you make sure `IPublicClientApplication` is also updated to show these are optional?

lib/msal-browser/src/utils/BrowserConstants.ts
sameerag: I had the exact q. What is the conclusion?

sameerag: Feel free to resolve if addressed.

tnorling: @sameerag Yes they do get added downstream, PR was updated to have the downstream addition utilize this constant as well

PR #2068, opened 2020-07-30
[msal-node] storage optimization

sameerag:
This PR revamps the storage strategy in MSAL Node:

* Introduces a new `cache` data structure which is a Key-Value store :`Record<string, object>`
* All `Storage` APIs now refer to this `cache` for fetching/reading/writing data
* Every time the `cache` changes (write/remove), `emitChange()` is registered
* Serialized persistence happens on demand, and conversion of this new `cache` to `inMemoryCache` and vice versa happens as needed
* Unit tests for `Storage` don't exist, added them
* Sample tested manually

**Followup**

* Similar changes will be to be followed up for `msal-browser`
* The util APIs may differ, as `msal-browser` stores the values as `string` instead of `object`


lib/msal-node/src/cache/Storage.ts
```diff
@@ -49,80 +39,118 @@ export class Storage extends CacheManager {
...
+        }
+
+        return inMemoryCache;
+    }
+
+    /**
+     * converts inMemoryCache to CacheKVStore
+     * @param inMemoryCache
+     */
+    inMemoryCacheToCache(inMemoryCache: InMemoryCache): CacheKVStore {
```
sameerag, line 79: Note: I need to test this, verify if this over writes existing values with same keys. Will test and change as needed. `spread` operator ideally should, but since we are pushing the data to `Record<string, object>`, need to test all use cases.

lib/msal-node/src/cache/Storage.ts
```diff
@@ -49,80 +39,118 @@ export class Storage extends CacheManager {
...
+
+        let inMemoryCache: InMemoryCache = {
+            accounts: {},
+            idTokens: {},
+            accessTokens: {},
+            refreshTokens: {},
+            appMetadata: {},
+        };
+
+        for (let key in cache) {
```
jmckennon, line 56: Should this be a switch statement?

lib/msal-node/src/cache/Storage.ts
```diff
@@ -49,80 +39,118 @@ export class Storage extends CacheManager {
...
      * @param value
      * @param type
-     * @param inMemory
      */
-    setItem(
-        key: string,
-        value: string | object,
-        type?: string
-    ): void {
+    setItem(key: string, value: ValidCacheType, type?: string) {
```
jmckennon, line 144: If `type` is not provided, will this store and return a string in `getItem`?

lib/msal-react-native-android-poc/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@azuread/msal-react-native-android-poc",
-	"version": "1.0.0",
```
jmckennon, line 3: Was this intentionally committed?

lib/msal-node/src/cache/Storage.ts
sameerag: I don't think we can equate a value: `cache[key` which will be the `switch(value)` to `instanceof <some object>`. Hence the choice of `if`.

lib/msal-node/src/cache/Storage.ts
sameerag: The idea is to get rid of type altogether. I kept it until tests make it obsolete.

lib/msal-react-native-android-poc/package-lock.json
sameerag: nope. reverted.

lib/msal-common/src/utils/Constants.ts
```diff
@@ -229,9 +229,13 @@ export enum CredentialType {
 export enum CacheSchemaType {
```
tnorling, line 229: This will be removed in the browser PR, right?

lib/msal-node/test/cache/Storage.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+                client_info: 'eyJ1aWQiOiJ1aWQxIiwgInV0aWQiOiJ1dGlkMSJ9',
+            },
+        };
+        let account = new AccountEntity();
+        account = CacheManager.toObject(account, newMockAccount);
+
+        nodeStorage.setItem(accountKey, account);
+        const fetchedAccount = nodeStorage.getItem(accountKey);
+
+        expect(account).toEqual(fetchedAccount);
```
tnorling, line 100: I think it would be good to also test instanceOf here as well since we're expecting getItem to return specific types

lib/msal-node/test/cache/Storage.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+        expect(account).toBeInstanceOf(AccountEntity);
+        expect(account.clientInfo).toBe(
+            'eyJ1aWQiOiJ1aWQiLCAidXRpZCI6InV0aWQifQ=='
+        );
+
+        const newInMemoryCache = nodeStorage.getInMemoryCache();
+        expect(newInMemoryCache.accounts[accountKey]).toEqual(account);
+    });
+
+    it('setItem() and getItem() tests', () => {
```
tnorling, line 78: Can we also specify in the titles which part they are testing? e.g. Account, AccessToken, etc?

lib/msal-node/test/cache/Storage.spec.ts
sameerag: Added

lib/msal-node/test/cache/Storage.spec.ts
sameerag: well I distributed the data types to be random, will add.

PR #2072, opened 2020-07-30
Merge previous PR changes and add token story

AndrewCraswell:
Merges all the work for msal-react hackathon project into the hackathon branch. Also adds an Acquire Token storybook entry.


experimental/msal-react/.prettierignore
```diff
@@ -0,0 +1 @@
+src/index.ts
```
AndrewCraswell, line 1: Prettier was having issues with `export type`, which for some reason wasn't being resolved with an ignore statement.

experimental/msal-react/src/MsalAuthentication.tsx
```diff
@@ -0,0 +1,74 @@
...
+        () => ({
+            error,
+            msal,
+        }),
+        [error, msal]
+    );
+}
+
+export const MsalAuthentication: React.FunctionComponent<IMsalAuthenticationProps> = props => {
+    const { username, loginHandler, children } = props;
```
jasonnutter, line 60: TODO: change to full account object instead of just username.

jasonnutter: To be consistent with MSAL Angular, add props for `interactionType` and an instance of `PopupRequest | RedirectRequest`: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/64a4f9e868e63346dfd711dec717abe7fd14d949/samples/msal-angular-v2-samples/angular10-browser-sample/src/app/msal/msal.guard.config.ts#L8

experimental/msal-react/src/Templates.tsx
```diff
@@ -0,0 +1,39 @@
...
+            <React.Fragment>
+                {getChildrenOrFunction(children, context)}
+            </React.Fragment>
+        );
+    }
+    return null;
+};
+
+export const AuthenticatedTemplate: React.FunctionComponent<IMsalTemplateProps> = props => {
+    const { children, username } = props;
```
jasonnutter, line 27: TODO: change to full account object instead of username.

jasonnutter: Make specific account optional?

AndrewCraswell:
Great to see some additional tasks tracked here. A couple apps under our team's purview is looking to migrate to this new React pattern relatively quickly. I'm looking to take a weekend in the coming month to iterate over it some more and test in our app. Would love to incorporate some of this feedback so we can get something basic working in production and test the new Context API pattern out in the wild.

experimental/msal-react/src/index.ts
```diff
@@ -0,0 +1,22 @@
+// TODO: How do we get the token before executing an API call in pure TypeScript, outside of a React component context?
```
jasonnutter, line 1: @AndrewCraswell Can you clarify this please?

experimental/msal-react/src/index.ts
```diff
@@ -0,0 +1,22 @@
...
+// TODO: How do we raise the `error` state into the MsalContext, where changes will allow all subscribed components to update?
+// TODO: How do we represent the current state of the authentication process (Authenticated, InProgress, IsError, Unauthenticated)?
```
jasonnutter, line 3: The `AuthenticationState` enum from react-aad-msal may not work the same way in v2, given multiple account support. We will expose an event API to know when an auth event is in progress, and the isAuthenticated hook can answer the question about whether or not you are authenticated.

jasonnutter: Auth errors are per request, so I'm not sure we want to include a global singleton error state.

experimental/msal-react/src/index.ts
```diff
@@ -0,0 +1,22 @@
+// TODO: How do we get the token before executing an API call in pure TypeScript, outside of a React component context?
+// TODO: How do we raise the `error` state into the MsalContext, where changes will allow all subscribed components to update?
```
jasonnutter, line 2: If I'm understanding correct, event API should address this?

AndrewCraswell: This is exactly right. Today we have to wrap the MSAL API to intercept values and add to the Context Provider. When the event API goes live a wrapper won't be necessary, so the consumer just needs to call the MSAL functions as usual.

I also realized that while the event API isn't necessarily a blocker to continue building out the Alpha React MSAL library. I've started experimenting with creating a temp wrapper around MSAL that raises events when interacting with the MSAL API. For now this works, and so long as we agree early on a contract for which events MSAL will dispatch, then the demo can be built out accordingly. When the Event API goes live we just need to remove the temp wrapper and access MSAL API directly.

This was the approach I had been planning to use short-term for our apps in production.

tnorling:
Merging this work into our repo & will start working through some of the TODOs

experimental/msal-react/src/index.ts
AndrewCraswell: Clarified in comment below

experimental/msal-react/src/index.ts
AndrewCraswell: Should be fine so long as we don't create scenarios where the `forceLogin` prop triggers an endless loop.

PR #2073, opened 2020-07-30
Add SSR support to msal-browser

### LINKED ISSUE #1892: Angular SSR with  MSAL.JS CDN - working with types

## Library

## Description
Hello, I am going to attempt to get Angular Universal working with MSAL.JS.  I will have to use the CDN version to make this work as I already tried with non-CDN and there were too many issues.

Do you have the types I could install?

I am declaring the global MSAL variable like this.
declare var msal: any;

However, you can see in this screenshot I would need to have any types all over the place since the MSAL namespace is not recongized.

image

Examples: 
Angular 9/10

### LINKED ISSUE #2052: ReferenceError: window is not defined break SSR

## Library

## Framework
n/a
## Description
When trying to SSR the code, `window` not being defined causes the SSR to fail

## Error Message
```
ReferenceError: window is not defined
    at Function.BrowserUtils.getBrowserNetworkClient (PATH_TO_PROJECT/node_modules/@azure/msal-browser/src/utils/BrowserUtils.ts:67:9)
```
## Security

## Regression
- [x] Did this behavior work before? 
Version:
1.3
## MSAL Configuration

n/a

## Reproduction steps
import MSAL and try SSR'ing

## Expected behavior
the code should be more sensitive to `window` not being defined

## Browsers/Environment
node. Yes, I know this a browser-based package, trying to SSR for the browser! Perhaps delaying the call that checks for `window` (or probably `fetch`) until runtime would help.

Thanks!

jasonnutter:
This PR allows for `PublicClientApplication` to be instantiated in an environment where `window` is not available (e.g. server-side rendering for React, Angular, etc.). Note, this will not allow for login or token operations to happen server-side, as those APIs will now throw specific errors if they are invoked in such an environment.

TODO:
- [x] Test in real React SSR environment
- [x] Test in real Angular SSR environment
- [x] Ensure all code that could be run during module resolution is SSR-safe.


lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -578,6 +592,18 @@ export class PublicClientApplication implements IPublicClientApplication {
         return this.initializeAuthorizationRequest(request);
     }
 
+    /**
+     * Helper to validate app environment before making a silent request
```
jasonnutter, line 596: I guess this comment is incorrect as it validates silent and interactive requests.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -260,6 +271,8 @@ export class PublicClientApplication implements IPublicClientApplication {
      * @returns {Promise.<AuthenticationResult>} - a promise that is fulfilled when this function has completed, or rejected if an error was raised. Returns the {@link AuthResponse} object
      */
     async acquireTokenPopup(request: PopupRequest): Promise<AuthenticationResult> {
+        this.preflightBrowserEnvironmentCheck();
```
jasonnutter, line 274: This is done outside the `try` because we don't want the `this.browserStorage.cleanRequest()` call to run in non-browser environments.

jasonnutter: @pkanher617 Is it okay if this check (which checks for browser environment and if we're in an iframe) is done outside catch, specifically for the iframe request? My understanding is the `cleanRequest` call is a consideration for the other errors that are potentially thown in the `preflightInteractiveRequest`, but is not needed for the iframe check.

tnorling (approved):

sameerag:
One comment: How is SSR supported when we are blocking non-browser environments? The instantiation of `PublicClientApplication` is done with an empty object. So does this mean SSR has a PCA instance -> does not access window -> When client component is mounted, the `window` is initialized?

may be a  naive q from me, just curious how this PR helps SSR and for what operations!

sameerag (approved):
One q: approved otherwise

mbrevda:
[seems so](https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/2073/files#diff-4bc968f7eb828bab330109a01e72c95bR103)

mlynam:
I think the main issue here is you are declaring a configuration setting at global scope in your config: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/84a5f5aed0ad1626e6d918fb2ef0393837aac3c6/lib/msal-browser/src/config/Configuration.ts#L100-L108. This causes browser dependent code to fire. For my use case, it would be better to delay the evaluation of the window conditioned code until I actually invoke some msal code. I can conditionally stub a component during SSR rendering if this was the case.

jasonnutter:
> I think the main issue here is you are declaring a configuration setting at global scope in your config:
> 
> https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/84a5f5aed0ad1626e6d918fb2ef0393837aac3c6/lib/msal-browser/src/config/Configuration.ts#L100-L108
> 
> This causes browser dependent code to fire _when the module is resolving._ For my use case, it would be better to delay the evaluation of the window conditioned code until I actually invoke some msal code. I can conditionally stub a component during SSR rendering if this was the case.

Good callout, I'll make sure that gets addressed.

mbrevda:
Was this merged?

jrmcdona:
Oh yeah, I need this one!

mlynam:
I also need this fixed. Is there any work the community can help with to move this along?

tnorling:
This was closed unintentionally by our bot. Apologies for the delay, we do intend to pick this back up soon.

mbrevda:
@mlynam if it's any help, I mocked out msal during ssr pending this PR

mlynam:
@mbrevda thanks man I did too. Hopefully it lands soon!

jrmcdona:
@mbrevda @mlynam  did you mock out the types and using CDN to get it to work?

mbrevda:
There is no need to have MSAL enabled during SSR. Here is how I stubed it:

```js
// temporoarly override module laoding logic to mock @azure/msal-browser
// workaround for https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/2052
var Module = require('module')
var originalLoad = Module._load
Module._load = (request, parent, isMain) => {
  if (request === '@azure/msal-browser') {
    return {
      PublicClientApplication: class PublicClientApplication {
        getAllAccounts() {}
      }
    }
  }
  delete global.window
  return originalLoad(request, parent, isMain)
}

// SSR code...
```

jmckennon (approved):
Code looks good, probably a bit to late for this feedback but it feels a bit unfortunate to have this SSR support interwoven with what we would consider to be standard browser flow/usage, feels like they could live in parallel somehow rather than being mixed together, if that makes sense

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -65,25 +68,34 @@ export abstract class ClientApplication {
...
-        // Set the configuration.
-        this.config = buildConfiguration(configuration);
+        /*
+         * If loaded in an environment where window is not available,
+         * set internal flag to false so that further requests fail.
+         * This is to support server-side rendering environments.
+         */
+        this.isBrowserEnvironment = typeof window !== "undefined";
+
+        if (this.isBrowserEnvironment) {
```
DarylThayil, line 78: might be cleaner to do the exit clause insteadof nesting

if (!this.isBrowserEnv) {
return
}

.... do normal stuff

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -95,7 +107,7 @@ export abstract class ClientApplication {
      * @returns {Promise.<AuthenticationResult | null>} token response or null. If the return value is null, then no auth redirect was detected.
      */
     async handleRedirectPromise(): Promise<AuthenticationResult | null> {
-        return this.handleRedirectResponse();
+        return this.isBrowserEnvironment ? this.handleRedirectResponse() : null;
```
DarylThayil, line 110: any chance this breaks client code that expects this to be a promise by returning null ?

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -259,6 +272,13 @@ export abstract class ClientApplication {
      * @returns {Promise.<AuthenticationResult>} - a promise that is fulfilled when this function has completed, or rejected if an error was raised. Returns the {@link AuthResponse} object
      */
     acquireTokenPopup(request: PopupRequest): Promise<AuthenticationResult> {
+        try {
```
DarylThayil, line 275: Is this related to SSR?

lib/msal-browser/src/app/ClientApplication.ts
tnorling: Since the function is declared as async this will resolve null. This essentially behaves the same as if there was no hash in the url

lib/msal-browser/src/app/ClientApplication.ts
tnorling: Yes, the `preflightBrowserEnvironmentCheck` will throw an error if the window object is not defined. We don't want people to call any APIs in a non-browser environment

lib/msal-browser/src/app/ClientApplication.ts
tnorling: I like this suggestion, will update

jasonnutter:
Looks good, thanks for taking over @tnorling!

jrmcdona:
Will there be much change in the implementation on the consumers end?

tnorling:
@jrmcdona No, once the next release goes out, importing msal-browser and initializing PublicClientApplication should no longer throw an error in SSR environments, everything else should stay the same. You will just need to ensure that your app is not calling any of the token acquisition APIs (i.e. loginRedirect, loginPopup, etc.) in your server render path. If, for some reason, that's not possible, you can catch the new error that this PR introduces.

kostenickj:
Will this be released into @azure/msal-browser soon? SSR is a deal breaker for our app and having to edit the code manually on every update has been tedious...

tnorling:
@kostenickj Yes, this will be included in version 2.2.1 which will be released very soon.

mbrevda:
Thanks!

maugerc:
hello @tnorling, you indicate the 2.2.1 version of msal-browser for this fix, but it's already release one year ago, you talk to 2.21.0 version, when will it release ?

Issue #2055, opened 2020-07-29
handleRedirectPromise throws a state mismatch error when it is invoked multiple times


jasonnutter:
Please follow the issue template below. 

## Library
- [ ] `msal@1.x.x` or `@azure/msal@1.x.x`
- [x] `@azure/msal-browser@2.x.x`
- [ ] `@azure/msal-angular@0.x.x`
- [ ] `@azure/msal-angular@1.x.x`
- [ ] `@azure/msal-angularjs@1.x.x`

## Upcoming Library
- [ ] `@azure/msal-node@1.x`

## Source
- [x] Internal (MSAL.js team)
- [ ] Customer request

## Description

If an application invokes `handleRedirectPromise` multiple times on the same page, a state mismatch error is thrown. This causes problems in environments like React, where a component could be rendered multiple times, or multiple components could invoke this promise. We should ensure that no errors are thrown when this happens, at a minumum.

## Priority

Medium

## Acceptance Criteria

1. When `handleRedirectPromise` is invoked multiple times on the same page, no errors are thrown.

PR #2078, opened 2020-07-31
handleRedirect fixes

tnorling:
Builds on #2045, refactors and fixes the following bugs in the redirect flow:

- #2055 State mismatch error when calling `handleRedirectPromise` multiple times
- `currentUrl` and `loginRequestUrl` being evaluated as not equal if one has a trailing slash and the other does not
- When `loginRequestUrl` is not in the cache, msal redirects to the homepage but would not process the hash

### DIFF
--- a/lib/msal-browser/src/app/PublicClientApplication.ts
+++ b/lib/msal-browser/src/app/PublicClientApplication.ts
@@ -130,54 +130,62 @@ export class PublicClientApplication implements IPublicClientApplication {
         return this.handleRedirectResponse();
     }
 
+    private getRedirectResponseHash(): string {
+        // Get current location hash from window or cache.
+        const { location: { hash } } = window;
+        const isResponseHash = UrlString.hashContainsKnownProperties(hash);
+        const cachedHash = this.browserStorage.getItem(this.browserStorage.generateCacheKey(TemporaryCacheKeys.URL_HASH), CacheSchemaType.TEMPORARY) as string;
+        this.browserStorage.removeItem(this.browserStorage.generateCacheKey(TemporaryCacheKeys.URL_HASH));
+
+        const responseHash = isResponseHash ? hash : cachedHash;
+        if (responseHash) {
+            // Deserialize hash fragment response parameters.
+            const serverParams: ServerAuthorizationCodeResponse = StringUtils.queryStringToObject<ServerAuthorizationCodeResponse>(responseHash);
+            const requestStateObj: RequestStateObject = ProtocolUtils.parseRequestState(this.browserCrypto, serverParams.state);
+            const platformStateObj: BrowserStateObject = BrowserProtocolUtils.parseBrowserRequestState(this.browserCrypto, requestStateObj.libraryState.platformState);
+            if (platformStateObj.interactionType !== InteractionType.REDIRECT) {
+                return null;
+            } else {
+                BrowserUtils.clearHash();
+                return responseHash;
+            }
+        }      
+
+        return null;
+    }
+
     /**
      * Checks if navigateToLoginRequestUrl is set, and:
      * - if true, performs logic to cache and navigate
      * - if false, handles hash string and parses response
      */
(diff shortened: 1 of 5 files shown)


lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -130,54 +130,62 @@ export class PublicClientApplication implements IPublicClientApplication {
...
+        const responseHash = isResponseHash ? hash : cachedHash;
+        if (responseHash) {
             // Deserialize hash fragment response parameters.
-            const hashUrlString: UrlString = new UrlString(hash);
-            const serverParams: ServerAuthorizationCodeResponse = hashUrlString.getDeserializedHash<ServerAuthorizationCodeResponse>();
+            const serverParams: ServerAuthorizationCodeResponse = StringUtils.queryStringToObject<ServerAuthorizationCodeResponse>(responseHash);
             const requestStateObj: RequestStateObject = ProtocolUtils.parseRequestState(this.browserCrypto, serverParams.state);
             const platformStateObj: BrowserStateObject = BrowserProtocolUtils.parseBrowserRequestState(this.browserCrypto, requestStateObj.libraryState.platformState);
-            if (StringUtils.isEmpty(platformStateObj.interactionType) || platformStateObj.interactionType !== InteractionType.REDIRECT) {
+            if (platformStateObj.interactionType !== InteractionType.REDIRECT) {
```
pkanher617, line 146: Not sure why the empty check was removed, should probably add it back.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -130,54 +130,62 @@ export class PublicClientApplication implements IPublicClientApplication {
...
+            return null;
         }
 
-        const cachedHash = this.browserStorage.getItem(this.browserStorage.generateCacheKey(TemporaryCacheKeys.URL_HASH), CacheSchemaType.TEMPORARY) as string;
         const loginRequestUrl = this.browserStorage.getItem(this.browserStorage.generateCacheKey(TemporaryCacheKeys.ORIGIN_URI), CacheSchemaType.TEMPORARY) as string;
         const loginRequestUrlNormalized = UrlString.removeHashFromUrl(loginRequestUrl || "");
         const currentUrlNormalized = UrlString.removeHashFromUrl(window.location.href);
 
-        if (loginRequestUrlNormalized === currentUrlNormalized && this.config.auth.navigateToLoginRequestUrl) {
+        if (loginRequestUrlNormalized === currentUrlNormalized || !this.config.auth.navigateToLoginRequestUrl) {
```
pkanher617, line 172: You don't need the second check if you're going to check for truthiness inside the if. One reason I made these ifs top level is to make the logic easier to follow. I'd prefer to revisit this together.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -130,54 +130,62 @@ export class PublicClientApplication implements IPublicClientApplication {
...
-            return this.handleHash(isResponseHash ? hash : cachedHash);
-        }
-
-        if (isResponseHash && !BrowserUtils.isInIframe()) {
+            if (this.config.auth.navigateToLoginRequestUrl && loginRequestUrl.indexOf("#") > -1) {
+                // Replace current hash with non-msal hash, if present
+                BrowserUtils.replaceHash(loginRequestUrl);
+            }
+            return this.handleHash(responseHash);
+        } else if (!BrowserUtils.isInIframe()) {
```
pkanher617, line 179: Should this be an `else if`?

lib/msal-browser/src/app/PublicClientApplication.ts
tnorling: It's redundant isn't it? `"" !== "redirect"` and `null !== "redirect"`

lib/msal-browser/src/app/PublicClientApplication.ts
tnorling: Reverted to be top level

PR #2082, opened 2020-07-31
Add Beachball to repo

pkanher617:


change/@azure-msal-browser-2020-07-31-16-28-44-add-beachball-config.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "none",
+  "comment": "Adding beachball to the build steps",
+  "packageName": "@azure/msal-browser",
+  "email": "prithvi.kanherkar@gmail.com",
```
sameerag, line 5: Is this intentional? Should we have a team email that we can add here if really needed?

jasonnutter:
Can you please provide some background on how this works? Thanks!

change/@azure-msal-browser-2020-07-31-16-28-44-add-beachball-config.json
pkanher617: I think this is based on the github account making the commit. I need to change my primary email on my account I think?

PR #2096, opened 2020-08-04
Update azure-pipelines-extensions.yml

sangonzal:
Update triggers to exclude extensions pipelines

### DIFF
--- a/azure-pipelines-extensions.yml
+++ b/azure-pipelines-extensions.yml
@@ -21,3 +21,5 @@ pr:
         - extensions/*
       exclude: 
         - '*'
+trigger:
+    - none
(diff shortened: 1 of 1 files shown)


azure-pipelines-extensions.yml
```diff
@@ -13,6 +13,11 @@ jobs:
...
+      include:
+        - '*'
     paths:
-        include:
-            - extensions/*
\ No newline at end of file
+      include:
+        - extensions/*
+      exclude: 
+        - '*'
```
pkanher617, line 23: add a newline

sangonzal: Added newline at end of file. For some reason it doesn't show Files Changed tab, but if you look at raw file you'll see it.

Issue #1845, opened 2020-06-29
[Angular v10][Version 1.0.0] Library build not working(ModuleWithProviders<T>)


sebastiannordby:
## Library
- [X] `@azure/msal-angular@1.0.0`
- [ ] `msal@1.x.x` or `@azure/msal@1.x.x`
- [ ] `@azure/msal-browser@2.x.x`
- [ ] `@azure/msal-angular@0.x.x`
- [ ] `@azure/msal-angularjs@1.x.x`

## Description
When building a library this error occurs, [this is because of deprecation in version 10 of Angular.](https://angular.io/guide/migration-module-with-providers)

## Error Message
ERROR: node_modules/@azure/msal-angular/src/msal.module.d.ts:5:86 - error TS2314: Generic type 'ModuleWithProviders<T>' requires 1 type argument(s).

## Regression
- [X ] Did this behavior work before? 
Version: < 10

**Edit(Temp fix):**
Adding the following property inside "compilerOptions" in your tsconfig.json will make your library buildable.
"skipLibCheck": true,

jasonnutter:
@PapaBigBear Thanks for the information. We have not had a chance to test and update MSAL Angular for compatibility with Angular v10, but we'll make sure this is fixed.

devemanager:
### How to fixed!
ERROR in node_modules/@azure/msal-angular/src/msal.module.d.ts:5:86 - error TS2314: Generic type 'ModuleWithProviders<T>' requires 1 type argument(s).     

static forRoot(config: Configuration, angularConfig?: MsalAngularConfiguration): ModuleWithProviders;

danielzach:
It's been a while since Angular 10 was released. The abovementioned build error is actually the only one after updating to Angular 10 in my case, so this is a blocker. Any update on this?

PR #2102, opened 2020-08-05
fix: Added generic argument to ModuleWithProviders

jpleclerc:
Angular 10 now requires the generic arguments for ModuleWithProviders otherwise it wont compile.


jpleclerc:
#1845

jpleclerc:
Nobody is looking at it ?

DHFW:
Please fix this, this is holding us back from upgrading to Angular 10.

Issue #1766, opened 2020-06-08
Reading roles/groups from token


graphicsxp:
## Library
- [ ] `@azure/msal-angular@1.x.x`

## Description
How to read roles or groups from token returned by msal ? 
I'm doing the following : `this.msalService.getAccount().idTokenClaims["roles"]; `  but in returns a string as the defined type by StringDict. However at runtime when I debug the application it is not a string, it is an array of strings (the roles), which is exactly what I need. 
But typescript won't compile if I try to cast it since StringDict returns a string. What is the correct way to get roles then ?

derisen:
@graphicsxp could you take a look at the [sample here](https://github.com/Azure-Samples/ms-identity-javascript-angular-spa-dotnetcore-webapi-roles-groups/tree/derisen/chapter1). There, we demonstrate how to implement a `RoleGuard` service.

graphicsxp:
Yes indeed. I still find it confusing. The sample uses the .include function of string to check if the role is part of the route data. However it only works because String and Array share an `include` method in their signature.
For example, what if we want to allow multiple roles to access a routes. Or just one role but assuming in the tokenClaims there are many roles. The include method wouldn't work there and we'd need to use  `find` or `filter`, which don't exist on the String type.
any insights ?

graphicsxp:
ok, I will use one of these workarounds for the time being until the type definition is fixed.

Thank you !

benc-uk:
Any update? Still seeing idTokenClaims as a StringDict with string values, making accessing the _array_ of roles quite tricky

jasonnutter:
@benc-uk We haven't a chance to address this, but you can try the workaround above to get an array for those values.

PR #2105, opened 2020-08-05
[msal-common] Fix ID Token Claims type

### LINKED ISSUE #2084: Incorrect idTokenClaims Type

## Library

## Framework
N/A

## Description
ID token claims include a mix of strings/numbers but the type used (StringDict) only allows string values

MsalTypes.d.ts(5, 5): The expected type comes from this index signature. */

## Security

## Regression
Version:

## MSAL Configuration

```js
```

## Reproduction steps
```ts
import { AuthenticationResult } from '@azure/msal-browser';

const idTokenClaims: AuthenticationResult['idTokenClaims'] = {
  exp: 1596411774067 // error
  // ...
}
```

## Expected behavior
Allow numbers and strings to be used for id token claims

## Browsers/Environment
- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari
- [ ] IE

technical-boy:
This PR:

- Fixes #2084 
- Changes the type for AuthenticationResult.idTokenClaims from StringDict to generic object in order to allow non-string values for ID Token claims

### DIFF
--- a/change/@azure-msal-common-2020-08-11-16-45-31-fix-id-token-claims-type.json
+++ b/change/@azure-msal-common-2020-08-11-16-45-31-fix-id-token-claims-type.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Change AuthenticationResult.idTokenClaims type from StringDict to generic Object type to allow for non-string valued ID Token Claims. (#2105)",
+  "comment": "Change AuthenticationResult.idTokenClaims type from StringDict to generic Object type to allow for non-string valued ID Token Claims. Also updates exp claim type in IdTokenClaims type from string to number (#2105)",
   "packageName": "@azure/msal-common",
   "email": "hemoral@microsoft.com",
   "dependentChangeType": "patch",
--- a/change/@azure-msal-common-2020-08-11-16-45-31-fix-id-token-claims-type.json
+++ b/change/@azure-msal-common-2020-08-11-16-45-31-fix-id-token-claims-type.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Change AuthenticationResult.idTokenClaims type from StringDict to generic Object type to allow for non-string valued ID Token Claims. (#2105)",
+  "packageName": "@azure/msal-common",
+  "email": "hemoral@microsoft.com",
+  "dependentChangeType": "patch",
+  "date": "2020-08-11T23:45:31.158Z"
+}
--- a/change/@azure-msal-common-2020-08-11-16-45-31-fix-id-token-claims-type.json
+++ b/change/@azure-msal-common-2020-08-11-16-45-31-fix-id-token-claims-type.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Change AuthenticationResult.idTokenClaims type from StringDict to generic Object type to allow for non-string valued ID Token Claims. Also updates exp claim type in IdTokenClaims type from string to number (#2105)",
+  "comment": "Update typing of IdTokenClaims (#2105)",
   "packageName": "@azure/msal-common",
   "email": "hemoral@microsoft.com",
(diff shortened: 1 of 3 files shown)


lib/msal-common/src/response/AuthenticationResult.ts
```diff
@@ -27,7 +26,7 @@ export type AuthenticationResult = {
     scopes: Array<string>;
     account: AccountInfo;
     idToken: string;
-    idTokenClaims: StringDict;
+    idTokenClaims: Object;
```
tnorling, line 29: We have an IdTokenClaims type, can we use that instead of the generic "Object"?

technical-boy: We decided to go with a generic Object given that ID Token Claims can conceivably contain claims of any type (excluding functions). The IdTokenClaims type wouldn't solve this particular issue because it defines all the listed claims as string-typed.

tnorling: In that case, can we update the typing in `IdTokenClaims`? Each specific claim should in theory have their own type and we should make sure our internal type complies with what the server is sending. e.g. We can expect `exp` to always be `number`, `iss` to always be `string`, etc.

jasonnutter: We could attempt to explicitly type all of the possible claims, but my understanding is that those may change, and there are also many optional claims that applications can enable. Ultimately, we want to fix bugs like #2084 and #1766 where we have incorrectly asserted at all claims will be strings. And it would be easier and future-proof to only guarantee that it will be an object, instead of attempting to document all possible claims.

tnorling: This value of this is set to an object that is type `IdTokenClaims` in `ResponseHandler` and the `IdToken` constructor so I'm not sure changing it from `StringDict` to `Object` here is really going to solve this problem on it's own. We would either need to change the types in all places in the code or update the type for each claim.

technical-boy: I agree that only changing it here doesn't solve the problem. I was initially unaware that the IdTokenClaims type existed in the library, so I only knew to update the AuthenticationResult type that was throwing the error in #2084.  So if we were to decide on a generic Object type to replace IdTokenClaims, I'd apply that change in the ResponseHandler and the IdToken constructor (and anywhere else the IdTokenClaims type is found).

Having said that, is there any value in explicitly typing each claim? I don't think I have enough context to foresee what kind of future bug may arise for having the claims ambiguously typed as generic Object attributes (maybe the fact that developers won't immediately know which claims are supported by default?)

change/@azure-msal-common-2020-08-11-16-45-31-fix-id-token-claims-type.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Change AuthenticationResult.idTokenClaims type from StringDict to generic Object type to allow for non-string valued ID Token Claims. Also updates exp claim type in IdTokenClaims type from string to number (#2105)",
```
tnorling, line 3: nit: In the interest of easily digestible change files we should aim to keep these comments pretty small. People can then open the relevant PR for more detail. Something like "Update typing of IdTokenClaims (#2105)" would be fine here.

pkanher617: +1

technical-boy: Cool, I'll keep this in mind for the future as well!

technical-boy, line 3:
```suggestion
  "comment": "Update typing of IdTokenClaims (#2105)",
```

PR #2112, opened 2020-08-06
[msal-node][msal-common] Adds support for client credential grant

sangonzal:
- Adds support client credential grant flow to MSAL Node
- Updates response handler to not rely on accounts for flows that acquire tokens for an application instead of an user
- Fixes bug in response handler where it was relying on client info for ADFS accounts


lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -90,6 +90,8 @@ export class ResponseHandler {
             if (!StringUtils.isEmpty(this.clientInfo.uid) && !StringUtils.isEmpty(this.clientInfo.utid)) {
                 this.homeAccountIdentifier = `${this.clientInfo.uid}.${this.clientInfo.utid}`;
             }
+        } else {
+            this.homeAccountIdentifier = "";
```
tnorling, line 94: Should we attempt to use `idToken.claims.oid || idToken.claims.sub` as a fallback in this scenario? And then if that doesn't exist, make it blank?

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -152,12 +153,20 @@ export class ResponseHandler {
...
+            );
         }
 
         // AccessToken
         let cachedAccessToken: AccessTokenEntity = null;
         if (!StringUtils.isEmpty(serverTokenResponse.access_token)) {
-            const responseScopes = ScopeSet.fromString(serverTokenResponse.scope);
+            
+            // If scopes not returned in server response, use request scopes
+            const responseScopes = serverTokenResponse.scope ? ScopeSet.fromString(serverTokenResponse.scope) : new ScopeSet(requestScopes || []);
```
tnorling, line 169: What is the scenario where the server wouldn't return scopes on an access token?

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -172,7 +181,7 @@ export class ResponseHandler {
                 env,
                 serverTokenResponse.access_token,
                 this.clientId,
-                idTokenObj.claims.tid,
+                idTokenObj ? idTokenObj.claims.tid : authority.tenant,
```
tnorling, line 184: `authority.tenant` could be `common` right? Is that ok?

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -237,13 +249,15 @@ export class ResponseHandler {
...
-            uniqueId: idTokenObj.claims.oid || idTokenObj.claims.sub,
-            tenantId: idTokenObj.claims.tid,
+            uniqueId: uid,
+            tenantId: tid,
             scopes: responseScopes,
-            account: cacheRecord.account.getAccountInfo(),
-            idToken: idTokenObj.rawIdToken,
-            idTokenClaims: idTokenObj.claims,
+            account: cacheRecord.account ? cacheRecord.account.getAccountInfo() : null,
+            idToken: idTokenObj ? idTokenObj.rawIdToken : "",
```
pkanher617, line 259: Thinking about this a bit more - should we check cache for a current id token and return it here?

lib/msal-common/src/client/ClientCredentialClient.ts
```diff
@@ -0,0 +1,86 @@
...
+
+        if (this.config.clientCredentials.clientSecret) {
+            parameterBuilder.addClientSecret(this.config.clientCredentials.clientSecret);
+        }
+
+        if (this.config.clientCredentials.clientAssertion) {
+            const clientAssertion = this.config.clientCredentials.clientAssertion;
+            parameterBuilder.addClientAssertion(clientAssertion.assertion);
+            parameterBuilder.addClientAssertionType(clientAssertion.assertionType);
+        }
```
pkanher617, line 82: Do we have empty/null checks on the clientCredentials params?

lib/msal-common/src/response/ResponseHandler.ts
pkanher617: I think we cache with `common` but we should verify this.

lib/msal-common/src/response/ResponseHandler.ts
pkanher617: The client credentials flow does not return scopes in the response. I believe MSAL Java has a similar handling pattern for this.

lib/msal-common/src/response/ResponseHandler.ts
pkanher617: Somewhat related to my comment below about falling back to cached id token.

sangonzal: Chatted offline, will split validateTokenResponse into two functions.

lib/msal-common/src/client/ClientCredentialClient.ts
sangonzal: Yes it is in ConfidentialClientApplication.ts

lib/msal-common/src/response/ResponseHandler.ts
sangonzal: Chatted offline. Decided not to do this. Id token could be expired.

lib/msal-common/src/response/ResponseHandler.ts
sangonzal: Yes, it would just be cached as common.

lib/msal-common/src/response/ResponseHandler.ts
sangonzal: After further looking into this and hearing back about scenarios where client_info can be expected, it is not necessary that to make this change. Currently, `ResponseHandler` checks that client_info was returned when: 
- Id token was returned and is not an ADFS authority. 

In these cases, client_info_empty_error will be thrown if no client_info was returned.

change/@azure-msal-node-2020-08-14-17-01-57-msal-node-client-credential-grant.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
pkanher617, line 2: This should probably be a pre-release

change/msal-node-client-credentials-2020-08-14-17-01-57-msal-node-client-credential-grant.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "minor",
```
pkanher617, line 2: This is a sample, should be "none" change type. We don't want to release samples.

change/@azure-msal-node-2020-08-14-17-01-57-msal-node-client-credential-grant.json
```diff
@@ -0,0 +1,7 @@
...
+  "packageName": "@azure/msal-node",
+  "email": "email not defined",
```
pkanher617, line 4: Need to fix this, otherwise this will show up in the changelog.

pkanher617:
Could you make sure to add the .beachballrc file to the new sample you have created?

change/msal-node-client-credentials-2020-08-14-17-01-57-msal-node-client-credential-grant.json
sangonzal: Updated

change/@azure-msal-node-2020-08-14-17-01-57-msal-node-client-credential-grant.json
sangonzal: Added email.

change/@azure-msal-node-2020-08-14-17-01-57-msal-node-client-credential-grant.json
sangonzal: Updated to prerelase

PR #2114, opened 2020-08-06
getAccountByHomeId + fill username in B2C scenarios

tnorling:
- Adds `getAccountByHomeId` API which will be the recommended single account API going forward
- Since B2C use cases don't return the `preferred_username` claim, use the first element of the `emails` claim when available

General Notes:
- `getAccountByUsername` is provided as a "convenience" and should be considered less reliable than `getAccountByHomeId`
- If a custom policy is intentionally configured to return more than 1 email in the `emails` claim, the username on the account object will only be the first email returned. B2C team says this is a very uncommon scenario and all basic policies return only 1 email


lib/msal-browser/docs/logout.md
```diff
@@ -49,7 +49,7 @@ The logout API also accepts an object for different configuration options:
 ## Code Snippet
 
 ```javascript
-const currentAccount = getAccountByUsername(username);
+const currentAccount = getAccountById(homeAccountId);
```
jasonnutter, line 52: Should this be something like `msalInstance.getAccountbyId`?

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -509,12 +509,32 @@ export class PublicClientApplication implements IPublicClientApplication {
...
+     * This API is provided for convenience but getAccountById should be used for best reliability
      * @returns {@link AccountInfo} - the account object stored in MSAL
      */
     getAccountByUsername(userName: string): AccountInfo|null {
+        if (StringUtils.isEmpty(userName)) {
+            return null;
+        }
+
+        const allAccounts = this.getAllAccounts();
+        return allAccounts && allAccounts.length ? allAccounts.filter(accountObj => accountObj.username.toLowerCase() === userName.toLowerCase())[0] || null : null;
```
jasonnutter, line 522: This line is hard to read, can it be broken up?

lib/msal-common/src/cache/entities/AccountEntity.ts
```diff
@@ -150,7 +150,7 @@ export class AccountEntity {
                 ? idToken.claims.oid
                 : idToken.claims.sid;
             account.localAccountId = localAccountId;
-            account.username = idToken.claims.preferred_username;
+            account.username = idToken.claims.preferred_username || (idToken.claims.emails? idToken.claims.emails[0]: "");
```
jasonnutter, line 153: Any chance this is too opinionated (taking the first email)?

lib/msal-common/src/response/AuthenticationResult.ts
```diff
@@ -4,6 +4,7 @@
  */
 
 import { AccountInfo } from "../account/AccountInfo";
+import { IdTokenClaims } from "../account/IdTokenClaims";
```
jasonnutter, line 7: This doesn't look like its used.

lib/msal-browser/docs/login-user.md
```diff
@@ -68,12 +68,19 @@ When a login call has succeeded, you can use the `getAllAccounts()` function to
 const myAccounts: AccountInfo[] = msalInstance.getAllAccounts();
 ```
 
-If you know the username of the account, you can also retrieve the account information by using the the `getAccountByUsername()` API:
+If you know the account information, you can also retrieve the account information by using the the `getAccountByUsername()` and `getAccountById()` APIs:
```
jasonnutter, line 71:
```suggestion
If you know the account information, you can also retrieve the account information by using the `getAccountByUsername()` or `getAccountById()` APIs:
```

lib/msal-browser/FAQ.md
```diff
@@ -156,3 +158,9 @@ Please see the documentation on [Tenancy in Azure Active Directory](https://docs
 ## My application has multiple resources it needs to access to. How should I handle scopes for access tokens?
 
 Please see the doc about resources and scopes [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/resources-and-scopes.md)
+
+# B2C
+
+## Why is `getAccountByUsername()` returning null, even though I'm signed in?
+
+In order to use `getAccountByUsername()` in B2C scenarios you must enable your `idTokens` to return the `emails` claim in your B2C tenant. MSAL will fill the `username` field on the `AccountInfo` object with the first element of the array returned on the `emails` claim. In most cases this array will only have one element, however, if you notice that your idTokens are returning more than one email on this claim, ensure you are calling `getAccountByUsername` with the first email.
```
jasonnutter, line 166: Is there documentation we can link to for enabling this claim for B2C?

lib/msal-common/src/cache/entities/AccountEntity.ts
tnorling: We talked to the B2C team and they said in 99% of use cases there will only be 1 email ever returned. The reason it's an array is for the other 1% where someone has deliberately configured their custom policy to return more than 1. It's also one of the reasons we're specifically calling out that `getAccountById` should be considered more reliable than `getAccountByUsername`

jasonnutter: Makes sense, can you please add that as a comment here?

lib/msal-browser/docs/login-user.md
sameerag: Since `getAccountById()` is the reliable call in all cases, can we please rephrase this as: If the user has the account information, they can also retrieve the account information by calling the API:  `getAccountById()`, if the user is certain that the account has a `upn` claim, they can retrieve the account information with the API: ``getAccountByUsername()`. Please note for B2C, this is what is expected when you make this call .... << add details here or a link to the B2C docs once they are created>>

lib/msal-browser/FAQ.md
sameerag: I don't think so, but I think it should be added. cc @nickgmicrosoft

lib/msal-browser/docs/logout.md
tnorling: Good catch! Fixed.

lib/msal-browser/src/app/PublicClientApplication.ts
tnorling: Broke it up into if/else to make the logic more clear

lib/msal-common/src/cache/entities/AccountEntity.ts
tnorling: Done

lib/msal-common/src/response/AuthenticationResult.ts
tnorling: Thanks, removed!

lib/msal-browser/FAQ.md
tnorling: Added a short description underneath but would be great to have a more generic doc on docs.microsoft.com that applies to more than just this specific use case.

lib/msal-browser/docs/login-user.md
```diff
@@ -68,12 +68,19 @@ When a login call has succeeded, you can use the `getAllAccounts()` function to
 const myAccounts: AccountInfo[] = msalInstance.getAllAccounts();
 ```
 
-If you know the username of the account, you can also retrieve the account information by using the the `getAccountByUsername()` API:
+If you know the account information, you can also retrieve the account information by using the `getAccountByUsername()` or `getAccountById()` APIs:
```
sameerag, line 71: Can this be instead: 

If you know the account information, you can also retrieve the account information by using the `getAccountById()` or `getAccountByUsername()` APIs. We recommend `getAccountById()` as the default lookup API. Please note that `getAccountByUserName()` can be called for `idToken.preferred_name` for AAD and the first of `idToken.emails` claim for B2C if configured. 

... something that makes it obvious our preference and what they are choosing.

tnorling: Is the note below (Lines 80-83) insufficient or do you just want it moved up into this line?

PR #2115, opened 2020-08-06
[msal-browser] Storage optimization

sameerag:
This PR improves `getCache()` and `setCache()` implementations of the `msal-browser` component by minimizing the references to the exact cache type and introducing helpers as needed.

Todo:
* Remove `type` wherever it isn't needed
* Pass/add UT as needed


lib/msal-browser/src/cache/BrowserStorage.ts
```diff
@@ -427,4 +406,63 @@ export class BrowserStorage extends CacheManager {
...
+        if (StringUtils.isEmpty(value)) {
+            return null;
+        }
+
+        let parsedValue;
+        // handle strings - TEMPORARY Values; check cookies for these
+        try {
+            parsedValue = JSON.parse(value);
+        } catch (e) {
+            const itemCookie = this.getItemCookie(key);
```
tnorling, line 427: Should this be inside the `if` below?

lib/msal-browser/src/cache/BrowserStorage.ts
```diff
@@ -109,80 +126,42 @@ export class BrowserStorage extends CacheManager {
...
             }
+        } else {
+            stringifiedValue = JSON.stringify(value);
         }
+
+        // actual window storage
+        this.windowStorage.setItem(key, stringifiedValue);
     }
 
+    // getItem ->
```
technical-boy, line 150: If this comment is meant as documentation, can we make it a block comment? I assume the linter will complain about multi-lined comments.

lib/msal-browser/src/cache/BrowserStorage.ts
technical-boy: +1

lib/msal-common/src/cache/entities/CredentialEntity.ts
```diff
@@ -82,11 +82,11 @@ export class CredentialEntity {
     generateType(): number {
         switch (this.credentialType) {
```
technical-boy, line 83: Since the known-cases in these switch statements have no behavior logic, they are essentially mapping Credential Types/Authority types to cache value types. Would using a map instead of a switch statement be a better approach? You could still handle the default case by checking if the result of accessing a key on the object/map is undefined.

lib/msal-node/src/cache/serializer/SerializerTypes.ts
```diff
@@ -3,7 +3,10 @@
  * Licensed under the MIT License.
  */
 
-import { AccountCache, IdTokenCache, AccessTokenCache, RefreshTokenCache, AppMetadataCache } from "@azure/msal-common";
+import { AccountCache, IdTokenCache, AccessTokenCache, RefreshTokenCache, AppMetadataCache, IdTokenEntity, AccessTokenEntity, RefreshTokenEntity, AppMetadataEntity, AccountEntity } from "@azure/msal-common";
+
+export type ValidCacheType = AccountEntity | IdTokenEntity | AccessTokenEntity | RefreshTokenEntity | AppMetadataEntity | string;
```
technical-boy, line 8: Is this a duplicated export (same type is exported in cache/utils/CacheTypes)?

lib/msal-browser/src/cache/BrowserStorage.ts
```diff
@@ -109,80 +126,42 @@ export class BrowserStorage extends CacheManager {
      * @param value
      */
     setItem(key: string, value: string | object, type: string): void {
```
pkanher617, line 128: We should change this signature to:
```javascript
setItem(key: string, value: ValidCacheType): void {
```

lib/msal-browser/src/cache/BrowserStorage.ts
```diff
@@ -427,4 +406,63 @@ export class BrowserStorage extends CacheManager {
...
+     * @param value
+     */
+    private getItemHelper(key: string, value: string): ValidCacheType {
+
+        // handle null cases
+        if (StringUtils.isEmpty(value)) {
+            return null;
+        }
+
+        let parsedValue;
```
pkanher617, line 422: Should be ValidCacheType

lib/msal-browser/src/cache/BrowserStorage.ts
```diff
@@ -427,4 +406,63 @@ export class BrowserStorage extends CacheManager {
...
+        } catch (e) {
+            const itemCookie = this.getItemCookie(key);
+            if (this.cacheConfig.storeAuthStateInCookie) {
+                return itemCookie;
+            }
+            return value;
+        }
+
+        // Account
+        if (parsedValue.includes("authorityType")) {
```
pkanher617, line 435: `includes` isn't supported in IE

lib/msal-browser/src/cache/BrowserStorage.ts
sameerag: temporary: will delete

lib/msal-browser/src/cache/BrowserStorage.ts
```diff
@@ -427,4 +401,65 @@ export class BrowserStorage extends CacheManager {
...
+    /**
+     * helper to retrieve the value with type from the cache
+     * @param key
+     * @param value
+     */
+    private getItemHelper(key: string, value: string): ValidCacheType {
+
+        // handle null cases
+        if (StringUtils.isEmpty(value)) {
+            return null;
```
jasonnutter, line 414: If the value from browser storage is empty, we should not immediately return null, as it doesnt give the library the opportunity to see if the value is stored in a cookie. Instead, if `value` is `null` from local/session storage and `storeAuthStateInCookie` is true, then check for the value stored in a cookie.

lib/msal-browser/src/cache/BrowserStorage.ts
```diff
@@ -427,4 +401,65 @@ export class BrowserStorage extends CacheManager {
...
+                const itemCookie = this.getItemCookie(key);
+                return itemCookie;
+            }
+            return value;
+        }
+
+        // Account
+        if (parsedValue.hasOwnProperty(AUTHORITY_TYPE)) {
+            const accountEntity = new AccountEntity();
+            return CacheManager.toObject(accountEntity, parsedValue) as AccountEntity;
```
jasonnutter, line 433: All of this casting business logic should be pulled into helper functions one level up (i.e. functions that call `getItem`).

sameerag:
This PR will be addressed with #2415. Please follow that PR instead.

PR #2128, opened 2020-08-11
Change /token header type to Record

tnorling:
Axios network module in msal-node is silently ignoring and not sending headers because they are type Map. This aims to fix that by changing the type to Record.


lib/msal-browser/src/network/FetchClient.ts
```diff
@@ -57,17 +57,17 @@ export class FetchClient implements INetworkModule {
...
-            headers.append(key, value);
+        Object.keys(options.headers).forEach((key) => {
+            headers.append(key, options.headers[key]);
         });
         return headers;
     }
 
-    private getHeaderMap(headers: Headers): Map<string, string> {
-        const headerMap = new Map<string, string>();
+    private getHeaderDict(headers: Headers): Record<string, string> {
```
sameerag, line 66: Shouldn't this be named `setHeaders()` rather?

sameerag: never mind!

tnorling:
> LGTM. One general q: we use a map to generate all our query params we send to `/authorize` endpoint. Does this affect that?

It shouldn't since in that case the Map is just an internal structure to msal.js that will be turned into a string before actually being used

Issue #2118, opened 2020-08-08
MSAL-Browser 2.0.1: Incorrectly deserializes URL Hash resulting In 'platformStateObj is null' exception


wdspider:
>> Please follow the issue template below. Failure to do so will result in a delay in answering your question.

## Library
- [ ] `msal@1.x.x` or `@azure/msal@1.x.x`
- [x] `@azure/msal-browser@2.0.1`
- [ ] `@azure/msal-angular@0.x.x`
- [ ] `@azure/msal-angular@1.x.x`
- [ ] `@azure/msal-angularjs@1.x.x`

**Important**: Please fill in your exact version number above, e.g. `msal@1.1.3`.

## Framework
TypeScript / React / Material UI

## Description
MSAL redirects to AD B2C fine which returns back to the app as expected.  However, MSAL is not processing the url hash properly.  The problem seems to lie within this method:
```js
PublicClientApplication.prototype.getRedirectResponseHash = function () {
        // Get current location hash from window or cache.
        var hash = window.location.hash;
        var isResponseHash = UrlString.hashContainsKnownProperties(hash);
        var cachedHash = this.browserStorage.getItem(this.browserStorage.generateCacheKey(TemporaryCacheKeys.URL_HASH), CacheSchemaType.TEMPORARY);
        this.browserStorage.removeItem(this.browserStorage.generateCacheKey(TemporaryCacheKeys.URL_HASH));
        var responseHash = isResponseHash ? hash : cachedHash;
        if (responseHash) {
            // Deserialize hash fragment response parameters.
            var serverParams = UrlString.getDeserializedHash(responseHash);
            var platformStateObj = BrowserProtocolUtils.extractBrowserRequestState(this.browserCrypto, serverParams.state);
            if (platformStateObj.interactionType !== InteractionType.REDIRECT) {
                return null;
            }
            else {
                BrowserUtils.clearHash();
                return responseHash;
            }
        }
        return null;
    };
```
particularly, these two lines:
```js
            var serverParams = UrlString.getDeserializedHash(responseHash);
            var platformStateObj = BrowserProtocolUtils.extractBrowserRequestState(this.browserCrypto, serverParams.state);
```

The first line successfully deserializes the hash into this serverParams object:
![image](https://user-images.githubusercontent.com/5971649/89702171-f1490080-d8f2-11ea-82af-e9d46ba31675.png)
[...]

PR #2129, opened 2020-08-11
[msal-browser] Fix bugs in redirect flow

pkanher617:
Fixes bugs raised in #2118 : 
- Hash is incorrectly deserialized because '#' symbol is not removed.
- Cache is not cleared if browser back button is pressed

### DIFF
--- a/lib/msal-common/src/url/UrlString.ts
+++ b/lib/msal-common/src/url/UrlString.ts
@@ -181,12 +181,12 @@ export class UrlString {
     /**
      * Check if the hash of the URL string contains known properties
      */
-    static hashContainsKnownProperties(urlString: string): boolean {
-        if (StringUtils.isEmpty(urlString)) {
+    static hashContainsKnownProperties(hash: string): boolean {
+        if (StringUtils.isEmpty(hash)) {
             return false;
         }
 
-        const parameters: ServerAuthorizationCodeResponse = UrlString.getDeserializedHash(urlString);
+        const parameters: ServerAuthorizationCodeResponse = UrlString.getDeserializedHash(hash);
         return !!(
             parameters.code ||
             parameters.error_description ||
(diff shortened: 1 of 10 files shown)


lib/msal-common/src/url/UrlString.ts
```diff
@@ -145,15 +140,34 @@ export class UrlString {
...
+
     static constructAuthorityUriFromObject(urlObject: IUri): UrlString {
         return new UrlString(urlObject.Protocol + "//" + urlObject.HostNameAndPort + "/" + urlObject.PathSegments.join("/"));
     }
 
     /**
-     * Returns deserialized portion of URL hash
+     * Returns URL hash as server auth code response object.
      */
     static getDeserializedHash(hash: string): ServerAuthorizationCodeResponse {
```
tnorling, line 165: Would "urlString" be a more apt name instead of "hash"?

pkanher617: no because this function specifically takes a hash string. we never pass the entire url here

tnorling: In that case let's update `hashContainsKnownProperties` to be consistent with that. 

Also, why are we then checking if the hash is empty on both line 166 and 170?

pkanher617: > In that case let's update hashContainsKnownProperties to be consistent with that.

`hashContainsKnownProperties` does have use cases where the value passed to it is a full url in `@azure/msal-browser`. We can move it to that package, but I figured it would be fine to leave this here since it is a protocol level API.

> Also, why are we then checking if the hash is empty on both line 166 and 170?

It's a bit of a weird scenario, but line 166 checks the value that is passed in. line 170 checks that the deserialized hash is parsed correctly. That's why one scenario throws an error and one returns an empty string. I will add comments.

tnorling: The way I read this is: 
Url passed in -> !Url.isEmpty -> Attempt to extract the hash from the url -> Attempt to deserialize the hash

I would maybe try to rename some things because `parseHash` is really just taking the right side of the anchor, it's not parsing anything. 

I'd also suggest maybe returning the passed "hash" in `parseHash` if there is no anchor tag, rather than having 2 `isEmpty` calls adding to the confusion

pkanher617: resolved offline.

Issue #2001, opened 2020-07-20
loginPopup fails in mobile safari


michaelmesser:
>> Please follow the issue template below. Failure to do so will result in a delay in answering your question.

## Library
- [ ] `msal@1.x.x` or `@azure/msal@1.x.x`
- [x] `@azure/msal-browser@2.0.0-beta.4`
- [ ] `@azure/msal-angular@0.x.x`
- [ ] `@azure/msal-angular@1.x.x`
- [ ] `@azure/msal-angularjs@1.x.x`

**Important**: Please fill in your exact version number above, e.g. `msal@1.1.3`.

## Description
In mobile safari `loginPopup` fails. I belief this is because `window.open` is called from async. Maybe [this](https://stackoverflow.com/questions/20696041/window-openurl-blank-not-working-on-imac-safari) could fix it?

## Error Message

popup_window_error: Error opening the popup window. This can happen if you are using IE or if popups are blocked in the browser. Details:
BrowserAuthError: empty_window_error: window.open returned null or undefined window object

## Security
- [ ] Is this issue security related? No

## Regression
- [ ] Did this behavior work before? Did not try with a previous version
Version:

## MSAL Configuration

```js
const myMSALObj = new msal.PublicClientApplication({
  auth: {
    clientId: "{}",
    authority:
      "https://login.microsoftonline.com/{}",
    redirectUri: "https://{}",
  },
  cache: {
    cacheLocation: "localStorage",
    storeAuthStateInCookie: false,
  },
});

const request = {
  scopes: ["openid", "profile", "User.Read"],
};
```

## Reproduction steps
```js
const { idToken } = await myMSALObj.loginPopup(request);
```

## Browsers/Environment
- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari
- [ ] IE
- [x] Other (Mobile Safari)

michaelmesser:
So on mobile safari this fails:
```js
setTimeout(() => {
          window.windowRef = window.open();
          window.windowRef.location = "https://google.com";
}, 1000);
```
but this works:
```js
window.windowRef = window.open();
setTimeout(() => {
          window.windowRef.location = "https://google.com";
}, 1000);
```

Both snippets are run in an onclick handler. So this issue can be fixed.

michaelmesser:
```javascript
const windowHack = () => {
  const old_open = window.open;
  const myWindow = window.open();
[...]

PR #2132, opened 2020-08-11
[msal-browser] Open popup change with configuration

jo-arroyo:
This PR addresses issue #2001 , where `loginPopup` was throwing an error and unable to open an popup in mobile Safari, despite being a result of user interaction. This PR aims to allow `loginPopup` to be called and a popup opened despite popup blocking being on, by making the following changes:

- `loginPopup` continues to call `acquireTokenPopup` with either request passed in or default request.
- Adding logic to `acquireTokenPopup` to check configuration to open a popup window in most cases before calling `acquireTokenPopupAsync`.
- Renaming `acquireTokenPopup` to `acquireTokenPopupAsync` and making it a private function. It now takes in the request and an optional popup window.
- Adding `asyncPopups` to Configurations. This flag defaults to false, and can be set to true to not open a popup upon first calling `acquireTokenPopup`, but delayed until later. This is to mitigate scenarios that do not support opening a blank popup, such as with Office Add-ons and PWA.
- Separating out the popup opening logic from `openPopup` into a helper function, `openSizedPopup`. This allows correctly-sized popups to be opened from `acquireTokenPopup` or `openPopup`.
- Adding and adjusting unit tests to reflect above changes.
- Update documentation.

**Note:** This fix has been tested in mobile Safari, Safari, Firefox, Chrome, Edge. It has not been tested for Office Add-Ons.


lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -301,7 +301,24 @@ export class PublicClientApplication implements IPublicClientApplication {
      *
      * @returns {Promise.<AuthenticationResult>} - a promise that is fulfilled when this function has completed, or rejected if an error was raised. Returns the {@link AuthResponse} object
      */
-    async acquireTokenPopup(request: PopupRequest): Promise<AuthenticationResult> {
+    acquireTokenPopup(request: PopupRequest): Promise<AuthenticationResult> {
+        // delayOpenPopup flag is true. Acquires token without first opening popup.
+        if (this.config.system.delayOpenPopup) {
```
jo-arroyo, line 306: Note: open to suggestions on a better configuration name?

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -51,13 +51,15 @@ export type CacheOptions = {
...
  * - loadFrameTimeout             - maximum time the library should wait for a frame to load
+ * - delayOpenPopup               - Sets whether popup is open as soon as loginPopup is called. By default, this flag is set to false. Set to true for native apps and PWA.
  */
 export type BrowserSystemOptions = SystemOptions & {
     loggerOptions?: LoggerOptions;
     networkClient?: INetworkModule;
     windowHashTimeout?: number;
     iframeHashTimeout?: number;
     loadFrameTimeout?: number;
+    delayOpenPopup?: boolean;
```
jo-arroyo, line 62: Note: Open to suggestions for a better name for this?

lib/msal-browser/test/interaction_handler/PopupHandler.spec.ts
```diff
@@ -242,4 +242,101 @@ describe("PopupHandler.ts Unit Tests", () => {
...
+				redirectUri: `${TEST_URIS.DEFAULT_INSTANCE}/`,
+				code: "thisIsATestCode",
+				scopes: TEST_CONFIG.DEFAULT_SCOPES,
+				codeVerifier: TEST_CONFIG.TEST_VERIFIER,
+				authority: `${Constants.DEFAULT_AUTHORITY}/`,
+				correlationId: RANDOM_TEST_GUID
+            };
+
+            expect(() => popupHandler.initiateAuthRequest("http://localhost/#/code=hello", testRequest)).to.throw(BrowserAuthErrorMessage.emptyWindowError.desc);
+            expect(() => popupHandler.initiateAuthRequest("http://localhost/#/code=hello", testRequest)).to.throw(BrowserAuthError);
```
jasonnutter, line 309: I don't think you need both of these assertions, the first one should be enough.

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -51,13 +51,15 @@ export type CacheOptions = {
  * - windowHashTimeout            - sets the timeout for waiting for a response hash in a popup
  * - iframeHashTimeout            - sets the timeout for waiting for a response hash in an iframe
  * - loadFrameTimeout             - maximum time the library should wait for a frame to load
+ * - delayOpenPopup               - Sets whether popup is open as soon as loginPopup is called. By default, this flag is set to false. Set to true for native apps and PWA.
```
tnorling, line 54: nit: I would try to rephrase the first sentence to provide both scenarios and make it more generic as it applies to `acquireTokenPopup` as well.  For example: " when set to false, blank popup is opened before anything else happens. When set to true, popup is opened when making the network request"

lib/msal-browser/src/config/Configuration.ts
jasonnutter: `asyncPopups` is the only goodish name I can think of at the moment.

lib/msal-browser/src/interaction_handler/PopupHandler.ts
```diff
@@ -111,25 +111,19 @@ export class PopupHandler extends InteractionHandler {
...
-            const width = window.innerWidth || document.documentElement.clientWidth || document.body.clientWidth;
-            const height = window.innerHeight || document.documentElement.clientHeight || document.body.clientHeight;
-            const left = Math.max(0, ((width / 2) - (popUpWidth / 2)) + winLeft);
-            const top = Math.max(0, ((height / 2) - (popUpHeight / 2)) + winTop);
-
-            // open the window
-            const popupWindow = window.open(urlNavigate, title, "width=" + popUpWidth + ", height=" + popUpHeight + ", top=" + top + ", left=" + left);
+            let popupWindow;
+            // Popup window passed in, setting url to navigate to
+            if (popup) {
```
tnorling, line 118: Out of curiosity what's the reason why we open a blank popup first by default? Why would we not want to change the code to have a single code path where we just always open the popup directly to the /authorize endpoint?

jasonnutter:
This looks great, thanks for the description and the tests! Interested to hear other suggestions for the name, but otherwise looks good!

lib/msal-browser/src/interaction_handler/PopupHandler.ts
```diff
@@ -147,6 +141,25 @@ export class PopupHandler extends InteractionHandler {
         }
     }
 
+    static openSizedPopup(urlNavigate: string = "about:blank"): Window|null {
```
tnorling, line 144: For consistency's sake I would personally prefer we make this optional and have a line underneath here that checks if `StringUtils.isEmpty(urlNavigate)` and sets it to the default if it is

jasonnutter: I personally like using default values instead of `isEmpty`, more readable imo.

lib/msal-browser/src/config/Configuration.ts
jo-arroyo: Sounds good! Changed.

lib/msal-browser/test/interaction_handler/PopupHandler.spec.ts
jo-arroyo: Noted. Removed

lib/msal-browser/src/interaction_handler/PopupHandler.ts
jo-arroyo: It appears that a number of browsers will block popups that are opened in asynchronous code paths even if they wouldn't otherwise be blocked (e.g. as a result of user interaction). The mitigation is to synchronously open a blank popup and asynchronously update it with the url. The issue reported has more information on this: #2001

tnorling: Sorry, guess I misunderstood the behavior today vs. the fix. My question was more around, if that's the case why don't we always open a blank popup synchronously and update asynchronously with the url? What's the reasoning behind creating an additional code path and parameter vs. just changing the existing behavior to fix the issue on those browsers? Would that break the other browsers that are currently not broken?

lib/msal-browser/src/interaction_handler/PopupHandler.ts
pkanher617: +1 for this notation, it is a bit more readable.

pkanher617: `StringUtils.isEmpty` check should still be added in case `openSizedPopup("")` is called.

jasonnutter: > `StringUtils.isEmpty` check should still be added in case `openSizedPopup("")` is called.

I'm not too worried about that happening, to be honest.

lib/msal-browser/src/interaction_handler/PopupHandler.ts
jo-arroyo: Oh, I see what you're asking. The fix doesn't break with other browsers, so we are having it as the default code path. However, it seems that some desktop apps and progressive web apps have problems with blank popups, so in msal-core we had to stop opening blank popups entirely and open them later with the url. Having two code paths in this fix tries make sure that both kinds of issues are covered. Here's the other PR and related issues for desktop apps and PWA: #1667

lib/msal-browser/src/interaction_handler/PopupHandler.ts
jasonnutter: Also, I believe calling `window.open` with `""` is functionality equivalent to `"about:blank"`

lib/msal-browser/src/interaction_handler/PopupHandler.ts
tnorling: Oh that's right, makes sense. Thanks for clearing that up!

lib/msal-browser/src/interaction_handler/PopupHandler.ts
tnorling: This is fine. I generally agree with this being more readable if it's just one or a small number of params. I think my comment was more around keeping a consistent code pattern (I don't think I've seen this anywhere else in our code). If we want to start doing this more, I'm all for it 👍

jasonnutter:
Can you add this option to this page? https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/configuration.md#system-config-options

lib/msal-browser/docs/configuration.md
```diff
@@ -81,6 +82,7 @@ const msalInstance = new PublicClientApplication(msalConfig);
 | `windowHashTimeout` | Timeout in milliseconds to wait for popup authentication to resolve. | integer (milliseconds) | `60000` |
 | `iframeHashTimeout` | Timeout in milliseconds to wait for iframe authentication to resolve | integer (milliseconds) | `6000` |
 | `loadFrameTimeout` | Timeout in milliseconds to wait for the iframe to load in the window. | integer (milliseconds) | In IE or Edge: `500`, in all other browsers: `0` |
+| `asyncPopups` | Sets whether popups are opened asynchronously. When set to false, blank popups are opened before anything else happens. When set to true, popups are opened when making the network request. Can be set to true for scenarios that do not support popups, e.g. desktop apps or progressive web apps | boolean | `false` |
```
jasonnutter, line 85:
```suggestion
| `asyncPopups` | Sets whether popups are opened asynchronously. When set to false, blank popups are opened before anything else happens. When set to true, popups are opened when making the network request. Can be set to true for scenarios where `about:blank` is not supported, e.g. desktop apps or progressive web apps | boolean | `false` |
```

PR #2141, opened 2020-08-13
[msal-common] Add validation to cache entities

sameerag:
Adds validation for a cache entity. 

This is needed as `msal-browser` saves the values in the `storage` as `JSON strings` instead of `objects` and when retrieving the `values`, they need to be assigned as the object instances of the `cache entities`.

**Context:**
This PR is the first of the series of changes being made to optimize and simplify `getItem()` and `setItem()` implementations in `msal-browser` and `msal-node`

### DIFF
--- a/lib/msal-common/src/cache/entities/AppMetadataEntity.ts
+++ b/lib/msal-common/src/cache/entities/AppMetadataEntity.ts
@@ -2,16 +2,16 @@
  * Copyright (c) Microsoft Corporation. All rights reserved.
  * Licensed under the MIT License.
  */
-import { APP_META_DATA, Separators } from "../../utils/Constants";
+import { APP_METADATA, Separators } from "../../utils/Constants";
 
 /**
  * APP_META_DATA Cache
- * 
+ *
  * Key:Value Schema:
- * 
+ *
  * Key: appmetadata-<environment>-<client_id>
- * 
- * Value: 
+ *
+ * Value:
  * {
  *      clientId: client ID of the application
  *      environment: entity that issued the token, represented as a full host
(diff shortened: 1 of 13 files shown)


lib/msal-common/src/cache/entities/AppMetadataEntity.ts
```diff
@@ -27,7 +27,19 @@ export class AppMetadataEntity {
...
         return appMetaDataKeyArray.join(Separators.CACHE_KEY_SEPARATOR).toLowerCase();
     }
+
+    /**
+     * Validates an entity: checks for all expected params
+     * @param entity
+     */
+    static isAppMetadataEntity(key: string, entity: object): boolean {
+        return (
+            (key.indexOf(APP_METADATA) !== -1) &&
```
tnorling, line 40: If the key is expected to start with "appmetadata" you can check if `key.indexOf(APP_METADATA) === 0` here

lib/msal-common/src/cache/entities/AppMetadataEntity.ts
```diff
@@ -27,7 +27,19 @@ export class AppMetadataEntity {
...
-        const appMetaDataKeyArray: Array<string> = [APP_META_DATA, this.environment, this.clientId];
+        const appMetaDataKeyArray: Array<string> = [APP_METADATA, this.environment, this.clientId];
         return appMetaDataKeyArray.join(Separators.CACHE_KEY_SEPARATOR).toLowerCase();
     }
+
+    /**
+     * Validates an entity: checks for all expected params
+     * @param entity
+     */
+    static isAppMetadataEntity(key: string, entity: object): boolean {
```
tnorling, line 38: Whats the difference between this function and `isAppMetadata` in `CacheManager`?

lib/msal-common/src/cache/entities/AppMetadataEntity.ts
sameerag: yes. Good call.

lib/msal-common/src/cache/entities/AppMetadataEntity.ts
sameerag: We are also checking the entity.

tnorling: Curious what the scenarios are for having both functions? Would `isAppMetadataEntity` be enough to cover all scenarios?

sameerag: There may not be. These are the validations we are providing and why:
* **Key based validation** - applicable in `getByFilter` APIs in `CacheManager`, we should be able to fetch the key and determine if we want to fetch the value; it is performance degrading to fetch the value for all keys in a loop. This is currently supported by `getCredentialType()` and `isAppMetadata()` in `CacheManager` today. Moving them to `entities` as `static APIs` may make more sense now.

* **Value based validation** - This is what we are adding now. This comes to play when we are in `getItem()` fetching either the instance(`msal-node`) or a JSON object(`msal-browser`), and verify the value that we are fetching is indeed what we want.

tnorling: Got it, that makes sense. Thanks!

PR #2148, opened 2020-08-13
B2C Sample + E2E Test 2.0

tnorling:
B2C sample and E2E tests

Future enhancements: 
- Test social account login (Google currently requires MFA, will need to work with lab team to figure out how we can bypass that)
- Add forgot password flow and test


LarsKemmann:
I'm working through an Angular setup with MSAL.js 2.0 and B2C for a client, and found my way to the _angular10-browser-sample_ in this PR. First, thank you for all the great work on this! :) Second, can you provide some guidance on how to call a B2C-protected API from Angular in that sample? The sample currently has:

```
protectedResourceMap.set('https://graph.microsoft.com/v1.0/me', ['user.read'])
```

This doesn't work since B2C tokens can't be used to make MS Graph API calls. I tried setting up my own but the following doesn't seem to work correctly, and instead I get an infinite redirect loop with B2C and the app -- I'm guessing the app doesn't consider the signed-in user to have the required scopes? I did grant admin consent to the app and ensured that the redirect URLs are configured with SPA, and that the implicit grant is disabled.

```
protectedResourceMap.set('https://aspnetcoreapib2c.azurewebsites.net/hello', ['https://vibraniumtechdemob2c.onmicrosoft.com/api/demo.read'])
```

tnorling:
@LarsKemmann Can you please open an issue? Thanks!

pkanher617 (approved):

jasonnutter (approved):

PR #2151, opened 2020-08-14
AT Proof-Of-Possession #1: Adding req_cnf to the token request

pkanher617:
This PR is the first for the Proof-of-Possession feature in MSAL.js 2.x. 

This PR adds the following:

- Adding the `req_cnf` parameter to the access token request in the authorization code flow
- Adding authenticationType to the request object
- Generating and exporting keypairs in the browser package

### DIFF
--- a/lib/msal-browser/src/crypto/BrowserCrypto.ts
+++ b/lib/msal-browser/src/crypto/BrowserCrypto.ts
@@ -112,6 +112,7 @@ export class BrowserCrypto {
      * @param data 
      */
     private async getSubtleCryptoDigest(algorithm: string, data: Uint8Array): Promise<ArrayBuffer> {
+        console.log(algorithm);
         return window.crypto.subtle.digest(algorithm, data);
     }
 
--- a/lib/msal-browser/src/crypto/BrowserCrypto.ts
+++ b/lib/msal-browser/src/crypto/BrowserCrypto.ts
@@ -165,7 +165,6 @@ export class BrowserCrypto {
         return new Promise((resolve: any, reject: any) => {
             const msExportKey = window["msCrypto"].subtle.exportKey(format, key);
             msExportKey.addEventListener("complete", (e: { target: { result: ArrayBuffer; }; }) => {
-                // TODO: Check with Jason for details?
                 const resultBuffer: ArrayBuffer = e.target.result;
 
                 const resultString = BrowserStringUtils.utf8ArrToString(new Uint8Array(resultBuffer))
--- a/lib/msal-browser/src/crypto/BrowserCrypto.ts
+++ b/lib/msal-browser/src/crypto/BrowserCrypto.ts
@@ -17,13 +17,8 @@ const S256_HASH_ALG = "SHA-256";
 const MODULUS_LENGTH = 2048;
 // Public Exponent
 const PUBLIC_EXPONENT: Uint8Array = new Uint8Array([0x01, 0x00, 0x01]);
-
-/**
- * Used for exporting keys
- */
-export enum KeyFormat {
-    jwk = "jwk"
-}
+// JWK Key Format string
+const KEY_FORMAT_JWK = "jwk";
 
 /**
  * This class implements functions used by the browser library to perform cryptography operations such as
@@ -33,23 +28,17 @@ export class BrowserCrypto {
 
     private _keygenAlgorithmOptions: RsaHashedKeyGenParams;
 
-    constructor(keygenAlgName?: string, hashAlg?: string, modLength?: number, publicExp?: Uint8Array) {
+    constructor() {
         if (!(this.hasCryptoAPI())) {
             throw BrowserAuthError.createCryptoNotAvailableError("Browser crypto or msCrypto object not available.");
         }
 
-        const keygenConfigName: string = StringUtils.isEmpty(keygenAlgName) ? PKCS1_V15_KEYGEN_ALG : keygenAlgName;
-        const keygenConfigHash: string = StringUtils.isEmpty(hashAlg) ? S256_HASH_ALG : hashAlg;
-        const keygenConfigModulusLength: number = modLength ? modLength : MODULUS_LENGTH;
-        const keygenConfigPublicExponent: Uint8Array = publicExp ? publicExp : PUBLIC_EXPONENT;
-
         this._keygenAlgorithmOptions = {
-            name: keygenConfigName,
-            hash: keygenConfigHash,
-            modulusLength: keygenConfigModulusLength,
-            publicExponent: keygenConfigPublicExponent
+            name: PKCS1_V15_KEYGEN_ALG,
+            hash: S256_HASH_ALG,
+            modulusLength: MODULUS_LENGTH,
+            publicExponent: PUBLIC_EXPONENT
         };
-        console.log(this._keygenAlgorithmOptions);
     }
 
     /**
@@ -92,8 +81,8 @@ export class BrowserCrypto {
      * @param key 
      * @param format 
      */
-    async exportKey(key: CryptoKey, format: KeyFormat): Promise<JsonWebKey> {
-        return this.hasIECrypto() ? this.msCryptoExportKey(key, format) : window.crypto.subtle.exportKey(format, key);
+    async exportJwk(key: CryptoKey): Promise<JsonWebKey> {
+        return this.hasIECrypto() ? this.msCryptoExportJwk(key) : window.crypto.subtle.exportKey(KEY_FORMAT_JWK, key);
     }
 
     /**
@@ -107,14 +96,14 @@ export class BrowserCrypto {
      * Checks whether IE crypto (AKA msCrypto) is available.
      */
     private hasIECrypto(): boolean {
-        return !!window["msCrypto"];
+        return "msCrypto" in window;
     }
 
     /**
      * Check whether browser crypto is available.
      */
     private hasBrowserCrypto(): boolean {
-        return !!window.crypto;
+        return "crypto" in window;
     }
 
     /**
@@ -161,9 +150,9 @@ export class BrowserCrypto {
      * @param key 
      * @param format 
      */
-    private async msCryptoExportKey(key: CryptoKey, format: KeyFormat): Promise<JsonWebKey> {
+    private async msCryptoExportJwk(key: CryptoKey): Promise<JsonWebKey> {
         return new Promise((resolve: any, reject: any) => {
-            const msExportKey = window["msCrypto"].subtle.exportKey(format, key);
+            const msExportKey = window["msCrypto"].subtle.exportKey(KEY_FORMAT_JWK, key);
             msExportKey.addEventListener("complete", (e: { target: { result: ArrayBuffer; }; }) => {
                 const resultBuffer: ArrayBuffer = e.target.result;
 
--- a/lib/msal-browser/src/crypto/BrowserCrypto.ts
+++ b/lib/msal-browser/src/crypto/BrowserCrypto.ts
@@ -1,12 +1,29 @@
 /*
  * Copyright (c) Microsoft Corporation. All rights reserved.
  * Licensed under the MIT License.
+ * 
  */
 import { BrowserStringUtils } from "../utils/BrowserStringUtils";
 import { BrowserAuthError } from "../error/BrowserAuthError";
-
+import { StringUtils } from "@azure/msal-common";
+/**
+ * See here for more info on RsaHashedKeyGenParams: https://developer.mozilla.org/en-US/docs/Web/API/RsaHashedKeyGenParams
+ */
+// RSA KeyGen Algorithm
+const PKCS1_V15_KEYGEN_ALG = "RSASSA-PKCS1-v1_5";
 // SHA-256 hashing algorithm
-const HASH_ALG = "SHA-256";
+const S256_HASH_ALG = "SHA-256";
+// MOD length for PoP tokens
+const MODULUS_LENGTH = 2048;
+// Public Exponent
+const PUBLIC_EXPONENT: Uint8Array = new Uint8Array([0x01, 0x00, 0x01]);
+
+/**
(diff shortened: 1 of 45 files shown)


lib/msal-browser/src/crypto/BrowserCrypto.ts
```diff
@@ -1,23 +1,55 @@
...
-
+import { StringUtils } from "@azure/msal-common";
+/**
+ * See here for more info on RsaHashedKeyGenParams: https://developer.mozilla.org/en-US/docs/Web/API/RsaHashedKeyGenParams
+ */
+// RSA KeyGen Algorithm
+const PKCS1_V15_KEYGEN_ALG = "RSASSA-PKCS1-v1_5";
 // SHA-256 hashing algorithm
-const HASH_ALG = "SHA-256";
+const S256_HASH_ALG = "SHA-256";
```
jasonnutter, line 15: I think `HASH_ALG` is a better name. I personally don't think constants like this need to include hints about their value.

lib/msal-browser/src/crypto/BrowserCrypto.ts
```diff
@@ -1,23 +1,55 @@
...
  */
 import { BrowserStringUtils } from "../utils/BrowserStringUtils";
 import { BrowserAuthError } from "../error/BrowserAuthError";
-
+import { StringUtils } from "@azure/msal-common";
+/**
+ * See here for more info on RsaHashedKeyGenParams: https://developer.mozilla.org/en-US/docs/Web/API/RsaHashedKeyGenParams
+ */
+// RSA KeyGen Algorithm
+const PKCS1_V15_KEYGEN_ALG = "RSASSA-PKCS1-v1_5";
```
jasonnutter, line 13: Same here, could be `KEYGEN_ALG`

lib/msal-browser/src/crypto/BrowserCrypto.ts
```diff
@@ -1,23 +1,55 @@
...
+// MOD length for PoP tokens
+const MODULUS_LENGTH = 2048;
+// Public Exponent
+const PUBLIC_EXPONENT: Uint8Array = new Uint8Array([0x01, 0x00, 0x01]);
+
+/**
+ * Used for exporting keys
+ */
+export enum KeyFormat {
+    jwk = "jwk"
```
jasonnutter, line 25: If we are only using JWK, this probably doesn't need to be an enum (to be consistent with the other constants above).

lib/msal-browser/src/crypto/BrowserCrypto.ts
```diff
@@ -1,23 +1,55 @@
...
 /**
  * This class implements functions used by the browser library to perform cryptography operations such as
  * hashing and encoding. It also has helper functions to validate the availability of specific APIs.
  */
 export class BrowserCrypto {
 
-    constructor() {
+    private _keygenAlgorithmOptions: RsaHashedKeyGenParams;
+
+    constructor(keygenAlgName?: string, hashAlg?: string, modLength?: number, publicExp?: Uint8Array) {
```
jasonnutter, line 36: Why are these optional? Seems like they should all be required.

lib/msal-browser/src/crypto/BrowserCrypto.ts
```diff
@@ -1,23 +1,55 @@
...
+        const keygenConfigModulusLength: number = modLength ? modLength : MODULUS_LENGTH;
+        const keygenConfigPublicExponent: Uint8Array = publicExp ? publicExp : PUBLIC_EXPONENT;
+
+        this._keygenAlgorithmOptions = {
+            name: keygenConfigName,
+            hash: keygenConfigHash,
+            modulusLength: keygenConfigModulusLength,
+            publicExponent: keygenConfigPublicExponent
+        };
+        console.log(this._keygenAlgorithmOptions);
```
jasonnutter, line 52: Remove.

lib/msal-browser/src/crypto/BrowserCrypto.ts
```diff
@@ -1,23 +1,55 @@
...
+
+    constructor(keygenAlgName?: string, hashAlg?: string, modLength?: number, publicExp?: Uint8Array) {
         if (!(this.hasCryptoAPI())) {
             throw BrowserAuthError.createCryptoNotAvailableError("Browser crypto or msCrypto object not available.");
         }
+
+        const keygenConfigName: string = StringUtils.isEmpty(keygenAlgName) ? PKCS1_V15_KEYGEN_ALG : keygenAlgName;
+        const keygenConfigHash: string = StringUtils.isEmpty(hashAlg) ? S256_HASH_ALG : hashAlg;
+        const keygenConfigModulusLength: number = modLength ? modLength : MODULUS_LENGTH;
+        const keygenConfigPublicExponent: Uint8Array = publicExp ? publicExp : PUBLIC_EXPONENT;
```
jasonnutter, line 44: It is a bit cleaner to use default values for these.

lib/msal-browser/src/crypto/BrowserCrypto.ts
```diff
@@ -63,6 +103,20 @@ export class BrowserCrypto {
         return this.hasIECrypto() || this.hasBrowserCrypto();
     }
 
+    /**
+     * Checks whether IE crypto (AKA msCrypto) is available.
+     */
+    private hasIECrypto(): boolean {
+        return !!window["msCrypto"];
```
jasonnutter, line 110: A slightly cleaner way:
```suggestion
        return "msCrypto" in window;
```

lib/msal-browser/src/crypto/BrowserCrypto.ts
```diff
@@ -63,6 +103,20 @@ export class BrowserCrypto {
...
+    }
+
+    /**
+     * Check whether browser crypto is available.
+     */
+    private hasBrowserCrypto(): boolean {
+        return !!window.crypto;
```
jasonnutter, line 117:
```suggestion
        return "crypto" in window;
```

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -173,6 +174,11 @@ export class AuthorizationCodeClient extends BaseClient {
         parameterBuilder.addGrantType(GrantType.AUTHORIZATION_CODE_GRANT);
         parameterBuilder.addClientInfo();
 
+        if (request.authenticationScheme === AuthenticationType.POP) {
+            const popTokenGenerator = new PopTokenGenerator(this.cryptoUtils);
+            parameterBuilder.addPopToken(await popTokenGenerator.generateCnf());
```
jasonnutter, line 179: Can you factor the `generateCnf` call out to a variable?

lib/msal-common/src/crypto/PopTokenGenerator.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    private cryptoUtils: ICrypto;
+
+    constructor(cryptoUtils: ICrypto) {
+        this.cryptoUtils = cryptoUtils;
+    }
+
+    async generateCnf(): Promise<string> {
+        const reqCnf: ReqCnf = {
+            kid: await this.cryptoUtils.getPublicKeyThumprint(),
```
jasonnutter, line 30: Can the `keyPublicKeyThumbprint` be factored to a variable? Also, spelling.

samples/msal-browser-samples/VanillaJSTestApp2.0/app/default/authConfig.js
```diff
@@ -35,7 +35,8 @@ const msalConfig = {
 
 // Add here scopes for id token to be used at MS Identity Platform endpoints.
 const loginRequest = {
-    scopes: ["User.Read"]
+    scopes: ["User.Read"],
+    authenticationType: "pop"
```
jasonnutter, line 39: Should show using the enum.

lib/msal-browser/src/crypto/BrowserCrypto.ts
pkanher617: One reason I did this is because without this we do not know what this constant is referring to without looking at the value. This improves readability and makes dev a little easier.

lib/msal-browser/src/crypto/BrowserCrypto.ts
pkanher617: see above comment.

lib/msal-browser/src/crypto/BrowserCrypto.ts
pkanher617: I can rename this.

lib/msal-browser/src/crypto/BrowserCrypto.ts
pkanher617: I made them optional and set defaults - I can set them from the calling class, just thought it was easier to keep it all in the BrowserCrypto.ts file.

lib/msal-common/src/request/AuthorizationUrlRequest.ts
```diff
@@ -32,6 +32,7 @@ import { BaseAuthRequest } from "./BaseAuthRequest";
  * - nonce                      - A value included in the request that is returned in the id token. A randomly generated unique value is typically used to mitigate replay attacks.
  */
 export type AuthorizationUrlRequest = BaseAuthRequest & {
+    authenticationScheme?: AuthenticationType,
```
tnorling, line 35: Should we call this `authenticationType` to match the name of the type and avoid confusion?

samples/msal-browser-samples/VanillaJSTestApp2.0/app/default/authConfig.js
tnorling, line 39: If you want to leave the parameter name above, this should be `authenticationScheme`

lib/msal-common/src/crypto/ICrypto.ts
```diff
@@ -35,4 +35,8 @@ export interface ICrypto {
      * Generate PKCE codes for OAuth. See RFC here: https://tools.ietf.org/html/rfc7636
      */
     generatePkceCodes(): Promise<PkceCodes>;
+    /**
+     * Generates an JWK RSA S256 Thumbprint
+     */
+    getPublicKeyThumprint(): Promise<string>;
```
tnorling, line 41: Need to change this in a number of places
```suggestion
    getPublicKeyThumbprint(): Promise<string>;
```

tnorling:
Do we have docs, or are you working on docs, that explain the difference between Pop and Bearer and when people should use one over the other?

lib/msal-common/src/request/AuthorizationUrlRequest.ts
pkanher617: changed to authenticationScheme to match other MSAL libs. I can change the name of the type as well.

pkanher617:
> Do we have docs, or are you working on docs, that explain the difference between Pop and Bearer and when people should use one over the other?

Docs will be part of separate PR, but I plan on including them before we merge this feature.

lib/msal-browser/src/crypto/BrowserCrypto.ts
pkanher617: done.

lib/msal-browser/src/crypto/BrowserCrypto.ts
pkanher617: changed.

lib/msal-browser/src/crypto/BrowserCrypto.ts
pkanher617: done.

lib/msal-browser/src/crypto/BrowserCrypto.ts
pkanher617: done.

lib/msal-browser/src/crypto/BrowserCrypto.ts
pkanher617: done.

lib/msal-browser/src/crypto/BrowserCrypto.ts
pkanher617: done.

lib/msal-common/src/client/AuthorizationCodeClient.ts
pkanher617: done.

lib/msal-common/src/crypto/PopTokenGenerator.ts
pkanher617: done.

samples/msal-browser-samples/VanillaJSTestApp2.0/app/default/authConfig.js
pkanher617: done.

pkanher617: done

lib/msal-common/src/crypto/ICrypto.ts
pkanher617: done

lib/msal-browser/src/crypto/BrowserCrypto.ts
sangonzal: My 2c: I would also prefer to be more precise with the naming. Updating this here makes it more readable. Same thing below.

lib/msal-browser/src/crypto/BrowserCrypto.ts
```diff
@@ -63,6 +92,20 @@ export class BrowserCrypto {
         return this.hasIECrypto() || this.hasBrowserCrypto();
     }
 
+    /**
+     * Checks whether IE crypto (AKA msCrypto) is available.
+     */
+    private hasIECrypto(): boolean {
+        return "msCrypto" in window;
```
sangonzal, line 99: Looks like "msCrypto" is used in multiple places, could probably be made into a constant.

PR #2152, opened 2020-08-14
Customize E2E Tests

tnorling:
- Moves test utils files outside the msal-browser directory for easier reuse across other libraries
- Adds configurable sample which loads in a static json file containing the msal config and request objects. The e2e test file can create different json files for each test, allowing it to test many different config combinations.
- Usage of the lab API is more flexible, enabling us to test different accounts and app registrations dynamically

Next Steps:
- Determine which samples are no longer needed and can be replaced by an example json config file. Samples showing scenarios that require code to be organized differently (such as onPageLoad sample) may need to remain
- Update e2e test to use lab API to fill in clientId, authority, etc.
- Update 1.x samples/tests to use new framework


pkanher617 (approved):

jmckennon (approved):
LGTM!

PR #2153, opened 2020-08-15
AT Proof-Of-Possession #2: Handle proof-of-possession tokens correctly in responses

pkanher617:
This PR is a follow up to PR #2151, and implements the response handling part of the proof-of-possession flow in msal-common.

This PR specifically adds the following:

- Renaming IdToken and IdTokenClaims to more generic class names
- Update request objects to include the resource request method and URI, and pass these to the sign token method
- Adds token signing interface method
- Updates cache method to include token type
- Updates response handling flow to handle proof-of-possession flow


lib/msal-common/src/crypto/ICrypto.ts
```diff
@@ -37,6 +39,13 @@ export interface ICrypto {
     generatePkceCodes(): Promise<PkceCodes>;
     /**
      * Generates an JWK RSA S256 Thumbprint
+     * @param resourceRequestMethod 
+     * @param resourceRequestUri 
+     */
+    getPublicKeyThumprint(resourceRequestMethod: string, resourceRequestUri: string): Promise<string>;
```
jo-arroyo, line 45: nit: thumbprint spelling error (saw something similar in PR #1, may already be corrected on other PR)

lib/msal-common/src/account/JwtToken.ts
```diff
@@ -11,39 +11,39 @@ import { ICrypto } from "../crypto/ICrypto";
...
+        this.claims = JwtToken.extractTokenClaims(rawToken, crypto);
     }
 
     /**
      * Extract IdToken by decoding the RAWIdToken
      *
      * @param encodedIdToken
      */
-    static extractIdToken(encodedIdToken: string, crypto: ICrypto): IdTokenClaims {
+    static extractTokenClaims(encodedIdToken: string, crypto: ICrypto): TokenClaims {
```
tnorling, line 34:
```suggestion
    static extractTokenClaims(encodedToken: string, crypto: ICrypto): TokenClaims {
```

lib/msal-common/src/account/JwtToken.ts
```diff
@@ -11,39 +11,39 @@ import { ICrypto } from "../crypto/ICrypto";
...
         // id token will be decoded to get the username
         const decodedToken: DecodedJwt = StringUtils.decodeJwt(encodedIdToken);
```
tnorling, line 36:
```suggestion
        const decodedToken: DecodedJwt = StringUtils.decodeJwt(encodedToken);
```

lib/msal-common/src/error/ClientAuthError.ts
```diff
@@ -191,18 +191,18 @@ export class ClientAuthError extends AuthError {
...
     /**
      * Creates an error thrown when the id token string is null or empty.
      * @param invalidRawTokenString
      */
-    static createIdTokenNullOrEmptyError(invalidRawTokenString: string) : ClientAuthError {
-        return new ClientAuthError(ClientAuthErrorMessage.nullOrEmptyIdToken.code,
-            `${ClientAuthErrorMessage.nullOrEmptyIdToken.desc} Raw ID Token Value: ${invalidRawTokenString}`);
+    static createTokenNullOrEmptyError(invalidRawTokenString: string) : ClientAuthError {
+        return new ClientAuthError(ClientAuthErrorMessage.nullOrEmptyToken.code,
+            `${ClientAuthErrorMessage.nullOrEmptyToken.desc} Raw ID Token Value: ${invalidRawTokenString}`);
```
tnorling, line 205:
```suggestion
            `${ClientAuthErrorMessage.nullOrEmptyToken.desc} Raw Token Value: ${invalidRawTokenString}`);
```

lib/msal-common/src/crypto/ICrypto.ts
pkanher617: corrected in previous PR

lib/msal-common/src/account/JwtToken.ts
```diff
@@ -11,39 +11,39 @@ import { ICrypto } from "../crypto/ICrypto";
 /**
  * Id Token representation class. Parses id token string and generates claims object.
  */
-export class IdToken {
+export class JwtToken {
```
jasonnutter, line 14: This is redundant, and I don't think the name should include JWT, as JWT is an implementation detail.

pkanher617: JWTs are a specific format of token, and this class is built to that specific implementation. So I'm not sure I see the issue...?

lib/msal-common/src/error/ClientAuthError.ts
pkanher617: done.

lib/msal-common/src/account/JwtToken.ts
pkanher617: done

lib/msal-common/src/account/JwtToken.ts
pkanher617: done.

lib/msal-common/src/index.ts
```diff
@@ -7,8 +7,8 @@ export { AuthOptions, SystemOptions, LoggerOptions, DEFAULT_SYSTEM_OPTIONS } fro
 export { ClientConfiguration } from "./config/ClientConfiguration";
 // Account
 export { AccountInfo } from "./account/AccountInfo";
-export { IdToken } from "./account/IdToken";
-export { IdTokenClaims } from "./account/IdTokenClaims";
+export { JwtToken } from "./account/JwtToken";
+export { TokenClaims } from "./account/TokenClaims";
```
sangonzal, line 11: This is a breaking change for common, and thus we would bump the major version. Not sure if a name change is worth a major version bump - should perhaps be discussed as a team.

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -34,12 +34,10 @@ export class RefreshTokenClient extends BaseClient {
         );
 
         responseHandler.validateTokenResponse(response.body);
-        const tokenResponse = responseHandler.handleServerTokenResponse(
+        return await responseHandler.handleServerTokenResponse(
```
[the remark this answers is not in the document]
pkanher617: I've handled this in a later PR, this PR only covers the authorization code 2 legged flow, not the refresh flow.

lib/msal-common/src/index.ts
pkanher617: msal-common is an internal package, and thus I don't think it really matters whether or not this is bumped as a major version. Happy to discuss this as a team but this shouldn't be a blocker.

jasonnutter: Yes, but beacuse its a dependency, new versions can be picked up in customer apps, so we should absolutely avoid breaking changes, in my opinion.

This can be addressed by exporting both:

```suggestion
export { JwtToken } from "./account/JwtToken";
export { JwtToken as IdToken } from "./account/JwtToken";
export { TokenClaims } from "./account/TokenClaims";
export { TokenClaims as IdTokenClaims  } from "./account/TokenClaims";
```

lib/msal-common/src/account/JwtToken.ts
jasonnutter: You're right, but generally speaking, I personally try to avoid leaking implementation details, including in names.

Also, JwtToken is redundant. Maybe `AuthToken`?

pkanher617: done.

lib/msal-common/src/index.ts
pkanher617: done.

lib/msal-common/src/client/DeviceCodeClient.ts
```diff
@@ -47,12 +47,14 @@ export class DeviceCodeClient extends BaseClient {
 
         // Validate response. This function throws a server error if an error is returned by the server.
         responseHandler.validateTokenResponse(response);
-        const tokenResponse = responseHandler.handleServerTokenResponse(
+        return await responseHandler.handleServerTokenResponse(
             response,
-            this.authority
+            this.authority,
+            "",
+            "",
```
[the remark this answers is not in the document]
pkanher617: done.

PR #2154, opened 2020-08-15
AT Proof-Of-Possession #3: Browser Implementation of token signing and key storage

pkanher617:
This PR is a follow up to PR #2153, and includes the `msal-browser` implementation of the token signing and key storage APIs introduced in the previous code.

Specifically, this PR includes the following:

- Browser implementation of key storage using IndexedDB
- Browser implementation of jwt token signing

### DIFF
--- a/lib/msal-browser/src/cache/DatabaseStorage.ts
+++ b/lib/msal-browser/src/cache/DatabaseStorage.ts
@@ -1,0 +1,84 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+interface IDBOpenDBRequestEvent extends Event {
+    target: IDBOpenDBRequest & EventTarget;
+}
+
+interface IDBOpenOnUpgradeNeededEvent extends IDBVersionChangeEvent {
+    target: IDBOpenDBRequest & EventTarget;
+}
+
+interface IDBRequestEvent extends Event {
+    target: IDBRequest & EventTarget;
+}
+
+/**
+ * Storage wrapper for IndexedDB storage in browsers: https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API
+ */
+export class DatabaseStorage<T>{
+    private _db : IDBDatabase;
+    private _dbName: string;
+    private _tableName: string;
+    private _version: number;
+
+    constructor(dbName: string, tableName: string, version: number) {
+        this._dbName = dbName;
+        this._tableName = tableName;
+        this._version = version;
+    }
+
+    /**
+     * Opens IndexedDB instance.
+     */
+    async open(): Promise<void> {
+        return new Promise((resolve, reject) => {
+            // TODO: Add timeouts?
+            const openDB = window.indexedDB.open(this._dbName, this._version);
+            openDB.addEventListener("upgradeneeded", (e: IDBOpenOnUpgradeNeededEvent) => {
+                e.target.result.createObjectStore(this._tableName);
+            });
+            openDB.addEventListener("success", (e: IDBOpenDBRequestEvent) => {
+                this._db = e.target.result;
+                resolve();
+            });
+
+            openDB.addEventListener("error", error => reject(error));
+        });
+    }
+
+    /**
+     * Retrieves item from IndexedDB instance.
+     * @param key 
+     */
+    async get(key: string): Promise<T> {
+        return new Promise((resolve, reject) => {
+            // TODO: Add timeouts?
+            const transaction = this._db.transaction([this._tableName], "readonly");
+
+            const objectStore = transaction.objectStore(this._tableName);
+            const dbGet = objectStore.get(key);
+            dbGet.addEventListener("success", (e: IDBRequestEvent) => resolve(e.target.result));
+            dbGet.addEventListener("error", e => reject(e));
+        });
+    }
+
+    /**
+     * Adds item to IndexedDB under given key
+     * @param key 
+     * @param payload 
+     */
+    async put(key: string, payload: T): Promise<T> {
+        return new Promise((resolve: any, reject: any) => {
+            // TODO: Add timeouts?
+            const transaction = this._db.transaction([this._tableName], "readwrite");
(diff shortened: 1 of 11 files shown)


lib/msal-browser/src/crypto/CryptoOps.ts
```diff
@@ -26,23 +34,20 @@ export class CryptoOps implements ICrypto {
     private static EXTRACTABLE: boolean = true;
     private static POP_HASH_LENGTH = 43; // 256 bit digest / 6 bits per char = 43
 
+    private static DB_VERSION = 1;
+    private static DB_NAME = "msal.db";
+    private static TABLE_NAME =`${CryptoOps.DB_NAME}.keys`;
```
tnorling, line 39: These should be in constants

lib/msal-browser/src/cache/DatabaseStorage.ts
```diff
@@ -0,0 +1,84 @@
...
+}
+
+interface IDBRequestEvent extends Event {
+    target: IDBRequest & EventTarget;
+}
+
+/**
+ * Storage wrapper for IndexedDB storage in browsers: https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API
+ */
+export class DatabaseStorage<T>{
```
tnorling, line 21: Since someone might want to use IndexedDB as a storage option in the future, should we have this be an implementation of CacheManager? I understand there's complexities around async storage but curious if others think it's worth the investment to figure it out now instead of refactoring later?

lib/msal-browser/src/crypto/CryptoOps.ts
pkanher617: These aren't used outside the context of the crypto APIs, so I chose to keep them here.

lib/msal-browser/src/cache/DatabaseStorage.ts
DarylThayil, line 21: Nit Pick Could we call this IndexDatabaseStorage to be more specific

lib/msal-browser/src/crypto/BrowserCrypto.ts
```diff
@@ -177,6 +207,44 @@ export class BrowserCrypto {
...
     }
 
+    /**
+     * IE Helper function for importKey
+     * @param key 
+     * @param format 
+     * @param extractable 
+     * @param usages 
+     */
+    private async msCryptoImportKey(keyBuffer: ArrayBuffer, extractable: boolean, usages: Array<KeyUsage>): Promise<CryptoKey> {
```
jmckennon, line 217: Should all of the IE stuff be its own file?

lib/msal-browser/src/cache/DatabaseStorage.ts
jasonnutter: That is leaking implementation details, which I think we should avoid doing when naming things when possible.

lib/msal-browser/src/crypto/BrowserCrypto.ts
```diff
@@ -77,14 +75,41 @@ export class BrowserCrypto {
...
+            this.msCryptoImportKey(keyBuffer, extractable, usages) 
+            : window.crypto.subtle.importKey(KEY_FORMAT_JWK, key, this._keygenAlgorithmOptions, extractable, usages);
+    }
+
+    /**
+     * Signs given data with given key
+     * @param key 
+     * @param data 
+     */
+    async sign(key: CryptoKey, data: ArrayBuffer): Promise<ArrayBuffer> {
```
DarylThayil, line 107: nit pick
I wonder if this could be more specific "signToken"

DarylThayil:
mostly looks good, I was rely on @jasonnutter  for final sign off

lib/msal-browser/src/cache/DatabaseStorage.ts
pkanher617: I think async storage will be a separate discussion. For now I think I will leave this as a specific implementation in our lib.

lib/msal-browser/src/crypto/BrowserCrypto.ts
pkanher617: I was trying to keep naming consistent with the browser crypto APIs.

lib/msal-browser/src/crypto/BrowserCrypto.ts
pkanher617: I can separate it into it's own file for readability, I thought it would be easier to have all the crypto APIs in a single file.

lib/msal-browser/src/cache/DatabaseStorage.ts
```diff
@@ -0,0 +1,84 @@
...
+            openDB.addEventListener("error", error => reject(error));
+        });
+    }
+
+    /**
+     * Retrieves item from IndexedDB instance.
+     * @param key 
+     */
+    async get(key: string): Promise<T> {
+        return new Promise((resolve, reject) => {
```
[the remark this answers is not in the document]
pkanher617: done.

lib/msal-browser/src/cache/DatabaseStorage.ts
```diff
@@ -0,0 +1,84 @@
...
+}
+
+/**
+ * Storage wrapper for IndexedDB storage in browsers: https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API
+ */
+export class DatabaseStorage<T>{
+    private _db : IDBDatabase;
+    private _dbName: string;
+    private _tableName: string;
+    private _version: number;
```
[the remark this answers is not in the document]
pkanher617: done.

lib/msal-browser/src/cache/DatabaseStorage.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    constructor(dbName: string, tableName: string, version: number) {
+        this._dbName = dbName;
+        this._tableName = tableName;
+        this._version = version;
+    }
+
```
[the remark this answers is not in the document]
pkanher617: resolved offline.

PR #2157, opened 2020-08-17
[msal-node][msal-common] Support for On-behalf-of flow

sangonzal:
Adds support for [on-behalf-of flow](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-on-behalf-of-flow).


samples/msal-node-samples/on-behal-0f/web-app/index.js
```diff
@@ -0,0 +1,80 @@
...
+
+const SERVER_PORT = process.env.PORT || 3000;
+
+// Before running the sample, you will need to replace the values in the config, 
+// including the clientSecret
+const config = {
+    auth: {
+        clientId: "12d77c73-d09d-406a-ae0d-3d4e576f7d9b",
+        authority: "https://login.microsoftonline.com/90b8faa8-cc95-460e-a618-ee770bee1759",
+        clientSecret: "",
```
jasonnutter, line 18: I would use the `dotenv` package with `.env` file to have the secret loaded via environment variables.

jasonnutter: https://www.npmjs.com/package/dotenv

lerna.json
```diff
@@ -17,7 +17,7 @@
   "packages": [
     "lib/*",
     "experimental/*",
-    "samples/**/*"
+    "samples/**/**/*"
```
pkanher617, line 20: shouldn't be needed

samples/msal-node-samples/on-behal-0f/web-app/index.js
pkanher617: This is probably something we should do across the sample framework (@technical-boy)

lib/msal-common/src/cache/entities/AccountEntity.ts
```diff
@@ -143,6 +146,7 @@ export class AccountEntity {
 
         account.environment = env;
         account.realm = idToken.claims.tid;
+        account.oboAssertion = oboAssertion;
```
sameerag, line 149: Is this not needed for ADFS cases?

lib/msal-common/src/client/OnBehalfOfClient.ts
```diff
@@ -0,0 +1,173 @@
...
+            oboAssertion: request.oboAssertion
+        };
+
+        const credentialCache: CredentialCache = this.cacheManager.getCredentialsFilteredBy(accessTokenFilter);
+        const idTokens = Object.keys(credentialCache.idTokens).map(key => credentialCache.idTokens[key]);
+        // When acquiring a token on behalf of an application, there might not be an id token in the cache
+        if (idTokens.length < 1) {
+            return null;
+        }
+        return idTokens[0] as AccessTokenEntity;
```
sameerag, line 112: Is this a typo? This should be `IdTokenEntity`.

lib/msal-common/src/client/OnBehalfOfClient.ts
```diff
@@ -0,0 +1,173 @@
...
+        const credentialCache: CredentialCache = this.cacheManager.getCredentialsFilteredBy(accessTokenFilter);
+        const accessTokens = Object.keys(credentialCache.accessTokens).map(key => credentialCache.accessTokens[key]);
+        if (accessTokens.length < 1) {
+            return null;
+        }
+        return accessTokens[0] as AccessTokenEntity;
+    }
+    
+    private readIdTokenFromCache(request: OnBehalfOfRequest): IdTokenEntity {
+        const accessTokenFilter: CredentialFilter = {
```
sameerag, line 98: Same as below, shouldn't this be 'idTokenFilter'?

lib/msal-common/src/client/OnBehalfOfClient.ts
sangonzal: Yep, this should be IdTokenEntity. I've updated. Same as above.

lib/msal-common/src/cache/entities/AccountEntity.ts
sangonzal: Yes good point, I have added for ADFS2019 accounts as well.

sameerag (approved):

lerna.json
sangonzal: Yep, you're right, reverted this change.

samples/msal-node-samples/on-behal-0f/web-app/index.js
sangonzal: Since all the samples are in the process of being rewritten, I'll let @technical-boy decide how to handle storing secrets in a way that works for all of the samples/tests.

PR #2169, opened 2020-08-19
Client Capabilities Support

tnorling:
Adds support for ClientCapabilities.

Developer provides list of capabilities on auth.config object which MSAL internally translates into a claims request and sent to both /authorize and /token endpoints on every request

Providing claims request to a silent request will result in a network call.


tnorling:
cc. @AndreiZe

jasonnutter:
I think this should only be set on the request, and not in the config, given MSAL can be used for many resources, and not all of them will have this enabled (or need the same value).

tnorling:
> I think this should only be set on the request, and not in the config, given MSAL can be used for many resources, and not all of them will have this enabled (or need the same value).

This design is in accordance with the spec. There's also a slight important nuance between providing the values on the config vs. on the request. When provided on the request the cache is automatically skipped, which is not the case when only provided in the config.

PR #2183, opened 2020-08-24
getRedirectResponseHash checks for interaction in progress

tnorling:
After a successful login hitting the back button will go back to eSTS which then redirects back to the app with a new response hash. However, in this scenario all the temporary cache items have already been cleared from the first success so MSAL.js will potentially throw a variety of errors related to expected cache items being missing. This PR aims to address this by checking if interaction is in progress before proceeding in processing the hash.

Open Question for the team: In the scenario where a valid response comes back in the hash but we do not have temp cache items available (suggesting either the request didn't originate on this domain OR the original request was already processed) should we silently return null (as shown in this PR), should we throw some sort of error or should we log something to the console?


lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -166,7 +166,12 @@ export abstract class ClientApplication {
                 return null;
             } else {
                 BrowserUtils.clearHash();
-                return responseHash;
+                if (this.interactionInProgress()) {
```
pkanher617, line 169: We should do this check at the beginning, so that we don't have to do any cache lookups.

tnorling: Issue is that we want to clear the hash if we aren't expecting it but we don't want to clear it if interactionType != Redirect. So I think it needs to stay here unless anyone has a suggestion for solving that problem.

pkanher617: > Issue is that we want to clear the hash if we aren't expecting it

I don't think we should be clearing the hash if we aren't expecting it, that seems like it could break non-msal hashes.

tnorling: That's why this check is being done here. Because by this point we've already determined that the hash **is** an msal hash

DarylThayil:
I think I like not throwing an error here because we don't really want people to have to write code to catch an error that could be just an anomaly.  For now I think the log is good

lib/msal-browser/src/app/ClientApplication.ts
tnorling: Resolved offline. Interaction check only checks for this `clientId`, in theory a parallel PCA using a different clientId could have initiated the request in which case we would not want to clear the hash. Moved to the top.

pkanher617: Resolved offline - multiple app use cases may break if we clear, so probably better to move the check before we clear.

PR #2187, opened 2020-08-25
Update CDN for msal-browser and msal

pkanher617:


.npmrc
```diff
@@ -1 +1 @@
-//registry.npmjs.org/:_authToken=${NODE_AUTH_TOKEN}
\ No newline at end of file
+#//registry.npmjs.org/:_authToken=${NODE_AUTH_TOKEN}
```
jasonnutter, line 1: Why is this commented out?

pkanher617: This is causing issues in builds. We don't need this to be enabled currently.

PR #2189, opened 2020-08-25
Separate Cache Lookup and Token Refresh in SilentFlowClient

tnorling:
This PR mainly refactors the SilentFlowClient and provides 2 additional APIs

Changes include:
- Addition of `acquireCachedToken` and `refreshToken` function in `SilentFlowClient`
- Original `acquireToken` function refactored to call the 2 new APIs
- Move the token cache lookups to `CacheManager`
- Add `getCacheRecord` function to `CacheManager` to do all the token cache lookups in one place


lib/msal-common/src/cache/entities/IdTokenEntity.ts
```diff
@@ -53,6 +57,27 @@ export class IdTokenEntity extends CredentialEntity {
     }
 
     /**
+     * fetches idToken from cache if present
+     * @param request
+     */
+    static readIdTokenFromCache(cacheManager: CacheManager, clientId: string, account: AccountInfo, realm?: string): IdTokenEntity {
```
pkanher617, line 63: Maybe these should be helpers on CacheManager instead?

lib/msal-common/src/client/SilentFlowClient.ts
```diff
@@ -35,6 +31,25 @@ export class SilentFlowClient extends BaseClient {
      * @param request
      */
     public async acquireToken(request: SilentFlowRequest): Promise<AuthenticationResult> {
+        let result: AuthenticationResult;
+        try {
+            result = this.acquireCachedToken(request);
```
jasonnutter, line 36: Only needs to be `return this.acquireCachedToken(request)`;

lib/msal-common/src/cache/entities/IdTokenEntity.ts
```diff
@@ -53,6 +57,27 @@ export class IdTokenEntity extends CredentialEntity {
...
+        if (StringUtils.isEmpty(realm)) {
+            const accountKey: string = AccountEntity.generateAccountCacheKey(account);
+            const cachedAccount = cacheManager.getAccount(accountKey);
+            realm = cachedAccount && cachedAccount.realm ? cachedAccount.realm : realm;
```
jasonnutter, line 67: Would prefer to not reassign `realm` (even if its empty).

jasonnutter: The `isEmpty` check is only needed for the `realm` value, the preceding two lines don't need to be guarded by it.

pkanher617: the cached account is not needed if the realm is not empty. this saves a cache lookup.

jasonnutter: What if a helper function was used?

```js
const accountRealm = StringUtils.isEmpty(realm) ? this.getRealmForAccount(account) : realm;
```

lib/msal-common/src/cache/entities/IdTokenEntity.ts
tnorling: Yeah I considered that as well. I initially thought this made more organizational sense but after thinking more on it I'm inclined to agree with you

lib/msal-common/src/client/SilentFlowClient.ts
tnorling: Done.

lib/msal-common/src/cache/entities/IdTokenEntity.ts
tnorling: Done.

lib/msal-common/src/cache/entities/IdTokenEntity.ts
tnorling: Done.

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -168,10 +186,71 @@ export abstract class CacheManager implements ICacheManager {
...
+     * @param inputRealm 
+     */
+    getAccessTokenEntity(clientId: string, account: AccountInfo, scopes: ScopeSet): AccessTokenEntity | null {
+        const accessTokenFilter: CredentialFilter = {
+            homeAccountId: account.homeAccountId,
+            environment: account.environment,
+            credentialType: CredentialType.ACCESS_TOKEN,
+            clientId,
+            realm: account.tenantId,
+            target: scopes.printScopesLowerCase()
```
sameerag, line 225: We are storing `scopes` in lower case and `look-up` in lower case. We need to address this, as I observed in cache-compat test cases, other libraries store the case as is. We may need to talk about this.

My take: Look up should be case insensitive but storing should be as is.

tnorling: 100% agree. As this PR is scoped to a refactor, let's take functional changes in a follow up PR

PR #2190, opened 2020-08-25
Broker #3: Popup requests from embedded application

pkanher617:
Adds the ability to make popup requests from an embedded application by sending a message to the top frame broker.


lib/msal-common/src/client/BrokerAuthorizationCodeClient.ts
```diff
@@ -0,0 +1,168 @@
...
+import { StringUtils } from "../utils/StringUtils";
+import { ClientAuthError } from "../error/ClientAuthError";
+import { ResponseHandler } from "../response/ResponseHandler";
+import { BrokerAuthenticationResult } from "../response/BrokerAuthenticationResult";
+import { GrantType } from "../utils/Constants";
+
+/**
+ * Oauth2.0 Authorization Code client implementing the broker protocol for browsers.
+ */
+export class BrokerAuthorizationCodeClient extends AuthorizationCodeClient {
```
jasonnutter, line 20: I would argue against this approach, as now we have to remember to add any new parameters into two classes. Could the existing client be augmented to easily support the need broker parameters instead?

pkanher617: I agree. I can make some updates to make this more streamlined so we have less copied code.

pkanher617: done.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -80,7 +89,7 @@ export class PublicClientApplication implements IPublicClientApplication {
 
     // Broker Objects
     private embeddedApp: BrokerClient;
-    private broker: BrokerManager;
+    private broker: boolean;
```
tnorling, line 92: Can we call this `isBroker` instead?

tnorling: Is this even needed since we have the config option `actAsBroker`?

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -123,22 +132,85 @@ export class PublicClientApplication implements IPublicClientApplication {
...
+            this.listenForBrokerMessage();
         } else if (this.config.system.brokerOptions.allowBrokering) {
-            this.embeddedApp = new BrokerClient(this.config.system.brokerOptions, this.logger, this.config.auth.clientId,  version);
+            this.embeddedApp = new BrokerClient(this.config.system.brokerOptions, this.logger, this.config.auth.clientId, version, this.browserStorage);
             this.logger.verbose("Acting as child");
             this.embeddedApp.initiateHandshake();
         }
     }
 
+    private listenForBrokerMessage(): void {
```
tnorling, line 152: Thought we landed on having a `BrokerManager` class that extends `PublicClientApplication`. Did I misunderstand or did that decision change?

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -121,6 +122,42 @@ export class ResponseHandler {
...
+        const cacheRecord = this.generateCacheRecord(serverTokenResponse, idTokenObj, authority, requestStateObj && requestStateObj.libraryState);
+        const refreshTokenRecord: CacheRecord = {
+            accessToken: null,
+            idToken: null,
+            refreshToken: cacheRecord.refreshToken,
+            account: null
+        };
+
+        // Save refresh token
+        this.cacheStorage.saveCacheRecord(refreshTokenRecord);
```
tnorling, line 152: Can't you just use `this.cacheStorage.saveCredential(cacheRecord.refreshToken)` instead?

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: Good point, I did this temporarily.

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: Just doing things incrementally to see what changes need to be made.

lib/msal-common/src/response/ResponseHandler.ts
pkanher617: Good point will change.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -49,15 +49,20 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
+     */
     private initializeBrokering(): void {
         if (this.config.system.brokerOptions.actAsBroker) {
-            this.broker = new BrokerManager(this.config.system.brokerOptions, this.logger, version);
-            this.logger.verbose("Acting as Broker");
-            this.broker.listenForMessage();
+            this.broker = new BrokerClientApplication(this.config);
         } else if (this.config.system.brokerOptions.allowBrokering) {
-            this.embeddedApp = new BrokerClient(this.config.system.brokerOptions, this.logger, this.config.auth.clientId,  version);
+            this.embeddedApp = new EmbeddedClientApplication(this.config.system.brokerOptions, this.logger, this.config.auth.clientId, version, this.browserStorage);
```
tnorling, line 59: Any reason we aren't just passing the whole config object here?

lib/msal-browser/src/broker/EmbeddedClientApplication.ts
```diff
@@ -4,72 +4,107 @@
...
+            this.brokeringEnabled = true;
         } catch (e) {
             this.logger.error(e);
             this.brokeringEnabled = false;
             throw e;
         }
     }
 
+    private async preflightBrokerRequest(): Promise<void> {
+        if (!this.brokeringEnabled) {
```
tnorling, line 62: Can we call this something like `brokerCommunicationEstablished` ? I was initially confused why we were checking for falsey `brokeringEnabled`

lib/msal-common/src/client/BrokerAuthorizationCodeClient.ts
```diff
@@ -0,0 +1,81 @@
...
+        const tokenResponse = responseHandler.handleBrokeredServerTokenResponse(response.body, this.authority, cachedNonce, cachedState);
+
+        return tokenResponse;
+    }
+
+    /**
+     * This API validates the `AuthorizationCodeUrlRequest` and creates a URL
+     * @param request
+     */
+    protected createAuthCodeUrlQueryString(request: AuthorizationUrlRequest): string {
```
tnorling, line 52: Why override the existing function? Can't we just add the params we need to `AuthorizationUrlRequest` and add them to the existing function on `AuthorizationCodeClient`?

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: Mostly because we aren't really using much from there aside from clientid and brokerOptions, but we can pass the entire config if it's easier.

lib/msal-common/src/client/BrokerAuthorizationCodeClient.ts
pkanher617: Subclassing and overriding allows us to keep the broker code separate from the general authorization code flow. This way we don't affect msal-node flows either.

lib/msal-browser/src/app/PublicClientApplication.ts
tnorling: I think it may make it cleaner and more future-proof if we do need to access more config options in the future.

lib/msal-browser/src/broker/BrokerClientApplication.ts
```diff
@@ -0,0 +1,125 @@
...
+                case BrokerMessageType.AUTH_REQUEST:
+                    this.logger.verbose("Broker auth request received");
+                    return await this.handleBrokerAuthRequest(clientMessage);
+                default:
+                    return;
+            }
+        }
+    }
+
+    /* eslint-disable */
```
DarylThayil, line 58: what are we disabling? Can you disable a specific rule not just the whole eslint?

lib/msal-browser/src/broker/BrokerClientApplication.ts
```diff
@@ -0,0 +1,125 @@
...
+    /**
+     * Handle a broker handshake request from a child.
+     * @param clientMessage 
+     */
+    private async handleBrokerHandshake(clientMessage: MessageEvent): Promise<void> {
+        const validMessage = BrokerHandshakeRequest.validate(clientMessage);
+        this.logger.verbose(`Broker handshake validated: ${validMessage}`);
+        const brokerHandshakeResponse = new BrokerHandshakeResponse(version);
+
+        // @ts-ignore
```
DarylThayil, line 68: Same can we ignore a specific rule

lib/msal-browser/src/broker/EmbeddedClientApplication.ts
```diff
@@ -4,72 +4,107 @@
...
+                throw BrowserAuthError.createBrokeringDisabledError(e);
+            }
+        }
+    }
+
     async sendPopupRequest(request: PopupRequest): Promise<AuthenticationResult> {
-        return this.sendRequest<AuthenticationResult>(request, InteractionType.POPUP, 6000);
+        await this.preflightBrokerRequest();
+
+        const brokerAuthResultMessage = await this.sendRequest(request, InteractionType.POPUP, 60000);
```
DarylThayil, line 76: Please define this number somewhere and reference it. even if just at the top of the file
```js
const POPUP_TIMEOUT_LIMIT = 6000
```

DarylThayil: Should this be configurable?

lib/msal-browser/src/app/PublicClientApplication.ts

lib/msal-browser/src/broker/EmbeddedClientApplication.ts
pkanher617: done.

lib/msal-browser/src/broker/EmbeddedClientApplication.ts
pkanher617: done.

lib/msal-common/src/client/BrokerAuthorizationCodeClient.ts
tnorling: After some more consideration I've come around to this. I think this approach makes sense.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -49,15 +49,20 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
+     */
     private initializeBrokering(): void {
         if (this.config.system.brokerOptions.actAsBroker) {
-            this.broker = new BrokerManager(this.config.system.brokerOptions, this.logger, version);
-            this.logger.verbose("Acting as Broker");
-            this.broker.listenForMessage();
+            this.broker = new BrokerClientApplication(this.config);
         } else if (this.config.system.brokerOptions.allowBrokering) {
-            this.embeddedApp = new BrokerClient(this.config.system.brokerOptions, this.logger, this.config.auth.clientId,  version);
+            this.embeddedApp = new EmbeddedClientApplication(this.config, this.logger, version, this.browserStorage);
```
tnorling, line 59: One more nitpick: `version` can just be imported directly in `EmbeddedClientApplication`

lib/msal-browser/src/broker/BrokerAuthResult.ts
```diff
@@ -0,0 +1,30 @@
+import { BrokerMessage } from "./BrokerMessage";
+import { BrokerAuthenticationResult } from "@azure/msal-common";
```
sameerag, line 2: Nit: We have two files named `BrokerAuthResult` and `BrokerAuthenticationResult`, is the difference `common` and `browser`?

tnorling: We should probably rename `BrokerAuthResult` to `BrokerAuthResponse`

lib/msal-browser/src/broker/BrokerClientApplication.ts
```diff
@@ -0,0 +1,125 @@
...
+        window.addEventListener("message", this.handleBrokerMessage.bind(this));
+    }
+
+    /**
+     * 
+     * @param message 
+     */
+    private async handleBrokerMessage(message: MessageEvent): Promise<void> {
+        // Check that message is a BrokerHandshakeRequest
+        const clientMessage = BrokerMessage.validateMessage(message);
```
sameerag, line 43: Are we not throwing and logging the error for invalid messages?This applies to all broker requests and responses.

lib/msal-browser/src/broker/BrokerClientApplication.ts
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    /* eslint-disable */
+    /**
+     * Handle a broker handshake request from a child.
+     * @param clientMessage 
+     */
+    private async handleBrokerHandshake(clientMessage: MessageEvent): Promise<void> {
+        const validMessage = BrokerHandshakeRequest.validate(clientMessage);
+        this.logger.verbose(`Broker handshake validated: ${validMessage}`);
```
sameerag, line 65: What happens if the `validMessage` above is `null`?

jasonnutter: I don't think that should happen.

lib/msal-browser/src/broker/BrokerClientApplication.ts
jasonnutter: No, because the window could receive messages from other frames that are unrelated to brokering, and those should be silently ignored.

lib/msal-browser/src/broker/BrokerClientApplication.ts
jasonnutter: Typescript isn't liking the request signature for `clientMessage.source.postMessage`, which is why this is here.

lib/msal-browser/src/broker/BrokerClientApplication.ts
pkanher617: `handleBrokerHandshake` is only called if the clientMessage is a truthy value.

lib/msal-browser/src/broker/BrokerClientApplication.ts
sameerag: Got it!

lib/msal-browser/src/broker/BrokerMessage.ts
```diff
@@ -16,7 +16,7 @@ export abstract class BrokerMessage {
 
     static validateMessage(message: MessageEvent): MessageEvent|null {
         if (message.data && message.data.messageType) {
-            if (message.data.messageType in BrokerMessageType) {
+            if (Object.values(BrokerMessageType).indexOf(message.data.messageType) > -1) {
```
jasonnutter, line 19: Object.values not supported in all browsers

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -127,6 +127,49 @@ export class ResponseHandler {
...
+            }
+        }
+
+        // save the response tokens
+        let requestStateObj: RequestStateObject = null;
+        if (!StringUtils.isEmpty(cachedState)) {
+            requestStateObj = ProtocolUtils.parseRequestState(this.cryptoObj, cachedState); 
+        }
+
+        const cacheRecord = this.generateCacheRecord(serverTokenResponse, idTokenObj, authority, requestStateObj && requestStateObj.libraryState);
```
sameerag, line 157: There are some changes that will come here with FOCI. Currently they are made in `generateCacheRecord` and hence may not impact this code. We should test though once both features are integrated.

lib/msal-browser/src/broker/BrokerMessage.ts
pkanher617: done.

lib/msal-browser/src/app/PublicClientApplication.ts

PR #2192, opened 2020-08-26
[msal-browser] Add SsoSilentRequest for ssoSilent

jo-arroyo:
This PR changes the request type for `ssoSilent` for msal-browser. The type `SsoSilentRequest` has scopes as an optional parameter, as scopes will not be passed when `ssoSilent` is called. 

Samples and tests have also been updated to reflect this change.


lib/msal-browser/src/request/SsoSilentRequest.ts
```diff
@@ -0,0 +1,34 @@
...
+ * - loginHint                  - Can be used to pre-fill the username/email address field of the sign-in page for the user, if you know the username/email address ahead of time. Often apps use this parameter during re-authentication, having already extracted the username from a previous sign-in using the preferred_username claim.
+ * - sid                        - Session ID, unique identifier for the session. Available as an optional claim on ID tokens.
+ * - domainHint                 - Provides a hint about the tenant or domain that the user should use to sign in. The value of the domain hint is a registered domain for the tenant.
+ * - extraQueryParameters       - String to string map of custom query parameters.
+ * - nonce                      - A value included in the request that is returned in the id token. A randomly generated unique value is typically used to mitigate replay attacks.
+ */
+export type SsoSilentRequest = Omit<AuthorizationUrlRequest, "scopes"> & {
```
pkanher617, line 32: Didn't know about Omit, did some reading, would it be better to use Partial? https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype

jo-arroyo: After discussions with Prithvi, decided not to use Partial in case of future changes to BaseRequest.

technical-boy: I didn't know about these Utility types, very cool!

jasonnutter:
@jo-arroyo FYI merge conflicts

PR #2202, opened 2020-08-27
Address msal-node tsdx warning

tnorling:
Addresses warning when building msal-node:

`[tsdx]: Your rootDir is currently set to "./". Please change your rootDir to "./src".
TSDX has deprecated setting tsconfig.compilerOptions.rootDir to "./" as it caused buggy output for declarationMaps and occassionally for type declarations themselves.
You may also need to change your include to remove "test", which also caused declarations to be unnecessarily created for test files.`

Also hopefully fixes intermittent build error:

`EBUSY: resource busy or locked, copyfile 'D:\a\1\s\lib\msal-node\dist\src\utils\NetworkUtils.d.ts' -> 'D:\a\1\s\lib\msal-node\dist\utils\NetworkUtils.d.ts'`


lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -24,10 +24,10 @@ import {
 import { Configuration, buildAppConfiguration } from '../config/Configuration';
 import { CryptoProvider } from '../crypto/CryptoProvider';
 import { Storage } from '../cache/Storage';
-import { version } from '../../package.json';
 import { Constants as NodeConstants } from './../utils/Constants';
 import { TokenCache } from '../cache/TokenCache';
 import { ClientAssertion } from "../client/ClientAssertion";
+const pjson = require("../../package.json");
```
sameerag, line 30: I thought we are avoiding `require` except in tests.

technical-boy: +1

tnorling: Hmm then we may need to think of something because we can't import package.json from outside the root

pkanher617: I think it's ok for this scenario. Did some testing and there isn't any way this will work currently because `tsdx` has issues with `.` as the rootDir. This is one reason I don't think we should have used `tsdx`.

tnorling: So even adding a `@ts-ignore` in order to import, silently changes the `rootDir` back to `./` in the background. We'll move forward with the `require` for now and revisit later

PR #2206, opened 2020-08-28
[msal-core] Fix cached ID Token retrieval for acquireTokenSilent

### LINKED ISSUE #2185: MSAL 1.4.0 was released, and is causing infinite redirection loop
## Library
- [x] `msal@1.4.0` 

## Framework
React

## Description
We noticed an upgrade of msal to the version 1.4.0 today in our dev environment. It was released at the same day (25. Aug. 2020). The login with underlying msal library does not work anymore and is ending in infinite redirection loop to login.microsoft.com. 

More, after short search on this project side I cannot find any migration guide or release note to the released version. Was this version released by accident? Could someone clarify?

For now we changed the version back to the latest 1.3.x and everything works fine.

## Security

## Regression
Version:

## MSAL Configuration

```js
```

## Reproduction steps
```js
```

## Browsers/Environment
- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari
- [ ] IE

hectormmg:
This PR:
- Adds logic to always try to include an ID Token in the `acquireTokenSilent` auth response, regardless of `response_type`
- Adds `getIdToken` method to the `AuthCache` class to support retrieving ID Tokens from the cache
- Adds several utility private methods to UserAgentApplication dedicated to building an appropriate cache auth response based on the response type
- Adds exhaustive testing for the above updates to behavior
- Fixes #2185 which is caused by a cache miss when response_type is "id_token"


lib/msal-core/src/cache/AuthCache.ts
```diff
@@ -206,6 +208,31 @@ export class AuthCache extends BrowserStorage {// Singleton
         return results;
     }
 
+    /**
+     * Get idToken in the cache
+     * @param  
+     */
+    getIdToken(clientId: string, homeAccountIdentifier: string): AccessTokenValue {
+        const key = Object.keys(window[this.cacheLocation]).find((key) => {
+            const keyMatches = key.match(clientId) && key.match(homeAccountIdentifier);
```
tnorling, line 217: The idToken cacheKey should be a standard format of `msal.clientId.idtoken`

Since this is in AuthCache you can just call `this.generateCacheKey(PersistentCacheKeys.IDTOKEN, true)` to get the idtoken key

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1356,9 +1355,49 @@ export class UserAgentApplication {
      */
     private getCachedToken(serverAuthenticationRequest: ServerRequestParameters, account: Account): AuthResponse {
         this.logger.verbose("GetCachedToken has been called");
-        let accessTokenCacheItem: AccessTokenCacheItem = null;
         const scopes = serverAuthenticationRequest.scopes;
+        const authority = (serverAuthenticationRequest.authority) ? serverAuthenticationRequest.authority : this.authority;
```
jasonnutter, line 1359:
```suggestion
        const authority = serverAuthenticationRequest.authority || this.authority;
```

lib/msal-core/src/cache/AuthCache.ts
```diff
@@ -206,6 +208,24 @@ export class AuthCache extends BrowserStorage {// Singleton
...
+        const idTokenKey: AccessTokenKey = {
+            authority: authority,
+            clientId: clientId,
+            scopes: undefined,
+            homeAccountIdentifier: homeAccountIdentifier
+        };
+
+        const idTokenKeyString = JSON.stringify(idTokenKey);
+        const idToken = this.getItem(idTokenKeyString);
+        const idTokenValue = JSON.parse(idToken) as AccessTokenValue;
```
jasonnutter, line 224: What should happen if `JSON.parse` throws?

change/msal-2020-09-10-13-11-57-idtoken-cache-lookup.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Fix ID Token cache lookup",
```
tnorling, line 3: Can we put the PR number in here?

lib/msal-core/src/cache/AuthCache.ts
```diff
@@ -206,6 +208,24 @@ export class AuthCache extends BrowserStorage {// Singleton
...
+        
+        return new AccessTokenCacheItem(idTokenKey, idTokenValue);
```
tnorling, line 226: Can we add a comment here explaining why a function for getting an idToken is returning an object called AccessToken?

lib/msal-core/src/cache/AuthCache.ts
tnorling: IMO this should be in a try/catch and have a logger message + return null if it throws

lib/msal-core/src/cache/AuthCache.ts
tnorling, line 226: We should check the expiration and only return if it is not expired similar to the access token scenario

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1356,9 +1355,49 @@ export class UserAgentApplication {
...
+                expiresOn: null,
+                account: account,
+                accountState: aState,
+                fromCache: true
+            };
+
+            this.logger.verbose("Response generated and token set");
+        }
+        
+        return (idTokenObj) ? ResponseUtils.setResponseIdToken(authResponse, idTokenObj) : authResponse;
```
tnorling, line 1397: We need to assert that what was requested is being returned if it exists and is not expired. Couple scenarios to keep in mind which I do not see covered here:

1. If `id_token` or `id_token_token` is requested and there is no cached `idToken` or it is expired we need to return `null`
2. If token or `id_token_token` is requested and there is no cached `accessToken` or it is expired we need to return `null`

I think it may be better to make the above if/else block a switch on the `responseType` and in each case build the complete response, verify all the parts you need are there and return either response or `null`.

tnorling (changes requested):

lib/msal-core/src/cache/AuthCache.ts
technical-boy: What's your opinion on renaming AccessTokenCacheItem to TokenCacheItem? Other than the overhead involved, is there a philosophical/technical reason not to make it a more general type?

lib/msal-core/src/cache/AuthCache.ts
technical-boy: Implemented try/catch + returning null. The logger message was already there after the call to `getIdToken` in `acquireTokenSilent`. Updated the logic around the call as well to handle this possible exception.

change/msal-2020-09-10-13-11-57-idtoken-cache-lookup.json
technical-boy: Done!

lib/msal-core/src/cache/AuthCache.ts
technical-boy: Refactored the expiration validation a bit, now we remove the ID Token from the cache and return null if expired over the offset.

change/msal-2020-09-10-13-11-57-idtoken-cache-lookup.json
```diff
@@ -0,0 +1,8 @@
...
+  "packageName": "msal",
+  "email": "hectormgdev@gmail.com",
```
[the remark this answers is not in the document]
technical-boy: Fixed, thanks for catching this!

lib/msal-core/src/UserAgentApplication.ts
technical-boy: I think the latest changes should cover these observations/suggestions, thanks for the help!

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -1356,11 +1355,119 @@ export class UserAgentApplication {
...
+            case ResponseTypes.id_token_token:
+                this.logger.verbose("Returning Access Token and ID Token AuthResponse cache result");
+                return this.buildIdTokenTokenAuthResponse(idToken, serverAuthenticationRequest, scopes, account);
+            default: // ResponseTypes.token
+                this.logger.verbose("Returning Access Token AuthResponse cache result");
+                return this.buildTokenAuthResponse(idToken, serverAuthenticationRequest, scopes, account);
+        }
+    }
+
+    private buildTokenAuthResponse(idToken: IdToken, serverAuthenticationRequest: ServerRequestParameters, scopes: Array<string>, account: Account): AuthResponse {
```
tnorling, line 1378: Could we turn these into one generic static function on `ResponseUtils`?

technical-boy: I tried at first, but the refactor wasn't trivial. There are calls to private methods like `getCachedToken()` and `getAccountState`. I'll call those methods before calling these utility functions and pass the results as arguments.

technical-boy: Done!

technical-boy: Oh sorry, misread the part about making it a single method. Let me pass the switch into a utility function and handle it in ResponseUtils.

lib/msal-core/src/utils/ResponseUtils.ts
```diff
@@ -31,4 +34,28 @@ export class ResponseUtils {
...
+                    account: account,
+                    scopes: scopes,
+                    accountState: accountState
+                };
+                
+                authResponse = ResponseUtils.setResponseIdToken(authResponse, idToken);
+                return (authResponse.idToken) ? authResponse : null;
+            case ResponseTypes.id_token_token:
+                authResponse = ResponseUtils.setResponseIdToken(authResponse, idToken);
+                return (authResponse && authResponse.idToken) ? authResponse : null;
```
tnorling, line 53: should we also be checking that `authResponse.accessToken` is truthy here?

tnorling (approved):
One comment but otherwise looks good! Nice work

lib/msal-core/src/utils/ResponseUtils.ts
hectormmg: Adding the check, thanks for catching it. I'm fairly certain authResponse would be `null` if there was no access token, then `setResponseIdToken` would return null immediately if that was the case, making this line correctly return `null` in the end. However, I think it's a good idea to code defensively against future changes in `getCachedToken` that may not be caught by the current logic.

jmckennon (approved):
LGTM

jsmed-renson:
Thank you, @hectormmg and everyone else who helped to fix this! It seems to work perfectly. 💯

PR #2209, opened 2020-08-28
AT Proof-Of-Possession #4: Renewal of Proof-of-possession tokens

pkanher617:
This PR is a follow up to PR #2154, and includes the implementation of the refresh token flow for proof-of-possession access tokens.

### DIFF
--- a/lib/msal-common/src/client/RefreshTokenClient.ts
+++ b/lib/msal-common/src/client/RefreshTokenClient.ts
@@ -10,9 +10,10 @@ import { Authority, NetworkResponse } from "..";
 import { ServerAuthorizationTokenResponse } from "../response/ServerAuthorizationTokenResponse";
 import { RequestParameterBuilder } from "../request/RequestParameterBuilder";
 import { ScopeSet } from "../request/ScopeSet";
-import { GrantType } from "../utils/Constants";
+import { GrantType, AuthenticationType } from "../utils/Constants";
 import { ResponseHandler } from "../response/ResponseHandler";
 import { AuthenticationResult } from "../response/AuthenticationResult";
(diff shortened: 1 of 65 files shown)


lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -36,20 +37,24 @@ export class RefreshTokenClient extends BaseClient {
         responseHandler.validateTokenResponse(response.body);
         return await responseHandler.handleServerTokenResponse(
             response.body,
-            this.authority
+            this.authority,
+            "",
+            "",
```
sangonzal, line 42: These should perhaps be null instead of empty strings.

sangonzal (approved):
:shipit:

lib/msal-common/src/client/RefreshTokenClient.ts
jasonnutter: Semantically, both are the same (falsey). If we pass null, strictnullChecks would require the type to be `string | null`, which  results in downstream work to account for that.

sangonzal: Makes sense. I would argue that empty strings and null are two separate things, and that these values here are actually null as opposed to empty string and so the precision is worth the extra work. Ultimately its a nitpick either way, since you mentioned we are using StringUtils.isEmpty(), so functionally it doesn't matter.

sangonzal: I'll add that null is already used in some other places, so whichever one is picked should be consistent throughtout.

pkanher617: Fixed in the latest push, no longer sending empty/null params

PR #2223, opened 2020-09-02
Add telemetry header size limit

tnorling:
This PR addresses several things in regards to Telemetry:

- Replaces references to `ServerTelemetryCacheValue` to use the new `ServerTelemetryEntity` introduced in a previous release
- Replaces hard limit of 3 errors cached at a time to align with the spec implemented by all MSALs
- Header size is limited to 4KB and will contain an overflow flag in `platformFields`
- Currently no limit on how much is cached...should we add one? I believe it's unlikely this will become an issue
- After successful request to server, the data that was sent will be removed from cache, the data that was left behind will be retained.


lib/msal-common/src/telemetry/server/ServerTelemetryManager.ts
```diff
@@ -35,36 +37,41 @@ export class ServerTelemetryManager {
...
-        const failedRequests = lastRequests.failedRequests.join(SERVER_TELEM_CONSTANTS.VALUE_SEPARATOR);
-        const errors = lastRequests.errors.join(SERVER_TELEM_CONSTANTS.VALUE_SEPARATOR);
-        const platformFields = lastRequests.errorCount;
+        const maxErrors = ServerTelemetryManager.maxErrorsToSend(lastRequests);
+        const failedRequests = lastRequests.failedRequests.slice(0, 2*maxErrors).join(SERVER_TELEM_CONSTANTS.VALUE_SEPARATOR);
+        const errors = lastRequests.errors.slice(0, maxErrors).join(SERVER_TELEM_CONSTANTS.VALUE_SEPARATOR);
+        const errorCount = lastRequests.errors.length;
+
+        // Indicate whether this header contains all data or partial data
+        const overflow = maxErrors < errorCount ? "1" : "0";
```
jasonnutter, line 52: Are 1 and 0 true/false here? Adding a constant for both might help with context, as these are magic strings, technically.

lib/msal-common/src/telemetry/server/ServerTelemetryManager.ts
```diff
@@ -73,19 +80,58 @@ export class ServerTelemetryManager {
...
+    static maxErrorsToSend(serverTelemetryEntity: ServerTelemetryEntity): number {
+        let i;
+        let maxErrors = 0;
+        let dataSize = 0;
+        const errorCount = serverTelemetryEntity.errors.length;
+        for (i = 0; i < errorCount; i++) {
+            const apiId = serverTelemetryEntity.failedRequests[2*i];
+            const correlationId = serverTelemetryEntity.failedRequests[2*i + 1];
+            const errorCode = serverTelemetryEntity.errors[i];
+            dataSize += apiId.toString().length + correlationId.toString().length + errorCode.length + 3; // Add 3 to account for commas
```
jasonnutter, line 126: Some more comments might be helpful here, not clear to me what/why is happening.

lib/msal-common/src/cache/entities/ServerTelemetryEntity.ts
```diff
@@ -5,10 +5,18 @@
...
 export class ServerTelemetryEntity {
-    failedRequests: Array<string|number>;
-    errors: string[];
-    errorCount: number;
-    cacheHits: number;
+    public failedRequests: Array<string|number>;
+    public errors: string[];
+    public cacheHits: number;
+
+    static initializeServerTelemetryEntity(): ServerTelemetryEntity {
```
jasonnutter, line 12: Why not a constructor?

lib/msal-common/src/telemetry/server/ServerTelemetryManager.ts
tnorling: Added comments. This function is determining how many of our cached errors can be added to the header sent to the server, based on the number of characters.

lib/msal-common/src/telemetry/server/ServerTelemetryManager.ts
tnorling: Done

lib/msal-common/src/cache/entities/ServerTelemetryEntity.ts
tnorling: Forgot to change this when I updated the design. Thanks for the callout!

jasonnutter (approved):

sameerag (approved):

PR #2245, opened 2020-09-08
Throw error if more than 1 access token is matched

sangonzal:


sameerag (approved):
Lgtm

lib/msal-common/src/client/OnBehalfOfClient.ts
```diff
@@ -88,12 +89,16 @@ export class OnBehalfOfClient extends BaseClient {
...
+        const numAccessTokens = accessTokens.length;
+        if (numAccessTokens < 1) {
             return null;
+        } else if (numAccessTokens > 1) {
+            throw ClientAuthError.createMultipleMatchingTokensInCacheError();
         }
         return accessTokens[0] as AccessTokenEntity;
     }
-    
+
```
pkanher617, line 101: Should this be done in the silent flow as well?

pkanher617 (approved):
Would like to see some tests

lib/msal-common/src/client/OnBehalfOfClient.ts
sangonzal: Yes. This logic is no longer in SilentFlowClient, it was refactored into CacheManager. I updated CacheManager as part of this PR.

sangonzal:
Added tests

Issue #2247, opened 2020-09-09
Base64 encode/decode operation works incorrectly for strings which contain percent-encoded symbols like %20


winperec:
## Library
- [ ] `msal@1.x.x` or `@azure/msal@1.x.x`
- [x] `@azure/msal-browser@2.1.0`
- [ ] `@azure/msal-angular@0.x.x`
- [ ] `@azure/msal-angular@1.x.x`
- [ ] `@azure/msal-angularjs@1.x.x`

## Description
Base64 encode/decode operation works incorrectly for strings which contain URL encoded symbols like %20
## Error Message
As a result if the application URL contains spaces or any encoded symbols you get the error like:
> AADSTS500112: The reply address 'https://appdomain/path with spaces' does not match the reply address 'https://appdomain/path%20with%20spaces' 
## Security
- [ ] Is this issue security related?

## Regression
- [ ] Did this behavior work before? 
Version:

## MSAL Configuration

```js
{
            auth: {
                authority: 'https://login.microsoftonline.com/common,
                clientId: clientAppId,
                navigateToLoginRequestUrl: false,
            },

            cache: {
                cacheLocation: 'localStorage',
                storeAuthStateInCookie: BrowserHelper.isIE(),
            },
        }
// For Azure B2C issues, please include your policies.
```

## Reproduction steps
```js
const crypto = new BrowserCrypto()
const input = 'test%20input'
const decoded = crypto.base64Decode(crypto.base64Encode(input))
// decoded is 'test input' but not 'test%20input'
```

## Expected behavior
The decoded string should be equal to initial input
## Browsers/Environment
- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari
- [x] IE
- [ ] Other (Please add browser name here)

jmckennon:
Thanks for the information, I'll take a look at your PR today and see if I can repro the issue and use the fix from your PR.

winperec:
Maybe even better to reuse base64 encode/decode from msal-core as far as I understood it is more mature? (though both versions look like black magic :))

PR #2248, opened 2020-09-09
#2247 Base64 encode/decode operation works incorrectly for strings with percent-encoded symbols

winperec:


jmckennon (approved):
This looks good, I verified that `decodeURIComponent` was indeed breaking the decode by replacing something like `%20` with a space, as you would expect. @pkanher617 can you verify that `decodeURIComponent` doesn't need to be moved elsewhere, outside of a `Base64Decode`? My suspicion is no as all of the unit tests are still passing, but I wanted to make sure.

pkanher617 (approved):
Looks great. Tested on my end and everything seems to work as expected. Thanks for submitting this!

winperec:
Guys thanks for approving! any ETA on where the new version is available?
@pkanher617 @pkanher617

Issue #2042, opened 2020-07-28
Preventing 3rd Party Cookies block when using MSAL Angular (1.x)


HockeyJustin:
Please follow the issue template below. Failure to do so will result in a delay in answering your question.

## Library
- [x] `@azure/msal-angular@1.x.x`

## Description
I have created an application using `@azure/msal-angular@1.x.x`  (1.0.0-beta.5).

This has been fine for local development, but in iOS (everywhere), private browsing and other machines, there is an issue of 3rd party cookies being blocked. This prevents the whole application from working (it's pretty much all authenticated).

<img width="194" alt="3rd party cookie block" src="https://user-images.githubusercontent.com/4604271/88680779-c7206300-d0e8-11ea-8bf1-5c0b6444c90d.png">

What can I do to get msal-angular to work on all browsers, without users having to turn off browser-wide settings/ defaults please?

Do I need to remove `msal-angular` and go with another library?

If so, is the recommendation to use [msal-browser](https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev/lib/msal-browser) which supports the Auth Code with PKCE, or follow [this tutorial](https://docs.microsoft.com/en-us/samples/azure-samples/active-directory-b2c-javascript-msal-singlepageapp/active-directory-b2c-javascript-msal-singlepageapp/) released earlier this month?

I'm hoping msal-angular has a solution, but I can't see any updates planned on the roadmap.

jo-arroyo:
@HockeyJustin We're aware of the issues with 3rd party cookies with `msal-angular` as it still depends on MSAL.js 1.0. We will upgrade `msal-angular` to work with MSAL.js 2.0. However, that work has not yet been scheduled. We will update our roadmap when we schedule that work. For now, you can use MSAL.js 2.0 directly if you need to use the auth code flow.

HockeyJustin:
@jo-arroyo . Thanks for getting back to me.

This is a big issue for us. Looking at the [MSAL.js 2.0 sample](https://docs.microsoft.com/en-us/samples/azure-samples/active-directory-b2c-javascript-msal-singlepageapp/active-directory-b2c-javascript-msal-singlepageapp/), I would almost need to make my own version of `msal-angular` to get this to work within the angular ecosystem (e.g. Route Guards etc).

I'm not keen on doing all that work, only to then scrap it and use the official msal-angular 2.x.

[...]

PR #2252, opened 2020-09-09
loginRedirect for Angular v2 Sample

jo-arroyo:
This PR adds `loginRedirect` to the angular v2 sample, in response to concerns raised about the sample in issue #2042 . 
It also adds config files for both the guard and interceptor.


samples/msal-angular-v2-samples/angular10-browser-sample/src/app/home/home.component.ts
```diff
@@ -7,9 +8,10 @@ import { Component, OnInit } from '@angular/core';
 })
 export class HomeComponent implements OnInit {
 
-  constructor() { }
+  constructor(private authService: MsalService) { }
 
   ngOnInit(): void {
+    this.authService.handleRedirectObservable();
```
jasonnutter, line 14: I think it would be good to call subscribe and console.log the result/error at least.

jasonnutter (approved):
One thing otherwise looks good!

samples/msal-angular-v2-samples/angular10-browser-sample/src/app/app.module.ts
```diff
@@ -24,6 +27,16 @@ function MSALInstanceFactory(): IPublicClientApplication {
   });
 }
 
+function MSALInterceptorConfigFactory(): MsalInterceptorConfig {
+  const protectedResourceMap = new Map<string, Array<string>>()
+  protectedResourceMap.set('https://graph.microsoft.com/v1.0/me', ['user.read'])
```
pkanher617, line 32: This should probably be done in a config file right?

jasonnutter: `app.module.ts` is essentially a config file for the app. We could put it in a separate file and have this module pull it in, but easier to just put it here.

samples/msal-angular-v2-samples/angular10-browser-sample/src/app/home/home.component.ts
jo-arroyo: Done

jmckennon (approved):
LGTM!

superman-lopez:
I am running into an issue with this sample, I am not sure if it's related to this merge or something else:
When using `interactionType: InteractionType.REDIRECT` the authentication is successful, but something off after the redirect as calling `this.authService.getAllAccounts()` returns no accounts.  This is also obvious in the Angular web app as the Login is still showing and the Logout button is hidden.  Calling the Microsoft Graph in the `/profile` location will trigger a failing acquireToken call.

When I refresh Chrome, calling `this.authService.getAllAccounts()` returns the correct value.  In Safari the browser refresh kicks off a acquireToken loop.

When I use `interactionType: InteractionType.POPUP` in interceptor config the redirect immediately shows the correctly refreshed Angular web app and the Logout button is showing rather than the Login button.

PR #2255, opened 2020-09-10
Remove debug package dependency from msal-common

HarryGifford:
It looks like the package `debug` is not used in `msal-common`. Removing it allows us to avoid tracking `ms` and `debug` as external dependencies.

### DIFF
--- a/lib/msal-common/package.json
+++ b/lib/msal-common/package.json
@@ -10,7 +10,7 @@
     "type": "git",
     "url": "https://github.com/AzureAD/microsoft-authentication-library-for-js.git"
   },
-  "version": "1.2.0",
+  "version": "1.2.1",
   "description": "Microsoft Authentication Library for js",
   "keywords": [
     "implicit",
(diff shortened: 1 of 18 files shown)


lib/msal-common/package.json
```diff
@@ -10,7 +10,7 @@
     "type": "git",
     "url": "https://github.com/AzureAD/microsoft-authentication-library-for-js.git"
   },
-  "version": "1.2.0",
+  "version": "1.2.1",
```
jasonnutter, line 13: @pkanher617 Should this be bumped manually or does beachball take care of that?

pkanher617: I'm not sure why the version is being bumped in this PR, this should only be done before a release.

HarryGifford: @pkanher617 Ok, thanks. I see the build fails for me anyways, so I will open an issue and abandon this PR instead.

PR #2264, opened 2020-09-10
acquireTokenSilent calls ssoSilent

tnorling:
- Move `acquireTokenByRefreshToken` to `RefreshTokenClient`
- `acquireTokenSilent` uses:
  1. `acquireCachedToken` on `SilentFlowClient`
  2. `acquireTokenByRefreshToken` on `RefreshTokenClient`
  3. `ssoSilent` if refresh token is expired
- `account` is added as optional parameter on `AuthorizationUrlRequest` to accommodate the lack of `loginHint` and `sid` on the `SilentFlowRequest` object.
- Auth Code Url parameter builder prefers `sid` over `loginHint` and will generate `login_hint` from `account.username` if neither are provided.

### DIFF
--- a/lib/msal-common/src/client/AuthorizationCodeClient.ts
+++ b/lib/msal-common/src/client/AuthorizationCodeClient.ts
@@ -231,16 +231,17 @@ export class AuthorizationCodeClient extends BaseClient {
             parameterBuilder.addPrompt(request.prompt);
         }
 
-        if (request.loginHint) {
-            parameterBuilder.addLoginHint(request.loginHint);
-        }
-
         if (request.domainHint) {
             parameterBuilder.addDomainHint(request.domainHint);
         }
 
+        // Add sid or loginHint with preference for sid -> loginHint -> username of AccountInfo object
         if (request.sid) {
             parameterBuilder.addSid(request.sid);
+        } else if (request.loginHint) {
+            parameterBuilder.addLoginHint(request.loginHint);
+        } else if (request.account && request.account.username) {
+            parameterBuilder.addLoginHint(request.account.username);
         }
 
         if (request.nonce) {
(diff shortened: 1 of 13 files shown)


lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -231,16 +231,17 @@ export class AuthorizationCodeClient extends BaseClient {
...
         if (request.domainHint) {
             parameterBuilder.addDomainHint(request.domainHint);
         }
 
+        // Add sid or loginHint with preference for sid -> loginHint -> username of AccountInfo object
         if (request.sid) {
             parameterBuilder.addSid(request.sid);
+        } else if (request.loginHint) {
+            parameterBuilder.addLoginHint(request.loginHint);
+        } else if (request.account && request.account.username) {
```
sameerag, line 243: `Account` can also have `sid`? Used to in msal-core, not sure how relevant it is for browser!

jasonnutter: We should use `sid` if it is present in the id token.

sameerag (approved):
One comment, approved.

lib/msal-common/src/client/AuthorizationCodeClient.ts
sameerag: Then, we need to account for `sid` in `account`

tnorling: That's a good suggestion but will take it in a follow up PR

pkanher617: +1 for a follow up PR

pkanher617 (approved):
LGTM

PR #2265, opened 2020-09-10
[msal-node-extensions] Update exceptions and logging. Updates locking behavior. 

sangonzal:
- Updates exception messages
- Improves logging
- Updates lock to also catch EPERM error when trying to acquire lock


change/@azure-msal-node-extensions-2020-09-10-15-19-52-msal-extensions-update-error.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "prerelease",
+  "comment": "Update error message",
+  "packageName": "@azure/msal-node-extensions",
+  "email": "email not defined",
```
jo-arroyo, line 5: Nit: wondering if you wanted to add your email here? Or blank for a reason

extensions/msal-node-extensions/src/utils/Constants.ts
```diff
@@ -17,6 +17,12 @@ export const Constants = {
      */
     ENOENT_ERROR: "ENOENT",
 
+    /**
+     * Operation not permitted. AN attempt was made to perform an operation taht requires
```
jo-arroyo, line 21: Nit: spelling

change/@azure-msal-node-extensions-2020-09-10-15-19-52-msal-extensions-update-error.json
sangonzal: Yep good catch, don't know why beachball didn't add it.

extensions/msal-node-extensions/src/utils/Constants.ts
sangonzal: Updated

extensions/msal-node-extensions/src/lock/CrossPlatformLock.ts
```diff
@@ -62,14 +63,20 @@ export class CrossPlatformLock {
...
+                await fs.unlink(this.lockFilePath);
+                await this.lockFileHandle.close();
+                this.logger.info("lockfile deleted");
+            } else {
+                this.logger.warning("lockfile handle does not exist, so lockfile could not be deleted");
+            }
         } catch (err) {
             if (err.code == Constants.ENOENT_ERROR) {
-                this.logger.warning("Tried to unlock but Lockfile does not exist");
+                this.logger.warning("Tried to unlock but lockfile does not exist");
```
sameerag, line 76: Difference is that .lockfile missing vs unable to unlock?

sameerag (approved):
Approved. One question: Can we also add time stamps to performance testing, which may aid in benchmarking as you have the framework now?

jo-arroyo (approved):

extensions/msal-node-extensions/src/lock/CrossPlatformLock.ts
sangonzal: Yes. We try to delete the .lockfile. If it fails because it's not there, then it means that the file is unlocked, but we log a warning message. If it fails for whatever other reason, file still might be locked and therefore we throw.

PR #2267, opened 2020-09-10
Add default scopes to silent requests

tnorling:
- Adds `openid` and `profile` to all requests during parameter builder step. eSTS expects that these scopes are present on all requests. 
- Removes these scopes from the cache lookup logic to prevent cache misses if the service does not return these scopes in the response.
- Removes these scopes from intersecting scope logic to prevent possible unintended removal of tokens that contain these scopes.


pkanher617 (approved):
Looks good. Make sure we add/update tests before we merge.

jo-arroyo (approved):
LGTM!

PR #2290, opened 2020-09-15
Properly support multiple concurrent auth code requests

jasonnutter:
Refactors temporary cache logic for state to support multiple concurrent iframe requests. Removes logic to attempt to reuse an iframe based on scope, in the interest of simplicity in the library.

TODO:
- [x] Helper functions
- [x] Test with all flows
- [x] Unit tests


lib/msal-browser/src/cache/BrowserStorage.ts
```diff
@@ -331,16 +334,43 @@ export class BrowserStorage extends CacheManager {
...
+
+    /**
+     * Creates full cache key for the request state
+     * @param stateString State string for the request
+     */
+    generateStateKey(stateString: string): string {
+        // Use the library state id to key temp storage for uniqueness for multiple concurrent requests
+        const {
+            libraryState: {
+                id: stateId
```
jasonnutter, line 369: Using stateId since 1) its different from teh state string itself 2) it wont have encoding issues 3) its shorter (important for cookies)

pkanher617 (approved):

lib/msal-browser/src/interaction_handler/RedirectHandler.ts
```diff
@@ -55,13 +55,16 @@ export class RedirectHandler extends InteractionHandler {
         // Interaction is completed - remove interaction status.
         this.browserStorage.removeItem(this.browserStorage.generateCacheKey(BrowserConstants.INTERACTION_STATUS_KEY));
 
+        const hashUrlString = new UrlString(locationHash);
+        // Deserialize hash fragment response parameters.
+        const serverParams: ServerAuthorizationCodeResponse = UrlString.getDeserializedHash(hashUrlString.getHash());
```
tnorling, line 60: We should consider making a helper function to get the state from the hash. We're doing this in a couple places and I've seen a couple customer requests to provide a method to extract the state (scenarios such as subdomain routing would require them to handle their own redirects and they use state to do that)

jasonnutter: Yep, working on this now.

jasonnutter: Added.

pkanher617: We already return the state in the response. We can make a helper for this but there shouldn't be any reason that they can't use the state to do the subdomain today.

tnorling: In this scenario they aren't going as far as getting the response because they need to do another redirect themselves before processing the hash. This doesn't need to be exposed as a public API in this PR but wanted to point it out as another use case for a helper. We can talk offline if you want to clarify the use case I'm referring to here

tnorling (approved):

PR #2296, opened 2020-09-16
Update/cleanup to readme for MSAL.js and MSAL Node.

mvrak:


mvrak:
Combining with sample submission.

PR #2302, opened 2020-09-16
Scopes stored case sensitive and matched case insensitive

sameerag:
Scopes in common cache scenarios have to adhere to the below rules:

* Stored with case sensitivity (we store what we get from the server)
* Matched with case insensitivity (to account for shared cache)

This PR accounts for the above rules for scopes.


tnorling:
Looks good. Can we add some unit tests specifically testing the case sensitivity/insensitivity? Looks like one of the E2E tests needs to be updated as well.

tnorling (approved):

PR #2335, opened 2020-09-23
Network Interface is exported for custom http calls in AzureSDK

sameerag:
AzureSDK has a requirement to customize their network calls and `msal-common` already has an interface which facilitates it. This PR exports the needed interfaces in `msal-node` to limit the package dependency.

### DIFF
--- a/change/@azure-msal-common-2020-09-23-11-49-08-public-network-interface.json
+++ b/change/@azure-msal-common-2020-09-23-11-49-08-public-network-interface.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Remove null in function return types to be compatible with ICacheManager.ts",
+  "comment": "Remove null in function return types to be compatible with ICacheManager.ts (#2335)",
   "packageName": "@azure/msal-common",
   "email": "sameera.gajjarapu@microsoft.com",
   "dependentChangeType": "patch",
--- a/change/@azure-msal-common-2020-09-23-11-49-08-public-network-interface.json
+++ b/change/@azure-msal-common-2020-09-23-11-49-08-public-network-interface.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Remove null in function return types to be compatible with ICacheManager.ts",
+  "packageName": "@azure/msal-common",
+  "email": "sameera.gajjarapu@microsoft.com",
+  "dependentChangeType": "patch",
+  "date": "2020-09-23T18:48:44.045Z"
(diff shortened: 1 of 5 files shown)


lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -177,7 +177,7 @@ export abstract class CacheManager implements ICacheManager {
      * retrieve a credential - accessToken, idToken or refreshToken; given the cache key
      * @param key
      */
-    getCredential(key: string): CredentialEntity | null {
+    getCredential(key: string): CredentialEntity {
```
tnorling, line 180: Why were these changes made? The interface should be updated to include the null return instead

sameerag: An object can be null, why are we specifically making null as an acceptable type? Either way is fine for me, the changes were made because AzureSDK is seeing these issues in their repo:
    
_../../../common/temp/node_modules/.pnpm/@azure/msal-common@1.3.0/node_modules/@azure/msal-common/dist/src/cache/CacheManager.d.ts(78,5): error TS2416: Property 'getCredential' in type 'CacheManager' is not assignable to the same property in base type 'ICacheManager'. 
 Type '(key: string) => CredentialEntity | null' is not assignable to type '(key: string) => CredentialEntity'. 
 Type 'CredentialEntity | null' is not assignable to type 'CredentialEntity'. 
 Type 'null' is not assignable to type 'CredentialEntity'._

change/@azure-msal-common-2020-09-23-11-49-08-public-network-interface.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Remove null in function return types to be compatible with ICacheManager.ts",
```
tnorling, line 3: Can we update this as well and add the PR number at the end?

tnorling (approved):

lib/msal-common/src/cache/CacheManager.ts
sameerag: Done. Made them uniform and kept the `null`. Lets revisit this when we enable `strict null checks` in the library.

change/@azure-msal-common-2020-09-23-11-49-08-public-network-interface.json
sameerag: Done

technical-boy (approved):
LGTM!

PR #2343, opened 2020-09-25
Add enhanced documentation about responseMode in browser

jmckennon:


jasonnutter (approved):

lib/msal-browser/src/request/PopupRequest.ts
```diff
@@ -14,7 +14,7 @@ import { AuthorizationUrlRequest } from "@azure/msal-common";
  * - correlationId              - Unique GUID set per request to trace a request end-to-end for telemetry purposes.
  * - redirectUri                - The redirect URI where authentication responses can be received by your application. It must exactly match one of the redirect URIs registered in the Azure portal.
  * - extraScopesToConsent       - Scopes for a different resource when the user needs consent upfront.
- * - responseMode               - Specifies the method that should be used to send the authentication result to your app. Can be query, form_post, or fragment. If no value is passed in, it defaults to query.
+ * - responseMode               - Specifies the method that should be used to send the authentication result to your app. Fragment is the only valid option for msal-browser.
```
tnorling, line 17: If this is not configurable we should consider removing it from the request object

tnorling (approved):
One general comment which we can discuss outside of this PR

lib/msal-browser/src/request/PopupRequest.ts
jmckennon: It is configurable, it just won't change anything in the browser library. There are other options that are relevant for node.

tnorling: Yeah I guess what I'm suggesting is that we don't expose it in browser. Imo it's not a good developer experience to expose params people can't configure or to make params optional that are required, which we do as well. I think we need to rethink what should be kept internal and what should be exposed to the user. But that can be a separate discussion and shouldn't be a blocker for this PR for now, just wanted to raise it.

sameerag: I agree. Browser configuration is separate and public only for SPAs, if you don't want to support other mode, may be we should default internally?

sameerag (approved):
Approved with comments. @tnorling What we proposed though will be a breaking change and `msal-browser` has GA'ed, so may not be feasible.

lib/msal-browser/src/request/PopupRequest.ts
jmckennon: Sounds good--will keep note of this, but resolving and merging for now.

PR #2344, opened 2020-09-25
[msal-node] Add Auth Code with PKCE sample for msal-node

hectormmg:
This PR adds an extended Authorization Code flow sample that shows how PKCE should be used with MSAL Node's public APIs.

For clarity, the sample does not implement PKCE code generation logic, that is left to the client developer.

The scope of this sample is limited to showing how one would add their PKCE codes to an authorization request made through MSAL Node.


sangonzal (approved):

samples/msal-node-samples/standalone-samples/auth-code-pkce/index.js
```diff
@@ -0,0 +1,86 @@
...
+ * For this sample, you can either implement your own PKCE code generation logic or use an existing tool
+ * to manually generate a Code Verifier and Code Challenge, plugging them into the pkceCodes object below.
+ * 
+ * For details on implementing your own PKCE code generation logic, consult the 
+ * PKCE specification https://tools.ietf.org/html/rfc7636#section-4
+ */
+
+const PKCE_CODES = {
+    CHALLENGE_METHOD: "S256", // Use SHA256 Algorithm
+    VERIFIER: "", // Generate a code verifier for the Auth Code Request first
```
pkanher617, line 45: I think it would be good to include a sample showing how to generate these values. You can use the node crypto libraries to do this.

pkanher617 (changes requested):
See comment about PKCE params

samples/msal-node-samples/standalone-samples/auth-code-pkce/index.js
sameerag: Node `crypto` APIs need some more work to be made public @pkanher617. We found some bugs and concluded the activity may need some more work.

We want the users to default to PKCE for public clients and hence this sample is being merged now, but we are tracking the activity to provide helper APIs with node crypto libraries later this quarter.

sameerag (approved):

pkanher617 (approved):

samples/msal-node-samples/standalone-samples/auth-code-pkce/index.js
pkanher617: Yeah that's not what I meant by this. You can easily generate PKCE parameters using the node crypto library in the sample. We can do it in a follow up but this sample is incomplete.

samples/msal-node-samples/standalone-samples/auth-code-pkce/index.js
```diff
@@ -0,0 +1,86 @@
...
+const msal = require('@azure/msal-node');
+
+const SERVER_PORT = process.env.PORT || 3000;
+
+// Before running the sample, you will need to replace the values in the config, 
+// including the clientSecret
+const config = {
+    auth: {
+        clientId: "89e61572-2f96-47ba-b571-9d8c8f96b69d",
+        authority: "https://login.microsoftonline.com/5d97b14d-c396-4aee-b524-c86d33e9b660"
```
sameerag, line 15: Shouldn't we add `clientSecret` here?

PR #2348, opened 2020-09-28
Update cache interface and cache plugin API

sangonzal:
- Updates `ICachePlugin` from `readFromPersistence` and `writeToPersistence` to `beforeCacheAccess` and `afterCacheAccess`
- Removes `readFromPersistence` and `writeToPersistence` from public API. Instead, `beforeCacheAccess` and `afterCacheAccess` are now triggered inside MSAL.
- Updates cache behavior when saving tokens to the cache. If saving tokens from an acquireTokenByRefreshToken operation, if the account which was not used to refresh the tokens is not present in the cache, then MSAL returns null and does not store the refresh tokens in the cache. 
- Updates msal-node-extensions `PersistenceCachePlugin` to reflect new `ICachePlugin`
- Updates msal-node `getAllAccounts` and `removeAccounts` to be async.


lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -200,6 +200,8 @@ export abstract class ClientApplication {
                 cpu: process.arch || "",
                 os: process.platform || "",
             },
+            persistencePlugin: this.config.cache!.cachePlugin,
```
hectormmg, line 203: Just out of curiosity because I'm new to the non-null assertion operator, what does it do here? Is it just a way to tell the compiler that there's no way `this.config.cache` could be null or undefined at this point?

hectormmg (approved):
Looks great!

lib/msal-node/src/client/ClientApplication.ts
sangonzal: Yep, that's right.

sameerag (approved):
Approved with comments. Can we please update the samples as needed and documentation around this usage if any?

change/@azure-msal-node-2020-09-29-10-21-20-msal-node-update-cache-interface.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "prerelease",
+  "comment": "Update TokenCache interface (#2348)",
+  "packageName": "@azure/msal-node",
+  "email": "email not defined",
+  "dependentChangeType": "patch",
```
pkanher617, line 6: change this to "none"

change/@azure-msal-node-extensions-2020-09-29-10-21-20-msal-node-update-cache-interface.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "prerelease",
+  "comment": "Update PersistenceCachePlugin (#2348)",
+  "packageName": "@azure/msal-node-extensions",
+  "email": "email not defined",
+  "dependentChangeType": "patch",
```
pkanher617, line 6: change to "none"

pkanher617 (approved):
make sure to update the changefiles, otherwise lgtm

PR #2355, opened 2020-09-29
[msal-core] Fix acquireTokenSilent closing pop up window from concurrent acquireTokenPopupCall

### LINKED ISSUE #2116: acquireTokenPopup fails if used concurrently with acquireTokenSilent

## Library
- [x] `msal@1.x.x` or `@azure/msal@1.x.x`

msal@1.3.2
https://alcdn.msauth.net/lib/1.3.2/js/msal.min.js

## Framework
None, basic JavaScript application

## Description
When MSAL tries to acquire a token via acquireTokenSilent and acquireTokenPopup concurrently, if acquireTokenSilent returns a response it automatically closes the popup before user has a chance to interact with it. This results in acquireTokenPopup failing. In our usage the acquireTokenPopup is used when user has not yet consented to scopes, but acquireTokenSilent is trying to acquire a token for different scopes that user has already consented to.

## Error Message
ClientAuthError: User cancelled the flow.

## Security

## Regression
It worked with ADAL
Version:

## MSAL Configuration

```js
// Repros with a basic configuration
var application = new Msal.UserAgentApplication({
    auth: {
        clientId: "<APP_ID>",
        redirectUri: "https://localhost"
    }
});
```

## Reproduction steps
```js
// Get token that requires consent. Fails with ClientAuthError: User cancelled the flow.
application.acquireTokenPopup({
    scopes: ["https://onenote.com/Notes.Read"]
}).then((result) => {
    console.log(result)
}).catch((ex) => {
    console.error(ex);
});

// Gets the token silently successfully.
application.acquireTokenSilent({
    scopes: ["https://graph.microsoft.com/user.read"]
}).then((result) => {
    console.log(result)
}).catch((ex) => {
    console.error(ex);
});
```
## Code causing issue
```
   private handleAuthenticationResponse(hash: string): void {
        this.logger.verbose("HandleAuthenticationResponse has been called");

        // retrieve the hash
        const locationHash = hash || window.location.hash;

        const stateInfo = this.getResponseState(locationHash);
        this.logger.verbose("Obtained state from response");

        const tokenResponseCallback = window.callbackMappedToRenewStates[stateInfo.state];
        this.processCallBack(locationHash, stateInfo, tokenResponseCallback);

        // If current window is opener, close all windows
        WindowUtils.closePopups();
    }
```
 **WindowUtils.closePopups()** is closing all popups opened by MSAL when any response is received from a different request.

## Expected behavior
I would expect that I can acquire both tokens as they are for different resources, I need two separate tokens. The getTokenSilent call succeeding should not impact the result of my interactive flow I need for another feature. We expect MSAL to allow concurrent token requests.

## Browsers/Environment
- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari
- [ ] IE

hectormmg:
This PR:
- Removes call to `WindowUtils.closePopups()` inside `handleAuthenticationResponse()`
- Moves call to `WindowUtils.closePopups()` in `acquireTokenHelper()` outside of if block that caused it to only be called when used with Angular framework
- Fixes #2116


jasonnutter (approved):

pkanher617 (approved):
approved but would like to see some tests for this in the near future.

jasonnutter (approved):

tnorling (approved):

PR #2356, opened 2020-09-29
msal-react #1: account changes

tnorling:
- Adds a stubbed instance of PCA to msal-browser which will throw if any APIs are invoked. This is used as the default instance for the react context and can be used in other framework libraries
- Updates `useIsAuthenticated` hook to accept an `AccountIdentifiers` object and call `getAccountByHomeId` if `homeAccountId` is provided
- Removes the `[]` fallback when calling `getAllAccounts` in `msal-react` as `msal-browser` was previously updated to return an empty array instead of null


lib/msal-react/src/useIsAuthenticated.ts
```diff
@@ -1,21 +1,36 @@
-import { useState, useEffect } from 'react';
+import { AccountInfo, IPublicClientApplication } from "@azure/msal-browser";
+import { useState, useEffect } from "react";
 
-import { useMsal } from './MsalProvider';
-import { isAuthenticated } from './utilities';
+import { useMsal } from "./MsalProvider";
 
-export function useIsAuthenticated(username?: string): boolean {
+function isAuthenticated(instance: IPublicClientApplication, account?: Partial<AccountInfo>): boolean {
```
tnorling, line 6: This probably belongs in msal-browser. Open questions: 
- Should we create a new type containing optional parameters: username, homeAccountId, and possibly accountId in the future. Or what's shown here an optional AccountInfo object?
- Should the lookup logic try to match every parameter provided, or give precedence to certain params as shown here?

lib/msal-react/src/MsalContext.tsx
```diff
@@ -1,5 +1,6 @@
 import * as React from "react";
 import { IPublicClientApplication, AccountInfo } from "@azure/msal-browser";
+import { PublicClientApplication } from "../../msal-browser/dist/src";
```
jasonnutter, line 3: This should come from the `@azure/msal-browser` import.

lib/msal-react/src/MsalContext.tsx
```diff
@@ -10,47 +11,12 @@ export interface IMsalContext {
...
-            return Promise.reject();
-        },
-        logout: () => {
-            return Promise.reject();
-        },
-        ssoSilent: () => {
-            return Promise.reject();
-        },
-    },
+    instance: new PublicClientApplication({}),
```
jasonnutter, line 19: What happens if one of the APIs are invoked on this default instance?

tnorling: What is the scenario where that would happen? Imo the default should never be used. The reason I removed the stubbed functions was that I don't want that to be another thing we need to maintain when we add/remove APIs from msal-browser.

lib/msal-react/src/MsalContext.tsx
tnorling: Thanks for catching this!

lib/msal-react/src/MsalContext.tsx
jasonnutter: I think if you use the MsalContext but don't provide an instance. One thing I would be concerned about are any side effect that are caused by instantiating `PublicClientApplication`. 

Also, I don't see maintaining the stub implementation as a big deal, given we would already need to update the wrapper instance if new APIs are added.

tnorling: I don't think this is ever instantiated if you have a Provider above and I think using the Context without a Provider should be unsupported. I was hoping to also get rid of the wrapper instance once the event API is added...

tnorling: Agreed to have the stubbed instance live in `msal-browser`, allowing it to be used by other framework libraries and forcing us to update the stubbed instance when we update the PCA interface.

pkanher617 (approved):

lib/msal-react/package.json
```diff
@@ -1,33 +1,46 @@
...
-    "@azure/msal-browser": "^2.0.0",
+    "@azure/msal-browser": "^2.2.0",
     "react": ">=16"
   },
-  "name": "@azure/msal-react",
-  "author": "Jason Nutter",
   "module": "dist/msal-react.esm.js",
   "devDependencies": {
-    "@azure/msal-browser": "^2.0.0",
+    "@azure/msal-browser": "^2.2.0",
```
pkanher617, line 43: should be 2.3

jmckennon (approved):

PR #2369, opened 2020-10-01
msal-react #2: Login handler

tnorling:
- Updates `useMsalAuthentication` hook to accept an `interactionType` and `request` object to trigger login if user is not already authenticated
- `useHandleRedirect` hook returns a tuple of `[response, error]`
- Replaces wrapped msal instance with event callbacks


lib/msal-react/src/Constants.ts
```diff
@@ -0,0 +1,8 @@
+/*
+ * Interaction type of the API
+ */
+export enum InteractionType {
```
jasonnutter, line 4: This should be added to browser (I think there is an existing enum that can be exported).

lib/msal-react/src/MsalAuthentication.tsx
```diff
@@ -1,26 +1,25 @@
...
     msal: IMsalContext;
 };
 
-// TODO: Add optional argument for the `request` object?
-export function useMsalAuthentication(
-    args: IMsalAuthenticationProps = {}
-): MsalAuthenticationResult {
-    const { username, loginHandler = defaultLoginHandler } = args;
+export function useMsalAuthentication(args: IMsalAuthenticationProps = {}): MsalAuthenticationResult {
+    const { username, interactionType = InteractionType.POPUP, authenticationRequest } = args;
```
jasonnutter, line 22: For both Angular and React, we may want to consider making interactionType required.

tnorling: I think that's reasonable. It would make it more clear to the user what's happening instead of silently defaulting to popups

jamiehaywood: hi guys, a question about the use of `enum` instead of a string literal type for `InteractionType`?

```
interactionType = 'popup'
```
seems a lot cleaner to me, and also benefits from enhanced vscode intellisense as well as not having to import the `InteractionType` due to the string literal type being inferred by TypeScript.

With your blessing, happy to raise a PR for any of these changes.

lib/msal-react/src/MsalAuthentication.tsx
```diff
@@ -33,24 +31,32 @@ export function useMsalAuthentication(
...
+            return msal.instance.loginRedirect(authenticationRequest as RedirectRequest).catch(error => {
+                setError(error);
+            });
+        } else if (interactionType === InteractionType.SILENT) {
+            return msal.instance.ssoSilent(authenticationRequest as SsoSilentRequest).catch(error => {
+                setError(error);
+            }); 
+        } else {
+            return null;
+        }
```
jamiehaywood, line 48:
```suggestion
        switch (interactionType) {
            case InteractionType.POPUP:
                return msal.instance.loginPopup(authenticationRequest as PopupRequest).catch(error => {
                    setError(error);
                });
            case InteractionType.REDIRECT:
                return msal.instance.loginRedirect(authenticationRequest as RedirectRequest).catch(error => {
                    setError(error);
                });
            case InteractionType.SILENT:
                return msal.instance.ssoSilent(authenticationRequest as SsoSilentRequest).catch(error => {
                    setError(error);
                }); 
            default:
                return null;
        }
```

lib/msal-react/src/MsalAuthentication.tsx
```diff
@@ -63,8 +69,8 @@ export function useMsalAuthentication(
 }
 
 export const MsalAuthentication: React.FunctionComponent<IMsalAuthenticationProps> = props => {
-    const { username, loginHandler, children } = props;
-    const { msal } = useMsalAuthentication({ username, loginHandler });
+    const { username, interactionType, authenticationRequest, children } = props;
+    const { msal } = useMsalAuthentication({ username, interactionType, authenticationRequest });
```
jamiehaywood, line 73:
```suggestion
export const MsalAuthentication: React.FunctionComponent<IMsalAuthenticationProps> = ({ username, interactionType, authenticationRequest, children }) => {
    const { msal } = useMsalAuthentication({ username, interactionType, authenticationRequest });
```

jamiehaywood: destructuring in the params I feel looks cleaner

lib/msal-react/src/MsalAuthentication.tsx
jasonnutter: I would argue that enums provide better intellisense, given they provide a clear definition of the supported values. We also want to avoid magic strings whenever possible, especially given this enum is used internally by the library. And this enum does map to string values: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/src/utils/BrowserConstants.ts#L74-L78

lib/msal-browser/src/index.ts
```diff
@@ -1,6 +1,7 @@
 // App and config
 export { PublicClientApplication } from "./app/PublicClientApplication";
 export { Configuration } from "./config/Configuration";
+export { InteractionType } from "./utils/BrowserConstants";
```
jasonnutter, line 4: We may want to change the casing of the values to capital-case  ("Popup" instead of "POPUP") now that this being exported.

lib/msal-react/src/MsalAuthentication.tsx
tnorling: thanks for the suggestion! updated!

lib/msal-browser/src/index.ts
tnorling: Do you mean for the enum keys? The values are all lowercase "popup"

lib/msal-react/src/MsalAuthentication.tsx
jamiehaywood: Thanks for the explanation. Completely understand about the decision particularly around the `magic string` point 😀

lib/msal-browser/src/index.ts
tnorling: Done

lib/msal-react/src/MsalAuthentication.tsx
```diff
@@ -17,7 +18,7 @@ type MsalAuthenticationResult = {
 };
 
 export function useMsalAuthentication(msalProps: IMsalAuthenticationProps): MsalAuthenticationResult {
-    const { interactionType, username, authenticationRequest } = msalProps;
+    const { interactionType, username, homeAccountId, authenticationRequest } = msalProps;
```
jamiehaywood:
```suggestion
export function useMsalAuthentication({ interactionType, username, homeAccountId, authenticationRequest }: IMsalAuthenticationProps): MsalAuthenticationResult {
```

lib/msal-react/src/useMsalAuthentication.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { useMsal } from "./MsalProvider";
+import { IMsalContext } from "./MsalContext";
+
+type MsalAuthenticationResult = {
+    msal: IMsalContext;
+};
+
+export function useMsalAuthentication(
+    interactionType: InteractionType, 
+    authenticationRequest?: PopupRequest|RedirectRequest|SsoSilentRequest,
```
jmckennon, line 13: Are both `interactionType` and `authenticationRequest` necessary? Can we not infer one from the other?

lib/msal-react/src/useMsalAuthentication.tsx
```diff
@@ -0,0 +1,44 @@
...
+    authenticationRequest?: PopupRequest|RedirectRequest|SsoSilentRequest, 
+    accountIdentifier?: AccountIdentifiers
+): MsalAuthenticationResult {
+    const msal = useMsal();
+    const isAuthenticated = useIsAuthenticated(accountIdentifier);
+
+    const login = useCallback(() => {
+        switch (interactionType) {
+            case InteractionType.Popup:
+                return msal.instance.loginPopup(authenticationRequest as PopupRequest);
```
jmckennon, line 22: Since `authenticationRequest` is an optional, does this work by supplying a default `PopupRequest` if none is given?

jmckennon (approved):
Few comments but looks good!

lib/msal-react/src/useMsalAuthentication.tsx
tnorling: The request object is also optional the login APIs, if not provided msal-browser will fill in things like `authority` and `redirectUri` from the values provided on the PublicClientApplication config and default the `scopes` to `openid` and `profile`

lib/msal-react/src/useMsalAuthentication.tsx
tnorling: Since types are stripped at runtime and these also share a lot of properties we cannot infer the `interactionType` from just the type of the `authenticationRequest`

lib/msal-react/src/useMsalAuthentication.ts
```diff
@@ -0,0 +1,51 @@
...
+    const [result, setResult] = useState<AuthenticationResult|null>(null);
+    const [error, setError] = useState<AuthError|null>(null);
+
+    const login = useCallback((loginType: InteractionType, request?: PopupRequest|RedirectRequest|SsoSilentRequest): Promise<AuthenticationResult|null> => {
+        switch (loginType) {
+            case InteractionType.Popup:
+                return instance.loginPopup(request as PopupRequest);
+            case InteractionType.Redirect:
+                // This promise is not expected to resolve due to full frame redirect
+                return instance.loginRedirect(request as RedirectRequest).then(null);
```
jasonnutter, line 33: Are you looking to return `null` is this throws? Think it needs to be:
```suggestion
                return instance.loginRedirect(request as RedirectRequest).then(() => null);
```

jasonnutter (approved):
One thing otherwise looks good!

lib/msal-react/src/useMsalAuthentication.ts
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+import { useCallback, useEffect, useState } from "react";
+import { PopupRequest, RedirectRequest, SsoSilentRequest, InteractionType, AuthenticationResult, AuthError } from "@azure/msal-browser";
+import { AccountIdentifiers, useIsAuthenticated } from "./useIsAuthenticated";
+import { useMsal } from "./MsalProvider";
+
+type MsalAuthenticationResult = {
+    login: Function; 
+    result: AuthenticationResult|null|void;
```
tnorling, line 13:
```suggestion
    result: AuthenticationResult|null;
```

lib/msal-react/src/useMsalAuthentication.ts
tnorling: Thanks for catching this. This is to prevent returning `void`

PR #2393, opened 2020-10-08
Check for Headers class when configuring network client

### LINKED ISSUE #2391: FetchClient uses Headers without checking for existence

## Library
- [x] `@azure/msal-browser@2.2.1`

## Description
The FetchClient code uses Headers without first checking to see if they are defined/exist. For browsers such as IE11 where fetch may be polyfilled but Headers is not (such as unfetch is widely used, 2.5M weekly downloads) this will prevent MSAL from being able to authenticate. We are proposing adding a check for Headers existence either before FetchClient is used or within getFetchHeaders.
## Browsers/Environment
- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari
- [x] IE

jasonnutter:
BrowserUtils should check for all globals it uses before using the fetch client. This is to unblock scenarios where `fetch` is defined but `Headers` is not (e.g. by the `unfetch` package).


tnorling (approved):

change/@azure-msal-browser-2020-10-07-17-32-50-msal-browser-fetch-headers.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Check for Headers class when configuring network client",
+  "packageName": "@azure/msal-browser",
+  "email": "janutter@microsoft.com",
+  "dependentChangeType": "patch",
```
pkanher617, line 6: Could you change this to "none"? I am looking at a way to make an auto config.

pkanher617 (approved):
One small change otherwise looks good.

change/@azure-msal-browser-2020-10-07-17-32-50-msal-browser-fetch-headers.json
jasonnutter: Why? This is a bugfix.

PR #2394, opened 2020-10-08
Add event API to msal-browser

jo-arroyo:
Adds event API to msal-browser.

Note: merge `angular-sample-add-events` PR into this before merging to dev.


change/@azure-msal-browser-2020-10-07-18-01-18-browser-add-broadcast.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "minor",
+  "comment": "Add event api to msal-browser (#2394)",
+  "packageName": "@azure/msal-browser",
+  "email": "joarroyo@microsoft.com",
+  "dependentChangeType": "patch",
```
jasonnutter, line 6: I would consider this a minor version, as it adds a new feature set to the library.

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -108,7 +116,20 @@ export abstract class ClientApplication {
      * @returns {Promise.<AuthenticationResult | null>} token response or null. If the return value is null, then no auth redirect was detected.
      */
     async handleRedirectPromise(): Promise<AuthenticationResult | null> {
-        return this.isBrowserEnvironment ? this.handleRedirectResponse() : null;
+        if (this.isBrowserEnvironment) {
+            return this.handleRedirectResponse()
+                .then((result: AuthenticationResult) => {
+                    if (result) {
+                        this.broadcastEvent(BroadcastEvent.HANDLE_REDIRECT_SUCCESS, InteractionType.REDIRECT, result);
```
tnorling, line 123: I think we should also broadcast the `ACQUIRE_TOKEN_SUCCESS/FAILURE` and `LOGIN_SUCCESS/FAILURE` events here. The tricky thing is going to be determining whether we are returning a login response or a token response. Maybe if we can figure that out the `HANDLE_REDIRECT_SUCCESS/FAILURE` events become redundant?

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -128,6 +149,8 @@ export abstract class ClientApplication {
             return null;
         }
 
+        this.broadcastEvent(BroadcastEvent.HANDLE_REDIRECT_START, InteractionType.REDIRECT);
```
tnorling, line 152: This should be removed as it's also broadcasted below in the `handleHash` function

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -454,6 +494,7 @@ export abstract class ClientApplication {
      * @param {@link (EndSessionRequest:type)} 
      */
     async logout(logoutRequest?: EndSessionRequest): Promise<void> {
+        this.broadcastEvent(BroadcastEvent.LOGOUT_START, null, logoutRequest);
```
tnorling, line 497: Should we wrap the `authClient.getLogoutUri(validLogoutRequest)` below in a try/catch that would broadcast `LOGOUT_SUCCESS` or `LOGOUT_FAILURE` event? Or would it be better to find a way to broadcast these events when we return from the server redirect?

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -76,8 +88,11 @@ export class PublicClientApplication extends ClientApplication implements IPubli
             // Telemetry manager only used to increment cacheHits here
             const serverTelemetryManager = this.initializeServerTelemetryManager(ApiId.acquireTokenSilent_silentFlow, silentRequest.correlationId);
             const silentAuthClient = await this.createSilentFlowClient(serverTelemetryManager, silentRequest.authority);
-            return await silentAuthClient.acquireCachedToken(silentRequest);
+            const cachedToken = await silentAuthClient.acquireCachedToken(silentRequest);
+            this.broadcastEvent(BroadcastEvent.ACQUIRE_TOKEN_SUCCESS, InteractionType.SILENT, cachedToken);
+            return cachedToken;
         } catch (e) {
+            this.broadcastEvent(BroadcastEvent.ACQUIRE_TOKEN_FAILURE, InteractionType.SILENT, null, e);
```
tnorling, line 95: This should be moved to the bottom of the catch block in `acquireTokenByRefreshToken` as this just represents a cache miss, not an actual failure

lib/msal-browser/src/app/ClientApplication.ts
jasonnutter: >  The tricky thing is going to be determining whether we are returning a login response or a token response.

In the Angular sample, we did this by looking at whether or not an account is logged in, and if so, then assume the redirect happened as a result of calling acquireTokenRedirect.

lib/msal-browser/src/event/BroadcastMessage.ts
```diff
@@ -0,0 +1,16 @@
+import { AuthenticationResult, AuthError, EndSessionRequest } from "@azure/msal-common";
+import { BroadcastEvent } from "./BroadcastEvent";
+import { InteractionType } from "../utils/BrowserConstants";
+import { PopupRequest, RedirectRequest, SilentRequest, SsoSilentRequest } from "..";
+
+export type BroadcastMessage = {
+    type: BroadcastEvent;
```
jasonnutter, line 7: We might want to give this field a more descriptive name, given `interactionType` is also present.

lib/msal-browser/test/app/PublicClientApplication.spec.ts
```diff
@@ -1789,4 +1791,80 @@ describe("PublicClientApplication.ts Class Unit Tests", () => {
...
+
+    describe("broadcastEvent and addEventCallback tests", () => {
+        it("can add an event callback and broadcast to it", () => {
+            let receivedMessage: BroadcastMessage;
+            const subscriber = (message) => {receivedMessage = message};
+            pca.addEventCallback(subscriber);
+            pca.broadcastEvent(BroadcastEvent.LOGIN_START, InteractionType.POPUP);
+
+            expect(receivedMessage.type).to.deep.eq(BroadcastEvent.LOGIN_START);
+            expect(receivedMessage.interactionType).to.deep.eq(InteractionType.POPUP);
```
jasonnutter, line 1803: For tests with callbacks like this, it is often better to use the `done` callback that is passed into the test, and invoke that in the callback when the message is received. This is because with tests that rely on callbacks or async behavior, that the test can complete before the `expect` statements are running, potentially resulting in a false positive.

```js
it("can add an event callback and broadcast to it", (done) => {
  const subscriber = (message) => {
    expect(message.type).to.deep.eq(BroadcastEvent.LOGIN_START);
    expect(message.interactionType).to.deep.eq(InteractionType.POPUP);
    done();
  }
  
  pca.addEventCallback(subscriber);
  pca.broadcastEvent(BroadcastEvent.LOGIN_START, InteractionType.POPUP);
});
```

change/@azure-msal-browser-2020-10-07-18-01-18-browser-add-broadcast.json
jo-arroyo: Do you mean major version?

lib/msal-browser/src/app/ClientApplication.ts
jo-arroyo: Done.

lib/msal-browser/test/app/PublicClientApplication.spec.ts
jo-arroyo: Makes sense. Modified.

change/@azure-msal-browser-2020-10-07-18-01-18-browser-add-broadcast.json
jasonnutter: Nope, minor.

lib/msal-browser/src/event/BroadcastMessage.ts
jo-arroyo: Sure. Changed it. Feel free to let me know if you think of a better name than eventType.

lib/msal-browser/src/app/ClientApplication.ts
tnorling: This may not work for v2 since we support multiple accounts now. We may need to leave ourselves a temp cache entry or include a flag or something in state

change/@azure-msal-browser-2020-10-07-18-01-18-browser-add-broadcast.json
jo-arroyo: Ah, I was looking at the wrong line. Changed.

tnorling: I believe `dependentChangeType` denotes how versions for libraries that depend on this one should be bumped. @pkanher617 can confirm. Currently it's just samples that depend on browser, so it should be none for now?

lib/msal-browser/src/app/PublicClientApplication.ts
jo-arroyo: Got it. Moved.

change/@azure-msal-browser-2020-10-07-18-01-18-browser-add-broadcast.json
jasonnutter: Whoops, I read the diff wrong. I think @pkanher617 wants this set to "none" for now.

lib/msal-browser/src/app/ClientApplication.ts
jasonnutter: Right. Something in the state object will likely be simplest.

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -771,5 +810,36 @@ export abstract class ClientApplication {
...
+        const message: BroadcastMessage = {
+            eventType: eventType,
+            interactionType: interactionType || null,
+            payload: payload || null,
+            error: error || null,
+            timestamp: Date.now()
+        };
+
+        this.eventCallbacks.forEach((callback) => {
+            callback(message);
```
jasonnutter, line 830: I think a verbose log message would be useful here (a message indicating an event with the given eventType is probably good enough to start).

change/@azure-msal-browser-2020-10-07-18-01-18-browser-add-broadcast.json
jo-arroyo: Okay. Updated.

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -771,5 +810,36 @@ export abstract class ClientApplication {
...
+        });
+    }
+
+    /**
+     * Adds event callbacks to array
+     * @param callback 
+     */
+    addEventCallback(callback: Function) {
```
jasonnutter, line 838: I think a type should be added for the callback that shows the parameter it will be passed. I think it would be:

`export type EventCallbackFunction = (message: BroadcastMessage) => void;`

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -46,6 +48,9 @@ export abstract class ClientApplication {
     // Flag to indicate if in browser environment
     protected isBrowserEnvironment: boolean;
 
+    // Callback for subscribing to events
+    private eventCallbacks: Function[];
```
jasonnutter, line 52: And then this would be `private eventCallbacks: EventCallbackFunction[]`

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -771,5 +810,36 @@ export abstract class ClientApplication {
...
 
+    /**
+     * Broadcasts events by calling callback with broadcast message
+     * @param eventType 
+     * @param interactionType 
+     * @param payload 
+     * @param error 
+     */
+    broadcastEvent(eventType: BroadcastEvent, interactionType?: InteractionType, payload?: BroadcastPayload, error?: BroadcastError) {
+        const message: BroadcastMessage = {
```
jasonnutter, line 821: We have a few different terms we are using, "broadcast", "event", "message". I think it would be good to not use "broadcast" (as it is Angular nomenclature) and pick one of the other two. Let's chat about it this afternoon.

jasonnutter (changes requested):
A few comments, looking good!

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -52,6 +53,9 @@ export class PublicClientApplication extends ClientApplication implements IPubli
      * @param {@link (RedirectRequest:type)}
      */
     async loginRedirect(request?: RedirectRequest): Promise<void> {
+        if (this.isBrowserEnvironment) {
```
jasonnutter, line 56: If we want to prevent events from happening server side, I think it would be cleaner to put this check in the `broadcastEvent` method itself.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -52,6 +53,9 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
+            this.broadcastEvent(BroadcastEvent.LOGIN_START, InteractionType.REDIRECT, request);
```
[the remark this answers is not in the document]
jasonnutter: And I think that's probably fine, I imagine they will be handled similarily. I think we'll just want to document it well.

lib/msal-browser/src/app/ClientApplication.ts
pkanher617: Broker is going to use "message" pretty heavily as well, so it may be good to make sure we can differentiate between event broadcast messages and window frame messages easily.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -63,6 +67,18 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
+        if (this.isBrowserEnvironment) {
+            this.broadcastEvent(BroadcastEvent.LOGIN_START, InteractionType.POPUP, request);
+            return this.acquireTokenPopup(request || DEFAULT_REQUEST)
+                .then((result) => {
+                    this.broadcastEvent(BroadcastEvent.LOGIN_SUCCESS, InteractionType.POPUP, result);
+                    return result;
+                })
+                .catch((e) => {
+                    this.broadcastEvent(BroadcastEvent.LOGIN_FAILURE, InteractionType.POPUP, null, e);
+                    throw e;
```
jasonnutter, line 79: I don't think there needs to be two different blocks, these should be able to hang off of the existing `acquireTokenPopup` invocation below.

lib/msal-browser/src/app/PublicClientApplication.ts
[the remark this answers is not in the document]
pkanher617: it could be a pitfall if someone is expecting something to happen on both starts, but as long as the success/error event only fires once I guess it is ok.

lib/msal-browser/src/app/PublicClientApplication.ts
jo-arroyo: Done.

lib/msal-browser/src/app/PublicClientApplication.ts
jo-arroyo: Cleaned up!

lib/msal-browser/src/app/ClientApplication.ts
jo-arroyo: Renamed to EventMessage, EventType, emitEvent after discussion.

lib/msal-browser/src/app/ClientApplication.ts
jo-arroyo: Done.

lib/msal-browser/src/app/ClientApplication.ts
jo-arroyo: Done.

lib/msal-browser/src/app/ClientApplication.ts
jo-arroyo: Done.

lib/msal-browser/src/app/ClientApplication.ts
jo-arroyo: We added the try/catch. Still open to suggestions on how it could be done after the server redirect, but the try/catch works for now.

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -771,5 +816,42 @@ export abstract class ClientApplication {
...
     }
 
+    /**
+     * Emits events by calling callback with event message
+     * @param eventType 
+     * @param interactionType 
+     * @param payload 
+     * @param error 
+     */
+    emitEvent(eventType: EventType, interactionType?: InteractionType, payload?: EventPayload, error?: EventError) {
```
jo-arroyo, line 826: Open to discussion: wanted to have this be protected, but found we could not test for it.

jasonnutter: I'll find a workaround next week.

lib/msal-browser/src/app/ClientApplication.ts
tnorling: > Is this mainly for telemetry? The event API doesn't really need to distinguish between login/acquireToken for programmatic reasons.

No, this is because people may subscribe to one event over the other and may care if a login happened or a token acquisition happened. I think it's also wise to have consistency between redirect and popup scenarios where possible

lib/msal-browser/src/app/ClientApplication.ts
jasonnutter: Marking as protected doesn't end up breaking tests.

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -382,7 +474,8 @@ export abstract class ClientApplication {
             // Create authorize request url
             const navigateUrl = await authClient.getAuthCodeUrl(silentRequest);
 
-            return await this.silentTokenHelper(navigateUrl, authCodeRequest, authClient);
+            const silentTokenResult = await this.silentTokenHelper(navigateUrl, authCodeRequest, authClient);
+            return silentTokenResult;
```
jasonnutter, line 478: Meant to revert this.

lib/msal-browser/src/app/IPublicClientApplication.ts
```diff
@@ -12,8 +12,9 @@ export interface IPublicClientApplication {
     acquireTokenPopup(request: PopupRequest): Promise<AuthenticationResult>;
     acquireTokenRedirect(request: RedirectRequest): Promise<void>;
     acquireTokenSilent(silentRequest: SilentRequest): Promise<AuthenticationResult>;
-    getAccountByUsername(userName: string): AccountInfo | null;
+    addEventCallback(callback: Function): void;
     getAccountByHomeId(homeAccountId: string): AccountInfo | null;
+    getAccountByUsername(userName: string): AccountInfo | null;
```
jasonnutter, line 17: Alphabetizing.

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -324,8 +384,23 @@ export abstract class ClientApplication {
             ThrottlingUtils.removeThrottle(this.browserStorage, this.config.auth.clientId, authCodeRequest.authority, authCodeRequest.scopes);
 
             // Handle response from hash string.
-            return await interactionHandler.handleCodeResponse(hash);
+            const result = await interactionHandler.handleCodeResponse(hash);
+
+            // If logged in, emit acquire token events
+            if (isLoggedIn) {
```
tnorling, line 390: We should check against the difference between the original account state and current state here to determine if an additional user was logged in or not

jasonnutter (approved):

tnorling (approved):

PR #2396, opened 2020-10-08
msal-react #3: uses event API

tnorling:


lib/msal-react/src/MsalProvider.tsx
```diff
@@ -3,91 +3,50 @@ import { useContext } from "react";
...
+        instance.addEventCallback((message: BroadcastMessage) => {
+            const eventTypes = [
+                BroadcastEvent.LOGIN_SUCCESS,
+                BroadcastEvent.ACQUIRE_TOKEN_SUCCESS,
+                BroadcastEvent.HANDLE_REDIRECT_SUCCESS,
+                BroadcastEvent.SSO_SILENT_SUCCESS,
+                BroadcastEvent.LOGOUT_SUCCESS
+            ];
+    
+            if (eventTypes.indexOf(message.type) > -1) {
```
jasonnutter, line 37: I think a switch statement would be cleaner here

tnorling: I think once we add a 2nd thing this callback needs to do that would make sense (error handling maybe?)

tnorling: done

PR #2404, opened 2020-10-09
Fix unexpected interaction_required error when using redirect flow

### LINKED ISSUE #2403: Inconsistent behavior of handleRedirectPromise causes spurious errors

## Library
- [x] `@azure/msal-browser@2.1.0`

## Description

Depending on the configuration value of `auth.navigateToLoginRequestUrl` and your redirect url, `handleRedirectPromise` may (or may not) return `null` when there is actually valid authentication data present. `handleRedirectPromise` actually redirects the browser to a new location and returns `null`. However, the redirect does not happen fast enough to avoid client code from seeing the `null` return value and initiating a new login redirect. This results in a `BrowserAuthError` with the code `interaction_in_progress`.

The heart of the problem is the idea that async code that calls a redirect does not need to return any result because you're leaving the page anyway. However, in practice the redirect may take long enough that the javascript code continues to execute with invalid results, leading to brief flashes of spurious error conditions.

## Security

## Regression
Version:

Unknown

## MSAL Configuration

```js
const msalConfig = {
    auth: {
        clientId: 'XXXXXXXXXX',
        authority: 'https://login.microsoftonline.com/XXXXXXXXXXXXX',
        redirectUri: `${window.location.origin}/learn/oauth`,
        // navigateToLoginRequestUrl is true by default
    },
    cache: {
    },
};

```

## Reproduction steps
```js

// inside authManager.js we have a login() function that returns Promise<AccountInfo
const login = (loginRequest = options.loginRequest, interactionType = loginInteractionType) => {
    return _authApp.handleRedirectPromise() .then(handleResponse(loginRequest, interactionType));
};

// note overall similarity to sample code at
// https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/38bb90e4e94a8d89363d6266d91390d33910d2f0/samples/msal-browser-samples/VanillaJSTestApp2.0/app/onPageLoad/auth.js#L23
const handleResponse = (loginRequest, interactionType) => (resp) => {
    const account = resp ? resp.account : getAccount(); // getAccount is basically getAllAccounts()[0]
    if (account) { // <--- PROBLEM: resp is null, account is null, so we call _authApp.loginRedirect(loginRequest) again
        return account;
    }

    if (interactionType === 'popup') {
        return _authApp.loginPopup(loginRequest).then((authResult) => authResult.account);
    } else {
        return _authApp.loginRedirect(loginRequest);
        // since the previous line also resolves to undefined instead on an AccountInfo, we actually include this workaround
        // to prevent the site from briefly rendering with a null account:
        // return _authApp.loginRedirect(loginRequest).then(() => new Promise((resolve) => setTimeout(resolve, 60000))).then(() => { throw new Error('Redirect failed') })
    }
};

// inside site index.js
// we often briefly see the interaction_in_progress error here
authManager
    .login()
    .then((account) => {
        ReactDOM.render(<App user={account} />, container);
    })
    .catch((error) => {
        ReactDOM.render(
            <div>Authentication failed: {error.toString()}.</div>,
            container
        );
    });
```

The problem is caused by:
* The user is signing in for the first time
* `auth.navigateToLoginRequestUrl` is `true`
* Your redirect page is not the same as the page you initially visited to initiate the redirect (for example, the redirect page is /oauth but the user visited /home)

The problem is here: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/679dab0146a70dde87dee58781bcbcdadf20397d/lib/msal-browser/src/app/ClientApplication.ts#L136

Given these conditions, control falls to the second `else if` on line 145. At this point there is a valid response hash available, but the code calls `BrowserUtils.navigateWindow(homepage, true);` without returning the processed hash. Again, the assumption is that you're leaving the page anyway so you don't need to return `this.handleHash(responseHash)`. But in practice this leads to spurious errors.

## Expected behavior

No spurious error conditions.

## Browsers/Environment
- [x ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari
- [ ] IE

tnorling:
Fixes #2403
When `navigateToLoginRequestUrl` is true and `handleRedirectPromise` needs to redirect back to the start page, it will return null before the page actually redirects. This potentially causes client code that is reliant on `handleRedirectPromise` to run. In a common scenario client code would see that no accounts are logged in and would call `loginRedirect` again, causing an `interaction_required` error to show up in the console before the redirect to the start page occurs.

This PR addresses this by delaying the resolution of `handleRedirectPromise` if a redirection has been triggered. If the redirect is successful the first call to `handleRedirectPromise` should never return as the page has changed. If for some reason the redirect fails, it will return null after some time.


lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -160,6 +160,9 @@ export abstract class ClientApplication {
                 // Navigate to page that initiated the redirect request
                 BrowserUtils.navigateWindow(loginRequestUrl, true);
             }
+
+            // Due to navigation this should not resolve, delay to ensure code dependant on handleRedirectPromise doesn't run before navigation
+            return new Promise(resolve => setTimeout(resolve, 5000, null));
```
jasonnutter, line 165: Not sure how I feel about an arbitrary delay like this. This would impact all customers right? Even those not calling `loginRedirect` after `handleRedirectPromise`?

jasonnutter: Ah, this only runs in scenarios where we are redirecting again, correct?

jasonnutter: Can the `5000` be put into a constant?

tnorling: This line is only hit when we need to go from redirectUri -> redirectStartPage. We don't actually expect a return in this scenario, but due to timing it does today.

jasonnutter: Another approach might to have `BrowserUtils.navigateWindow` return a promise and perform this delay

jasonnutter: Which would help prevent this bug from coming up in other scenarios.

lib/msal-browser/src/utils/BrowserUtils.ts
```diff
@@ -25,6 +26,11 @@ export class BrowserUtils {
         } else {
             window.location.assign(urlNavigate);
         }
+
+        // To block code from running after navigation, this should not throw if navigation succeeds
+        setTimeout(() => {
+            throw BrowserAuthError.createNavigationFailedError();
```
jasonnutter, line 32: Will this end up in teh catch of `handleRedirectPromise`?

lib/msal-browser/src/utils/BrowserUtils.ts
```diff
@@ -25,6 +26,11 @@ export class BrowserUtils {
...
+        }, BrowserConstants.NAVIGATION_TIMEOUT_MS);
```
jasonnutter, line 33: Also, 5 seconds may not be enough time...

lib/msal-browser/src/utils/BrowserUtils.ts
tnorling: Only if navigation does not occur

lib/msal-browser/src/app/ClientApplication.ts
pkanher617: +1 to move the delay to navigateWindow if it's required. I also think we should consider making this configurable.

lib/msal-browser/src/utils/BrowserUtils.ts
pkanher617: Rather than a constant can we make it a config param?

tnorling: I think throwing here makes it clear what happened rather than silently failing (meaning we failed to navigate away from the page). Is there a scenario you can think of where we would not want to throw here? We could make the timeout something much bigger, and configurable, if the concern is the potential of taking longer to leave the page.

change/@azure-msal-browser-2020-10-09-12-03-06-handleRedirectPromise-fix.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Fix unexpected interaction_required error in redirect flow (#2404)",
+  "packageName": "@azure/msal-browser",
+  "email": "thomas.norling@microsoft.com",
+  "dependentChangeType": "patch",
```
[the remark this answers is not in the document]
tnorling: Done

lib/msal-browser/src/utils/BrowserUtils.ts
tnorling: Changed to a log message instead of an error. Making it configurable is a little messy and maybe not necessary if we set the timeout to something fairly large (I set it to 30 seconds now, but feel free to suggest something different)

jmckennon (approved):

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -155,10 +155,10 @@ export abstract class ClientApplication {
                 // Cache the homepage under ORIGIN_URI to ensure cached hash is processed on homepage
                 this.browserStorage.setItem(this.browserStorage.generateCacheKey(TemporaryCacheKeys.ORIGIN_URI), homepage, CacheSchemaType.TEMPORARY);
                 this.logger.warning("Unable to get valid login request url from cache, redirecting to home page");
-                BrowserUtils.navigateWindow(homepage, true);
+                BrowserUtils.navigateWindow(homepage, this.logger, true);
```
pkanher617, line 158: We should try to revisit logger so we don't have to pass the handle to utils in the future.

lib/msal-browser/src/utils/BrowserUtils.ts
pkanher617: I still think making it configurable is the right way to go but if we want to leave it for a later PR that's fine.

pkanher617 (approved):
A few small comments for future PRs but otherwise lgtm.

lib/msal-browser/src/utils/BrowserUtils.ts
```diff
@@ -20,12 +21,17 @@ export class BrowserUtils {
...
-    static navigateWindow(urlNavigate: string, noHistory?: boolean): void {
+    static navigateWindow(urlNavigate: string, logger: Logger, noHistory?: boolean): void {
         if (noHistory) {
             window.location.replace(urlNavigate);
         } else {
             window.location.assign(urlNavigate);
         }
+
+        // To block code from running after navigation, this should not throw if navigation succeeds
+        setTimeout(() => {
```
jasonnutter, line 32: This won't block the code from executing, it just makes it so a log message gets printed out. I like the idea of making this function return a promise, and resolving that promise this time.

lib/msal-browser/src/utils/BrowserUtils.ts
```diff
@@ -20,12 +21,17 @@ export class BrowserUtils {
...
+            logger.warning("Expected to navigate away from the current page but timeout occurred.");
+        }, BrowserConstants.NAVIGATION_TIMEOUT_MS);
```
jasonnutter, line 34:
```suggestion
        return new Promise((resolve, reject) => {
             setTimeout(() => {
                 logger.warning("Expected to navigate away from the current page but timeout occurred.");
                 resolve();
             }, BrowserConstants.NAVIGATION_TIMEOUT_MS);
        });
```

lib/msal-browser/src/utils/BrowserUtils.ts
tnorling: Required a little bit of refactor of the `RedirectHandler` but implemented your suggestion. Thanks!

lib/msal-browser/src/utils/BrowserUtils.ts
tnorling: Was able to make the timeout configurable

lib/msal-browser/src/utils/BrowserUtils.ts
```diff
@@ -20,12 +20,20 @@ export class BrowserUtils {
...
             window.location.replace(urlNavigate);
         } else {
             window.location.assign(urlNavigate);
         }
+
+        // To block code from running after navigation, this should not throw if navigation succeeds
+        return new Promise((resolve) => {
+            setTimeout(() => {
+                logger.warning("Expected to navigate away from the current page but timeout occurred.");
+                resolve();
```
jasonnutter, line 34: Would `resolve(null)` mitigate the need for possibly returning `void` in the public api?

jasonnutter: (would be fine with that if so)

tnorling: yes, but most places that call `navigateWindow` actually do return void (loginRedirect, logout, RedirectHandler) If we want to not return void from `handleRedirectPromise` we can just await this there then resolve null.

jasonnutter: > If we want to not return void from handleRedirectPromise we can just await this there then resolve null instead of returning this directly

That works.

tnorling: Done

jasonnutter (approved):

PR #2407, opened 2020-10-10
Change Angular v2 sample to use msal-browser's event API

jo-arroyo:
Changing events in the Angular v2 sample to use the event API from msal-browser.


samples/msal-angular-v2-samples/angular10-browser-sample/src/app/msal/msal.broadcast.service.ts
```diff
@@ -1,25 +1,39 @@
...
-
-    this._msalSubject.next({type , payload});
+  /**
+   * Broadcasts events from msal-angular, same shape as events from msal-browser
+   * @param eventType 
+   * @param interactionType 
+   * @param payload 
+   * @param error 
+   */
+  broadcastAngularEvent(eventType: EventType, interactionType?: InteractionType, payload?: EventPayload, error?: EventError) {
```
jo-arroyo, line 28: To consider: If we are broadcasting events from msal-angular that are not the events listed in the msal-browser eventType, we may have to create a new type for the eventType, eventMessage, eventPayload, etc.

jasonnutter: I think this function can be deleted since we won't be doing any Angular-specific broadcasting at this time.

jo-arroyo: Makes sense. Removed.

samples/msal-angular-v2-samples/angular10-browser-sample/src/app/msal/msal.broadcast.service.ts
```diff
@@ -1,25 +1,39 @@
...
-  constructor() {
-    this._msalSubject = new Subject<MsalBroadcastMessage>();
+  constructor(
+    @Inject(MSAL_INSTANCE) private msalInstance: IPublicClientApplication
+  ) {
+    this._msalSubject = new Subject<EventMessage>();
     this.msalSubject$  = this._msalSubject.asObservable();
+    this.msalInstance.addEventCallback((message: EventMessage) => {
+      this._msalSubject.next(message);
+    });
```
jasonnutter, line 18: I think doing in the constructor is fine for now. We may want to think about moving it to its own "subscribe" or "initialize" function if there are problems (e.g. if this constructor gets invoked multiple times thus suscribing twice). Don't need to worry about for the purposes of this sample for now, but we'll want to keep keep it in mind for v2.

jasonnutter (approved):

jasonnutter:
Ah you may want to wait to merge this until the next browser version is published, so that the sample can have the version of msal-browser updated accordingly.

jmckennon (approved):

PR #2411, opened 2020-10-12
Add CDN doc for msal-core

jasonnutter:
Adds dedicated document for CDN usage for MSAL.js v1, including updated hashes for 1.4.1.

This will replace this page: https://github.com/AzureAD/microsoft-authentication-library-for-js/wiki/MSAL-JS-CDN-URL-and-SRI-Hash

### DIFF
--- a/lib/msal-core/cdn-upload.js
+++ b/lib/msal-core/cdn-upload.js
@@ -1,6 +1,7 @@
-const dotenv = require('dotenv');
-const { BlobServiceClient } = require('@azure/storage-blob');
+const dotenv = require("dotenv");
+const { BlobServiceClient } = require("@azure/storage-blob");
 const pkg = require("./package.json");
+const { ConsoleLogger } = require("typedoc/dist/lib/utils");
 
 dotenv.config({
     path: "../../.env"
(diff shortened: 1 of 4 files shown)


lib/msal-core/docs/cdn.md
```diff
@@ -0,0 +1,145 @@
...
+Microsoft offers CDN builds hosted in two different regions, US West and Europe North.
+
+CDN Domain          | Region       | Example
+--------------------| -------------| --------
+`alcdn.msauth.net`  | US West      | `https://alcdn.msauth.net/lib/<version>/js/msal.min.js`
+`alcdn.msftauth.net`| Europe North | `https://alcdn.msftauth.net/lib/<version>/js/msal.min.js`
+
+### CDN fallback
+
+In the unlikely event that a CDN build is broken or the CDN itself in inaccessible, an application can use the other CDN region as a fallback:
```
jasonnutter, line 41:
```suggestion
In the unlikely event that a CDN build is broken or the CDN itself is inaccessible, an application can use the other CDN region as a fallback:
```

lib/msal-core/docs/cdn.md
```diff
@@ -0,0 +1,145 @@
...
+* Use the Microsoft CDN instead of third-party CDNs.
+* Use the CDN region nearest to your users.
+* Use the `async` or `defer` attributes to not block page rendering.
+* Use the `integrity` attribute to ensure integrity of CDN builds.
+* IE11 support requires a Promise polyfill ([more information](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/samples/msal-core-samples/VanillaJSTestApp/app/ie11-sample/README.md) on IE11 support).
+
+## Basic Usage
+
+```html
+<script type="text/javascript" src="https://alcdn.msauth.net/lib/1.4.1/js/msal.min.js"></script>
```
jasonnutter, line 18: Trying to minimize the number of places where we would need to update with the latest version number.

lib/msal-core/cdn-upload.js
```diff
@@ -1,6 +1,7 @@
-const dotenv = require('dotenv');
-const { BlobServiceClient } = require('@azure/storage-blob');
+const dotenv = require("dotenv");
+const { BlobServiceClient } = require("@azure/storage-blob");
 const pkg = require("./package.json");
+const { ConsoleLogger } = require("typedoc/dist/lib/utils");
```
jasonnutter, line 4: Whoops, should be deleted.

tnorling (approved):

lib/msal-core/docs/cdn.md
```diff
@@ -0,0 +1,145 @@
+# CDN Usage of MSAL.js
+
+In addition to npm, `msal` can be consumed from Microsoft-hosted CDNs.
+
+## Best Practices
+
```
pkanher617, line 6: A TOC might be useful here since there are a lot of sections.

lib/msal-core/docs/cdn.md
pkanher617: We could keep the one here generic, and update the one in the README with the latest.

pkanher617 (approved):
A couple suggestions for the doc but otherwise looks good.

PR #2425, opened 2020-10-14
msal-react sample

tnorling:
Adds 2 msal-react samples:
- Default sample showing `Authenticated/UnauthenticatedTemplates` and logging in triggered by user interaction
- On page load sample using `MsalAuthenticationTemplate` and Error Boundary


jmckennon (approved):
Looks good, not sure if this is applicable but do we want a small amount of unit tests similar to how `msal-angular` does them in those samples?

tnorling:
> Looks good, not sure if this is applicable but do we want a small amount of unit tests similar to how `msal-angular` does them in those samples?

Yes, E2E tests will be added at some point before GA

package.json
```diff
@@ -24,6 +24,8 @@
     "lerna": "^3.20.2",
     "nyc": "^14.1.1",
     "puppeteer": "^2.1.1",
+    "react": "^16.13.1",
+    "react-dom": "^16.13.1",
```
jasonnutter, line 28: Why is this added to the root package.json?

tnorling: I was getting errors because the library and sample were using different instances of react even though they were the same version. Installing at the root seemed to fix it but I still want to play around with it to see if there's a better solution

tnorling: Fixed the issues I was having by having the sample import the libraries and react locally. Removed from the root.

PR #2426, opened 2020-10-14
Msal node export errors

sameerag:
This PR adds 'error' types in msal-node export types for users to discern between different error types and to be able to use `instanceOf` to classify their errors.


tnorling (approved):

hectormmg (approved):
LGTM!

PR #2427, opened 2020-10-14
Broker Redirect: Option 1

pkanher617:
Cache response in broker, and gives to embedded app on next API invocation with the same parameters


jasonnutter:
Closing in favor of #2428

PR #2430, opened 2020-10-15
Quickstart revision

mvrak:
Adding detail and in depth sample info for several standalone samples.  Some tutorial content.


samples/msal-node-samples/standalone-samples/README.md
```diff
@@ -1,12 +1,16 @@
 # MSAL Node Standalone Samples
 
-The sample applications contained in this directory are independent samples of MSAL Node usage, covering all of the authorization flows that MSAL Node currently supports. While all the samples include default configuration values that work out-of-the-box, their configuration is easily customizable to use your own Azure AD app registration.
+The sample applications contained in this directory are independent samples of MSAL Node usage, covering each of the authorization flows that MSAL Node currently supports. MSAL Node covers many situations, and most developers will only need one of these samples.
```
sameerag, line 3:
```suggestion
The sample applications contained in this directory are independent samples of MSAL Node usage, covering each of the authorization flows that MSAL Node currently supports. MSAL Node covers many scenarios, and most developers will only need one of these samples.
```

samples/msal-node-samples/standalone-samples/auth-code/readme.md
```diff
@@ -0,0 +1,203 @@
+# MSAL Node Standalone Sample:  Auth Code
+
+The sample applications contained in this directory are independent samples of MSAL Node usage, covering each of the authorization flows that MSAL Node currently supports. To get started with this sample, first follow the general instructions [here](../readme.me).
+
+Once MSAL Node is installed, and you have the right files, come here to learn about this scenario.
+
+### How is this scenario used?
+The Auth Code flow is most commonly used for a web app that signs in users.  General information about this scenario is available [here](https://docs.microsoft.com/en-us/azure/active-directory/develop/scenario-web-app-sign-user-overview?tabs=aspnetcore).
```
sameerag, line 8: Should we mention PKCE here?

samples/msal-node-samples/standalone-samples/auth-code/readme.md
```diff
@@ -0,0 +1,203 @@
...
+
+### Import the Configuration Object
+
+If you set up the sample with your app registration, you may be able to copy this object directly into your application.  
+
+
+```js
+const config = {
+    auth: {
+        clientId: "12d77c73-d09d-406a-ae0d-3d4e576f7d9b",
```
sameerag, line 71: Do we want to keep this `GUID`? Or have the user fill it?

samples/msal-node-samples/standalone-samples/auth-code/readme.md
sameerag: I meant, this seems to be `confidential client auth-code flow` from the `client-secret` desc below. Are we distinguishing between pca and cca apps here in this section?

samples/msal-node-samples/standalone-samples/auth-code/readme.md
```diff
@@ -0,0 +1,203 @@
...
+const msal = require('@azure/msal-node');
+```
+
+### Initialize MSAL Node at runtime
+
+
+Initialize the app object within your web app.
+
+```js
+const pca = new msal.ConfidentialClientApplication(config);
```
sameerag, line 101: `pca` and `client secret` - is that a correct combo?

samples/msal-node-samples/standalone-samples/README.md
```diff
@@ -1,12 +1,16 @@
...
+
+## Should I use this sample:
```
hectormmg, line 5: Maybe we should add a list of reasons of why people should use these samples instead of the one we use for automated tests. Otherwise we may want to remove this section?

samples/msal-node-samples/standalone-samples/README.md
```diff
@@ -20,23 +24,40 @@ HTTP:
...
 $ cd samples/msal-node-samples/standalone-samples/auth-code
 ```
 
+### Which sample should I review?
+
+Review our [scenario docs](https://docs.microsoft.com/en-us/azure/active-directory/develop/authentication-flows-app-scenarios) to pick a sample. 
+
+Continue to the appropriate sample to learn how it works, or to build it in to your existing app.
+
+- [Auth Code](auth-code/readme.md)  [web app]
```
hectormmg, line 55: Putting these in a table might be a good idea, columns can be "Sample" and "Usage Scenario" or something like that.

samples/msal-node-samples/standalone-samples/README.md
```diff
@@ -20,23 +24,40 @@ HTTP:
...
+- [Client Credentials](client-credentials/readme.md) [console app]
+- [Device Code](device-code/readme.md)
+- [On Behalf Of](on-behalf-of/web-app/readme.md) [calling MS Graph]
+- [Silent Flow](silent-flow/readme.md)
+
 ### Configure the application
 
 Each application has an `index.js` file, which is considered the entrypoint of the application. This file both contains the MSAL configuration and the application code that shows how to use each authorization flow. The sample works out-of-the-box with the default configuration in the `index.js` files, but you can customize it to use your own AzureAD app registration by changing the configuration values to match your app registration's. Below you'll find an example of what the main MSAL configuration object looks like. This is the object you'll want to customize to match your app registration.
```
hectormmg, line 63: Like we discussed, we may want to remove the "out-of-the-box" claim until we replace the client IDs with the ones for lab app registrations that any user can authenticate with.

samples/msal-node-samples/standalone-samples/README.md
mvrak: I'd prefer to defer this topic to later.  I think the section is fine for now.  I'm not fully sure if I can explain the automated testing sample in the way I was able to explain these.

samples/msal-node-samples/standalone-samples/README.md
mvrak: I thought of that, but hit a roadblock.  Not all flows are 1:1 with the terminology that an app developer may be thinking.  Using a hint like this doesn't say "this is what it is used for" rather, it is just a hint.

So, this is purposefully just a weaker "hint" statement rather than something I would put into a table.

samples/msal-node-samples/standalone-samples/README.md
mvrak: I agree, I will remove the out of the box claims.

samples/msal-node-samples/standalone-samples/auth-code/readme.md
mvrak: I didn't include language to distinguish cca and pca here.  Better covered on the docs site, the question is whether a developer would be wondering about that in the middle of this flow.  

I could mention PKCE, but didn't as I see there is another sample with a PKCE label?  We could have a separate PR for that.

samples/msal-node-samples/standalone-samples/auth-code/readme.md
mvrak: I should change the variable name.

samples/msal-node-samples/standalone-samples/auth-code/readme.md
mvrak: the guidance does indicate to replace the sample ID.  Since this is a cca, they would not have the secret anyways, so I have no concerns keeping it

mvrak: I prefer the "sample data" approach to the "ENTER ID HERE".  It maintains more context.

mvrak:
@hectormmg @sameerag let me know If I need to resolve those conflicts or if there is a more standard method.  I'm sure those package files are usually in conflict for branches like this.

sameerag:
> @hectormmg @sameerag let me know If I need to resolve those conflicts or if there is a more standard method. I'm sure those package files are usually in conflict for branches like this.

Unfortunately there is no getting around resolving the conflicts afaik. You can ignore the package-lock and just ensure the package.json is up-to-date. In this case pushing the latest once you pull from dev and run `npm i` would be the most preferred method.

samples/msal-node-samples/standalone-samples/auth-code/readme.md
sameerag: The only advantage with a `placeholder string` is that it fails with `invalid client_id` forcing the user to add it.

sameerag (approved):
Approved with one comment.

samples/msal-node-samples/standalone-samples/auth-code/readme.md
mvrak: so we can add the string _placeholder after the GUID?

samples/msal-node-samples/standalone-samples/auth-code/readme.md
mvrak: its not and good catch, i called this out when i did the demo sample for support.  easy 3 spots to change pca to cca

hectormmg (approved):

PR #2435, opened 2020-10-15
AuthCode with Non-PKCE flow clients

noant:
Hi! We found that msal.js not works with non-pkce auth code flow. It means that I can't use msal.js and Chinese Azure Ad B2C, because now it accepts only implicit flow and auth code flow with client secret. So, to provide client secret we should call token endpoint through our custom endpoint with addiction of client secret. Net Core Web Api example:
```
        [HttpPost]
        [AllowAnonymous]
        public async Task<ObjectResult> ExchangeCn()
        {
            using (var sr = new StreamReader(HttpContext.Request.Body)) 
            {
                var requestBody = sr.ReadToEnd();
                
                if (!requestBody.Contains($"client_id=<clientId>"))
                    throw new HttpRequestException("Invalid client_id parameter");

                var extendedRequestBody = $"{requestBody}&client_secret=<clientSecret>";

                var tokenRequestContent = new StringContent(extendedRequestBody, Encoding.UTF8, "application/x-www-form-urlencoded");

                foreach (var header in HttpContext.Request.Headers.Where(x => x.Key.StartsWith("x-client")))
                    tokenRequestContent.Headers.Add(header.Key, header.Value.ToArray());


                using var client = HttpClientFactory.Create(new HttpClientHandler() { AllowAutoRedirect = true });
                var response = await client.PostAsync("https://<domain>.b2clogin.cn/<tenant>/b2c_1_sign_1_cn/oauth2/v2.0/token", tokenRequestContent);

                return new ObjectResult(await response.Content.ReadAsStringAsync());
            }
        }
```

I added some classes (AuthorizationCodeClientNonPkce, RefreshTokenClientNonPkce) that can be used to access to our custom endpoint instead of the token endpoint. Provider settings example:

```
export const config = {
    auth: {
        clientId: process.env.MSAL_CLIENT_ID,
        authority: SIGN_IN_AUTHORITY,
        redirectUri: window.location.origin,
        postLogoutRedirectUri: window.location.origin,
        navigateToLoginRequestUrl: false,
        nonPkceAuthCodeExchangeEndpoint: 'https://<my_custom_endpoint>/api/v1.2/auth/exchange_cn', <<< custom endpoint parameter
        useNonPkceAuthCodeExchangeEndpoint: true, <<< if true - use custom endpoint to exchange auth code to tokens
        knownAuthorities: ['https://<some_known_authority>']
    },
    cache: {
        storeAuthStateInCookie: !!window.document.documentMode, // true for IE
    },
    system: {
        loadFrameTimeout: 30000,
        // tokenRenewalOffsetSeconds: 0,
    },
};
```

I have already created my personal npm packages forked from msal-browser and msal-common because this functionality is very important to my organization.
https://www.npmjs.com/package/msal-common-ext
https://www.npmjs.com/package/msal-browser-ext
So in this branch, package.json in msal-browser refers to standard msal-common package and for msal-browser debugging, msal-common link should be changed to currently locally created msal-common package with new classes. In dev branch in my fork msal-browser already uses msal-common-ext package from npm.

### DIFF
--- a/lib/msal-browser/package-lock.json
+++ b/lib/msal-browser/package-lock.json
@@ -1,5 +1,5 @@
 {
-    "name": "@azure/msal-browser",
+    "name": "msal-browser-ext",
     "version": "2.3.0",
     "lockfileVersion": 1,
     "requires": true,
@@ -137,13 +137,6 @@
             "dev": true,
             "requires": {
                 "tslib": "^1.9.3"
-            }
-        },
-        "@azure/msal-common": {
-            "version": "file:../msal-common/azure-msal-common-1.5.0.tgz",
-            "integrity": "sha512-sGuvCph4qVA2oq9ycvgUw68E9kfjysJtp9xqHxdRXKQ9JobpvAGlLedDHqJWtBJHQv077hnGZ93IZTfcb+O7rw==",
-            "requires": {
-                "debug": "^4.1.1"
             }
         },
         "@azure/storage-blob": {
@@ -3836,6 +3829,14 @@
             "version": "2.1.2",
             "resolved": "https://registry.npmjs.org/ms/-/ms-2.1.2.tgz",
             "integrity": "sha512-sGkPx+VjMtmA6MX27oA4FBFELFCZZ4S4XqeGOXCv68tT+jb3vk/RyaKWP0PTKyWtmLSM0b+adUTEvbs1PEaH2w=="
+        },
+        "msal-common-ext": {
+            "version": "1.5.0",
+            "resolved": "https://registry.npmjs.org/msal-common-ext/-/msal-common-ext-1.5.0.tgz",
+            "integrity": "sha512-0gDMACnXOYQQJJ8oYEehZG7rTHVVqRRxWXrKo7CuV+rkyDA6dPsHQCQmuTZXFImUYLwQg460P6NJch2rkC6JQg==",
+            "requires": {
+                "debug": "^4.1.1"
+            }
         },
         "neo-async": {
             "version": "2.6.1",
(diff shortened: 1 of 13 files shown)


pkanher617:
@noant Currently we cannot accept this. MSAL.js is a moniker for multiple libraries, so I'm not sure which library you are using, but the two that are supported for Single-Page Application flows are the implicit flow ([@azure/msal](https://www.npmjs.com/package/@azure/msal) or [msal](https://www.npmjs.com/package/msal)) and the authorization code flow ([@azure/msal-browser](https://www.npmjs.com/package/@azure/msal-browser)). I want to reiterate, these are for browser-only flows, with no backend component. We cannot use Authorization Code Flow without PKCE because of the inherent security flaws that are exposed by browsers and mobile devices. You can read more about this [here](https://tools.ietf.org/html/rfc7636#section-1).

The .NET web API you are describing above does not fall under this. What you are probably looking for is our [@azure/msal-node](https://www.npmjs.com/package/@azure/msal-node) library, which is currently in alpha, and will be in public preview shortly. This will support the flows you are describing above (including auth code flow without PKCE).

jasonnutter:
> It means that I can't use msal.js and Chinese Azure Ad B2C, because now it accepts only implicit flow and auth code flow with client secret.

Azure B2C should now support Auth Code Flow w/ PKCE for SPAs. Please open a support ticket if this does not appear to be working.

noant:
> 
> 
> > It means that I can't use msal.js and Chinese Azure Ad B2C, because now it accepts only implicit flow and auth code flow with client secret.
> 
> Azure B2C should now support Auth Code Flow w/ PKCE for SPAs. Please open a support ticket if this does not appear to be working.

I already did it, but it did not affect anything =) I think I should choose another oauth service for my organization, because it is too many problems with azure b2c.

jasonnutter:
Closing, as we will not be supported Auth Code without PKCE in MSAL Browser.

PR #2440, opened 2020-10-16
Update FAQ with note about B2C multiple access token workaround

tnorling:


lib/msal-browser/FAQ.md
```diff
@@ -185,3 +186,39 @@ You can read more about this behavior [here](https://docs.microsoft.com/azure/ac
 ## Why am I not signed in when returning from an invite link?
 
 MSAL.js will only process tokens which it originally requested. If your flow requires that you send a user a link they can use to sign up, you will need to ensure that the link points to your app, not the B2C service directly. An example flow can be seen in the [working with B2C](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/working-with-b2c.md) doc.
+
+## Why is there no access token returned from `acquireTokenSilent`?
+
+When exchanging a refresh token for a new access token the requested scopes must match the scopes that were requested when the refresh token was obtained. MSAL.js only holds one refresh token at a time. The B2C service is working on allowing refresh tokens to be redeemed for any consented scope but until that feature is rolled out these are the current workarounds:
```
nickgmicrosoft, line 192: Does the "When exchanging a refresh token for a new access token the requested scopes must match the scopes that were requested when the refresh token was obtained." apply only to Azure AD B2C or for AAD as well? If that part is also B2C only, maybe we can condense to:

```suggestion
Azure AD B2C currently requires refresh tokens to be redeemed with the same scopes that were requested when the refresh token is first obtained. If your application requires different behavior, workarounds include: 
```

nickgmicrosoft (changes requested):

lib/msal-browser/FAQ.md
tnorling: Thanks for the suggestion, this sounds better!

jmckennon (approved):

sameerag (approved):
Lgtm. One thing, can we please start adding `description` to all PRs we create?

PR #2448, opened 2020-10-18
[msal-node] update samples path

sameerag:
Update the samples path in the [Readme](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/README.md_


change/@azure-msal-node-2020-10-18-14-05-15-update-samples-path.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "prerelease",
```
tnorling, line 2:
```suggestion
  "type": "none",
```

tnorling (approved):

jasonnutter (approved):
Should address the beachball change file suggestion, otherwise looks good.

change/@azure-msal-node-2020-10-18-14-05-15-update-samples-path.json
sameerag: Done

PR #2469, opened 2020-10-21
Fix loginRedirect Angular 10 Sample Code

reesemclean:
The current msal.guard in the Angular 10 sample code does not use the scopes passed into the MsalGuardConfig during redirect login since it overwrites the given scopes with an empty array.

This updates the sample msal.guard to make sure all the values from the MsalGuardConfig are respected.


jasonnutter:
Good catch, thanks!

PR #2483, opened 2020-10-23
[msal-common] Add missing default headers to device code

sameerag:
`Content-type` and `Telemetry` headers are missing in `device code flow` as we are not calling the default headers API correctly. Fixing the issue with this PR.

Note: We mock all our network calls in unit tests and hence couldn't catch this in our testing. 

Todo:
Adding these in end-to-end tests is a good idea. Will try to see if the current framework allows this.

### DIFF
--- a/change/@azure-msal-common-2020-10-23-08-46-53-set-default-headers.json
+++ b/change/@azure-msal-common-2020-10-23-08-46-53-set-default-headers.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Add missing default headers to device code",
+  "packageName": "@azure/msal-common",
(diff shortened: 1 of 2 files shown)


pkanher617 (approved):

pkanher617 (commented):
Any tests that need to be updated?

jasonnutter (approved):

sameerag:
> Any tests that need to be updated?

Not sure. May be in E2E. Cannot add to UT because we mock our network calls. Introducing `nock` would be a good idea, so may be defer the testing into a separate issue would be a good idea since this PR cannot wait and is blocking Azure SDK.

PR #2495, opened 2020-10-27
Allow hash to be passed into handleRedirectPromise, reset non-msal after processing

### LINKED ISSUE #2268: Login Redirect fails in Angular 10 - MSAL Browser Sample
## Library
- [x] `@azure/msal-browser@2.1.0`

## Framework
Angular: I am testing the Angular 10 - MSAL Browser Sample

## Description
The Angular 10 - MSAL Browser Sample has been updated to include the ability to do a login redirect rather than a login popup.  However using this login redirect, the post login redirect (in the sample set to `http://localhost:4200`) does not correctly refresh and the Angular Web App (or specifically `MsalService`) is not aware of the successful login.  Calling `this.authService.getAllAccounts().length` in the `AppComponent` returns a length of 0.  After refreshing the Chrome browser, the Web Application is able to process the successful login.  In Safari browser the refresh also fails.

Issue is not present when login is provided with popup.

The sample uses `@azure/msal-browser` version 2.0.2, but I have tried 2.1.0 with the same results.

## Security

## Regression
Version: Not aware

## MSAL Configuration

```js
auth: {
}
```

## Reproduction steps
1. clone the `Angular 10 - MSAL Browser` Sample
2. `npm install` and `npm start`
3. visit `http://localhost:4200` and start login redirect by clicking "Login"
4. After providing Azure AD credentials, browser is redirected to `http://localhost:4200`, note that the "Login" button still shows and the "Logout" button is hidden.

Calling the Microsoft Graph in the `/profile` location will trigger a failing acquireToken call.

When I refresh Chrome, calling authService.getAllAccounts() returns the correct value. In Safari the browser refresh kicks off a failing acquireToken loop.

When I use interactionType: `InteractionType.POPUP` in the interceptor config, the post login redirect immediately shows the correctly refreshed Angular web app and the Logout button is showing rather than the Login button.  So, this seems related to the `InteractionType.REDIRECT` configuration only.

## Browsers/Environment
- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [x] Safari
- [ ] IE

jasonnutter:
When using hash routing in Angular and enabling `navigateToLoginRequestUrl`, msal will put restore the non-msal hash before the msal hash has been processed, resulting in Angular navigating too early. Furthermore, the Angular router will clear the msal hash before `handleRedirectPromise` has a chance to parse it, so `handleRedirectPromise` can now be given an explicit hash to process. This allows an app (or the MSAL Angular library) to capture the hash before the Angular router clears it, and then process that hash when `handleRedirectPromise` is called.


change/@azure-msal-browser-2020-10-26-18-16-36-handleredirectpromise-hash.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Allow hash to be passed into handleRedirectPromise, reset non-msal after processing",
+  "packageName": "@azure/msal-browser",
+  "email": "janutter@microsoft.com",
+  "dependentChangeType": "patch",
```
jasonnutter, line 6:
```suggestion
  "dependentChangeType": "none",
```

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -115,10 +115,10 @@ export abstract class ClientApplication {
      * auth flows.
      * @returns {Promise.<AuthenticationResult | null>} token response or null. If the return value is null, then no auth redirect was detected.
      */
-    async handleRedirectPromise(): Promise<AuthenticationResult | null> {
+    async handleRedirectPromise(hash?: string): Promise<AuthenticationResult | null> {
```
jasonnutter, line 118: Should probably update the jsdoc comments to make it clear hash is optional

tnorling (approved):
Looks good. Can we add a note to the docs that states the hash is optional and should only be provided explicitly if there's a valid reason to?

jasonnutter:
> Looks good. Can we add a note to the docs that states the hash is optional and should only be provided explicitly if there's a valid reason to?

Yep, totally agree.

samples/msal-angular-v2-samples/angular10-browser-sample/src/app/msal/msal.service.ts
```diff
@@ -27,10 +28,18 @@ interface IMsalService {
...
-        @Inject(MSAL_INSTANCE) private msalInstance: IPublicClientApplication
-    ) {}
+        @Inject(MSAL_INSTANCE) private msalInstance: IPublicClientApplication,
+        private location: Location
+    ) {
+        // Cache the code hash before Angular router clears it
+        const hash = this.location.path(true).split('#').pop();
+        if (hash) {
+            this.redirectHash = `#${hash}`;
+        }
```
jasonnutter, line 41: FYI @jo-arroyo

pkanher617 (approved):
docs updates should happen with this but seems ok to me.

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -113,6 +113,7 @@ export abstract class ClientApplication {
      * Event handler function which allows users to fire events after the PublicClientApplication object
      * has loaded during redirect flows. This should be invoked on all page loads involved in redirect
      * auth flows.
+     * @param hash Hash to process. Defaults to the current value of window.location.hash. Only needs to be provided explicitly if current value does not contain response to be handled.
```
jo-arroyo, line 116:
```suggestion
     * @param hash Hash to process. Defaults to the current value of window.location.hash. Only needs to be provided explicitly response to be handled is not contained in the current value.
```

jo-arroyo (approved):

lib/msal-browser/docs/initialization.md
```diff
@@ -113,6 +113,8 @@ This will also allow you to retrieve tokens on page reload. See the [onPageLoad
 
 It is not recommended to use both interaction types in a single application.
 
+**Note:** `handleRedirectPromise` will optionally accept a hash value to be processed, defaulting to the current value of `window.location.hash`. This parameter only needs to be provided in scenarios where the current value of `window.location.hash` does not contain the redirect response that needs to be processed. For almost all scenarios, applications should not need to provide this parameter explicitly.
```
pkanher617, line 116:
```suggestion
**Note:** `handleRedirectPromise` will optionally accept a hash value to be processed, defaulting to the current value of `window.location.hash`. This parameter only needs to be provided in scenarios where the current value of `window.location.hash` does not contain the redirect response that needs to be processed. **For almost all scenarios, applications should not need to provide this parameter explicitly.**
```

pkanher617 (approved):
One suggestion, looks good otherwise

tnorling (approved):

PR #2496, opened 2020-10-27
[msal-node] Remove "debug"

sameerag:
Since `debug` isn't currently supported in `msal-common` and `msal-browser`, removing it in `msal-node`. This will be added in future if all platforms align.


hectormmg (approved):

change/@azure-msal-node-2020-10-26-20-16-45-remove-debug.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "none",
```
tnorling, line 2: Since you are changing behavior inside the library
```suggestion
  "type": "prerelease",
```

tnorling (approved):
One comment, otherwise LGTM

tnorling:
Should we remove from msal-common as well?

sameerag:
> Should we remove from msal-common as well?

Well ... we can remove the commented code I suppose. I do want to have this fleshed out soon, we are removing it from `msal-node` as `msal-browser` had some issues integrating `debug`.

PR #2502, opened 2020-10-27
Add SRI script for core

jasonnutter:
Core PR for #2487


pkanher617 (approved):
one thing to check is whether docs are updated if the upload fails.

tnorling (approved):
This is pretty cool! Would we be able to do something similar to update the version number in the Constants file?

sameerag (approved):

jasonnutter:
> This is pretty cool! Would we be able to do something similar to update the version number in the Constants file?

Ah yeah forgot about that.

jasonnutter:
I'll address the core version bump and better error handling incrementally in a follow up PR.

PR #2505, opened 2020-10-27
msal-react useAccount hook

tnorling:
- Adds `useAccount` hook which internally calls `getAccountByUsername` or `getAccountByHomeId`
- Updates `useIsAuthenticated` to use new hook or `accounts` from context (this addresses a potential bug)
-  Replaces `loginInProgress` boolean with an `inProgress` enum describing what is in progress
- Provider calls `handleRedirectPromise`  and `useHandleRedirect` hook removed - this gives the library tighter control over timing and prevents potential bugs where context isn't properly updated. User will be expected to call `acquireTokenSilent` when they need a token 
- Allows user to provide an error component or loading component to `MsalAuthenticationTemplate`
- Fixes bugs in `useMsalAuthentication` hook causing login loops and state updates after unmount


lib/msal-react/src/MsalProvider.tsx
```diff
@@ -47,6 +47,8 @@ export function MsalProvider({instance, children}: MsalProviderProps) {
             }
         });
 
+        instance.handleRedirectPromise();
```
jasonnutter, line 50: Why is this needed here?

lib/msal-react/src/hooks/useAccount.ts
```diff
@@ -0,0 +1,31 @@
...
+        return instance.getAccountByUsername(accountIdentifiers.username);
+    }
+
+    return null;
+}
+
+export function useAccount(accountIdentifiers: AccountIdentifiers): AccountInfo | null {
+    const { instance, loginInProgress } = useMsal();
+
+    const [account, setAccount] = useState<AccountInfo|null>(null);
```
jasonnutter, line 24: Any concerns with keeping this in sync with `instance`?

lib/msal-react/src/MsalProvider.tsx
jasonnutter: I see the description now. What if an app needs to wait until `handleRedirectPromise` is completed in order to render? How would that work (given there is no longer a hook they can use)?

tnorling: I guess it depends what their render condition is. If the app needs to render based on whether or not a user is signed in it can use the `useIsAuthenticated` hook which would update after `handleRedirectPromise` finishes. 

If a component needs a token and needs to ensure `handleRedirectPromise` is done before calling `acquireTokenSilent` we will probably need to add an `interactionInProgress` boolean alongside the existing `loginInProgress` and we might need to move up the handleRedirect Start event.

lib/msal-react/src/MsalContext.tsx
```diff
@@ -19,6 +20,7 @@ export interface IMsalContext {
 const defaultMsalContext: IMsalContext = {
     instance: stubbedPublicClientApplication,
     loginInProgress: false,
+    interactionInProgress: false,
```
jasonnutter, line 23: Might consider a single value whose value is an enum instead of multiple booleans.

tnorling: Added. Thanks for the suggestion, I like this better!

PR #2512, opened 2020-10-28
Update Request Types

tnorling:
Updates request types to separate API surface from the request objects needed internally.


change/@azure-msal-browser-2020-11-06-14-12-06-update-request-types.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
```
sameerag, line 2:
```suggestion
  "type": "minor",
```

lib/msal-browser/src/index.ts
```diff
@@ -27,9 +28,6 @@ export {
     AuthenticationScheme,
     // Account
     AccountInfo,
-    // Request
-    AuthorizationUrlRequest,
```
jasonnutter, line 31: This would be a breaking change. You can do thins: `export { CommonAuthorizationCodeRequest as AuthorizationUrlRequest }`

change/@azure-msal-browser-2020-11-06-14-12-06-update-request-types.json
jasonnutter: Why? API surface isnt changing.

lib/msal-browser/src/index.ts
tnorling: Prithvi and I talked about this. It's technically a breaking change anyway because we're changing the underlying type but seeing as how people shouldn't be using this we wanted to remove it, if we need to bump the version accordingly, we can.

jasonnutter: It looks like the exported names in common are also changing. We need to guard against breaking apps that pick up teh new version of common but not the new version of browser.

jasonnutter: If the main reason for this is naming, then I would say leverage the `as` renaming for imports and exports, e.g `import { AuthorizationCodeRequest as CommonAuthorizationCodeRequest } from '@azure/msal-browser';`

jasonnutter: Otherwise if we need to make this a breaking change in common, we should do that.

tnorling: Not just naming, we're changing required vs. optional params on each of the types to ensure that requests passed around internally have all the required params without making them required on the API surface. It's a breaking change either way.

jasonnutter: It certainly seems like common should be a breaking change. Given `AuthorizationUrlRequest` shouldn't be used directly, I'm less concern about that changing from the browser export.

jasonnutter: For the browser request types that built off of `AuthorizationUrlRequest` (e.g. popuprequest), are we confident that the params that were being used haven't changed in a breaking way? e.g. making a param in popuprequest that was being used being changed from an optional to non-optional property.

change/@azure-msal-common-2020-11-06-14-12-06-update-request-types.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "minor",
```
jasonnutter, line 2:
```suggestion
  "type": "major",
```

lib/msal-browser/src/index.ts
tnorling: Yes because all those request types use the `Partial` utility type to make everything optional and then explicitly declares it's required types, if any.

change/@azure-msal-browser-2020-11-06-14-12-06-update-request-types.json
tnorling: Removing the export of `AuthorizationUrlRequest` may be worth making this a minor update

lib/msal-browser/src/index.ts
tnorling: Now being exported directly from browser

change/@azure-msal-browser-2020-11-06-14-12-06-update-request-types.json
tnorling: Will remain patch type

change/@azure-msal-common-2020-11-06-14-12-06-update-request-types.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "major",
```
pkanher617, line 2: Are we ok with doing msal-common@v2.0.0 right now?

pkanher617 (approved):
Looks ok, one question about versioning but otherwise we can merge.

Issue #2507, opened 2020-10-28
Expose OID aka localAccountId in AccountInfo


kostrse:
Please follow the issue template below. Failure to do so will result in a delay in answering your question.

## Library
- [ ] `msal@1.x.x` or `@azure/msal@1.x.x`
- [x] `@azure/msal-browser@2.x.x`
- [ ] `@azure/msal-angular@0.x.x`
- [ ] `@azure/msal-angular@1.x.x`
- [ ] `@azure/msal-angularjs@1.x.x`

## Description

In the MSAL v2 library, `AccountInfo` exposes `homeAccountId` which is user's OID in the home tenant, but doesn't expose the user's real identifiers like OID or SUB (aka `localAccountId` aka `uniqueId`).

Could you please expose it via public interface?

Even better would be to return specifically OID and/or SUB if they present, rather than abstracted out value `localAccountId`.

The problem is that, at least in our particular scenario, our app communicates with Graph API, and with other APIs, and in every case the OID is the actual unique identifier of the user, regardless of whether it's a normal user or a guest user from another tenant.

Or at least what is the way to retrieve the cached id-token for a given `AccountInfo` to obtain the OID without re-triggering the login process since you already have it in cache (`BrowserStorage`).

I understand that I could get it from `AuthenticationResult` after a successful login and store it on my own, but it's just basically re-implementation of own cache manager which msal already has (`BrowserStorage`).

sameerag:
@kostrse Thanks for raising this. I think this can be done, will soon raise a PR for the same.

CerosDev:
My solution is to perform a `acquireTokenSilent` to retrieve the Id Token. From my understanding, this should get the one from the cache if it's still valid. 

But doing so requires you to convert your function to async/await.

It would be great if the `getAllAccounts` could return the cached Id Token too. I believe this was the case in msal-core (msal 1.x)

kostrse:
I tried `acquireTokenSilent` as a workaround but it isn't perfect solution.

id-token doesn't have concept of expiration but `acquireTokenSilent` is meant for access tokens, so it triggers refresh of access token if it has expired even though I do not need access token itself.

I also have to provide some scopes, `acquireTokenSilent` cannot request access token without any scopes, even though I do not need them.

Access to collection of received id claims would be great and flexible enough since different application may request different claims (profile, sid, etc.).

TimoWilhelm:
[...]

PR #2516, opened 2020-10-29
[msal-common] Fixing a bug and adding `localAcccountId` in AccountInfo interface

sameerag:
* Currently when creating an account, `localAccountId` is supposed to be `oid` or `sub` in the absence of the former. There is a typo in the code where we are storing `sid` instead of `sub`. This PR rectifies this.

* There is also an ask from the community (#2507) to expose `localAccountId` in the `AccountInfo`. This PR adds that functionality.

Note: Unit Tests added


change/@azure-msal-common-2020-10-29-03-03-01-local-account-id.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Fixing a bug and adding `localAcccountId` in AccountInfo interface (#2516)",
+  "packageName": "@azure/msal-common",
+  "email": "sameera.gajjarapu@microsoft.com",
+  "dependentChangeType": "patch",
```
jasonnutter, line 6:
```suggestion
  "dependentChangeType": "none",
```

jasonnutter (approved):
Looks good! Thanks for tests!

change/@azure-msal-common-2020-10-29-03-03-01-local-account-id.json
sameerag: I don't understand why this doesn't reflect. I can swear made sure I added "none" this time. :-(

lib/msal-common/src/cache/entities/AccountEntity.ts
```diff
@@ -178,7 +179,7 @@ export class AccountEntity {
         oboAssertion?: string
     ): AccountEntity {
         const account: AccountEntity = new AccountEntity();
-        
+
         account.authorityType = (authority.authorityType === AuthorityType.Adfs) ? CacheAccountType.ADFS_ACCOUNT_TYPE : CacheAccountType.GENERIC_ACCOUNT_TYPE;
         account.homeAccountId = idToken.claims.sub;
```
tnorling, line 184: Can we add `localAccountId` to the generic account generator as well?

lib/msal-common/src/account/AccountInfo.ts
```diff
@@ -17,4 +17,5 @@ export type AccountInfo = {
     tenantId: string;
     username: string;
     name?: string;
+    localAccountId?: string;
```
kostrse, line 20: While OID is optional claim (requires `profile`), I think that SUB (subject) is always present in ID tokens.
So this field could be safely not nullable? Or I miss something?

sameerag: Yes. Except for `B2C` custom policies where there can be some exceptions. I am checking the data with B2C service team and will close this soon.

TimoWilhelm:
Just wanted to clarify my comment [in the related issue](https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/2507#issuecomment-718914484).

I'm facing the same issue as @kostrse because `AccountInfo` for one account that is invited as a guest user in multiple AAD tenants has the same `homeTenantId` and is ambiguous. So it would be great to add an additional convenience method next to the others in [ClientApplication.ts](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/36c6bc0bd6daba906f9d77bce31a543c137e3bf4/lib/msal-browser/src/app/ClientApplication.ts#L569-L614) since with this PR we can now use `localAccountId` as an actually unique identifier.

```ts
    /**
     * Returns the signed in account matching localAccountId.
     * (the account object is created at the time of successful login)
     * or null when no matching account is found
     * @returns {@link AccountInfo} - the account object stored in MSAL
     */
    getAccountByLocalId(localAccountId: string): AccountInfo|null {
        const allAccounts = this.getAllAccounts();
        if (!StringUtils.isEmpty(localAccountId) && allAccounts && allAccounts.length) {
            return allAccounts.filter(accountObj => accountObj.localAccountId === localAccountId)[0] || null;
        } else {
            return null;
        }
    }

```

sameerag:
> Just wanted to clarify my comment [in the related issue](https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/2507#issuecomment-718914484).
> 
> I'm facing the same issue as @kostrse because `AccountInfo` for one account that is invited as a guest user in multiple AAD tenants has the same `homeTenantId` and is ambiguous. So it would be great to add an additional convenience method next to the others in [ClientApplication.ts](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/36c6bc0bd6daba906f9d77bce31a543c137e3bf4/lib/msal-browser/src/app/ClientApplication.ts#L569-L614) since with this PR we can now use `localAccountId` as an actually unique identifier.
> 
> ```ts
>     /**
>      * Returns the signed in account matching localAccountId.
>      * (the account object is created at the time of successful login)
>      * or null when no matching account is found
>      * @returns {@link AccountInfo} - the account object stored in MSAL
>      */
>     getAccountByLocalId(localAccountId: string): AccountInfo|null {
>         const allAccounts = this.getAllAccounts();
>         if (!StringUtils.isEmpty(localAccountId) && allAccounts && allAccounts.length) {
>             return allAccounts.filter(accountObj => accountObj.localAccountId === localAccountId)[0] || null;
>         } else {
>             return null;
>         }
>     }
> ```

Thanks for the details @TimoWilhelm. I talked to @pkanher617  and we agree we can provide this functionality now. The upcoming tenant profiles feature will enhance more queries like `account -> fetch all tenants this user is logged in with` or `all accounts in a single tenant`; 

This functionality of `getAccountBylocalAccountId()` need not wait until we push that one out.

Will be updating the PR soon.

pkanher617 (approved):

lib/msal-common/src/cache/entities/AccountEntity.ts
sameerag: Done

lib/msal-common/src/account/AccountInfo.ts
sameerag: We decided to set it to "" in case of no `sub` use case, which shouldn't occur in any of the usual scenarios.

CerosDev:
Are we expecting a release of this soon? #2524 is a major issue

sameerag:
Yes, We will release a patch soon.

PR #2518, opened 2020-10-29
Add handleRedirect End Event

tnorling:
Adds handleRedirect end event and moves handleRedirect start event to the beginning of `handleRedirectPromise`


lib/msal-browser/src/event/EventType.ts
```diff
@@ -15,6 +15,7 @@ export enum EventType {
     SSO_SILENT_SUCCESS = "msal:ssoSilentSuccess",
     SSO_SILENT_FAILURE = "msal:ssoSilentFailure",
     HANDLE_REDIRECT_START = "msal:handleRedirectStart",
+    HANDLE_REDIRECT_END = "msal:handleRedirectEnd",
```
jasonnutter, line 18: Do we want to do success and failure?

tnorling: My feeling was that success and failure are covered by the Login and AcquireToken success and failure events

jasonnutter (approved):

hectormmg (approved):

PR #2520, opened 2020-10-29
[msal-node] Add getLogger and setLogger

jo-arroyo:
This PR adds getLogger and setLogger functions to msal-node.


change/@azure-msal-node-2020-10-29-13-30-16-msal-node-set-logger.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "minor",
```
hectormmg, line 2: For msal-node these still fall under "prerelease" if I'm not mistaken.
```suggestion
  "type": "prerelease",
```

hectormmg (approved):

change/@azure-msal-node-2020-10-29-13-30-16-msal-node-set-logger.json
jo-arroyo: Oh, good to know! Thanks!

sameerag (approved):

PR #2523, opened 2020-10-29
Restore string to cacheLocation type

jasonnutter:
Unfortunately, using the value of a value in an enum is not good enough for typescript, so we need to restore `string` as a possible type. The `BrowserCacheManager` will [handle a string literal value correctly](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/src/cache/BrowserCacheManager.ts#L52-L66) at runtime.

### DIFF
--- a/change/@azure-msal-browser-2020-10-29-16-03-05-msal-browser-cache-location-string.json
+++ b/change/@azure-msal-browser-2020-10-29-16-03-05-msal-browser-cache-location-string.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Restore string to cacheLocation type",
+  "packageName": "@azure/msal-browser",
+  "email": "janutter@microsoft.com",
+  "dependentChangeType": "patch",
+  "date": "2020-10-29T23:03:05.675Z"
+}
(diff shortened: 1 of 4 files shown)


change/@azure-msal-browser-2020-10-29-16-03-05-msal-browser-cache-location-string.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Restore string to cacheLocation type",
```
jasonnutter, line 3:
```suggestion
  "comment": "Restore string to cacheLocation type (#2523)",
```

change/@azure-msal-browser-2020-10-29-16-03-05-msal-browser-cache-location-string.json
```diff
@@ -0,0 +1,8 @@
...
+  "packageName": "@azure/msal-browser",
+  "email": "janutter@microsoft.com",
+  "dependentChangeType": "patch",
```
jasonnutter, line 6:
```suggestion
  "dependentChangeType": "none",
```

pkanher617 (approved):
LGTM

jo-arroyo (approved):

sameerag (approved):
So it is a String `enum` value compared to user input (string) runtime?

PR #2525, opened 2020-10-30
msal-react unit tests

tnorling:
Adding basic unit tests to msal-react library


pkanher617 (approved):

pkanher617:
If we want to use jest across the lib we should schedule some time to migrate the other libs.

PR #2533, opened 2020-11-02
sets homeAccountIdentifier to idTokenObj.claims.sub for AD FS workflo…

### LINKED ISSUE #2532: acquireTokenSilent doesn't work for AD FS workflow

## Library
- [x] `@azure/msal-browser@2.5.1`

## Description

User successfully acquired accessToken and refresh token, but there is no client_info in the response
https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/fef918b8cdbff3dc82aea3ca41118a0130ac1d40/lib/msal-common/src/response/ResponseHandler.ts#L115
so homeAccountId will stay empty on the other hand we set homeAccountId on the account here
https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/fef918b8cdbff3dc82aea3ca41118a0130ac1d40/lib/msal-common/src/cache/entities/AccountEntity.ts#L183

later when I'm trying to refresh token it's not possible to get refreshToken from the storage, as there is no homeAccountId associated with it

So proposal is to do the same as we do for account, set it from `idTokenObj.claims.sub` for AD FS workflow.

Thoughts? I'll create PR...

MariMax:
…w, fixes #2532

### DIFF
--- a/lib/msal-common/src/response/ResponseHandler.ts
+++ b/lib/msal-common/src/response/ResponseHandler.ts
@@ -133,6 +133,10 @@ export class ResponseHandler {
                     throw ClientAuthError.createNonceMismatchError();
                 }
             }
+
+            if (authority && authority.authorityType === AuthorityType.Adfs && !this.homeAccountIdentifier) {
+                this.homeAccountIdentifier = idTokenObj.claims.sub || "";
+            }
         }
 
         // save the response tokens


jasonnutter:
I believe will be addressed in #2516. Can you please take a look at that PR?

MariMax:
I doubt #2516 will help problem is acquireTokenSilent cannot find details in cache, because it trying to get it based on 
account.homeAccountId and account.environment see here
https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/29c5e1a46f8ac4265a96070d5a50222763c2b211/lib/msal-common/src/cache/CacheManager.ts#L542

fixing localAccountId won't help in this case

sameerag:
Thank you for this @MariMax. We made changes for #2516 to add this use case. I just pushed the changes, introduced a centralized helper API to generate `homeAccountId`. Can you please check that?

MariMax:
@sameerag yep that should work, thank you!
handled in #2516 closing

PR #2542, opened 2020-11-03
Change label for bug reports

jasonnutter:
Currently, bug reports are given the bug label by default. Proposing bug reports are marked as questions until they are confirmed as bugs.


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -2,7 +2,7 @@
 name: Bug report
 about: Broken or unintended behavior with one of our libraries.
 title: ''
-labels: bug
+labels: question
```
jasonnutter, line 5: We could also add another label such as "unconfirmed bug", which gets replaced with "bug" once confirmed.

tnorling: I like this option, start with `Bug: Unconfirmed` and then manually change to `Question` and `Bug: Confirmed` or just `Bug`

Big fan of changing this regardless of what we decide to change it to!

sameerag: I am leaning towards "unconfirmed" too instead of "question". +1 for changing this.

jo-arroyo: +1 on unconfirmed!

jasonnutter:
```suggestion
labels: bug-unconfirmed
```

jo-arroyo (approved):

tnorling (approved):

PR #2545, opened 2020-11-04
Add navigateFrameWait and change loadFrameTimeout to browser to match core behavior

### LINKED ISSUE #2314: loadFrameTimeout not doing what it is documented to in msal-browser?
## Library
- [X] `@azure/msal-browser@2.0.2`

## Framework
React
## Description
Noticed this while I was reading what `loadFrameTimeout` does, in the documentation:
https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/90e0231e749e11c200290322abdcc32b67cddfb3/lib/msal-browser/src/config/Configuration.ts#L55

But in the code https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/2c75dcfdc3af95ce942433d950ceb95de085c3da/lib/msal-browser/src/interaction_handler/SilentHandler.ts#L105, I see it is being used in a setTimeout. How I'm understanding this is that this value delays the load of the iframe. Is this expected?
## Error Message
No error, just wondering if the config is doing what it says it does in the documentation 
## Security

## Regression
Version:

## MSAL Configuration

```js
```

## Reproduction steps
```js
```

## Browsers/Environment
- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari
- [ ] IE

jasonnutter:
This PR adds back `navigateFrameWait` from v1 and changes `loadFrameTimeout` to function as it did in v1. Currently in v2, `loadFrameTimeout` is being used for what `navigateFrameWait` did in v1, which is semantically incorrect and creates confusion (#2314). This is especially to important to fix before Angular and React release alphas, as we will have a large number of users from v1 migrating to v2.

Given the existence in v2 of `windowHashTimeout` and `iframeHashTimeout`, `loadFrameTimeout` will be used to set the defaults for those values if they are not provided.

PR also fixes the behavior of `loadFrame` to add the iframe to the dom then async set the src, instead of async doing both at the same. This is to match to intent of the [original change in v1](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/v1.0.0/lib/msal-core/src/UserAgentApplication.ts#L1616-L1629).

And the error message for iframe timeouts has been fixed to say "iframe" instead of "popup".


lib/msal-browser/src/config/Configuration.ts
```diff
@@ -125,9 +127,10 @@ export function buildConfiguration({ auth: userInputAuth, cache: userInputCache,
         ...DEFAULT_SYSTEM_OPTIONS,
         loggerOptions: DEFAULT_LOGGER_OPTIONS,
         networkClient: BrowserUtils.getBrowserNetworkClient(),
-        windowHashTimeout: DEFAULT_POPUP_TIMEOUT_MS,
-        iframeHashTimeout: DEFAULT_IFRAME_TIMEOUT_MS,
-        loadFrameTimeout: BrowserUtils.detectIEOrEdge() ? 500 : 0,
+        // If loadFrameTimeout is provided, use that as default.
+        windowHashTimeout: (userInputSystem && userInputSystem.loadFrameTimeout) || DEFAULT_POPUP_TIMEOUT_MS,
+        iframeHashTimeout: (userInputSystem && userInputSystem.loadFrameTimeout) || DEFAULT_IFRAME_TIMEOUT_MS,
```
jasonnutter, line 132: An argument could be made that if `loadFrameTimeout` is set to a really low number (due to how its used in v2) that is smaller than the existing defaults, we should use the defaults (to prevent unintentional timeouts).

change/@azure-msal-browser-2020-11-04-11-03-20-navigateFrameWait-browser.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Add navigateFrameWait and change loadFrameTimeout to browser to match core behavior",
+  "packageName": "@azure/msal-browser",
+  "email": "janutter@microsoft.com",
+  "dependentChangeType": "patch",
```
jasonnutter, line 6:
```suggestion
  "dependentChangeType": "none",
```

change/@azure-msal-browser-2020-11-04-11-03-20-navigateFrameWait-browser.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Add navigateFrameWait and change loadFrameTimeout to browser to match core behavior",
```
jasonnutter, line 3:
```suggestion
  "comment": "Add navigateFrameWait and change loadFrameTimeout to browser to match core behavior (#2545)",
```

lib/msal-browser/src/error/BrowserAuthError.ts
```diff
@@ -45,10 +45,14 @@ export const BrowserAuthErrorMessage = {
         code: "user_cancelled",
         desc: "User cancelled the flow."
     },
-    monitorWindowTimeoutError: {
+    monitorPopupTimeoutError: {
         code: "monitor_window_timeout",
         desc: "Token acquisition in popup failed due to timeout."
     },
+    monitorIframeTimeoutError: {
+        code: "monitor_window_timeout",
```
jasonnutter, line 53: Keeping error code the same as there are likely apps relying on this error code for certain error handling logic.

lib/msal-browser/src/config/Configuration.ts
jasonnutter: At a minimum there should probably be a log message.

pkanher617: I see the default value as a way to guarantee that unintentional timeouts would not happen, and the user can fine-tune this for their application. I am ok with a log message but would probably vote against overwriting smaller values.

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -66,6 +67,7 @@ export type BrowserSystemOptions = SystemOptions & {
     windowHashTimeout?: number;
     iframeHashTimeout?: number;
     loadFrameTimeout?: number;
```
pkanher617, line 69: Should we add backwards compatibility/deprecation log messages for existing users using this parameter?

jo-arroyo (approved):
LGTM!

lib/msal-browser/src/config/Configuration.ts
pkanher617: Nvm, backwards compat isn't an option since it is used as defaults for the hashTimeouts. However I think a deprecation/change-of-behavior log message would be good.

pkanher617 (approved):
One comment about logs for the `loadFrameTimeout` parameter, but otherwise seems fine. Changing the public API surface should also warrant some changes to the docs for configuration.

lib/msal-browser/src/config/Configuration.ts
pkanher617: Sorry I misread this code. I would prefer not to overwrite user values, but in this case behavior of the parameter is changing, and it may break users using both `loadFrameTimeout` and `windowHashTimeout/iframeHashTimeout`.

jasonnutter: Adding log messages for now, will look at overwriting user values if we get feedback accordingly.

Issue #2452, opened 2020-10-19
msal-browser v 2.3.1 & v2.1.0: logout() works but logout(logoutRequest) throws error


willzhan:
>> Please follow the issue template below. Failure to do so will result in a delay in answering your question.

## Library
- [ ] `@azure/msal-browser@2.1.0`
- [ ] `@azure/msal-browser@2.3.1`

**Important**: Please fill in your exact version number above, e.g. `msal@1.1.3`.

## Description
Try to use logout API: logout() works well, while logout(logoutRequest) throws error. I have made sure that logoutRequest.account = the login account.

### msal-browser v 2.3.1: 

Uncaught (in promise) ClientAuthError: no_account_found: No account found in cache for given key.
    at X.u [as constructor] (https://alcdn.msauth.net/browser/2.3.1/js/msal-browser.min.js:2:4330)
    at new X (https://alcdn.msauth.net/browser/2.3.1/js/msal-browser.min.js:2:10991)
    at Function.X.createNoAccountFoundError (https://alcdn.msauth.net/browser/2.3.1/js/msal-browser.min.js:2:10352)
    at Un.St.removeAccount (https://alcdn.msauth.net/browser/2.3.1/js/msal-browser.min.js:2:40894)
    at Rr.getLogoutUri (https://alcdn.msauth.net/browser/2.3.1/js/msal-browser.min.js:2:66489)
    at io.<anonymous> (https://alcdn.msauth.net/browser/2.3.1/js/msal-browser.min.js:2:129952)
    at https://alcdn.msauth.net/browser/2.3.1/js/msal-browser.min.js:2:1919
    at Object.next (https://alcdn.msauth.net/browser/2.3.1/js/msal-browser.min.js:2:2024)
    at r (https://alcdn.msauth.net/browser/2.3.1/js/msal-browser.min.js:2:795)

### msal-browser v 2.1.0 error: 

[...]

Issue #2127, opened 2020-08-10
allow "logout" of a specific application but not end the entire AAD session, which logs you out of all apps


aappddeevv:
Please follow the issue template below. Failure to do so will result in a delay in answering your question.

## Library
- [ ] `msal@1.x.x` or `@azure/msal@1.x.x`
- [X ] `@azure/msal-browser@2.x.x`
- [ ] `@azure/msal-angular@0.x.x`
- [ ] `@azure/msal-angular@1.x.x`
- [ ] `@azure/msal-angularjs@1.x.x`

## Description
If one is using outlook on the web as well as an app that is leverage AAD, if you logout the app, you are also logging out of the outlook application. The feature request is for a way to just "logout" (perhaps clear the app specific storage) the app without logging out of the entire AAD session.

I have a few apps that are based on a registered app in AAD in the same tenant as the outlook web app (e.g. O365). Logging out of one app seems to log you out of all apps, which to me at least, was unexpected. I had expected to only be logged out of the account/app that was specified in the `logout` call.

technical-boy:
Hi @aappddeevv . Thanks for the feedback, we'll take this under consideration. As you mention, if you'd like this functionality now, you can manually clear or delete values from local storage.

aappddeevv:
Of course, however, I do not want to tell my users to go clear local storage. I would need to duplicate the key naming logic to clear those entries programattically.

jasonnutter:
This will be addressed in #2669, which will add the `onRedirectNavigate` callback to the logout API.

PR #2546, opened 2020-11-05
Logout should only clear local cache when run inside iframe

### LINKED ISSUE #2145: How to disable post-logout redirect

## Library

## Description
The default logout behavior redirects users back to the current location.href. This doesn't work for my app because if a user is not authenticated we auto redirect users to login creating a confusing UX on logout. Instead, I would like to not redirect the user at all after the logout is successful. How can I disable the default redirect behavior? I've tried setting the msal config `postLogoutRedirectUri: ''` without success. 
I believe our current auth solution achieves this behavior by simply not providing the redirect param.

jasonnutter:
When an application's logout broadcast url is loaded (known as "Logout URL" in the portal, and is different than the post logout redirect uri), we should be advising applications to invoke the MSAL logout function to clear local cache. While this can be done today, the logout function should not attempt to redirect, as the logout broadcast url will be loaded inside an iframe.

This PR will also address the bugs raised in:
- #2452
- #2145 

The application's logout broadcast URL will be given the session `sid` and `iss` as query parameters. Guidance from the server is to inspect cached accounts and if there is an account that matches the provided `sid`, only remove that account (this will require exposing `idTokenClaims` in `AccountInfo`). If no account matches the given `sid` (likely because `sid` was not enabled for the app), then clear all accounts. 

TODO:
- [ ] Tests
- [ ] Documenation about logout and logout broadcast url


lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -105,24 +104,16 @@ export class AuthorizationCodeClient extends BaseClient {
             throw ClientConfigurationError.createEmptyLogoutRequestError();
         }
 
-        if (logoutRequest.account) {
-            // Clear given account.
-            this.cacheManager.removeAccount(AccountEntity.generateAccountCacheKey(logoutRequest.account));
-        } else {
-            // Clear all accounts and tokens
-            this.cacheManager.clear();
-        }
```
jasonnutter, line 114: Clearing cache and creating logout url should be separated.

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -554,12 +554,33 @@ export abstract class ClientApplication {
...
+             */
+            if (logoutRequest && logoutRequest.account) {
+                // Clear given account.
+                this.browserStorage.removeAccount(AccountEntity.generateAccountCacheKey(logoutRequest.account));
+            } else {
+                // Clear all accounts and tokens
+                this.browserStorage.clear();
+            }
+
+            if (!BrowserUtils.isInIframe()) {
```
tnorling, line 570: There was also an ask in #2127 to only clear cache, regardless of iframe or not. Can we provide a flag on the `EndSessionRequest` object to skip the redirect or do something similar to `onRedirectNavigate`?

jasonnutter: Yeah, we can do that.

jasonnutter: The `onRedirectNavigate` PR will add this for logout.

PR #2560, opened 2020-11-07
Support relative redirectUris

### LINKED ISSUE #1896: Would it be possible to use relative paths for the redirectUri?
## Library
- [x] `@azure/msal-browser@2.0.0-beta-3`

## Description

Using a relative path for the `redirectUri` fails:

```typescript
export const login = async () => {
  if (isLoginPopup) {
    return auth.loginPopup({ redirectUri: '/', scopes: allScopes })
  }
    return auth.loginRedirect({ redirectUri: '/', scopes: allScopes })
}
```
image

Would it be possible to support relative paths to be used in the redirectUri? Things like `/`, `/about`, `/settings`, .... instead of a FQDN  like`http://localhost:8080/`. It would be great if the library was able to stitch these two together behind the scenes.

tnorling:
Adds support in `msal-browser` for providing relative paths on the `redirectUri` and `postLogoutRedirectUri` parameters

If a relative path is provided, `msal-browser` will create the absolute path as part of the request validation step

Closes #1896


lib/msal-common/src/url/UrlString.ts
```diff
@@ -157,6 +157,17 @@ export class UrlString {
 
         return match[2];
     }
+
+    static getAbsoluteUrl(relativeUrl: string, baseUrl: string): string {
+        if (relativeUrl[0] === Constants.FORWARD_SLASH) {
```
jasonnutter, line 162: So this would only support relative urls that start with `/`? In theory we could also support relative redirect uris that start with a folder path, for example.

tnorling: Not sure I know what that would look like. Don't all `redirectUris` need to be in the format `https://domain.com/path` ? All we're trying to do here is make the `https://domain.com` part optional.

jasonnutter: All of these are technically relative urls:

- `/path/auth.html`
- `path`
- `./path`
- `path1/path2`
- `auth.html`

tnorling: Gotcha, guess I misunderstood your original comment. I thought about this but I feel like supporting some of those makes url validation a lot harder. Let's say someone passes in just `myApp.com` as a `redirectUri`. String matching logic intended to match something like `auth.html` as a relative path may turn the `redirectUri` into `https://myApp.com/myApp.com` instead of the intended `https://myApp.com`. I think it's safer and less prone to unexpected behavior to tell people if they want to use relative paths it needs to start with `/` (which I think is the most common anyway)

jasonnutter: Yeah, makes sense, and we can always add in enhanced relative uri logic later if needed.

pkanher617 (approved):

sameerag (approved):

PR #2569, opened 2020-11-09
Update issue templates to include new libraries

jo-arroyo:
Updating Github issue templates to include Node, Angular v2, and React.

PR #2572, opened 2020-11-09
Update issue templates

mvrak:
Added references for msal-node beta to all templates.


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -15,6 +15,7 @@ assignees: ''
 - [ ] `@azure/msal-angular@0.x.x`
 - [ ] `@azure/msal-angular@1.x.x`
 - [ ] `@azure/msal-angularjs@1.x.x`
+- [ ] `@azure/msal-node@1.x.x-beta.x`
```
jasonnutter, line 18: Small thing: I would say we don't need to have the prerelease names include, given someone may still be using alpha, and so that we don't have to update against once GA.

```suggestion
- [ ] `@azure/msal-node@1.x.x`
```

tnorling:
FYI these changes are also included in PR #2569

.github/ISSUE_TEMPLATE/bug_report.md
mvrak: @sameerag @hectormmg  What do you think?  Since you proposed using the beta designation?  

I don't personally think this will affect the qualify of the submissions or our ability to address them.  I'm open to either choice or just saying ...msal-node@beta

jo-arroyo:
Closing as issues template was updated in PR #2569 .

PR #2587, opened 2020-11-11
Support id_token_hint on logout requests

### LINKED ISSUE #996: Provide config option to include id_token in B2C logout requests
## I'm submitting a...

<pre><code>
[ ] Bug report  
[ ] Performance issue
[X] Feature request
[ ] Other... Please describe:
</code></pre>

## Library version

<pre><code>
Library version: 1.0.4

</code></pre>

To provide additional security and avoid CSRF, it is possible to configure Azure AD B2C to require the `id_token_hint` parameter during logout. When this is turned on, logouts via MSAL.js will fail. MSAL should provide a configuration option to include this parameter when initiating log out.

See this docs article under "Secure your logout redirect". https://docs.microsoft.com/en-us/azure/active-directory-b2c/active-directory-b2c-reference-oidc#secure-your-logout-redirect

tnorling:
In B2C scenarios if a policy has the "Require Id Token on Logout Requests" option enabled, users must provide an `id_token_hint` parameter on logout requests. This PR adds this parameter to the `EndSessionRequest` object and updates the logout url generator to use parameter builder like we do for the /authorize request.

[More info on id_token_hint](https://docs.microsoft.com/en-us/azure/active-directory-b2c/openid-connect#secure-your-logout-redirect)

Closes #996


change/@azure-msal-browser-2020-11-10-16-13-10-idTokenHint.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "none",
```
jasonnutter, line 2: I think this can be a patch version since it is adding functionality (although doesn't really matter as there will be other patch changes, I suppose).

jasonnutter (approved):

change/@azure-msal-browser-2020-11-10-16-13-10-idTokenHint.json
tnorling: The browser changes are only doc updates all the code changes are in common

sameerag (approved):
LGTM. One comment: Do we restrict in the code that this is only for B2C? I did not think so, and hence in that case did we test this for AAD and see it works as expected?

tnorling:
> LGTM. One comment: Do we restrict in the code that this is only for B2C? I did not think so, and hence in that case did we test this for AAD and see it works as expected?

Thanks for the callout, this was tested and did not have any adverse affects in AAD scenarios

PR #2589, opened 2020-11-11
Add lib name & version to Logger

tnorling:
Currently all log messages output the version of `msal-common` being used i.e.
`[Wed, 11 Nov 2020 19:37:04 GMT] :  : 1.7.1 : Info - Emitting event: msal:handleRedirectStart`

This PR adds lib name and version to Logger constructor so that log messages properly reflect where they are coming from i.e.
`[Wed, 11 Nov 2020 19:50:54 GMT] :  : @azure/msal-browser@2.6.1 : Info - Emitting event: msal:handleRedirectStart`
`[Wed, 11 Nov 2020 19:54:27 GMT] :  : @azure/msal-common@1.7.1 : Info - Navigate url is empty`


lib/msal-common/src/logger/Logger.ts
```diff
@@ -51,12 +50,21 @@ export class Logger {
...
     private localCallback: ILoggerCallback;
 
-    constructor(loggerOptions: LoggerOptions) {
+    // Package name implementing this logger
+    private packageName: string;
+
+    // Package version implementing this logger
+    private packageVersion: string;
+
+    constructor(loggerOptions: LoggerOptions, packageName: string, packageVersion: string) {
```
jasonnutter, line 59: I think we need to `packageName` and `packageVersion` optional, to prevent breaking if an app picks up the new version of common but not the new version of browser/node.

tnorling: yep, good call. Changed.

hectormmg (approved):

lib/msal-common/src/logger/Logger.ts
```diff
@@ -51,12 +51,21 @@ export class Logger {
...
+
+    constructor(loggerOptions: LoggerOptions, packageName?: string, packageVersion?: string) {
         if (loggerOptions) {
             this.localCallback = loggerOptions.loggerCallback;
             this.piiLoggingEnabled = loggerOptions.piiLoggingEnabled;
             this.level = loggerOptions.logLevel;
         }
+
+        this.packageName = packageName || Constants.EMPTY_STRING;
+        this.packageVersion = packageVersion || Constants.EMPTY_STRING;
```
jasonnutter, line 68: Could they default to values from `pkg`? Or do we want to avoid using `pkg`?

jasonnutter: Ah that would grab common's data, yeah?

jasonnutter: Yep missed that in the description, nvm!

pkanher617 (approved):

sameerag (approved):

PR #2590, opened 2020-11-12
Scope beachball to lib folder

tnorling:
When we bump versions before publish beachball also bumps dependencies in the `package.json` files in the samples directory. This is undesirable as samples are no longer in scope for lerna and our E2E tests need to reference local builds. This PR scopes beachball  commands to only affect the `lib` directory.

Please callout other samples that should be updated to pull packages locally instead of from npm


hectormmg (approved):

pkanher617:
> Please callout other samples that should be updated to pull packages locally instead of from npm

I think this will require changes to all packages that are referencing MSAL.js packages from npm. So, every node sample, the browser Typescript sample, etc.

hectormmg:
> > Please callout other samples that should be updated to pull packages locally instead of from npm
> 
> I think this will require changes to all packages that are referencing MSAL.js packages from npm. So, every node sample, the browser Typescript sample, etc.

I see why samples used in end-to-end tests need to use the local versions, but I think the standalone node usage samples can use the latest published package.

sameerag:
My two cents: In Philosophy, yes, standalone samples can use the published version. However, since node e2e isn't robust yet, can we retain the ability to test locally for the standalone samples for now? cc @hectormmg

pkanher617:
> > > Please callout other samples that should be updated to pull packages locally instead of from npm
> > 
> > 
> > I think this will require changes to all packages that are referencing MSAL.js packages from npm. So, every node sample, the browser Typescript sample, etc.
> 
> I see why samples used in end-to-end tests need to use the local versions, but I think the standalone node usage samples can use the latest published package, so I think msal-node is covered here.

It does make local dev a little more challenging because we can't run samples based on local packages anymore (since samples are being removed from lerna). If the samples used in e2e tests can be run locally it should be fine.

tnorling:
> It does make local dev a little more challenging because we can't run samples based on local packages anymore (since samples are being removed from lerna). If the samples used in e2e tests can be run locally it should be fine.

To be clear the samples have already been removed from lerna, it was done a while back. This PR is just stopping beachball from overwriting the local file with the published version when we run beachball:bump (breaking e2e tests)

pkanher617 (approved):

sameerag (approved):

jasonnutter:
> > It does make local dev a little more challenging because we can't run samples based on local packages anymore (since samples are being removed from lerna). If the samples used in e2e tests can be run locally it should be fine.
> 
> To be clear the samples have already been removed from lerna, it was done a while back. This PR is just stopping beachball from overwriting the local file with the published version when we run beachball:bump (breaking e2e tests)

My understanding is that the samples were only not in scope for the `lerna run` command, but are still in scope for `lerna bootstrap`. Not sure if that changes anything about this PR.

jasonnutter (approved):

tnorling:
> My understanding is that the samples were only not in scope for the `lerna run` command, but are still in scope for `lerna bootstrap`. Not sure if that changes anything about this PR.

We should test this but I think it's out of scope for both. We removed it from scope to make build faster since running `npm install` from the root of the repo was installing sample dependencies (puppeteer takes over a full minute by itself)

PR #2596, opened 2020-11-13
Create Angular 11 sample for msal-angular-v2

jo-arroyo:
Creating sample using Angular 11 and typescript 4 for msal-angular-v2 library. To be merged into `msal-angular-v2` branch.


jasonnutter (approved):
Looks good! You can also update the ranges in the MSAL Angular peerDepencies block to include Angular 11.

PR #2599, opened 2020-11-13
Add script to create Github release based on beachball changelog

jasonnutter:
Adds a script to programmatically create Github releases based on the latest release for each library. Used to create the most recent releases: https://github.com/AzureAD/microsoft-authentication-library-for-js/releases


.github/create-releases.js
```diff
@@ -0,0 +1,114 @@
...
+    const release = {
+        ...repoMeta,
+        tag_name,
+        name: `${name} v${version}`,
+        prerelease: !!prerelease,
+        body: body.join("\n")
+    };
+
+    // Gets latest releases and checks if there is a matching releases for the given library version
+    // TODO: improve check to handle release not being on the first page
```
pkanher617, line 94: what does this mean?

.github/create-releases.js
```diff
@@ -0,0 +1,114 @@
...
+
+    const release = {
+        ...repoMeta,
+        tag_name,
+        name: `${name} v${version}`,
+        prerelease: !!prerelease,
+        body: body.join("\n")
+    };
+
+    // Gets latest releases and checks if there is a matching releases for the given library version
```
pkanher617, line 93: This is only checking releases that we have created entries for in Github right? Should we also add a check to see if the release is on npm?

.github/create-releases.js
jasonnutter: The releases API is paginated, and right now were only looking for an existing release on the first page of results.

.github/create-releases.js
jasonnutter: We could, although going forward there should be a 1:1 between the two.

tnorling (approved):

.github/create-releases.js
jasonnutter: I'll look at checking npm incrementally, as npm will throw an error if you try to republish the same version.

jmckennon (approved):

PR #2602, opened 2020-11-16
[msal-common] Enable Strict null checks 

sameerag:
Current Configuration:

```javascript
        "strict": true,
        "strictNullChecks": true,
        "strictFunctionTypes": true,
        "strictPropertyInitialization": false,
        "noImplicitThis": true,
        "noImplicitAny": true,
        "alwaysStrict": true,
        "noUnusedLocals": true,
        "noUnusedParameters": false,
        "noFallthroughCasesInSwitch": true,
```


tnorling (approved):

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -111,19 +117,20 @@ export function buildConfiguration({ auth: userInputAuth, cache: userInputCache,
...
     // Default logger options for browser
     const DEFAULT_LOGGER_OPTIONS: LoggerOptions = {
         loggerCallback: (): void => {},
+        logLevel: LogLevel.Info,
         piiLoggingEnabled: false
     };
 
     // Default system options for browser
-    const DEFAULT_BROWSER_SYSTEM_OPTIONS: BrowserSystemOptions = {
+    const DEFAULT_BROWSER_SYSTEM_OPTIONS: Required<Omit<BrowserSystemOptions, "loadFrameTimeout">> = {
```
jasonnutter, line 133: Why is `loadFrameTimeout` omitted?

tnorling: The only place it's used is to set the `windowHashTimeout` and `iframeHashTimeout` below, it's not needed anywhere else.

jasonnutter: Could the default by set to `0` so that it doesn't need to be omitted?

tnorling: Done

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -86,6 +86,12 @@ export type Configuration = {
     system?: BrowserSystemOptions
 };
 
+export type BrowserConfiguration = {
```
pkanher617, line 89: Why are we not updating the existing type?

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -415,11 +413,11 @@ export abstract class CacheManager implements ICacheManager {
                 return;
             }
 
-            if (!StringUtils.isEmpty(environment) && !this.matchEnvironment(entity, environment)) {
+            if (!environment || !this.matchEnvironment(entity, environment)) {
```
pkanher617, line 416: This will return if no environment is passed, which is not what we want. Should match the filters above `!!environment && !this.matchEnvironment(entity, environment)`

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -661,7 +659,7 @@ export abstract class CacheManager implements ICacheManager {
      * @param homeAccountId
      */
     private matchHomeAccountId(entity: AccountEntity | CredentialEntity, homeAccountId: string): boolean {
-        return entity.homeAccountId && homeAccountId === entity.homeAccountId;
+        return !!(entity.homeAccountId && homeAccountId === entity.homeAccountId);
```
pkanher617, line 662: Why are we adding a double boolean here?

lib/msal-common/src/cache/CacheManager.ts
tnorling: Ah thanks, I updated all the checks above, must have missed these down here

lib/msal-common/src/cache/CacheManager.ts
tnorling: So I'm not sure I understand why but the compiler thinks the return type of the expression is `string | boolean`. I can't think of a scenario where `entity.homeAccountId` would be returned here since its an &&... but in any case prepending `!!` forces type `boolean`

lib/msal-common/src/cache/CacheManager.ts
tnorling: Fixed.

hectormmg (approved):

jmckennon (approved):
LGTM! Big fan of these changes

jasonnutter:
> In what situations is this value `null`?

@jviney @josmithua I believe this is because in code, the value set for `account` is initially set to `null` and then dynamically filled in after processing the response.

josmithua:
So it might still be null if it fails to process the response? Or will an exception be thrown? If the account property will never be null when returning from the function then I think the interface should reflect that

PR #2610, opened 2020-11-17
Improvements to redirect response handling

tnorling:
This PR updates the `useMsalAuthentication` hook to react to the login success/failure events instead of promise resolve/catch. This is done so that redirect responses can be consumed by the hook as well as responses from the `login` callback this hook provides.

This PR also updates the returned `login` callback to default to the values provided to the hook instead of requiring these params be provided again


pkanher617 (approved):

samples/msal-react-samples/onPageLoad/package.json
```diff
@@ -3,8 +3,8 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@azure/msal-react": "1.0.0-alpha.0",
-    "@azure/msal-browser": "2.7.0",
+    "@azure/msal-react": "file:../../../lib/msal-react",
```
pkanher617, line 6: You shouldn't need to do this, lerna can take care of bootstrapping?

pkanher617: Also beachball bumps the dependencies

tnorling: We removed samples from lerna to prevent things like puppeteer from being installed on every build

pkanher617: Sorry I think I reviewed the PRs in the wrong order.

lib/msal-react/src/hooks/useMsalAuthentication.ts
```diff
@@ -21,46 +21,56 @@ export function useMsalAuthentication(
...
             // Ensure login is only called one time from within this hook, any subsequent login attempts should use the callback returned
             setHasBeenCalled(true);
-            login(interactionType, authenticationRequest)
-                .then(result => {
-                    isMounted.current && setResponse([result, null]);
-                })
-                .catch(error => {
-                    isMounted.current && setResponse([null, error]);
-                });
+            login().catch(() => {
```
sameerag, line 68: Why do we still need the catch?

tnorling: Otherwise errors will throw and potentially break application flow since we are surfacing the errors in a different way

lib/msal-react/src/hooks/useMsalAuthentication.ts
```diff
@@ -21,46 +21,56 @@ export function useMsalAuthentication(
...
                 throw "Invalid interaction type provided.";
         }
+    }, [instance, interactionType, authenticationRequest]);
+
+    useEffect(() => {
+        const callbackId = instance.addEventCallback((message: EventMessage) => {
+            switch(message.eventType) {
+                case EventType.LOGIN_SUCCESS:
+                case EventType.SSO_SILENT_SUCCESS:
+                    message.payload && setResponse([message.payload as AuthenticationResult, null]);
```
jasonnutter, line 50: Nitpick (readability):
```suggestion
                    if (message.payload) {
                        setResponse([message.payload as AuthenticationResult, null]);
                    }
```

lib/msal-react/src/hooks/useMsalAuthentication.ts
```diff
@@ -21,46 +21,56 @@ export function useMsalAuthentication(
...
+                    break;
+                case EventType.LOGIN_FAILURE:
+                case EventType.SSO_SILENT_FAILURE:
+                    message.error && setResponse([null, message.error as AuthError]);
+                    break;
+            }
+        });
+
+        return () => {
+            callbackId && instance.removeEventCallback(callbackId);
```
jasonnutter, line 60: Same here

jasonnutter (commented):
Any additional tests needed?

lib/msal-react/src/hooks/useMsalAuthentication.ts
tnorling: Done

lib/msal-react/src/hooks/useMsalAuthentication.ts
tnorling: Done

jasonnutter (approved):

PR #2620, opened 2020-11-18
[msal-angular] Add log messages

jo-arroyo:
This PR:
- Adds `clone` function and unit tests to `msal-common` Logger (Note: this function is also duplicated on the `dev` branch)
- Adds log messages to msal-angular
- Updates logging documentation
- Adds logging to the angular 10 sample configuration


lib/msal-angular/src/msal.guard.ts
```diff
@@ -48,13 +50,18 @@ export class MsalGuard implements CanActivate {
 
     private loginInteractively(url: string): Observable<boolean> {
         if (this.msalGuardConfig.interactionType === InteractionType.Popup) {
+            this.authService.logger.verbose("Guard - logging in by popup");
```
jo-arroyo, line 53: Unsure if we want to specify in the logs how it's logging in (popup or redirect) or leave it to be more generic, since the logs from msal-browser are related to events, which just state that login has started, without specifying interaction type

lib/msal-angular/src/msal.service.ts
```diff
@@ -29,9 +32,10 @@ export class MsalService implements IMsalService {
         if (hash) {
             this.redirectHash = `#${hash}`;
         }
+        this.logger = this.instance.getLogger();
```
tnorling, line 35: I think instead of getting the existing msal-browser logger we should create a new logger with the same options, that way the log messages coming from `msal-angular` and coming from `msal-browser` are properly distinguished with relevant versions, etc.

lib/msal-angular/src/msal.guard.ts
jasonnutter: I think indicating the interaction type is a good idea.

lib/msal-angular/src/msal.service.ts
jo-arroyo: Done!

lib/msal-common/src/logger/Logger.ts
```diff
@@ -68,6 +68,13 @@ export class Logger {
         this.packageVersion = packageVersion || Constants.EMPTY_STRING;
     }
 
+    /**
+     * Create new Logger with existing configurations.
+     */
+    public clone(packageName: string, packageVersion: string): Logger {
+        return new Logger({loggerCallback: this.localCallback, piiLoggingEnabled: this.piiLoggingEnabled, logLevel: this.level}, packageName, packageVersion);
```
jasonnutter, line 75: Can we add tests for this?

jo-arroyo: Done.

jasonnutter (approved):

Issue #2606, opened 2020-11-17
acquireTokenSilent does not return the token field


jakubgrzywaczewski:
## Library
- [x] `msal@1.4.4`

## Framework
React 16.11.3 

## Description

So far I have used msal.js version 1.3.2 for authorization via b2c. After upgrading to version 1.4.4, I noticed a problem with the response returned by the **acquireTokenSilent** method - the **accessToken** field is missing. The token itself appears in the **idToken**.**rawIdToken** field.

## Regression
- [x] Did this behavior work before? 
Version: 1.3.2

## MSAL Configuration

```js
const msalConfig = {
  auth: {
    clientId: Settings.default.msalClientId,
    authority: Settings.default.msalAuthority,
    redirectUri: `${document.location.origin}/signin/b2c`,
    validateAuthority: false,
  },
  cache: {
    cacheLocation,
    storeAuthStateInCookie: false,
  },
};

const msalAuthenticationParameters = {
  scopes: [Settings.default.msalClientId],
};

const msalInstance= new UserAgentApplication(msalConfig);

const msalAuthenticationParameters = {
  scopes: [Settings.default.msalClientId],
};

```

## Reproduction steps
```js
export function fetchWithMsal(
  url: string,
  request: RequestInit,
): Promise<Response> {
  let resultRequest = request;

  const result = msalInstance
    .acquireTokenSilent(msalAuthenticationParameters)
    .then((response) => {
      if (!response.accessToken) {
        return undefined;
      }
      resultRequest = generateRequestProperties({
        accessToken: response.idToken.rawIdToken,
        request,
      });

      return fetch(url, resultRequest);
    })
    .catch((e) => {
      msalInstance.acquireTokenRedirect(msalAuthenticationParameters);
    });

  return result as Promise<Response>;
}
```

## Expected behavior
Return accessToken field with token.

pkanher617:
@jakubgrzywaczewski If you are sending `msalClientId` as the scope, you are requesting the scopes for openid and profile, which are idToken scopes. Please check the network trace to make sure, but this should only give you an id_token from the STS. 

Previously, the library returned this token as an access token as well. We have since changed that behavior to be in-line with what the server is sending back.

If you are seeing an access token in the network trace and it is not being returned, we can investigate this further.

jakubgrzywaczewski:
[...]

PR #2631, opened 2020-11-21
[Docs] update FAQs for B2C

derisen:
This PR adds FAQ items to **msal-core** and **msal-browser** for:

1) getting B2C support
2) queries on using `clientId` as scope in `acquireToken*` and the expected behavior (ref [issue](https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/2606))


lib/msal-browser/FAQ.md
```diff
@@ -43,6 +43,8 @@
 1. [I logged out of my application. Why am I not asked for credentials when I try to log back in?](#i-logged-out-of-my-application-why-am-i-not-asked-for-credentials-when-i-try-to-log-back-in)
 1. [Why am I not signed in when returning from an invite link?](#why-am-i-not-signed-in-when-returning-from-an-invite-link)
 1. [Why is there no access token returned from acquireTokenSilent?](#why-is-there-no-access-token-returned-from-acquiretokensilent)
+1. [I used to be able to get an access token when using acquireTokenSilent with clientId as scope, now I can't. Why?](#i-used-to-be-able-to-get-an-access-token-when-using-acquiretokensilent-with-clientId-as-scope-now-i-can't-why)
```
tnorling, line 46: I think this is only relevant for msal-core

lib/msal-core/docs/FAQ.md
```diff
@@ -379,6 +381,17 @@ You can read more about this behavior [here](https://docs.microsoft.com/azure/ac
 
 MSAL.js will only process tokens which it originally requested. If your flow requires that you send a user a link they can use to sign up, you will need to ensure that the link points to your app, not the B2C service directly. An example flow can be seen in the [working with B2C](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/working-with-b2c.md) doc.
 
+## I used to be able to get an access token when using acquireTokenSilent with clientId as scope, now I can't. Why?
+
+If you are sending `clientId` as the scope, you are requesting scopes for `openid` and `profile`, which are **ID token** scopes.
```
tnorling, line 386: I would suggest making this more general than just B2C specific. i.e. `Why is the idToken or accessToken field in the response empty?` and and move this into either the `Tokens` section or the `Scopes and Resources` section.

lib/msal-browser/FAQ.md
derisen: Aah, I'll double check with the 2.x sample

lib/msal-core/docs/FAQ.md
derisen: I had the impression that this is a quirk with the B2C server. Will confirm with AAD sample and revise accordingly 
cc @hectormmg

lib/msal-browser/FAQ.md
tnorling: What I mean is that `msal-browser` has always behaved this way so I'm not sure we need to callout a change. The doc linked in the answer was also written specifically for `msal-core` so I would prefer to leave this out of the `msal-browser` FAQ until we have a chance to write a similar doc for `msal-browser`

derisen: got it 👍

lib/msal-core/docs/FAQ.md
derisen: done!

lib/msal-core/docs/FAQ.md
```diff
@@ -285,6 +287,12 @@ Please refer to our performance guide [here](https://github.com/AzureAD/microsof
 
 Token lifetimes are 1 hour and the session lifetime is 24 hours. This means that if no requests have been made in 24 hours, you will need to login again before requesting a new token.
 
+## Why is the accessToken field in the response empty?
+
+If you are sending `clientId` as the scope, you are requesting scopes for `openid` and `profile`, which are **ID token** scopes. Previously, MSAL.js would get an **ID token** and put it into the `accessToken` and `idToken` fields (both with the same value). We have since changed that behavior to be in-line with what the AAD/B2C server is sending back. Now MSAL.js gets the same **ID token** but puts it under `idToken` field and leaves the `accessToken` field empty.
```
tnorling, line 292:
```suggestion
If you are requesting scopes `clientId`, `openid` and/or `profile`, these are **ID token** scopes. Previously, in versions <1.4.0, MSAL.js would get an **ID token** and put it into the `accessToken` and `idToken` fields (both with the same value). In version 1.4.0 this behavior was changed to be in-line with what the AAD/B2C server is sending back. Now MSAL.js gets the same **ID token** but puts it under `idToken` field and leaves the `accessToken` field empty. If you need an access token you should request a scope other than `clientId`, `openid` or `profile`
```

lib/msal-core/docs/FAQ.md
```diff
@@ -379,6 +387,17 @@ You can read more about this behavior [here](https://docs.microsoft.com/azure/ac
 
 MSAL.js will only process tokens which it originally requested. If your flow requires that you send a user a link they can use to sign up, you will need to ensure that the link points to your app, not the B2C service directly. An example flow can be seen in the [working with B2C](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/working-with-b2c.md) doc.
 
+## I used to be able to get an access token when using acquireTokenSilent with clientId as scope, now I can't. Why?
```
tnorling, line 390: This can now be removed since the section above covers it. Or just link to the above if you feel like keeping it.

derisen: my bad. done!

tnorling (approved):

sameerag (approved):
We do not yet have a FAQ for `msal-node`. @hectormmg can you please address this and follow up with @derisen to add this B2C FAQ in the same?

Not blocking approval though.

jmckennon (approved):

Issue #2494, opened 2020-10-27
Allow timeout configurability with Device code 


jonathandturner:
Please follow the issue template below. Failure to do so will result in a delay in answering your question.

## Library
 - msal-node

## Description
Something that we'd like to be able to do is set a timeout for the device code credential so that once the timeout has been reached, we return control and can give an error to the user. It seems that some MSAL languages, like Python (https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/identity/azure-identity/azure/identity/_credentials/device_code.py#L85-L88), are able to use these timeouts, and it'd be great to be able to set them from node also.

sameerag:
Thanks @jonathandturner for raising this. I think this is possible. Will update this space once the specifics are clear. Tracking this PR with node project.

PR #2656, opened 2020-11-25
Adding request timeout for device code flow

samuelkubai:
Adding timeout value to the device code request

Issue: #2494


lib/msal-common/test/client/DeviceCodeClient.spec.ts
```diff
@@ -174,5 +174,22 @@ describe("DeviceCodeClient unit tests", async () => {
...
+
+            let deviceCodeResponse = null;
+            const request: DeviceCodeRequest = {
+                scopes: [...TEST_CONFIG.DEFAULT_GRAPH_SCOPE, ...TEST_CONFIG.DEFAULT_SCOPES],
+                deviceCodeCallback: (response) => deviceCodeResponse = response,
+                timeout: DEVICE_CODE_RESPONSE.interval - 2, // Setting a timeout less the interval polling time to ensure the timeout has expired before the second poll
+            };
+
+            const client = new DeviceCodeClient(config);
+            await expect(client.acquireToken(request)).to.be.rejectedWith(`${ClientAuthErrorMessage.DeviceCodeExpired.desc}`);
```
pkanher617, line 192: Might be better to reject with a different error (specific to the user request timeout being hit, instead of the default `expires_in` value).

samuelkubai: I agree a different error would communicate much clearer to our users.

change/@azure-msal-node-2020-11-25-19-29-14-device-code-timeout.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "prerelease",
```
tnorling, line 2: Looks like msal-node changes are only package-lock? This should be type none in that case

change/@azure-msal-common-2020-11-25-19-29-14-device-code-timeout.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "minor",
```
tnorling, line 2:
```suggestion
  "type": "patch",
```

samuelkubai: @tnorling should the addition of a feature be classified as a `patch`?

change/@azure-msal-node-2020-11-25-19-29-14-device-code-timeout.json
samuelkubai: Change this to a `none`.

lib/msal-common/test/client/DeviceCodeClient.spec.ts
samuelkubai: @pkanher617 please re-review, added a specific error for this timeout

change/@azure-msal-common-2020-11-25-19-29-14-device-code-timeout.json
sameerag: Since this isn't technically a full feature but an additional functionality added to polling, I would prefer a `patch` too. I do see that this case is a little gray area though.

lib/msal-common/src/error/ClientAuthError.ts
```diff
@@ -142,6 +142,10 @@ export const ClientAuthErrorMessage = {
     tokenRefreshRequired: {
         code: "token_refresh_required",
         desc: "Cannot return token from cache because it must be refreshed. This may be due to one of the following reasons: forceRefresh parameter is set to true, claims have been requested, there is no cached access token or it is expired."
+    },
+    userTimeoutReached: {
+        code: "user_timeout_reached",
```
sameerag, line 147: should we mention for device code polling to clarify use case?

change/@azure-msal-common-2020-11-25-19-29-14-device-code-timeout.json
samuelkubai: Cool, that's fine with me too, I don't have too strong of an opinion on either for this use case.

lib/msal-common/src/client/DeviceCodeClient.ts
```diff
@@ -143,6 +143,7 @@ export class DeviceCodeClient extends BaseClient {
         const requestBody = this.createTokenRequestBody(request, deviceCodeResponse);
         const headers: Record<string, string> = this.createDefaultTokenRequestHeaders();
 
+        const userSpecifiedTimeout = request.timeout ? TimeUtils.nowSeconds() + request.timeout : undefined;
```
sameerag, line 146: Shouldn't we validate the user set timeout? `request.timeout <= deviceCodeExpirationTime` should always hold. Can we throw a configuration error if this isn't respected?

samples/msal-node-samples/standalone-samples/device-code/package.json
```diff
@@ -12,6 +12,7 @@
   "author": "",
   "license": "MIT",
   "dependencies": {
-    "@azure/msal-node": "file:../../../../lib/msal-node"
+    "@azure/msal-node": "file:../../../../lib/msal-node",
+    "@azure/msal-common": "file:../../../../lib/msal-common"
```
tnorling, line 16: This line is not needed as msal-common will be included with msal-node

lib/msal-common/src/error/ClientAuthError.ts
samuelkubai: I agree that makes the error message more clear, I have added it @sameerag

samples/msal-node-samples/standalone-samples/device-code/package.json
samuelkubai: You are right, I added this as changes to the local `msal-common` were not available to the samples.

lib/msal-common/src/client/DeviceCodeClient.ts
samuelkubai: Since the `deviceCodeExpirationTime` is provided at runtime, won't throwing a configuration error be too rough of an experience in the case Azure changes their `deviceCodeExpirationTime`?

I will however still add that validation to ensure the user timeout is only respected if it's shorter than the device code expiration time

pkanher617: I agree throwing an error might be too extreme here: would probably be better to just use the shortest time.

tnorling (approved):
Looks good! Congrats on your first PR!

lib/msal-common/src/client/DeviceCodeClient.ts
hectormmg: I also agree. In the case where we go to the default we should do a verbose log explaining why the user provided timeout was ignored. Also I think this is the kind of configuration option that may warrant a note on the Device Code flow section of the Request doc in msal-common.

samuelkubai: Two great points @hectormmg, went ahead and updated the docs as well as added a verbose log for when the user timeout is ignored.

sameerag: Thanks. Looks good for me.

lib/msal-common/src/error/ClientAuthError.ts
sameerag: Great!

sameerag (approved):

hectormmg (approved):
Looks great to me!

jasonnutter (approved):

PR #2669, opened 2020-11-30
Add onRedirectNavigate to redirect operations in browser

### LINKED ISSUE #2127: allow "logout" of a specific application but not end the entire AAD session, which logs you out of all apps

### LINKED ISSUE #2664: Does MSAL support authentication within Chrome extensions?
## Library

### tl;dr
Does MSAL support authentication within Chrome extensions using the ```chrome.identity.launchWebAuthFlow``` api? Link: https://developer.chrome.com/apps/identity#method-launchWebAuthFlow If so, how so? If not, would you be open to a contribution/PR to do so?

### Long version
I'm writing a plain, vanilla, JS Chrome extension and I'd like to add MSA authentication to it. I started experimenting with MSAL to do this but ran into a few troubles. The first one was while configuring my AAD App registration: The Single-page Application Platform Type only supports redirect uris with https (or http) schemas. Chrome extensions require a chrome-extension:// redirect uri schema. Fortunately, I found that I could add a Mobile and Desktop Platform Type to the App Registration and supply the needed style/schema of redirect url there. Unfortunately, even with this work-around I ran into trouble. This time during the auth code/token exchange I believe. MSAL began throwing an error that read something like 'cross-origin token exchanges are only allowed in Single-page Application client types'. This issue lead me to investigate, and discover, how Google supports external authentication within Chrome extensions and that's through the ```chrome.identity.launchWebAuthFlow``` api mentioned above.

The launchWebAuthFlow api takes the first url used in the authentication flow along with a callback function that's called when a redirect from the auth flow is received. Upon execution the api call causes Chrome to render a new popup using the login url that you provided. Chrome then supports redirects for your extension through https://<yourextensionid>.chromiumapp.com/*. When a redirect is received at that url it calls your callback function and provides the full url of the redirect as a parameter (which you can then parse and proceed with the authentication).

I put together a crude proof-of-concept that suggests that this approach/flow would work with MSAL but the approach doesn't seem to be supported within MSAL currently. I think you'd need two new functions on IPublicClientApplication to support this. One that works similar to loginRedirect but just returns the url to kick off the authentication flow instead of trying to redirect the current page and another that works similarly to handleRedirectPromise but takes the hash of the redirect url received in the launchWebAuthFlow callback method. This brings me to my question above: Is Chrome extension authentication supported in MSAL through the ```chrome.identity.launchWebAuthFlow``` api (or some other way that I'm not seeing)? If so, how so? If not, would you be open to a contribution/PR to do so?

jasonnutter:
Adds `onRedirectNavigate` callback from v1 (#1691).

This adds a new callback for `loginRedirect`, `acquireTokenRedirect`, and `logout` to allow applications to get the URL that would be navigated to, and to stop navigation. This is to support scenarios where applications want to perform the redirect themselves, instead of relying on MSAL.

Usage:

```js
msal.loginRedirect({
  onRedirectNavigate: url => {
    // url that would be navigated to
    console.log(url);
   
     // Applications must explicitly return false to stop navigation
     return false;
  }
})
```

Also includes a sample Chromium extension demonstrating the usage.

TODO:

- [x] Test with Chrome extension


change/@azure-msal-browser-2020-11-30-15-01-27-browser-onredirectnavigate.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Add onRedirectNavigate to redirect operations in browser",
+  "packageName": "@azure/msal-browser",
+  "email": "janutter@microsoft.com",
+  "dependentChangeType": "patch",
```
jasonnutter, line 6:
```suggestion
  "comment": "Add onRedirectNavigate to redirect operations in browser (#2669)",
  "packageName": "@azure/msal-browser",
  "email": "janutter@microsoft.com",
  "dependentChangeType": "none",
```

tnorling (commented):
Could you update the request docs with this new param + maybe a usage example?

jasonnutter:
> Could you update the request docs with this new param + maybe a usage example?

I will definitely add a usage example. Do we still need the standalone docs for request/response? Would be nice to have those docs fully automated (typedoc, in this case).

tnorling:
> > Could you update the request docs with this new param + maybe a usage example?
> 
> I will definitely add a usage example. Do we still need the standalone docs for request/response? Would be nice to have those docs fully automated (typedoc, in this case).

I think the request docs are out of date anyway. I'm all for automating it if they're still easy to find (I don't know where the typedocs live today)

jasonnutter:
> > > Could you update the request docs with this new param + maybe a usage example?
> > 
> > 
> > I will definitely add a usage example. Do we still need the standalone docs for request/response? Would be nice to have those docs fully automated (typedoc, in this case).
> 
> I think the request docs are out of date anyway. I'm all for automating it if they're still easy to find (I don't know where the typedocs live today)

Yeah, the request/response docs are very much out of date, and so I would propose deleting them, since the typedocs will always be update to date in theory. cc: @pkanher617 

https://azuread.github.io/microsoft-authentication-library-for-js/ref/msal-browser/

lib/msal-browser/test/app/PublicClientApplication.spec.ts
```diff
@@ -1875,6 +1907,26 @@ describe("PublicClientApplication.ts Class Unit Tests", () => {
...
+            const logoutUriSpy = sinon.stub(AuthorizationCodeClient.prototype, "getLogoutUri").returns(testLogoutUrl);
+            sinon.stub(BrowserUtils, "navigateWindow").callsFake((urlNavigate: string, timeout: number, logger: Logger, noHistory: boolean) => {
+                // If onRedirectNavigate does not stop navigatation, this will be called, failing the test as done will be invoked twice
+                return Promise.resolve(done());
+            });
+            pca.logout({
+                onRedirectNavigate: (url) => {
+                    expect(url).to.be.eq(testLogoutUrl);
+                    done();
+                    return false;
```
jasonnutter, line 1920: FYI @pkanher617 this can be how apps stop the redirect from happening during logout

tnorling:
> Yeah, the request/response docs are very much out of date, and so I would propose deleting them, since the typedocs will always be update to date in theory. cc: @pkanher617
> 
> Or, just have the existing markdown files just link to the typedocs.
> 
> https://azuread.github.io/microsoft-authentication-library-for-js/ref/msal-browser/

Let's delete the docs and update the links to those docs to point to the typedocs?

jasonnutter:
> > Yeah, the request/response docs are very much out of date, and so I would propose deleting them, since the typedocs will always be update to date in theory. cc: @pkanher617
> > Or, just have the existing markdown files just link to the typedocs.
> > https://azuread.github.io/microsoft-authentication-library-for-js/ref/msal-browser/
> 
> Let's delete the docs and update the links to those docs to point to the typedocs?

Done.

tnorling (approved):

jo-arroyo (approved):

pkanher617 (approved):

bryik:
This PR closed issue #2127, but I do not see how it solves the problem described by that issue.

I guess `msalInstance.logout()` could be provided with an `onRedirectNavigate()` callback that clears local storage and prevents the logout redirect from occurring e.g.

```js
msalInstance.logout({
  onRedirectNavigate: () => {
    window.localStorage.clear();
    return false;
  }
});
```

Is that the recommended way to logout of a specific application without logging out of all apps tied to an AAD session?

imo it would be great to have this mentioned in the [logout doc](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/logout.md). I was surprised that calling `logout()` in a little test app signed me out of Outlook.

jasonnutter:
> Is that the recommended way to logout of a specific application without logging out of all apps tied to an AAD session?

Yes, using `onRedirectNavigate` and returning `false` is how you can stop the logout navigation. Note, you do not need to manually clear localStorage or sessionStorage (unless there are other non-MSAL.js things you want to remove), as MSAL.js will still clear MSAL.js entries from browser storage.

> imo it would be great to have this mentioned in the logout doc. I was surprised that calling logout() in a little test app signed me out of Outlook.

Yes, I have it on my list to update that document as apart of #2546.

bryik:
Perfect, thank you @jasonnutter!

kireet3213:
Hey there. I set onRedirectNavigate to false with updated msal-browser and msal-react packages but I fail to authenticate the user with  msalInstance.handleRedirectPromise(redirectUrl). It says that there is no redirection in progress, returning null.  Is there some other way to get it authenticated in this library?

PR #2670, opened 2020-11-30
[msal-common] Add clone to Logger

jo-arroyo:
This PR adds a `clone` function to the `Logger.ts`. This function will be used by wrapper libraries (such as `msal-angular`) to add a new Logger to log wrapper library-specific logs, while using the same logger configurations set up for `msal-browser`. This PR also adds unit tests. 

Note: This function has also been duplicated on the `msal-angular-v2` branch.


tnorling (approved):
Thanks for doing this!

jasonnutter (approved):

PR #2671, opened 2020-12-01
Prevent unnecessary context updates and re-renders

tnorling:
Arrays of objects are not referentially stable, meaning every time we update the `accounts` in context, React thinks state has changed and will re-render the component tree, even if no accounts have been added or removed from the array. This also means anything that takes a dependency on `accounts` (such as a `useEffect` hook) will unnecessarily run again.

This PR addresses this by adding a comparison check between the value of `getAllAccounts()` to the current accounts in context and only updating state if an account has been added or removed.

Also fixes:
- An entry in the getting started docs which was affected by this. 
- `useAccount` considers all `accountIdentifiers` when looking for a match


lib/msal-react/src/utils/utilities.ts
```diff
@@ -22,3 +22,19 @@ export function getChildrenOrFunction<T>(
...
+ * @param arrayA 
+ * @param arrayB 
+ */
+export function arraysAreEqual(arrayA: Array<object>, arrayB: Array<object>) {
+    if (arrayA.length !== arrayB.length) {
+        return false;
+    }
+
+    return arrayA.every((element, index) => {
+        return JSON.stringify(element) === JSON.stringify(arrayB[index]);
```
jasonnutter, line 38: This is potentially expensive, could the comparison just be on a single property (e.g. localAccountId)?

lib/msal-react/src/utils/utilities.ts
```diff
@@ -22,3 +22,19 @@ export function getChildrenOrFunction<T>(
...
+    });
```
jasonnutter, line 39:
```suggestion
    return arrayA.every((element, index) => JSON.stringify(element) === JSON.stringify(arrayB[index]));
```

lib/msal-react/src/utils/utilities.ts
tnorling: Yeah I considered that, do we think the other properties on `AccountInfo` are stable? We know `localAccountId` and `homeAccountId` should never change, but is it possible someone could change their `username` for example? That could potentially leave us in a bad state if we're only updating on the presence or absence of a single property.

jasonnutter: > We know localAccountId and homeAccountId should never change

I think a check for one or both of these properties should be fine.

jasonnutter: v1 already solved this actually: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-core/src/Account.ts#L81

tnorling: @jasonnutter Updated. Thoughts?

lib/msal-react/src/utils/utilities.ts
```diff
@@ -22,3 +24,26 @@ export function getChildrenOrFunction<T>(
...
 type SetComplement<A, A1 extends A> = SetDifference<A, A1>;
 export type Subtract<T extends T1, T1 extends object> = Pick<T,SetComplement<keyof T, keyof T1>>;
+
+/**
+ * Helper function to determine whether 2 arrays are equal
+ * Used to avoid unnecessary state updates
+ * @param arrayA 
+ * @param arrayB 
+ */
+export function accountArraysAreEqual(arrayA: Array<AccountInfo>, arrayB: Array<AccountInfo>): boolean {
```
jasonnutter, line 34: Can you add tests for this?

lib/msal-react/src/utils/utilities.ts
jasonnutter: I think it would be good to have tests, but like it otherwise!

lib/msal-react/src/utils/utilities.ts
tnorling: Done. Still want to figure out how to properly test that the component is rendering and rerendering when I expect it to, but I'll tackle that at some point before GA

jasonnutter: Sounds good!

jasonnutter (approved):

lib/msal-react/src/utils/utilities.ts
tnorling: Will have to add `idTokenClaims` to the comparison once your PR is merged. That's an example of a potentially unstable value we would want to update state on.

jo-arroyo (approved):
Looks good!

PR #2593, opened 2020-11-12
Add allowRedirectInIframe flag to browser

jasonnutter:
By default, MSAL will block redirect operations if they are done inside an iframe, as we know the server will not support rendering the login UX in an iframe. However, there is at least one notable expection, the B2C embedded login UX feature (where the B2C login UI is allowed to be rendered inside an iframe), where invoking loginRedirect inside an iframe needs to be supported. This flag enables applications to opt out of that check if they know they need this functionality.

PR #2677, opened 2020-12-01
[msal-angular] Add redirectUri check to guard

jo-arroyo:
This PR updates the guard to check if redirectUri is set to a page with MSAL guard and logs a warning. 
Also updates exports from msal-browser.

Changes being made to msal-browser are duplicated on the `dev` branch in #2680

### DIFF
--- a/lib/msal-angular/src/msal.guard.ts
+++ b/lib/msal-angular/src/msal.guard.ts
@@ -79,6 +79,7 @@ export class MsalGuard implements CanActivate {
         /*
          * If a page with MSAL Guard is set as the redirect for acquireTokenSilent,
          * short-circuit to prevent redirecting or popups.
+         * TODO: Update to allow running in iframe once allowRedirectInIframe is implemented
          */
         if (UrlString.hashContainsKnownProperties(window.location.hash) && BrowserUtils.isInIframe()) {
             this.authService.getLogger().warning("Guard - redirectUri set to page with MSAL Guard. It is recommended to not set redirectUri to a page that requires authentication.");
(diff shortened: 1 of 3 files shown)


lib/msal-angular/src/msal.guard.ts
```diff
@@ -76,6 +76,15 @@ export class MsalGuard implements CanActivate {
         }
         this.authService.getLogger().verbose("MSAL Guard activated");
 
+        /*
+         * If a page with MSAL Guard is set as the redirect for acquireTokenSilent,
+         * short-circuit to prevent redirecting or popups.
+         */
+        if (UrlString.hashContainsKnownProperties(window.location.hash) && BrowserUtils.isInIframe()) {
```
jasonnutter, line 83: Can you add a TODO here, as this will need to be updated once #2593 is merged. (as we'll need to allow the guard to run in iframes if `allowRedirectInIframe` is set to `true`).

jasonnutter (approved):
One TODO that we'll need to address later, but looks good otherwise!

lib/msal-angular/src/msal.guard.ts
jo-arroyo: Makes sense. Added.

lib/msal-browser/src/index.ts
```diff
@@ -42,5 +43,7 @@ export {
     Logger,
     LogLevel,
     // Protocol Mode
-    ProtocolMode
+    ProtocolMode,
+    // Utils
+    UrlString
```
tnorling, line 48: This change will need to be made on the dev branch before putting out the next release of angular. Is there a 2nd PR for this?

tnorling (approved):
One general comment/question about keeping branches in sync, but Looks Good.

lib/msal-browser/src/index.ts
jo-arroyo: Yes, good point! I put a PR up now, #2680

PR #2678, opened 2020-12-01
Add matchPattern string util for wildcard matching for urls

jasonnutter:
For broker and Angular, we need logic for wildcard matching for urls (for the interceptor and for `allowedBrokerDomains`). This adds a simple helper function to do pattern matching with support for wildcards, without the need for a third-party library.


jo-arroyo (approved):
Looks good!

tnorling (approved):

change/@azure-msal-common-2020-12-01-15-07-04-matchPattern-common.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Add matchPattern string util for wildcard matching for urls",
```
tnorling, line 3:
```suggestion
  "comment": "Add matchPattern string util for wildcard matching for urls (#2678)",
```

PR #2680, opened 2020-12-02
[msal-browser] Add exports to index.ts

jo-arroyo:
This PR matches the changes being done in #2677, adding additional exports to msal-browser `index.ts`, to be used in the msal-angular-v2 guard.


change/@azure-msal-browser-2020-12-02-09-16-10-browser-new-exports.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "none",
```
jasonnutter, line 2:
```suggestion
  "type": "patch",
```

tnorling (approved):

jasonnutter (approved):

PR #2683, opened 2020-12-03
[msal-angular] replace minimatch

jo-arroyo:
This PR:
- Removes `minimatch` and `path` from msal-angular dependencies
- Replaces minimatch usage with `StringUtils.matchPattern` from msal-browser
- Updates documentation on wildcard matching
- Improves interceptor tests related to wildcard matching
- IE11 compatibility added to angular11-sample-app

Changes made to msal-browser are mirrored on #2693.

### DIFF
--- a/lib/msal-angular/docs/v2-docs/initialization.md
+++ b/lib/msal-angular/docs/v2-docs/initialization.md
@@ -89,7 +89,7 @@ export class AppModule {}
 
 Using MsalInterceptor is optional and you can write your own interceptor if you choose to. Alternatively, you can also explicitly acquire tokens using the acquireToken APIs.
 
-As of `@azure/msal-angular@2`, `protectedResourceMap` supports wildcard patterns that are supported by [minimatch](https://github.com/isaacs/minimatch). `unprotectedResources` is deprecated and no longer an option for configuration. Instead, setting a scope value of `null` on a resource will prevent it from getting tokens.
+As of `@azure/msal-angular@2`, `protectedResourceMap` supports using `*` for wildcards. `unprotectedResources` is deprecated and no longer an option for configuration. Instead, setting a scope value of `null` on a resource will prevent it from getting tokens.
 
 **Note:** When using wildcards, if multiple matching entries are found in the `protectedResourceMap`, the first match found will be used (based on the order of the `protectedResourceMap`).
 
--- a/lib/msal-angular/package.json
+++ b/lib/msal-angular/package.json
@@ -19,10 +19,7 @@
   },
   "main": "./dist/bundles/azure-msal-angular.umd.js",
   "typings": "./dist/azure-msal-angular.d.ts",
-  "dependencies": {
-    "minimatch": "^3.0.4",
-    "path": "^0.12.7"
-  },
+  "dependencies": {},
   "devDependencies": {
     "@angular-devkit/build-angular": "~0.1000.8",
     "@angular-devkit/build-ng-packagr": "~0.1000.8",
--- a/lib/msal-angular/package.json
+++ b/lib/msal-angular/package.json
@@ -19,7 +19,6 @@
   },
   "main": "./dist/bundles/azure-msal-angular.umd.js",
   "typings": "./dist/azure-msal-angular.d.ts",
-  "dependencies": {},
   "devDependencies": {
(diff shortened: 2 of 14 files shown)


lib/msal-angular/docs/v2-docs/initialization.md
```diff
@@ -89,7 +89,7 @@ export class AppModule {}
 
 Using MsalInterceptor is optional and you can write your own interceptor if you choose to. Alternatively, you can also explicitly acquire tokens using the acquireToken APIs.
 
-As of `@azure/msal-angular@2`, `protectedResourceMap` supports wildcard patterns that are supported by [minimatch](https://github.com/isaacs/minimatch). `unprotectedResources` is deprecated and no longer an option for configuration. Instead, setting a scope value of `null` on a resource will prevent it from getting tokens.
+As of `@azure/msal-angular@2`, `protectedResourceMap` supports using `*` for wildcards. `unprotectedResources` is deprecated and no longer an option for configuration. Instead, setting a scope value of `null` on a resource will prevent it from getting tokens.
```
jasonnutter, line 92: It would be good to call this out in the upgrade guide also (in addition to the fact we no longer support just putting your root domain in the protected resource map, must use wildcard matching)

jo-arroyo: Done.

lib/msal-angular/package.json
```diff
@@ -19,10 +19,7 @@
   },
   "main": "./dist/bundles/azure-msal-angular.umd.js",
   "typings": "./dist/azure-msal-angular.d.ts",
-  "dependencies": {
-    "minimatch": "^3.0.4",
-    "path": "^0.12.7"
-  },
+  "dependencies": {},
```
hectormmg, line 22: Nit: Maybe remove the dependencies key?

hectormmg (approved):
LGTM!

lib/msal-angular/package.json
jo-arroyo: Good point! Thank you!

lib/msal-angular/src/msal.guard.ts
```diff
@@ -79,9 +79,8 @@ export class MsalGuard implements CanActivate {
         /*
          * If a page with MSAL Guard is set as the redirect for acquireTokenSilent,
          * short-circuit to prevent redirecting or popups.
-         * TODO: Update to allow running in iframe once allowRedirectInIframe is implemented
          */
-        if (UrlString.hashContainsKnownProperties(window.location.hash) && BrowserUtils.isInIframe()) {
+        if (UrlString.hashContainsKnownProperties(window.location.hash)) {
```
jasonnutter, line 83: Sorry, I wasn't clear about the changes here. We want to keep the `isInIframe` check, but only if `allowRedirectInIframe` is `false`.
```suggestion
        if (UrlString.hashContainsKnownProperties(window.location.hash) && BrowserUtils.isInIframe() && !allowRedirectInIframe) {
```

I think for now I would revert this change, and we can talk about how to handle this incrementally, since may require some other work to expose the `allowRedirectInIframe` value.

jo-arroyo: Okay makes sense. Reverting

tnorling (approved):
Looks good, just a friendly reminder to also make relevant changes to the dev branch :)

jasonnutter:
> Looks good, just a friendly reminder to also make relevant changes to the dev branch :)

Since this is just for Angular v2, I don't think a separate PR is needed for dev.

tnorling:
> > Looks good, just a friendly reminder to also make relevant changes to the dev branch :)
> 
> Since this is just for Angular v2, I don't think a separate PR is needed for dev.

This contains a new export from msal-browser, need to make sure that is available in the next release

jasonnutter:
> > > Looks good, just a friendly reminder to also make relevant changes to the dev branch :)
> > 
> > 
> > Since this is just for Angular v2, I don't think a separate PR is needed for dev.
> 
> This contains a new export from msal-browser, need to make sure that is available in the next release

Ah right, missed that.

jo-arroyo:
> Looks good, just a friendly reminder to also make relevant changes to the dev branch :)

Thanks for the reminder, @tnorling !

PR #2687, opened 2020-12-03
Change the PkceGenerator code challenge encoding

samuelkubai:
Change the code challenge encoding from `ascii` to `base64` as the crypto buffer doesn't properly support encoding to anything other than `hex`, `latin1` or `base64`.


hectormmg (approved):
LGTM!

lib/msal-node/src/crypto/PkceGenerator.ts
```diff
@@ -37,7 +37,8 @@ export class PkceGenerator {
      */
     private generateCodeChallengeFromVerifier(codeVerifier: string): string {
         return EncodingUtils.base64EncodeUrl(
-            this.sha256(codeVerifier).toString("ascii")
+            this.sha256(codeVerifier).toString("base64"),
```
sameerag, line 40: the spec says: 
`S256
code_challenge = BASE64URL-ENCODE(SHA256(ASCII(code_verifier)))` [here](https://tools.ietf.org/html/rfc7636#section-4.2). Does this mean we are violating it by not following `ascii`? I am not too clear on how this is usually handled, let me dig a little more.

pkanher617: The `ASCII` transform is already done by this point, the toString needs to output the base64 string (which is the `BASE64URL-ENCODE` portion of the algorithm you posted).

pkanher617 (approved):
I think the tests for PKCE is already done in node but want to make sure before we merge this @samuelkubai.

PR #2692, opened 2020-12-03
[msal-node] Use CryptoProvider to generate PKCE codes in Auth Code PKCE sample application

hectormmg:
This PR adds the use of the CryptoProvider class to actually generate the PKCE codes in the auth-code-pkce AAD standalone sample and the EndToEndTestApp Auth Code scenario.

### DIFF
--- a/samples/msal-node-samples/standalone-samples/auth-code-pkce/index.js
+++ b/samples/msal-node-samples/standalone-samples/auth-code-pkce/index.js
@@ -25,53 +25,68 @@ const config = {
     }
 };
 
-/**
- * PKCE Setup
- * 
- * MSAL enables PKCE in the Authorization Code Grant Flow by including the codeChallenge and codeChallengeMethod parameters
- * in the request passed into getAuthCodeUrl() API, as well as the codeVerifier parameter in the
- * second leg (acquireTokenByCode() API).
- * 
- * Generating the codeVerifier and the codeChallenge is the client application's responsiblity.
- * For this sample, you can either implement your own PKCE code generation logic or use an existing tool
- * to manually generate a Code Verifier and Code Challenge, plugging them into the pkceCodes object below.
- * 
- * For details on implementing your own PKCE code generation logic, consult the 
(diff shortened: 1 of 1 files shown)


jasonnutter:
Out of curiousity, why do we make applications perform these steps themselves, instead of abstracting it away like we do for MSAL Browser?

derisen:
@hectormmg can we do the same for [b2c-auth-code-pkce](https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev/samples/msal-node-samples/standalone-samples/b2c-auth-code-pkce) sample?

hectormmg:
> Out of curiousity, why do we make applications perform these steps themselves, instead of abstracting it away like we do for MSAL Browser?

@jasonnutter I may be wrong here but I believe it was left out of the scope since msal-node would mostly be used for confidential clients. We've talked about abstracting it away eventually but it's not currently a priority.

>@hectormmg can we do the same for b2c-auth-code-pkce sample?

@derisen yes, I'll try to make it part of this PR if it doesn't get too large!

jasonnutter (approved):
For the node samples, I might suggest looking at something like `express-promise-router` to make the promise code in the handlers cleaner: https://github.com/express-promise-router/express-promise-router

samples/msal-node-samples/standalone-samples/auth-code-pkce/index.js
```diff
@@ -25,53 +25,68 @@ const config = {
...
-    }).catch((error) => console.log(JSON.stringify(error)));
+        // Add PKCE code challenge and challenge method to authCodeUrl request objectgit st
+        const authCodeUrlParameters = {
+            scopes: ["user.read"],
+            redirectUri: "http://localhost:3000/redirect",
+            codeChallenge: app.locals.pkceCodes.challenge, // PKCE Code Challenge
+            codeChallengeMethod: app.locals.pkceCodes.challengeMethod // PKCE Code Challenge Method
+        };
+
+        console.log(authCodeUrlParameters);
```
samuelkubai, line 74: Do we want to leave this log in?

PR #2698, opened 2020-12-05
Add action to deploy typedocs

jasonnutter:
This action will automatically build and deploy typedocs on pushes to master branch.

You can see it worked here: https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/gh-pages/ref

### DIFF
--- a/.github/workflows/typedoc.yml
+++ b/.github/workflows/typedoc.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v1
 
-      - name: Clean Install
+      - name: Install
         env:
           RUNNING_NODE_CI: 1
         run: npm ci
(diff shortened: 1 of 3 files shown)


.github/workflows/typedoc.yml
```diff
@@ -0,0 +1,70 @@
...
+          RUNNING_NODE_CI: 1
+        run: npm ci
+
+      - name: Build
+        run: npm run build
+
+      - name: Generate
+        run: npx lerna run doc:generate
+
+      - name: Deploy - msal-core
```
tnorling, line 32: If these deploy steps are all the same but for different libs you could use a matrix strategy instead. You can take a look at our CI steps to see how this is done [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/.github/workflows/build-steps.yml#L23)

The benefit is that each one will run in parallel and if one fails it won't stop deployment of the others

jasonnutter: Good to know, I'll look at doing this incrementally.

pkanher617 (approved):

tnorling (approved):

PR #2703, opened 2020-12-07
allow custom event use like v1, allows use of teams auth error handling

Venipa:
according to [https://docs.microsoft.com/en-us/microsoftteams/platform/tabs/how-to/authentication/auth-tab-aad](https://docs.microsoft.com/en-us/microsoftteams/platform/tabs/how-to/authentication/auth-tab-aad)

to allow users to emit/broadcast a event with login failure/success

### DIFF
--- a/lib/msal-angular/src/msal.broadcast.service.ts
+++ b/lib/msal-angular/src/msal.broadcast.service.ts
@@ -23,6 +23,6 @@ export class MsalBroadcastService {
         });
     }
     broadcast(message: EventMessage) {
-      this._msalSubject.next(message);
+        this._msalSubject.next(message);
     }
 }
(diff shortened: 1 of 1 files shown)


jasonnutter:
@Venipa Thanks for the contribution! Can you please add tests for this new API?

lib/msal-angular/src/msal.broadcast.service.ts
```diff
@@ -22,4 +22,7 @@ export class MsalBroadcastService {
             this._msalSubject.next(message);
         });
     }
+    broadcast(message: EventMessage) {
```
jasonnutter, line 25: Can you please add a return type?
```suggestion
    broadcast(message: EventMessage): void {
```

jasonnutter (changes requested):

jasonnutter:
@Venipa Can you please retarget this PR for our dev branch? And please include tests, thanks!

Issue #2705, opened 2020-12-07
Faulty url normalization causes an infinite redirect loop


jdolearydl:
## Library
- [x ] `@azure/msal-browser@2.7.0`

## Framework
N/A

## Description
ClientApplication.ts can cause an infinite redirect loop when loginRequestUrlNormalized doesn't equal currentUrlNormalized due to query string `?`.
For example: `"http://localhost:8000/observation/field?" !== "http://localhost:8000/observation/field"`

When ClientApplication.ts line 179
`if (loginRequestUrlNormalized === currentUrlNormalized && this.config.auth.navigateToLoginRequestUrl) {`
is false due to this mistaken mismatch, it eventually goes into line 208
                `await BrowserUtils.navigateWindow(loginRequestUrl, this.config.system.redirectNavigationTimeout, this.logger, true);` which navigates back to loginRequestUrl which starts the process all over and creates an infinite redirect loop.

This all happens because calling `UrlString.removeHashFromUrl('http://localhost:8000/observation/field?')` returns `'http://localhost:8000/observation/field?/'` instead of `'http://localhost:8000/observation/field/'`.
If removeHashFromUrl or possibly UrlString.canonicalizeUri were to also remove the queryString this bug would be avoided.

## Error Message
None

## MSAL Configuration
```js
{
    auth: {
        clientId: redacted,
        authority: redacted,
        redirectUri: "http://localhost:8000/login"
    },
    cache: {
        cacheLocation: "localStorage",
        storeAuthStateInCookie: false,
    }
}
```

## Reproduction steps

Simple way: write a test to prove that removeHashFromUrl will turn `'http://localhost:8000/foo?'` into `'http://localhost:8000/foo?/` which is obviously not the same url.

Full Reproduction: Start the **redirect flow** from a page such as `http://localhost:8000/observation/field?` which also calls  `OAuth.MSAL.handleRedirectPromise()`.  The `?` of the query string will cause the normalized urls to mismatch and create an infinite loop. (Note: it doesn't make a difference that the config redirectUri is not the same as the ORIGIN_URI)

## Expected behavior
It should sign the user in, not enter into an infinite redirect loop

## Browsers/Environment
- [x] Chrome
- [x] Firefox
- [x] Edge
- [ ] Safari
- [ ] IE
- [ ] Other (Please add browser name here)

jdolearydl:
[...]

PR #2707, opened 2020-12-07
Fix empty query string bug

tnorling:
When a redirect is initiated from a page with an empty ? in the url it will lead to an infinite loop

`clearHash` will remove the empty ? before comparison of loginRequestUrl to current url so the 2 will never be equal

This PR fixes this by removing the ? from the end of a url before comparison

### DIFF
--- a/lib/msal-common/src/url/UrlString.ts
+++ b/lib/msal-common/src/url/UrlString.ts
@@ -40,10 +40,16 @@ export class UrlString {
     static canonicalizeUri(url: string): string {
         if (url) {
             url = url.toLowerCase();
-        }
 
-        if (url && !StringUtils.endsWith(url, "/")) {
-            url += "/";
+            if (StringUtils.endsWith(url, "?")) {
+                url = url.slice(0, -1);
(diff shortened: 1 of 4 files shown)


jo-arroyo (approved):

jasonnutter (approved):

pkanher617 (approved):

lib/msal-common/src/url/UrlString.ts
```diff
@@ -40,10 +40,16 @@ export class UrlString {
...
             url = url.toLowerCase();
-        }
 
-        if (url && !StringUtils.endsWith(url, "/")) {
-            url += "/";
+            if (StringUtils.endsWith(url, "?")) {
+                url = url.slice(0, -1);
+            } else if (StringUtils.endsWith(url, "?/")) {
+                url = url.slice(0, -2);
+            }
```
jdolearydl, line 48: I'm curious about the intent here?  The example that I gave in #2705 only had a `?/` because this function (canonicalilzeUrl) added a `/` after the `?`.  Under what other circumstances would a `?/` need to be removed?

PR #2711, opened 2020-12-07
Add Brokered ssoSilent requests

pkanher617:
Adds the capability to broker `ssoSilent()` requests from the embedded app with the top frame's MSAL broker.


lib/msal-browser/src/broker/client/BrokerClientApplication.ts
```diff
@@ -151,17 +152,25 @@ export class BrokerClientApplication extends ClientApplication {
 
             switch (validMessage.interactionType) {
                 case InteractionType.Silent:
-                    return this.brokeredSilentRequest(validMessage, clientMessage.ports[0], this.brokerAccount);
+                    return this.brokeredSsoSilentRequest(validMessage, clientMessage.ports[0]);
```
jasonnutter, line 155: I'm not sure if want all silent requests going to ssoSilent. For example, if the broker already has an auth code, I think the broker can just make an acquireTokenSilent call, given we wouldn't need to open an iframe to get the needed auth state.

jasonnutter: The way I thought about it was:

- embedded app calls acquireTokenSilent -> broker calls acquireTokenSilent
- embedded app calls ssoSilent -> broker calls acquireTokenSilent, if that fails, then call ssoSilent.

pkanher617: This is to emulate the behavior of acquireTokenSilent today, which does the following in the normal case
1. Check cache for current access token
2. expire access token if valid refresh token is available
3. call ssoSilent if refresh token is expired or unavailable

In the case of the broker, we want to do the same thing, but the embeddedApp is not the one doing step 3, the broker is.

tnorling (approved):

jasonnutter (approved):

jmckennon (approved):

jo-arroyo (approved):

Issue #2720, opened 2020-12-09
ClientAssertion.parseCertificate fails to parse cert data with newlines


eanders-MS:
## Library
- [x] `@azure/msal-node@1.0.0-beta.2`

## Framework

NodeJS

## Description

It is common for certs exported from pfx to contain newlines. For example, OpenSSL exports certs in the format:
```
-----BEGIN CERTIFICATE-----
MIIFSDCCBDCg........................................
IqNDl5bnkmR0........................................
eUd4y23epjUW........................................
BnozMiiGCHuG........................................
jGqq8qpdA6wT........................................
...
-----END CERTIFICATE-----
```

The regex in ClientAssertion.parseCertificate is unable to parse such certs with newlines due to the `.` term's default behavior of not matching newlines. This can be fixed by the addition of the `s` (dotall) flag, like this:

```ts
    const regexToFindCerts = /-----BEGIN CERTIFICATE-----\n(.+?)\n-----END CERTIFICATE-----/gms;
```
at https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/cdb7e1d7b4afbb5908175ee7ee1059085bd95dff/lib/msal-node/src/client/ClientAssertion.ts#L111

## MSAL Configuration

```ts
        const config: msal.Configuration = {
            auth: {
                clientId: td.serverSetting("AAD_1P_CLIENT_ID"),
                authority: "https://login.microsoftonline.com/common/",
                clientCertificate: {
                    thumbprint: (new Buffer(CERT_THUMBPRINT, "base64")).toString(),
                    privateKey: td.serverSetting("CERT_PRIVATE_KEY"),
                    x5c: td.serverSetting("CERT_PUBLIC_CERTIFICATE")
                }
            }
        };
        this.client = new msal.ConfidentialClientApplication(config);
```

## Reproduction steps
1. Export certs from a pfx using this OpenSSL command:
    `$ openssl pkcs12 -in mycert.pfx -nokeys -out cert.pem`
2. Set this value as your MSAL client configuration's `auth.clientCertificate.x5c` value.
3. Initialize a ConfidentialClientApplication with this setting.
4. Notice the certificate failed to parse.

## Expected behavior
ClientAssertion.parseCertificate would be able to handle newlines.

PR #2721, opened 2020-12-09
ClientAssertion.parseCertificate - handle newlines

eanders-MS:
It is common for certs exported from pfx to contain newlines. For example, OpenSSL exports certs in the format:
```
-----BEGIN CERTIFICATE-----
MIIFSDCCBDCg........................................
IqNDl5bnkmR0........................................
eUd4y23epjUW........................................
BnozMiiGCHuG........................................
jGqq8qpdA6wT........................................
...
-----END CERTIFICATE-----
```

The regex in `ClientAssertion.parseCertificate` was unable to parse such certs due to the `.` term's default behavior of not matching newlines (even when the multiline flag is present). This is fixed by the addition of the s (dotall) flag.

Fixes https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/2720

### DIFF
--- a/change/@azure-msal-node-e1216fab-db96-4b5e-b319-17f3bcffa13b.json
+++ b/change/@azure-msal-node-e1216fab-db96-4b5e-b319-17f3bcffa13b.json
@@ -1,6 +1,6 @@
 {
   "type": "prerelease",
-  "comment": "ClientAssertion.parseCertificate - allow newlines in cert",
+  "comment": "ClientAssertion.parseCertificate - allow newlines in cert (#2721).",
   "packageName": "@azure/msal-node",
   "email": "email not defined",
   "dependentChangeType": "patch"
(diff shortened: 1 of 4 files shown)


jasonnutter:
@eanders-MS Please generate a beachball changelog file. This can be done by navigating to the root of the repo, running `npm install`, then running `npm run beachball:change`. You will want to select `prerelease`, thanks!

eanders-MS:
Looks like my changes to the unit test are failing. How do I run them locally? I was hoping it would succeed in CI, but that didn't happen.

change/@azure-msal-node-e1216fab-db96-4b5e-b319-17f3bcffa13b.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "ClientAssertion.parseCertificate - allow newlines in cert",
```
jmckennon, line 3:
```suggestion
  "comment": "ClientAssertion.parseCertificate - allow newlines in cert (#2721).",
```

This adds in the PR number to the changelog as well.

jmckennon (changes requested):
Looks good to me, nice catch on this functionality. We appreciate the help with the fix. Will approve after the PR number is added to the changelog file.

tnorling (approved):

jasonnutter (approved):
Thanks for the fix!

jmckennon (approved):

PR #2728, opened 2020-12-10
Add active account option in msal-browser

pkanher617:
This PR adds the ability to set an active account for an instance of MSAL. This account MUST be set by the application using the MSAL library, and will be used when no account is given to any of our APIs. 

This account is cleared when logout is called, either with no account or a matching account. If a non-matching account is passed to logout, the active account is kept intact.


lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -49,6 +49,9 @@ export abstract class ClientApplication {
     // Flag to indicate if in browser environment
     protected isBrowserEnvironment: boolean;
 
+    // Sets the account to use if no account info is given
+    private activeAccountId: string | null;
```
jasonnutter, line 53: I think we shuold consider making this name reflect which id is used:

```suggestion
    private activeLocalAccountId: string | null;
```

lib/msal-browser/src/app/IPublicClientApplication.ts
```diff
@@ -28,6 +28,8 @@ export interface IPublicClientApplication {
     ssoSilent(request: SsoSilentRequest): Promise<AuthenticationResult>;
     getLogger(): Logger;
     setLogger(logger: Logger): void;
+    setActiveAccount(account: AccountInfo): void;
+    getActiveAccount(): AccountInfo;
```
tnorling, line 32:
```suggestion
    getActiveAccount(): AccountInfo | null;
```

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -644,6 +652,29 @@ export abstract class ClientApplication {
         }
     }
 
+    /**
+     * Sets the account to use as the active account. If no account is passed to the acquireToken APIs, then MSAL will use this active account.
+     * @param account 
+     */
+    setActiveAccount(account: AccountInfo): void {
```
tnorling, line 659:
```suggestion
    setActiveAccount(account: AccountInfo | null): void {
```

lib/msal-browser/src/app/IPublicClientApplication.ts
```diff
@@ -28,6 +28,8 @@ export interface IPublicClientApplication {
     ssoSilent(request: SsoSilentRequest): Promise<AuthenticationResult>;
     getLogger(): Logger;
     setLogger(logger: Logger): void;
+    setActiveAccount(account: AccountInfo): void;
```
tnorling, line 31:
```suggestion
    setActiveAccount(account: AccountInfo | null): void;
```

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -644,6 +652,29 @@ export abstract class ClientApplication {
...
+        if (!account) {
+            this.activeAccountId = null;
+            return;
+        }
+        this.activeAccountId = account.localAccountId;
```
jasonnutter, line 664: Nitpick:

```suggestion
        this.activeAccountId = account ? account.localAccountId : null;
```

lib/msal-browser/src/app/ClientApplication.ts
pkanher617: done.

lib/msal-browser/src/app/IPublicClientApplication.ts
pkanher617: done.

lib/msal-browser/src/app/ClientApplication.ts
pkanher617: done.

lib/msal-browser/src/app/IPublicClientApplication.ts
pkanher617: done.

lib/msal-browser/src/app/ClientApplication.ts
pkanher617: done.

lib/msal-common/src/cache/entities/AccountEntity.ts
```diff
@@ -275,4 +275,21 @@ export class AccountEntity {
...
     }
+
+    /**
+     * Helper function to determine whether 2 arrays are equal
+     * Used to avoid unnecessary state updates
+     * @param arrayA 
+     * @param arrayB 
+     */
+    static accountInfoIsEqual(accountA: AccountInfo, accountB: AccountInfo): boolean {
+        if (!accountA || !accountB) {
```
tnorling, line 286: What if they're both null?

tnorling (approved):

lib/msal-angular/package-lock.json
```diff
@@ -420,24 +420,6 @@
...
-    },
-    "@azure/msal-common": {
-      "version": "2.0.0",
-      "resolved": "https://registry.npmjs.org/@azure/msal-common/-/msal-common-2.0.0.tgz",
-      "integrity": "sha512-d1RNcJb+P1EGzMHtgbZoVlHLQWjlVfr504jywNk9YEfoq8Hw3BxJ0wepu+1w0hc64D8zG0wljcvHaIH1jTn2SA==",
-      "dev": true,
-      "requires": {
-        "debug": "^4.1.1"
-      }
-    },
```
jasonnutter, line 440: Can you revert this?

jasonnutter (approved):

jasonnutter:
FYI @jo-arroyo once this gets merged, we should update the MSAL Angular v2 interceptor to use the active account (potentially falling back to the first account in getAllAccounts).

lib/msal-common/src/cache/entities/AccountEntity.ts
```diff
@@ -275,4 +275,21 @@ export class AccountEntity {
...
+     */
+    static accountInfoIsEqual(accountA: AccountInfo | null, accountB: AccountInfo | null): boolean {
+        if (!accountA || !accountB) {
+            return false;
+        }
+        return (accountA.homeAccountId === accountB.homeAccountId) && 
+            (accountA.localAccountId === accountB.localAccountId) &&
+            (accountA.username === accountB.username) &&
+            (accountA.tenantId === accountB.tenantId) &&
+            (accountA.environment === accountB.environment);
```
sameerag, line 293: Shouldn't we respect environment aliases here?

pkanher617: For now this is just an exact match. It is only being used to clear the active account. If we want to add environment aliasing for other scenarios we can, but I think for logout it is fine.

sameerag: okay!

sameerag (approved):
Lgtm. Please note the aliases change in case this use case is extended later.

PR #2758, opened 2020-12-15
Authority Metadata Caching

tnorling:
Updates authority instantiation to add caching of responses from instance discovery and `.well-known/openid-configuration` network calls. In `msal-browser` this is cached in-memory and will need to be refreshed on every page load. This solves several issues related to using more than 1 authority.

Also adds `authorityMetadata` configuration option from version 1 allowing developers to provide endpoint metadata and skip the network call. 

- [x] 100% Authority test coverage
- [x] Doc updates


hectormmg (approved):
LGTM!

lib/msal-browser/docs/performance.md
```diff
@@ -28,3 +28,25 @@ const msalInstance = new msal.PublicClientApplication({
     }
 });
 ```
+
+## Bypass authority metadata resolution
+
+By default, during the process of retrieving a token MSAL.js will make a network request to retrieve metadata from the authority configured for the request. If you would like to skip this network request, you can provide the required metadata in the configuration of `PublicClientApplication`.
+
+**Important:** It is your application's responsibility to ensure it is using correct, up-to-date authority metadata. Failure to do so may result in your application not working correctly.
```
sameerag, line 36: Should we explicitly mention that this is only if they choose to provide the values to `msal`?

sameerag (approved):
lgtm

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -749,17 +750,13 @@ export abstract class ClientApplication {
...
+        const discoveredAuthority = await this.getDiscoveredAuthority(requestAuthority);
+
         return {
             authOptions: {
                 clientId: this.config.auth.clientId,
                 authority: discoveredAuthority,
-                knownAuthorities: this.config.auth.knownAuthorities,
-                cloudDiscoveryMetadata: this.config.auth.cloudDiscoveryMetadata,
-                clientCapabilities: this.config.auth.clientCapabilities,
-                protocolMode: this.config.auth.protocolMode
```
jasonnutter, line 762: Why are these removed here?

jasonnutter: Not needed downstream?

tnorling: Not needed on the auth code client anymore. They're only needed on the authority instance which is initialized in the platform libs

pkanher617 (approved):

PR #2762, opened 2020-12-15
Fix useMsalAuthentication redirect error handling

### LINKED ISSUE #2750: useMsalAuthentication with interactionType Redirect infinite redirect on access denied
## Library
- [x] `@azure/msal-browser@2.7.0`
- [x] `@azure/msal-react@1.0.0.alpha1`

## Framework
- [ ] Angular
- [x] React 17
- [ ] Other

## Description
When I use `useMsalAuthentication` with interactionType `Redirect` together with an account that does not have access to the registered App, the `useMsalAuthentication` hook will automatically redirect even after unsuccessful login.

I do not have the same issue when using interactionType `Popup`, where I can clearly see in the code that `hasBeenCalled` is set to `true`, and an `error` is returned from the hook.

## Error Message
No error message.

## MSAL Configuration

```ts
const configuration: Configuration = {
  auth: {
    clientId: "", //App id removed from bug report
  },
};
const pca = new PublicClientApplication(configuration);
```

## Reproduction steps

I have an Azure AD SPA Registered App that requires User assignment. So only certain users are allowed to authenticate, based on a security group.

The code below will work as intended if the account I logged in with has access to Azure AD Registered App.

If the account I try to log in with does not have access, the page will keep redirecting me to the microsoft authentication page. 

```ts
const { login, result, error } = useMsalAuthentication(
    InteractionType.Redirect
  );
  const isAuthenticated = useIsAuthenticated();
  if (error) {
    return (
      <div>
        There has been an error: Code: ${error.errorCode}, Message: $
        {error.errorMessage}, Submessage: ${error.suberror}
      </div>
    );
  }
  if (!isAuthenticated) {
    return null;
  }
  return (
    <div>Authenticated</div>
```

I tried debugging `useMsalAuthentication`, and it seems the `hasBeenCalled` will not be true, and the error message will not be set.

Both of these do work when I set the interaction type to `Popup`. 

This bug only affects `Redirect`.

## Expected behavior
When a user does not have access to the app, I expect to get an error message as part of the `useMsalAuthentication` hook, so that I can display the error message to the user, instead of redirecting them to log in again.

## Identity Provider

- [x] Azure AD
- [ ] Azure B2C Basic Policy
- [ ] Azure B2C Custom Policy
- [ ] ADFS
- [ ] Other

## Browsers/Environment

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [x] Safari
- [ ] IE

## Regression

Version:

## Security

## Source

- [ ] Internal (Microsoft)
- [x] Customer request

tnorling:
Using `useMsalAuthentication` or `MsalAuthenticationTemplate` with redirect could cause redirect loops in scenarios where the service returned an error. This was due to the error not preventing another login call.

This PR fixes #2750


jasonnutter (approved):

jo-arroyo (approved):

leandrocanabarro:
Hello, when will the version be deployed?

PR #2771, opened 2020-12-16
msal-browser testing sample using msal-node to obtain tokens

tnorling:
SPAs using `msal-browser` to obtain tokens don't have an easy way to test their protected pages or APIs without using something like puppeteer to navigate through the AAD sign-in pages. This PR demonstrates how someone could use msal-node's ROPC flow to obtain tokens on the backend and pre-populate local/session storage with tokens. 

One new API is exposed on the msal-node `TokenCache` to retrieve the Key Value store to make this easier.


lib/msal-node/src/cache/TokenCache.ts
```diff
@@ -88,6 +88,10 @@ export class TokenCache implements ISerializableTokenCache {
         }
     }
 
+    getKVStore(): CacheKVStore {
```
jasonnutter, line 91: Do you think we should add tests for this?

jasonnutter (approved):
One question, otherwise looks good!

sameerag (commented):
Exposing ROPC to users should strictly go with warnings to discourage this for any other authentication purposes except for testing.

pkanher617:
> Exposing ROPC to users should strictly go with warnings to discourage this for any other authentication purposes except for testing.

I agree, I think we should make it clear that this should only be done for tests, not in a production app.

tnorling:
> > Exposing ROPC to users should strictly go with warnings to discourage this for any other authentication purposes except for testing.
> 
> I agree, I think we should make it clear that this should only be done for tests, not in a production app.

I also agree. I've added a warning to the top of the sample readme

pkanher617 (approved):

sameerag (approved):

PR #2782, opened 2020-12-17
Fix types in Angular v2 B2C sample

jasonnutter:
`BaseAuthRequest` no longer works in this scenario, instead should be either `PopupRequest` or `RedirectRequest`.


jo-arroyo (approved):

tnorling (approved):
Great catch!

PR #2789, opened 2020-12-18
Allow applications to disable postLogoutRedirectUri in MSAL Browser

### LINKED ISSUE #2145: How to disable post-logout redirect

jasonnutter:
Today MSAL Browser will fallback to the current page for teh `postLogoutRedirectUri`, which prevents applications from opting out of that behavior. This PR adds the ability to set `postLogoutRedirectUri` to `null` (in the logout request or in the configuration) to prevent setting a `post_logout_redirect_uri` on the logout request.


tnorling (approved):
Looks good. Would this potentially be a breaking change for people who are not providing a `postLogoutRedirectUri` explicitly today? Is patch the right designation?

jasonnutter:
@tnorling I can see that argument, however, I would say that the inability to not set a post logout redirect URI is a bug. I suppose one way we could make this somewhat backwards compatible is to make it so that setting `null` as the `postLogoutRedirectURI` (which I doubt anyone is doing today) is the way to turn off the post logout redirect.

tnorling:
@jasonnutter I think I would slightly prefer `null` over empty string, however, either approaches are fine. I would also be fine with making this a minor version update and calling this out as "potentially breaking".

pkanher617 (approved):
I agree that we should probably call this out as a breaking change, I am fine with doing it as a minor version. If we opt to make this backwards compatible instead (which might be a better option here) I think I would prefer using `null`.

jasonnutter:
Updated to make non-breaking: you must explicitly pass `null` to disable post logout redirect.

jo-arroyo (approved):

pkanher617 (approved):

manavsharmaprog:
Found this fix after a lot of Google search.

> May I please request you to bring the documentation up to speed with this small trick of passing **null** in postlogoutredirecturi.


My use case was that I was logging out and got redirected back to login screen which caused the interaction_in_progress to be set and giving me an error on Chrome Dev console. Clearing the cookie manually didn't look like a good idea either.
I know there're some better ways to handle redirects but a lot of people may end up bumping into a similar use case as mine.

PR #2790, opened 2020-12-18
Fix bugs in Broker flow

pkanher617:
This PR fixes the following bugs in the broker flow:
- caches tokens with the correct client id (previously was always cached with broker client id)
- issue with temp cache items not clearing is no longer present
- set and use the active account specifically for broker scenarios
- add support for wildcard domains in the allowed broker origins
- match the `preferred_interaction_type` behavior to what is described in the design document
- Errors lost their typing after being sent in `postMessage`


lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -187,4 +187,26 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
+        if (this.broker) {
+            this.broker.setActiveAccount(account);
+        }
+        super.setActiveAccount(account);
+    }
+
+    /**
+     * Gets the currently active account
+     */
+    getActiveAccount(asBroker?: boolean): AccountInfo | null {
```
tnorling, line 205: These changes should be in the `ExperimentalClient`, right?

pkanher617: See [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/2807/files#diff-a22a6fa751365f69045c4a587ace7c208315decc7f08f8b5af791bc9b00057a2R217-R236)

lib/msal-browser/src/broker/msg/resp/BrokerHandshakeResponse.ts
```diff
@@ -35,8 +36,11 @@ export class BrokerHandshakeResponse extends BrokerMessage {
             message.data.messageType === BrokerMessageType.HANDSHAKE_RESPONSE &&
             message.data.version) {
             // TODO, verify version compatibility
-            if (trustedBrokerDomains.indexOf(message.origin) < 0) {
-                // TODO make this a browser Error
+            const matchedDomains = trustedBrokerDomains.filter((domain: string) => {
+                return StringUtils.matchPattern(domain, message.origin);
```
tnorling, line 40: There's a `getDomainFromUrl` function on `UrlString` which may safer than just checking if one string is in another string

samples/msal-browser-samples/BrokerTestApp2.0/app/brokeredWidget/script.js
```diff
@@ -49,7 +49,77 @@ function handleResponse(resp) {
...
+                const contentElement = document.getElementsByClassName("myContent")[0];
+                contentElement.innerHTML = "I am now trying to fetch profile data";
+        
+                setTimeout(async () => {
+                    let exit = false;
+                    const loginReq = { 
+                        scopes: ["openid", "profile", "User.Read"],
+                        loginHint: "idlab@msidlab4.onmicrosoft.com" 
+                    };
+                    // await myMSALObj.ssoSilent(loginReq).then(() => {
```
tnorling, line 62: Why is this commented out?

lib/msal-browser/src/broker/msg/resp/BrokerHandshakeResponse.ts
pkanher617: I'm using `StringUtils.matchPattern()` so that we can support wildcard patterns in the config (i.e. `https://*.contoso.com`).

lib/msal-browser/src/app/PublicClientApplication.ts
tnorling: Could we rebase this PR into that one then? Or remove these changes here?

lib/msal-browser/src/broker/msg/resp/BrokerHandshakeResponse.ts
tnorling: Gotcha, thanks!

samples/msal-browser-samples/BrokerTestApp2.0/app/brokeredWidget/script.js
pkanher617: Fixed in #2806 but I can uncomment here also. Was done for testing.

lib/msal-browser/src/app/PublicClientApplication.ts
pkanher617: The experimental API stuff was decided after these changes were already made, which is why they are in a subsequent PR. I could potentially rebase but I think that is going to cause more issues, and take much longer. Removing these changes isn't possible because we are using these APIs in this PR.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -187,4 +187,26 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
+        if (asBroker) {
```
jasonnutter, line 206: What are the use cases for setting this to `true` or `false`?

pkanher617: If you are attempting to get the broker's active account, pass `true`, otherwise pass `false`. Basically, if you are in a brokered scenario or not. I had to do this since the broker's active account is different from the PCA's active account, and there isn't a way for these two apps to share them today.

jasonnutter: I think those should always be the same account.

jasonnutter: It was brought up in email that storing the active account in browser storage would be valuable, sounds like it would help here as well.

pkanher617: Resolved offline

lib/msal-browser/src/broker/msg/resp/BrokerHandshakeResponse.ts
```diff
@@ -35,8 +36,11 @@ export class BrokerHandshakeResponse extends BrokerMessage {
             message.data.messageType === BrokerMessageType.HANDSHAKE_RESPONSE &&
             message.data.version) {
             // TODO, verify version compatibility
-            if (trustedBrokerDomains.indexOf(message.origin) < 0) {
-                // TODO make this a browser Error
+            const matchedDomains = trustedBrokerDomains.filter((domain: string) => {
```
pkanher617, line 39: implicit return

lib/msal-browser/src/broker/msg/resp/BrokerAuthResponse.ts
```diff
@@ -29,27 +32,55 @@ export class BrokerAuthResponse extends BrokerMessage {
...
-            // TODO: verify version compat
-
-            return new BrokerAuthResponse(message.data.interactionType, message.data.result, message.data.error);
+            return new BrokerAuthResponse(message.data.interactionType, message.data.result, BrokerAuthResponse.detectError(message));
         }
 
         return null;
     }
 
+    static detectError(message: MessageEvent): AuthError {
```
tnorling, line 41: I think we discussed moving some of this logic to either its own file in /error or into one of the existing error files. Could we do that in this PR?

lib/msal-browser/src/broker/msg/resp/BrokerHandshakeResponse.ts
```diff
@@ -35,8 +36,9 @@ export class BrokerHandshakeResponse extends BrokerMessage {
             message.data.messageType === BrokerMessageType.HANDSHAKE_RESPONSE &&
             message.data.version) {
             // TODO, verify version compatibility
-            if (trustedBrokerDomains.indexOf(message.origin) < 0) {
-                // TODO make this a browser Error
+            const matchedDomains = trustedBrokerDomains.filter((domain: string) => StringUtils.matchPattern(domain, message.origin));
```
tnorling, line 39: Can we double check and add a test to make sure this won't match urls that contain the search string anywhere other than the domain part? e.g. shouldn't match if the url is in the query string

jasonnutter (approved):

lib/msal-browser/src/broker/msg/resp/BrokerAuthResponse.ts
pkanher617: I've done that in a separate PR. There are a lot of moving parts and changing it here will end up causing issues down the line.

lib/msal-browser/src/broker/msg/resp/BrokerHandshakeResponse.ts
pkanher617: The tests for this I think are done already for the `matchPattern` API. If we need to update them I will take a look, but these tests are going to be part of a separate PR.

tnorling (approved):

PR #2792, opened 2020-12-19
Enable strict compiler option in msal-browser

tnorling:
Enables the TypeScript `strict` compiler option for `msal-browser`


lib/msal-browser/src/app/IPublicClientApplication.ts
```diff
@@ -12,20 +12,20 @@ import { EndSessionRequest } from "../request/EndSessionRequest";
...
     getAccountByLocalId(localId: string): AccountInfo | null;
     getAccountByUsername(userName: string): AccountInfo | null;
     getAllAccounts(): AccountInfo[];
     handleRedirectPromise(hash?: string): Promise<AuthenticationResult | null>;
-    loginPopup(request?: PopupRequest): Promise<AuthenticationResult>;
+    loginPopup(request?: PopupRequest): Promise<AuthenticationResult | null>;
     loginRedirect(request?: RedirectRequest): Promise<void>;
     logout(logoutRequest?: EndSessionRequest): Promise<void>;
-    ssoSilent(request: SsoSilentRequest): Promise<AuthenticationResult>;
+    ssoSilent(request: SsoSilentRequest): Promise<AuthenticationResult | null>;
```
jasonnutter, line 28: I don't think any of these can actually be null though?

jasonnutter: Ah, its because some of the underlying functions can return `null`, yeah?

tnorling: Yeah that's right. It looks like there's only one place in one of the functions in common that returns null. If this is something we want to change it would be relatively low impact.

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -83,33 +81,34 @@ export abstract class ClientApplication {
...
+        // Array of events
+        this.eventCallbacks = new Map();
 
         // Initialize logger
         this.logger = new Logger(this.config.system.loggerOptions, name, version);
 
+        // Initialize the network module class.
+        this.networkClient = this.config.system.networkClient;
+
+        if (!this.isBrowserEnvironment) {
```
tnorling, line 99: The changes in the constructor have implications on SSR and I think this creates risk that we may break SSR in the future. Open to feedback on a better way to both ensure SSR is supported and properties on PCA are "definitely" initialized

lib/msal-browser/src/app/IPublicClientApplication.ts
jasonnutter: I would prefer that we are able to make this change without needing to change the public API, especially give these APIs will never actually return `null` to the user.

lib/msal-browser/src/app/ClientApplication.ts
jasonnutter: There should be unit tests that make sure the constructor doesnt throw when `window` isn't defined.

lib/msal-browser/src/app/IPublicClientApplication.ts
tnorling: Done.

pkanher617 (approved):

sameerag (approved):

PR #2806, opened 2020-12-22
Add caching for broker responses in memory, lookup by origin and thumbprint

pkanher617:
This PR adds the ability for the broker to cache responses from a redirect. It also uses embedded app origin to look up these responses, preventing child apps from receiving responses that do not belong to them.


lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -311,17 +316,17 @@ export abstract class ClientApplication {
             // Create acquire token url.
             const navigateUrl = await authClient.getAuthCodeUrl(validRequest);
 
-            const redirectStartPage = (request && request.redirectStartPage) || window.location.href;
+            const redirectStartPage = (validRequest && userRedirectStartPage) || window.location.href;
```
tnorling, line 319:
```suggestion
            const redirectStartPage = userRedirectStartPage || window.location.href;
```

lib/msal-browser/src/broker/client/BrokerClientApplication.ts
```diff
@@ -30,24 +27,44 @@ import { BrokerSsoSilentRequest } from "../request/BrokerSsoSilentRequest";
...
+     * has loaded during redirect flows. This should be invoked on all page loads involved in redirect
+     * auth flows.
+     * @param hash Hash to process. Defaults to the current value of window.location.hash. Only needs to be provided explicitly if the response to be handled is not contained in the current value.
+     * @returns {Promise.<AuthenticationResult | null>} token response or null. If the return value is null, then no auth redirect was detected.
+     */
+    async handleRedirectPromise(hash?: string): Promise<BrokerAuthenticationResult | null> {
+        const brokerResponse = super.handleRedirectPromise(hash) as Promise<BrokerAuthenticationResult>;
+        this.currentBrokerRedirectResponse = this.waitForBrokeredResponse(brokerResponse);
+        const redirectResponse = await this.currentBrokerRedirectResponse;
+        this.currentBrokerRedirectResponse = undefined;
```
tnorling, line 47: This feels a little hacky but I can't think of an alternative at the moment. This may be something we should do in the base `handleRedirectPromise` to prevent the race conditions we hear about from time to time.

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -57,6 +57,8 @@ export type BrokerOptions = {
     preferredInteractionType?: InteractionType.Popup | InteractionType.Redirect | InteractionType.None;
     allowBrokering?: boolean;
     trustedBrokerDomains?: string[];
+    brokerRedirectStartPage?: string;
```
tnorling, line 60: Any reason a brokered scenario would not want to use the same `redirectStartPage` and `onRedirectNavigate` as the non-brokered scenario?

pkanher617: `redirectStartPage` and `onRedirectNavigate` shouldn't be set by the embedded application, since they are not the ones doing the redirect. The broker app should specify A. which page the redirect starts on and B. what happens after a redirect.

pkanher617: We block all redirect APIs in an embedded scenario anyway, so these config options don't get used today in an embedded scenario.

lib/msal-browser/src/broker/client/BrokerClientApplication.ts
pkanher617: It is a bit hacky, I needed a way to wait for the response to finish in case the broker receives a request for the redirect response before it is done trading for the tokens. Happy to hear alternatives. The reason I haven't done this in the base `handleRedirectPromise()` is because I am trying to keep broker specific code away from the base functions.

tnorling: I'm advocating for making this not broker specific. We have scenarios where if you call `handleRedirectPromise` twice at the same time it has the potential to create a race condition and break. I think moving this logic up one level, into the base function would solve both problems. Ultimately what you're trying to prevent here is having more than one instance of `handleRedirectPromise` running concurrently, right?

lib/msal-browser/src/config/Configuration.ts
tnorling: Sorry should have been more clear, why do we need separate `redirectStartPage` and `brokerRedirectStartPage` params instead of having the broker just use `redirectStartPage` regardless of whether it is getting tokens for itself or for an embedded app?

lib/msal-browser/src/broker/client/BrokerClientApplication.ts
pkanher617: That is more of a side effect, the main reason is that I need a way to determine if we are currently trading for tokens or not, so that when an embedded app makes a request we can wait for the tokens to be retrieved before returning a response back from the broker.

lib/msal-browser/src/config/Configuration.ts
pkanher617: redirectStartPage is set in the context of the client making the request, and is only a parameter on the request. The broker doesn't have any control over what the embedded application sends as the redirectStartPage, but we don't want the embeddedApp telling the broker to use the redirectStartPage for its own app - the broker should specify that themselves. The easiest way to do that is to add a broker config option for it in the constructor config.

tnorling: Got it. Forgot the params are only set on the request. Makes sense, thanks for clarifying

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -337,49 +342,17 @@ export abstract class ClientApplication {
...
-            // Since this function is synchronous we need to reject
-            return Promise.reject(e);
-        }
-
-        // asyncPopups flag is true. Acquires token without first opening popup. Popup will be opened later asynchronously.
-        if (this.config.system.asyncPopups) {
-            return this.acquireTokenPopupAsync(request);
-        } else {
-            // asyncPopups flag is set to false. Opens popup before acquiring token.
-            const popup = PopupHandler.openSizedPopup();
```
jasonnutter, line 359: Popups opened by the broker should also include this logic.

lib/msal-browser/src/broker/client/BrokerClientApplication.ts
```diff
@@ -30,24 +27,44 @@ import { BrokerSsoSilentRequest } from "../request/BrokerSsoSilentRequest";
...
+        if (redirectResponse) {
+            if (!redirectResponse.tokensToCache) {
+                return redirectResponse;
+            }
+        }
+        return null;
```
jasonnutter, line 53:
```suggestion
        return (redirectResponse && !redirectResponse.tokensToCache) ? redirectResponse : null;
```

lib/msal-browser/src/broker/client/BrokerClientApplication.ts
```diff
@@ -107,19 +124,31 @@ export class BrokerClientApplication extends ClientApplication {
...
+            if (this.currentBrokerRedirectResponse) {
+                await this.currentBrokerRedirectResponse;
+            }
+
+            const memCacheKeys = this.browserStorage.getMemoryKeys();
+            const embeddedAppKey = `${MemoryCacheKeys.BROKER_RESPONSE}.${clientMessage.origin}`
+            const cachedResponseKeys = memCacheKeys.filter((cacheKey) => cacheKey.indexOf(embeddedAppKey));
+            if (cachedResponseKeys.length >= 1) {
+                if (cachedResponseKeys.length > 1) {
+                    this.logger.error("Too many responses found for the origin, sending back the first one found. You may need to call acquireToken() again.")
```
jasonnutter, line 137: `acquireToken`?

lib/msal-browser/src/broker/client/BrokerClientApplication.ts
```diff
@@ -136,13 +165,23 @@ export class BrokerClientApplication extends ClientApplication {
...
+                await this.currentBrokerRedirectResponse;
+            }
+
+            const reqThumbprint: RequestThumbprint = {
+                authority: validMessage.request.authority,
+                clientId: validMessage.embeddedClientId,
+                scopes: validMessage.request.scopes
+            };
+            const responseThumbprint = `${validMessage.embeddedAppOrigin}.${this.browserCrypto.base64Encode(JSON.stringify(reqThumbprint))}`;
+            const responseCacheKey = `${MemoryCacheKeys.BROKER_RESPONSE}.${responseThumbprint}`;
```
jasonnutter, line 178: Can you factor these into helper functions? Would help readability.

lib/msal-browser/src/app/ClientApplication.ts
pkanher617: We will need to add that property to the broker options then, since the broker will need to specify the popup in constructor.

lib/msal-browser/src/config/Configuration.ts
pkanher617, line 60: Add Omit type for redirect/popup params

pkanher617: Think about adding in a later version?

lib/msal-browser/src/broker/client/BrokerClientApplication.ts
```diff
@@ -30,24 +27,44 @@ import { BrokerSsoSilentRequest } from "../request/BrokerSsoSilentRequest";
...
-     * 
+     * Event handler function which allows users to fire events after the PublicClientApplication object
+     * has loaded during redirect flows. This should be invoked on all page loads involved in redirect
+     * auth flows.
+     * @param hash Hash to process. Defaults to the current value of window.location.hash. Only needs to be provided explicitly if the response to be handled is not contained in the current value.
+     * @returns {Promise.<AuthenticationResult | null>} token response or null. If the return value is null, then no auth redirect was detected.
+     */
+    async handleRedirectPromise(hash?: string): Promise<BrokerAuthenticationResult | null> {
+        const brokerResponse = super.handleRedirectPromise(hash) as Promise<BrokerAuthenticationResult>;
+        this.currentBrokerRedirectResponse = this.waitForBrokeredResponse(brokerResponse);
```
pkanher617, line 45: Add comments

lib/msal-browser/src/broker/client/BrokerClientApplication.ts
```diff
@@ -107,19 +124,31 @@ export class BrokerClientApplication extends ClientApplication {
...
-            if (brokerResult) {
-                // TODO: Replace with in-memory cache lookup
-                this.cachedBrokerResponse = undefined;
-                const brokerAuthResponse: BrokerAuthResponse = new BrokerAuthResponse(InteractionType.Redirect, brokerResult);
-                this.logger.info(`Sending auth response: ${brokerAuthResponse}`);
-                clientPort.postMessage(brokerAuthResponse);
-                clientPort.close();
-                return;
+            if (this.currentBrokerRedirectResponse) {
+                await this.currentBrokerRedirectResponse;
```
pkanher617, line 129: Add comments

lib/msal-browser/src/app/ClientApplication.ts
pkanher617: Will resolve in a separate PR once the fixes for popup flow has been merged to dev

lib/msal-browser/src/broker/client/BrokerClientApplication.ts
pkanher617: done.

lib/msal-browser/src/broker/client/BrokerClientApplication.ts
pkanher617: done.

lib/msal-browser/src/config/Configuration.ts
pkanher617: done.

lib/msal-browser/src/broker/client/BrokerClientApplication.ts
pkanher617: done.

lib/msal-browser/src/broker/client/BrokerClientApplication.ts
pkanher617: fixed

lib/msal-browser/src/broker/client/BrokerClientApplication.ts
pkanher617: fixed

lib/msal-browser/src/app/ClientApplication.ts
pkanher617: done.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -3,7 +3,7 @@
  * Licensed under the MIT License.
  */
 
-import { AccountInfo, AuthenticationResult, SilentFlowRequest } from "@azure/msal-common";
+import { AccountInfo, AuthenticationResult, AuthorizationUrlRequest, PromptValue, SilentFlowRequest } from "@azure/msal-common";
```
tnorling, line 6: Once you update from dev these request types should be imported from within browser, FYI

tnorling (approved):

lib/msal-browser/src/app/PublicClientApplication.ts

PR #2807, opened 2020-12-23
Experimental API for PublicClientApplication

pkanher617:
This PR adds an experimental API for the PublicClientApplication class. This is done by creating a new class called ExperimentalClientAPI, which exposes any APIs which are in an "experimental" phase but should still be usable by developers. You can see the usage of the experimental API in the sample files.


tnorling (approved):

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -88,6 +88,7 @@ export type BrowserSystemOptions = SystemOptions & {
 };
 
 export type ExperimentalOptions = {
+    enableExperimentalApi?: boolean;
```
pkanher617, line 91: change to enable

pkanher617: possible hang off of `system` instead of `experimental`?

pkanher617: chose not to hang off system as it starts putting experimental references where they don't need to be. Easier to group under a single config.

Issue #2614, opened 2020-11-18
Subsequent calls to acquireTokenPopup break the first invocation and prevent authentication


jviney:
## Library
- [x] `@azure/msal-common@1.7.2`
- [x] `@azure/msal-browser@2.7.0`

## Framework

AngularJS.

## Description

The first call to `acquireTokenPopup` works correctly. If there is a subsequent call to `acquireTokenPopup` before the first one has completed, it causes the first popup to go blank, and raises a `BrowserAuthError.interaction_in_progress` error.

## MSAL Configuration

```js
new PublicClientApplication({
  auth: {
      clientId: "client-id",
    },
    cache: {
      cacheLocation: "localStorage",
      storeAuthStateInCookie: false,
    }
})
```

## Reproduction steps
```js
application.acquireTokenPopup()

setTimeout(() => {
  application.acquireTokenPopup()
}, 500)
```

## Expected behavior

Subsequent calls to `acquireTokenPopup` should raise the `BrowserAuthError.interaction_in_progress` without interrupting the first auth flow.

## Browsers/Environment
- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari
- [ ] IE
- [ ] Other (Please add browser name here)

jdolearydl:
I observed the similar behavior; however, for me it was the second, most recent popup that goes blank.

jasonnutter:
PR to fix issues related to blank popup windows is linked above. This will allow separate tabs to open their own popups, and prevent the first popup from breaking when a second one is opened.

PR #2842, opened 2021-01-08
Fix handling of multiple popup windows

jasonnutter:
This PR fixes the following bugs:

1. If the same app (using local storage) is opened in multiple tabs, only the first app instance is able to open a popup, due to temporary cache values being shared between tabs. This is addressed by moving temporary cache values to their own sessionStorage or memoryStorage instance so that they are not shared between tabs, and by setting a dynamic popup name so that the second tab won't try to open its url in the popup window opened by the first tab.
2. If an app invokes one of the popup APIs twice in a row without letting the first request finish, the popup for the first request will be broken, due to MSAL initially opening `about:blank` in the already opened popup (see #2614). This is fixed by moving the interaction in progress check above the code that initially opens the popup to `about:blank`.

There will be a follow up discussion about further enhancements we can make to the ergonomics of handling multiple concurrent popup requests.


tnorling (approved):
Should we add a couple E2E tests for these scenarios?

jasonnutter:
> Should we add a couple E2E tests for these scenarios?

Looks like some e2e tests are broked, so may not need to add new ones.

pkanher617 (approved):
I also think we should add e2e tests (I'm not seeing where they are broken?)

gabriel-kohen-by:
Hi Folks,

I'm using the libraries with the following versions:
```
    "@azure/msal-browser": "^2.12.0",
    "@azure/msal-react": "^1.0.0-beta.0"
```

The config object is:
```
const msalProviderConfig = {
  type: 'popup',
  msalConfig: {
    auth: {
      clientId,
      authority,
      redirectUri,
      postLogoutRedirectUri,
    },
    system: {
      loggerOptions: {
        loggerCallback: getLoggerCallback(),
      },
    },
  },
  silentRequestConfig: {
    scopes: [scopes, 'offline_access'],
  },
  endSessionRequestConfig: {},
  loginRequestConfig: {
    scopes: [scopes, 'offline_access'],
  },
};
```
Post logout redirect URL is `${window.location.origin}/` and redirectUrl is `redirectUri`
The as you can see seems to be failing intermittently. See the attached recording where all is working fine for several times until around the 50th second mark of the video the about:blank window stays hanging.
[Video clip recreating the issue](https://photos.app.goo.gl/A8oK5NQCPj21hYJr6)
Any idea what I'm might be doing wrong?
Thank you for the otherwise rock solid library.

@tnorling @jo-arroyo @jasonnutter @pkanher617

tnorling:
@gabriel-kohen-by Can you please open an issue? Thanks!

PR #2845, opened 2021-01-08
Send wrapper SKU and version on telemetry headers

tnorling:
Adds an `initializeWrapperLibrary` API to `msal-browser` to be called by `msal-react`, `msal-angular` and future wrappers. Accepts the SKU and version and passes it down to the `ServerTelemetryManager` on each request.

The `x-client-current-telemetry` header will add the sku and version to the "platform fields" at the end like so:

`x-client-current-telemetry: 2|865,0|@azure/msal-react,1.0.0-alpha.1`


lib/msal-angular/src/constants.ts
```diff
@@ -11,6 +11,7 @@ export const MSAL_GUARD_CONFIG = new InjectionToken<string>("MSAL_GUARD_CONFIG")
 
 export const MSAL_INTERCEPTOR_CONFIG = new InjectionToken<string>("MSAL_INTERCEPTOR_CONFIG");
 
-export const name = "@azure/msal-angular";
-
-export const version = "2.0.0-alpha.1";
+export const Constants = {
+    SKU: "@azure/msal-angular",
+    VERSION: "2.0.0-alpha.3"
```
jasonnutter, line 16: Lets try to solve pulling this in dynamically tomorrow.

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -1034,5 +1040,21 @@ export abstract class ClientApplication {
...
     }
+
+    /**
+     * Called by wrapper libraries (Angular & React) to set SKU and Version passed down to telemetry, logger, etc.
+     * @param sku 
+     * @param version 
+     */
+    initializeWrapperLibrary(sku: string, version: string): void {
+        // Validate the SKU passed in is one we expect
+        if (sku.match(/^@azure\/msal\-[a-z]+$/g) && version.match(/^[0-9a-z\.\-]+$/g)) {
```
jasonnutter, line 1051: Could this be done without regex?

jasonnutter: Also, what would be the downside if "bad" skus and versions were provided?

tnorling: Mostly didn't want bad data polluting our telemetry data but also didn't want to open a door to sending unsanitized data to eSTS/Kusto. I'm positive they sanitize the data as well but I think it's good practice to not blindly accept user provided data as-is

tnorling: Updated to use an enum instead.

lib/msal-angular/src/constants.ts
tnorling: Addressed in #2915

lib/msal-browser/src/utils/BrowserConstants.ts
```diff
@@ -87,3 +87,9 @@ export const DEFAULT_REQUEST: RedirectRequest|PopupRequest = {
 
 // JWK Key Format string (Type MUST be defined for window crypto APIs)
 export const KEY_FORMAT_JWK = "jwk";
+
+// Supported wrapper SKUs
+export enum WrapperSKUs {
```
tnorling, line 92:
```suggestion
export enum WrapperSKU {
```

pkanher617 (approved):

jo-arroyo (approved):
LGTM

PR #2851, opened 2021-01-11
API extractor for msal-node

sameerag:
Introducing [API extractor](https://api-extractor.com/pages/overview/intro/) for `msal-node`.

This PR adds support for commands: `npm run api` and `npm run api:local` for production and local runs  to generate an api-extractor report.

**Future:**
* Working on adding support to generate a versioned `api-report.md` for every public release. This would help with tracking API changes both for the dev team and the consumers of this library.
* Add support for all libraries
* Consolidate all api-report files in the <root>/api/


jasonnutter:
Should the `/temp` folder be added to gitignore? Doesn't seem like those are files that should be committed.

lib/msal-node/src/cache/NodeStorage.ts
```diff
@@ -23,8 +23,9 @@ import { InMemoryCache, JsonCache, CacheKVStore } from "./serializer/SerializerT
 
 /**
  * This class implements Storage for node, reading cache from user specified storage location or an  extension library
+ * @public
  */
-export class Storage extends CacheManager {
+export class NodeStorage extends CacheManager {
```
sameerag, line 28: Had to rename this as `Storage` is conflicting with some global symbol while generating the `index.d.ts` and auto generating the name as `Storage_2` in the final API report. There is an open issue in github regarding this: https://github.com/microsoft/rushstack/issues/1830 

Henceforth had to rename. It also is consistent with our general naming convention for Browser too.

sameerag:
Note: Naming the `request` as the same in `msal-common` and `msal-node` internally renames the base request as `request_2`. Eg: `DeviceCodeRequest` if defined in both packages `msal-common` and `msal-node`, when compiled and linked,  is differentiated in the rollup .d.ts as `DeviceCodeRequest` and `DeviceCodeRequest_2`. 

API report ends up using the same names. To avoid this, renamed the `msal-common` request names.

lib/msal-browser/src/event/EventMessage.ts
```diff
@@ -3,7 +3,7 @@
  * Licensed under the MIT License.
  */
 
-import { AuthenticationResult, AuthError, EndSessionRequest } from "@azure/msal-common";
+import { AuthenticationResult, AuthError, CommonEndSessionRequest } from "@azure/msal-common";
```
sameerag, line 6: Q: Why is the `EventPayload` based on `msal-common` `EndSessionRequest` instead of `msal-browser`?

lib/msal-common/src/index.ts
```diff
@@ -51,15 +51,15 @@ export { ICrypto, PkceCodes, DEFAULT_CRYPTO_IMPLEMENTATION } from "./crypto/ICry
...
-export { UsernamePasswordRequest } from "./request/UsernamePasswordRequest";
+export { CommonAuthorizationUrlRequest } from "./request/CommonAuthorizationUrlRequest";
+export { CommonAuthorizationCodeRequest } from "./request/CommonAuthorizationCodeRequest";
+export { CommonRefreshTokenRequest } from "./request/CommonRefreshTokenRequest";
+export { CommonClientCredentialRequest } from "./request/CommonClientCredentialRequest";
+export { CommonOnBehalfOfRequest } from "./request/CommonOnBehalfOfRequest";
+export { CommonSilentFlowRequest } from "./request/CommonSilentFlowRequest";
+export { CommonDeviceCodeRequest } from "./request/CommonDeviceCodeRequest";
+export { CommonEndSessionRequest } from "./request/CommonEndSessionRequest";
+export { CommonUsernamePasswordRequest } from "./request/CommonUsernamePasswordRequest";
```
jasonnutter, line 62: These would all warrant a breaking change in common, fyi.

lib/msal-node/src/cache/NodeStorage.ts
jasonnutter: Is this why the other common exports needed to be renamed?

lib/msal-angular/package-lock.json
```diff
@@ -14209,4 +14209,4 @@
       "dev": true
     }
   }
-}
+}
```
jasonnutter, line 14212: This change can be discarded, so that the angular change file isn't needed.

lib/msal-browser/src/event/EventMessage.ts
jasonnutter: Not sure, looks like it should be able to use the one from browser. @jo-arroyo can confirm.

change/@azure-msal-common-2021-01-28-23-58-32-api-extractor.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "none",
```
jasonnutter, line 2: Assuming the exports are renamed: 

```suggestion
  "type": "major",
```

lib/msal-browser/src/event/EventMessage.ts
jo-arroyo: Hmm... must have been overlooked. Should be `EndSessionRequest` from `msal-browser` instead.

jasonnutter:
I would advocate for a different folder name than `temp`, makes it seems like files that shouldn't be committed. What about `api`?

> Note: Naming the `request` as the same in `msal-common` and `msal-node` internally renames the base request as `request_2`. Eg: `DeviceCodeRequest` if defined in both packages `msal-common` and `msal-node`, when compiled and linked, is differentiated in the rollup .d.ts as `DeviceCodeRequest` and `DeviceCodeRequest_2`.
> 
> API report ends up using the same names. To avoid this, renamed the `msal-common` request names.

Makes sense, unfortunate that API extractor doesn't handle naming conflicts more gracefully. That being said, I don't necessarily mind the `_2` naming, especially if it looks like it will be addresses in API extractor. Ultimately though, if we need to do a breaking change, that's fine.

jasonnutter (changes requested):
A few comments and questions. Needs to be a breaking change for common if exports are renamed.

jasonnutter:
A couple other questions:

1. I don't see the API Extractor package installed anywhere. Did you install it globally on your computer? If so, I think it should be included as dev dependency so it can be run on CI.
2. Does pinning to this version address the name collision issue? https://github.com/angular/angular/commit/c42f35eb2f20bf5da13d141a9933755fd7808694

lib/msal-node/src/cache/NodeStorage.ts
sameerag: Similar reasons, added a different note for the reason below.

lib/msal-browser/src/event/EventMessage.ts
sameerag: Great! Can change this and `angular` dependency can be thrown out. Thanks.

lib/msal-common/src/index.ts
sameerag: Yes, they would.

sameerag:
> A couple other questions:
> 
> 1. I don't see the API Extractor package installed anywhere. Did you install it globally on your computer? If so, I think it should be included as dev dependency so it can be run on CI.

I am working on making this a part of CI, and every PR would need to consider generating this automatically and emit warnings if the public surface changes. I can see if I can include those changes here.

> 2. Does pinning to this version address the name collision issue? [angular/angular@c42f35e](https://github.com/angular/angular/commit/c42f35eb2f20bf5da13d141a9933755fd7808694)

Let me try this and update.

sameerag:
> 1. angular/angular@c42f35e

This did not help. I see a different issue, where the `CommonRequest` is not rolled up and a warning for export. I would want to do some R&D here, but not with this PR. Also I would say staying on the latest is more beneficial if we want to make this a standard across our libraries.

> 2. I don't see the API Extractor package installed anywhere. Did you install it globally on your computer? If so, I think it should be included as dev dependency so it can be run on CI.

Update: 
* `package.json` is now updated with `api-extractor` as a dev-dependency. 
* Added support for command: `npm run api:local` and `npm run api` for production and local runs for api-extractor report
* Removed `temp` folder and added it to `.gitignore`

lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -252,7 +277,7 @@ export abstract class ClientApplication {
 
     /**
      * Generates a request with the default scopes & generates a correlationId.
-     * @param authRequest
+     * @param authRequest - BaseAuthrequest for initialization
```
jo-arroyo, line 280:
```suggestion
     * @param authRequest - BaseAuthRequest for initialization
```

lib/msal-node/src/client/IConfidentialClientApplication.ts
```diff
@@ -11,15 +11,37 @@ import { OnBehalfOfRequest } from "../request/OnBehalfOfRequest";
...
 import { SilentFlowRequest } from "../request/SilentFlowRequest";
 import { TokenCache } from "../cache/TokenCache";
+
+/**
+ * Interface for the ConfidentialClientApplication class defining the public API signatures
+ * @public
+ */
 export interface IConfidentialClientApplication {
+
+    /** creates the URL of the authorization request */
```
jo-arroyo, line 21:
```suggestion
    /** Creates the URL of the authorization request */
```

lib/msal-node/src/client/IPublicClientApplication.ts
```diff
@@ -12,15 +12,37 @@ import { SilentFlowRequest } from "../request/SilentFlowRequest";
 import { UsernamePasswordRequest } from "../request/UsernamePasswordRequest";
 import { TokenCache } from "../cache/TokenCache";
 
+/**
+ * Interface for the PublicClientApplication class defining the public API signatures
+ * @public
+ */
 export interface IPublicClientApplication {
+
+    /** creates the URL of the authorization request */
```
jo-arroyo, line 21:
```suggestion
    /** Creates the URL of the authorization request */
```

lib/msal-node/src/cache/TokenCache.ts
```diff
@@ -237,6 +241,11 @@ export class TokenCache implements ISerializableTokenCache, ITokenCache {
         };
     }
 
+    /**
+     * helper to merge new cache with the old one
```
jo-arroyo, line 245:
```suggestion
     * Helper to merge new cache with the old one
```

lib/msal-node/src/cache/TokenCache.ts
```diff
@@ -247,6 +256,10 @@ export class TokenCache implements ISerializableTokenCache, ITokenCache {
         return finalState;
     }
 
+    /**
+     * helper to overlay as a part of cache merge
```
jo-arroyo, line 260:
```suggestion
     * Helper to overlay as a part of cache merge
```

lib/msal-node/src/client/ClientAssertion.ts
```diff
@@ -22,12 +23,22 @@ export class ClientAssertion {
     private jwtAudience: string;
     private publicCertificate: Array<string>;
 
+    /**
+     * initialize the ClientAssertion class from the clientAssertion passed by the user
```
jo-arroyo, line 27:
```suggestion
     * Initialize the ClientAssertion class from the clientAssertion passed by the user
```

lib/msal-node/src/client/ClientAssertion.ts
```diff
@@ -22,12 +23,22 @@ export class ClientAssertion {
...
+     * @param assertion - refer https://tools.ietf.org/html/rfc7521
+     */
     public static fromAssertion(assertion: string): ClientAssertion {
         const clientAssertion = new ClientAssertion();
         clientAssertion.jwt = assertion;
         return clientAssertion;
     }
 
+    /**
+     * initialize the ClientAssertion class from the certificate passed by the user
```
jo-arroyo, line 37:
```suggestion
     * Initialize the ClientAssertion class from the certificate passed by the user
```

lib/msal-node/src/client/ClientAssertion.ts
```diff
@@ -92,13 +111,16 @@ export class ClientAssertion {
         return this.jwt;
     }
 
+    /**
+     * utility API to check expiration
```
jo-arroyo, line 115:
```suggestion
     * Utility API to check expiration
```

lib/msal-node/src/cache/NodeStorage.ts
```diff
@@ -35,17 +36,24 @@ export class Storage extends CacheManager {
         this.logger = logger;
     }
 
+    /**
+     * queue up callbacks
```
jo-arroyo, line 40:
```suggestion
     * Queue up callbacks
```

jo-arroyo (approved):
Just a few nits for comment consistency

lib/msal-node/package.json
```diff
@@ -62,6 +62,8 @@
     "husky": "^4.2.3",
     "tsdx": "^0.14.1",
     "tslib": "^1.10.0",
+    "typescript": "^3.8.3",
```
jasonnutter, line 65: Duplicate

jasonnutter:
> > 1. I don't see the API Extractor package installed anywhere. Did you install it globally on your computer? If so, I think it should be included as dev dependency so it can be run on CI.
> 
> Update:
> 
> * `package.json` is now updated with `api-extractor` as a dev-dependency.
> * Added support for command: `npm run api:local` and `npm run api` for production and local runs for api-extractor report
> * Removed `temp` folder and added it to `.gitignore`

My understanding is that the `npm run api:local` command is something that would be run on every PR (e.g. as a pre-commit hook) to catch unintentional API changes before being merged, is that correct or no?

lib/msal-node/package.json
sameerag: Addressed

jasonnutter (approved):

pkanher617 (approved):

hectormmg (approved):

PR #2865, opened 2021-01-12
Fix initial state values in react hooks

### LINKED ISSUE #2846: useIsAuthenticated is always initially false
## Library

- [x] `@azure/msal-react@1.0.0-alpha.1`

## Framework

- [ ] Angular
- [x] React 16.14
- [ ] Other

## Description

When I use the `useIsAuthenticated` hook, on the first call the result is always false, even if I am actually logged in. This is because the `useEffect` call inside `useIsAuthenticated` isn't called until after the initial state has been returned. The initial state should probably be set to `isAuthenticated(allAccounts, account, matchAccount)` instead of `false`.

## Source

- [ ] Internal (Microsoft)
- [x] Customer request

tnorling:
We recently changed the initial state values inside the `msal-react` hooks to be static values in order to support SSR. This broke scenarios where the hook is rendered **after** the initial render of `MsalProvider` such as is the case when using `react-router-dom` or other router libraries.

To mitigate this problem this PR changes the initial state values to be a static value only if the `inProgress` value is `"startup"`. This will continue to support SSR since the `inProgress` value will always be `"startup"` when rendered on the server.

Also updates the React Router sample to use the router links that surfaced this problem.

Fixes #2846

### DIFF
--- a/change/@azure-msal-react-bd81c4ec-0359-4caf-ac48-3b6cc98fc388.json
+++ b/change/@azure-msal-react-bd81c4ec-0359-4caf-ac48-3b6cc98fc388.json
@@ -1,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "Fix initial state values (#2865)",
+  "packageName": "@azure/msal-react",
+  "email": "thomas.norling@microsoft.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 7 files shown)


jasonnutter (approved):

jo-arroyo (approved):
LGTM

pkanher617 (approved):

jonaspraem:
Doesn't seem to be fixed

PR #2911, opened 2021-01-20
Fix Error Handling in Broker Scenario

pkanher617:
This PR fixes the way errors are messaged in the broker flow by deconstructing the error object before sending it through the message channel.


lib/msal-browser/src/cache/BrowserCacheManager.ts
```diff
@@ -420,7 +420,7 @@ export class BrowserCacheManager extends CacheManager {
         const key = this.generateCacheKey(cacheKey);
 
         const value = this.internalStorage.getItem(key);
-        if (value) {
+        if (!value) {
```
tnorling, line 423: Yikes...nice catch

lib/msal-common/src/error/ClientAuthError.ts
```diff
@@ -169,10 +169,12 @@ export const ClientAuthErrorMessage = {
  * Error thrown when there is an error in the client code running on the browser.
  */
 export class ClientAuthError extends AuthError {
+    
+    static CLIENT_AUTH_ERR_NAME: string = "ClientAuthError";
```
tnorling, line 173: Can we put all these in an enum in Constants?

tnorling (approved):
One minor suggestion. Looks good otherwise

lib/msal-common/src/error/ClientAuthError.ts
pkanher617: I thought it would be easier to have them part of the error class instead (since we only reference them in that context). `Constants.ts` is more for generic strings used across the app.

lib/msal-common/src/error/AuthError.ts
```diff
@@ -20,6 +20,7 @@ export const AuthErrorMessage = {
  */
 export class AuthError extends Error {
 
+    static AUTH_ERROR_NAME: string = "AuthError";
```
jasonnutter, line 23: I would suggest just `Name` as the static field for all classes. `AuthError.AUTH_ERROR_NAME` is redundant.

jasonnutter (approved):
One comment otherwise looks good.

PR #2915, opened 2021-01-20
Update package SKU and version as part of release process

tnorling:
Adds a script that copies `name` and `version` properties from `package.json` into a new file `src/version.json` during the prepack phase of release. This allows us to use this data in the library without emitting package.json in the released package and resolves tsdx warnings about setting `rootDir` to `./src`


lib/msal-browser/src/version.json
```diff
@@ -0,0 +1 @@
+{"name":"@azure/msal-browser","version":"2.9.0"}
```
pkanher617, line 1: Is there a reason we can't just take the version from the package.json?

lib/msal-common/package.json
```diff
@@ -54,7 +54,8 @@
     "build:modules:watch": "rollup -cw",
     "build": "npm run clean && npm run build:modules",
     "build:all": "npm run build",
-    "prepack": "npm run build"
+    "prepack": "npm run updateVersion && npm run build",
+    "updateVersion": "node ../../release-scripts/updateVersion.js"
```
pkanher617, line 58: Would it be better to run this after `beachball:bump`? Or a combination of the two.

lib/msal-browser/src/version.json
tnorling: We shouldn't be emitting things in the build that live outside the root directory and some tools like tsdx require that the root directory is set to "./src". Though this isn't causing problems for browser today it is for other libs so the change here is mainly to keep all our libraries aligned

release-scripts/updateVersion.js
```diff
@@ -0,0 +1,17 @@
...
+const { name, version } = require(packageJsonPath);
+const fs = require("fs");
+
+const data = {
+    name: name,
+    version: version
+};
+
+const outputFilepath = path.resolve(process.cwd(), "src", "version.json");
+fs.writeFileSync(outputFilepath, JSON.stringify(data));
```
jasonnutter, line 17: Can you add some basic success and error handling?

lib/msal-common/package.json
tnorling: Done

release-scripts/updateVersion.js
tnorling: Done

jo-arroyo (approved):

jasonnutter (approved):

hectormmg (approved):

PR #2917, opened 2021-01-21
Documentation for mobile framework support

DarylThayil:
Add a doc to address questions we get around mobile support

### DIFF
--- a/docs/markdown/js-mobile-frameworks.md
+++ b/docs/markdown/js-mobile-frameworks.md
@@ -2,7 +2,7 @@
 The short answer is, no.  The more nuanced answer is that while we do not support these frameworks explicitly, there is no reason that our libraries can not work with these frameworks.
 
 #### Making it work
-If the libraries do not work out of the box you may need to write some glue code to adapt our libraries to the framework.  If in your investigation you find some parameters / features that would make working with these frameworks easier, we accept contributions to our libraries.  We will not make any large archtiectural changes to support this, but if we can accomodate we will.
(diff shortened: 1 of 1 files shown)


jasonnutter:
This might be better in the main README file, thoughts?

docs/markdown/js-mobile-frameworks.md
```diff
@@ -0,0 +1,8 @@
+### Does Msal's Javascript platform support mobile frameworks such as cordova, ionic, or react native. 
+The short answer is, no.  The more nuanced answer is that while we do not support these frameworks explicitly, there is no reason that our libraries can not work with these frameworks.
+
+#### Making it work
+If the libraries do not work out of the box you may need to write some glue code to adapt our libraries to the framework.  If in your investigation you find some parameters / features that would make working with these frameworks easier, we accept contributions to our libraries.  We will not make any large archtiectural changes to support this, but if we can accomodate we will.
```
hectormmg, line 5:
```suggestion
If the libraries do not work out of the box you may need to write some glue code to adapt our libraries to the framework.  If in your investigation you find some parameters / features that would make working with these frameworks easier, we accept contributions to our libraries.  We will not make any large architectural changes to support this, but if we can accommodate we will.
```

DarylThayil:
@jasonnutter  open to suggestions, maybe we can link from there but it wasn't clear where this belonged.

jasonnutter:
I would be up for a repo-level FAQ (which could also then link to FAQs for each library).

tnorling:
> I would be up for a repo-level FAQ (which could also then link to FAQs for each library).

+1 for this suggestion

PR #2936, opened 2021-01-25
Invoke interaction if MSAL Interceptor returns null access token

jasonnutter:
There is a known issue with the B2C service where it is only capable of issuing RTs that work for one resource, and calling the token endpoint with a mismatched RT/AT resource combo results in teh service returning a successful response with a null access token. To mitigate, the interceptor will invoke interaction, which will acquire a new RT for the relevant resource.


jo-arroyo (approved):
LGTM!

tnorling (approved):

Issue #2923, opened 2021-01-21
Property 'canLoad' in type 'MsalGuard' is not assignable to the same property in base type 'CanLoad'


DrJuriko:
Hello,

I wonder, am I the only one who have this problem?

## Library
- [x] "@azure/msal-angular": "^2.0.0-alpha.3",
- [x] "@azure/msal-browser": "^2.9.0",

## Framework
- [x] Angular
- [ ] React
- [ ] Other

## Description
Angular CLI: 9.1.13
Node: 12.18.3
OS: win32 x64

Angular: 9.1.13
... animations, cli, common, compiler, compiler-cli, core, forms
... language-service, platform-browser, platform-browser-dynamic
... router
Ivy Workspace: Yes

Package                            Version
------------------------------------------------------------
@angular-devkit/architect          0.901.13
@angular-devkit/build-angular      0.901.13
@angular-devkit/build-ng-packagr   0.901.13
@angular-devkit/build-optimizer    0.901.13
@angular-devkit/build-webpack      0.901.13
@angular-devkit/core               9.1.13
@angular-devkit/schematics         9.1.13
@angular/cdk                       9.2.4
@angular/material                  9.2.4
@ngtools/webpack                   9.1.13
@schematics/angular                9.1.13
@schematics/update                 0.901.13
ng-packagr                         9.1.5
rxjs                               6.5.5
typescript                         3.8.3
webpack                            4.42.0

## Error Message

    ERROR in node_modules/@azure/msal-angular/msal.guard.d.ts:29:5 - error TS2416: Property 'canLoad' in type 'MsalGuard' is not assignable to the same property in base type 'CanLoad'.
      Type '() => Observable<boolean | UrlTree>' is not assignable to type '(route: Route, segments: UrlSegment[]) => boolean | Promise<boolean> | Observable<boolean>'.
        Type 'Observable<boolean | UrlTree>' is not assignable to type 'boolean | Promise<boolean> | Observable<boolean>'.
          Type 'Observable<boolean | UrlTree>' is not assignable to type 'Observable<boolean>'.
            Type 'boolean | UrlTree' is not assignable to type 'boolean'.
              Type 'UrlTree' is not assignable to type 'boolean'.
    
    29     canLoad(): Observable<boolean | UrlTree>;

## Reproduction steps
I just tried to serve my test app:

ng serve my-app

[...]

PR #2948, opened 2021-01-28
[msal-angular] Update guard with version detection for canLoad

jo-arroyo:
This PR updates the MSAL Guard to check the version of Angular used by applications to determine whether to return a `loginFailedRoute` or boolean in the `canLoad` interface. 

This PR addresses issue #2923.

### DIFF
--- a/lib/msal-angular/src/msal.guard.ts
+++ b/lib/msal-angular/src/msal.guard.ts
@@ -5,7 +5,7 @@
 
 import { CanActivate, ActivatedRouteSnapshot, RouterStateSnapshot, CanActivateChild, CanLoad, UrlTree, Router } from "@angular/router";
 import { MsalService } from "./msal.service";
-import { Injectable, Inject } from "@angular/core";
+import { Injectable, Inject, VERSION } from "@angular/core";
 import { Location } from "@angular/common";
 import { InteractionType, BrowserConfigurationAuthError, BrowserUtils, UrlString, PopupRequest, RedirectRequest, AuthenticationResult } from "@azure/msal-browser";
 import { MsalGuardConfiguration } from "./msal.guard.config";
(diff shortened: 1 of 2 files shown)


jasonnutter (approved):
Looks good! Curious if there is an easy way to test this (might be best tested with e2e tests).

lib/msal-angular/src/msal.guard.ts
```diff
@@ -139,8 +143,9 @@ export class MsalGuard implements CanActivate, CanActivateChild, CanLoad {
         return this.activateHelper(state);
     }
 
-    canLoad(): Observable<boolean|UrlTree> {
+    canLoad(): Observable<boolean> {
         this.authService.getLogger().verbose("Guard - canLoad");
+        // @ts-ignore
```
tnorling, line 148: Instead of ignoring the compiler errors can we force the return of boolean on the return? 
`return !!this.activateHelper()`

jasonnutter: No, because we want the routetree returned for Angular 10 and up.

jasonnutter: We're just telling Angular that it will be a boolean, to allow it to have the function signature that is compatible with Angular 9.

tnorling: hmm ok, not the biggest fan of using ts-ignore but if it's unavoidable I suppose it's fine for now

tnorling (approved):

anuj-shukla:
When will this update be available as I'm facing this issue when trying to use alpha-3 package.

jo-arroyo:
@anuj-shukla The latest versions of msal-browser and msal-angular were released last night. Please update using `npm install @azure/msal-browser@2.11.0 @azure/msal-angular@alpha` and give the new version a try.

PR #2949, opened 2021-01-28
Fix unable to change B2C policy

tnorling:
The authority changes in `msal-common` version 3.0.0 caused a regression. Since we now cache the authorization endpoint, token endpoint, etc. and subsequent requests do a cache lookup based on domain, switching B2C policies broke (and potentially switching tenants). In a scenario where you attempt to use a sign-in policy and then switch to the reset password policy, the cached sign-in policy will be used because the domain stayed the same.

This PR fixes this by adding the following:
- Add `canonical_authority` to the cached metadata representing the authority url used to generate the cache entry
- If the requested authority and cached `canonical_authority` have different url lengths, skip cache and go to the network. 
i.e. `login.microsoftonline.com/common` vs `login.microsoftonline.com/tfp/b2clogin.com/susi` should be treated as completely different
- If the authorization/token/end_session/etc. endpoints contain a segment also contained in the cached `canonical_authority` but not contained in the requested authority, replace the mismatched segment.

For example: 
- Authorization Endpoint: `https://login.microsoftonline.com/te/msidlabb2c.onmicrosoft.com/b2c_1_sisopolicy/oauth2/authorize`
- Cached `canonical_authority: "https://login.microsoftonline.com/te/msidlabb2c.onmicrosoft.com/b2c_1_sisopolicy/"` 
- Requested authority `https://login.microsoftonline.com/te/msidlabb2c.onmicrosoft.com/b2c_1_password_reset/`

Here `b2c_1_sisopolicy` is in both the cached authority and authorization url but does not equal `b2c_1_password_reset` so we replace that segment to form `https://login.microsoftonline.com/te/msidlabb2c.onmicrosoft.com/b2c_1_password_reset/oauth2/authorize`


pkanher617 (approved):

sameerag:
What happens for `login.microsoftonline.com/common` vs `login.microsoftonline.com/<tenant_id>` if this is the only difference between configuration and request? Do we update the cache with the `tenant_id`? or do we go to network every single time?

tnorling:
> What happens for `login.microsoftonline.com/common` vs `login.microsoftonline.com/<tenant_id>` if this is the only difference between configuration and request? Do we update the cache with the `tenant_id`? or do we go to network every single time?

We don't update the cache (even in policy scenarios) but this fix will replace the tenant as well, so no network call.

sameerag (approved):

sameerag:
As discussed, lets pick up the more robust fix in the upcoming weeks!

PR #2952, opened 2021-01-29
Fix token timestamp calculation

### LINKED ISSUE #2921: msal-browser caches tokens with incorrectly-calculated expiry, causing expired tokens to be returned to client
## Library

- (Verified in `msal-browser`, but likely affects all)

## Framework

- [x] Angular
- [x] React
- (Affects all)

## Description
If no `libraryState` is provided to `generateCacheRecord`, a fallback timestamp is created and expiration of the cache entry is based on that. This is problematic because the fallback timestamp is created _after_ the token has been requested and received from the server.

If code execution is suspended for any reason (this can happen if a laptop is closed, the tab is sent to the background to save on CPU, etc) after the token request is made and before the fallback timestamp is generated, then when code execution is later resumed, the current time will be used to generate the fallback timestamp upon which the cache entry expiration is calculated. When this happens, MSAL will consider the newly-cached token to be valid for its full `expires_in` duration, when in fact, the token could already be expired. MSAL then returns the stale token in response to acquireTokenSilent until its incorrectly-calculated expiration has passed. 

https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/ff607022fd1982b281fa81247bf29815f907f511/lib/msal-common/src/response/ResponseHandler.ts#L114
https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/ff607022fd1982b281fa81247bf29815f907f511/lib/msal-common/src/response/ResponseHandler.ts#L141
https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/ff607022fd1982b281fa81247bf29815f907f511/lib/msal-common/src/response/ResponseHandler.ts#L218

## Error Message

Stale token is returned via `acquireTokenSilent`, and continues to be returned until `expires_in` from the time it was cached by MSAL.

## MSAL Configuration

```js
```

## Reproduction steps
1. Acquire a token, pause code execution (simulating tab sent to background or laptop closed) after token is received and before fallback timestamp is created
2. Wait for `expires_in` seconds
3. Resume code execution

## Expected behavior
To solve this problem in ADAL, expiration timestamps were calculated and included on token request state. After the response came back, the state's timestamp was read to calculate cache expiration. 

The same should be done in MSAL. 

Potential solution 1: Update the `/token` endpoint to support responding with the `ts` of request state, or returning the entire request state. Update `handleServerTokenResponse` to base the expiration timestamp on `ts` instead of falling back to a new timestamp.

Potential solution 2: An alternate approach could avoid updating the `/token` endpoint by instead ensuring every type of request has created request state (perhaps in `initializeBaseRequest`)(containing timestamp), and passing this state through to `handleServerTokenResponse`, which would consume the timestamp of the state directly and have no need to fall back to a new timestamp.

## Identity Provider

- [x] Azure AD
- [x] Azure B2C Basic Policy
- [x] Azure B2C Custom Policy
- [x] ADFS
- (Affects all)

## Browsers/Environment

- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari
- [x] IE
- [x] Other (All)

## Regression

- [x] Did this behavior work before?
Version: Worked in ADAL after a fix, but never worked in MSAL to my knowledge, expect for the authorization code flows.

## Security

## Source

- [x] Internal (Microsoft)
- [ ] Customer request

pkanher617:
The library currently uses the timestamp from the state parameter to calculate the baseline for token lifetime. This causes two issues:

- token lifetime for access token renewal through refresh token can be thrown off if execution is paused after the POST request is made, but before the token is retrieved (see #2921)
- token lifetimes can be off if the POST request is not made immediately after the code has been retrieved from the /authorize call. For example, if the user takes a long time to enter their credentials.

This PR fixes this by calculating timestamp directly before a POST request is made, and using this timestamp in the ResponseHandler.ts.

Should fix #2921.


sameerag (approved):
Lgtm.

jasonnutter:
What about for interactive calls? Namely redirect, where the we lose page state. Seems like timestamp in state should be used for that.

jessemartin:
Thanks @pkanher617!

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -211,11 +211,8 @@ export class ResponseHandler {
             // If scopes not returned in server response, use request scopes
             const responseScopes = serverTokenResponse.scope ? ScopeSet.fromString(serverTokenResponse.scope) : new ScopeSet(requestScopes || []);
 
-            // Expiration calculation
-            const currentTime = TimeUtils.nowSeconds();
-
             // If the request timestamp was sent in the library state, use that timestamp to calculate expiration. Otherwise, use current time.
```
tnorling, line 214: nit: update this comment

tnorling (approved):

jo-arroyo (approved):

lib/msal-common/src/response/ResponseHandler.ts

PR #2953, opened 2021-01-29
ssoSilent should leverage cache credentials if possible

### LINKED ISSUE #2925: MSAL Browser: ssoSilent should leverage cached auth state if possible
## Library

## Source

- [x] Internal (Microsoft)
- [ ] Customer request

## Description

Today, the `ssoSilent` API will always attempt to invoke a network request, regardless of current cache state. This means that a network request may be made when it is not necessary, which results in a needless perf penalty.

## Priority

Medium

## Acceptance Criteria

1. If the application has valid auth state in the cache (valid id token, refresh token), invoking ssoSilent should not result in a network request, by default.
2. Setting `forceRefresh` to `true` should result in a network request, like with `acquireTokenSilent`.

jasonnutter:
Today, the `ssoSilent` API always attempts a network request in a hidden iframe, regardless of cache state. This results in an unnecessary performance penalty if there are tokens in the cache. This PR changes `ssoSilent` to call `acquireTokenSilent`, which will check for cached tokens, and invoke `acquireTokenByIframe` if a new refresh token is needed. Passing `forceRefresh` will bypassing this caching logic, similar to `acquireTokenSilent`.

Also adds the convience method `getAccountBySessionId`, which `ssoSilent` will use to check if there is a cached account for the given session id.


lib/msal-browser/src/request/SsoSilentRequest.ts
```diff
@@ -25,5 +25,8 @@ import { AuthorizationUrlRequest } from "@azure/msal-common";
  * - domainHint                 - Provides a hint about the tenant or domain that the user should use to sign in. The value of the domain hint is a registered domain for the tenant.
  * - extraQueryParameters       - String to string map of custom query parameters.
  * - nonce                      - A value included in the request that is returned in the id token. A randomly generated unique value is typically used to mitigate replay attacks.
+ * - forceRefresh               - Force a network request, instead of leveraging cached credentials. Defaults to false.
```
jasonnutter, line 28: An argument could be made that this should default to `true`, which is the current behavior. I personally think the semantics of `forceRefresh` should align with the other libraries, regardless of the current behavior of `ssoSilent`, as most apps will welcome the improved default performance of `ssoSilent`.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -69,6 +70,85 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
+        this.preflightBrowserEnvironmentCheck(InteractionType.Silent);
+        this.emitEvent(EventType.SSO_SILENT_START, InteractionType.Silent, request);
+
+        try {
+            // Build cached account by loginHint, sid, or the request itself
+            const accountByLoginHint = request.loginHint && this.getAccountByUsername(request.loginHint);
+            
+            const accountBySid = request.sid && this.getAccountBySessionId(request.sid);
+            if (request.sid && !accountBySid) {
+                this.logger.warning("ssoSilent - sid provided to ssoSilent, but no cached account found. If this is unexpected, confirm sid is enabled as an optional id token claim for your application");
```
jasonnutter, line 98: This message should probably check `forceRefresh`, as its not needed if were bypassing the cache.

pkanher617:
How is this different from the way acquireTokenSilent works today?

jasonnutter:
> How is this different from the way acquireTokenSilent works today?

The API is different (account is required, or an active account), and `acquireTokenSilent` doesn't work if the cache is empty (no_tokens_found error).

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -676,6 +647,22 @@ export abstract class ClientApplication {
...
+    /**
+     * Returns the signed in account matching the given session id,
+     * or null when no matching account is found.
+     * Note: sid must be enabled as an optional id token claim.
+     * @param sid Session ID (optional id token claim)
+     * @returns {@link AccountInfo} - the account object stored in MSAL
+     */
+    public getAccountBySessionId(sid: string): AccountInfo | null {
+        const allAccounts = this.getAllAccounts();
+        if (sid && allAccounts && allAccounts.length) {
```
tnorling, line 659: `getAllAccounts` should always return type array, length check should be enough
```suggestion
        if (sid && allAccounts.length) {
```

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -69,6 +70,85 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
+            }
+
+            const account = request.account || accountByLoginHint || accountBySid;
```
tnorling, line 101: imo `sid` should come before `loginHint`
```suggestion
            const account = request.account || accountBySid || accountByLoginHint;
```

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -69,6 +70,85 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
+
+            // Only checked for cached credentials if forceRefresh is off and there is an account for the request
+            if (!request.forceRefresh && account) {
```
tnorling, line 104: forceRefresh could also mean acquire using a refresh token, right? Might not need to check it here as `acquireTokenSilent` should handle that logic

lib/msal-browser/src/request/SsoSilentRequest.ts
tnorling: I'm fine with defaulting to false and just mentioning it in the changelog as "potentially breaking"

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -69,6 +70,85 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
+     */
+    async ssoSilent(request: SsoSilentRequest): Promise<AuthenticationResult> {
+        this.preflightBrowserEnvironmentCheck(InteractionType.Silent);
+        this.emitEvent(EventType.SSO_SILENT_START, InteractionType.Silent, request);
+
+        try {
+            // Build cached account by loginHint, sid, or the request itself
+            const accountByLoginHint = request.loginHint && this.getAccountByUsername(request.loginHint);
+            
+            const accountBySid = request.sid && this.getAccountBySessionId(request.sid);
```
tnorling, line 96: Are we requiring that one of `loginHint`, `sid` or `account` are passed? If so should we check that here? If not should we attempt to fallback to `getActiveAccount`?

lib/msal-browser/src/app/ClientApplication.ts
jasonnutter: Indeed. This was copy pasta from one of the other `getAccountBy` methods, so I'll update those accordingly as well.

lib/msal-browser/src/app/PublicClientApplication.ts
jasonnutter: Yes, and that will check will get made downstream: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/src/app/ClientApplication.ts#L483

jasonnutter: And yeah, probably good to fallback to `getActiveAccount`.

lib/msal-browser/src/app/PublicClientApplication.ts
jasonnutter: Agreed.

lib/msal-browser/src/app/PublicClientApplication.ts
jasonnutter: I figured it would be best to avoid `acquireTokenSilent` completely if `forceRefresh` is on, as it would semantically mean "give me a new set of tokens, including a new RT", as `forceRefresh` for `acquireTokenSilent` will still use the existing RT.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -69,6 +70,85 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
+     *
+     * @returns {Promise.<AuthenticationResult>} - a promise that is fulfilled when this function has completed, or rejected if an error was raised. Returns the {@link AuthResponse} object
+     */
+    async ssoSilent(request: SsoSilentRequest): Promise<AuthenticationResult> {
+        this.preflightBrowserEnvironmentCheck(InteractionType.Silent);
+        this.emitEvent(EventType.SSO_SILENT_START, InteractionType.Silent, request);
+
+        try {
+            // Build cached account by loginHint, sid, or the request itself
+            const accountByLoginHint = request.loginHint && this.getAccountByUsername(request.loginHint);
```
sameerag, line 94: `sid` should come before `loginHint` as per the eSTS agreement.

lib/msal-browser/src/app/PublicClientApplication.ts
sameerag: +1 for fallback

lib/msal-browser/src/app/PublicClientApplication.ts
sameerag: Just FYI rule of thumb: `user passed account` > `cached account` and `sid` > `loginHint` in any account read.

lib/msal-browser/src/app/PublicClientApplication.ts
sameerag: The `forceRefresh` across apps mean only `AT` refresh and not necessarily `RT` considering long lived RT support for all other MSALs. I would leave up to you to interpret that in the way you see fit for `ssoSilent` but this may cause confusion for users when used against both these APIs. (`ssoSilent` and `acquireTokenSilent`) differently.

lib/msal-browser/src/request/SsoSilentRequest.ts
jasonnutter: Thought about this over the weekend, and I think I will default to true, to maintain existing behavior, which apps may be relying on. Will document this and denote this behavior will likely change in future major version.

lib/msal-browser/src/app/PublicClientApplication.ts
jasonnutter: Actually, I do not think we should use getActiveAccount, as ssoSilent could be used to switch between accounts, and so we should only use account hints that are explicitly provided in the request.

lib/msal-browser/src/app/PublicClientApplication.ts
jasonnutter: Is there a precedent for a flag to force an RT refresh?

jasonnutter: I could see an argument for a different flag if we wanted to support the ability to use the cached RT but want to force getting a new AT (which isn't possible, with how I've designed it).

jasonnutter: .NET has `forceLogin` (which is for setting `prompt=login`): https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/search?q=forcelogin

Would that be acceptable here?

jasonnutter: `skipCache` could also work.

jasonnutter: A bit literal, but arguably less confusing/overloaded than `forceRefresh` or `forceLogin`.

jasonnutter: Decided to go with `skipCache`, defaulting to `true`.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -69,6 +70,90 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
+                    .catch(error => {
+                        /*
+                         * This error is returned when there is an empty cache. 
+                         * In this scenario, fall through to iframed request.
+                         */
+                        if (error.errorCode === ClientAuthErrorMessage.noTokensFoundError.code) {
+                            return null;
+                        }
+
+                        throw error;
```
tnorling, line 128: Might have to take another look at this later for telemetry purposes (it can be an incremental update)

jasonnutter:
Closing, as we decided that apps should invoke `acquireTokenSilent` first, and then handle `no_tokens_found` with `ssoSilent`.

Issue #2927, opened 2021-01-22
msal-react acquiring tokens prior to API calls


richardrobberse:
## Library
- [X] `@azure/msal-react@1.0.0-alpha.3`

## Description
I'm facing some issues with the way tokens have to be acquired in msal-react.

In short: I have all my async data fetch logic written in Redux actions (thunks to be more precise), my components only dispatch actions to trigger the data fetch and the data flows in once it's retrieved from the server.
Those async actions require a valid access token and as stated in the docs I should retrieve a fresh token for every API call, meaning that I'd have to bloat every component with access token logic and pass the the token to the action that fetches the data.
I could put this logic in a hook and call the hook in every component, but the logic that acquires the token (`IPublicClientApplication.acquireTokenSilent`) is async, meaning that I'd have to return some kind of promise or state object from the hook and wait for it to complete, before I can fetch my data.

Ideally I'd like to have a valid token in my Redux store at all times and keep my components "clean" from token logic OR have the ability to retrieve a token outside of a React component (but this won't be possible since it's based on React Context)
Any ideas on how to achieve a clean and scalable solution?

Thanks in advance!

## Source

- [ ] Internal (Microsoft)
- [X] Customer request

tnorling:
@richardrobberse I'm not familiar enough with Redux 

> I could put this logic in a hook and call the hook in every component, but the logic that acquires the token (IPublicClientApplication.acquireTokenSilent) is async, meaning that I'd have to return some kind of promise or state object from the hook and wait for it to complete, before I can fetch my data.

Every solution will require that you wait for `acquireTokenSilent` to complete before fetching your data. From a performance perspective this should not be a big deal because `acquireTokenSilent` will first attempt to retrieve a valid token from local/session/in-memory storage so in most cases the promise will resolve almost immediately. This is typically the recommended approach, can you clarify what about it does not work for you?

> have the ability to retrieve a token outside of a React component (but this won't be possible since it's based on React Context)

You actually can do this using the `PublicClientApplication` instance you pass into the `MsalProvider` or by creating a new one with the same configuration options. The only things you cannot access from outside the React Context are the components and hooks exposed by `msal-react`. For APIs like `acquireTokenSilent` and `getAllAccounts` this should be fine to do, calling login APIs outside the context, however, may lead to some issues updating state inside the context and would not be recommended.

richardrobberse:
@tnorling 
[...]

PR #2960, opened 2021-01-31
Added section about how to acquire an access token outside of a React component

richardrobberse:
@tnorling Here's a proposal based on what we've discussed here #2927, hope this is a valuable addition.

This is my first contribution to this repository, I've followed the contributing docs so please let me know if I should do things differently.

### DIFF
--- a/lib/msal-react/docs/getting-started.md
+++ b/lib/msal-react/docs/getting-started.md
@@ -226,3 +226,35 @@ export function App() {
     }
 }
 ```
+
+## Acquiring an access token outside of a React component
+
+If you require an access token outside of a React component you can directly call the `acquireTokenSilent` function on the `PublicClientApplication`.
+The `PublicClientApplication` can be reused but might as well be recreated for every API call. 
+We do not recommend calling functions that change the user's state outside of a component because those functions rely on the React Context API to work properly.
+
+Keep in mind that the user has to be signed in before you can acquire a token outside of a React component.
+
+```javascript
+import { Configuration, PublicClientApplication } from '@azure/msal-browser'
+
+const acquireAccessToken = async (): Promise<string> => {
+  const configuration: Configuration = {
+    auth: {
+      clientId: "client-id"
+      redirectUri: "redirect uri",
+    },
+  }
+
+  const pca = new PublicClientApplication(cfg)
+  const accounts = pca.getAllAccounts()
+
+  const authResult = await pca.acquireTokenSilent({
+    scopes: ["User.Read"],
+    accounts[0],
+  })
+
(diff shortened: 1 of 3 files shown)


lib/msal-react/docs/getting-started.md
```diff
@@ -226,3 +226,35 @@ export function App() {
...
+      redirectUri: "redirect uri",
+    },
+  }
+
+  const pca = new PublicClientApplication(cfg)
+  const accounts = pca.getAllAccounts()
+
+  const authResult = await pca.acquireTokenSilent({
+    scopes: ["User.Read"],
+    accounts[0],
```
jasonnutter, line 254:
```suggestion
    account: accounts[0],
```

jasonnutter: Or, you can show using the `setActiveAccountApi` (which can be used to set the account after login).

lib/msal-react/docs/getting-started.md
```diff
@@ -226,3 +226,35 @@ export function App() {
...
+If you require an access token outside of a React component you can directly call the `acquireTokenSilent` function on the `PublicClientApplication`.
+The `PublicClientApplication` can be reused but might as well be recreated for every API call. 
+We do not recommend calling functions that change the user's state outside of a component because those functions rely on the React Context API to work properly.
+
+Keep in mind that the user has to be signed in before you can acquire a token outside of a React component.
+
+```javascript
+import { Configuration, PublicClientApplication } from '@azure/msal-browser'
+
+const acquireAccessToken = async (): Promise<string> => {
```
tnorling, line 241: With regards to my comment above, documentation should demonstrate passing an existing pca instance to the function rather than instantiating a new one
```suggestion
const acquireAccessToken = async (pca: PublicClientApplication): Promise<string> => {
```

lib/msal-react/docs/getting-started.md
```diff
@@ -226,3 +226,35 @@ export function App() {
...
+    auth: {
+      clientId: "client-id"
+      redirectUri: "redirect uri",
+    },
+  }
+
+  const pca = new PublicClientApplication(cfg)
+  const accounts = pca.getAllAccounts()
+
+  const authResult = await pca.acquireTokenSilent({
```
tnorling, line 252: Wrap this call in an if/else checking for `accounts.length > 0` and potentially throwing an error if no accounts are signed in

tnorling (changes requested):

tnorling:
I've pulled this PR into #3002 so I can resolve the comments and merge. Thanks again for your contribution!

PR #2961, opened 2021-01-31
Check useAccount() for undefined accounts[0] value

olafwrieden:
Added `{}` as default value in `useAccount()` because an error is otherwise thrown on first render due to `accounts[0]` being undefined until a successful authentication has been achieved. See #2773 where this issue has been reported.


tnorling (commented):
Thanks for your contribution! Could you please run `npm run beachball:change` in the root directory and go through the prompts to generate the change files? You can select change type "None".

Thanks!

PR #2965, opened 2021-02-03
msal-react FAQ and class component doc

tnorling:
Adds an initial FAQ for `msal-react` + a doc for use with class components. Closes #2933

Issue #2933, opened 2021-01-25
Does MSAL-React support class components?


riktimmondal:
## Library

- [ ] `msal@1.x.x` or `@azure/msal@1.x.x`
- [ ] `@azure/msal-browser@2.x.x`
- [ ] `@azure/msal-node@1.x.x`
- [X] `@azure/msal-react@1.0.0-alpha.3`
- [ ] `@azure/msal-angular@0.x.x`
- [ ] `@azure/msal-angular@1.x.x`
- [ ] `@azure/msal-angular@2.x.x`
- [ ] `@azure/msal-angularjs@1.x.x`

## Description

Am using https://www.npmjs.com/package/@azure/msal-react in my react project. The library provides hooks to perform auth very easliy.

So in functional component, for getting access token,
```
if (account && inProgress === "none") {
instance.acquireTokenSilent({
...loginRequest,
account: account
}).then((response) => {
callMsGraph(response.accessToken).then(response => setGraphData(response));
});
}
```
needs to be called where ```const { instance, accounts, inProgress } = useMsal();``` is used.

But I need to call api to fetch data from class components.
So, how to achieve the same functionality in class component
## Source

- [X] Internal (Microsoft)
- [ ] Customer request

tnorling:
@riktimmondal When using class components you have 2 options: 

1. Use the raw context as shown [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-react/docs/getting-started.md#consuming-the-raw-context)
2. Use the `withMsal` higher order component to wrap your class component and inject the context into its props. I see now that this approach has not yet been documented, I will make sure to add this to our docs.

PR #2968, opened 2021-02-03
[msal-react] adding new sample with class components

derisen:
This PR adds a new sample to msal-react demonstrating the usage of class components.

This is a conversion of the current [msal-react quickstart sample](https://github.com/Azure-Samples/ms-identity-javascript-react-spa).

Refers to: #2965
Closes: [#2933](https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/2933)


tnorling:
@derisen I was thinking about this last night and I think actually what I would suggest instead of a standalone sample is to add a page to the existing [react-router-sample](https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev/samples/msal-react-samples/react-router-sample) that demonstrates using a class component. I can work on porting your changes here over to that sample

derisen:
@tnorling aah, that makes sense too (I've missed that sample). Ok then, I'll leave the decision to you. I can certainly add a new page there with classes (a 'mails' page perhaps?)

tnorling:
> @tnorling aah, that makes sense too (I've missed that sample). Ok then, I'll leave the decision to you. I can certainly add a new page there with classes (a 'mails' page perhaps?)

@derisen I'd suggest just duplicating the /profile page but re-implement as a class component so that it's easy to compare the 2. Our microsoft accounts don't have consent to the Mail.Read scope making testing more challenging for the mail stuff unfortunately.

derisen:
> > @tnorling aah, that makes sense too (I've missed that sample). Ok then, I'll leave the decision to you. I can certainly add a new page there with classes (a 'mails' page perhaps?)
> 
> @derisen I'd suggest just duplicating the /profile page but re-implement as a class component so that it's easy to compare the 2. Our microsoft accounts don't have consent to the Mail.Read scope making testing more challenging for the mail stuff unfortunately.

Sounds good!

derisen:
Closed by #2973

Issue #2195, opened 2020-08-26
Incorrect access token for API scope when using a personal Microsoft account


jviney:
## Library
- [x] `@azure/msal-browser@2.1.0`
- [x] `@azure/msal-common@1.2.0`

## Framework

AngularJS 1.6.1

## Description

Setup:

* Frontend SPA (AngularJS) and a backend. Both access Graph APIs on behalf of the user.
* 1 Azure app.
* API scope under "Expose an API" in Azure (eg: api://1ae23a84-215a-4bd5-ae99-d54f1b7b211a/access_as_user).

In the browser, login with:

```js
msal.acquireTokenPopup({
  prompt: 'select_account',
  redirectUri: "https://local.com/redirect_microsoft.html",
  scopes: ["profile", "email", "openid", "offline_access", "user.read", "files.readwrite.all"],

  // Get consent for this scope, but it won't be included in the access token
  extraScopesToConsent: ["api://1ae23a84-215a-4bd5-ae99-d54f1b7b211a/access_as_user"]
})
```

This works correctly so far for personal and business Microsoft accounts.

Next, try to get an access token for the API scope. The returned token will be passed to our API, which uses the OBO flow to exchange it for its own access and refresh token. We store the refresh token for future use.

```js
msal.acquireTokenSilent({
  scopes: ["api://1ae23a84-215a-4bd5-ae99-d54f1b7b211a/access_as_user"],
  account: account // from the result of acquireTokenPopup()
})
```

In `@azure/msal-browser` 2.0.2 this step would fail for personal Microsoft accounts with the error `ClientAuthError: null_or_empty_id_token: The idToken is null or empty`. It succeeds in 2.1.0 even though the id token is not present. It works for non-personal Microsoft accounts in both versions, and includes an id token. I don't know why personal accounts do not get an id token here.

The access token is then passed to our API to perform the OBO flow. It works for non-personal Microsoft accounts, but fails for personal accounts with the error:

```
{
    "error": "invalid_request",
    "error_description": "AADSTS90023: Cross-origin token redemption is permitted only for the 'Single-Page Application' client-type.\r\nTrace ID: 32db7dc0-ab90-4a03-94bc-69522f919100\r\nCorrelation ID: 3b10fa8d-0527-4cad-a459-206f4086a731\r\nTimestamp: 2020-08-26 07:27:14Z",
    "error_codes": [
        90023
    ],
    "timestamp": "2020-08-26 07:27:14Z",
[...]

PR #2969, opened 2021-02-03
Ignore OIDC scopes in cache logic

pkanher617:
This PR fixes an issue identified in #2195 where the MSAL.js SDK should be adding the `offline_access` scope to all calls. It also adds logic to ignore all OIDC scopes when looking up or replacing access tokens.


tnorling (approved):

sameerag (approved):
Lgtm

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -803,8 +803,8 @@ export abstract class CacheManager implements ICacheManager {
         const entityScopeSet: ScopeSet = ScopeSet.fromString(entity.target);
         const requestTargetScopeSet: ScopeSet = ScopeSet.fromString(target);
 
-        if (!requestTargetScopeSet.containsOnlyDefaultScopes()) {
-            requestTargetScopeSet.removeDefaultScopes(); // ignore default scopes
+        if (!requestTargetScopeSet.containsOnlyOIDCScopes()) {
+            requestTargetScopeSet.removeOIDCScopes(); // ignore OIDC scopes
```
hpsin, line 807: I'm hoping that the containsOnly part will still support userinfo tokens, which are tokens requested with scope=openid. Will this cache those tokens, issued with scope=openid [profile email]

lib/msal-common/src/utils/Constants.ts
```diff
@@ -37,9 +38,20 @@ export const Constants = {
...
     NOT_DEFINED: "not_defined",
     EMPTY_STRING: "",
-    FORWARD_SLASH: "/" 
+    FORWARD_SLASH: "/"
 };
 
+export const OIDC_DEFAULT_SCOPES = [
+    Constants.OPENID_SCOPE,
+    Constants.PROFILE_SCOPE,
+    Constants.OFFLINE_ACCESS_SCOPE
```
hpsin, line 47: It's a breaking change to behavior, but adding email here would be _really_ nice. I've never understood why MSAL doesn't request the user's email address.

lib/msal-common/src/utils/Constants.ts
```diff
@@ -37,9 +38,20 @@ export const Constants = {
...
+];
+
+export const OIDC_SCOPES = [
+    ...OIDC_DEFAULT_SCOPES,
+    Constants.EMAIL_SCOPE
```
hpsin, line 52: Idea for the future, to make this more standards based.  These values can be sourced from the authority config file:

https://login.microsoftonline.com/microsoft.com/v2.0/.well-known/openid-configuration

,"scopes_supported":["openid","profile","email","offline_access"],

lib/msal-common/src/cache/CacheManager.ts
hpsin: Nevermind, I see it in the tests below, seems like.

pkanher617: Yes, this will only remove OIDC scopes if they are included with non-OIDC scopes.

lib/msal-common/src/utils/Constants.ts
pkanher617: Will look at adding this in the future, thanks for the suggestion

lib/msal-common/src/utils/Constants.ts
pkanher617: @sameerag @jasonnutter possibly something to add to next major version?

jasonnutter: Why would this be breaking?

jasonnutter: Seems additive/incremental.

jasonnutter:
Does MSAL Node need to be updated to also use this logic?

lib/msal-common/src/utils/Constants.ts
hpsin: Adding the email claim out of no where will slightly change the consent screen, and would trigger new consent prompts for existing apps.

pkanher617:
> Does MSAL Node need to be updated to also use this logic?

Thanks for catching this, `msal-node` was already sending all three scopes, I updated the code to use the object from common.

lib/msal-common/src/utils/Constants.ts
jasonnutter: Ah, makes sense.

jasonnutter: > @sameerag @jasonnutter possibly something to add to next major version?

Yeah, worth noting.

jasonnutter (approved):

PR #2971, opened 2021-02-03
msal-angular update documentation: add ssoSilent and remove cors-api

jo-arroyo:
This PR: 
- Adds documentation on how to use `ssoSilent()` with msal-angular
- Removes the `cors-api` documentation


lib/msal-angular/docs/v2-docs/ssosilent.md
```diff
@@ -0,0 +1,54 @@
+# Silent login with ssoSilent()
+
+If you already have a session that exists with the authentication server, you can use the ssoSilent() API to make request for tokens without interaction. You will need to pass a `login_hint` (which can be retrieved from the account object `username` property or the `upn` claim in the ID token) in the request object in order to successfully obtain a token silently.
```
tnorling, line 3:
```suggestion
If you already have a session that exists with the authentication server, you can use the ssoSilent() API to make request for tokens without interaction. You will need to pass a `loginHint` (which can be retrieved from the account object `username` property or the `upn` claim in the ID token) in the request object in order to successfully obtain a token silently.
```

Let's also callout that the `loginHint` can also come from the `emails` claim for B2C use cases and that you could alternatively use `sid` instead for AAD use cases.

lib/msal-angular/docs/v2-docs/ssosilent.md
```diff
@@ -0,0 +1,54 @@
...
+};
+
+this.authService.ssoSilent(silentRequest)
+    .subscribe({
+        next: (result) => console.log("Success!"), // Handle result
+        error: (error) => console.log(error) // Handle error
+    });
+```
+
+If ssoSilent() fails, it will throw an `interaction_required` error. We recommend handling this error by logging in interactively. Here is an example of ssoSilent() being used in an application's `app.component.ts`:
```
tnorling, line 18:
```suggestion
If ssoSilent() fails, we recommend handling this error by logging in interactively. Here is an example of ssoSilent() being used in an application's `app.component.ts`:
```

There may be scenarios where it's not an `interaction_required` error, imo we should fallback to interactive for any error (which is also what the example shows below)

lib/msal-angular/docs/v2-docs/ssosilent.md
jasonnutter, line 3:
```suggestion
If you already have a session that exists with the authentication server, you can use the `ssoSilent()` API to make a request for tokens without interaction. You will need to pass a `login_hint` (which can be retrieved from the account object `username` property or the `upn` claim in the ID token) in the request object in order to successfully obtain a token silently.
```

jo-arroyo: Good idea. Added.

jo-arroyo: Thanks for catching that. Changed.

tnorling (approved):

jasonnutter (approved):

PR #2994, opened 2021-02-08
Fix typing for expiresOn field

### LINKED ISSUE #2991: Null expiresOn for tokens?
## Library

- [x] `@azure/msal-browser@2.10.0`

## Description

When can AuthenticationResult.expiresOn be null? I get this object back from handleRedirectPromise, acquireTokenSilent, etc.

```
const msalConfig: Configuration = {
  auth: {
    clientId: 'XXX',
    authority: 'https://login.microsoftonline.com/consumers/',
    postLogoutRedirectUri: `${window.location.origin}/logout`,
  },
  cache: { cacheLocation: 'localStorage' },
};
```

pkanher617:
Fixes #2991.

Also throws if no access token entity is present in `generateAuthenticationResult`


lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -316,6 +316,8 @@ export class ResponseHandler {
             responseScopes = ScopeSet.fromString(cacheRecord.accessToken.target).asArray();
             expiresOn = new Date(Number(cacheRecord.accessToken.expiresOn) * 1000);
             extExpiresOn = new Date(Number(cacheRecord.accessToken.extendedExpiresOn) * 1000);
+        } else {
+            throw ClientAuthError.createAccessTokenEntityNullError();
```
jasonnutter, line 320: This might cause a problem for angular with the b2c issue where it returns a blank access token: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-angular/src/msal.interceptor.ts#L62

It might be better to wait until b2c fixes that issue before we include this specific change, given there could be apps (and wrapper libraries) that are relying upon a blank access token to be returned, instead of a new error.

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -299,7 +299,7 @@ export class ResponseHandler {
         resourceRequestUri?: string): Promise<AuthenticationResult> {
         let accessToken: string = "";
         let responseScopes: Array<string> = [];
-        let expiresOn: Date | null = null;
+        let expiresOn: Date;
```
tnorling, line 302: This implicitly becomes `Date | undefined`, what was the original issue?

pkanher617: Below we assign the value in the AuthenticationResult object to this variable, which caused a type mismatch.

lib/msal-common/src/response/ResponseHandler.ts
pkanher617: Do we know when that fix is going to get implemented? In the meantime that means we DO have scenarioes where no access token is returned.

jasonnutter: This quarter, I believe. And yes, it means we do, although its technically a bug and not by design.

hectormmg (approved):

tnorling (approved):

jasonnutter (approved):

PR #2684, opened 2020-12-03
[msal-node] Increase E2E Test Coverage for msal-node

hectormmg:
This PR:
- Configures `auth-code`, `device-code` and `silent-flow` samples to run E2E tests for AAD and ADFS scenarios
- Adds E2E test cases for the above sample applications
- Adds Node e2e workflow to test with Node v10, v12 and v14

PR #2998, opened 2021-02-09
update node version support in package.json

sameerag:
Introducing "engines" in `package.json` for `msal-node` to update `Node.js` version support.

```javascript
 "engines" : {
    "node" : ">=14.0.0"
  }
```


lib/msal-node/package.json
```diff
@@ -70,5 +70,8 @@
     "axios": "^0.21.1",
     "jsonwebtoken": "^8.5.1",
     "uuid": "^8.3.0"
+  },
+  "engines" : {
+    "node" : ">=14.0.0"
```
jasonnutter, line 75: I think it would be good to add to the readme detailing what are strategy is and which verisons are supported (i.e. a major version of MSAL node will support the active LTS releases and up at the time it was GA'd)

jasonnutter: One thing to consider: latest version of Azure Functions supports Node 10 and 12 https://docs.microsoft.com/en-us/azure/azure-functions/functions-versions

jasonnutter: Given those are maintenance LTS, we may want to support those in v1.

tnorling, line 75: Should we specify the specific versions to guard against breaking changes in a future node version? Similar to what we do in React/Angular

sameerag: Sure. I was going with the active releases for `Node.js`. Given that there could be scripts dependent on these, we can start with `10.0.0` I suppose, since we already tested against `>=10.0.0`.

@tnorling  I no longer see version testing in `node` E2E tests. Did we revert the changes or am I checking the wrong script?

tnorling: Hasn't been merged yet. It's in Hector's PR #2684

hectormmg, line 75:
```suggestion
    "node" : ">=10.0.0 <16.0.0"
```

jasonnutter:
Per our discussion last night, can we add a table to the readme (similar [msal-angular](https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev/lib/msal-angular#version-support)) where we detailed which versions of Node are supported in which versions of MSAL Node? I think our policy should be when a major version of MSAL Node is released as stable, it should support all LTS and active versions (today that is 10-15, as done in this PR), drop support for any previously supported Node versions that have reached end of life, and should not support prerelease/preview/pending versions until they are stable, as denoted here: https://nodejs.org/en/about/releases/

lib/msal-node/package.json
sameerag: Updated the docs.

lib/msal-node/README.md
```diff
@@ -73,6 +74,19 @@ Before using `@azure/msal-node` you will need to register your app in the azure
...
 ```
+## Node Version Support
+MSAL Node will follow the [Long Term Support (LTS) schedule of the Node.js project](https://nodejs.org/en/about/releases/). Our support plan is as follows.
+
+Any major MSAL Node release:
+- Will support active LTS versions
+- Will drop support for any previously supported Node versions that have reached end of life
+- Will not support prerelease/preview/pending versions until they are stable
+
+| MSAL Node version | MSAL support status     | Supported Angular versions |
```
jasonnutter, line 85:
```suggestion
| MSAL Node version | MSAL support status     | Supported Node versions |
```

lib/msal-node/README.md
```diff
@@ -73,6 +74,19 @@ Before using `@azure/msal-node` you will need to register your app in the azure
 ```javascript
 npm install @azure/msal-node
 ```
+## Node Version Support
+MSAL Node will follow the [Long Term Support (LTS) schedule of the Node.js project](https://nodejs.org/en/about/releases/). Our support plan is as follows.
```
jasonnutter, line 78:
```suggestion
MSAL Node will follow the [Long Term Support (LTS) schedule of the Node.js project](https://nodejs.org/about/releases/). Our support plan is as follows.
```

lib/msal-node/README.md
```diff
@@ -73,6 +74,19 @@ Before using `@azure/msal-node` you will need to register your app in the azure
...
+
+Any major MSAL Node release:
+- Will support active LTS versions
```
jasonnutter, line 81:
```suggestion
- Will support current LTS and active versions
```

lib/msal-node/README.md
```diff
@@ -73,6 +74,19 @@ Before using `@azure/msal-node` you will need to register your app in the azure
...
+- Will drop support for any previously supported Node versions that have reached end of life
+- Will not support prerelease/preview/pending versions until they are stable
+
+| MSAL Node version | MSAL support status     | Supported Angular versions |
+|-------------------|-------------------------|----------------------------|
+| 1.x.x             | Active development      | 10(maintanence), 12, 14    |
```
jasonnutter, line 87:
```suggestion
| 1.x.x             | Active development      | 10, 12, 14, 15    |
```

I don't think you need to label versions as maintenance, as the Node.js page details that.

lib/msal-node/README.md
sameerag: oops!

lib/msal-node/README.md
sameerag: We do not want to list odd numbers, as per the norm??

jasonnutter: Ah okay, then we need to adjust the line above and the `engines` field.

lib/msal-node/package.json
```diff
@@ -70,5 +70,8 @@
     "axios": "^0.21.1",
     "jsonwebtoken": "^8.5.1",
     "uuid": "^8.3.0"
+  },
+  "engines" : {
+    "node" : ">=10.0.0 <16.0.0"
```
jasonnutter, line 75: If we just want to support LTS even release versions, it should be: 

```suggestion
    "node" : "10 || 12 || 14"
```

lib/msal-node/README.md
jasonnutter: Suggestion updated

lib/msal-node/README.md
```diff
@@ -73,6 +74,19 @@ Before using `@azure/msal-node` you will need to register your app in the azure
 ```javascript
 npm install @azure/msal-node
 ```
+## Node Version Support
+MSAL Node will follow the [Long Term Support (LTS) schedule of the Node.js project](https://nodejs.org/about/releases/). Our support plan is as follows.
+
+Any major MSAL Node release:
+- Will support Maintenance LTS, Active LTS, and Current versions of Node
```
jasonnutter, line 81:
```suggestion
- Will support Maintenance LTS and Active LTS versions of Node (even-number releases)
```

jasonnutter, line 81:
```suggestion
- Will support stable (even-numbered) Maintenance LTS, Active LTS, and Current versions of Node
```

lib/msal-node/README.md
sameerag: What I think is this, we will support the range in `engines` since it is usually in between for any node updates, but specify we may not upgrade MSAL Node for odd numbered Node.js updates. I haven't made up my mind yet, we can discuss this and resolve offline.

sameerag: As discussed offline, updating to even versions for now. Let's wait on the feedback and update as needed.

jasonnutter (approved):

hectormmg (approved):

PR #3004, opened 2021-02-10
Add logs to msal-browser

jo-arroyo:
This PR:
- Adds log messages to BrowserCacheManager.ts
- Updates log messages in RedirectHandler.ts


change/@azure-msal-browser-3e900bbd-5fb6-4800-9cf5-beeaefbc5e47.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
tnorling, line 2:
```suggestion
  "type": "patch",
```

lib/msal-browser/src/cache/BrowserCacheManager.ts
```diff
@@ -288,6 +295,7 @@ export class BrowserCacheManager extends CacheManager {
      * @param appMetadataKey
      */
     getAppMetadata(appMetadataKey: string): AppMetadataEntity | null {
+        this.logger.verbose("BrowserCacheManager.getAppMetadata called");
         const value = this.getItem(appMetadataKey);
         if (!value) {
             return null;
```
tnorling, line 301: In each of the getters should we also log whether or not we successfully found something in the cache?

tnorling (approved):
One non-blocking comment. Can be done incrementally if that makes more sense.

lib/msal-browser/src/cache/BrowserCacheManager.ts
pkanher617: +1

jo-arroyo: Makes sense. I added cache hit logs, and it makes the logs a lot longer. If we discover it creates too much noise, I'm happy to remove them later.

PR #3005, opened 2021-02-10
Refresh in feature

pkanher617:


lib/msal-common/src/client/SilentFlowClient.ts
```diff
@@ -60,14 +59,23 @@ export class SilentFlowClient extends BaseClient {
...
             throw ClientAuthError.createRefreshRequiredError();
-        } else {
-            if (this.config.serverTelemetryManager) {
-                this.config.serverTelemetryManager.incrementCacheHits();
-            }
-            return await this.generateResultFromCacheRecord(cacheRecord, request.resourceRequestMethod, request.resourceRequestUri);
+        } else if (cacheRecord.accessToken.refreshOn && TimeUtils.isTokenExpired(cacheRecord.accessToken.refreshOn, 0)) {
+            // Refresh on time has passed, start token renewal process and return current access token
+            const refreshTokenClient = new RefreshTokenClient(this.config);
+            refreshTokenClient.acquireTokenByRefreshToken(request);
```
jasonnutter, line 71: Make sure all errors from this call are caught.

jasonnutter (approved):

lib/msal-common/src/client/SilentFlowClient.ts
```diff
@@ -59,14 +58,20 @@ export class SilentFlowClient extends BaseClient {
         const environment = request.authority || this.authority.getPreferredCache();
         const cacheRecord = this.cacheManager.readCacheRecord(request.account, this.config.authOptions.clientId, requestScopes, environment);
 
-        if (this.isRefreshRequired(request, cacheRecord.accessToken)) {
```
tnorling, line 62: What was the reason for removing the helper function?

pkanher617: Removed it because there was no easy way to reconcile refresh due to token expiration or refresh due to refreshIn time. I can add it back if it makes sense but I figured it was easier to just list it in the function itself.

jasonnutter (approved):

lib/msal-common/src/client/SilentFlowClient.ts
pkanher617: done, caught by calling function

lib/msal-common/src/response/ServerAuthorizationTokenResponse.ts
```diff
@@ -26,6 +26,7 @@ export type ServerAuthorizationTokenResponse = {
     token_type?: string;
     scope?: string;
     expires_in?: number;
+    refresh_in?: number;
```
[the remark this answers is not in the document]
pkanher617: done.

PR #3012, opened 2021-02-11
Await loginRedirect in MSAL Guard to prevent race conditions

jasonnutter:
When `false` is returned from an Angular Guard, the default behavior is for the app to be redirect back to the home page. In theory, `loginRedirect` could take longer than expected, causing the app to be redirected to the home page and then redirected to the login screen. This can be problematic if the home page uses a guard, or automatically redirects to a page use the guard. This should help mitigate this race condition by letting the `loginRedirect` operation complete first (which should happen, as that promise is set to take 30 seconds, but the redirect should always happen first, not giving the promise a chance to resolve).

### DIFF
--- a/lib/msal-angular/src/msal.guard.spec.ts
+++ b/lib/msal-angular/src/msal.guard.spec.ts
@@ -169,10 +169,11 @@ describe('MsalGuard', () => {
       })
     ));
 
-    const listener = jasmine.createSpy();
-    guard.canActivate(routeMock, routeStateMock).subscribe(listener);
-    expect(listener).toHaveBeenCalledWith(false);
-    done();
+    guard.canActivate(routeMock, routeStateMock)
+        .subscribe(result => {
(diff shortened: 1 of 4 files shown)


tnorling (approved):

lib/msal-angular/src/msal.guard.spec.ts
```diff
@@ -169,10 +169,11 @@ describe('MsalGuard', () => {
...
     ));
 
-    const listener = jasmine.createSpy();
-    guard.canActivate(routeMock, routeStateMock).subscribe(listener);
-    expect(listener).toHaveBeenCalledWith(false);
-    done();
+    guard.canActivate(routeMock, routeStateMock)
+        .subscribe(result => {
+            expect(result).toBeFalse;
+            done();
```
jasonnutter, line 175: @jo-arroyo FYI

jo-arroyo (approved):

PR #3013, opened 2021-02-11
Ensure scrollbars are enabled for browser popups

jasonnutter:
This ensures that we are enabling scrollbars for popup interactions. Modern browsers seem to set this by default, but IE11 does not. This change was originally done in v1 in #1426, and we forgot to bring it forward to v2. This is an accessibility consideration, namely.


tnorling (approved):

pkanher617 (approved):
Title confused me

jasonnutter:
Whoops, fixed.

Issue #2626, opened 2020-11-20
Provide reference to popup


kevindelange:
## Library
`@azure/msal-browser@2.x.x`

## Description

Hi,

It would be awesome if we could have a reference to the popup (loginPopup method, which eventually fire's a `window.open()`).

In the current state I believe there is no way we can acces this reference in our (angular 10) application, using msal-browser.

When the popup opens with a B2C policy, it may happen that the popup gets lost by the user. When they then try to open the popup again (for instance, to login or change account details), an error gets thrown: `interaction_in_progress`. This is because the popup is already open. It would be nice to have a reference to the popup, so we can close it, before opening it again.

I'm sorry if i've missed the option to get the reference, or cancel a policy if there's an active one.

jasonnutter:
> When the popup opens with a B2C policy, it may happen that the popup gets lost by the user. 

Can you please clarify what you mean by "gets lost?"

kevindelange:
> > When the popup opens with a B2C policy, it may happen that the popup gets lost by the user.
> 
> Can you please clarify what you mean by "gets lost?"

Clicking next to the popup, so it's not visible anymore may happen (by accident), and then when you try to open it again, because you think it's gone, breaks the flow.

Also, when in need of supplying an authorization code from an email in the popup, and switching to another tab (to copy over the email) it can be confusing for some people that this popup is still open. Mainly because it's a new smaller window, and not an open tab.

The way we currently solved this, is checking if a popup is opened by the user, and then if they click the button again (while a popup is active) we refresh the page. It will probably won't happen that often, but this workaround is not so nice. I think it would be better if we could have a reference to the popup, so we can cancel the `window.open()` before opening a new one.

If there are other solutions to this problem that I am not aware of, could you please point them out? I haven't found any.

jasonnutter:
Update: we're working on some improvements to how popups are handled (#2842), and will fix this scenario incrementally where the same popup request is made twice in a row.

pkanher617:
@kevindelange Have you been able to try out this feature since the PR referenced above was merged?

PR #3044, opened 2021-02-19
Add logoutPopup API

tnorling:
Adds `logoutPopup` API as described in the design proposal: #3123


lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -649,6 +661,82 @@ export abstract class ClientApplication {
         }
     }
 
+    /**
+     * Clears local cache for the current user then opens a popup window prompting the user to sign-out of the server
+     * @param logoutRequest 
+     */
+    async logoutPopup(logoutRequest: EndSessionRequest): Promise<void> {
```
jasonnutter, line 668: This needs to not be marked as `async` to properly mitigate the blocking performed by Safari.

jasonnutter: (it should/can still return a Promise)

tnorling: Done

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -702,6 +715,101 @@ export abstract class ClientApplication {
...
+     * @param logoutRequest 
+     */
+    logoutPopup(logoutRequest?: EndSessionPopupRequest): Promise<void> {
+        let validLogoutRequest: CommonEndSessionRequest;
+        try {
+            this.preflightBrowserEnvironmentCheck(InteractionType.Popup);
+            this.logger.verbose("logoutPopup called");
+            validLogoutRequest = this.initializeLogoutRequest(logoutRequest);
+        } catch (e) {
+            // Since this function is syncronous we need to reject
```
jo-arroyo, line 733:
```suggestion
            // Since this function is synchronous we need to reject
```

lib/msal-browser/src/event/EventMessage.ts
```diff
@@ -16,7 +16,7 @@ export type EventMessage = {
     timestamp: number;
 };
 
-export type EventPayload = PopupRequest | RedirectRequest | SilentRequest | SsoSilentRequest | EndSessionRequest | AuthenticationResult | null;
+export type EventPayload = PopupRequest | RedirectRequest | SilentRequest | SsoSilentRequest | EndSessionRequest | AuthenticationResult | Window | null;
```
jasonnutter, line 19: Why do we emit the window object?

lib/msal-browser/src/interaction_handler/PopupHandler.ts
```diff
@@ -65,145 +63,35 @@ export class PopupHandler extends InteractionHandler {
...
-                    return;
-                } else if (ticks > maxTicks) {
-                    // Timeout error
-                    this.cleanPopup(popupWindow);
-                    clearInterval(intervalId);
-                    reject(BrowserAuthError.createMonitorPopupTimeoutError());
-                    return;
-                }
-            }, BrowserConstants.POLL_INTERVAL_MS);
+            this.popupUtils.monitorPopupForSameOrigin(popupWindow).then(() => {
```
jasonnutter, line 66: Once we are on the same origin, can we just return the hash/url, instead of polling again?

lib/msal-browser/src/request/EndSessionPopupRequest.ts
```diff
@@ -0,0 +1,20 @@
...
+ * - account                - Account object that will be logged out of. All tokens tied to this account will be cleared.
+ * - postLogoutRedirectUri  - URI to navigate to after logout page.
+ * - authority              - Authority to send logout request to.
+ * - correlationId          - Unique GUID set per request to trace a request end-to-end for telemetry purposes.
+ * - idTokenHint            - ID Token used by B2C to validate logout if required by the policy
+ * - onRedirectNavigate     - Callback that will be passed the url that MSAL will navigate to. Returning false in the callback will stop navigation.
+ */
+export type EndSessionPopupRequest = Partial<CommonEndSessionRequest> & {
+    authority?: string;
+    redirectMainWindowTo?: string;
```
jasonnutter, line 19: I'm not sure I like the naming. `postLogoutPopupRedirectUri`? Or is that too similar to `postLogoutRedirectUri`?

lib/msal-browser/src/event/EventMessage.ts
tnorling: I figured the `POPUP_OPENED` event was a good opportunity to address #2626 and would also allow applications to write their own logic to close the popup if it doesn't get closed for whatever reason.

lib/msal-browser/src/request/EndSessionPopupRequest.ts
tnorling: Open to changing the name but would like to pick something that doesn't have "logout" in it if we ever decide to add a similar feature to `loginPopup`

lib/msal-browser/src/event/EventMessage.ts
jasonnutter: I see. Can we create a new type, e.g. `PopupEvent`, that contains the window object as a property? To future proof in case we want/need to include other info.

lib/msal-browser/src/interaction_handler/PopupHandler.ts
tnorling: Wasn't sure why the hash extraction was inside the polling previously so didn't want to change it. But I've pulled it out now.

lib/msal-browser/src/event/EventMessage.ts
tnorling: Done

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -702,6 +715,99 @@ export abstract class ClientApplication {
...
+
+            const popupUtils = new PopupUtils(this.browserStorage, this.logger);
+            // Open the popup window to requestUrl.
+            const popupWindow = popupUtils.openPopup(logoutUri, popupName, popup);
+            this.emitEvent(EventType.POPUP_OPENED, InteractionType.Popup, {popupWindow}, null);
+
+            try {
+                // Don't care if this throws an error (User Cancelled)
+                await popupUtils.monitorPopupForSameOrigin(popupWindow);
+            } catch {}
```
jasonnutter, line 789: A verbose log message would probably be useful in this catch (and right after the away in the try)

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -702,6 +715,99 @@ export abstract class ClientApplication {
...
+
+            popupUtils.cleanPopup(popupWindow);
+
+            if (redirectMainWindowTo) {
```
jasonnutter, line 793: Can you add a log message for the if and else cases for this block?

lib/msal-browser/src/app/ClientApplication.ts
tnorling: Done

lib/msal-browser/src/app/ClientApplication.ts
tnorling: Done

jo-arroyo (approved):

lib/msal-browser/src/request/EndSessionPopupRequest.ts
jasonnutter: How about `mainWindowRedirectUri`?

jasonnutter (approved):
One suggestion on naming, and some e2e test seem to be failing, otherwise looks good!

lib/msal-browser/src/request/EndSessionPopupRequest.ts
tnorling: Done

tnorling:
> One suggestion on naming, and some e2e test seem to be failing, otherwise looks good!

Angular E2E tests are still running on published versions of the lib afaik so it is expected that they fail on this PR. This is something we need to fix before GA cc: @jo-arroyo

lib/msal-browser/docs/request-response-object.md
```diff
@@ -11,5 +11,6 @@ The MSAL Browser library has a set of configuration options that can be used to
 | `acquireTokenSilent` | [SilentRequest](https://azuread.github.io/microsoft-authentication-library-for-js/ref/modules/_azure_msal_browser.html#silentrequest) | [AuthenticationResult](https://azuread.github.io/microsoft-authentication-library-for-js/ref/modules/_azure_msal_browser.html#authenticationresult) |
 | `loginPopup` | [PopupRequest](https://azuread.github.io/microsoft-authentication-library-for-js/ref/modules/_azure_msal_browser.html#popuprequest) | [AuthenticationResult](https://azuread.github.io/microsoft-authentication-library-for-js/ref/modules/_azure_msal_browser.html#authenticationresult) |
 | `loginRedirect` | [RedirectRequest](https://azuread.github.io/microsoft-authentication-library-for-js/ref/modules/_azure_msal_browser.html#redirectrequest) | [AuthenticationResult](https://azuread.github.io/microsoft-authentication-library-for-js/ref/modules/_azure_msal_browser.html#authenticationresult) (via `handleRedirectPromise`) |
-| `logout` | [EndSessionRequest](https://azuread.github.io/microsoft-authentication-library-for-js/ref/modules/_azure_msal_browser.html#endsessionrequest) | `void` |
```
ralphtheninja, line 14: This is technically a new major version. Assuming you're going with semver that is.

tnorling: Yes you're right that this would be a breaking change, however, the `logout` API has not been removed from the library, just from the docs (to encourage people to use the new APIs going forward). This API will be removed in the next major version

Issue #3059, opened 2021-02-22
Chromium extension sample fails because redirect uris don't match


arcadiogarcia:
## Library

- [ ] `msal@1.x.x` or `@azure/msal@1.x.x`
- [x] `@azure/msal-browser@2.x.x`
- [ ] `@azure/msal-node@1.x.x`
- [ ] `@azure/msal-react@1.x.x`
- [ ] `@azure/msal-angular@0.x.x`
- [ ] `@azure/msal-angular@1.x.x`
- [ ] `@azure/msal-angular@2.x.x`
- [ ] `@azure/msal-angularjs@1.x.x`

## Framework

- [ ] Angular
- [ ] React
- [ ] Other

## Description

The code in provided Chromium extension sample fails to log in, it compails about the redirect_uri not matching.

The cause of this seems to be that the chromium redirect uri is missing in the parameters of the original auth url, after adding it it logins correctly.

## Error Message

index.es.js:8747 POST https://login.microsoftonline.com/common/oauth2/v2.0/token 400 (Bad Request)
index.es.js:524 Uncaught (in promise) ServerError: invalid_grant: 70000 - [2021-02-22 19:34:37Z]: AADSTS70000: The provided value for the 'redirect_uri' is not valid. The value must exactly match the redirect URI used to obtain the authorization code.
Trace ID: e32a97ed-a580-43a6-b931-37ecd3081b00
Correlation ID: 40c82ca9-081b-46f5-8965-ab768228a3c5
Timestamp: 2021-02-22 19:34:37Z - Correlation ID: 40c82ca9-081b-46f5-8965-ab768228a3c5 - Trace ID: e32a97ed-a580-43a6-b931-37ecd3081b00
    at t [as constructor] (chrome-extension://okmgenonbcmbmhkaljgdaaadkjjaipjm/flyout.bundle.js:32:6666)
    at new t (chrome-extension://okmgenonbcmbmhkaljgdaaadkjjaipjm/flyout.bundle.js:32:42748)
    at e.validateTokenResponse (chrome-extension://okmgenonbcmbmhkaljgdaaadkjjaipjm/flyout.bundle.js:32:60761)
    at t.<anonymous> (chrome-extension://okmgenonbcmbmhkaljgdaaadkjjaipjm/flyout.bundle.js:32:65741)
    at chrome-extension://okmgenonbcmbmhkaljgdaaadkjjaipjm/flyout.bundle.js:32:1863
[...]

PR #3060, opened 2021-02-22
Add redirectUri parameter to chromium extension sample

arcadiogarcia:
See #3059, the chromium sample fails to login because the redirect uri used to obtain the auth code doesn't match the one used late ,in the token call, this PR modifies the sample and the readme to explicitly pass the right redirect uri.


jasonnutter:
The `redirectUri` is set on the `PublicClientApplication` instance, which result in that `redirectUri` being set for all requests, so you should not need to set the `redirectUri` on the request itself unless if you are using a different `redirectUri`.

PR #3068, opened 2021-02-23
[msal-node]: Add regional authorities

samuelkubai:
Add support for regional authorities in msal-node


lib/msal-common/src/authority/RegionDiscovery.ts
```diff
@@ -0,0 +1,91 @@
...
+ */
+
+import { INetworkModule } from "../network/INetworkModule";
+import { IMDSBadResponse } from "../response/IMDSBadResponse";
+
+export class RegionDiscovery {
+    // Network interface to make requests with.
+    protected networkInterface: INetworkModule;
+    // The IMDS endpoint to retrieve region information.
+    protected static IMDS_ENDPOINT = "http://169.254.169.254/metadata/instance/compute/location";
```
tnorling, line 13: Is this the same for everyone? Can we put these in the constants file?

lib/msal-common/src/authority/RegionDiscovery.ts
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    /**
+     * Detect the region from the application's environment.
+     * 
+     * @returns Promise<string | null>
+     */
+    public async detectRegion(): Promise<string | null> {
+        // Detect region from the process environment variable
+        let autodetectedRegionName = process.env[RegionDiscovery.REGION_ENV_NAME];
```
tnorling, line 32: This will not be available in browser, please find another way to get the region env or move to node

lib/msal-common/src/authority/RegionDiscovery.ts
```diff
@@ -0,0 +1,91 @@
...
+     *  
+     * @returns Promise<string | null>
+     */
+    private async getCurrentVersion(): Promise<string | null> {
+        try {
+            const response = await this.networkInterface.sendGetRequestAsync<IMDSBadResponse>(`${RegionDiscovery.IMDS_ENDPOINT}?format=json`, RegionDiscovery.IMDS_OPTIONS);
+
+            // When IMDS endpoint is called without the api version query param, bad request response comes back with latest version.
+            if (response.status === 400) {
+                return response.body["newest-versions"][0];
```
tnorling, line 83: Please add validation that response.body is truthy and contains "newest-versions" which has at least 1 member

lib/msal-common/src/response/IMDSBadResponse.ts
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+export type IMDSBadResponse = {
+    error: string;
+    "newest-versions": Array<string>;
```
tnorling, line 8:
```suggestion
    newest-versions: Array<string>;
```

lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -312,14 +315,14 @@ export abstract class ClientApplication {
...
+    private async createAuthority(authorityString: string, prefferedAzureRegionOptions?: PrefferedAzureRegionOptions): Promise<Authority> {
         this.logger.verbose("createAuthority called");
         const authorityOptions: AuthorityOptions = {
             protocolMode: this.config.auth.protocolMode!,
             knownAuthorities: this.config.auth.knownAuthorities!,
             cloudDiscoveryMetadata: this.config.auth.cloudDiscoveryMetadata!,
             authorityMetadata: this.config.auth.authorityMetadata!
         };
-        return await AuthorityFactory.createDiscoveredInstance(authorityString, this.config.system!.networkClient!, this.storage, authorityOptions);
+        return await AuthorityFactory.createDiscoveredInstance(authorityString, this.config.system!.networkClient!, this.storage, authorityOptions, prefferedAzureRegionOptions);
```
tnorling, line 326: Is this feature applicable to browser? If so please make the changes there as well.

change/@azure-msal-angular-bd9bc5d3-a8d8-4d07-83b4-246e0a84455f.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "updating package-lock",
```
tnorling, line 3: Can you please remove the package lock changes?

tnorling (changes requested):

lib/msal-common/src/authority/Authority.ts
```diff
@@ -36,13 +38,19 @@ export class Authority {
     private authorityOptions: AuthorityOptions;
     // Authority metadata
     private metadata: AuthorityMetadataEntity;
+    // Region discovery service
+    private regionDiscovery: RegionDiscovery;
+    // Authority preffered azure region options
+    private prefferedAzureRegionOptions: PrefferedAzureRegionOptions | null;
 
-    constructor(authority: string, networkInterface: INetworkModule, cacheManager: ICacheManager, authorityOptions: AuthorityOptions) {
+    constructor(authority: string, networkInterface: INetworkModule, cacheManager: ICacheManager, authorityOptions: AuthorityOptions, prefferedAzureRegionOptions?: PrefferedAzureRegionOptions) {
```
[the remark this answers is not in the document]
samuelkubai: Thanks for this suggestion @tnorling it was a good improvement to the change intended.

lib/msal-common/src/authority/RegionDiscovery.ts
samuelkubai: @tnorling how do you propose we move the code to `msal-node` whilst being so tightly knit with the `Authority class`

change/@azure-msal-angular-bd9bc5d3-a8d8-4d07-83b4-246e0a84455f.json
samuelkubai: Removed.

lib/msal-common/src/authority/RegionDiscovery.ts
tnorling: Can you include this value on `PreferredAzureRegionOptions`?

samuelkubai: @sameerag what's you thought on adding this value to the `PreferredAzureRegionOptions`, the implication would be that we are moving the region discovery from the environment variable to the consumer of the API.

samuelkubai: Pro:
- It allows them to get the environment variable in whatever way that suites their environment

Con:
- It shifts the responsibility to the users of the library

tnorling: You can still set it in `msal-node`, without user providing it if you like, since we are passing this value into authority instantiation. But we should not be putting code in common that is not supported in downstream libraries.

samuelkubai: Gotcha, that makes sense, thanks @tnorling, we'll work on that improvement

samuelkubai: @tnorling I have implemented the changes you suggested please re-review.

lib/msal-common/src/authority/Authority.ts
```diff
@@ -435,4 +453,36 @@ export class Authority {
...
+     */
+    static buildRegionalAuthorityString(host: string, region: string): string {
+        // Create and validate a Url string object with the initial authority string
+        const globalUrl = new UrlString(host);
+        globalUrl.validateAsUri();
+
+        // Include the query string portion of the url
+        return UrlString.constructAuthorityUriFromObject({
+            ...globalUrl.getUrlComponents(),
+            HostNameAndPort: `${region}.login.microsoft.com`
```
sameerag, line 471: Did we test this for all regions? This pattern had some issues before if I recall with `/token` end point I assumed. As long as we are dynamically reading the `token endpoint` and not constructing, we should be fine though.

lib/msal-common/src/authority/AzureRegionConfiguration.ts
```diff
@@ -0,0 +1,17 @@
...
+import { AzureRegion } from "./AzureRegion";
+
+/*
+ * AzureRegionConfiguration
+ * - preferredAzureRegion       - Preferred azure region from the user 
+ * - environmentRegionFunc      - Environment specific way of fetching the region from the environment
+ */
+export type AzureRegionConfiguration = {
+    preferredAzureRegion?: AzureRegion,
+    environmentRegionFunc: () => string | undefined;
```
tnorling, line 15: Why does this need to be a function? Can you just pass in the string?

samples/msal-node-samples/client-credentials/index.js
```diff
@@ -7,10 +7,19 @@ var msal = require('@azure/msal-node');
 
 const config = {
     auth: {
-        clientId: "12d77c73-d09d-406a-ae0d-3d4e576f7d9b",
-        authority: "https://login.microsoftonline.com/90b8faa8-cc95-460e-a618-ee770bee1759",
-        clientSecret: ""
-   } 
+        clientId: "48f55727-da76-4dbb-9908-d9335f239e20",
+        authority: "https://login.microsoftonline.com/eefab79a-7793-4852-be01-105ecb6f6d4e/",
+        clientSecret: "epFuS.3-QDTU3DGI-QQs8UwQx-34jy.u9y",
```
sameerag, line 12: This should not be open. Can we share it in keyvault and programmatically read it? cc @derisen

samuelkubai: Hi @sameerag, don't we have all samples sharing their client information publicly and if we put in in a key vault wouldn't users of the sample be able to access the details all the same?

lib/msal-common/src/authority/AzureRegionConfiguration.ts
samuelkubai: That's a good point, I agree we don't need to pass in a function we can just pass in the string.

samples/msal-node-samples/client-credentials/index.js
jasonnutter: Anyone using samples that require secrets must provide their own. Please remove this, and invalidate the secret from the app registration.

samuelkubai: I have removed them and will invalidate them.

lib/msal-common/src/error/ClientAuthError.ts
```diff
@@ -491,6 +495,13 @@ export class ClientAuthError extends AuthError {
         return new ClientAuthError(ClientAuthErrorMessage.noAuthorizationCodeFromServer.code, ClientAuthErrorMessage.noAuthorizationCodeFromServer.desc);
     }
 
+    /**
+     * Throws error when no azure region is detected and fallback is provided
+     */
+    static createNoAzureRegionDetectedError(): ClientAuthError {
```
tnorling, line 501: Where is this used?

samuelkubai: This was previously used before the API changed, removing it as it is no longer used.

tnorling (approved):
Please remove the package-lock for samples. Otherwise, LGTM nice work!

jasonnutter:
@samuelkubai Is this ready? @sameerag we will need an approval from you, I think.

lib/msal-common/src/authority/Authority.ts
```diff
@@ -257,6 +261,19 @@ export class Authority {
...
+
+                const azureRegion = this.authorityOptions.azureRegionConfiguration.preferredAzureRegion === Constants.AZURE_REGION_AUTO_DISCOVER_FLAG 
+                    ? autodetectedRegionName 
+                    : this.authorityOptions.azureRegionConfiguration.preferredAzureRegion;
+
+                if (azureRegion) {
+                    metadata = Authority.replaceWithRegionalInformation(metadata, azureRegion);
+                }
+            }
+
```
sameerag, line 276: So this code fragment ensures we do the instance_discovery with global authority and then add regional_data right? Can we please add a test case for it?

lib/msal-common/src/authority/AzureRegionConfiguration.ts
```diff
@@ -0,0 +1,17 @@
...
+
+import { AzureRegion } from "./AzureRegion";
+
+/*
+ * AzureRegionConfiguration
+ * - preferredAzureRegion       - Preferred azure region from the user 
+ * - environmentRegionFunc      - Environment specific way of fetching the region from the environment
+ */
+export type AzureRegionConfiguration = {
+    preferredAzureRegion?: AzureRegion,
```
sameerag, line 14: rename to `azureRegion`?

lib/msal-common/src/authority/RegionDiscovery.ts
```diff
@@ -0,0 +1,86 @@
...
+     */
+    public async detectRegion(environmentRegion: string | undefined): Promise<string | null> {
+        // Initialize auto detected region with the region from the envrionment 
+        let autodetectedRegionName = environmentRegion;
+
+        // Call the local IMDS endpoint for applications running in azure vms
+        if (!autodetectedRegionName) {
+            try {
+                const response = await this.getRegionFromIMDS(Constants.IMDS_VERSION);
+                if (response.status === 200) {
```
sameerag, line 34: can we have constants:
``` javascript
const  responseCodes = {
 httpSuccess: 200,
 httpFailure: 400
} 
```

instead here?

lib/msal-common/src/request/CommonClientCredentialRequest.ts
```diff
@@ -4,14 +4,17 @@
...
- * - skipCache               - Skip token cache lookup and force request to authority to get a a new token. Defaults to false.
+ * - scopes                             - Array of scopes the application is requesting access to.
+ * - authority                          - URL of the authority, the security token service (STS) from which MSAL will acquire tokens.
+ * - correlationId                      - Unique GUID set per request to trace a request end-to-end for telemetry purposes.
+ * - skipCache                          - Skip token cache lookup and force request to authority to get a a new token. Defaults to false.
+ * - preferredAzureRegionOptions        - Options of the user's preferred azure region
  */
 export type CommonClientCredentialRequest = BaseAuthRequest & {
     skipCache?: boolean;
+    preferredAzureRegion?: AzureRegion;
```
sameerag, line 19: Bogdan's comment is to make this common across ConfidentialClients. I do think we can add this for all requests separately when the use cases are supported, can you please check once though?

samuelkubai: I don't understand the ask here, could you shed more light on it?

lib/msal-common/src/authority/RegionDiscovery.ts
samuelkubai: Sure, thing.

lib/msal-common/src/authority/AzureRegionConfiguration.ts
samuelkubai: Sure thing.

lib/msal-common/src/request/CommonClientCredentialRequest.ts
sameerag: Addressed.

PR #3072, opened 2021-02-23
Memoize handleRedirectPromise

tnorling:
Multiple concurrent calls to `handleRedirectPromise` can cause race conditions and confusing errors when the first invocation clears temporary cache before the 2nd can read it. The official recommendation up until now has been "dont call handleRedirectPromise more than 1 time per page load" but this can be confusing and easy to overlook when using frameworks (react, angular, etc.) that may re-render components and call `handleRedirectPromise` again unexpectedly.

This PR aims to solve this once and for all by having all invocations of `handleRedirectPromise` on a single `PublicClientApplication` instance return the same promise and ensure that the hash processing logic is only run once.

Note: This does not solve scenarios where your page has multiple instances of `PublicClientApplication` calling `handleRedirectPromise` or where `PublicClientApplication` is re-instantiated and and calls `handleRedirectPromise` again (your app should still only instantiate `PublicClientApplication` once and should not be re-instantiated as a result of a re-render)

### DIFF
--- a/lib/msal-browser/src/app/ClientApplication.ts
+++ b/lib/msal-browser/src/app/ClientApplication.ts
@@ -126,7 +126,13 @@ export abstract class ClientApplication {
         this.logger.verbose("handleRedirectPromise called");
         const loggedInAccounts = this.getAllAccounts();
         if (this.isBrowserEnvironment) {
+            /**
+             * Store the promise on the PublicClientApplication instance if this is the first invocation of handleRedirectPromise,
+             * otherwise return the promise from the first invocation. Prevents race conditions when handleRedirectPromise is called
+             * several times concurrently.
+             */
             if (typeof this.redirectResponse === "undefined") {
+                this.logger.verbose("handleRedirectPromise has been called for the first time, storing the promise");
                 this.redirectResponse = this.handleRedirectResponse(hash)
                     .then((result: AuthenticationResult | null) => {
                         if (result) {
@@ -155,7 +161,9 @@ export abstract class ClientApplication {
 
                         throw e;
                     });
-            } 
+            } else {
+                this.logger.verbose("handleRedirectPromise has been called previously, returning the result from the first call");
+            }
             
             return this.redirectResponse;
         }
--- a/lib/msal-browser/src/app/ClientApplication.ts
+++ b/lib/msal-browser/src/app/ClientApplication.ts
@@ -55,7 +55,7 @@ export abstract class ClientApplication {
     private eventCallbacks: Map<string, EventCallbackFunction>;
 
     // Redirect Response Object
-    private redirectResponse: Promise<AuthenticationResult | null> | undefined;
+    private redirectResponse: Map<string, Promise<AuthenticationResult | null>>;
 
     /**
      * @constructor
@@ -99,6 +99,9 @@ export abstract class ClientApplication {
         // Initialize the network module class.
         this.networkClient = this.config.system.networkClient;
 
+        // Initialize redirectResponse Map
+        this.redirectResponse = new Map();
+
         if (!this.isBrowserEnvironment) {
             this.browserStorage = DEFAULT_BROWSER_CACHE_MANAGER(this.config.auth.clientId, this.logger);
             this.browserCrypto = DEFAULT_CRYPTO_IMPLEMENTATION;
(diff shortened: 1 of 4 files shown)


lib/msal-browser/test/app/PublicClientApplication.spec.ts
```diff
@@ -484,6 +484,111 @@ describe("PublicClientApplication.ts Class Unit Tests", () => {
                 expect(window.sessionStorage.length).to.be.eq(4);
             });
 
+            it.only("Multiple concurrent calls to handleRedirectPromise return the same promise", async () => {
```
jasonnutter:
```suggestion
            it("Multiple concurrent calls to handleRedirectPromise return the same promise", async () => {
```

tnorling: You looked too fast 🤣  Fixed

pkanher617: remove `.only`

pkanher617: sorry didn't refresh

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -54,6 +54,9 @@ export abstract class ClientApplication {
     // Callback for subscribing to events
     private eventCallbacks: Map<string, EventCallbackFunction>;
 
+    // Redirect Response Object
+    private redirectResponse: Promise<AuthenticationResult | null> | undefined;
```
pkanher617, line 58: If the page is re-rendered won't this variable be re-rendered also? Doesn't this only solve `handleRedirectPromise` being called multiple times in a single page load?

tnorling: Yes, that's right. More specifically this solves `handleRedirectPromise` being called multiple times per pca instance.

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -123,34 +126,38 @@ export abstract class ClientApplication {
...
         if (this.isBrowserEnvironment) {
-            return this.handleRedirectResponse(hash)
-                .then((result: AuthenticationResult | null) => {
-                    if (result) {
-                        // Emit login event if number of accounts change
-                        const isLoggingIn = loggedInAccounts.length < this.getAllAccounts().length;
-                        if (isLoggingIn) {
-                            this.emitEvent(EventType.LOGIN_SUCCESS, InteractionType.Redirect, result);
-                            this.logger.verbose("handleRedirectResponse returned result, login success");
+            if (typeof this.redirectResponse === "undefined") {
```
jasonnutter, line 129: Can you please add a comment explaining why this is necessary?

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -123,34 +126,38 @@ export abstract class ClientApplication {
...
-                    }
-                    this.emitEvent(EventType.HANDLE_REDIRECT_END, InteractionType.Redirect);
-
-                    throw e;
-                });
+                        this.emitEvent(EventType.HANDLE_REDIRECT_END, InteractionType.Redirect);
+
+                        throw e;
+                    });
+            }
```
jasonnutter, line 158: I think it would be good to add some verbose logs here indicating the existing promise is being reused

jasonnutter (approved):
A couple comments otherwise looks good!

lib/msal-browser/src/app/ClientApplication.ts
tnorling: Done

lib/msal-browser/src/app/ClientApplication.ts
tnorling: Done

lib/msal-browser/src/app/ClientApplication.ts
pkanher617: in the PR description it says that it may solve issues when frameworks re-render components, we should update this in case someone gets confused thinking this will solve re-rendered instances.

pkanher617 (approved):
1 minor comment, otherwise looks good

sameerag (approved):

lib/msal-browser/src/app/ClientApplication.ts
sameerag: For a min, I thought the same. :-D

tnorling: Page refresh is not a broken scenario. This is only to guard against concurrent calls to handleRedirectPromise throwing errors

PR #3074, opened 2021-02-23
Update msal-angular preinstall script to force npm-force-resolutions version

jo-arroyo:
Msal-Angular has a preinstall script that runs `npm-force-resolutions` to force npm to install certain versions of dependencies. A recent update to the `npm-force-resolutions` package is causing this preinstall script to fail, so this PR forces the version of `npm-force-resolutions` to 0.0.3.

### DIFF
--- a/change/@azure-msal-angular-7c84ff5a-7773-4dde-a551-3665adb6e468.json
+++ b/change/@azure-msal-angular-7c84ff5a-7773-4dde-a551-3665adb6e468.json
@@ -1,5 +1,5 @@
 {
-  "type": "patch",
+  "type": "prerelease",
   "comment": "Update msal-angular preinstall script to force npm-force-resolutions version (#3074)",
   "packageName": "@azure/msal-angular",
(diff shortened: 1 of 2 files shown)


change/@azure-msal-angular-7c84ff5a-7773-4dde-a551-3665adb6e468.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
jasonnutter, line 2: I think this can be a prerelease change
```suggestion
  "type": "prerelease",
```

tnorling (approved):

change/@azure-msal-angular-7c84ff5a-7773-4dde-a551-3665adb6e468.json
jo-arroyo: Yes. Makes sense. Changed.

jasonnutter (approved):

PR #3076, opened 2021-02-24
Added period on authority link.

jaymendez:
Without the period, other people would just copy and it would result in an error.


tnorling:
@jaymendez Thanks for catching this and correcting it! Could you please run `npm run beachball:change` in the root of the repo and select change type `none` to generate the change files as per our [contributing guidelines](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/contributing.md#beachball-check)?

PR #3089, opened 2021-02-25
[msal-common] Add support for custom client claims in SHR

hectormmg:
This PR:
- Adds optional clientClaims attribute to `SignedHttpRequest` type
- Adds the clientClaims parameter to ResponseHandler.handleServerTokenResponse calls
- Adds the `client_claims` claim to the PoP Access Token JWT which is signed and returned as the SHR
- Adds `x5c_ca` as a known ID Token claim to TokenClaims type
- Adds Custom Client Claims doc to msal-browser docs


sameerag (approved):
Can we add docs? Is this an enhancement to POP?

hectormmg:
> Can we add docs? Is this an enhancement to POP?

I'm currently working on AT PoP docs separately. I'll cover this there as well. And yes, this is an enhancement for AT PoP.

lib/msal-common/src/request/BaseAuthRequest.ts
```diff
@@ -17,6 +18,7 @@ export type BaseAuthRequest = {
     correlationId: string;
     scopes: Array<string>;
     claims?: string;
+    clientClaims?: string;
```
pkanher617, line 21: `clientClaims` may get confused with `claims`, maybe use `requestClientClaims` or `shrClientClaims`?

hectormmg: The claim name comes from the cross-platform spec, so I think for parity's sake we should keep `clientClaims`. Still open to discussing this further.

lib/msal-common/src/client/SilentFlowClient.ts
```diff
@@ -65,15 +65,15 @@ export class SilentFlowClient extends BaseClient {
             if (this.config.serverTelemetryManager) {
                 this.config.serverTelemetryManager.incrementCacheHits();
             }
-            return await this.generateResultFromCacheRecord(cacheRecord, request.resourceRequestMethod, request.resourceRequestUri);
+            return await this.generateResultFromCacheRecord(cacheRecord, request.resourceRequestMethod, request.resourceRequestUri, request.clientClaims);
```
tnorling, line 68: Can we condense this to just pass request?

lib/msal-common/src/request/BaseAuthRequest.ts
tnorling: I agree, can we call this something like `shrClaims`?

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -51,6 +51,7 @@ export class RefreshTokenClient extends BaseClient {
             reqTimestamp,
             request.resourceRequestMethod,
             request.resourceRequestUri,
+            request.clientClaims,
```
tnorling, line 54: Also condense these into just passing `request` and destructure inside the function

tnorling (commented):
Couple nitpicks. Can you also add a short doc describing what this is for and how to use it?

lib/msal-common/src/request/BaseAuthRequest.ts
pkanher617: I meant the variable on the request object. I understand other platforms are also implementing this feature but I don't think that should stop us from using descriptive naming.

lib/msal-common/src/client/SilentFlowClient.ts
hectormmg: Should I make it a `BaseAuthRequest` to cover every auth request type?

tnorling: Yeah that's fine for now

lib/msal-common/src/request/BaseAuthRequest.ts
hectormmg: Changed to `shrClaims`

lib/msal-common/src/client/SilentFlowClient.ts
hectormmg: Done!

lib/msal-browser/docs/shr-client-claims.md
```diff
@@ -0,0 +1,48 @@
...
+
+Given that the payload of a `Signed HTTP Request` has a specific format, rather than allowing custom claim names which could collide with any of the [first-order SHR claims](https://tools.ietf.org/html/draft-ietf-oauth-signed-http-request-03#section-3), MSAL Browser allows custom client claims to be passed in as a string, which will be added to the Signed HTTP Request `JWT` as the value of a claim called `client_claims`. This string will usually be a serialized object containing the custom claims, but MSAL will not parse or verify the string in any way.
+
+Given that MSAL does not cache `Signed HTTP Requests` (the access token secret is cached and the SHR payload is built and signed just-in-time whenever `acquireTokenSilent` is called), custom client claims will **not** be cached either. This means that the claims must also be passed into every `acquireTokenSilent` call in order for them to be added to the resulting `SignedHTTPRequest`.
```
tnorling, line 7: Does passing client claims automatically result in a cache miss like passing `claims` does?

hectormmg: No, the client claims aren't taken into account for the caching logic. They go straight into the SHR "envelope".

tnorling (approved):

PR #3094, opened 2021-02-26
Move /token headers to POST body

tnorling:
Telemetry and library info headers on the /token endpoint are causing an OPTIONS request which is an unnecessary performance hit. In discussion with the server team we've decided to move these headers to the POST body in order to make /token requests "simple requests" in the context of CORS. We've also made the decision to reduce the maximum size that the telemetry headers can be and how much we cache at any given point.

Do not merge until the server changes have been fully deployed.

### DIFF
--- a/lib/msal-common/src/client/AuthorizationCodeClient.ts
+++ b/lib/msal-common/src/client/AuthorizationCodeClient.ts
@@ -168,6 +168,15 @@ export class AuthorizationCodeClient extends BaseClient {
         // add code: user set, not validated
         parameterBuilder.addAuthorizationCode(request.code);
 
+        // Add library metadata
+        parameterBuilder.addLibraryInfo(this.config.libraryInfo);
+
+        parameterBuilder.addThrottling();
+        
+        if (this.serverTelemetryManager) {
+            parameterBuilder.addServerTelemetry(this.serverTelemetryManager);
+        }
+
         // add code_verifier if passed
         if (request.codeVerifier) {
             parameterBuilder.addCodeVerifier(request.codeVerifier);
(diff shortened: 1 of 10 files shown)


jasonnutter:
@tnorling I believe server support has shipped, can you please confirm?

tnorling:
> @tnorling I believe server support has shipped, can you please confirm?

Looks like we're still waiting for rollout to complete on 3 more servers

sameerag (approved):
Thanks for the tests. Can we change the PR desc to /authorize and /token end points? As we seem to do both?

tnorling:
> Thanks for the tests. Can we change the PR desc to /authorize and /token end points? As we seem to do both?

Not sure I understand, can you clarify what you mean by this? All params on /authorize endpoint go in the query string as it's a GET request (and this hasn't been changed). This PR specifically moves params from the /token headers to the body of the POST request.

jasonnutter:
This looks good. One thought: would it be possible to add a test to assert that we aren't adding extra headers that would result in a preflight request? That would help protect against someone accidently introducing a regression that results in the preflight request being made again by the browser.

hectormmg (approved):
LGTM!

sameerag:
> > Thanks for the tests. Can we change the PR desc to /authorize and /token end points? As we seem to do both?
> 
> Not sure I understand, can you clarify what you mean by this? All params on /authorize endpoint go in the query string as it's a GET request (and this hasn't been changed). This PR specifically moves params from the /token headers to the body of the POST request.

Sorry my mistake. I thought I saw `AuthorizationUrlRequest` and misunderstood. Yes, makes sense.

tnorling:
> This looks good. One thought: would it be possible to add a test to assert that we aren't adding extra headers that would result in a preflight request? That would help protect against someone accidently introducing a regression that results in the preflight request being made again by the browser.

Good idea! Added.

pkanher617 (approved):
+1 for tests checking headers, otherwise looks great

PR #3116, opened 2021-03-01
Ensure indexeddb is cleared on logout

jasonnutter:
Today, anything stored in IndexedDB is not cleared on logout. This PR makes the following changes:

- Adds `clear` operation to `DatabaseStorage` class.
- Makes `clearCache` operation asynchronous to handle async `DatabaseStorage.clear` function
- Moves cache clearing into logout function, instead of `getLogoutUri` (since `getLogoutUri` is other synchronous).

### DIFF
--- a/lib/msal-browser/src/app/ClientApplication.ts
+++ b/lib/msal-browser/src/app/ClientApplication.ts
@@ -666,9 +666,20 @@ export abstract class ClientApplication {
 
         try {
             this.emitEvent(EventType.LOGOUT_START, InteractionType.Redirect, logoutRequest);
+            
+            // create logout string and navigate user window to logout.
             const authClient = await this.createAuthCodeClient(serverTelemetryManager, logoutRequest && logoutRequest.authority);
-            // create logout string and navigate user window to logout. Auth module will clear cache.
             const logoutUri: string = authClient.getLogoutUri(validLogoutRequest);
+
+            // Clear caches
+            if (validLogoutRequest.account) {
+                // Clear given account.
+                this.browserStorage.removeAccount(AccountEntity.generateAccountCacheKey(validLogoutRequest.account));
+            } else {
+                // Clear all accounts and tokens
+                await this.browserStorage.clear();
+            }
+
             this.emitEvent(EventType.LOGOUT_SUCCESS, InteractionType.Redirect, validLogoutRequest);
 
             if (!validLogoutRequest.account || AccountEntity.accountInfoIsEqual(validLogoutRequest.account, this.getActiveAccount())) {
--- a/lib/msal-browser/src/cache/BrowserCacheManager.ts
+++ b/lib/msal-browser/src/cache/BrowserCacheManager.ts
@@ -11,6 +11,7 @@ import { BrowserStorage } from "./BrowserStorage";
 import { MemoryStorage } from "./MemoryStorage";
 import { IWindowStorage } from "./IWindowStorage";
 import { BrowserProtocolUtils } from "../utils/BrowserProtocolUtils";
+import { CryptoOps } from "../crypto/CryptoOps";
 
 /**
  * This class implements the cache storage interface for MSAL through browser local or session storage.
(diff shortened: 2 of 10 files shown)


lib/msal-browser/src/cache/BrowserCacheManager.ts
```diff
@@ -490,7 +491,9 @@ export class BrowserCacheManager extends CacheManager {
             }
         });
 
+        debugger;
         this.internalStorage.clear();
+        await (this.cryptoImpl as CryptoOps).clearCache();
```
jasonnutter, line 496: This is another example of why I think the pop functions should be in their own class, instead of in `CryptoOps`. I think I'll make a PR demonstating what that might look like.

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -666,9 +666,20 @@ export abstract class ClientApplication {
...
             const logoutUri: string = authClient.getLogoutUri(validLogoutRequest);
+
+            // Clear caches
+            if (validLogoutRequest.account) {
+                // Clear given account.
+                this.browserStorage.removeAccount(AccountEntity.generateAccountCacheKey(validLogoutRequest.account));
+            } else {
+                // Clear all accounts and tokens
+                await this.browserStorage.clear();
+            }
```
pkanher617, line 681: This is done in msal-common in the AuthorizationCodeClient.ts class, may be redundant to have it here also?

https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/df0a58b788daba179156cff6703d25d339a11f9a/lib/msal-common/src/client/AuthorizationCodeClient.ts#L121-L127

lib/msal-browser/src/cache/DatabaseStorage.ts
```diff
@@ -110,4 +110,28 @@ export class DatabaseStorage<T>{
...
+            await this.open();
+        }
+
+        return new Promise<void>((resolve: any, reject: any) => {
+            if (!this.db) {
+                return reject(BrowserAuthError.createDatabaseNotOpenError());
+            }
+
+            const transaction = this.db.transaction([this.tableName], "readwrite");
+            const objectStore = transaction.objectStore(this.tableName);
```
jasonnutter, line 128: I think we should factor this boilerplate into a helper.

lib/msal-browser/src/crypto/CryptoOps.ts
```diff
@@ -153,4 +153,11 @@ export class CryptoOps implements ICrypto {
 
         return `${tokenString}.${encodedSignature}`;
     }
+
+    /**
+     * Clears crypto cache storage (IndexedDB)
+     */
+    async clearCache() {
```
jasonnutter, line 160: See above, this should really be a in pop-specific class.

lib/msal-browser/src/app/ClientApplication.ts
jasonnutter: That block will be removed.

lib/msal-browser/src/cache/DatabaseStorage.ts
```diff
@@ -110,4 +110,28 @@ export class DatabaseStorage<T>{
             dbPut.addEventListener("error", e => reject(e));
         });
     }
+
+    /**
+     * Remove all records from the database
+     */
+    async clear(): Promise<void> {
```
jasonnutter, line 117: Today, we don't use any accounts hints for storing things in indexeddb, maybe we should? That way we can properly clear records if the the app provides an account to the logout function.

jasonnutter: One easy way to do this would be to create separate DB's for each user.

lib/msal-browser/src/app/ClientApplication.ts
pkanher617: Assuming you mean from `msal-common`, make sure there are no side-effects for `msal-node`

jasonnutter: Correct, for sure.

lib/msal-browser/src/cache/DatabaseStorage.ts
hectormmg: I think it would be useful. At the moment we only use the `req_cnf` claim (for the keys at least). Maybe we can think up a cache key naming scheme like we have for artifacts cached in localStorage.

PR #3124, opened 2021-03-02
Add the correct type to `load()` method in IPersistence class

sameerag:
We have a mismatch in the return types for the function `load()` in `FilePersistence.ts` and other implementations of `IPersistence` header. This PR corrects it.


sadasant (approved):
Looks good to me!

(What are all of those "Unchanged files with check annotations"? I have never seen that before)

jasonnutter (approved):

hectormmg (approved):

PR #3144, opened 2021-03-04
Update react docs & samples with new packages

tnorling:
- Updates references to `msal-browser` and `msal-react` to `@azure/msal-browser` and `@azure/msal-react`
- Updates samples to use the new beta package
- Updates build/test workflow to not run when changefiles are added


jo-arroyo (approved):

.github/workflows/build-steps.yml
```diff
@@ -13,6 +13,7 @@ on:
       - '**.md'
       - '**/.vscode/*.json'
       - 'samples/**'
+      - 'change/**'
```
pkanher617, line 16: Are we sure we want to ignore change-files? If there is a new change-file shouldn't we assume that a build should be run?

tnorling: The presence of a changefile on its own should not trigger a build. The presence of changed files in the libs/samples/workflow should be what triggers a new build

tnorling: Sorry, not samples in this case, but in E2E case

pkanher617: Makes sense

pkanher617:
We should make sure to update the main page README and update any references to alpha to beta (for both React and Angular)

pkanher617 (approved):
One comment about references to `alpha`, otherwise looks good.

PR #3145, opened 2021-03-04
Update workflows to cancel in-progress runs

tnorling:
Adds a step to CI and E2E workflows to cancel in-progress and queued workflow runs on the same branch. This will help speed things up when you make several back to back pushes to a PR, ensuring that previous workflow runs aren't holding up the most recent push due to concurrency limits.


jasonnutter (approved):

sameerag (approved):
lgtm

Issue #2944, opened 2021-01-27
Handle network failure in acquireTokenSilent


k-mehta:
## Library

- [ ] `msal@1.x.x` or `@azure/msal@1.x.x`
- [x] `@azure/msal-browser@2.8.0`
- [ ] `@azure/msal-node@1.x.x`
- [ ] `@azure/msal-react@1.x.x`
- [ ] `@azure/msal-angular@0.x.x`
- [ ] `@azure/msal-angular@1.x.x`
- [ ] `@azure/msal-angular@2.x.x`
- [ ] `@azure/msal-angularjs@1.x.x`

## Framework

- [ ] Angular
- [x] React
- [ ] Other

## Description
It appears that MSAL is not gracefully handling the scenario where a token has expired and the user does not have internet connectivity. I am seeing a generic "TypeError: Failed to fetch" when I call PublicClientApplication.acquireTokenSilent(...).

## Error Message
Right before that error is thrown, I see the following in my console:

```
VM818:1 POST https://login.microsoftonline.com/consumers/oauth2/v2.0/token net::ERR_NAME_NOT_RESOLVED
(anonymous)	@	VM818:1
(anonymous)	@	index.es.js:8170
step	@	index.es.js:74
(anonymous)	@	index.es.js:55
(anonymous)	@	index.es.js:48
__awaiter	@	index.es.js:44
FetchClient.sendPostRequestAsync	@	index.es.js:8164
(anonymous)	@	index.es.js:3454
step	@	index.es.js:175
(anonymous)	@	index.es.js:156
(anonymous)	@	index.es.js:149
__awaiter$1	@	index.es.js:145
NetworkManager.sendPostRequest	@	index.es.js:3448
(anonymous)	@	index.es.js:3532
step	@	index.es.js:175
(anonymous)	@	index.es.js:156
(anonymous)	@	index.es.js:149
__awaiter$1	@	index.es.js:145
BaseClient.executePostToTokenEndpoint	@	index.es.js:3528
(anonymous)	@	index.es.js:5268
step	@	index.es.js:175
(anonymous)	@	index.es.js:156
fulfilled	@	index.es.js:146
Promise.then (async)		
step	@	index.es.js:148
(anonymous)	@	index.es.js:149
__awaiter$1	@	index.es.js:145
RefreshTokenClient.executeTokenRequest	@	index.es.js:5255
(anonymous)	@	index.es.js:5181
step	@	index.es.js:175
[...]

PR #3146, opened 2021-03-04
Add network error when fetch throws

tnorling:
When the fetch client fails due to network connectivity issues it throws a generic `TypeError`. This PR adds a slightly more meaningful error describing what happened.

Fixes: #2944


jasonnutter:
Is it possible to see if this was a CORS error? It would be nice to differentiate between CORS errors and no network errors, for example, and provide helpful error messages (esp for CORS, as not setting spa redirect is a common mistake).

lib/msal-common/src/network/NetworkManager.ts
```diff
@@ -31,7 +32,9 @@ export class NetworkManager {
      */
     async sendPostRequest<T>(thumbprint: RequestThumbprint, tokenEndpoint: string, options: NetworkRequestOptions): Promise<NetworkResponse<T>> {
         ThrottlingUtils.preProcess(this.cacheManager, thumbprint);
-        const response = await this.networkClient.sendPostRequestAsync<T>(tokenEndpoint, options);
+        const response = await this.networkClient.sendPostRequestAsync<T>(tokenEndpoint, options).catch(e => {
+            throw ClientAuthError.createNetworkError(tokenEndpoint, e);
+        });
```
jasonnutter, line 37: try/catch might be cleaner here

sameerag: +1

sameerag (commented):
I think this applies for any NW call right? Can we change the PR desc to reflect this?

sameerag:
> Is it possible to see if this was a CORS error? It would be nice to differentiate between CORS errors and no network errors, for example, and provide helpful error messages (esp for CORS, as not setting spa redirect is a common mistake).

Well ... we then need to track the error response codes. We can do it in `axios` but this will be platform specific code, we can handle it a layer up I suppose

tnorling:
> Is it possible to see if this was a CORS error? It would be nice to differentiate between CORS errors and no network errors, for example, and provide helpful error messages (esp for CORS, as not setting spa redirect is a common mistake).

I was hoping for this too but it seems that the CORS error is a separate error thrown by the browser, that indirectly causes the Fetch client to fail with `TypeError: Failed to fetch`. Haven't been able to find a way to catch that.

sameerag:
> > Is it possible to see if this was a CORS error? It would be nice to differentiate between CORS errors and no network errors, for example, and provide helpful error messages (esp for CORS, as not setting spa redirect is a common mistake).
> 
> I was hoping for this too but it seems that the CORS error is a separate error thrown by the browser, that indirectly causes the Fetch client to fail with `TypeError: Failed to fetch`. Haven't been able to find a way to catch that.

We can do it in the platform response code -> where we return the response, we can catch and throw the exact error

tnorling:
> > > Is it possible to see if this was a CORS error? It would be nice to differentiate between CORS errors and no network errors, for example, and provide helpful error messages (esp for CORS, as not setting spa redirect is a common mistake).
> > 
> > 
> > I was hoping for this too but it seems that the CORS error is a separate error thrown by the browser, that indirectly causes the Fetch client to fail with `TypeError: Failed to fetch`. Haven't been able to find a way to catch that.
> 
> We can do it in the platform response code -> where we return the response, we can catch and throw the exact error

I'm saying that `TypeError: Failed to fetch` **is** the exact error thrown by the platform network manager.

sameerag:
> > > > Is it possible to see if this was a CORS error? It would be nice to differentiate between CORS errors and no network errors, for example, and provide helpful error messages (esp for CORS, as not setting spa redirect is a common mistake).
> > > 
> > > 
> > > I was hoping for this too but it seems that the CORS error is a separate error thrown by the browser, that indirectly causes the Fetch client to fail with `TypeError: Failed to fetch`. Haven't been able to find a way to catch that.
> > 
> > 
> > We can do it in the platform response code -> where we return the response, we can catch and throw the exact error
> 
> I'm saying that `TypeError: Failed to fetch` **is** the exact error thrown by the platform network manager.

got it!

lib/msal-common/src/error/ClientAuthError.ts
```diff
@@ -225,6 +229,14 @@ export class ClientAuthError extends AuthError {
             `${ClientAuthErrorMessage.endpointResolutionError.desc} Detail: ${errDetail}`);
     }
 
+    /**
+     * Creates an error thrown when the fetch client throws
+     */
+    static createNetworkError(endpoint: string, errDetail: string): ClientAuthError {
+        return new ClientAuthError(ClientAuthErrorMessage.networkError.code,
+            `${ClientAuthErrorMessage.networkError.desc} Detail: ${errDetail} Attempted to reach: ${endpoint}`);
```
pkanher617, line 237: the `endpoint` part might be considered PII if it has something like loginHint in it.

tnorling: Right. This should just be the token endpoint which doesnt have a query string, but it may be worth stripping that off just in case

pkanher617: Ah yeah that's true. Probably worth stripping off for now (or truncating it?)

pkanher617 (approved):
One comment about error message, also maybe use try/catch instead of .catch, but otherwise looks good.

lib/msal-common/src/network/NetworkManager.ts
tnorling: Done

lib/msal-common/src/error/ClientAuthError.ts
tnorling: Done

pkanher617 (approved):

jasonnutter:
Do we need to add similar logic to the xhr client?

sameerag:
I can take up `axios` throwing or handling errors similarly with a separate PR. `XHRClient` may be should be addressed with this one.

lib/msal-browser/src/network/FetchClient.ts
```diff
@@ -37,11 +38,21 @@ export class FetchClient implements INetworkModule {
...
+        } catch (e) {
+            if (window.navigator.onLine) {
+                throw BrowserAuthError.createNetworkRequestFailedError(e, url);
+            } else {
+                throw BrowserAuthError.createNoNetworkConnectivityError();
+            }
+        }
         return {
             headers: this.getHeaderDict(response.headers),
             body: await response.json() as T,
```
jasonnutter, line 58: I'm curious if we should move this part into the try/catch, so that we can gracefully capture json parsing errors?

tnorling: Maybe a separate try catch so we can distinguish between parsing errors vs. network/CORS errors?

jasonnutter: Yeah that sounds good.

lib/msal-browser/src/network/FetchClient.ts
```diff
@@ -22,11 +23,16 @@ export class FetchClient implements INetworkModule {
             method: HTTP_REQUEST_TYPE.GET,
             headers: this.getFetchHeaders(options)
```
jasonnutter, line 24: We still want to wrap this in a try/catch, right?

tnorling: Done

jasonnutter (approved):

PR #3147, opened 2021-03-05
Uses sid from account if not present in request

pkanher617:
Fixes an issue where the app would not use sid from token claims. This should fix issues in `acquireTokenSilent` for `msal-browser` when the refresh token has expired, which would only send login_hint today when falling back to acquireTokenByIframe, resulting in inadvertent timeouts.


tnorling (approved):

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -248,10 +250,19 @@ export class AuthorizationCodeClient extends BaseClient {
         // Add sid or loginHint with preference for sid -> loginHint -> username of AccountInfo object
         if (request.sid) {
             parameterBuilder.addSid(request.sid);
+        } else if (request.account) {
+            const accountSid = this.extractAccountSid(request.account);
+            // If account and loginHint are provided, we will check account first for sid before adding loginHint
+            if (accountSid) {
+                parameterBuilder.addSid(accountSid);
+            } else if (request.loginHint) {
```
jasonnutter, line 258: Is this needed, given the same check is done below (265)?

pkanher617: It will be skipped if account and loginHint are both passed in the request, which is why I included it here. Open to suggestions for cleaning this up.

jasonnutter: Ah right, makes sense.

jasonnutter (approved):

lib/msal-browser/docs/login-user.md
```diff
@@ -97,7 +97,13 @@ These APIs will return an account object or an array of account objects with the
 
 ## Silent login with ssoSilent()
 
-If you already have a session that exists with the authentication server, you can use the ssoSilent() API to make request for tokens without interaction. You will need to pass a `login_hint` (which can be retrieved from the account object `username` property or the `upn` claim in the ID token) in the request object in order to successfully obtain a token silently.
+If you already have a session that exists with the authentication server, you can use the ssoSilent() API to make request for tokens without interaction. You will need to pass one of the following into the request object in order to successfully obtain a token silently.
+
+- `login_hint` (which can be retrieved from the account object `username` property or the `upn` claim in the ID token)
```
sameerag, line 102: can we please reverse this order? For perception of precedence.

sameerag (approved):
Approved with one minor comment!

lib/msal-browser/docs/login-user.md
pkanher617: so account, sid, login_hint?

pkanher617: done.

Issue #3149, opened 2021-03-05
prod build error for msal-angular in custom library


sidyes:
## Library

- [ ] `msal@1.x.x` or `@azure/msal@1.x.x`
- [x] `@azure/msal-browser@2.12.0`
- [ ] `@azure/msal-node@1.x.x`
- [ ] `@azure/msal-react@1.x.x`
- [ ] `@azure/msal-angular@0.x.x`
- [ ] `@azure/msal-angular@1.x.x`
- [x] `@azure/msal-angular@2.0.0-beta.0`
- [ ] `@azure/msal-angularjs@1.x.x`

## Framework

- [x] Angular
- [ ] React
- [ ] Other

## Description
I have integrated the msal angular solution into a library of a NX workspace. It is working just fine, when I use my library (that kind of wraps the msal lib) within an application.
But when I try to make a production build of this library, the build fails.

## Error Message

```
Building Angular Package

------------------------------------------------------------------------------
Building entry point 'my-shared-auth-lib'
------------------------------------------------------------------------------
× Compiling TypeScript sources through NGC
ERROR: Cannot determine the module for class MsalRedirectComponent in C:/path/to/project/node_modules/@azure/msal-angular/azure-msal-angular.d.ts! Add MsalRedirectComponent to the NgModule to fix it.

Cannot determine the module for class MsalRedirectComponent in C:/path/to/project/node_modules/@azure/msal-angular/azure-msal-angular.d.ts! Add MsalRedirectComponent to the NgModule to fix it.
```

## MSAL Configuration

```js
@NgModule({
  imports: [MsalModule, StoreModule],
  providers: [AzureAuthService, MsalService, MsalGuard, MsalBroadcastService]
})
export class SharedAzureAuthModule {
  static forRoot(
    azureConfig: AzureConfig
  ): ModuleWithProviders<SharedAzureAuthModule> {
    return {
      ngModule: SharedAzureAuthModule,
      providers: [
        {
          provide: HTTP_INTERCEPTORS,
          useClass: MsalInterceptor,
          multi: true,
        },
        {
          provide: MSAL_INSTANCE,
          useValue: getMsalInstanceConfig(azureConfig.msalInstance), // basically same setup as in the sample app but parametrized
        },
        {
          provide: MSAL_INTERCEPTOR_CONFIG,
          useValue: getMsalInterceptorConfig(azureConfig.msalInterceptor), // basically same setup as in the sample app but parametrized
        },
        {
[...]

PR #3161, opened 2021-03-08
fix: #3149 - Msal Redirect Component was not part of the Module

roopeshreddy:
This PR fix the dangling component added to the NgModule. As per the Angular docs, any dangling component which is not part of the module or referenced in any other component will be removed during the tree shaking.


lib/msal-angular/src/msal.module.ts
```diff
@@ -12,9 +12,10 @@ import { MsalGuard } from "./msal.guard";
 import { MsalBroadcastService } from "./msal.broadcast.service";
 import { MsalService } from "./msal.service";
 import { MSAL_INSTANCE , MSAL_GUARD_CONFIG, MSAL_INTERCEPTOR_CONFIG } from "./constants";
+import { MsalRedirectComponent } from './msal.redirect.component';
```
pkanher617, line 15: Change the `'` to `"` so the linter isn't erroring out.

jasonnutter:
@roopeshreddy Thanks for the PR! Can you please address the lint errors, and also run `npm run beachball:change` from the root of the repo, and complete the browsers and push the generated change files?

@jo-arroyo Can you please test this PR locally?

lib/msal-angular/src/msal.module.ts
roopeshreddy: This is taken care.

jasonnutter:
@roopeshreddy Can you discard/undo the changes to the package-lock files?

roopeshreddy:
> @roopeshreddy Can you discard/undo the changes to the package-lock files?

Done. I've reverted the package-local.json files

tnorling (approved):

Issue #3085, opened 2021-02-25
Configure B2C sample to work with PathLocationStrategy instead of HashLocationStrategy


shadyabouelmakarem:
## Library
- [x] `@azure/msal-angular@2.0.0-alpha.5`

## Description
How can I replicate the `angular11-b2c-sample` project without turning on the `HashLocationStrategy` ?

The issue I'm facing is that redirect URIs from Azure AD B2C always start with `#` which misses up the application because I'm using `PathLocationStrategy` instead of `HashLocationStrategy`.

I can set the `useHash: false` instead of `useHash: true` like this:

> RouterModule.forRoot(routes, {
      useHash: false,
      // Don't perform initial navigation in iframes
      initialNavigation: !isIframe ? 'enabled' : 'disabled',
    }),

But still, how can I interact with the redirect URIs starting with `#` ?

## Source

- [x] Customer request

jo-arroyo:
@shadyabouelmakarem `PathLocationStrategy` is fully supported by msal-angular. However, MSAL does not support redirectUris that have a hash, because that is where the response from the server is. MSAL will automatically handle parsing the response hash for you. 

> But still, how can I interact with the redirect URIs starting with `#` ?

Could you clarify what you mean by this?

eluchsinger:
@jo-arroyo  The sample uses HashLocation strategy (see [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/df0a58b788daba179156cff6703d25d339a11f9a/samples/msal-angular-v2-samples/angular11-b2c-sample/src/app/app-routing.module.ts#L40)). How is PathLocationStrategy supported? I have also pointed this out [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/2885#pullrequestreview-573056409).

jo-arroyo:
@eluchsinger The PathLocationStrategy should be supported by the library. The samples show the HashLocationStrategy, but you are able to turn that off or remove it from the `RouterModule.forRoot`. Are you having problems using the PathLocationStrategy with our library?

eluchsinger:
@jo-arroyo I'm having the problem that after logging in, it does not redirect correctly  to the protected page I tried to access. 

[...]

Issue #3047, opened 2021-02-19
Auto Redirect for login


dmccolloughOneGas:
## Library

- [ ] `msal@1.x.x` or `@azure/msal@1.x.x`
- [X] `@azure/msal-browser@2.11.0`
- [ ] `@azure/msal-node@1.x.x`
- [ ] `@azure/msal-react@1.x.x`
- [ ] `@azure/msal-angular@0.x.x`
- [ ] `@azure/msal-angular@1.x.x`
- [X] `@azure/msal-angular@2.0.0-alpha.6`
- [ ] `@azure/msal-angularjs@1.x.x`

## Description

I have followed [this repos](https://github.com/Azure-Samples/ms-identity-javascript-angular-spa-aspnetcore-webapi).
In fact. I downloaded the code, added a controller to the API and am making calls to the new controller.

I would like to know if it's possible in the Angular App to not have the user click on a login button to redirect to the Microsoft Login Page?

I have attempted this on my own in the angular app, by doing the following: 
In the app.component.ts file, in the ngOnInit() method. I added a call to this.login(); above the line this.checkAccount();

The angular app, will auto redirect to the Microsoft Login Page automatically, I can login and it will take me to the default page based on the routing map.

The odd behavior I see is once I'm on the default page if I click the browser refresh button, the page will just automatically start a refresh loop. It will do this refresh until you stop it. When it does this, I see in the console the error thats described in this issue: #3042 

I have read through the comments and see where a MsalRedirectComponent has been introduced, would this solve my problem? 

If so, can I get a link to a sample that uses the MsalRedirectComponent or the documentation for it?
The comments in 3042 mention a linked sample page, but I don't see a link.

I'm attaching my modified sample project for review.

Ideally I do not want the user to have to click a login button to get to the the Microsoft Login Page

Thanks

[ModifiedSampleApp.zip](https://github.com/AzureAD/microsoft-authentication-library-for-js/files/6011094/ModifiedSampleApp.zip)

jo-arroyo:
@dmccolloughOneGas Thanks for your question and for providing your sample app. I will be investigating this issue and will get back to you early next week.

jo-arroyo:
@dmccolloughOneGas Thanks for sending your sample app. I tried it out and was able to reproduce the issue. Our recommendations are as follows: 

[...]

PR #3174, opened 2021-03-09
Add MsalCustomNavigationClient and guidance for routing strategy for msal-angular

jo-arroyo:
This PR adds a `MsalCustomNavigationClient` for client-side routing and adds a sample and documentation regarding using `PathLocationStrategy` and `HashLocationStrategy` with logging on page load.

msal-angular
- Adds and exports `MsalCustomNavigationClient` to msal-angular
- Updates the `MsalRedirectComponent` with a verbose log message

Samples
- Updates `Angular10-sample-app` to use `PathLocationStrategy` and `MsalCustomNavigationClient`
- Updates samples README with each sample's routing strategy

Docs
- Updates performance docs regarding `MsalCustomNavigationClient` usage
- Updates FAQ with guidance on protecting initial route with `MsalGuard` for invoking login when app initializes
- Updates public-apis doc with `MsalRedirectComponent` and `MsalCustomNavigationClient`

This PR addresses issues: #3047, #3085, #3293.


lib/msal-angular/src/msal.navigation.client.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+export class MsalCustomNavigationClient extends NavigationClient {
+    private router: Router;
+
+    constructor(router: Router) {
+        super();
+        this.router = router;
+    }
+
+    async navigateInternal(url:string, options: NavigationOptions): Promise<boolean> {
```
jo-arroyo, line 22: Do we need these `NavigationOptions` here? I know they're part of msal-browser's `NavigationClient`, but they are failing the linter here because they're not being used.

jasonnutter: I don't think you should need to provide it if its not used? Does typescript complain if you leave it out?

lib/msal-angular/src/msal.navigation.client.ts
```diff
@@ -0,0 +1,34 @@
...
+        const urlComponents = new UrlString(url).getUrlComponents();
+        
+        // Ensures url does not end with / character
+        const joinedPathSegments = urlComponents.PathSegments.join("/");
+
+        // Adds query to url if present
+        const newUrl = urlComponents.QueryString ? `/${joinedPathSegments}?${urlComponents.QueryString}` : `/${joinedPathSegments}`;
```
jasonnutter, line 29: For the guard, the angular location service provided some helpful apis to normalize urls. Do those help at all here? 

https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/6650148dae23ec568619a4ed8fe2a9077d1e0e8c/lib/msal-angular/src/msal.guard.ts#L44

lib/msal-angular/src/msal.navigation.client.ts
tnorling: You should use at least the history flag

lib/msal-angular/src/msal.navigation.client.ts
jo-arroyo: Good idea, I adjusted it to use the Location service instead

lib/msal-angular/src/msal.navigation.client.ts
jo-arroyo: Noted. `navigateUrl` now uses `noHistory` flag.

jasonnutter (approved):

tnorling (approved):

PR #3188, opened 2021-03-11
Don't migrate tokens for different clientid

### LINKED ISSUE #3172: MSALGuard Does Not Validate Token Audiance
## Library
@azure/msal-angular 1.1.1

## Framework
Angular 11

## Description

We have multiple websites on the same domain running under Kuberneties with SSO, when we switch between websites the token is not refreshed and it tries to use the same token, probably as they are in the same domain, for example

test.xx.yy/webapp1
test.xx.yy/webapp2

The Azure clientids are different for both websites and MSAL Angular should check the audiance (aud) claim of the token against the clientid of the application and refresh the token if they are different. This is what we have done creating our own guard which compares the aud and clientid and if different call login which results in a silent refresh getting the appropriate token. We keep roles in the token and this causes an issue as without the correct token, the token does not hold the correct roles.

## Error Message
 None

## Reproduction steps

Create two websites under the same domain, switch between the websites and in the token check the aud matches the clientid of the website you have navigated to.

```

## Expected behavior
 The token should be for the right website, the aud claim of the token should match the Azure clientid of the website

## Identity Provider

-] Azure AD


## Browsers/Environment

-  Chrome

## Security

- [Yes ] Is this issue security related?

### PR DESCRIPTION
**Pull request by tnorling**
Msal 1.x has generic cache entries `msal.id_token` etc. that does not distinguish different clientIds for legacy purposes. Migration code copies these cache entries into `msal.clientId.id_token` etc. without first checking if the cached tokens match the clientId for the current app. This causes a bug in apps that have multiple clientIds on a single domain where tokens can cross app boundaries and cause confusing errors.

This PR fixes this by comparing the cached generic id_token aud claim to the current clientId and determines whether or not the generic cache entries should be migrated.

Fixes #3172

### CODE REVIEW

File: lib/msal-core/src/cache/AuthCache.ts [hunk 1]
```diff
@@ -39,14 +40,27 @@ export class AuthCache extends BrowserStorage {// Singleton
...
-        const errorDescValue = super.getItem(errorDescKey);
 
-        const values = [idTokenValue, clientInfoValue, errorValue, errorDescValue];
-        const keysToMigrate = [PersistentCacheKeys.IDTOKEN, PersistentCacheKeys.CLIENT_INFO, ErrorCacheKeys.ERROR, ErrorCacheKeys.ERROR_DESC];
+        let idToken;
 
-        keysToMigrate.forEach((cacheKey, index) => this.duplicateCacheEntry(cacheKey, values[index], storeAuthStateInCookie));
+        if (idTokenValue) {
+            try {
+                idToken = new IdToken(idTokenValue);
```
jasonnutter, line 48: What would cause this to throw?

jasonnutter (approved):

lib/msal-core/src/cache/AuthCache.ts
tnorling: Potentially if the raw idToken string cannot be decoded and/or JSON parsed

pkanher617 (approved):

McInnesIan:
When will this be published as npm package so I can use it and remove my fix

tnorling:
@McInnesIan This was released yesterday in version 1.4.8. [Release Notes](https://github.com/AzureAD/microsoft-authentication-library-for-js/releases/tag/msal-v1.4.8)

PR #3202, opened 2021-03-12
Propagate request scopes to response handler

dopsonbr:
My OAuth provider doesn't always return `scopes` in the token response. When scopes are returned MSAL errors. When troubleshooting the errors I noticed the below comment in ResponseHandler.ts.
``` typescript
// If scopes not returned in server response, use request scopes
const responseScopes = serverTokenResponse.scope ? ScopeSet.fromString(serverTokenResponse.scope) : new ScopeSet(requestScopes || []);
```
My change propagates the request scopes so they can be used when the response doesn't include them.


tnorling:
@dopsonbr Thanks for calling out this gap. This will be addressed in #3089 where a change will be made to pass the entire request object down to `handleServerTokenResponse`. Please track that PR for resolution, thanks!

PR #3203, opened 2021-03-12
Change default headers when authority uses oidc protocol mode

dopsonbr:
Not all OAuth providers support non-standard headers on CORS requests. This change removes the non-standard headers when the authority is configured in OIDC protocol mode. The AAD behavior is unchanged. 

Another way to implement this would be to make the pre-flight OPTIONS check to the authority and inspect the `Access-Control-Allow-Headers` response. Then set the default headers based on that. I'm open to implementing this in a different way or another way that allows library consumers to override the default headers.

### DIFF
--- a/lib/msal-common/src/client/BaseClient.ts
+++ b/lib/msal-common/src/client/BaseClient.ts
@@ -3,19 +3,20 @@
  * Licensed under the MIT License.
  */
 
-import { ClientConfiguration, buildClientConfiguration, CommonClientConfiguration } from "../config/ClientConfiguration";
-import { INetworkModule } from "../network/INetworkModule";
-import { NetworkManager, NetworkResponse } from "../network/NetworkManager";
-import { ICrypto } from "../crypto/ICrypto";
-import { Authority } from "../authority/Authority";
-import { Logger } from "../logger/Logger";
-import { AADServerParamKeys, Constants, HeaderNames } from "../utils/Constants";
-import { ServerAuthorizationTokenResponse } from "../response/ServerAuthorizationTokenResponse";
-import { CacheManager } from "../cache/CacheManager";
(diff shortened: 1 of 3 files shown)


dopsonbr:
Is there a project formatter I can run? I didn't see one and didn't mean to change the formatting of the imports

jasonnutter:
> Is there a project formatter I can run? I didn't see one and didn't mean to change the formatting of the imports

If you run `npm install` in the root, you can run linting as defined in the root `package.json` for each lib.

jasonnutter:
Thanks for the PR! We'll take a look.

> Another way to implement this would be to make the pre-flight OPTIONS check to the authority and inspect the Access-Control-Allow-Headers response. Then set the default headers based on that. I'm open to implementing this in a different way or another way that allows library consumers to override the default headers.

We're actually looking at removing the need for the OPTIONS request, so that approach will likely not be feasable.

Also, please run `npm run beachball:change` from the root of the repo and follow the prompts.

tnorling:
@dopsonbr There are potentially use cases where these headers should be sent with the OIDC protocol mode (like going to login.live.com directly instead of going through the login.microsoftonline.com/common endpoint) One way you can override the default headers today is to implement your own Network Client and pass it into the `PublicClientApplication` constructor using [this param](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/src/config/Configuration.ts#L72). This will give you full control over the headers that are sent on network requests.

And as @jasonnutter said, we are planning to remove the OPTIONS request as soon as some server changes are completed to support it. The PR for that is #3094

tnorling:
Closing as we'll be removing the OPTIONS request soon and there's a workaround until then.

PR #3229, opened 2021-03-17
[msal-node] Fix the device code cancellation

samuelkubai:
Use `Object.assign()` instead of object de-structuring to ensure the request object remains linked to the original request object. To allow for the cancellation flag.


samples/msal-node-samples/standalone-samples/device-code/index.js
```diff
@@ -18,6 +18,7 @@ const deviceCodeRequest = {
     deviceCodeCallback: (response) => (console.log(response.message)),
     scopes: ["user.read"],
     timeout: 20,
+    cancel: false,
```
sameerag, line 21: So how do we make cancel=true? What is the usage? When does an app know it should not poll anymore?

samuelkubai: Cancel is designed to be set `true` after executing this request, on this sample you can find that scenario [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/74f2dfce8804e25035cd3ba21d195530a3fba450/samples/msal-node-samples/standalone-samples/device-code/index.js#L30) where we set a timeout to set the `cancel` to `true` after a few seconds.

samuelkubai: The usage is when the request is made and we are polling, at any point after polling has begun the user of the library can program a cancel by changing the flag from `false` to `true`

jasonnutter:
Fix looks good, is it possible to add a test that demonstrates the bug and fix?

samuelkubai:
Thanks for pointing it out @jasonnutter, I'll work on adding a test for this fix.

samuelkubai:
@jasonnutter, I'm having a few issues adding the necessary e2e test and thus propose to merge this pull request to unblock the dependents while I work on adding the test scenario.

jasonnutter:
> @jasonnutter, I'm having a few issues adding the necessary e2e test and thus propose to merge this pull request to unblock the dependents while I work on adding the test scenario.

I see, no problem. Any unit tests you can add?

jasonnutter:
> > > @jasonnutter, I'm having a few issues adding the necessary e2e test and thus propose to merge this pull request to unblock the dependents while I work on adding the test scenario.
> > 
> > 
> > I see, no problem. Any unit tests you can add?
> 
> @jasonnutter I think this use case is best covered with a e2e test

Makes sense.

jasonnutter (approved):

sameerag (approved):
lgtm

ramya-rao-a:
Hello!
Do we know when this fix will be released?

cc @sadasant

sameerag:
@ramya-rao-a We plan to make a release this week, waiting on one more PR. Mostly in a day or two.

PR #3234, opened 2021-03-17
Properly handle expiration timestamps when returned as strings

jasonnutter:
There has been at least one customer report and repro of the B2C token endpoint returning all properties as strings, including timestamps. Example:

```json
{
  "access_token":"eyJ0e...",
  "id_token":"eyJ0eX...",
  "token_type":"Bearer",
  "not_before":"1615993049",
  "expires_in":"3600",
  "expires_on":"1615996649",
  "resource":"resource-id-guid",
  "refresh_token":"eyJraWQiOiJ...",
  "scope":"scope1 scope2"
}
```

The MSAL response handling code is not expecting this, and as a result, the calculated expiration timestamps are invalid. When these tokens expire, MSAL still things they are valid, which results in MSAL returning already expired tokens back to applications. This fix adds logic to parse these values to numbers if they are returned as strings.


lib/msal-common/test/client/AuthorizationCodeClient.spec.ts
```diff
@@ -583,6 +583,84 @@ describe("AuthorizationCodeClient unit tests", () => {
...
+                authenticationScheme: AuthenticationScheme.BEARER
+            };
+
+            const authenticationResult = await client.acquireToken(authCodeRequest, {
+                code: authCodeRequest.code,
+                nonce: idTokenClaims.nonce,
+                state: testState
+            });
+
+            expect(authenticationResult.expiresOn?.toString()).to.not.equal("Invalid Date");
```
jasonnutter, line 662: Should add assertion for extExpiresOn as well

jasonnutter:
FYI @nickgmicrosoft I will file an issue with the B2C on this issue shortly.

pkanher617 (approved):

pkanher617 (approved):

jasonnutter:
Update: this issue was due to teh customer not properly setting the `SendTokenResponseBodyWithJsonNumbers` property in their custom policy, but we'll keep PR open just in case other customers run into a similar problem.

tnorling (approved):

PR #3249, opened 2021-03-19
Block nested popups

### LINKED ISSUE #3160: about:blank popup intermittently stays on on acquireTokenSilent/acquireTokenPopup
## Library

- [X] `@azure/msal-browser@2.x.x`
- [X] `@azure/msal-react@1.x.x`

## Framework

React 16.4
- [ ] Angular
- [X] React
- [ ] Other

## Description
Hi Folks,

I'm using the libraries with the following versions:
```
    "@azure/msal-browser": "^2.12.0",
    "@azure/msal-react": "^1.0.0-beta.0"
```

Post logout redirect URL is `${window.location.origin}/` and redirectUrl is `redirectUri`
The as you can see seems to be failing intermittently. See the attached recording where all is working fine for several times until around the 50th second mark of the video the about:blank window stays hanging.
[Video clip recreating the issue](https://photos.app.goo.gl/A8oK5NQCPj21hYJr6)
Any idea what I'm might be doing wrong?
Thank you for the otherwise rock solid library.

Might be related to: https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/2842

FYI: @tnorling , @jo-arroyo , @jasonnutter , @pkanher617 
_Originally posted by @gabriel-kohen-by in https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/2842#issuecomment-792864888_
## MSAL Configuration

```js
const msalProviderConfig = {
  type: 'popup',
  msalConfig: {
    auth: {
      clientId,
      authority,
      redirectUri,
      postLogoutRedirectUri,
    },
    system: {
      loggerOptions: {
        loggerCallback: getLoggerCallback(),
      },
    },
  },
  silentRequestConfig: {
    scopes: [scopes, 'offline_access'],
  },
  endSessionRequestConfig: {},
  loginRequestConfig: {
    scopes: [scopes, 'offline_access'],
  },
};
```

## Reproduction steps
Open a SPA page triggering getting of a token with 

## Expected behavior
Close any popup if we resolved the promise coming back from asking for an access token.

## Identity Provider

- [ ] Azure AD
- [ ] Azure B2C Basic Policy
- [X] Azure B2C Custom Policy
- [ ] ADFS
- [ ] Other

## Browsers/Environment

- [X] Chrome
- [ ] Firefox
- [ ] Edge
- [X] Safari
- [ ] IE

## Regression

- [X ] Did this behavior work before?
Version:
MSAL 1.X

## Security

## Source

- [ ] Internal (Microsoft)
- [X] Customer request

tnorling:
When the page used as the `redirectUri` invokes `loginPopup` on page load it can sometimes open another popup that never resolves. This PR adds a check to the `preflightBrowserEnvironmentCheck` to see if it's already in a popup and blocks subsequent requests. 

Fixes #3160


tnorling:
@jasonnutter Is this going to break any valid use cases @pkanher617 and I didn't consider?

lib/msal-browser/src/utils/BrowserUtils.ts
```diff
@@ -100,6 +100,15 @@ export class BrowserUtils {
         }
     }
 
+    /**
+     * Block redirectUri loaded in popup from calling AcquireToken APIs
+     */
+    static blockAcquireTokenInPopups(): void {
+        if (window.opener && window.opener !== window) {
```
jasonnutter, line 107: Two things:

1. I think you should pass in the interaction type, and only make this check if a popup request is being made, as redirect operations inside a popup should be allowed.
2. Make sure this test does result positives. Just beacuse a window has an opener doesnt mean its a popup.

jasonnutter: Yeah, this check caused problems in v1, and was part of the reason we put the interaction type in the state value. See https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/1413

jasonnutter (changes requested):
Please find a more reliable check for popups, and ensure this only blocks popup requests.

lib/msal-browser/src/utils/BrowserUtils.ts
```diff
@@ -100,6 +100,16 @@ export class BrowserUtils {
         }
     }
 
+    /**
+     * Block redirectUri loaded in popup from calling AcquireToken APIs
+     */
+    static blockAcquireTokenInPopups(): void {
+        // Popups opened by msal popup APIs are given a name that starts with "msal."
+        if (window.opener && window.opener !== window && window.name.indexOf(`${BrowserConstants.POPUP_NAME_PREFIX}.`) === 0) {
```
jasonnutter, line 108: Might be good to check truthiness of `window.name` to be safe.
```suggestion
        if (window.opener && window.opener !== window && typeof window.name === "string" && window.name.indexOf(`${BrowserConstants.POPUP_NAME_PREFIX}.`) === 0) {
```

jasonnutter (approved):

pkanher617 (approved):

PR #3258, opened 2021-03-19
Add passport-azure-ad code

jasonnutter:


jasonnutter:
@pkanher617 Since are leaning towards keeping the name, would you rather this go in the `lib` folder? Or would it be better to have the old/current version in the `maintenance` folder?

tnorling (approved):

sameerag:
> @pkanher617 Since are leaning towards keeping the name, would you rather this go in the `lib` folder? Or would it be better to have the old/current version in the `maintenance` folder?

I prefer `maintenance`

sameerag (approved):

Issue #3253, opened 2021-03-19
@azure/msal-angular - MsalBroadcastService: inProgress$ does not emit last value


mdarefull:
## Library

- @azure/msal-browser": "^2.12.1"
- @azure/msal-angular": "^2.0.0-beta.1",

## Framework

- [ X] Angular
- [ ] React
- [ ] Other

## Description
In the documentation for [handling redirects](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-angular/docs/v2-docs/redirects.md) it is recommended to use `MsalRedirectComponent `to handle the redirects and `MsalBroadcastService.inProgress$` for, quote `Pages that wish to perform functions following redirects...`

The problem is that inProgress$ won't emit the last value, so if the page subscribes after the redirect has been handled by MsalRedirectComponent, the subscription will never execute.
It is a nuance mainly because of race conditions, I cannot ensure a given component will subscribe before/after the redirects are handled causing the code to no execute. Not using the subscription means the opposite, the possibility for the code to run while the interaction is still in progress.

## Identity Provider

- [ ] Azure AD
- [X ] Azure B2C Basic Policy
- [ ] Azure B2C Custom Policy
- [ ] ADFS
- [ ] Other

jo-arroyo:
@mdarefull Thanks for your feedback. We are aware this has been an issue for our customers, and a fix with updated documentation scheduled for the next release.

PR #3269, opened 2021-03-19
Update inProgress observable and MsalGuard in msal-angular

jo-arroyo:
Previously, when (a) using path routing, and (b) when the user attempts to access a protected route upon first using the app, the `MsalGuard` is run before any eventCallbacks are registered with `msal-browser`. This meant that events in `MsalGuard` were not being captured by the eventCallback or the `inProgress$` observable, which in turn meant that components dependent on the `inProgress$` observable did not have access to all updates. This did not appear to be an issue with hash routing.

This PR:
- Updates the `MsalGuard` to subscribe to the `inProgress$` observable, allowing events in the `MsalGuard` to be captured
- Changes the `_inProgress` Subject to a `BehaviorSubject`, allowing the most recent `InteractionType` to be available upon subscription
- Adds unit tests for `inProgress$` observable
- Updates docs re. `inProgress$` observable
- Updates docs re. `MsalGuard` being a convenience feature
- Updates docs re. `MsalRedirectComponent` or `handleRedirectObservable` being mandatory when using redirects

Future work:
- Investigate whether the subject type `msalSubject$` requires a change, and whether `MsalGuard` should subscribe to the `msalSubject$` observable too.

This PR addresses concerns raised in issue #3253


lib/msal-angular/src/msal.broadcast.service.ts
```diff
@@ -22,8 +22,11 @@ export class MsalBroadcastService {
     ) {
         this._msalSubject = new Subject<EventMessage>();
         this.msalSubject$  = this._msalSubject.asObservable();
-        this._inProgress = new Subject<InteractionStatus>();
+
+        // InProgress as BehaviorSubject so most recent inProgress state will be available upon subscription
+        this._inProgress = new BehaviorSubject<InteractionStatus>(undefined);
```
jo-arroyo, line 27: Unsure whether to set it to undefined, null, or `Interaction.None`. Thoughts?

lib/msal-angular/docs/FAQ.md
```diff
@@ -79,7 +87,7 @@ Our [Angular 11](https://github.com/AzureAD/microsoft-authentication-library-for
...
 
-We recommend that all redirects are handled with the `MsalRedirectComponent` (see docs [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev/lib/msal-angular/docs/v2-docs/redirects.md)), and that any interaction or account validation should be done after  subscribing to the `inProgress$` subject and filtering for none. Please see our [sample](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/30b5ff95e2ff2cc827d98118004d92968bb67b3f/samples/msal-angular-v2-samples/angular11-sample-app/src/app/app.component.ts#L27) for an example. 
+We recommend that all redirects are handled with the `MsalRedirectComponent` (see docs [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev/lib/msal-angular/docs/v2-docs/redirects.md)), and that any interaction or account validation should be done after  subscribing to the `inProgress$` observable and filtering for `InteractionStatus.None`. Please see our [sample](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/30b5ff95e2ff2cc827d98118004d92968bb67b3f/samples/msal-angular-v2-samples/angular11-sample-app/src/app/app.component.ts#L27) for an example.
```
tnorling, line 90: I think it may be worth using stronger language than "recommend" here. I read some feedback that said it wasn't clear that it was mandatory to handle redirects explicitly either with `MsalRedirectComponent` or `handleRedirectObservable`. We can recommend `MsalRedirectComponent` as the best approach but we should be very clear that if you do not, you **must** call `handleRedirectObservable`

lib/msal-angular/docs/v2-docs/errors.md
```diff
@@ -25,7 +25,7 @@ In `@azure/msal-angular` there are 2 common scenarios when this can happen:
 
 We recommend that you bootstrap the the **`MsalRedirectComponent`** if you want to use redirects as this will handle all redirects for you. See the docs on redirects [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev/lib/msal-angular/docs/v2-docs/redirects.md) for how to use the `MsalRedirectComponent` or alternatively calling `handleRedirectObservable` manually.
```
tnorling, line 26: This section should also be updated if you make any changes suggested above.

lib/msal-angular/docs/v2-docs/errors.md
```diff
@@ -25,7 +25,7 @@ In `@azure/msal-angular` there are 2 common scenarios when this can happen:
 
 We recommend that you bootstrap the the **`MsalRedirectComponent`** if you want to use redirects as this will handle all redirects for you. See the docs on redirects [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev/lib/msal-angular/docs/v2-docs/redirects.md) for how to use the `MsalRedirectComponent` or alternatively calling `handleRedirectObservable` manually. 
 
-We also recommend that any interaction should be done after subscribing to the `inProgress$` subject and filtering for none. Please see our [events doc](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-angular/docs/v2-docs/events.md#the-inprogress-observable) for more details. 
+We also recommend that any interaction should be done after subscribing to the `inProgress$` observable and filtering for `InteractionStatus.None`. Please see our [events doc](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-angular/docs/v2-docs/events.md#the-inprogress-observable) for more details.
```
tnorling, line 28: Let's be more explicit that attempting interaction while another is in progress is not supported and checking for `InteractionStatus.None` is how you ensure you don't do this on accident.

lib/msal-angular/docs/v2-docs/initialization.md
```diff
@@ -33,6 +33,12 @@ export class AppModule {}
 
 You can add authentication to secure specific routes in your application by just adding `canActivate: [MsalGuard]` to your route definition. It can be added at the parent or child routes. When a user visits these routes, the library will prompt the user to authenticate.
 
+**Note:** While the `MsalGuard` is done with best effort, it is a convenience feature intended to improve the user experience and, as such, should not be relied upon for security. Attackers can potentially get around client-side guards, and you should ensure that the server does not return any data the user should not access.
+
+You may also need a route guard that addresses specific needs. We encourage you to write your own guard if `MsalGuard` does not meet all those needs.
+
+SEe this example of a route defined with the `MsalGuard`:
```
tnorling, line 40:
```suggestion
See this example of a route defined with the `MsalGuard`:
```

lib/msal-angular/src/msal.broadcast.service.ts
tnorling: When is the default value used? In react we always start with `InteractionStatus.Startup` knowing that we will always call `handleRedirectPromise` at some point which will set it to `None` when complete. You likely don't want it starting at `None` as that can provide a brief window when that would be inaccurate (i.e. handleRedirectPromise needs to resolve an in progress login)

lib/msal-angular/docs/v2-docs/initialization.md
jo-arroyo: Changed. Thanks!

lib/msal-angular/docs/v2-docs/errors.md
jo-arroyo: Makes sense. Updated.

lib/msal-angular/docs/v2-docs/errors.md
jo-arroyo: Done

lib/msal-angular/docs/FAQ.md
jo-arroyo: That makes sense. I'll make it more clear. Thanks

lib/msal-angular/src/msal.broadcast.service.ts
jasonnutter: Startup seems like a good first value.

jo-arroyo: The default value hasn't really been used, and only with changing it to a `BehaviorSubject` was an initial value available. While  `handleRedirectObservable` should always be called and set it to `None`, I agree that setting it to `Startup` would be more accurate.

tnorling (approved):

jasonnutter (approved):

PR #3281, opened 2021-03-22
fix(Msal-Angular): Dynamic interceptor authority

### LINKED ISSUE #3223: Wrong authority for access tokens with multiple tenants
## Library

- [X] `@azure/msal-browser@2.12.1`
- [X] `@azure/msal-angular@2.0.0-beta.1`

## Framework

- [X] Angular 10
- [ ] React
- [ ] Other

## Description

If you set the authority to `common` or `organizations`, the `this.msalService.instance.setActiveAccount(selectedAccount);`  does nothing for getting access tokens. Because the authority on the application is set to `organizations` in the application configuration. Even though you have set an **ActiveAccount** the resulting access token is from the primary tenant and not the tenent (account) I've selected

### Context:

User from **Tenant A** is registered as a guest in **Tenant B**. This user has the correct application assigned (in both tenants).
We got the user to login into the app, by executing the following code (tenant id is provided). So by changing the authority for this account we are able to trigger a second login to the new tenant.

```js
      this.msalService.loginRedirect({
        authority: `https://login.microsoftonline.com/${tenantId}`,
        scopes: ['openid', 'profile'],
      });
```

A call to `this.msalService.instance.getAllAccounts();` results in the following data, some guids changed because I don't know if they should be visable.

```json
 [
  {
    "homeAccountId": "a72319a6-89b8-4a81-a746-169abc03c4ba.94193037-eaf0-4afe-a6d1-5fa9790a0f90",
    "environment": "login.windows.net",
    "tenantId": "94193037-eaf0-4afe-a6d1-5fa9790a0f90",
    "username": "stephan@dont-send-me-spam.nl",
    "localAccountId": "a72319a6-89b8-4a81-a746-169abc03c4ba",
    "name": "Stephan van Rooij (Smartersoft B.V.)",
    "idTokenClaims": {
      "aud": "f46eccda-b21f-47d3-b6e2-88b62f55d23e",
      "iss": "https://login.microsoftonline.com/3774557c-67de-438d-b381-ad207388fec9/v2.0",
      "iat": 1615911126,
      "nbf": 1615911126,
      "exp": 1615915026,
      "aio": "AWQAm/8TAAAAmsg9GBNnH1kK+UkGS3gKixMp2itIfDJEgUh+7id/Q8U3u1iKm4Z+Z2kivNk1BHnApytQ4opurHNoN2orRHyy3lRyBoUAK3UZEHxYtht7axej/aVnb8NZ6kCGtA1dYrGl",
      "name": "Stephan van Rooij (Smartersoft B.V.)",
      ...
    }
  },
  {
    "homeAccountId": "a72319a6-89b8-4a81-a746-169abc03c4ba.94193037-eaf0-4afe-a6d1-5fa9790a0f90",
    "environment": "login.windows.net",
    "tenantId": "81c3be60-f652-4417-bf28-6b352312877b",
    "username": "stephan@dont-send-me-spam.nl",
    "localAccountId": "1d8fe87c-4688-4f10-9178-31fc6a1f8edd",
    "name": "Stephan van Rooij",
    "idTokenClaims": {
      "aud": "f46eccda-b21f-47d3-b6e2-88b62f55d23e",
      "iss": "https://login.microsoftonline.com/81c3be60-f652-4417-bf28-6b352312877b/v2.0",
      "iat": 1615907881,
      "nbf": 1615907881,
      "exp": 1615911781,
      ...
    }
  }
]
```

## Error Message

No error message, just the wrong (not expected) token

## MSAL Configuration

```js
// Basic configuration from the docs https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-angular/docs/v2-docs/initialization.md
// just set the authority to organizations
```

## Reproduction steps

```js
// Have the primary user login by the guard or whatever method you have in place
// Login the same user, but now for the other tenant (where he is a guest)
      this.msalService.loginRedirect({
        authority: `https://login.microsoftonline.com/${tenantB}`,
        scopes: ['openid', 'profile'],
      });
// After the user is logged in, switch the default account:
const wantedAccount = allAccounts.find(a => a.tenantId === tenantIdOfTenantB);
this.msalService.instance.setActiveAccount(wantedAccount);
```

## Expected behavior

I would expect the **MsalInterceptor** to get a new token for this user by calling `https://login.microsoftonline.com/tenant-id-of-active-account/oauth2/v2.0/token` instead it calls `https://login.microsoftonline.com/organizations/oauth2/v2.0/token`. Which results in a token for Tenant A (primary tenant for user), and not in a token for the explicitly set account.

If I manually request a token with the following code I'm getting the correct token:

```js
constructor(@Inject(MSAL_INTERCEPTOR_CONFIG) private msalInterceptorConfig: MsalInterceptorConfiguration, private msalService: MsalService) { }

this.msalService.instance.acquireTokenSilent({
      ...this.msalInterceptorConfig.authRequest,
      scopes: this.msalInterceptorConfig.protectedResourceMap.get('/api/*'),
      account: wantedAccount,
      authority: `https://login.microsoftonline.com/${wantedAccount.tenantId}`
    }).then((result) => {
      this.token = result.accessToken;
    })
```

To fix this for all multi-tenant applications I would suggest to do a string replace of `organizations` by the actual tenant id of the account, when requesting an access token. This makes it really easy to switch accounts (because you just have to set the ActiveAccount). And it doesn't do any harm it you replace `organizations` by the actual tenant id. It might even make it some millisecond faster, because the token endpoint doesn't have to resolve the tenant.

## Identity Provider

- [X] Azure AD
- [ ] Azure B2C Basic Policy
- [ ] Azure B2C Custom Policy
- [ ] ADFS
- [ ] Other

## Browsers/Environment

- [ ] Chrome
- [ ] Firefox
- [X] Edge (probably doesn't matter)
- [ ] Safari
- [ ] IE

## Regression

Version:

## Security

- [X] Is this issue security related?

Access tokens mean security, but in this case you're not getting access you shouldn't.

## Source

- [ ] Internal (Microsoft)
- [X] Customer request

svrooij:
Helps supporting true multi tenant applications, by allowing
overriding the authority at runtime.

Fixed #3223


jasonnutter:
@svrooij Thanks for the PR, we'll take a look and let you know if we have feedback.

svrooij:
Thanks @jasonnutter I was already discussing this with @tnorling in #3223

But since I missed the tests part and the coverage went down by 7% I'm going to add some tests later this evening (UTC+1 here).

svrooij:
@jasonnutter according to the [docs](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/contributing.md#creating-a-pr) I also have to do some **Beachball** stuff, is that something I should do or is that something you as a code owner should do? Since that probably will end up in the release notes somewhere.

jasonnutter:
> @jasonnutter according to the [docs](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/contributing.md#creating-a-pr) I also have to do some **Beachball** stuff, is that something I should do or is that something you as a code owner should do? Since that probably will end up in the release notes somewhere.

This is something you should do. In your branch, run `npm install` in the root of repo, and then run `npm run beachball:change` and follow the prompts, and then push the updated branch.

svrooij:
@jasonnutter already did that. Was a small change ☝️

tnorling:
@svrooij I think I would rather handle this in `msal-browser`, if possible, as this will likely affect more users than just Angular users. But I'm not sure what that would look like, we may need to add an API to `PublicClientApplication` to make changes to the base config after instantiation.

svrooij:
I partially agree with you @tnorling, in that this will most likely affect MSAL-browser users as well.

Since this is my first contribution I did not want to go ahead and change the public api of the most important part of this entire library. Also changing something big like that would require some discussions with the code owners I guess?

I asked myself, does this change break stuff for existing users and in what way can we make it that it will fit multiple use cases. If this way is liked by many users, we could also implement something similar for MSAL-browser.

lib/msal-angular/docs/multi-tenant.md
```diff
@@ -45,3 +45,37 @@ export class AppComponent implements OnInit {
...
+  const protectedResourceMap = new Map<string, Array<string>>();
+  protectedResourceMap.set("https://graph.microsoft.com/v1.0/me", ["user.read"]);
+  
+  return {
+    interactionType: InteractionType.Popup,
+    protectedResourceMap,
+    authRequest: {
+      authority: 'https://login.microsoftonline.com/organizations',
+    },
+    dynamicAuthority: (account: AccountInfo) => `https://login.microsoftonline.com/${account.tenantId ?? 'organizations'}`
```
jasonnutter, line 64: Instead of just changing the authority, we can maybe consider being able to change the entire `authRequest`? This would mean changing `authRequest` to support an object literal or a function (that takes in the account, like your PR shows) that returns an object literal. That would help address other scenarios for other properties on the `authRequest`.

jasonnutter:
> @svrooij I think I would rather handle this in `msal-browser`, if possible, as this will likely affect more users than just Angular users. But I'm not sure what that would look like, we may need to add an API to `PublicClientApplication` to make changes to the base config after instantiation.

For the purposes of MSAL Angular, I think we can limit the scope of this change to just changing the properties passed in teh requests made in teh interceptor and guard (the `authRequest` property), and not the base configuration.

lib/msal-angular/docs/multi-tenant.md
jasonnutter: Also, if we can make a similar change for MSAL Guard, I think that would beneficial.

svrooij: I did consider this, but came to the conclusion that you cannot change other properties of the request. The scope is already passed in and changing the redirect uri at runtime is weird since it requires registration at the azure portal.

Also you cannot change any properties on the guard, because either you have an account and the guard just says "ok go ahead" or you don't have an account and then there is no account to change it dynamically.

svrooij:
@jasonnutter what properties are considered changeable? I'm not very known to the library just yet.

lib/msal-angular/docs/multi-tenant.md
jasonnutter: Any of the properties for the declared types (`Omit<PopupRequest, "scopes"> | Omit<RedirectRequest, "scopes"> | Omit<SilentRequest, "scopes">;`) are changable.

jasonnutter: (We may consider adding `account` to teh omit list, cc: @jo-arroyo)

jasonnutter: For example:

```js
let authRequest;
if (typeof this.msalInterceptorConfig.authRequest === "function") {
  authRequest = this.msalInterceptorConfig.authRequest(account);
} else {
  authRequest = this.msalInterceptorConfig.authRequest;
}
```

svrooij: I know those properties can be changed because of the type definition, but I was wondering what properties would be reasonable to change.

And omitting the account might be a good idea 😉

jasonnutter: > I know those properties can be changed because of the type definition, but I was wondering what properties would be reasonable to change.

We designed it to allow any of the properties (other than `scopes`) to be set via `authRequest`, and I don't see any reason not to allow of those same properties to be set via a dynamic `authRequest`.

jasonnutter: In fact, a few of them (e.g. correlation id) may only be useful if they can be set dynamically.

jasonnutter: > And omitting the account might be a good idea

I would actually wait on that. Right now we support the active account or the first account from `getAllAccounts` array, but we may apps to have different logic.

svrooij: So the logic can be something like `dynamicAuthRequest: (originalAuthRequest: ..., httpReq: HttpEvent<any>): AuthRequest`

Then you can support overriding all properties you want. And you're right it does make sense to set the CorrelationId dynamically.

jasonnutter: I would personally prefer the `authRequest` property support two types (typed object or function that returns a typed object), instead of having two separate fields, as I think that will be simpler for us and applications.

@jo-arroyo @tnorling Thoughts?

jasonnutter: And as mentioned, I think we should have this change for both the guard and interceptor, given they have the same considerations around building the auth request.

svrooij: @jasonnutter it also makes sense to still have defaults defined. By having two types on the authRequest configuration you cannot use both at the same time. And have to use dynamic for everything.

jasonnutter: > By having two types on the authRequest configuration you cannot use both at the same time.

True, but if you need to set static and dynamic fields, both can be done in the dynamic version, e.g.

```js
authRequest: (httpReq, msalInstance) => {
  return {
     correlation: dyanmicCorrelationId,
     forceRefesh: true
  };
};
```

jasonnutter: > it also makes sense to still have defaults defined

Can you please clarify what you mean? The scope and account are the only properties that we set, we don't provide any additional defaults (aside from the defaults set internally by MSAL for acquireTokenSilent, etc).

tnorling: I like the idea of having one `authRequest` property that can either be static or a function. I can forsee lots of confusion about which is used, which should be used, what the difference is, etc. if we keep them separate.

svrooij: I changed it to be one property that is either a function or a static. And added the same functionality for the MsalGuard

lib/msal-angular/src/msal.interceptor.config.ts
```diff
@@ -3,10 +3,13 @@
...
+import { HttpRequest } from "@angular/common/http";
 import { PopupRequest, RedirectRequest, InteractionType, SilentRequest } from "@azure/msal-browser";
 
+export declare type MsalInterceptorAuthRequest = Omit<PopupRequest, "scopes"> | Omit<RedirectRequest, "scopes"> | Omit<SilentRequest, "scopes"> ;
+
 export type MsalInterceptorConfiguration = {
     interactionType: InteractionType.Popup | InteractionType.Redirect;
     protectedResourceMap: Map<string, Array<string>>;
-    authRequest?: Omit<PopupRequest, "scopes"> | Omit<RedirectRequest, "scopes"> | Omit<SilentRequest, "scopes">;
+    authRequest?: MsalInterceptorAuthRequest | ((originalAuthRequest: MsalInterceptorAuthRequest, req: HttpRequest<unknown>) => MsalInterceptorAuthRequest);
```
jasonnutter, line 14:
```suggestion
    authRequest?: MsalInterceptorAuthRequest | ((msalService: MsalService, req: HttpRequest<unknown>) => MsalInterceptorAuthRequest);
```

lib/msal-angular/docs/multi-tenant.md
```diff
@@ -45,3 +45,42 @@ export class AppComponent implements OnInit {
...
+  const protectedResourceMap = new Map<string, Array<string>>();
+  protectedResourceMap.set("https://graph.microsoft.com/v1.0/me", ["user.read"]);
+  
+  return {
+    interactionType: InteractionType.Popup,
+    protectedResourceMap,
+    authRequest: {
+      authority: 'https://login.microsoftonline.com/organizations',
+    },
+    dynamicAuthRequest: (authRequest, httpReq) => {
```
jasonnutter, line 64:
```suggestion
    authRequest: (msalService, httpReq) => {
```

lib/msal-angular/docs/multi-tenant.md
```diff
@@ -45,3 +45,42 @@ export class AppComponent implements OnInit {
...
+      return {
+        ...authRequest,
+        authority: `https://login.microsoftonline.com/${authRequest.account?.tenantId ?? 'organizations'}`
```
jasonnutter, line 67: Instead of pull this off of the authRequest, you should use the get account APIs on the MSAL instance (either getActiveAccount, getAllAccounts[0], etc) depending on if your app sets the active account.

lib/msal-angular/src/msal.interceptor.config.ts
svrooij: The MSAL service doesn't provide exposed configuration. And the select account functionality is already build into the interceptor. 
This way it seems more like a way to override then to replace (what you are proposing).

lib/msal-angular/src/msal.guard.config.ts
```diff
@@ -4,9 +4,12 @@
...
 
 import { PopupRequest, RedirectRequest,InteractionType } from "@azure/msal-browser";
+import { MsalService } from "./msal.service";
+
+export declare type MsalGuardAuthRequest = Partial<PopupRequest> | Partial<Omit<RedirectRequest, "redirectStartPage">>;
 
 export type MsalGuardConfiguration = {
     interactionType: InteractionType.Popup | InteractionType.Redirect;
-    authRequest?: Partial<PopupRequest> | Partial<Omit<RedirectRequest, "redirectStartPage">>;
+    authRequest?: MsalGuardAuthRequest | ((config: MsalGuardConfiguration, authService: MsalService) => MsalGuardAuthRequest);
```
jasonnutter, line 13: We would like to provide the authSerivce and `state`

```suggestion
    authRequest?: MsalGuardAuthRequest | ((msalService: MsalService, state: RouterStateSnapshot) => MsalGuardAuthRequest);
```

lib/msal-angular/src/msal.interceptor.config.ts
svrooij: Currently the code in the interceptor is configured like this:
```js
const authRequest = typeof this.msalInterceptorConfig.authRequest === "function"
            ? this.msalInterceptorConfig.authRequest({ account: account }, req)
            : { ...this.msalInterceptorConfig.authRequest, account };
```

And just above that part is the logic to get to correct account.

That means the authRequest can be overriden like:
```js
authRequest: (authRequest, httpReq) => { 
  return { ...authRequest, authority: 'something here' };
}
```

And you don't have to copy the select the correct account logic. How about I add the MsalService as well, that way you can do both. So get the account in a none default way and just use the default way and override only the stuff you want overridden?

lib/msal-angular/src/msal.guard.config.ts
svrooij: This makes perfect sense, applied

svrooij:
And also fixed some coverage 🎉 

image

svrooij:
@jasonnutter did you see my changes?

jasonnutter:
@svrooij Thanks, been busy but will look now.

svrooij:
Do you guys normally merge PRs or are they squashed? Since here are so many commits, maybe nicer if they are just merged as one commit.

lib/msal-angular/src/msal.interceptor.config.ts
```diff
@@ -3,10 +3,14 @@
...
 import { PopupRequest, RedirectRequest, InteractionType, SilentRequest } from "@azure/msal-browser";
+import { MsalService } from "./msal.service";
+
+export declare type MsalInterceptorAuthRequest = Omit<PopupRequest, "scopes"> | Omit<RedirectRequest, "scopes"> | Omit<SilentRequest, "scopes"> ;
 
 export type MsalInterceptorConfiguration = {
     interactionType: InteractionType.Popup | InteractionType.Redirect;
     protectedResourceMap: Map<string, Array<string>>;
-    authRequest?: Omit<PopupRequest, "scopes"> | Omit<RedirectRequest, "scopes"> | Omit<SilentRequest, "scopes">;
+    authRequest?: MsalInterceptorAuthRequest | ((originalAuthRequest: MsalInterceptorAuthRequest, msalService: MsalService, req: HttpRequest<unknown>) => MsalInterceptorAuthRequest);
```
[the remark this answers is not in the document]
jasonnutter: I guess I can see why it would be useful, but I would prefer it is put at the end:

```suggestion
    authRequest?: MsalInterceptorAuthRequest | ((msalService: MsalService, req: HttpRequest<unknown>, originalAuthRequest: MsalInterceptorAuthRequest) => MsalInterceptorAuthRequest);
```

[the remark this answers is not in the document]
svrooij: My original comment got lost somewhere, but this makes it possible to use the account selection logic from the interceptor and just override the properties you want.

If we remove the original auth request the developer is required to do the account selection, and this becomes more an provide your own implementation instead of change some properties to your liking.

By providing both the MSAL service and the computed authentication request the developer has a choice to either set own implementation or just override for instance the authority based on the selected account

jasonnutter:
> Do you guys normally merge PRs or are they squashed? Since here are so many commits, maybe nicer if they are just merged as one commit.

Normally just merge, but can squash this one since it has so many commits. Not a big deal for us, frankly.

lib/msal-angular/src/msal.interceptor.config.ts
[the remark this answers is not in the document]
svrooij: I’ll move it to the back and update the code calling this, tomorrow from behind my Pc cannot do that on my tablet while being on the sofa.

svrooij:
@jasonnutter I've tried to keep up with the changes on the dev branch, but after the last 2 merges, it's again out-of-date. Does that matter?

jasonnutter:
> @jasonnutter I've tried to keep up with the changes on the dev branch, but after the last 2 merges, it's again out-of-date. Does that matter? Invited you to my fork, so you should be able to fix merge stuff also.

Don't worry about that, we can handle keeping that up to date.

svrooij:
@jasonnutter Can I do something to move this PR forward? Or did I miss one of your comments? If you guys are busy that is also fine, I just have no clue how long a regular pr would take for this library.

lib/msal-angular/src/msal.guard.spec.ts
```diff
@@ -13,21 +13,26 @@ let routeStateMock: any = { snapshot: {}, url: '/' };
...
+let browserSystemOptions: BrowserSystemOptions;
 
 function MSALInstanceFactory(): IPublicClientApplication {
   return new PublicClientApplication({
     auth: {
       clientId: '6226576d-37e9-49eb-b201-ec1eeb0029b6',
       redirectUri: 'http://localhost:4200'
-    }
+    },
+    //system: browserSystemOptions
```
jasonnutter, line 25: Please remove

jasonnutter, line 25:
```suggestion
    }
```

lib/msal-angular/src/msal.guard.config.ts
```diff
@@ -3,10 +3,16 @@
...
+import { MsalService } from "./msal.service";
+
+export declare type MsalGuardAuthRequest = Partial<PopupRequest> | Partial<Omit<RedirectRequest, "redirectStartPage">>;
 
 export type MsalGuardConfiguration = {
     interactionType: InteractionType.Popup | InteractionType.Redirect;
-    authRequest?: Partial<PopupRequest> | Partial<Omit<RedirectRequest, "redirectStartPage">>;
+    authRequest?: MsalGuardAuthRequest | ((authService: MsalService, state: RouterStateSnapshot) => MsalGuardAuthRequest);
     loginFailedRoute?: string;
+    canActivate?: (msalService: MsalService, state?: RouterStateSnapshot) => Observable<boolean> | boolean;
```
jasonnutter, line 17: I'm not sure I understand why this is necessary? It seems like this would allow someone to implement their own guard logic, in which case, we would rather an app implement a custom guard, instead of using the MSAL Guard (which is intended to be generic, and will not work for all use cases).

jasonnutter: Furthermore, this feels like scope creep for this PR. I would prefer it is scoped to just the dynamic auth request, and have further enhacements be discussed/implemented separately.

lib/msal-angular/docs/multi-tenant.md
```diff
@@ -45,3 +45,39 @@ export class AppComponent implements OnInit {
     });
   }
 ```
+
+## Multi-tenant access tokens
```
jasonnutter, line 49: I think it would be good to include two sub sections here, one for the guard, and one for the interceptor.

jasonnutter:
> @jasonnutter Can I do something to move this PR forward? Or did I miss one of your comments? If you guys are busy that is also fine, I just have no clue how long a regular pr would take for this library.

Sorry, been busy. Left a few comments.

svrooij:
@jasonnutter I removed the feature creep and created a new issue for it #3406 I also changed the documentation and removed some comments in the tests that got left over from trying to debug some issue I was having.

jasonnutter:
> @jasonnutter I removed the feature creep and created a new issue for it #3406 I also changed the documentation and removed some comments in the tests that got left over from trying to debug some issue I was having.

Great, thanks! Will take a look.

jasonnutter (approved):
Looks good, thanks for addressing feedback! @jo-arroyo Can you please take a look when you get a chance?

lib/msal-angular/docs/multi-tenant.md
```diff
@@ -45,3 +45,46 @@ export class AppComponent implements OnInit {
     });
   }
 ```
+
+## Dynamic auth request
+
+By default the MsalGuard and the MsalInterceptor use the static properties you set in the config. Both can be configured with a method for the `authRequest`. This allows to dynamically change the parameters used for authentication.
```
jo-arroyo, line 51:
```suggestion
By default the MsalGuard and the MsalInterceptor use the static properties set in the config. Both can also be configured with a method for the `authRequest`, allowing the parameters used for authentication to be changed dynamically.
```

lib/msal-angular/docs/multi-tenant.md
```diff
@@ -45,3 +45,46 @@ export class AppComponent implements OnInit {
...
+
+### MsalInterceptor - dynamic auth request (multi tenant tokens)
+
+If you're using `organizations` or `common` all the tokens will be requested for the users' home tenant, this might not result in the desired effect. If you have this application where you're invited as a guest the tokens will be from the wrong authority.
```
jo-arroyo, line 55:
```suggestion
If `organizations` or `common` is used as the tenant, all tokens will be requested for the users' home tenant. However, this may not be the desired outcome. If a user is invited as a guest, the tokens may be from the wrong authority.
```

lib/msal-angular/docs/multi-tenant.md
```diff
@@ -45,3 +45,46 @@ export class AppComponent implements OnInit {
...
+
+Setting the `authRequest` in the **MsalInterceptorConfig** to a method allows you to dynamically change the auth request, and for instance set the authority based on the home tenant of the account (when using guest users).
```
jo-arroyo, line 57:
```suggestion
Setting the `authRequest` in the **MsalInterceptorConfig** to a method allows you to dynamically change the auth request. For instance, you may set the authority based on the home tenant of the account when using guest users.
```

lib/msal-angular/docs/multi-tenant.md
```diff
@@ -45,3 +45,46 @@ export class AppComponent implements OnInit {
...
+You can change all properties but be sure to always extend the `originalAuthRequest` like you see below.
```
jo-arroyo, line 58:
```suggestion
Properties on `authRequest` may be changed, but should always extend the `originalAuthRequest` like below:
```

lib/msal-angular/docs/v2-docs/configuration.md
```diff
@@ -270,3 +270,23 @@ fetch('/assets/configuration.json')
   }
 }
 ```
+
+## MsalGuard - dynamic auth request
+
+The **MsalGuard** also allows you to dynamically change the **authRequest** at runtime. This allow you to pick a different authority for some route, or to dynamically add scopes based on the **RouterStateSnapshot**.
```
jo-arroyo, line 276:
```suggestion
The **MsalGuard** also allows you to dynamically change the **authRequest** at runtime. This allow you to pick a different authority for a route, or to dynamically add scopes based on the **RouterStateSnapshot**.
```

lib/msal-angular/src/msal.interceptor.spec.ts
```diff
@@ -349,7 +353,60 @@ describe('MsalInterceptor', () => {
...
 
+
+  it("keeps original authority, https://login.microsoftonline.com/common", done => {
+    testInterceptorConfig.authRequest = {
+      authority: 'https://login.microsoftonline.com/common'
+    };
+    initializeMsal();
+    spyOn(PublicClientApplication.prototype, "getAllAccounts").and.returnValue([sampleAccountInfo]);
+    spyOn(PublicClientApplication.prototype, "acquireTokenSilent").and.callFake((silentRequest: SilentRequest) => new Promise((resolve) => {
+      // expect(silentRequest.authority).toBeUndefined();
```
jo-arroyo, line 364: Is this still needed here?

lib/msal-angular/src/msal.redirect.component.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+  TestBed.resetTestingModule();
+
+  TestBed.configureTestingModule({
+    declarations: [MsalRedirectComponent],
+    imports: [
+
+      MsalModule.forRoot(MSALInstanceFactory(), null, null)
+    ],
+    providers: [
+
```
jo-arroyo, line 34:
```suggestion
```

lib/msal-angular/src/msal.redirect.component.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+  });
+}
+
+function initializeMsal() {
+  TestBed.resetTestingModule();
+
+  TestBed.configureTestingModule({
+    declarations: [MsalRedirectComponent],
+    imports: [
+
```
jo-arroyo, line 30:
```suggestion
```

lib/msal-angular/src/msal.redirect.component.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+let authService: MsalService;
+let broadcastService: MsalBroadcastService;
+
+function MSALInstanceFactory(): IPublicClientApplication {
+  return new PublicClientApplication({
+    auth: {
+      clientId: '6226576d-37e9-49eb-b201-ec1eeb0029b6',
+      redirectUri: 'http://localhost:4200'
+    },
+    //system: browserSystemOptions
```
jo-arroyo, line 20: Is this still needed here?

jo-arroyo (commented):
Thanks for your work. I reviewed and have a few comments.

svrooij:
> @svrooij Thanks for your work. I reviewed and have a few comments.

@jo-arroyo Thank you for the feedback.

jo-arroyo (approved):

PR #3300, opened 2021-03-25
Make msal-browser and msal-common package preservedModules and mark sideEffects: false to enable tree-shaking for consumer

thoo1:
…this should makes the tree shaking work


thoo1:
I couldn't get test to pass locally but the tree shaking is working
for example, searching for OnBehalfOfClient in the built lib/msal-browser/dist/index.es.js should yield no results if treeshaking works
image

thoo1:
If folks would like to help improve this PR and get tests working, that's very welcome
I followed some of the instructions from this issue to make this PR: https://github.com/rollup/rollup/issues/2807#issuecomment-484798155

jasonnutter:
Quick question: how does this impact IE support?

badsyntax:
I also don't understand why the build process generates a bundle for the "module" field in `package.json`, it just doesn't make sense to me, but perhaps i'm missing something.

Bundlers rely on import's/export's to tree shake effectively (same as rollup), so I expect the "module" field to link to an index file that then imports the relevant modules, which in-turn imports other modules. The *consuming* build then will bundle and tree-shake.

I expect this library to publish ES modules (as in, seperate files with imports/exports) to allow my consuming build to tree-shake.

thoo1:
@badsyntax Yea this is where I also need to learn more about, I'm very new to how js libraries should export modules
@jasonnutter I'm not sure but I the babel plugin that is added should respect the .babelrc config file https://www.npmjs.com/package/@rollup/plugin-babel
Let me read further to see what the current babelrc file is configured to do, sorry, very new at this :)

thoo1:
after some reading, I found that it looks like for tree-shaking to work for a webpack project, the package shouldn't be bundled (in one file)
https://github.com/webpack/webpack/issues/9337#issuecomment-524966201
So we need to use preserveModules, I think this is what you were saying @badsyntax ?

thoo1:
this commit so far is not correct in enabling tree-shaking for the consumer, it is actually treeshaking the msal-common dependencies inside msal-browser, which wont do anything since the consumer will pull msal-common and msal-browser seperately

thoo1:
I think I may have gotten it to work, here is folder structure of the build in msal-browser and msal-common now that I've `preservedModules: true` in the rollup configs
msal-browser
image

msal-common
image

My project that uses msal-browser has decreased its bundle size about 150kB from this tree-shaking

TODO: uncomment out the other builds (cjs, minified build, etc)

badsyntax:
This is awesome work @thoo1! Over the weekend I will try your branch in my project to check the size reductions and verify the approach.

thoo1:
folder structure after latest changes (to enable cjs and umd builds too)
msal-browser:
image

msal-common:
image

jasonnutter:
> I also don't understand why the build process generates a bundle for the "module" field in `package.json`, it just doesn't make sense to me, but perhaps i'm missing something.
> 
> Bundlers rely on import's/export's to tree shake effectively (same as rollup), so I expect the "module" field to link to an index file that then imports the relevant modules, which in-turn imports other modules. The _consuming_ build then will bundle and tree-shake.
> 
> I expect this library to publish ES modules (as in, seperate files with imports/exports) to allow my consuming build to tree-shake.

We produce single-file builds for our CDN hosted version. cc: @pkanher617

ZuBB:
Also interested in this feature. in our case msal takes 2/5 of bundle size which is a lot.

@jasonnutter could you please tell us what do we miss here to have this merged?

jasonnutter:
Apologies, been busy and haven't had a chance to follow up here.

This brings up another question: could this be solved if we moved msal-browser and msal-common to tsdx (which we use for msal-react and msal-node)? We were already planning to do that in the next major version of MSAL Browser, but we could do that work now if it will address this issue.

cc @pkanher617

jasonnutter:
In the meantime, I'll pull down this branch and see how it looks. Thanks for taking the time @thoo1!

Issue #3192, opened 2021-03-11
MsalInterceptor not acquiring tokens when endpoint contains query parameters


derisen:
## Library

- [ ] `msal@1.x.x` or `@azure/msal@1.x.x`
- [ ] `@azure/msal-browser@2.x.x`
- [ ] `@azure/msal-node@1.x.x`
- [ ] `@azure/msal-react@1.x.x`
- [ ] `@azure/msal-angular@0.x.x`
- [ ] `@azure/msal-angular@1.x.x`
- [x] `@azure/msal-angular@2.0.0-beta.0`
- [ ] `@azure/msal-angularjs@1.x.x`

## Framework

- [x] Angular 11
- [ ] React
- [ ] Other

## Description
Some web API endpoints contain query parameters. MsalInterceptor does not seem to acquire tokens for such endpoints (e.g. [Azure Resource Manager API](https://docs.microsoft.com/en-us/rest/api/resources/tenants/list))

## Error Message

API response:

```
zone.js:3324 GET https://management.azure.com/tenants?api-version=2020-01-01 401
core.js:5980 ERROR 
HttpErrorResponse {headers: HttpHeaders, status: 401, statusText: "OK", url: "https://management.azure.com/tenants?api-version=2020-01-01", ok: false, …}
error:
error:
code: "AuthenticationFailed"
message: "Authentication failed. The 'Authorization' header is missing."
__proto__: Object
__proto__: Object
headers: HttpHeaders {normalizedNames: Map(0), lazyUpdate: null, lazyInit: ƒ}
message: "Http failure response for https://management.azure.com/tenants?api-version=2020-01-01: 401 OK"
name: "HttpErrorResponse"
ok: false
status: 401
statusText: "OK"
url: "https://management.azure.com/tenants?api-version=2020-01-01"
__proto__: HttpResponseBase
```

msal-angular logs:

```text
[Thu, 11 Mar 2021 02:58:43 GMT] :  : @azure/msal-browser@2.12.0 : Verbose - Emitting event to callback 80777888-4672-4c37-a974-70e698762e80: msal:handleRedirectEnd
auth-config.ts:30 [Thu, 11 Mar 2021 02:58:43 GMT] :  : @azure/msal-angular@2.0.0-beta.0 : Verbose - BroadcastService - msal:handleRedirectEnd results in setting inProgress to none
[...]

PR #3307, opened 2021-03-25
msal-angular interceptor endpoint fix

jo-arroyo:
This PR:
- Remove "in progress" from msal-angular README
- Changes `matchPattern` in `StringUtils` in msal-common to allow queries in string (addresses issue #3192)
- Changes MsalInterceptor `protectedResourceMap` type to allow for `null` (also issue #3192)
- Changes MsalInterceptor to check for relative paths if applicable
- Updates MsalInterceptor unit tests
- Updates msal-angular-v2 samples to use the latest versions of `@azure/msal-angular` and `@azure/msal-browser`


lib/msal-angular/src/msal.interceptor.config.ts
```diff
@@ -7,6 +7,6 @@ import { PopupRequest, RedirectRequest, InteractionType, SilentRequest } from "@
 
 export type MsalInterceptorConfiguration = {
     interactionType: InteractionType.Popup | InteractionType.Redirect;
-    protectedResourceMap: Map<string, Array<string>>;
+    protectedResourceMap: Map<string, Array<string>> | null;
```
jasonnutter, line 10: I think we want the `null` inside the map, like

```suggestion
    protectedResourceMap: Map<string, Array<string> | null>;
```

jasonnutter: Also, we could make the protectedResourceMap optional?

tnorling (approved):

jasonnutter:
Looks good, thanks for the tests!

PR #3315, opened 2021-03-26
Ignore account hints when prompt=select_account

tnorling:
AAD throws an error when you pass both `prompt=select_account` and a `login_hint` or `sid` as these are conflicting. This becomes problematic when someone wants to use `prompt=select_account` to switch users and has previously used the `setActiveAccount` API in browser (we quietly pass down the active account if none is provided to the interactive API).

This PR ignores account hints if `prompt=select_account` to prevent this error from happening unexpectedly.


lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -248,21 +248,26 @@ export class AuthorizationCodeClient extends BaseClient {
...
 
         // Add sid or loginHint with preference for sid -> loginHint -> username of AccountInfo object
-        if (request.sid) {
-            parameterBuilder.addSid(request.sid);
-        } else if (request.account) {
-            const accountSid = this.extractAccountSid(request.account);
-            // If account and loginHint are provided, we will check account first for sid before adding loginHint
-            if (accountSid) {
-                parameterBuilder.addSid(accountSid);
+        if (request.prompt !== PromptValue.SELECT_ACCOUNT) {
```
jasonnutter, line 251: Would we want to also ignore account hints for login?

jasonnutter (approved):

lib/msal-common/src/client/AuthorizationCodeClient.ts
tnorling: I went back and forth on that. In the `prompt=login` scenario `login_hint` could be used to pre-fill the email field so maybe just ignore `sid`?

jasonnutter: I think I've seen scenarios where an app would want to trigger the login prompt with the current user, so as long as we provide a way to set prompt=login with and without an account hint, we should be good.

jasonnutter: And yeah, sid can be ignored for interactive requests.

sameerag:
The server mandates that sid/login_hint should be used only for `prompt=none` use cases. Did we remove that check?

sameerag:
The server mandates that sid/login_hint should be used only for `prompt=none` use cases. Did we remove that check?

tnorling:
> The server mandates that sid/login_hint should be used only for `prompt=none` use cases. Did we remove that check?

AFAIK this was never present in 2.x. This PR only sends `sid` when `prompt=none`, however, `login_hint` is valid for `prompt=login` so will only be ignored if `prompt=select_account`.

jasonnutter (approved):

sameerag (approved):
Approved. One thing: test cases to ignore sid for prompt!=none (if user provides it), please check that once.

tnorling:
> Approved. One thing: test cases to ignore sid for prompt!=none (if user provides it), please check that once.

Yes, it's there (test cases to use login_hint over sid if prompt != none covers this as login_hint has lower priority than sid)

sameerag:
> Yes, it's there (test cases to use login_hint over sid if prompt != none covers this as login_hint has lower priority than sid)

I don't think we are testing this use case: Only `sid` and ignoring it if passed with `prompt != none`. This is different from making `sid` lower priority. Not blocking this PR, but this test case is critical to have around, in case something changes tomorrow.

tnorling:
> > Yes, it's there (test cases to use login_hint over sid if prompt != none covers this as login_hint has lower priority than sid)
> 
> I don't think we are testing this use case: Only `sid` and ignoring it if passed with `prompt != none`. This is different from making `sid` lower priority. Not blocking this PR, but this test case is critical to have around, in case something changes tomorrow.

Gotcha, added!

PR #3320, opened 2021-03-26
Make resourceRequestMethod and resourceRequestUri optional parameters for Access Token PoP

hectormmg:
This PR:

- Moves `authenticationScheme` into BaseAuthRequest so it is available for all request types
- Refactors proof-of-possession APIs to take a request parameter instead of the PoP-specific `resourceRequestMethod`, `resourceRequestUri` and `shrClaims`, extracting these attributes from the request before signing the SHR.
- Updates `PopTokenGenerator` spec to test that PoP tokens are signed even when PoP-specific attributes aren't configured


lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -1148,15 +1148,17 @@ export abstract class ClientApplication {
             browserState
         );
 
-        const authenticationScheme = request.authenticationScheme || AuthenticationScheme.BEARER;
+        // Set authenticationScheme to BEARER if not explicitly set in the request
+        if (!request.authenticationScheme) {
+            request.authenticationScheme = AuthenticationScheme.BEARER;
+        }
```
jasonnutter, line 1154: Can you please add verbose log messages to both sides of this conditional?

hectormmg: Done

lib/msal-common/src/request/BaseAuthRequest.ts
```diff
@@ -3,20 +3,24 @@
...
- * - authority               - URL of the authority, the security token service (STS) from which MSAL will acquire tokens. Defaults to https://login.microsoftonline.com/common. If using the same authority for all request, authority should set on client application object and not request, to avoid resolving authority endpoints multiple times.
- * - correlationId           - Unique GUID set per request to trace a request end-to-end for telemetry purposes.
  * - resourceRequestMethod      - HTTP Request type used to request data from the resource (i.e. "GET", "POST", etc.).  Used for proof-of-possession flows.
  * - resourceRequestUri         - URI that token will be used for. Used for proof-of-possession flows.
  */
 export type BaseAuthRequest = {
     authority: string;
     correlationId: string;
     scopes: Array<string>;
+    authenticationScheme?: AuthenticationScheme;
```
tnorling, line 23: What are the scenarios where this is not defined? It's always gotta be Bearer or PoP right?

tnorling: Is this just so this isn't a breaking change in common?

change/@azure-msal-common-7b4fc307-7a4d-40c4-9076-345d5f2c0dad.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Make SHR parameters optional (#3320)",
```
tnorling, line 3: Leaving off the parentheses enables links in the changelog
```suggestion
  "comment": "Make SHR parameters optional #3320",
```

lib/msal-common/src/request/BaseAuthRequest.ts
hectormmg: There's logic that sets it to bearer by default if it isn't set explicitly in the request, so adding it to the initial object is optional. I I think making it optional is hiding implementation details from clients that don't require POP.

tnorling: The platform libs have their own request objects that choose which parameters to expose to the end user. Making this required here shouldn't affect whether or not it's optional for users but helps the type checker enforce that it's defined internally

tnorling: That said, making it required here may be breaking for types that didn't already have it so it can wait until the next common major version bump.

pkanher617 (approved):

tnorling (approved):

PR #3327, opened 2021-03-29
Fix /consumers authority caching

tnorling:
Last year we made a change to ensure `/organizations` and `/common` authority are cached with the tenantId and don't result in cache misses when the tenantId is not explicitly provided. We missed the `/consumers` authority which is now potentially leaving the cache in a state that contains both `/consumers` and a tenanted authority for the same set of scopes and throws a multiple authority error.

Also removes multiple authority error which is no longer doing what it was originally intended to do (throw when multiple cache entries existed for the same scope but different authorities). If a cache entry is not found for the provided scopes and authority it should result in a cache miss regardless of what other tokens are available in the cache.


hectormmg (approved):

jasonnutter (approved):

sameerag (approved):
Lgtm. Approving this for now. I remember there is an MSA passthru use case that needs to be tested here, I will reach out with details. We can unblock this PR but lets follow through.

PR #3343, opened 2021-03-30
Update keytar version to the latest and add caret versioning!

sameerag:
Node extensions currently are based on [keytar](https://www.npmjs.com/package/keytar) 7.0.0, upgrade this to the latest version: 7.6.0 and add caret versioning for the same.


hectormmg (approved):

sameerag:
@samuelkubai This is ready to be tested now. Can you please confirm extensions work as expected after the `keytar` upgrade and we can merge this?

samuelkubai (approved):
I have tested it and it works great.

PR #3421, opened 2021-04-08
Update docs to clarify RT lifetime

tnorling:
Received some feedback that the wording in the docs around refresh token renewal was confusing. Making this more explicit.

### DIFF
--- a/lib/msal-browser/docs/token-lifetimes.md
+++ b/lib/msal-browser/docs/token-lifetimes.md
@@ -18,7 +18,7 @@ Access tokens in the browser have a default recommended expiration of 1 hour. Af
 
 ### Refresh tokens
 
-Refresh tokens given to Single-Page Applications are limited-time refresh tokens (usually 24 hours from the time of retrieval). This is a non-adjustable lifetime. Whenever a refresh token is used to renew an access token, a new refresh token is fetched with the renewed access token. Once a refresh token has expired, a new authorization code flow must be initiated to retrieve an authorization code and trade it for a new set of tokens. (TODO: Add docs for RT here)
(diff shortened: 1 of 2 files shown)


lib/msal-browser/docs/token-lifetimes.md
```diff
@@ -18,7 +18,7 @@ Access tokens in the browser have a default recommended expiration of 1 hour. Af
 
 ### Refresh tokens
 
-Refresh tokens given to Single-Page Applications are limited-time refresh tokens (usually 24 hours from the time of retrieval). This is a non-adjustable lifetime. Whenever a refresh token is used to renew an access token, a new refresh token is fetched with the renewed access token. Once a refresh token has expired, a new authorization code flow must be initiated to retrieve an authorization code and trade it for a new set of tokens. (TODO: Add docs for RT here)
+Refresh tokens given to Single-Page Applications are limited-time refresh tokens (usually 24 hours from the time of retrieval). This is a non-adjustable, non-sliding window, lifetime. Whenever a refresh token is used to renew an access token, a new refresh token is fetched with the renewed access token. This new refresh token will have a lifetime equal to the remaining lifetime of the original refresh token. Once a refresh token has expired, a new authorization code flow must be initiated to retrieve an authorization code and trade it for a new set of tokens.
```
jasonnutter, line 21: It might be worth pointing out the old RT is not invalidated (until it naturally expires)?

jasonnutter: @hpsin curious if you have thoughts here.

jasonnutter (approved):

lib/msal-browser/docs/token-lifetimes.md
tnorling: done

sameerag: The difference between the new RT and the old RT is just the expiry time? I am concerned about caching, if we cache a given RT and try to fetch a new accessToken by setting `forceRefresh: true`, we get a new RT is my conclusion from this statement. Can we make this bulleted and in as a usage sequence or scenario based than a concept blob?

tnorling: >we get a new RT is my conclusion from this statement

This is correct. You get a new RT, but with a shorter expiration and the new RT replaces the cached RT.

jo-arroyo (approved):

PR #3422, opened 2021-04-08
Update browserslist and docs around browser support

### LINKED ISSUE #3396: MsalProvider gives Content Security Policy directive: "script-src 'self'" error
### Core Library

@azure/msal-browser

### Core Library Version

2.12.1

### Wrapper Library

@azure/msal-react

### Wrapper Library Version

1.0.0-beta.1

### Description

When I add `<MsalProvider instance={msalInstance}>` I get a "Content Security Policy" error and my app is not loading anymore.
Locally it works fine, but after build and serve it on the Nginx docker container in Azure infrastructure I get this error. If I remove this context initialization the app starts to load again.

### Error Message

runtime.js:747 Uncaught EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self'".

### Msal Logs

The app is crashing before it is even loaded

### MSAL Configuration

```javascript
{
    auth: {
        clientId: "my-client-id",
        authority: b2cPolicies.authorities.trustFramework.authority,
        knownAuthorities: [b2cPolicies.authorityDomain],
        redirectUri: `${uri}/login`,
    },
    cache: {
        storeAuthStateInCookie: isIE || isEdge || isFirefox
    }
};
```

### Relevant Code Snippets

```javascript
const msalInstance = new PublicClientApplication(msalConfig);
function App() {

  return (
      <Provider store={store}>
        <ThemeProvider theme={lightTheme}>
          <AppContainer />
        </ThemeProvider>
      </Provider>
    </MsalProvider>
  );
}

ReactDOM.render(
  <React.StrictMode>
    <App />
  </React.StrictMode>,
  document.getElementById('root')
);
```

### Reproduction Steps

1. Add `<MsalProvider instance={msalInstance}>` as shown in examples and the app is not loading

### Expected Behavior

It should not crash an app

### Identity Provider

Azure B2C Custom Policy

Chrome, Safari

### Regression

_No response_

### Source

External (Customer)

tnorling:
- Updates docs to explicitly define what browsers we support
- Moves IE 11 caveats/known-issues from wiki to new markdown doc
- Adds `.browserslistrc` defining which browsers we support (this helps prevent certain polyfills for older browsers from being included)
- Adds postinstall step to update the browser list db with the most up to date information (open to feedback around where this should go, it should happen at minimum before releasing)

Closes #3396 by removing polyfill that was causing a conflict with `script-src` CSP.

This was tested and does not break ie11 compatibility as long as the ie11 polyfills are still included in any apps that need to support ie11 (demonstrated in the react-router sample and ie 11 sample).


lib/msal-react/package.json
```diff
@@ -55,5 +55,10 @@
     "tslib": "^2.0.0",
     "typescript": "^3.9.7"
   },
-  "dependencies": {}
+  "dependencies": {},
+  "browserslist" : [
+    ">2%",
+    "not dead",
+    "last 2 years"
```
jasonnutter, line 62: It would be good for us to formally document this and make sure we are consistent in all the client side libs (namely msal browser and its wrapper libs)

jasonnutter: https://browserslist.dev/?q=bGFzdCAyIHZlcnNpb25zLCBub3QgZGVhZCwgPjIl

tnorling: last 2 versions will still pick up IE11 and include this polyfill

So choices basically are: 
1. Our browserslist supports ie11 and we live with the broken CSP
2. Our browserslist supports ie11 and we stop using tsdx (not totally opposed to this as it seems it's not well maintained and pretty opinionated but it does make many things easier...🤷‍♂️)
3. Our browserslist doesn't support ie11 but we tell people we are still supporting it and they need to provide the polyfills (which they need to do anyway even with choices 1 and 2)

Side note: I'm not really sure why this polyfill is included if we're not using the functions it's supposed to polyfill...I would have expected the bundler to be smarter than that. Or maybe we are using these functions and I don't see where...

jasonnutter: MSAL Browser already requires polyfills for IE11, so I think it is fine to require that for MSAL React.

PR #3455, opened 2021-04-14
Fixed a minor typo - `match` was spelled as `mach`

apratus:
Original text:

`The intended recipient of an **Access Token** is represented by the `aud` claim; in case the value for the `aud` claim does not mach the resource`

Updated text:

`The intended recipient of an **Access Token** is represented by the `aud` claim; in case the value for the `aud` claim does not match the resource`


derisen:
@apratus thanks for catching this. Please check the [contribution guide](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/contributing.md#creating-a-pr) for preparing your PR (in particular, "Beachball check" section).

jasonnutter (approved):

jasonnutter:
Since this is a simple doc change, don't worry about the beachball step for now. Merging.

PR #3471, opened 2021-04-15
Add .browserslistrc and update docs around browser support

### LINKED ISSUE #3396: MsalProvider gives Content Security Policy directive: "script-src 'self'" error

tnorling:
- Updates docs to explicitly define what browsers we support
- Moves IE 11 caveats/known-issues from wiki to new markdown doc
- Adds `.browserslistrc` defining which browsers we support (this helps prevent certain polyfills for older browsers from being included)
- Adds postinstall step to update the browserslist db with the most up to date browser information (open to feedback around where this should go, it should happen at minimum before releasing)

Closes #3396 by removing polyfill that was causing a conflict with `script-src` CSP.

This was tested and does not break ie11 compatibility as long as the ie11 polyfills are still included in any apps that need to support ie11 (demonstrated in the react-router sample and ie 11 sample).


jasonnutter (approved):

lib/msal-browser/FAQ.md
```diff
@@ -59,15 +59,34 @@
...
 
-IE 11, Edge, Chrome, Firefox and Safari
+- Chrome
+- Edge (Chromium)
+- Firefox
+- Safari
+- Opera
 
-Keep these steps in mind when [using MSAL.js with IE](https://github.com/AzureAD/microsoft-authentication-library-for-js/wiki/Using-msal.js-with-Internet-Explorer).
+MSAL.js has also been tested and supports the following browsers with Promise polyfills:
```
sameerag, line 70: Isn't this contrasting the statement above: "Despite this, IE 11 is currently supported but we do not include Promise polyfills"

sameerag: Never mind, is this 1.0?

tnorling: This is true for both v1 and v2. We do not include the polyfills, developers need to provide them and this is described in more detail in the IE doc.

sameerag: The statement is confusing in that case. Can we always explicitly mention that we do not include polyfills?

tnorling: Updated.

sameerag (approved):

PR #3483, opened 2021-04-18
acquire-access-token-example

indiraraghavan1986:


jasonnutter:
@indiraraghavan1986 Thanks for the PR, however, our guidance is to use the MSAL Interceptor to acquire tokens and attach them to http requests, instead of doing it manually. Closing.

PR #3487, opened 2021-04-19
Bound RT #1 - Generate Session Transport Key

hectormmg:
This PR:

- Adds `refreshTokenBinding` feature flag to Client system configuration
- Adds `stkJwk` attribute to `BaseAuthRequest` type so the STK JWK thumbprint can be cached with the interactive request
- Extends msal-browser's `CryptoOps` and `BrowserCrypto` class to support Access Token and Refresh Token PoP key generation options
- Adds `retrieveAsymmetricKey` API to PopTokenGenerator in msal-common
- Adds STK generation to `ClientApplication`'s `initializeAuthorizationCodeRequest` method
- Refactors key generation configuration to allow different algorithms and usages
- Updates tests to mock out key generation step and test new functionality


lib/msal-browser/src/utils/CryptoConstants.ts
```diff
@@ -0,0 +1,67 @@
...
+// Cryptographic Algorithms used/supported
+export enum Algorithms  {
+    PKCS1_V15_KEYGEN_ALG = "RSASSA-PKCS1-v1_5",
+    RSA_OAEP = "RSA-OAEP",
+    AES_GCM = "AES-GCM",
+    DIRECT = "dir",
+    S256_HASH_ALG = "SHA-256",
+}
+
+// Numerical constants relating to biy/bytelength
```
hectormmg:
```suggestion
// Numerical constants relating to bit/byte length
```

lib/msal-browser/src/crypto/CryptoOps.ts
```diff
@@ -86,34 +85,56 @@ export class CryptoOps implements ICrypto {
...
+            // Generate Thumbprint for Private Key
+            const privateKeyJwk: JsonWebKey = await this.browserCrypto.exportJwk(keyPair.privateKey);
+            // Re-import private key to make it unextractable
+            unextractablePrivateKey = await this.browserCrypto.importJwk(keyOptions, privateKeyJwk, false, keyOptions.privateKeyUsage);
+        }
 
         // Store Keypair data in keystore
         await this.cache.put(publicJwkHash, {
-            privateKey: unextractablePrivateKey,
+            privateKey: unextractablePrivateKey as CryptoKey,
```
tnorling: This could be undefined, no? Rather than casting let's make sure it actually is the correct type

lib/msal-common/test/client/RefreshTokenClient.spec.ts
```diff
@@ -314,6 +315,45 @@ describe("RefreshTokenClient unit tests", () => {
             expect(result.includes(`${AADServerParamKeys.X_CLIENT_CPU}=${TEST_CONFIG.TEST_CPU}`)).toBe(true);
             expect(result.includes(`${AADServerParamKeys.X_MS_LIB_CAPABILITY}=${ThrottlingConstants.X_MS_LIB_CAPABILITY_VALUE}`)).toBe(true);
         });
+
+        it.only("adds req_cnf if authenticationScheme is POP", async () => {
```
tnorling:
```suggestion
        it("adds req_cnf if authenticationScheme is POP", async () => {
```

lib/msal-browser/src/crypto/CryptoOps.ts
hectormmg: I've updated this method so the check above also checks if the private key is present and then that'll take care of the `CryptoKeyPair` type requirement. I left the private key as a possible undefined thinking the same API was going to be used for symmetric keys later but that's not the case. There should always be a private key in the object stored by this method. Thanks!

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -38,6 +39,7 @@ export type BrowserAuthOptions = {
     navigateToLoginRequestUrl?: boolean;
     clientCapabilities?: Array<string>;
     protocolMode?: ProtocolMode;
+    refreshTokenBinding?: boolean;
```
jasonnutter: We may consider putting this under `system`.

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
```diff
@@ -39,6 +39,21 @@ export abstract class StandardInteractionClient extends BaseInteractionClient {
...
+        // Generate Session Transport Key if Refresh Token Binding is enabled
+        if (this.config.auth.refreshTokenBinding) {
+            this.logger.verbose("Refresh token binding enabled, attempting to generate Session Transport Key");
+            try {
+                const sessionTransportKeyThumbprint = await this.browserCrypto.getPublicKeyThumbprint(request, CryptoKeyTypes.stk_jwk);
+                request.stkJwk = sessionTransportKeyThumbprint;
+                this.logger.verbose("Successfully generated and stored Session Transport Key");
+            } catch(error) {
+                if(error instanceof BrowserAuthError) {
+                    this.logger.error(error.message);
```
jasonnutter: I don't think you should log the message as-is, instead write a hard-coded message and maybe pull off certain fields such as the errorCode.

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
```diff
@@ -39,6 +39,21 @@ export abstract class StandardInteractionClient extends BaseInteractionClient {
...
+                    this.logger.verbose("Could not generate Session Transport Key, refresh token will be unbound");
```
jasonnutter: This should probably be `error`.

lib/msal-browser/src/utils/CryptoConstants.ts
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * JWK Key Format string (Type MUST be defined for window crypto APIs)
+ */
+export enum CryptoKeyFormats  {
+    jwk = "jwk"
+}
+
+// Crypto Key Usage sets
+export const KEY_USAGES = {
```
jasonnutter: A lot of inconsistency in terms of casing here. Can these be capital case?

```suggestion
export const KeyUsages = {
```

lib/msal-browser/src/utils/CryptoConstants.ts
```diff
@@ -0,0 +1,67 @@
...
+        KEYPAIR: ["sign", "verify"],
+        PRIVATE_KEY: ["sign"]
+    },
+    RT_BINDING: {
+        KEYPAIR: ["encrypt", "decrypt"],
+        PRIVATE_KEY: ["decrypt"]
+    }
+};
+
+export const CRYPTO_KEY_CONFIG = {
```
jasonnutter: Same comment re: casing. Also, is there more specific types we can declare for these consts?

lib/msal-common/src/utils/Constants.ts
```diff
@@ -346,6 +346,14 @@ export enum PasswordGrantConstants {
     password = "password"
 }
 
+/**
+ * Supported Cryptographic Key Types
+ */
+export enum CryptoKeyTypes {
+    req_cnf = "req_cnf",
+    stk_jwk = "stk_jwk"
```
jasonnutter:
```suggestion
    ReqCnf = "req_cnf",
    StkJwk = "stk_jwk"
```

jasonnutter: Something like that least.

lib/msal-common/src/crypto/KeyManager.ts
```diff
@@ -0,0 +1,56 @@
...
+ *      - uhw: hardware storage
+ */
+type ReqCnf = {
+    kid: string;
+    xms_ksl: KeyLocation;
+};
+
+enum KeyLocation {
+    SW = "sw",
+    UHW = "uhw"
```
jasonnutter: What do "sw" and "uhw" stand for? That should probably be the enum key value, in the interest of readability.

lib/msal-common/src/crypto/PopTokenGenerator.ts
```diff
@@ -35,20 +18,6 @@ export class PopTokenGenerator {
         this.cryptoUtils = cryptoUtils;
     }
 
-    async generateCnf(request: SignedHttpRequestParameters): Promise<string> {
```
jasonnutter, line 38: Arguably a breaking change, given you are removing a public api from an exported class.

lib/msal-browser/test/utils/StringConstants.ts
```diff
@@ -227,3 +234,25 @@ export const testNavUrl = `https://login.microsoftonline.com/common/oauth2/v2.0/
...
 export const testLogoutUrl = `https://login.microsoftonline.com/common/oauth2/v2.0/logout?post_logout_redirect_uri=${encodeURIComponent(`${TEST_URIS.TEST_REDIR_URI}`)}`;
+
+// Crypto Key Usage sets
+export const KEY_USAGES = {
+    AT_BINDING: {
+        KEYPAIR: ["sign", "verify"],
+        PRIVATE_KEY: ["sign"]
+    },
+    RT_BINDING: {
+        KEYPAIR: ["encrypt", "decrypt"],
```
jasonnutter: usages?

jasonnutter: Also, casing does not need to be uppercase.

lib/msal-common/src/crypto/KeyManager.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export class KeyManager {
```
jasonnutter: Can this be `CryptoKeyManager`? KeyManager doesnt feel specific enough.

lib/msal-browser/src/utils/CryptoConstants.ts
hectormmg: I went with the all caps snake case for consts and pascal case for enums, but I think making them all pascal case would look better.

lib/msal-browser/src/config/Configuration.ts
hectormmg: What's the rationale? I'm not very familiar with the semantic difference here.

lib/msal-common/src/crypto/KeyManager.ts
hectormmg: I can definitely change the name to `CryptoKeyManager` and open to calling it something other than `Manager` if we can come up with something better.

Having said that, based on the fact that refactoring `generateCnf` out of `PopTokenGenerator` is a breaking change, should I just pass these APIs to `PopTokenGenerator` and remove this class?

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
hectormmg: Replaced the logging here, please let me know if the new changes are better or if you think they still need changing. I think checking the code and doing a hard-coded error log based on that might be different than what you meant (A generic hard-coded error message + calling out the error code as a detail).

lib/msal-browser/src/utils/CryptoConstants.ts
hectormmg: I've updated the casing on these, but if you'd prefer the internal properties to be camelCase instead of PascalCase let me know.

lib/msal-common/src/crypto/KeyManager.ts
hectormmg: Updated the enum keys to say `SoftwareStorage` and `HardwareStorage` which is what these refer to.

lib/msal-common/src/crypto/PopTokenGenerator.ts
hectormmg: I hadn't entirely considered this. Should we remove the `KeyManager` class entirely and keep using `PoPTokenGenerator` even if it becomes a misnomer? Truth is `KeyManager` only sort of wraps calls to `CryptoOps` anyway.

lib/msal-browser/test/utils/StringConstants.ts
hectormmg: Removed these since we can import them from the `CryptoConstants` file.

lib/msal-browser/src/config/Configuration.ts
hectormmg: Done.

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -122,7 +124,7 @@ export function buildConfiguration({ auth: userInputAuth, cache: userInputCache,
         postLogoutRedirectUri: "",
         navigateToLoginRequestUrl: true,
         clientCapabilities: [],
-        protocolMode: ProtocolMode.AAD
+        protocolMode: ProtocolMode.AAD,
```
hectormmg:
```suggestion
        protocolMode: ProtocolMode.AAD
```

lib/msal-common/src/crypto/KeyManager.ts
jasonnutter: Yeah, just adding `retrieveAsymmetricPublicKey` to `PopTokenGenerator` is fine.

lib/msal-common/src/crypto/PopTokenGenerator.ts
hectormmg: Kept generateCnf in PopTokenGenerator and also moved CryptoKeyManager APIs into PopTokenGenerator.

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
```diff
@@ -39,6 +39,25 @@ export abstract class StandardInteractionClient extends BaseInteractionClient {
...
+        if (this.config.system.refreshTokenBinding) {
+            this.logger.verbose("Refresh token binding enabled, attempting to generate Session Transport Key");
+            try {
+                const sessionTransportKeyThumbprint = await this.browserCrypto.getPublicKeyThumbprint(request, CryptoKeyTypes.StkJwk);
+                request.stkJwk = sessionTransportKeyThumbprint;
+                this.logger.verbose("Successfully generated and stored Session Transport Key");
+            } catch(error) {
+                if(error instanceof BrowserAuthError) {
+                    if (error.errorCode === "key_generation_failed") {
+                        this.logger.error("Failed to generate Session Transport Key, refresh token will be unbound.");
```
tnorling: Are we swallowing this error? If so, should we just log these messages where the errors are thrown instead of throwing the error and catching?

lib/msal-common/src/crypto/ICrypto.ts
```diff
@@ -65,6 +66,12 @@ export interface ICrypto {
      * @param plainText
      */
     hashString(plainText: string): Promise<string>;
+    /**
+     * Returns the public key from an asymmetric key pair stored in IndexedDB based on the
+     * public key thumbprint parameter
+     * @param keyThumbprint
+     */
+    getAsymmetricPublicKey(keyThumbprint: string): Promise<string>;
```
tnorling, line 74: I believe this would be a major change

lib/msal-common/src/crypto/PopTokenGenerator.ts
```diff
@@ -23,32 +24,41 @@ type ReqCnf = {
...
 
     async generateCnf(request: SignedHttpRequestParameters): Promise<string> {
-        const reqCnf = await this.generateKid(request);
+        const reqCnf = await this.generateKid(request, CryptoKeyTypes.ReqCnf);
         return this.cryptoUtils.base64Encode(JSON.stringify(reqCnf));
     }
 
-    async generateKid(request: SignedHttpRequestParameters): Promise<ReqCnf> {
-        const kidThumbprint = await this.cryptoUtils.getPublicKeyThumbprint(request);
+    async generateKid(request: SignedHttpRequestParameters, keyType: CryptoKeyTypes): Promise<ReqCnf> {
```
tnorling, line 43: This is a breaking change as well

lib/msal-browser/src/crypto/CryptoOps.ts
```diff
@@ -95,9 +94,24 @@ export class CryptoOps implements ICrypto {
...
+        let keyOptions: CryptoKeyOptions;
+
+        switch(keyType) {
+            case CryptoKeyTypes.StkJwk:
+                keyOptions = CryptoKeyConfig.RefreshTokenBinding;
+                break;
+            default:
+                keyOptions = CryptoKeyConfig.AccessTokenBinding;
+                break;
+        }
```
jasonnutter: I think a ternary would be cleaner here (unless if you think there will be more types here in the future).

jasonnutter:
```suggestion
        const keyOptions: CryptoKeyOptions = keyType === CryptoKeyTypes.StkJwk ? CryptoKeyConfig.RefreshTokenBinding : CryptoKeyConfig.AccessTokenBinding;
```

lib/msal-browser/src/crypto/CryptoOps.ts
```diff
@@ -198,4 +212,22 @@ export class CryptoOps implements ICrypto {
...
+        const cachedKeyPair = await this.cache.asymmetricKeys.getItem(keyThumbprint);
+
+        if (!cachedKeyPair) {
+            throw BrowserAuthError.createSigningKeyNotFoundInStorageError();
+        }
+
+        // Get public key as JWK
+        const publicKeyJwk = await this.browserCrypto.exportJwk(cachedKeyPair.publicKey);
+        return BrowserCrypto.getJwkString(publicKeyJwk);
+    }
```
jasonnutter, line 252: Would be great to see some more log messages in this class.

lib/msal-browser/src/error/BrowserAuthError.ts
```diff
@@ -156,6 +156,10 @@ export const BrowserAuthErrorMessage = {
     databaseUnavailable: {
         code: "database_unavailable",
         desc: "IndexedDB, which is required for persistent cryptographic key storage, is unavailable. This may be caused by browser privacy features which block persistent storage in third-party contexts."
+    },
+    keyGenerationFailed: {
+        code: "key_generation_failed",
+        desc: "Failed to generate cryptographic key"
```
jasonnutter, line 162: Anything else we can provide here?

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
```diff
@@ -39,6 +39,25 @@ export abstract class StandardInteractionClient extends BaseInteractionClient {
...
+        // Generate Session Transport Key if Refresh Token Binding is enabled
+        if (this.config.system.refreshTokenBinding) {
+            this.logger.verbose("Refresh token binding enabled, attempting to generate Session Transport Key");
+            try {
+                const sessionTransportKeyThumbprint = await this.browserCrypto.getPublicKeyThumbprint(request, CryptoKeyTypes.StkJwk);
+                request.stkJwk = sessionTransportKeyThumbprint;
+                this.logger.verbose("Successfully generated and stored Session Transport Key");
+            } catch(error) {
+                if(error instanceof BrowserAuthError) {
+                    if (error.errorCode === "key_generation_failed") {
```
jasonnutter: If you stick with this, can you compare against `BrowserAuthErrorMessage.keyGenerationFailed.code`?

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -78,6 +79,7 @@ export type BrowserSystemOptions = SystemOptions & {
     redirectNavigationTimeout?: number;
     asyncPopups?: boolean;
     allowRedirectInIframe?: boolean;
+    refreshTokenBinding?: boolean;
```
jasonnutter, line 82: Given this PR doesn't ship the entire bound RT feature (ie enabling this flag wont result in a bound RT), I think it would be good if this flag was not exposed here and instead exposed in the final PR.

jasonnutter: I would propose creating a new temp/staging branch off of dev, and retarget your PRs to that branch, and then when the feature is ready e2e, open a new PR from that branch to dev (which should just require simple sign off instead of a full review).

lib/msal-browser/src/crypto/CryptoOps.ts
hectormmg: I did originally add the switch to allow for other key types in the future, but we can just extend the ternary to a switch when that happens, so keeping it cleaner may be wiser.

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
hectormmg: @tnorling The reason I did it this way is that these messages are bound RT specific but the error throwing method here (getPublicKeyThumbprint) is also used for AT PoP key generation (so messages have to stay generic). We do "swallow" the error because we want to smoothly fall back to unbound RT if key generation fails instead of aborting the auth request.

What do you think?

lib/msal-common/src/crypto/PopTokenGenerator.ts
hectormmg: Updated change file

lib/msal-common/src/crypto/ICrypto.ts
hectormmg: Updated change file

lib/msal-browser/src/crypto/CryptoOps.ts
hectormmg: Added some more logging, trying to balance verbose logging with not having redundant logging since the APIs that call these APIs do their own logging around the calls for protocol-specific logging.

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
tnorling: Gotcha, thanks for the context. Can you add a brief comment explaining why we're ignoring these particular errors and then we also need to rethrow any error that doesn't match your conditionals here

hectormmg: Addes comments and refactored the logic to make it cleaner here

lib/msal-browser/src/error/BrowserAuthError.ts
hectormmg: Added a details variable to the error creation method to get some more info from the failed calls to generateKeyPair

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
hectormmg: Updated to access the code string programatically

lib/msal-browser/src/config/Configuration.ts
hectormmg: Added an intermediary branch called refresh-token-binding to merge into

tnorling (approved):

lib/msal-browser/src/error/BrowserAuthError.ts
jasonnutter: I dont see that update?

jasonnutter: Ah nvm, github wasn't showing it to me.

jasonnutter (approved):

PR #3500, opened 2021-04-20
Remove token binding key from key store when removing access token with auth scheme

hectormmg:
This PR:
- Fixes a bug where unused keys where being left in IndexedDB when access tokens with auth scheme (pop) where renewed or when a user logged out and the cache was cleared
- Extracts cache clearing logic out of `AuthorizationCodeClient.getLogoutUri` and into Interaction Clients so underlying credential removal can be made async
- Adds unit tests that cover these changes


lib/msal-browser/src/cache/DatabaseStorage.ts
```diff
@@ -110,4 +110,32 @@ export class DatabaseStorage<T>{
...
+                return reject(BrowserAuthError.createDatabaseNotOpenError());
+            }
+
+            const transaction = this.db.transaction([this.tableName], "readwrite");
+
+            const objectStore = transaction.objectStore(this.tableName);
+
+            const dbDelete = objectStore.delete(key);
+
+            dbDelete.addEventListener("success", (e: any) => {
```
jasonnutter, line 134: I'm about to disallow `any`:

```suggestion
            dbDelete.addEventListener("success", (e: Event) => {
```

lib/msal-browser/src/cache/DatabaseStorage.ts
```diff
@@ -110,4 +110,32 @@ export class DatabaseStorage<T>{
...
+    /**
+     * Removes item from IndexedDB under given key
+     * @param key
+     */
+    async delete(key: string): Promise<boolean> {
+        if (!this.dbOpen) {
+            await this.open();
+        }
+
+        return new Promise<boolean>((resolve, reject) => {
```
jasonnutter, line 123:
```suggestion
        return new Promise<boolean>((resolve: Function, reject: Function) => {
```

lib/msal-common/src/config/ClientConfiguration.ts
```diff
@@ -174,12 +174,14 @@ export function buildClientConfiguration(
...
+    
+    const loggerOptions = { ...DEFAULT_LOGGER_IMPLEMENTATION, ...userLoggerOption };
+    
     return {
         authOptions: buildAuthOptions(userAuthOptions),
         systemOptions: { ...DEFAULT_SYSTEM_OPTIONS, ...userSystemOptions },
-        loggerOptions: { ...DEFAULT_LOGGER_IMPLEMENTATION, ...userLoggerOption },
-        storageInterface: storageImplementation || new DefaultStorageClass(userAuthOptions.clientId, DEFAULT_CRYPTO_IMPLEMENTATION),
+        loggerOptions: loggerOptions,
+        storageInterface: storageImplementation || new DefaultStorageClass(userAuthOptions.clientId, DEFAULT_CRYPTO_IMPLEMENTATION, new Logger(loggerOptions)),
```
tnorling, line 184: You should pass in the name and version from packageMetadata here too

samples/package-lock.json
```diff
@@ -0,0 +1,6005 @@
+{
+    "name": "msal-js-samples",
```
tnorling, line 2: This package-lock was removed. Please delete and add a `.npmrc` file with the line `package-lock=false`

hectormmg: Will do, thanks!

lib/msal-common/src/config/ClientConfiguration.ts
hectormmg: Done!

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -554,6 +557,19 @@ export abstract class CacheManager implements ICacheManager {
...
         const key = credential.generateCredentialKey();
+
+        // Remove Token Binding Key from key store for Auth Scheme Credentials
+        if (credential.credentialType.toLowerCase() === CredentialType.ACCESS_TOKEN_WITH_AUTH_SCHEME.toLowerCase()) {
+            const accessTokenWithAuthSchemeEntity = credential as AccessTokenEntity;
+            const kid = accessTokenWithAuthSchemeEntity.keyId;
+
+            if (kid) {
+                this.cryptoImpl.removeTokenBindingKey(kid).catch((error) => {
+                    this.logger.verbose(`Crypto Keypair was not removed from key store: ${error}`);
```
jasonnutter, line 568:
```suggestion
                    this.logger.error(`Crypto Keypair was not removed from key store: ${error}`);
```

jasonnutter: And do we want to make this method async?

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -22,17 +22,20 @@ import { ThrottlingEntity } from "./entities/ThrottlingEntity";
...
 /**
  * Interface class which implement cache storage functions used by MSAL to perform validity checks, and store tokens.
  */
 export abstract class CacheManager implements ICacheManager {
     protected clientId: string;
     protected cryptoImpl: ICrypto;
+    protected logger: Logger;
 
-    constructor(clientId: string, cryptoImpl: ICrypto) {
+    constructor(clientId: string, cryptoImpl: ICrypto, logger: Logger) {
```
tnorling, line 35: This may be a breaking change

hectormmg: Do we expect someone to use the CacheManager class from common? If not, both NodeCacheManager and BrowserCacheManager have the logger as a constructor param, just changed it so they pass it to super instead of setting it for themselves. How likely is it to be a breaking change?

tnorling: Issue isn't someone using the CacheManager directly. Issue is that browser and node use it today. So if someone is using an old version of browser/node and the newest version of common this will throw an error when the platform lib calls `new CacheManager(clientId, cryptoImpl)` and common is now expecting `new CacheManager(clientId, cryptoImpl, logger)`.

tnorling: As I'm thinking about this more, we shouldn't be passing the logger down from the platform lib to common anyway. Common should be instantiating its own logger so that the log messages properly state they are coming from common.

hectormmg: That's a great point, I'll fix this. Thanks!

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -554,6 +557,19 @@ export abstract class CacheManager implements ICacheManager {
...
     removeCredential(credential: CredentialEntity): boolean {
         const key = credential.generateCredentialKey();
+
+        // Remove Token Binding Key from key store for Auth Scheme Credentials
+        if (credential.credentialType.toLowerCase() === CredentialType.ACCESS_TOKEN_WITH_AUTH_SCHEME.toLowerCase()) {
+            const accessTokenWithAuthSchemeEntity = credential as AccessTokenEntity;
+            const kid = accessTokenWithAuthSchemeEntity.keyId;
+
+            if (kid) {
+                this.cryptoImpl.removeTokenBindingKey(kid).catch((error) => {
```
jasonnutter, line 567: Can this function be made async?

hectormmg: Done

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -781,8 +784,12 @@ export abstract class ClientApplication {
             const authClient = await this.createAuthCodeClient(serverTelemetryManager, requestAuthority, validRequest.correlationId);
             browserRequestLogger.verbose("Auth code client created");
 
-            // create logout string and navigate user window to logout. Auth module will clear cache.
+            // Create logout string and navigate user window to logout.
             const logoutUri: string = authClient.getLogoutUri(validRequest);
+
+            // Clear cache on logout
+            await authClient.clearCacheOnLogout(validRequest);
```
jasonnutter, line 791: What should happen if this throws?

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -121,19 +121,30 @@ export class AuthorizationCodeClient extends BaseClient {
...
+                await this.cacheManager.removeAccount(AccountEntity.generateAccountCacheKey(logoutRequest.account));
+                this.logger.verbose("Cleared cache items belonging to the account provided in the logout request.");
+            } catch (error) {
+                this.logger.verbose("Account provided in logout request was not found.");
+            }
         } else {
             // Clear all accounts and tokens
-            this.cacheManager.clear();
+            this.logger.verbose("No account provided in logout request, clearing all cache items.");
+            await this.cacheManager.clear();
```
jasonnutter, line 146: Also curious to know what should happen if this throws

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -190,7 +190,7 @@ export abstract class CacheManager implements ICacheManager {
      * saves a cache record
      * @param cacheRecord
      */
-    saveCacheRecord(cacheRecord: CacheRecord): void {
+    async saveCacheRecord(cacheRecord: CacheRecord): Promise<void> {
```
jasonnutter, line 193: Is this a breaking change?

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -499,38 +501,42 @@ export abstract class CacheManager implements ICacheManager {
     /**
      * Removes all accounts and related tokens from cache.
      */
-    removeAllAccounts(): boolean {
+    async removeAllAccounts(): Promise<boolean> {
```
jasonnutter, line 504: Same question re: breaking change

jasonnutter:
Overall looks good. Two main questions are what should happens if there are errors (I would lean towards having things fail silently, as we don't want apps to get stuck in a state where they can't logout), and whether or not the changes in common are breaking changes (if so, we should make it a major bump for common).

lib/msal-common/src/client/AuthorizationCodeClient.ts
hectormmg: I'd be interested in your opinion on what we should do if it throws besides try/catching the error and logging that cache clearing failed. We can proceed with the ending the session with the server even if the cache isn't clear, but the user will still be functionally logged in (having credentials in the cache).

lib/msal-common/src/cache/CacheManager.ts
hectormmg: Correct me if I'm wrong, but `saveCacheRecord` is only used internally by the `ResponseHandler` in common, so this isn't a breaking change. However, do we need to update the method signature in `ICacheManager`? If so, I think that would constitute a breaking change. Although it doesn't seem that our platform libs implement `saveCacheRecord` since common has all the logic.

lib/msal-common/src/cache/CacheManager.ts
hectormmg: `removeAllAccounts is called by the `BrowserCacheManager` so I think this is a breaking change. How big of a deal would it be to make a major version bump for common? I don't believe there's anyway to correctly implement logout without making this `async`.

lib/msal-browser/src/app/ClientApplication.ts
hectormmg: The current implementation doesn't do error handling for the cache clearing, open to suggestions on how we should proceed if this part throws and the cache isn't cleared successfully. Is try/catch and logging the failure sufficient?

lib/msal-common/src/cache/CacheManager.ts
jasonnutter: Not a big deal if this it needed to do this properly.

lib/msal-common/src/client/AuthorizationCodeClient.ts
jasonnutter: Yeah, catching and logging is fine. Would prefer local cache potentially being out of date then the app being stuck.

change/@azure-msal-common-7f106e66-8f0f-4bfe-887f-3199473210de.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
jasonnutter, line 2:
```suggestion
  "type": "major",
```

Looks like we need this to be a major bump instead.

jasonnutter (changes requested):
Major bump needed for common breaking changes.

jasonnutter (approved):
Looks like there are merge conflicts, approved otherwise!

change/@azure-msal-node-dd66a214-eef6-48b8-b2cd-bf05ccbdda1c.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
```
tnorling, line 2:
```suggestion
  "type": "patch",
```

tnorling (approved):
2 minor suggestions. Approving otherwise

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -128,7 +128,11 @@ export class PublicClientApplication extends ClientApplication implements IPubli
         const silentRequest = silentCacheClient.initializeSilentRequest(request, account);
         this.eventHandler.emitEvent(EventType.ACQUIRE_TOKEN_START, InteractionType.Silent, request);
 
-        return silentCacheClient.acquireToken(silentRequest).catch(async () => {
+        return silentCacheClient.acquireToken(silentRequest).catch(async (error) => {
+            if (error instanceof BrowserAuthError && error.errorCode === BrowserAuthErrorMessage.signingKeyNotFoundInStorage.code) {
+                this.logger.verbose("Signing keypair for bound access token not found. Refreshing bound access token and generating a new crypto keypair.");
```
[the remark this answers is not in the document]
hectormmg: Done, thanks!

sameerag (approved):
Clarified all my questions offline, Please add docs stating the sync and async storage calls for different tokens vs clear being async always in the POP document. 

Approved, Good job with the tests.

PR #3502, opened 2021-04-21
Fix missing semicolon in code example

TastyDucks:
I noticed that an import statement ended with `l` instead of `;`.


jasonnutter:
Thanks!

PR #3505, opened 2021-04-21
Bound RT #2 - Request Bound Refresh Token

hectormmg:
This PR:
- Adds `stk_jwk` attribute to auth code URL and token request body in `AuthorizationCodeClient` in order to request bound refresh tokens from ESTS
- Adds an msal-browser sample to test refresh token binding
- Extends `AuthorizationCodeClient` tests to include stkJwk in every test request


samples/msal-browser-samples/VanillaJSTestApp2.0/app/boundRt/test/browser.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+        // Click Sign In
+        await page.click("#SignIn");
+        await page.waitForSelector("#loginRedirect");
+        await screenshot.takeScreenshot(page, "signInClicked");
+        // Click Sign In With Redirect
+        await page.click("#loginRedirect");
+        // Check URL to see if it contains stkJwk
+        await page.waitForNavigation({ waitUntil: "networkidle0", timeout: 10000});
+        await page.waitForSelector("input#i0116.input.text-box");
+        expect(page.url()).to.include("stk_jwk");
```
tnorling, line 71: Any way we can test the full E2E flow here, given that the integration tests should already cover this?

tnorling (approved):
One comment, looks good otherwise. Given this is going into a staging branch I consider this non-blocking for now

jasonnutter (approved):

lib/msal-common/src/request/RequestParameterBuilder.ts
```diff
@@ -383,6 +384,30 @@ export class RequestParameterBuilder {
         this.parameters.set(AADServerParamKeys.X_MS_LIB_CAPABILITY, ThrottlingConstants.X_MS_LIB_CAPABILITY_VALUE);
     }
 
+    /**
+     * Add stk_jwk thumbprint to query params
+     * @param stkJwkKid
+     */
+    addStkJwkThumbprint(stkJwkKid: string): void {
```
sameerag, line 391: What happens if this is null/empty? Do we handle that where we call this?

hectormmg: It just doesn't get added, it most likely won't get called if the stk_jwk hasn't been set, we do check before we call this.

samples/msal-browser-samples/VanillaJSTestApp2.0/app/boundRt/test/browser.spec.ts
hectormmg: These PRs slowly increase what is tested as the feature gets developed. The reason we can't test the whole flow is that the auth response at this point is encrypted and the decryption steps are added in subsequent PRs.

I'm relying heavily on the staging branch and the feature flag to be able to separate these PRs in a way that allow for the protocol to be reviewed "in order".

lib/msal-common/src/request/RequestParameterBuilder.ts
sameerag: Thanks

sameerag (approved):

PR #3506, opened 2021-04-21
Prevent MSAL Guard from navigating app to url with code hash

jasonnutter:
When the redirect URI was set to a page with the MSAL Guard, the guard could end up navigating the user to `/#code`, even though it should be cleared from the URL when processing the redirect. This change makes sure the user isn't navigated back to a url with the `#code=` hash response still in the url.

### DIFF
--- a/change/@azure-msal-angular-b0172103-6866-4f25-9d89-4084d829edd3.json
+++ b/change/@azure-msal-angular-b0172103-6866-4f25-9d89-4084d829edd3.json
@@ -1,6 +1,6 @@
 {
   "type": "prerelease",
-  "comment": "Prevent MSAL Guard from navigating app to url with code hash",
+  "comment": "Prevent MSAL Guard from navigating app to url with code hash #3506",
   "packageName": "@azure/msal-angular",
   "email": "janutter@microsoft.com",
   "dependentChangeType": "patch"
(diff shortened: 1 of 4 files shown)


lib/msal-angular/docs/v2-docs/known-issues.md
```diff
@@ -1,5 +1,8 @@
 # Known issues for MSAL Angular v2
 
+## 2.0.0-beta.4
+* When the MSAL Guard is used on the page used for the redirect URI, the `code=` hash may remain in the url. This is addressed in `2.0.0-beta.5`, expect when the MSAL Guard is used for `canLoad`. To mitigate this issue, applications should not put the MSAL Guard on the page used for the redirect URI.
```
tnorling, line 4:
```suggestion
* When the MSAL Guard is used on the page used for the redirect URI, the `code=` hash may remain in the url. This is addressed in `2.0.0-beta.5`, except when the MSAL Guard is used for `canLoad`. To mitigate this issue, applications should not put the MSAL Guard on the page used for the redirect URI.
```

change/@azure-msal-angular-b0172103-6866-4f25-9d89-4084d829edd3.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "Prevent MSAL Guard from navigating app to url with code hash",
```
tnorling, line 3:
```suggestion
  "comment": "Prevent MSAL Guard from navigating app to url with code hash #3506",
```

tnorling (approved):

jo-arroyo (approved):

PR #3509, opened 2021-04-22
Look for temp cache items in local storage as fallback

jasonnutter:
Since 2.10, temp cache items have been stored in session storage or in memory, instead of local storage (if configured). This can result in unexpected cache misses when upgrading between versions. This change will look in local storage if temp cache items aren't found in session/memory storage.


jo-arroyo (approved):

change/@azure-msal-browser-332c1b46-7af3-4c8e-a6d8-d34fc6f729fa.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Look for temp cache items in local storage as fallback",
```
tnorling, line 3:
```suggestion
  "comment": "Look for temp cache items in local storage as fallback #3509",
```

tnorling (approved):

PR #3527, opened 2021-04-23
Fix extra state update in useAccount hook

### LINKED ISSUE #3522: React B2C Sample - Token is requested twice
### Core Library

@azure/msal or msal

### Core Library Version

2.14.0

### Wrapper Library

@azure/msal-react

### Wrapper Library Version

1.0.0-beta.0

### Description

When requesting an access token, it is making 2 requests.

### Error Message

_No response_

### Msal Logs

_No response_

### MSAL Configuration

```javascript
export const msalConfig = {
    auth: {
        clientId: "9067c884-9fa6-414f-9aa4-a565b1cb46be",
        authority: "https://fabrikamb2c.b2clogin.com/fabrikamb2c.onmicrosoft.com/b2c_1_susi",
        knownAuthorities: ["fabrikamb2c.b2clogin.com"],
        redirectUri: "http://localhost:4200",
        postLogoutRedirectUri: "http://localhost:4200"
    }
};

export const loginRequest = {
    scopes: ["https://fabrikamb2c.onmicrosoft.com/helloapi/demo.read"]
};

export const forgotPasswordRequest = {
    authority: "https://fabrikamb2c.b2clogin.com/fabrikamb2c.onmicrosoft.com/b2c_1_reset"
}
```

### Relevant Code Snippets

```javascript
useEffect(() => {
        if (account && inProgress === "none") {
            instance.acquireTokenSilent({
                ...loginRequest,
                account: account
            }).then((response) => {
                setAtsResponse(response);
                console.log('Getting Token')
            });
        }
    }, [account, inProgress, instance]);
```

### Reproduction Steps

Replace the matching code within Protected.jsx as it has the console log section:
console.log('Getting Token')

Then run and look at the console log.

### Expected Behavior

It should only be ran once

### Identity Provider

Azure B2C Basic Policy

Chrome

### Regression

_No response_

### Source

External (Customer)

tnorling:
The `useAccount` hook has a minor bug that causes an extra state update which potentially causes unnecessary re-renders or effects to run again. 

This PR adds a `compareClaims` flag to the `accountInfoIsEqual` function if set to true the comparison will take into account the claims on each of the account objects passed. It specifically checks for the "iat" and "nonce" claims which should be sufficient to determine whether or not the account has changed.

Additionally, `useAccount` will compare the current account in state to the current account in the cache and only update if the account has changed.

Fixes #3522


lib/msal-common/src/cache/entities/AccountEntity.ts
```diff
@@ -277,19 +277,31 @@ export class AccountEntity {
...
         }
+
+        let claimsMatch = true; // default to true so as to not fail comparison below if compareClaims: false
+        if (compareClaims) {
+            const accountAClaims = (accountA.idTokenClaims || {}) as TokenClaims;
+            const accountBClaims = (accountB.idTokenClaims || {}) as TokenClaims;
+
+            // issued at timestamp and nonce are expected to change each time a new id token is acquired
+            claimsMatch = (accountAClaims.iat === accountBClaims.iat) &&
+            (accountAClaims.nonce === accountBClaims.nonce);
```
jasonnutter, line 297: Do we need to check both? Seems like checking nonce should be enough in theory?

tnorling: As far as I could tell `nonce` is optional according to the OIDC spec whereas `iat` is required. Included both just to provide a fallback in case one or the other is not present.

jasonnutter: Okay makes sense, thanks.

jo-arroyo (approved):

lib/msal-react/src/hooks/useAccount.ts
```diff
@@ -42,7 +42,10 @@ export function useAccount(accountIdentifiers: AccountIdentifiers): AccountInfo
     const [account, setAccount] = useState<AccountInfo|null>(initialStateValue);
 
     useEffect(() => {
-        setAccount(getAccount(instance, accountIdentifiers));
+        const currentAccount = getAccount(instance, accountIdentifiers);
+        if (!AccountEntity.accountInfoIsEqual(account, currentAccount, true)) {
+            setAccount(currentAccount);
+        }
     }, [inProgress, accountIdentifiers, instance]);
```
jasonnutter, line 49: Thoughts on the warning below?

tnorling: Thanks, addressed. May consider making that an error instead of a warning

jasonnutter (approved):

lib/msal-common/src/cache/entities/AccountEntity.ts
```diff
@@ -277,19 +277,31 @@ export class AccountEntity {
...
+        }
+
         return (accountA.homeAccountId === accountB.homeAccountId) && 
             (accountA.localAccountId === accountB.localAccountId) &&
             (accountA.username === accountB.username) &&
             (accountA.tenantId === accountB.tenantId) &&
-            (accountA.environment === accountB.environment);
+            (accountA.environment === accountB.environment) &&
```
sameerag, line 304: We provide environment aliasing, shouldn't that be accounted here?

tnorling: No, if it doesnt match that would mean the account object has changed so whatever is checking for equality should update

sameerag: Makes sense

sameerag (approved):

PR #3536, opened 2021-04-26
Fix npm audit warnings for node extensions

jasonnutter:


.github/workflows/npm-audit.yml
```diff
@@ -25,6 +26,19 @@ jobs:
     - name: npm audit (root)
       run: npm audit --production
 
+  audit-extensions:
+    runs-on: ubuntu-latest
```
sameerag, line 30: Is this git requirement? Ubuntu?

sameerag (approved):

.github/workflows/npm-audit.yml
jasonnutter: It is just for running the `npm audit` command, nothing else, so just using what we do for the other libs.

tnorling (approved):

PR #3544, opened 2021-04-27
Express.js example update

diberry:
* instructions in readme
* /App/data/cache.json - fixes [Error: ENOENT: no such file or directory, open 'C:\repos\code\sdk\microsoft-authentication-library-for-js\samples\msal-node-samples\ExpressTestApp\TestApp\App\data\cache.json']
* Add samples Uris so customer doesn't have to figure out what they are


diberry:
I would like to make a change to how the TS project and JS project in this sample are related so that the project can easily be deployed to Azure app service. There are a few ways to do that - what do you think?

jasonnutter:
> I would like to make a change to how the TS project and JS project in this sample are related so that the project can easily be deployed to Azure app service. There are a few ways to do that - what do you think?

Definitely open to feedback and ways to make it easier to use!

derisen (approved):
LGTM!

derisen:
> I would like to make a change to how the TS project and JS project in this sample are related so that the project can easily be deployed to Azure app service. There are a few ways to do that - what do you think?

@diberry sounds good! admittedly neglected the deployment aspect here..

diberry:
@derisen @jasonnutter This is a really awesome sample - I would like to build a tutorial around it to put in the [JS Dev Experience docs](https://docs.microsoft.com/en-us/azure/developer/javascript/). Any issue with that?

derisen:
> @derisen @jasonnutter This is a really awesome sample - I would like to build a tutorial around it to put in the [JS Dev Experience docs](https://docs.microsoft.com/en-us/azure/developer/javascript/). Any issue with that?

@diberry sounds good to me! There are a few improvements I'm planning to work on in the next couple of days e.g. simpler way of using certificates instead of client secrets and using [passport-azure-ad](https://github.com/AzureAD/passport-azure-ad) for token validation etc.

diberry:
#sign-off please merge

samples/msal-node-samples/ExpressTestApp/TestApp/appSettings.json
```diff
@@ -6,8 +6,8 @@
     },
     "settings": {
         "homePageRoute": "/home",
-        "redirectUri": "Enter_the_Redirect_Uri_Here",
-        "postLogoutRedirectUri": "Enter_the_Post_Logout_Redirect_Uri_Here"
+        "redirectUri": "http://localhost:4000/redirect",
```
sameerag, line 9: Why are we not leaving this customizable? This mandates the user to have `http://localhost:4000/redirect` as a specific redirectURI to be registered. Is the goal to have a one click working sample?

sameerag: Ah! I see now. We are doing it above but leaving it open here. Never mind.

sameerag (approved):

samples/msal-node-samples/ExpressTestApp/TestApp/appSettings.json
sameerag: Can we also add some comments, which will mention that this is customizable?

derisen: Can't put comments in JSON :( 
README mentions its for example though.

sameerag: Makes sense. Resolving.

Issue #3246, opened 2021-03-19
Use HTTP verbs when mapping protected resources


NDalvise:
## Library
- [x] `@azure/msal-angular@2.0.0-beta.1`
- [ ] `@azure/msal-browser@2.12.1`

## Description
In the Angular wrapper for MSAL the MsalInterceptorConfiguration maps the protected endpoints through the protectedResourceMap object. The mapping is done by matching a URI and an array of scopes.

Suppose to interact with a RESTful web API backend that expose an endpoint (https://example.com/api/resource). That endpoint produces different results based on the HTTP verb used to call it. By design the GET call is unprotected (everyone can call it) and the POST call is protected (only authorized users should call it).

In this scenario the Angular interceptor is not able to see the difference between the two calls because it works mapping the bare URI to a set of scopes. It could use the "HTTP verb + URI" pair to map endpoints that at the same URI have different authorization scopes based on the HTTP verb used.

## Source

- [ ] Internal (Microsoft)
- [x] Customer request

jasonnutter:
@NDalvise Thanks for the feature request, we have also been considering adding this enhacement. We'll add it to the backlog.

PR #3546, opened 2021-04-28
Update MsalInterceptor to accept HTTP methods in protectedResourceMap

jo-arroyo:
Currently, the Msal Interceptor attaches the auth header with GET requests even if it is not needed. This PR changes the `protectedResourceMap` to allow accepting an object with HTTP methods and scopes for that method, allowing different tokens to be attached to different HTTP methods.

This PR also adds unit tests and a `MsalInterceptor` doc, and addresses issue #3246.


lib/msal-angular/src/msal.interceptor.config.ts
```diff
@@ -11,6 +11,11 @@ export declare type MsalInterceptorAuthRequest = Omit<PopupRequest, "scopes"> |
 
 export type MsalInterceptorConfiguration = {
     interactionType: InteractionType.Popup | InteractionType.Redirect;
-    protectedResourceMap: Map<string, Array<string> | null>;
+    protectedResourceMap: Map<string, Array<string|ProtectedResourceScopes> | null>;
     authRequest?: MsalInterceptorAuthRequest | ((msalService: MsalService, req: HttpRequest<unknown>, originalAuthRequest: MsalInterceptorAuthRequest) => MsalInterceptorAuthRequest);
 };
+
+export type ProtectedResourceScopes = {
```
jo-arroyo, line 18: Open to suggestions of a different shape for this. Another possibility could be `type ProtectedResourceScopes = Record<string, string[]>`, which would support:
```javascript
["http://applicationD.com", [
  "all.scope",
  {"GET": ["read.scope"]}
]]
```

jasonnutter: I'll think about this overnight, but I think what you have here is good, as it allows us to easily add more things, and the typings will be more helpful.

lib/msal-angular/docs/v2-docs/msal-interceptor.md
```diff
@@ -0,0 +1,157 @@
...
+            ["https://myapplication.com/unprotected", null]
+            ["https://myapplication.com", ["custom.scope"]]
+        ]),
+    }
+    ```
+
+Other things to note regarding the `protectedResourceMap`:
+
+* **Wildcards**: `protectedResourceMap` supports using `*` for wildcards. When using wildcards, if multiple matching entries are found in the `protectedResourceMap`, the first match found will be used (based on the order of the `protectedResourceMap`). 
+* **Relative paths**: If there are relative resource paths in your application, you may need to provide the relative path in the `protectedResourceMap`. This also applies to issues that may arise with ngx-translate. Relative paths should not have a leading slash.
```
jasonnutter, line 147: "Relative paths should not have a leading slash." Is this always true? I feel like I've seen scenarios where a leading slash was needed.

lib/msal-angular/src/msal.interceptor.ts
```diff
@@ -122,19 +140,62 @@ export class MsalInterceptor implements HttpInterceptor {
...
+            });
+
+            // Only add to all scopes if scopes for endpoint and method is found
+            if (scopesForEndpoint.length > 0) {
+                allMatchedScopes.push(scopesForEndpoint);
             }
+        });
+
+        if (allMatchedScopes.length > 0) {
+            // Returns scopes for first matching endpoint
```
jasonnutter, line 197: I think it may be worth adding a warning log message here is more than 1 matched scopes are found.

jasonnutter (approved):
This is great, thanks for the tests and docs!

lib/msal-angular/docs/v2-docs/msal-interceptor.md
tnorling: Is this a decision we made or a limitation of a dependency? If it's the former I think we should mandate the opposite for consistency (we require a leading slash for redirectUris/postLogoutRedirectUris if you provide a relative url)

lib/msal-angular/src/msal.interceptor.config.ts
jasonnutter: Curious to hear other's thoughts, but I'm cool with the object shape defined here.

lib/msal-angular/docs/v2-docs/msal-interceptor.md
jo-arroyo: Looking back at customers who have experienced using relative paths, it looks like a leading slash works for some but not for others. We don't know what the relative paths in their apps are going to be, but putting a relative path without a leading slash in the `protectedResourceMap` will work for endpoints with or without a leading slash. 
(i.e. `relative` in the PRM will work for `/relative` and `relative` endpoints, but `/relative` in the PRM will not work for `relative` endpoints).

lib/msal-angular/src/msal.interceptor.ts
jo-arroyo: Sure. Added.

lib/msal-angular/docs/v2-docs/msal-interceptor.md
jo-arroyo: I've updated the guidance on this.

lib/msal-angular/src/msal.interceptor.config.ts
tnorling: I'm a fan of this approach

tnorling (approved):

PR #3553, opened 2021-04-29
Acquire with device code interval fix

### LINKED ISSUE #3476: Avoid calling asynchronous requests from `setInterval()`
### Core Library

@azure/msal-node

### Core Library Version

Latest

### Wrapper Library

Not Applicable

### Wrapper Library Version

Latest

### Description

Hello team! I'm reporting a bug which happens in the odd case that a network request might take longer than an interval. In such cases, users will experience their Node process not to close properly, and errors might not be able to bubble up appropriately.

This issue happens because `acquireTokenWithDeviceCode(` has a `setInterval` that triggers an async function in this line: [msal-common/src/client/DeviceCodeClient.ts#L156](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/44bc5edc3104400bef0e876c68ea529ce543e398/lib/msal-common/src/client/DeviceCodeClient.ts#L156). The interval is triggered with a duration of `pollingIntervalMilli`.

The function sent to that `setInterval` is an asynchronous function that calls and waits for `executePostToTokenEndpoint(`.

It could be the case that `executePostToTokenEndpoint(` can take longer than the interval's duration. In which case, `setInterval` will trigger the asynchronous function again, since `setInterval` does not wait for the internal function to finish.

I am aware that `pollingIntervalMilli` is set to be as many seconds as `deviceCodeResponse.interval` says ([source link](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/44bc5edc3104400bef0e876c68ea529ce543e398/lib/msal-common/src/client/DeviceCodeClient.ts#L148)). In the cases I've been able to test, this means 5 seconds. Depending on the network speed and the availability of the services, `executePostToTokenEndpoint(` might not be fulfilled on time.

For that `acquireTokenWithDeviceCode(`, this could be fixed by using an for loop, similar to:

```ts
// Assuming the for loop is within an async function, of course
private async acquireTokenWithDeviceCode(
  /* ... */

  // Instead of returning a promise and calling to setInterval inside of that promise...

  for (let i = 0; i <= maxIterations; i++) {
    if (request.cancel) {
      /* ... */
    } else if (userSpecifiedTimeout && userSpecifiedTimeout < deviceCodeExpirationTime && TimeUtils.nowSeconds() > userSpecifiedTimeout) {
      /* ... */
    } else if (TimeUtils.nowSeconds() > deviceCodeExpirationTime) {
      /* ... */
    } else {
      /* ... */
      const response = await this.executePostToTokenEndpoint(/* ... */);
    
      if (response.body && response.body.error === Constants.AUTHORIZATION_PENDING) {
        this.logger.info(response.body.error_description || "no_error_description");
      } else {
        return response.body;
      }
      
      // Here or somewhere near this we would wait some time.
      // In this example we use a new function: delay
      // The for loop won't start again before this promise is resolved.
      await delay(pollingIntervalMilli);
      
    }
  }

}
```

For the code above to work, `delay` could be defined as follows:

```ts
function delay<T>(t: number, value?: T): Promise<T | void> {
  return new Promise((resolve) => setTimeout(() => resolve(value), t));
}
```

This issue is not blocking us.

### Error Message

_No response_

### Msal Logs

_No response_

### MSAL Configuration

```javascript
Not applicable.
```

### Relevant Code Snippets

```javascript
See the description.
```

### Reproduction Steps

To reproduce this one would need to mock the HTTP layer to artificially introduce a delay longer than the interval duration.

### Expected Behavior

The polling of the device code endpoint should not happen before the interval finishes.

### Identity Provider

Other

Other

### Regression

_No response_

### Source

Internal (Microsoft)

sadasant:
This is a proposal. I can't get this repo to work locally yet! Lerna doesn't get installed in my VM.

Something similar to this would:
Fix #3476


lib/msal-common/src/client/DeviceCodeClient.ts
```diff
@@ -137,72 +137,73 @@ export class DeviceCodeClient extends BaseClient {
...
+
+      const userSpecifiedTimeout = request.timeout ? TimeUtils.nowSeconds() + request.timeout : undefined;
+      const deviceCodeExpirationTime = TimeUtils.nowSeconds() + deviceCodeResponse.expiresIn;
+      const pollingIntervalMilli = deviceCodeResponse.interval * 1000;
+
+      /**
+       * Breaks the polling with specific conditions.
+       */
+      const continuePolling = (): boolean => {
+        if (request.cancel) {
```
jasonnutter, line 154: FYI our lint rules ask for 4 spaces, instead of 2, which is why this diff looks so big.

samuelkubai: Fixed this and now the diff is as expected.

samuelkubai:
Fixed the failing test cases and linting errors and I think the pull request is ready for review by the team

lib/msal-common/src/client/DeviceCodeClient.ts
jasonnutter: Thanks!

lib/msal-common/src/client/DeviceCodeClient.ts
```diff
@@ -147,62 +147,69 @@ export class DeviceCodeClient extends BaseClient {
...
+            };
+            const response = await this.executePostToTokenEndpoint(
+                this.authority.tokenEndpoint,
+                requestBody,
+                headers,
+                thumbprint);
+
+            if (response.body && response.body.error === Constants.AUTHORIZATION_PENDING) {
+                // user authorization is pending. Sleep for polling interval and try again
+                this.logger.info(response.body.error_description || "no_error_description");
```
jasonnutter, line 199: I would prefer something more meaningful here. Also, errors should be logged using `this.logger.error`.

lib/msal-common/src/client/DeviceCodeClient.ts
```diff
@@ -147,62 +147,69 @@ export class DeviceCodeClient extends BaseClient {
...
+  
+                await delay(pollingIntervalMilli);
+            } else {
+                return response.body;
```
jasonnutter, line 203: Can we add a verbose log message here?

lib/msal-common/src/client/DeviceCodeClient.ts
```diff
@@ -147,62 +147,69 @@ export class DeviceCodeClient extends BaseClient {
...
+            }
+            return true;
+        };
+
+        /**
+         * Waits for t number of milliseconds
+         * @param t number
+         * @param value T
+         */
+        function delay<T>(t: number, value?: T): Promise<T | void> {
```
jasonnutter, line 177: Can this be defined somewhere more easily sharable?

jasonnutter: Also, this function and `continuePolling` are defined in different ways. I would propose they are both added as members of this class.

sameerag: +1 to making them class functions. `delay` can also be a utility function instantiated as needed.

lib/msal-common/src/client/DeviceCodeClient.ts
```diff
@@ -147,62 +147,69 @@ export class DeviceCodeClient extends BaseClient {
...
 
+        /**
+         * Breaks the polling with specific conditions.
+         * @param request CommonDeviceCodeRequest
+         * @param deviceCodeResponse DeviceCodeResponse
+         */
+        const continuePolling = (): boolean => {
+            if (request.cancel) {
+                this.logger.error("Token request cancelled by setting DeviceCodeRequest.cancel = true");
+                throw ClientAuthError.createDeviceCodeCancelledError();
```
sameerag, line 158: Do these errors throw and halt the application or are they supposed to bubble up in a promise as an error response to the client request?

sameerag: Never mind. We are doing `await delay()` below which should bubble this up.

lib/msal-common/src/client/DeviceCodeClient.ts
samuelkubai: Replaced the `"no_error_description"` with `"Authorization pending. Continue polling."`, the response description is still most meaningful here as it is phrased like `"AADSTS70016: OAuth 2.0 device flow error. Authorization is pending. Continue polling. Trace ID: 01707a0c-640b-4049-8cbb-ee2304dc0700 Correlation ID: 78b0fdfc-dd0e-4dfb-b13a-d316333783f6 Timestamp: 2020-03-26 22:54:14Z"` with a lot of useful information for debugging.

lib/msal-common/src/client/DeviceCodeClient.ts
samuelkubai: Moved the `delay` function to the `TimeUtils` utility class and made `continuePolling` a member of the class.

lib/msal-common/src/client/DeviceCodeClient.ts
samuelkubai: The message itself does not seem appropriate to categorize as an error message, I still think `this.logger.info` is the better categorization of the message.

lib/msal-common/src/client/DeviceCodeClient.ts
samuelkubai: Added.

sameerag (approved):

samuelkubai (approved):

PR #3558, opened 2021-04-29
Properly handle MSAL guard on redirect uri for hash and path routing

jasonnutter:
In #3506, this scenario was only handled for path routing. Furthermore, return `false` in the guard is not the correct behavior. Instead, it should return a url tree to the page without the hash response.

### DIFF
--- a/lib/msal-angular/src/msal.guard.spec.ts
+++ b/lib/msal-angular/src/msal.guard.spec.ts
@@ -26,7 +26,7 @@ function MSALInstanceFactory(): IPublicClientApplication {
     system: {
         loggerOptions: {
             loggerCallback: (level, message) => {
-                console.log(message)
+                // console.log(message)
             },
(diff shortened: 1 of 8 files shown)


lib/msal-angular/src/msal.guard.spec.ts
```diff
@@ -21,6 +22,15 @@ function MSALInstanceFactory(): IPublicClientApplication {
     auth: {
       clientId: '6226576d-37e9-49eb-b201-ec1eeb0029b6',
       redirectUri: 'http://localhost:4200'
+    },
+    system: {
+        loggerOptions: {
+            loggerCallback: (level, message) => {
+                console.log(message)
```
jasonnutter, line 29:
```suggestion
                // console.log(message)
```

jo-arroyo (approved):

tnorling (approved):

PR #3563, opened 2021-04-30
Preserve ES modules and enable treeshaking

tnorling:
This PR makes several changes to the rollup configs for `msal-common` and `msal-browser`:

- ES Module builds preserve the file structure to enable treeshaking - Thanks @thoo1!
- `msal-browser` does not bundle `msal-common` with the es module build
- Removes `msal-common` UMD bundle generation as we don't upload to CDN

Dependency updates:
- Updates `rollup` to version 2
- Replaces deprecated `rollup-plugin-node-resolve` package with `@rollup/plugin-node-resolve`
- Removes `rollup-plugin-json` as it's no longer needed

This is a first step towards reducing bundle size. Next step is to refactor some of our code to make it possible to be treeshaked.


lib/msal-browser/package.json
```diff
@@ -31,10 +32,7 @@
     "test": "test"
   },
   "files": [
-    "dist",
-    "lib/msal-browser.js",
-    "lib/msal-browser.js.map",
-    "lib/msal-browser.min.js"
```
jasonnutter, line 37: I added the cdn builds to npm so that if folks want the cdn builds, they can get them from npm. It shouldn't impact folks consuming non-cdn builds.

tnorling: Gotcha, I'll put this back then.

jasonnutter (approved):

pkanher617 (approved):

PR #3571, opened 2021-05-03
Reduce logs in BrowserCacheManager

jasonnutter:
These logs are very noisy and not really useful, unfortunately.

### DIFF
--- a/change/@azure-msal-browser-4e96ce29-ab2e-4b7a-bf40-125708bcb208.json
+++ b/change/@azure-msal-browser-4e96ce29-ab2e-4b7a-bf40-125708bcb208.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Reduce logs in BrowserCacheManager",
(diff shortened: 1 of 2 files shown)


change/@azure-msal-browser-4e96ce29-ab2e-4b7a-bf40-125708bcb208.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Reduce logs in BrowserCacheManager",
+  "packageName": "@azure/msal-browser",
+  "email": "janutter@microsoft.com",
+  "dependentChangeType": "patch"
```
jasonnutter, line 6:
```suggestion
  "dependentChangeType": "none"
```

tnorling (approved):

sameerag (approved):
Agree. I still think we need bring back `debug` in case someone wants to check the exact read/write or relevant data read/written. That can wait though. Approving.

PR #3576, opened 2021-05-04
[msal-node-extensions] Add the verification of persistence

samuelkubai:
Add the verifyPersistence method to each persistence implementation to enable our users to verify that the underlying persistence mechanisms are available for their environment.

### DIFF
--- a/extensions/msal-node-extensions/src/persistence/BasePersistence.ts
+++ b/extensions/msal-node-extensions/src/persistence/BasePersistence.ts
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { PersistenceError } from "../error/PersistenceError";
+import { Constants } from "../utils/Constants";
+import { IPersistence } from "./IPersistence";
+
+export abstract class BasePersistence {
+    public abstract createForPersistenceValidation(): Promise<IPersistence>; 
+
+    public async verifyPersistence(): Promise<boolean> {
+        // We are using a different location for the test to avoid overriding the functional cache
+        const persistenceValidator = await this.createForPersistenceValidation();
+
+        try {
+            persistenceValidator.save(Constants.PERSISTENCE_TEST_DATA);
+
+            const retrievedDummyData = await persistenceValidator.load();
+
+            if (!retrievedDummyData) {
+                throw PersistenceError.createCachePersistenceError(
+                    "Persistence check failed. Data was written but it could not be read. " +
+                    "Possible cause: on Linux, LibSecret is installed but D-Bus isn't running because it cannot be started over SSH."
+                );
+            }
+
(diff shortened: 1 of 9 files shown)


extensions/samples/msal-node-extensions/index.js
```diff
@@ -18,11 +18,16 @@ createPersistence().then((filePersistence) => {
...
             clientId: "99cab759-2aab-420b-91d8-5e3d8d4f063b",
             authority: "https://login.microsoftonline.com/90b8faa8-cc95-460e-a618-ee770bee1759",
-        },
-        cache: {
-            cachePlugin: new extensions.PersistenceCachePlugin(filePersistence)
-        },
+        }
     };
+
+    if (filePersistence.verifyPersistence()) {
```
sadasant, line 24: I believe this needs to be awaited!

samuelkubai: Yes, you are right, I have changed it.

sadasant (approved):
Looks good to me!

extensions/msal-node-extensions/src/persistence/BasePersistence.ts
```diff
@@ -0,0 +1,42 @@
...
+        // We are using a different location for the test to avoid overriding the functional cache
+        const persistenceValidator = await this.createForPersistenceValidation();
+
+        try {
+            persistenceValidator.save(Constants.PERSISTENCE_TEST_DATA);
+
+            const retrievedDummyData = await persistenceValidator.load();
+
+            if (!retrievedDummyData) {
+                throw PersistenceError.createCachePersistenceError(
```
sameerag, line 23: Is it possible in other platforms? And can we also compare this with a design that can throw `platform` specific errors instead?

samuelkubai: This cases are derived from the dotnet node-extensions, from my own tests, I have not been able to derive more cases than covered here.

sameerag: Well, approving for now. Let us work on a better solution eventually to segregate platform errors.

sameerag (approved):

hectormmg (approved):

lib/msal-common/src/authority/Authority.ts
```diff
@@ -464,7 +464,11 @@ export class Authority {
      * @returns bool
      */
     static isPublicCloudAuthority(host: string): boolean {
-        return Constants.KNOWN_PUBLIC_CLOUDS.includes(host);
+        for (let index = 0; index < Constants.KNOWN_PUBLIC_CLOUDS.length; index++) {
```
tnorling, line 467: Can you do this instead?
```suggestion
        return Constants.KNOWN_PUBLIC_CLOUDS.indexOf(host) >= 0;
```

tnorling (approved):

PR #3579, opened 2021-05-04
Add check for empty object string in claims request parameter

pkanher617:
This PR checks whether the object passed to the claims parameter is not an empty object. This prevents unnecessary cache skips when the empty object is provided, as we are only required to skip if additional claims are requested.


lib/msal-common/src/utils/Constants.ts
```diff
@@ -38,7 +38,8 @@ export const Constants = {
     AUTHORIZATION_PENDING: "authorization_pending",
     NOT_DEFINED: "not_defined",
     EMPTY_STRING: "",
-    FORWARD_SLASH: "/"
+    FORWARD_SLASH: "/",
+    EMPTY_OBJECT_STRING: "{}"
```
tnorling, line 42: What if someone passes "{ }"? Would it be better to do regex matching? Or at least remove whitespace before comparison?

sameerag: +1 to regex matching

lib/msal-common/test/account/AuthToken.spec.ts
```diff
@@ -77,8 +77,9 @@ describe("AuthToken.ts Class Unit Tests", () => {
         it("Throws error if rawIdToken is null or empty", () => {
             expect(() => new AuthToken("", cryptoInterface)).to.throw(ClientAuthErrorMessage.nullOrEmptyToken.desc);
             expect(() => new AuthToken("", cryptoInterface)).to.throw(ClientAuthError);
-
+            // @ts-ignore
```
sameerag, line 80: exception to function params being non-nullable?

sameerag: Can we just add an exception to tests?

pkanher617: this is more to test vanilla JS scenarios. Typescript will catch them regardless. We are making some changes to our tests, one of them may be to test vanilla JS and typescript separately.

lib/msal-common/src/utils/Constants.ts
pkanher617: Good point, I will make some changes.

lib/msal-common/test/account/AuthToken.spec.ts
pkanher617: > Can we just add an exception to tests?

Not sure what you mean.

sameerag: Allow null in function params for test files in lint rules.

lib/msal-common/src/utils/Constants.ts
jasonnutter: Could consider using `JSON.parse` here also.

jasonnutter: e.g. `Object.keys(JSON.parse(" {  } ")).length`

tnorling (approved):

lib/msal-common/src/utils/Constants.ts
pkanher617: done.

lib/msal-common/src/utils/StringUtils.ts
```diff
@@ -42,6 +42,20 @@ export class StringUtils {
...
+     * @param strObj 
+     */
+    static isEmptyObj(strObj?: string): boolean {
+        if (strObj && !StringUtils.isEmpty(strObj)) {
+            try {
+                const obj = JSON.parse(strObj);
+                return Object.keys(obj).length === 0;
+            } catch (e) {}
+        }
+        return true;
```
jasonnutter, line 56: Do we want to return true or false here? Seems like `false` should be teh default?

jasonnutter (approved):

PR #3587, opened 2021-05-05
[msal-core] Fix ID Token in browser history bug

### LINKED ISSUE #3367: Authentication redirect exposes Id token
## Library

- [x] `@azure/msal-angular@1.x.x`

## Description

I'm using the Angular MSAL library and Azure AD implicit flow to authorize users. 

After signing in through the Microsoft page, I first get a redirect url that exposes the id token and other response information in it before being redirected to the provided redirect uri. This url with the token hash is then saved in the browser history, which is an obvious security risk.

So far I had no luck finding a solution to the issue anywhere. 

## Source

- [ ] Internal (Microsoft)
- [x] Customer request

hectormmg:
This PR fixes a bug in msal v1 where the ID Token portion of the hash was being left in the browser history entry after authentication (for both reidrect and popup). Fixes #3367 .


lib/msal-core/src/utils/WindowUtils.ts
```diff
@@ -345,6 +348,7 @@ export class WindowUtils {
         // Office.js sets history.replaceState to null
         if (typeof history.replaceState === "function") {
             // Full removes "#" from url
+            window.location.replace(`${window.location.pathname}${window.location.search}`);
```
jasonnutter, line 351: I think this could be done before the `if` (i.e. for both when history.replaceState is available or not)

lib/msal-core/src/utils/WindowUtils.ts
```diff
@@ -139,7 +139,10 @@ export class WindowUtils {
                 if (href && UrlUtils.urlContainsHash(href)) {
                     logger.verbose("monitorPopupForHash found url in hash");
                     clearInterval(intervalId);
-                    resolve(contentWindow.location.hash);
+                    const hash = contentWindow.location.hash;
+                    contentWindow.location.hash = "";
+                    contentWindow.history.replaceState(null, null, `${contentWindow.location.pathname}${contentWindow.location.search}`);
```
jasonnutter, line 144: As with redirect, you should check if `history.replaceState` is a function first.

jasonnutter: Seems like this logic could be factored out.

lib/msal-core/src/utils/WindowUtils.ts
```diff
@@ -345,6 +348,7 @@ export class WindowUtils {
         // Office.js sets history.replaceState to null
```
jasonnutter, line 348: We could make this function generic and take in a window object, so that it could be used for popups and redirects

lib/msal-core/src/utils/WindowUtils.ts
tnorling, line 144: Are these 2 lines not redundant?

lib/msal-core/src/utils/WindowUtils.ts
tnorling: Is `history.replaceState` still needed at all if we use `window.location.replace`?

hectormmg: Yes, I tried multiple approaches and this is the only one that works. `window.location.replace` sets the url in the browser, but by then the previous URL is already in the history object, so the replaceState line is also necessary to remove it.

hectormmg: So, after a lot of fiddling, `history.replaceState` will work, but the hash has to be set to an empty string beforehand so the id_token is properly cleared from the history entry. With the latest changes, both redirect and popup use the same `clearUrlFragment` method and they're both leaving a single history entry without the id_token nor the octothorpe at the end of the URL.

lib/msal-core/src/utils/WindowUtils.ts
hectormmg: Done!

lib/msal-core/src/utils/WindowUtils.ts
```diff
@@ -341,13 +343,14 @@ export class WindowUtils {
...
+    static clearUrlFragment(contentWindow: Window): void {
         // Office.js sets history.replaceState to null
         if (typeof history.replaceState === "function") {
             // Full removes "#" from url
-            history.replaceState(null, null, `${window.location.pathname}${window.location.search}`);
+            contentWindow.location.hash = "";
+            contentWindow.history.replaceState(null, null, `${contentWindow.location.pathname}${contentWindow.location.search}`);
         } else {
-            window.location.hash = "";
+            contentWindow.location.hash = "";
```
jasonnutter, line 353: It looks like this line can always be run (before the if), then only thing run in the if is `contentWindow.history.replaceState`.

hectormmg: Right, thanks!

lib/msal-core/src/utils/WindowUtils.ts
```diff
@@ -341,13 +343,14 @@ export class WindowUtils {
     /**
      * Removes url fragment from browser url
      */
-    static clearUrlFragment(): void {
+    static clearUrlFragment(contentWindow: Window): void {
         // Office.js sets history.replaceState to null
         if (typeof history.replaceState === "function") {
```
jasonnutter, line 349:
```suggestion
        if (typeof contentWindow.history.replaceState === "function") {
```

jasonnutter (approved):
Looks good! Can you look and see if we need to make similar fixes in v2 (can be done in a separate PR).

PR #3594, opened 2021-05-06
Express.js sample - Fixes for local development and Azure app service deployment - DO NOT MERGE

diberry:


derisen (approved):
LGTM! one comment/suggestion

diberry:
Closing and leaving as is - since it is a good example of when to not use a straight forward Azure app service and instead use a container or a CICD pipeline. I'll use it as is in the tutorial - just a different tutorial.

Issue #3022, opened 2021-02-12
Could ILoggerCallback's message include correlationId if correlationId was provided to the request?


thoo1:
## Library

- [ ] `msal@1.x.x` or `@azure/msal@1.x.x`
- [X] `@azure/msal-browser@2.11.0`
- [ ] `@azure/msal-node@1.x.x`
- [ ] `@azure/msal-react@1.x.x`
- [ ] `@azure/msal-angular@0.x.x`
- [ ] `@azure/msal-angular@1.x.x`
- [ ] `@azure/msal-angular@2.x.x`
- [ ] `@azure/msal-angularjs@1.x.x`

## Description
`BaseAuthRequest` accepts a correlationId, which is useful in debugging logs in AAD. Could the logs that are provided in the `ILoggerCallback`, include this correlationId as well? That way we can see all logs for one request neatly

## Source

- [X] Internal (Microsoft)
- [ ] Customer request

thoo1:
Hi, wondering if there was any traction on this task? The reason I ask is because our app makes parallel calls to acquireTokenSilent and it is hard to figure out which log from msal corresponds to which call to acquireTokenSilent (the logs dont have correlationId)

tnorling:
@thoo1 We haven't had a chance to look into this yet but PRs are always welcome if you'd like to take a stab at it yourself.

PR #3601, opened 2021-05-06
Add correlationId to log messages

jo-arroyo:
This PR adds the correlation ID to log messages to make it easier to track log messages associated with requests. The correlation ID is either passed in or created per request, and will be added to log messages related to the request and passed down to flow client to use on the client's logger instance. Addresses #3022.

#### Msal-common
- Adjusts `Logger` to allow setting correlation ID on the logger instance from the `loggerOptions`
- Change `Logger.logMessage()` to set correlation ID if present. Precedence goes to the correlation ID passed in the log message, then to the correlation ID set on the Logger instance.
- Add unit tests for `Logger` correlation ID changes

#### Msal-browser
- Adds the correlation ID to msal-browser log messages related to requests
- Each request now creates a `browserRequestLogger`, with the correlation ID set when instantiated. This logger instance is used for all subsequent logs related to the request, and is passed into called functions and interaction handlers where applicable. 
  - Note that where the `browserRequestLogger` is not available, the correlation ID, if available, may still be passed for log messages.
  - This is a stop-gap feature, and may be replace when the `Logger` is redesigned. 
- Adds the correlation ID to `getClientConfiguration`, so correlation ID can be set on flow client logger instance
- Adjusts interaction handler unit tests to account for new `browserRequestLogger`

#### Msal-node
- Adds the correlation ID to msal-node log messages related to requests
- Adds the correlation ID to `buildOauthClientConfiguration`, so correlation ID can be set on flow client logger instance
- Changes cache and storage log messages to logLevel `trace` to reduce noisiness of `verbose` level log messages

#### Screenshot examples of new logs:

Screenshot 2021-06-02 125735

Screenshot 2021-06-02 125840

Note: We realize that we are knowingly taking on tech debt by passing in the correlation ID manually into log messages. In the future, we would like to refactor usage of the Logger so this will no longer be necessary.


jasonnutter:
Having the correlation id provided on the individual log messages (instead of being set on the logger itself) may be a better approach, to better support multiple concurrent requests.

sameerag:
> Having the correlation id provided on the individual log messages (instead of being set on the logger itself) may be a better approach, to better support multiple concurrent requests.

I was discussing this with @jo-arroyo! Since every request is a separate instance, will there be a case where parallel requests point to the same logger instance and mess up the correlation-id referenced? If so, we need to track correlation-id separately per request and add it to the logs as an extra parameter.  Better yet, have a request logger instance derived from the main class. My guess is this depends on the design of parallel requests.

We can discuss this @jo-arroyo @jasonnutter

jasonnutter:
> Since every request is a separate instance, will there be a case where parallel requests point to the same logger instance and mess up the correlation-id referenced

Not every request is guaranteed to have its own logger instance.

tnorling:
Did you consider generating a new logger instance per request (instead of using a single class property) and passing that down? It sounds like we can't avoid passing *something* down to functions if we want logging, I wonder if passing the logger instance would be better than passing the correlationId and then also needing to provide name and version on each message. Thoughts?

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -339,7 +339,8 @@ export abstract class ClientApplication {
                 throw BrowserAuthError.createNoCachedAuthorityError();
             }
 
-            const authClient = await this.createAuthCodeClient(serverTelemetryManager, currentAuthority);
+            const authClient = await this.createAuthCodeClient(serverTelemetryManager, currentAuthority, cachedRequest.correlationId);
+            authClient.logger.verbose("Auth code client created", "", name, version);
```
tnorling, line 343: It's not entirely clear to me why we need to use the authClient logger here instead of msal-browsers logger? Wouldn't it be cleaner to just call `authClient.logger.verbose("Auth code client created", cachedRequest.correlationId);` and not have to add name and version to the log messages themselves?

jasonnutter:
> Did you consider generating a new logger instance per request (instead of using a single class property) and passing that down? It sounds like we can't avoid passing _something_ down to functions if we want logging, I wonder if passing the logger instance would be better than passing the correlationId and then also needing to provide name and version on each message. Thoughts?

In the future (i.e. v3), I would like to have a model where when a request is made, the first thing that happens is the library creates a flow client instance (e.g. SilentFlowClient, RedirectFlowClient), and then everything downstream happens in that client, including log messages (as today each flow client includes its own logger). For example, today for a silent request, some things happen in the `SilentFlowClient`, but there are many things that do not.

tnorling:
> In the future (i.e. v3), I would like to have a model where when a request is made, the first thing that happens is the library creates a client instance (e.g. SilentFlowClient, AuthCodeClient), and then everything downstream happens in that client.

I'm actually working on a prototype for something very similar to this as we speak...FWIW I don't think it necessarily needs to wait until v3, but we'll see

> If everything for a given request happened in its client, it would be easy to pass the correlation id for that request into the client and attach it to the logger, and all log messages related to that request use the client's logger.

Agreed, however, everything does happen downstream from the API that was called. We could instantiate a new logger instance each time an API is called and pass this down to whatever functions need to use the logger (which we're already doing in many cases today). I'm not personally a fan of passing the correlationId down everywhere because once you're outside PCA you'll need to provide this *in addition to* the logger. Is there something I'm overlooking that would not make this feasible today?

lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -101,12 +101,13 @@ export abstract class ClientApplication {
...
         const authClientConfig = await this.buildOauthClientConfiguration(
-            validRequest.authority
+            validRequest.authority,
+            validRequest.correlationId
         );
-        this.logger.verbose("Auth client config generated");
         const authorizationCodeClient = new AuthorizationCodeClient(
             authClientConfig
         );
+        this.logger.verbose("Auth code client created", validRequest.correlationId);
```
jo-arroyo, line 110: As there are not many logs in `msal-node` that use the flow client logger, I've decided to use the global logger for now with the updated correlation ID instead of instantiating a new `nodeRequestLogger` instance for each request. This can, of course, be changed in future PRs as the `Logger` changes.

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -328,8 +328,9 @@ export abstract class ClientApplication {
      * @param state
      */
     private async handleHash(hash: string, state: string): Promise<AuthenticationResult> {
-        this.logger.verbose("handleHash called");
         const cachedRequest = this.browserStorage.getCachedRequest(state, this.browserCrypto);
+        const browserRequestLogger = new Logger({...this.config.system.loggerOptions, correlationId: cachedRequest.correlationId}, name, version);
```
tnorling, line 332: In order to prevent breaking changes to people calling the `setLogger` method we should probably call `this.logger.clone` with the new correlationId, etc. here

tnorling (commented):
Just one last comment and then I think this looks great!

lib/msal-browser/src/app/ClientApplication.ts
jo-arroyo: Makes sense. Updated.

lib/msal-common/src/logger/Logger.ts
```diff
@@ -71,8 +72,8 @@ export class Logger {
     /**
      * Create new Logger with existing configurations.
      */
-    public clone(packageName: string, packageVersion: string): Logger {
-        return new Logger({loggerCallback: this.localCallback, piiLoggingEnabled: this.piiLoggingEnabled, logLevel: this.level}, packageName, packageVersion);
+    public clone(packageName: string, packageVersion: string, correlationId?: string): Logger {
+        return new Logger({loggerCallback: this.localCallback, piiLoggingEnabled: this.piiLoggingEnabled, logLevel: this.level, correlationId: this.correlationId}, packageName, packageVersion);
```
jasonnutter, line 76:
```suggestion
        return new Logger({loggerCallback: this.localCallback, piiLoggingEnabled: this.piiLoggingEnabled, logLevel: this.level, correlationId: correlationId || this.correlationId}, packageName, packageVersion);
```

I think?

jo-arroyo: Yes, thanks for catching that.

tnorling (approved):
Looks good! Thanks for addressing feedback

sameerag (approved):
Lgtm. Can you please log a ADO item for node logger enhancements? When we add `debug` I want to address this.

PR #3623, opened 2021-05-12
fix cache expiry comparison

tripodsan:
`isAfter()` takes a date not a string.


AyaanRanosys:
@tripodsan 

> `isAfter()` takes a data not a string.

I think you mean "date" here. Right?

tripodsan:
> @tripodsan
> 
> > `isAfter()` takes a data not a string.
> 
> I think you mean "date" here. Right?

yes... I'm updating the description

jasonnutter:
Closing due to inactivity. We highly encourage ADAL Node applications to update to MSAL Node. If you would still like this PR merged, please include tests.

PR #3651, opened 2021-05-18
[msal-node] Fix log level propagation to msal-common

samuelkubai:
Allow the log level provided to msal-node to be propagated to msal-common


sameerag (approved):
Can we add tests? Do we test log levels or mock them everywhere? Just curious. Approved.

hectormmg (approved):

PR #3669, opened 2021-05-21
Switch to Codecov for test coverage reporting

tnorling:
Switches code coverage reporting to [Codecov](https://docs.codecov.io/docs). This allows us to upload code coverage results per library and will "carry forward" results for libraries that were unaffected by the PR. This also allows each library's README to display its own code coverage badge instead of the total for all our libraries (will update later)

PR comment will look something like this, showing each library's total coverage and the coverage changes for the specific files touched by the PR. (This comment is configurable if we want more or less information)

image

Also updates the CI workflow to only build and test libraries that were actually affected by the PR (e.g. changes to msal-common runs CI for common, browser, node, react and angular but changes to msal-core only runs CI for msal-core)

Coverage step can also be run on PRs from forks, unlike Coveralls

### DIFF
--- a/.github/workflows/build-steps.yml
+++ b/.github/workflows/build-steps.yml
@@ -19,22 +19,50 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  pre-check:
+    runs-on: ubuntu-latest
+    outputs: 
+      # Expose matched filters as job 'packages' output variable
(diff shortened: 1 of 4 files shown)


jasonnutter (approved):

jo-arroyo (approved):

sameerag:
Late to the party, love this @tnorling :)

PR #3716, opened 2021-06-04
Add docs and samples for front-channel logout

jasonnutter:
Add docs and samples for how to implement single-sign out via front-channel logout.

### DIFF
--- a/lib/msal-browser/docs/logout.md
+++ b/lib/msal-browser/docs/logout.md
@@ -114,15 +114,23 @@ const msal = new PublicClientApplication({
 })
 
 // Automatically on page load
-msal.logout({
+msal.logoutRedirect({
     onRedirectNavigate: () => {
-        // Return false if you would like to stop navigation after local logout
+        // Return false to stop navigation after local logout
         return false;
     }
 });
 ```
 
 Now when a user logouts out of another application, your application's front-channel logout url will be loaded in a hidden iframe, and MSAL.js will clear its cache to complete single-sign out.
+
+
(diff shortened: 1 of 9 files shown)


lib/msal-browser/docs/logout.md
```diff
@@ -86,6 +86,44 @@ await msalInstance.logoutPopup({
...
+    },
+    system: {
+        allowRedirectInIframe: true
+    }
+})
+
+// Automatically on page load
+msal.logout({
+    onRedirectNavigate: () => {
+        // Return false if you would like to stop navigation after local logout
```
jasonnutter, line 119:
```suggestion
        // Return false to stop navigation after local logout
```

tnorling (approved):

lib/msal-browser/docs/logout.md
```diff
@@ -86,6 +86,52 @@ await msalInstance.logoutPopup({
 });
 ```
 
+## Front-channel logout
```
pkanher617, line 89: Can we add a TOC to this file? There are quite a few sections and some things might get lost if we don't have a reference to them at the top of the file.

pkanher617 (approved):
small comment about the docs, otherwise looks good

lib/msal-browser/docs/logout.md
jasonnutter: TOC are now added by Github automatically to all md files: https://github.blog/changelog/2021-04-13-table-of-contents-support-in-markdown-files/

PR #3720, opened 2021-06-04
Fix node tests

tnorling:
The "Keep me signed in" page seems to have changed and a selector we were previously waiting for is no longer present. This PR addresses this by changing the selector we wait for.


hectormmg (approved):

sameerag (approved):
Approving. Can we talk to the UI team once to make sure if we need to look for anything else for KMSI?

PR #3729, opened 2021-06-09
MsalInterceptor relative url fix

### LINKED ISSUE #3682: MSAL Interceptor matches wrong endpoint configuration on request
### Core Library

### Core Library Version

2.14.2

### Wrapper Library

### Wrapper Library Version

2.0.0

### Description

Our Angular application does HTTP calls to two seperate backend services (A and B). Both of these backend services require a scope for every API call to that service. In the protected resource map this is configured as followed:

```

protectedResourceMap.set(
 'https://service-a-api.azurewebsites.net/api/*', [
 'api://service-a/user_impersonation',
]);

protectedResourceMap.set(
 'https://service-b-api.azurewebsites.net/api/*',
 ['api://service-b/user_impersonation']
);

protectedResourceMap.set(
 'api/*', 
 ['api://service-c/user_impersonation']
);

return protectedResourceMap;
```

We used this same configuration in msal-angular 1.1.2 without issues, however while updating to 2.0.0 we noticed that HTTP calls to Service A have the scope of Service B attached to them, resulting in a 401.

When debugging the MSAL interceptor we found that the wrong endpoint is returned in the function matchResourcesToEndpoint in the following else statement

```
else {                
  // Matching endpoint with both protected resource and relative url of protected resource                
  return StringUtils.matchPattern(normalizedKey, endpoint) || StringUtils.matchPattern(relativeNormalizedKey, endpoint);            }
```
The first check `StringUtils.matchPattern(normalizedKey, endpoint)` fails because everything before the wildcard (https://service-a-api.azurewebsites.net) is checked, but then in the second check `StringUtils.matchPattern(relativeNormalizedKey, endpoint)` the wrong endpoint is matched because the relativeNormalizedKey (/api/*) matches the endpoints of both backend services.

### Error Message

_No response_

### Msal Logs

_No response_

### MSAL Configuration

```javascript
export function MSALInstanceFactory(configService: ConfigurationService): IPublicClientApplication {
    const config = configService.getConfig();
        auth: {
            clientId: config.clientId,
            authority: `https://login.microsoftonline.com/${TENANT_ID}`,
            redirectUri: window.origin,
            postLogoutRedirectUri: window.origin,
        },
        cache: {
        },
    });
}

protectedResourceMap.set(
 'https://service-a-api.azurewebsites.net/api/*', [
 'api://service-a/user_impersonation',
]);

protectedResourceMap.set(
 'https://service-b-api.azurewebsites.net/api/*',
 ['api://service-b/user_impersonation']
);

protectedResourceMap.set(
 'api/*', 
 ['api://service-c/user_impersonation']
);

export function MSALInterceptorConfigFactory(configService: ConfigurationService): MsalInterceptorConfiguration {submodules,
    return {
        interactionType: InteractionType.Redirect,
        protectedResourceMap: protectedResourceMap ,
    };
}
```

### Relevant Code Snippets

```javascript
// Protected resource map

protectedResourceMap.set(
 'https://service-a-api.azurewebsites.net/api/*', [
 'api://service-a/user_impersonation',
]);

protectedResourceMap.set(
 'https://service-b-api.azurewebsites.net/api/*',
 ['api://service-b/user_impersonation']
);

protectedResourceMap.set(
 'api/*', 
 ['api://service-c/user_impersonation']
);

return protectedResourceMap;



// Interceptor function that returns wrong scope for endpoint
    private matchResourcesToEndpoint(protectedResourcesEndpoints: string[], endpoint: string): Array<string> {
        return protectedResourcesEndpoints.filter(key => {
            const normalizedKey = this.location.normalize(key);
            
            // Normalized key should include query strings if applicable
            const keyComponents = new UrlString(key).getUrlComponents();
            const relativeNormalizedKey = keyComponents.QueryString ? `${keyComponents.AbsolutePath}?${keyComponents.QueryString}` : this.location.normalize(keyComponents.AbsolutePath);

            // Relative endpoint not applicable, matching endpoint with protected resource. StringUtils.matchPattern accounts for wildcards
            if (relativeNormalizedKey === "" || relativeNormalizedKey === "/*") {
                return StringUtils.matchPattern(normalizedKey, endpoint);
            } else {
                // Matching endpoint with both protected resource and relative url of protected resource
                return StringUtils.matchPattern(normalizedKey, endpoint) || StringUtils.matchPattern(relativeNormalizedKey, endpoint);
            }
        });
    }
```

### Reproduction Steps

1. Use the default MSAL interceptor
2. Configure the MSAL interceptor with the provider protectedResourceMap
3. Do a request to service A
4. Get a 401 error because the token contains the scope for service B

### Expected Behavior

When doing a request to https://service-a-api.azurewebsites.net/api/foo, the MSAL interceptor should not match this request to the endpoint https://service-b-api.azurewebsites.net/api/* and thus retrieve the wrong scopes.

### Identity Provider

Azure AD / MSA

Chrome

### Regression

@azure/msal-angular 1.1.2

### Source

External (Customer)

jo-arroyo:
This PR addresses issue #3682 by ensuring that the `MsalInterceptor` includes the domain when matching relative endpoints.


tnorling (approved):

lib/msal-angular/src/msal.interceptor.ts
```diff
@@ -116,36 +117,51 @@ export class MsalInterceptor implements HttpInterceptor {
...
+            }
             
+            // Transforms relative urls to absolute urls, gets components
+            const keyLink = document.createElement("a");
+            keyLink.href = key;
+            const keyLinkComponents = new UrlString(keyLink.href).getUrlComponents();
+
+            const endpointLink = document.createElement("a");
+            endpointLink.href = endpoint;
+            const endpointLinkComponents = new UrlString(endpointLink.href).getUrlComponents();
```
jasonnutter, line 153: Can you factor this logic into a helper function?

jo-arroyo: Sure. I had to leave the `new UrlString...` part out of the helper, since the absolute endpoint is needed later on.

lib/msal-angular/src/msal.interceptor.ts
```diff
@@ -116,36 +117,59 @@ export class MsalInterceptor implements HttpInterceptor {
...
+        return matchingResources;
+    }
+
+    /**
+     * Transforms relative urls to absolute urls
+     * @param url 
+     * @returns 
+     */
+    private getAbsoluteUrl(url: string): string {
+        const link = document.createElement("a");
```
jasonnutter, line 170: Forgot that to use the `document` object in Angular, we are suppose to inject it. Details here: https://stackoverflow.com/a/37521339/4888559

jasonnutter, line 170: After following the instructions above: 
```suggestion
        const link = this.document.createElement("a");
```

jo-arroyo: Ok makes sense. Had to make additional changes because of it could not resolve `Document`. I added this fix: https://stackoverflow.com/questions/49513359/could-not-resolve-type-document-in-angular5.

jasonnutter (approved):

PR #3739, opened 2021-06-10
Update redirect guidance in msal-react

tnorling:
Expands msal-react FAQ entry regarding the handling of redirects

### DIFF
--- a/lib/msal-react/FAQ.md
+++ b/lib/msal-react/FAQ.md
@@ -45,7 +45,34 @@ Yes, `@azure/msal-react` supports both function and class components. Hooks, how
 
 ### How do I handle the redirect flow in a react app?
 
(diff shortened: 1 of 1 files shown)


jo-arroyo (approved):

lib/msal-react/FAQ.md
```diff
@@ -72,7 +72,7 @@ If you need direct access to the response object or error returned by a redirect
 
 1. Use the [event API](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-react/docs/events.md) to register a callback that will be invoked with the response when you return to your app. Make sure this is registered in a code path that is run __after__ the redirect as any callbacks registered __before__ the redirect will be lost.
 1. Use the [useMsalAuthentication hook](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-react/docs/hooks.md#usemsalauthentication-hook) to login. This hook will return the result or error when you are returned to your app.
-1. Call `handleRedirectPromise` which will resolve with the result or reject with an error if the page was loaded as a result of a redirect
+1. Call `handleRedirectPromise` which will resolve with the result or reject with an error if the page was loaded as a result of a redirect.
```
sameerag, line 75: What happens if the page isn't reloaded? Just curious if we need to document this or if it ever happens.

sameerag (approved):
Approving. Minor comment.

lib/msal-react/FAQ.md
tnorling: What do you mean by "reloaded"? If the page is loaded after a redirect you'll get the result or error, if not (i.e. load up the page for the first time) you'll get null

sameerag: Yeah, I meant if it never redirects - `load up the page first time` use case. Thanks.

PR #3747, opened 2021-06-13
Include refresh token in AuthenticationResult

cwardcode:
Currently, when calling any of the `acquireToken` functions to initiate an authorization code flow, the ID and Access tokens are returned, but the refresh token is not. This PR is to include the refresh token as part of the AuthenticationResult type, so that consumers may use the refresh token outside of this library if needed.


tnorling:
@cwardcode Thanks for your contribution but unfortunately we cannot accept this. Not including the refresh token on the public API surface was a conscious design choice made by all MSALs and one we are unlikely to reverse, at least in the near future. If you have a specific issue you're trying to solve please open an [Issue](https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/new/choose) and we would be happy to help you find alternative solutions.

Issue #3741, opened 2021-06-10
MSAL ConfidentialClientApplication using SNI (x5c) throwing "AADSTS700027: Client assertion contains an invalid signature"


r-delgadillo:
### Core Library

MSAL Node (@azure/msal-node)

### Core Library Version

1.1.0

### Wrapper Library

Not Applicable

### Wrapper Library Version

None

### Description

I'm having trouble acquiring a first-party AAD token using msal-node using SNI. I followed the example in the [documentation](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/319600946963c68d1bd8f6f001f05c3637213cfc/lib/msal-node/docs/initialize-confidential-client-application.md) but I keep getting an unhelpful error message.

```
"invalid_client: 700027 - [2021-06-10 22:21:00Z]: AADSTS700027: Client assertion contains an invalid signature. [Reason - The key was not found., Thumbprint of key used by client: '3BAFFDE57ADC4ABB93374403C529DD84776E9B0F', Please visit the Azure Portal, Graph Explorer or directly use MS Graph to see configured keys for app Id '9edfcdd9-0bc5-4bd4-b287-c3afc716aac7'. Review the documentation at https://docs.microsoft.com/en-us/graph/deployments to determine the corresponding service endpoint and https://docs.microsoft.com/en-us/graph/api/application-get?view=graph-rest-1.0&tabs=http to build a query request URL, such as 'https://graph.microsoft-ppe.com/beta/applications/9edfcdd9-0bc5-4bd4-b287-c3afc716aac7']\r\nTrace ID: 60dec111-cbde-4d7b-8781-72e62a760500\r\nCorrelation ID: 6fe920bc-4850-4137-849e-7e4046b86753\r\nTimestamp: 2021-06-10 22:21:00Z - Correlation ID: 6fe920bc-4850-4137-849e-7e4046b86753 - Trace ID: 60dec111-cbde-4d7b-8781-72e62a760500"
```

I verified we are passing in a valid secret, thumbprint, and x5c certificate for the request but I continue to get the error. We attempted to get a token using ADAL and @azure/identity libraries and had no problem with it. I'm not sure what the next steps are for attempting to debug this issue. Below is a snippet of the code. 

### Relevant Code Snippets

```javascript
const resource = 'https://serviceidentity.azure-int.net/';
[...]

PR #3775, opened 2021-06-21
update certificate & SNI docs

derisen:
* Updated [SNI doc](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/docs/sni.md) with usage example and potential issues (#3741)
* Reorganized certificate documentation (move conversion section from `key-vault-managed-identity.md` to `certificate-credentials.md`)
* Minor cleaning with cert samples


samples/msal-node-samples/auth-code-key-vault/index.js
```diff
@@ -4,7 +4,6 @@
  */
 
 const express = require('express');
-const forge = require('node-forge');
```
sameerag, line 7: Are we referencing `node-forge` only in the docs now?

sameerag (approved):
One minor comment. Lgtm

samples/msal-node-samples/auth-code-key-vault/index.js
derisen: The code sample itself wasn't actually using it, but you remined me to add it to the snippet for pfx-pem conversion

tnorling (approved):

PR #3783, opened 2021-06-22
Add monitor_window_timeout error to the error doc

tnorling:
Adds `monitor_window_timeout` error to the error doc and adds a link to the error message

### DIFF
--- a/lib/msal-browser/docs/errors.md
+++ b/lib/msal-browser/docs/errors.md
@@ -153,6 +153,7 @@ This error can be thrown when calling `ssoSilent`, `acquireTokenSilent`, `acquir
 
 1. The page you use as your `redirectUri` is removing or manipulating the hash
 1. The page you use as your `redirectUri` is automatically navigating to a different page
+1. You are being throttled by your identity provider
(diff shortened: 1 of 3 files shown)


lib/msal-browser/docs/errors.md
```diff
@@ -142,6 +143,66 @@ Remember that you will need to register this new `redirectUri` on your App Regis
...
+        clientId: "your-client-id"
+    },
+    system: {
+        windowHashTimeout: 9000, // Applies just to popup calls - In milliseconds
+        iframeHashTimeout: 9000, // Applies just to silent calls - In milliseconds
+        loadFrameTimeout: 9000 // Applies to both silent and popup calls - In milliseconds
+    }
+};
+```
+
```
jasonnutter, line 205: For AAD, I think it there should be a section talking about throttling, as it is the most common cause of this error. I think we should also talk about mitigations, which can include:

1. Retrying the request and/or
2. Invoking interaction

tnorling: Done

jasonnutter (approved):

sameerag (approved):

PR #3787, opened 2021-06-23
[msal-node-extensions] verifyPersistence should not finally return true

sadasant:
Doing finally and returning true takes predominance over throwing the error, so the error doesn’t bubble up properly.

verifyPersistence() should return true only if no error was encountered.


samuelkubai (approved):
LGTM

sameerag (approved):

PR #3788, opened 2021-06-23
Update accounts FAQ

### LINKED ISSUE #3751: [PublicClientApplication.getAllAccounts() - (msal-browser)] Proper documentation not available describing multiple logged in accounts scenarios.
### Core Library

### Wrapper Library

Not Applicable

### Documentation Location

https://docs.microsoft.com/en-us/azure/active-directory/develop/scenario-spa-acquire-token?tabs=javascript2; https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-java-get-remove-accounts-token-cache

### Description

PublicClientApplication.getAllAccounts() - (msal-browser):
No Documentation to address the developer/customer questions like below -
- When can PublicClientApplication.getAllAccounts() return more than one account?
- How does this work with SSO? If the user logged in using different accounts on two other websites before coming to mine, would I get two accounts? Are they sorted in any order?
- If PublicClientApplication.getAllAccounts() returns some account/accounts then does it guarantee that user is logged in? 
- If I want to check authentication on startup do I need to force loginRedirect/loginPopup or I can first check that getAllAccounts() already has something? What is the best practice?
- Link to Sample/documentation that demonstrates how to switch between multiple logged in accounts.

Please validate the below information and add it in the documentation along with additional info-
 - MSAL does not make assumptions about what account to use. It will either use the account passed into the API or the active account that was previously set by calling setActiveAccount. Each call to loginPopup or loginRedirect will always result in exactly one account. getAllAccounts is not guaranteed to be sorted in any order, nor is it guaranteed to return the same order across calls.

tnorling:
Updates FAQ and Accounts.md with some frequently asked questions regarding `getAllAccounts`

Closes #3751

### DIFF
--- a/lib/msal-browser/FAQ.md
+++ b/lib/msal-browser/FAQ.md
@@ -21,6 +21,12 @@
 1. [How to get single sign-on in my application with MSAL.js?](#how-to-get-single-sign-on-in-my-application-with-msaljs)
 1. [How can my application recognize a user after sign-in? How do I correlate users between applications?](#how-can-my-application-recognize-a-user-after-sign-in-how-do-i-correlate-users-between-applications)
 
+**[Accounts](#Accounts)**
+
+1. [In what scenarios will getAllAccounts return multiple accounts?](#in-what-scenarios-will-getallaccounts-return-multiple-accounts)
+1. [Is the result of getAllAccounts sorted in any order?](#is-the-result-of-getallaccounts-sorted-in-any-order)
+1. [If an account is returned by getAllAccounts does that mean the user has an active session on the server?](#if-an-account-is-returned-by-getallaccounts-does-that-mean-the-user-has-an-active-session-on-the-server)
+
 **[Configuration](#Configuration)**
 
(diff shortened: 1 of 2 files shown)


lib/msal-browser/FAQ.md
```diff
@@ -149,6 +155,22 @@ loginPopup().then((response) => {
...
+
+If you would like to force the server account selection screen you can pass `prompt: "select_account"` or `prompt: "login"` to the `acquireToken` or `login` API.
+
+## Is the result of `getAllAccounts` sorted in any order?
+
+No, accounts are not sorted nor are they guaranteed to maintain any particular order across multiple calls.
+
+## If an account is returned by `getAllAccounts` does that mean the user has an active session on the server?
+
+No, the account APIs reflect local account state only. If you need to ensure the user has an active session on the server you should call `acquireTokenSilent` and fallback to interaction if needed.
```
jasonnutter, line 172: `ssoSilent` would also work here.

jasonnutter (approved):

sameerag (approved):

PR #3796, opened 2021-06-25
Fix broken link in initialize-public-client-application.md

khomutoff:
Click on the 'PKCE sample' link on [initialize-public-client-application.md](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/docs/initialize-public-client-application.md) page returns a 404 status code. I found the correct link on [README.md](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/README.md) page https://github.com/AzureAD/microsoft-authentication-library-for-js/search?q=auth-code-pkce

### DIFF
--- a/lib/msal-node/docs/initialize-public-client-application.md
+++ b/lib/msal-node/docs/initialize-public-client-application.md
@@ -12,7 +12,7 @@ In this document:
 
 ## Initializing the PublicClientApplication object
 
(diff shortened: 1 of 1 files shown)


jasonnutter (approved):
Good catch, thanks!

PR #3826, opened 2021-07-02
Add support for MSR Crypto

jasonnutter:
Refactor browser crypto classes to make it more extensible, and add support for MSR Crypto polyfill, which will only be used if it is included in the page and if none of the other crypto interfaces are available (e.g. a spa running on a custom protocol such as for a hybrid mobile developement).

Leveraging MSR Crypto requires both enabling the new `system.cryptoOptions.useMsrCrypto` flag, and providing entropy via `system.cryptoOptions.entropy`.


peterzenz (approved):

sameerag (commented):
Couple generic comments:
* The interface class named as `ISubtleCrypto` doesn't sound right as one of the implementations is in house and has nothing to do with `SubtleCrypto` APIs. May be we can make it more generic?
* The names MsrCrypto and MsCrypto are confusing. For the latter do you mean in-house MSAL JS implementation (as in microsoft crypto?)? Can we at-least add comments/docs on differentiation, usage and context for each of these crypto libraries?

Everything else looks good.

lib/msal-browser/src/crypto/BrowserCrypto.ts
```diff
@@ -25,10 +29,17 @@ const PUBLIC_EXPONENT: Uint8Array = new Uint8Array([0x01, 0x00, 0x01]);
...
-        if (!(this.hasCryptoAPI())) {
-            throw BrowserAuthError.createCryptoNotAvailableError("Browser crypto or msCrypto object not available.");
+        if ("crypto" in window) {
+            this.subtleCrypto = new ModernBrowserCrypto();
+        } else if ("msrCrypto" in window) {
+            this.subtleCrypto = new MsrBrowserCrypto();
+        } else if ("msCrypto" in window) {
+            this.subtleCrypto = new MsBrowserCrypto();
+        } else {
+            throw BrowserAuthError.createCryptoNotAvailableError("Browser crypto, msCrypto, or msrCrypto objects not available.");
```
jasonnutter, line 42: TODO: add logging

jasonnutter:
> Couple generic comments:
> 
> * The interface class named as `ISubtleCrypto` doesn't sound right as one of the implementations is in house and has nothing to do with `SubtleCrypto` APIs. May be we can make it more generic?

The name was picked based on that we call `window.[crypto-interface].subtle.[method]`, and the method signatures are taken from the native platform APIs.

> * The names MsrCrypto and MsCrypto are confusing. For the latter do you mean in-house MSAL JS implementation (as in microsoft crypto?)? Can we at-least add comments/docs on differentiation, usage and context for each of these crypto libraries?
> 
> Everything else looks good.

Those names are based on the interfaces being `window.msCrypto` (IE11) and `window.msrCrypto` (Microsoft Research Crypto). And yeah, I'll what docs can be included.

tnorling:
Any tests you can add specifically for the new Crypto classes and selection logic?

lib/msal-browser/src/crypto/MsBrowserCrypto.ts
```diff
@@ -0,0 +1,94 @@
...
+    async exportKey(key: CryptoKey): Promise<JsonWebKey> {
+        return new Promise((resolve: Function, reject: Function) => {
+            const msExportKey = window["msCrypto"].subtle.exportKey(KEY_FORMAT_JWK, key);
+            msExportKey.addEventListener("complete", (e: { target: { result: ArrayBuffer; }; }) => {
+                const resultBuffer: ArrayBuffer = e.target.result;
+
+                const resultString = BrowserStringUtils.utf8ArrToString(new Uint8Array(resultBuffer))
+                    .replace(/\r/g, "")
+                    .replace(/\n/g, "")
+                    .replace(/\t/g, "")
```
Robbie-Microsoft: should these be combined into one expression? Additionally, should these be defined with "new RegExp()"?

jasonnutter: Potentially, but this is just copy paste (which has been working fine afaik), so I'd prefer not to change anything as part of this PR.

tnorling (approved):

jasonnutter:
Updated with sample app and considerations for seeding entropy needed for MSR Crypto.

jasonnutter:
Updated with sample app and considerations for seeding entropy needed for MSR Crypto.

tnorling (approved):

lib/msal-browser/src/crypto/BrowserCrypto.ts
```diff
@@ -28,36 +30,85 @@ const PUBLIC_EXPONENT: Uint8Array = new Uint8Array([0x01, 0x00, 0x01]);
...
+                this.logger.error("BrowserCrypto: Interface requires entropy but none provided.");
+                throw BrowserConfigurationAuthError.createEntropyNotProvided();
+            }
+
+            this.logger.verbose("BrowserCrypto: Entropy provided");
+            this.subtleCrypto.initPrng(this.cryptoOptions.entropy);
         }
 
-        this._keygenAlgorithmOptions = {
+        this.keygenAlgorithmOptions = {
```
peterzenz, line 75: Why the rename here?

peterzenz (approved):

lib/msal-browser/src/crypto/BrowserCrypto.ts
jasonnutter: Moving away from underscore prefixed properties, which we don't really do anywhere else.

PR #3834, opened 2021-07-06
[msal-node-extensions]: Add default cache location

samuelkubai:
Add a default cache location in the cases where the user of the library does not provide a cache location.

### DIFF
--- a/change/@azure-msal-node-extensions-832ab65b-4ac3-41b1-a9d3-ad58efdc6892.json
+++ b/change/@azure-msal-node-extensions-832ab65b-4ac3-41b1-a9d3-ad58efdc6892.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "feat: add a default cache location for msal-node-extensions",
+  "packageName": "@azure/msal-node-extensions",
+  "email": "samuelkamau@microsoft.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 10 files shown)


change/@azure-msal-node-extensions-832ab65b-4ac3-41b1-a9d3-ad58efdc6892.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
jasonnutter, line 2: Given extensions is still in alpha, this should be `prerelease`.
```suggestion
  "type": "prerelease",
```

extensions/msal-node-extensions/src/utils/Constants.ts
```diff
@@ -31,5 +31,32 @@ export const Constants = {
...
+    /**
+     * List of environment variables
+     */
+    ENVIRONMENT: {
+        HOME: "HOME",
+        LOGNAME: "LOGNAME",
+        USER: "USER",
+        LNAME: "LNAME",
+        USERNAME: "USERNAME",
+        PLATFORM: "platform",
```
sameerag, line 50: everything else in caps and only `platform` is different. Any particular reason?

sameerag: Nevermind. Is this how the system stores these in the `process.env()`?

sameerag (approved):
Lgtm

samuelkubai:
Closing this pull request after discussions on the correct way forward.

Issue #2884, opened 2021-01-15
Making multiple calls to acquireTokenSilent, may result in multiple requests to service POST /token, is there a way to use whichever request succeeded?


thoo1:
## Library

- [ ] `msal@1.x.x` or `@azure/msal@1.x.x`
- [X] `@azure/msal-browser@2.6.1`
- [ ] `@azure/msal-node@1.x.x`
- [ ] `@azure/msal-react@1.x.x`
- [ ] `@azure/msal-angular@0.x.x`
- [ ] `@azure/msal-angular@1.x.x`
- [ ] `@azure/msal-angular@2.x.x`
- [ ] `@azure/msal-angularjs@1.x.x`

## Description
Currently in our app, we make a call to `acquireTokenSilent` before any api request. We can make api requests in parallel which result in parallel calls to `acquireTokenSilent`. 
When needed, `acquireTokenSilent` will make network calls to POST /token to get a new token. Currently this results in multiple calls to POST /token.

In some of our logs, we see that one or two of them will fail with status code 0: Failed to fetch.
1. I'm not sure why status code 0 is happening for some requests, but other requests succeed
2. In the same batch of parallel requests to POST /token, if only 1 fails, does the library return the result of any previously successful request?
3. For example say request A and request B both start at the same time. Request A succeeds first, and then request B fails. Can the `acquireTokenSilent` method return the result of request A for both calls?

## Source

- [X] Internal (Microsoft)
- [ ] Customer request

jo-arroyo:
@thoo1 To confirm, do request A and request B have the same sets of scopes? 

MSAL 2.x doesn't have logic for sharing outgoing requests for silent calls, and makes distinct network requests. At the moment there is no way for both requests to return the result from one successful call if the requests are made in parallel.

thoo1:
@jo-arroyo yes, it would be for the same scope.
I guess my question is more a feature ask :D, wonder if it makes sense and could be done?

jo-arroyo:
@thoo1 That's a reasonable request and we will look into it.

JosephThomas:
Any ETA on this issue? I believe MSAL 1.x had support for this so I was surprised this regression wasn't mentioned anywhere.

jacknewberry:
If I understand your issue, I think I have solved this by saving the promise returned from getTokenSilently and returning this promise to the calling code rather than calling acquireTokenSilent again.

[...]

PR #3837, opened 2021-07-06
Msal-Browser supports parallel silent requests

jo-arroyo:
This adds a `silentRequest` Map to *PublicClientApplication.ts*, and assigns a unique thumbprint to every request made in `acquireTokenSilent`. These thumbprints are used to track requests made, and is designed to gracefully handle multiple concurrent `acquireTokenSilent` calls with the same request. 

This PR addresses #2884.

### DIFF
--- a/change/@azure-msal-browser-adabc718-b19b-4b98-a5b7-887465c75857.json
+++ b/change/@azure-msal-browser-adabc718-b19b-4b98-a5b7-887465c75857.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Msal-Browser supports parallel silent requests #3837",
+  "packageName": "@azure/msal-browser",
+  "email": "joarroyo@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/lib/msal-browser/src/app/PublicClientApplication.ts
+++ b/lib/msal-browser/src/app/PublicClientApplication.ts
@@ -3,7 +3,7 @@
  * Licensed under the MIT License.
  */
 
-import { AccountInfo, AuthenticationResult, CommonSilentFlowRequest, RequestThumbprint, ThrottlingUtils } from "@azure/msal-common";
+import { AccountInfo, AuthenticationResult, CommonSilentFlowRequest, RequestThumbprint } from "@azure/msal-common";
 import { Configuration } from "../config/Configuration";
 import { DEFAULT_REQUEST, ApiId, InteractionType } from "../utils/BrowserConstants";
 import { IPublicClientApplication } from "./IPublicClientApplication";
@@ -96,7 +96,7 @@ export class PublicClientApplication extends ClientApplication implements IPubli
             scopes: request.scopes,
             homeAccountIdentifier: account.homeAccountId
         };
-        const silentRequestKey = ThrottlingUtils.generateThrottlingStorageKey(thumbprint);
+        const silentRequestKey = JSON.stringify(thumbprint);
         let response = this.activeSilentTokenRequests.get(silentRequestKey);
         if (typeof response === "undefined") {
             this.logger.verbose("acquireTokenSilent called for the first time, storing active request", request.correlationId);
--- a/lib/msal-browser/src/app/PublicClientApplication.ts
+++ b/lib/msal-browser/src/app/PublicClientApplication.ts
@@ -21,7 +21,7 @@ import { version, name } from "../packageMetadata";
  */
 export class PublicClientApplication extends ClientApplication implements IPublicClientApplication {
 
-    // Redirect Response Object
+    // Active requests
     private activeSilentTokenRequests: Map<string, Promise<AuthenticationResult>>;
 
     /**
@@ -97,10 +97,10 @@ export class PublicClientApplication extends ClientApplication implements IPubli
             homeAccountIdentifier: account.homeAccountId
         };
         const silentRequestKey = JSON.stringify(thumbprint);
-        let response = this.activeSilentTokenRequests.get(silentRequestKey);
-        if (typeof response === "undefined") {
+        const cachedResponse = this.activeSilentTokenRequests.get(silentRequestKey);
+        if (typeof cachedResponse === "undefined") {
             this.logger.verbose("acquireTokenSilent called for the first time, storing active request", request.correlationId);
-            response = this.acquireTokenSilentAsync(request, account)
+            const response = this.acquireTokenSilentAsync(request, account)
                 .then((result) => {
                     this.activeSilentTokenRequests.delete(silentRequestKey);
                     return result;
@@ -110,10 +110,11 @@ export class PublicClientApplication extends ClientApplication implements IPubli
                     throw error;
                 });
             this.activeSilentTokenRequests.set(silentRequestKey, response);
+            return response;
         } else {
             this.logger.verbose("acquireTokenSilent has been called previously, returning the result from the first call", request.correlationId);
+            return cachedResponse;
         }
-        return response;
     }
 
     /**
--- a/lib/msal-browser/src/app/PublicClientApplication.ts
+++ b/lib/msal-browser/src/app/PublicClientApplication.ts
@@ -22,7 +22,7 @@ import { version, name } from "../packageMetadata";
 export class PublicClientApplication extends ClientApplication implements IPublicClientApplication {
 
     // Redirect Response Object
-    private silentRequest: Map<string, Promise<AuthenticationResult>>;
+    private activeSilentTokenRequests: Map<string, Promise<AuthenticationResult>>;
 
     /**
      * @constructor
@@ -48,7 +48,7 @@ export class PublicClientApplication extends ClientApplication implements IPubli
     constructor(configuration: Configuration) {
         super(configuration);
 
-        this.silentRequest = new Map();
+        this.activeSilentTokenRequests = new Map();
     }
 
     /**
@@ -97,18 +97,21 @@ export class PublicClientApplication extends ClientApplication implements IPubli
             homeAccountIdentifier: account.homeAccountId
         };
         const silentRequestKey = ThrottlingUtils.generateThrottlingStorageKey(thumbprint);
-        let response = this.silentRequest.get(silentRequestKey);
+        let response = this.activeSilentTokenRequests.get(silentRequestKey);
         if (typeof response === "undefined") {
-            response = this.acquireTokenSilentAsync(request, account).then((result) => {
-                this.silentRequest.delete(silentRequestKey);
-                return result;
-            }).catch((error) => {
-                this.silentRequest.delete(silentRequestKey);
-                throw error;
-            });
-            this.silentRequest.set(silentRequestKey, response);
+            this.logger.verbose("acquireTokenSilent called for the first time, storing active request", request.correlationId);
+            response = this.acquireTokenSilentAsync(request, account)
+                .then((result) => {
+                    this.activeSilentTokenRequests.delete(silentRequestKey);
+                    return result;
+                })
+                .catch((error) => {
+                    this.activeSilentTokenRequests.delete(silentRequestKey);
+                    throw error;
+                });
+            this.activeSilentTokenRequests.set(silentRequestKey, response);
         } else {
-            this.logger.verbose("acquireTokenSilent has been called previously, returning the result from the first call");
+            this.logger.verbose("acquireTokenSilent has been called previously, returning the result from the first call", request.correlationId);
         }
         return response;
     }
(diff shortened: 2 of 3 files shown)


lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -85,6 +90,36 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
         }
+        const thumbprint: RequestThumbprint = {
+            clientId: this.config.auth.clientId,
+            authority: request.authority || "",
+            scopes: request.scopes,
+            homeAccountIdentifier: account.homeAccountId
+        };
+        const silentRequestKey = ThrottlingUtils.generateThrottlingStorageKey(thumbprint);
+        let response = this.silentRequest.get(silentRequestKey);
+        if (typeof response === "undefined") {
```
jasonnutter, line 101: Can we add a logger statement here, similar to teh one below for the else?

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -21,6 +21,9 @@ import { version, name } from "../packageMetadata";
  */
 export class PublicClientApplication extends ClientApplication implements IPublicClientApplication {
 
+    // Redirect Response Object
+    private silentRequest: Map<string, Promise<AuthenticationResult>>;
```
jasonnutter, line 25: I would propose a more detailed name, e.g. `activeSilentTokenRequests` or something like that.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -85,6 +90,36 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
+        const silentRequestKey = ThrottlingUtils.generateThrottlingStorageKey(thumbprint);
+        let response = this.silentRequest.get(silentRequestKey);
+        if (typeof response === "undefined") {
+            response = this.acquireTokenSilentAsync(request, account).then((result) => {
+                this.silentRequest.delete(silentRequestKey);
+                return result;
+            }).catch((error) => {
+                this.silentRequest.delete(silentRequestKey);
+                throw error;
+            });
```
jasonnutter, line 108:
```suggestion
            response = this.acquireTokenSilentAsync(request, account)
                .then((result) => {
                    this.silentRequest.delete(silentRequestKey);
                    return result;
                })
                .catch((error) => {
                    this.silentRequest.delete(silentRequestKey);
                    throw error;
                });
```

Fixing up the spacing.

change/@azure-msal-browser-adabc718-b19b-4b98-a5b7-887465c75857.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
jasonnutter, line 2: I think this can be a patch, no api changes.

jasonnutter:
A few cosmetic things, but otherwise looks good! It would be great to have these same improvements for `ssoSilent` as well (which can be done in a separate PR).

change/@azure-msal-browser-adabc718-b19b-4b98-a5b7-887465c75857.json
sameerag: I still think it is a significant enough feature. Is the "minor" tag only for API changes? My understanding is "major" is needed for public API and other tags can be decided based on the feature impact??

lib/msal-browser/src/app/PublicClientApplication.ts
sameerag: +1 since we already have a `SilentRequest` type

change/@azure-msal-browser-adabc718-b19b-4b98-a5b7-887465c75857.json
tnorling: Probably doesn't matter for this PR since we're already going to do a minor version bump due to other PRs, but we should come to a consensus on this. I would personally consider this a minor feature rather than a patch

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -85,6 +90,36 @@ export class PublicClientApplication extends ClientApplication implements IPubli
         if (!account) {
             throw BrowserAuthError.createNoAccountError();
         }
+        const thumbprint: RequestThumbprint = {
+            clientId: this.config.auth.clientId,
+            authority: request.authority || "",
+            scopes: request.scopes,
+            homeAccountIdentifier: account.homeAccountId
+        };
+        const silentRequestKey = ThrottlingUtils.generateThrottlingStorageKey(thumbprint);
```
tnorling, line 99: Can we rename and/or move this function if this is going to start being used for things other than throttling?

jo-arroyo: @tnorling Would changing the name to `generateRequestStorageKey` be enough, or should it also be moved out of `ThrottlingUtils`? Aside from here, the only other place it is used is in `ThrottlingUtils`.

lib/msal-browser/src/app/PublicClientApplication.ts
jo-arroyo: Done

lib/msal-browser/src/app/PublicClientApplication.ts
jo-arroyo: Makes sense. Changed.

lib/msal-browser/src/app/PublicClientApplication.ts
tnorling: `generateThrottlingStorageKey` just does this: ``${ThrottlingConstants.THROTTLING_PREFIX}.${JSON.stringify(thumbprint)}`;`

Since we don't need the throttling prefix here, instead of moving/renaming can we just do: 
```suggestion
        const silentRequestKey = JSON.stringify(thumbprint);
```

jo-arroyo: Makes sense. Changed.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -73,7 +78,7 @@ export class PublicClientApplication extends ClientApplication implements IPubli
     }
 
     /**
-     * Silently acquire an access token for a given set of scopes. Will use cached token if available, otherwise will attempt to acquire a new token from the network via refresh token.
+     * Silently acquire an access token for a given set of scopes. Returnings currently processing promise if parallel requests are made.
```
tnorling, line 81:
```suggestion
     * Silently acquire an access token for a given set of scopes. Returns currently processing promise if parallel requests are made.
```

tnorling (approved):
Looks good!

change/@azure-msal-browser-adabc718-b19b-4b98-a5b7-887465c75857.json
jasonnutter: I would classify this as fixing broken behavior, given the expectation from customers that MSAL should more gracefully handle concurrent requests out of the box, given we don't define it as a requirement today that you have to make all acquireTokenSilent calls in serial.

jasonnutter: Already a queued up minor change, so we can keep this as-is, resolving.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -21,6 +21,9 @@ import { version, name } from "../packageMetadata";
  */
 export class PublicClientApplication extends ClientApplication implements IPublicClientApplication {
 
+    // Redirect Response Object
```
jasonnutter, line 24: Looks like this comment is outdated.

jasonnutter (approved):

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -85,6 +90,39 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
+            scopes: request.scopes,
+            homeAccountIdentifier: account.homeAccountId
+        };
+        const silentRequestKey = JSON.stringify(thumbprint);
+        let response = this.activeSilentTokenRequests.get(silentRequestKey);
+        if (typeof response === "undefined") {
+            this.logger.verbose("acquireTokenSilent called for the first time, storing active request", request.correlationId);
+            response = this.acquireTokenSilentAsync(request, account)
+                .then((result) => {
+                    this.activeSilentTokenRequests.delete(silentRequestKey);
```
pkanher617, line 105: Should we cache and return responses for requests with the same thumbprint for longer than the duration of the request? It might require some thought around lifecycle for cached responses but maybe something to consider. Might be too big for this PR though.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -85,6 +90,39 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
+            clientId: this.config.auth.clientId,
+            authority: request.authority || "",
+            scopes: request.scopes,
+            homeAccountIdentifier: account.homeAccountId
+        };
+        const silentRequestKey = JSON.stringify(thumbprint);
+        let response = this.activeSilentTokenRequests.get(silentRequestKey);
+        if (typeof response === "undefined") {
+            this.logger.verbose("acquireTokenSilent called for the first time, storing active request", request.correlationId);
+            response = this.acquireTokenSilentAsync(request, account)
```
pkanher617, line 103: Can we use `async/await` with a `try/catch` here instead of using `.then/.catch`?

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -85,6 +90,39 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
             throw BrowserAuthError.createNoAccountError();
         }
+        const thumbprint: RequestThumbprint = {
+            clientId: this.config.auth.clientId,
+            authority: request.authority || "",
+            scopes: request.scopes,
+            homeAccountIdentifier: account.homeAccountId
+        };
+        const silentRequestKey = JSON.stringify(thumbprint);
+        let response = this.activeSilentTokenRequests.get(silentRequestKey);
```
pkanher617, line 100: Might be good to not reassign the variable here. could do `cachedResponse` here and just response inside the `if` statement.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -44,6 +47,8 @@ export class PublicClientApplication extends ClientApplication implements IPubli
      */
     constructor(configuration: Configuration) {
         super(configuration);
+
+        this.activeSilentTokenRequests = new Map();
```
pkanher617, line 51: Might be good to use the current APIs for internal memory storage?

pkanher617: `MemoryStorage.ts`

pkanher617: Never mind, it's not a string:string Map here

change/@azure-msal-browser-adabc718-b19b-4b98-a5b7-887465c75857.json
pkanher617: +1 for minor

lib/msal-browser/src/app/PublicClientApplication.ts
tnorling: Isn't that what local/session storage is for?

pkanher617: not sure what you mean

tnorling: Why would we need to hold on to the `activeSilentTokenRequests` object longer than the duration of the request? After the request is complete any subsequent calls to ATS should pull from local/session storage

pkanher617: Was just thinking that caching the actual response object would be better than doing the entire lookup again, but it's probably an over-optimization here.

lib/msal-browser/src/app/PublicClientApplication.ts
jasonnutter: We need to cache the entire promise chain, while keeping the cleanup logic outside of acquireTokenSilentAsync, so try/catch won't work.

lib/msal-browser/src/app/PublicClientApplication.ts
jo-arroyo: Done.

lib/msal-browser/src/app/PublicClientApplication.ts
jo-arroyo: Thanks for spotting that. Changed.

pkanher617 (approved):

PR #3838, opened 2021-07-07
ROPC added for confidential clients

sameerag:
Add ROPC support for `Confidential Client Applications`.


lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -210,6 +213,38 @@ export abstract class ClientApplication {
...
+     *
+     * @param request - UsenamePasswordRequest
+     */
+    async acquireTokenByUsernamePassword(request: UsernamePasswordRequest): Promise<AuthenticationResult | null> {
+        this.logger.info("acquireTokenByUsernamePassword called", request.correlationId);
+        const validRequest: CommonUsernamePasswordRequest = {
+            ...request,
+            ...this.initializeBaseRequest(request)
+        };
+        const serverTelemetryManager = this.initializeServerTelemetryManager(ApiId.acquireTokenByUsernamePassword, validRequest.correlationId!);
```
jasonnutter, line 232: Looks like we should turn off whatever lint rule is raising the "Forbidden non-null assertion" warning/error.

jasonnutter (approved):
Looks good, any docs or samples that could be updated?

lib/msal-node/src/client/ClientApplication.ts
tnorling: We turned this rule on when we turned on `strictNullChecks` to remind us to fix all the non-null assertions in node. We should fix all these warnings and then set the lint rule to error

sameerag:
> Looks good, any docs or samples that could be updated?

Yes. Working on a sample update and a few docs, will add in sometime.

lib/msal-node/src/client/ClientApplication.ts
jasonnutter: Makes sense, sounds good.

sameerag: Let's log this in ADO @tnorling. All pending work under strict-null-checks.

pkanher617 (approved):
Should make sure we update the relevant docs inside the repo, as well as on docs.microsoft.com

sameerag:
> Should make sure we update the relevant docs inside the repo, as well as on docs.microsoft.com

I checked and docs are already up to date. We do not add any docs in MSAL JS, but redirect it to docs.microsoft.com and we do mention that it is supported for both public and confidential clients.

samuelkubai (approved):
LGTM

PR #3855, opened 2021-07-08
Adding experimental client API

pkanher617:
This PR adds the ExperimentalPublicClientApplication.ts class, which will be used for future bleeding edge features without disturbing the current PublicClientApplication.ts surface.


lib/msal-browser/test/app/ExperimentalPublicClientApplication.spec.ts
```diff
@@ -0,0 +1,2937 @@
...
+import { NavigationClient } from "../../src/navigation/NavigationClient";
+import { NavigationOptions } from "../../src/navigation/NavigationOptions";
+import { PopupUtils } from "../../src/utils/PopupUtils";
+import { EndSessionPopupRequest } from "../../src/request/EndSessionPopupRequest";
+import { EventMessage } from "../../src/event/EventMessage";
+import { EventHandler } from "../../src/event/EventHandler";
+import { ExperimentalPublicClientApplication } from "../../src/app/ExperimentalPublicClientApplication";
+import { ClientApplication } from "../../src/app/ClientApplication";
+
+describe("ExperimentalPublicClientApplication.ts Class Unit Tests", () => {
```
jasonnutter, line 32: Is this a carbon copy of the PCA test class?

pkanher617: yes, for now

jasonnutter: It will probably get annoying to have to update this test class ever time something is added to PCA, so I wonder if we can find a way to ensure EPCA has the same functionality without needing to redo all the unit tests.

jasonnutter:
Missing an export of EPCA from the `index.ts` (unless you are intentionally leaving that out until we add the first experiemental feature.

lib/msal-browser/test/app/ExperimentalPublicClientApplication.spec.ts
tnorling: +1 Can we just assert that each API exists on EPCA instead? The functionality may change anyway as we "experiment"

sameerag: Is there a way we can detect changes and test? My concern would be `Codecov` but @tnorling's suggestion probably will address that.

pkanher617:
> Missing an export of EPCA from the `index.ts` (unless you are intentionally leaving that out until we add the first experiemental feature.

Yes, I was waiting until we had something to put in here before we start exporting this.

PR #3857, opened 2021-07-10
Doc update "acquireToken" to "loginRedirect"

zeerorg:
This PR is a documentation update.
# Description
I think the statement: "You may need to pass `prompt: "select_account"` or `prompt: "login"` to the **acquireToken** API in order for AAD to display the account selection screen after the first interaction" is wrong since we would need to select the account during the "login" phase rather than the acquireToken phase ?

Thanks,
Rishabh


lib/msal-browser/docs/accounts.md
```diff
@@ -53,7 +53,7 @@ async function getTokenPopup(request, account) {
...
 * If you have a multiple accounts scenario, please modify the [sample](../../../samples/msal-browser-samples/VanillaJSTestApp2.0/app/default/auth.js) (in `handleResponse()`) to list all cached accounts and choose a specific account.
 * If an application wants to retrieve an account based on the `username`, it needs to save the `username` (from the response of a `loginAPI` for a specific user) prior to using `getAccountByUsername()` API.
-* `getAllAccounts()` will return multiple accounts if you have made several interactive token requests and the user has selected different accounts in two or more of those interactions. You may need to pass `prompt: "select_account"` or `prompt: "login"` to the acquireToken API in order for AAD to display the account selection screen after the first interaction.
+* `getAllAccounts()` will return multiple accounts if you have made several interactive token requests and the user has selected different accounts in two or more of those interactions. You may need to pass `prompt: "select_account"` or `prompt: "login"` to the loginRedirect/loginPopup API in order for AAD to display the account selection screen after the first interaction.
```
tnorling, line 56: Prompts are valid for any interactive API call whether its "login" or "acquireToken"

```suggestion
* `getAllAccounts()` will return multiple accounts if you have made several interactive token requests and the user has selected different accounts in two or more of those interactions. You may need to pass `prompt: "select_account"` or `prompt: "login"` to the interactive acquireToken or login API in order for AAD to display the account selection screen after the first interaction.
```

zeerorg:
@tnorling  Got it, updated the PR.

PR #3867, opened 2021-07-15
Fix edge case causing acquireTokenSilent to hang in v1

tnorling:
Msal.js v1 uses the global window object to store callbacks and promises, which return the response/error back to the developer. There exists an edge case today that can cause `acquireTokenSilent` to hang and never return anything if another instance of `UserAgentApplication` is instantiated **while** another `acquireTokenSilent` call is in progress. This is due to the fact that the UAA constructor clears the global window objects used to track in progress requests, resulting in the in-progress request losing the reference to the promise it needs to resolve.

This PR addresses this unique scenario by checking if the objects are already defined on the window before instantiating them. It also updates the cleanup logic to delete stale keys rather than just setting the value to `null`.

### DIFF
--- a/lib/msal-core/test/UserAgentApplication.spec.ts
+++ b/lib/msal-core/test/UserAgentApplication.spec.ts
@@ -163,6 +163,14 @@ describe("UserAgentApplication.ts Class", function () {
         };
     };
 
+    afterEach(() => {
+        sinon.restore();
+        window.activeRenewals = {};
+        window.renewStates = [];
+        window.callbackMappedToRenewStates = {};
+        window.promiseMappedToRenewStates = {};
+    });
+
     describe("Telemetry in UserAgenApplication", () => {
         it("configure telemtry in UAA happy case smoke test", () => {
             msal = new UserAgentApplication({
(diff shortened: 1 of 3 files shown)


jasonnutter (approved):
Global state strikes again...thanks for debugging this!

lib/msal-core/test/UserAgentApplication.spec.ts
```diff
@@ -2838,6 +2844,33 @@ describe("UserAgentApplication.ts Class", function () {
...
+        it("acquireTokenSilent returns even if a new UserAgentApplication instance is instantiated", (done) => {
+            // Tests a bug whereby the instantiation of a 2nd UserAgentApplication was causing in-progress acquireTokenSilent calls on other instances to hang
+            const request = {
+                scopes: [TEST_CONFIG.MSAL_CLIENT_ID], 
+                loginHint: TEST_LOGIN_HINT,
+                forceRefresh: true
+            }
+            const TEST_LIBRARY_STATE_SILENT = RequestUtils.generateLibraryState(Constants.interactionTypeSilent);
+            sinon.stub(RequestUtils, "validateAndGenerateState").returns(TEST_LIBRARY_STATE_SILENT + "|");
+            sinon.stub(CryptoUtils, "createNewGuid").returns("123523")
```
peterzenz, line 2857: Does it matter if this isn't a valid guid, or does this just pad with 0s?

peterzenz (approved):

lib/msal-core/test/UserAgentApplication.spec.ts
tnorling: It doesn't matter, just needs to be a string. In this case it's forcing acquireTokenSilent to set the nonce to "123523" because that's what our test IdToken has as its nonce.

PR #3871, opened 2021-07-16
Move acquireToken code into 'InteractionClients'

tnorling:
Moves the bulk of the acquireToken code into individual InteractionClient classes, scoped to a single request.


hectormmg (approved):
Lgtm

jo-arroyo (approved):

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -4,30 +4,26 @@
...
 import { EndSessionRequest } from "../request/EndSessionRequest";
 import { BrowserConfigurationAuthError } from "../error/BrowserConfigurationAuthError";
-import { PopupUtils } from "../utils/PopupUtils";
 import { EndSessionPopupRequest } from "../request/EndSessionPopupRequest";
 import { INavigationClient } from "../navigation/INavigationClient";
-import { NavigationOptions } from "../navigation/NavigationOptions";
 import { EventHandler } from "../event/EventHandler";
+import { PopupClient } from "../interaction_client/PopupClient";
+import { RedirectClient } from "../interaction_client/RedirectClient";
+import { SilentIframeClient } from "../interaction_client/SilentIframeClient";
```
sameerag, line 25: `SilentIframe` and `SilentRefresh` - aren't they non-interactive by definition? Will this cause confusion?

tnorling: Open to other suggestions. This was what we came up with during design, keeping in mind the precedence set by the existing `SilentHandler` which extends `InteractionHandler` as well

sameerag: I am thinking, can we have a separate module for silent cases:
`"../silent_client/SilentIframeClient"`;

pkanher617: I think that's just over-optimization, should be fine to leave it as interaction_client.

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -417,83 +216,25 @@ export abstract class ClientApplication {
...
-            const state = this.validateAndExtractStateFromHash(hash, InteractionType.Popup, validRequest.correlationId);
-
-            // Remove throttle if it exists
-            ThrottlingUtils.removeThrottle(this.browserStorage, this.config.auth.clientId, authCodeRequest.authority, authCodeRequest.scopes);
-
-            // Handle response from hash string.
-            const result = await interactionHandler.handleCodeResponse(hash, state, authClient.authority, this.networkClient);
+        const popupClient = new PopupClient(this.config, this.browserStorage, this.browserCrypto, this.logger, this.eventHandler, this.navigationClient);
 
+        return popupClient.acquireToken(request).then((result) => {
```
pkanher617, line 237: I'm not sure I see the value of mixing .then and async/await, but we can change this later.

pkanher617 (approved):
Couple things that might need some updates in the future, but looks ok for now. Please make sure tests pass before merging.

lib/msal-browser/src/app/ClientApplication.ts
tnorling: Can't do async/await here unfortunately. We made the popup APIs synchronous a while back to enable users to configure popups to be opened synchronously or asynchronously.

lib/msal-browser/src/app/ClientApplication.ts
sameerag: Well ... :), don't have anything to add. Lets leave it at that for now but if we have two/three silent clients, I do think names matter.

change/@azure-msal-browser-30eb62a8-70f1-4fc5-b9d1-12fd3447960e.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
jasonnutter, line 2: AFAIK this doesn't change the API surface, right?

tnorling: Right, this is all internal implementation details. Made it minor just for the sheer volume of changes, acknowledging the small risk that something might break as a result

sameerag (approved):

jasonnutter (approved):

PR #3895, opened 2021-07-23
Add API to sideload tokens to msal-browser

jo-arroyo:
This PR:
- Adds `TokenCache` interface, with `loadTokens` API
- Adds `getTokenCache` to ClientApplication
- Adds new server response type `ExternalTokenResponse` to msal-common
- Unit tests for `TokenCache`
- Testing docs for `TokenCache` usage
- Update `AccountEntity` to allow `environment` to be used, and `authority` to be optional

Note:
- Moving account and cache utility APIs from `ClientApplication.ts` to `TokenCache.ts` will be done in another PR.


lib/msal-browser/src/cache/TokenCache.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    private loadIdToken(idToken: string, homeAccountId: string, environment: string, tenantId: string, options: LoadTokenOptions): void {
+
+        const idTokenEntity = IdTokenEntity.createIdTokenEntity(homeAccountId, environment, idToken, this.config.auth.clientId, tenantId);
+
+        if (this.isBrowserEnvironment) {
+            this.logger.verbose("TokenCache - loading id token");
+            this.storage.setIdTokenCredential(idTokenEntity);
+        } else {
+            if (options.callback) {
```
tnorling, line 77: Should a provided callback be given precedence over default window storage (this.isBrowserEnvironment condition)?

lib/msal-browser/src/cache/ITokenCache.ts
```diff
@@ -0,0 +1,14 @@
...
+ */
+
+import { ServerAuthorizationROPCResponse } from "@azure/msal-common";
+import { SilentRequest } from "../request/SilentRequest";
+import { LoadTokenOptions } from "./TokenCache";
+
+export interface ITokenCache {
+
+    /** API to side-load tokens to MSAL cache */
+    loadTokens(request: SilentRequest, response: ServerAuthorizationROPCResponse, options: LoadTokenOptions): void;
```
jasonnutter, line 13: Admittedly not the biggest fan of this name, will stew on this.

jo-arroyo: Yeah, definitely open to suggestions on any of the naming.

lib/msal-browser/src/cache/TokenCache.ts
jo-arroyo: In our design discussions, we decided to store in local/session storage by default, but to provide a way to use a callback if it wasn't available. Is there a reason we'd want the callback to take precedence?

tnorling: I don't think we should make assumptions about how people would use this API. If we offer a callback override we should honor it. It feels weird to offer an option that would be totally ignored in some scenarios.

lib/msal-browser/src/cache/ITokenCache.ts
sameerag: The filename stays right? `TokenCache` is a public interface for msal-node. For uniformity I think it helps to have similar interface implementations named the same way. Function names, yes, there could be more meaningful naming I guess.

lib/msal-browser/src/cache/TokenCache.ts
jo-arroyo: That's fair. Updated.

lib/msal-browser/docs/testing.md
```diff
@@ -0,0 +1,143 @@
...
+MSAL Browser starting version 2.17.0 has added the `loadTokens()` API, which allows the loading of id tokens and access tokens to the MSAL cache, which can then be read out using `acquireTokenSilent()`. 
+
+**Note: This is an advanced feature that is intended for testing purposes only. Loading tokens to your application's cache may cause your app to break.**
+
+## The loadTokens() API
+
+The `loadToken()` API can be accessed by calling `getTokenCache()` on MSAL Browser's `PublicClientApplication` instance. 
+
+```js
+myMSALObj.getTokenCache().loadTokens(silentRequest, serverResponse, loadTokenOptions);
```
sameerag, line 12: Is it better to make it:

```suggestion
const msalTokenCache = myMSALObj.getTokenCache();
msalTokenCache.loadTokens(silentRequest, serverResponse, loadTokenOptions);
```

sameerag: Essentially the whole sample can now use the `msalTokenCache` to make other calls to fetch data:

```javascript
msalTokenCache.getAllAccounts()
```

lib/msal-browser/docs/testing.md
```diff
@@ -0,0 +1,143 @@
...
+```
+
+`loadTokens()` takes in a request of type `SilentRequest`, a response of type `ServerAuthenticationROPCRespones`, and options of type `LoadTokenOptions`.
+
+See the type definitions for each below:
+
+- [`SilentRequest`](https://azuread.github.io/microsoft-authentication-library-for-js/ref/modules/_azure_msal_browser.html#silentrequest)
+- [`ServerAuthenticationROPCResponse`](https://azuread.github.io/microsoft-authentication-library-for-js/ref/modules/_azure_msal_common.html#serverauthorizationropcresponse)
```
sameerag, line 20: Do we need to add how to fetch this or is it already covered?

lib/msal-browser/docs/testing.md
```diff
@@ -0,0 +1,143 @@
...
+};
+
+OR
+
+const silentRequest: SilentRequest = {
+    scopes: ["User.Read", "email"],
+    authority: "https://login.microsoftonline.com/your-tenant-id"
+};
+
+const serverResponse: ServerAuthorizationROPCResponse = {
```
sameerag, line 88: Curious: Is there a use case where the server response contains only `idToken`? My understanding is that ROPC fetches all three. Of course it is also scope dependent?

sameerag: Or is this we giving options to the users to be able to pass any combo of these tokens?

jo-arroyo: As far as I know, ROPC fetches all three, but we are scoping `loadTokens` to just id tokens and access tokens at the moment, with access tokens being optional. Users can pass in just an id token, or an id token and an access token.

lib/msal-browser/docs/testing.md
jo-arroyo: Good point. I changed the doc to match. I felt this PR was getting to big, so I will move `getAllAccounts()` in another PR, but yes, eventually they will use the tokenCache to make other calls to fetch data.

lib/msal-browser/docs/testing.md
jo-arroyo: Thanks for pointing that out. I added instructions for importing.

lib/msal-browser/test/error/BrowserAuthError.spec.ts
```diff
@@ -283,4 +283,17 @@ describe("BrowserAuthError Unit Tests", () => {
         expect(err.name).toBe("BrowserAuthError");
         expect(err.stack?.includes("BrowserAuthError.spec.ts")).toBe(true);
     });
+
+    it.only("createUnableToLoadTokenError()", () => {
```
tnorling, line 287:
```suggestion
    it("createUnableToLoadTokenError()", () => {
```

samples/msal-browser-samples/VanillaJSTestApp2.0/app/customizable-e2e-test/auth.js
```diff
@@ -92,3 +92,28 @@ async function getTokenSilently() {
         });
     }
 }
+
+async function loadTokens() {
```
tnorling, line 96: I would not put this in the sample itself because in a real world use case you wouldn't want to have your actual app use this API. Instead the test file should call the ROPC flow and load the tokens into the cache using the puppeteer APIs

lib/msal-browser/docs/testing.md
sameerag: Got it. Can we please call this out in the docs? Basically when users make their service call, they will get the `refreshToken`. We want them to understand we are intentionally not accepting `refreshToken` for side loading tokens.

lib/msal-browser/src/cache/TokenCache.ts
```diff
@@ -0,0 +1,157 @@
...
+    loadTokens(request: SilentRequest, response: ServerAuthorizationROPCResponse, options: LoadTokenOptions): void {
+        this.logger.info("TokenCache - loadTokens called");
+
+        if (!response.id_token) {
+            throw BrowserAuthError.createUnableToLoadTokenError("Please ensure server response includes id token.");
+        }
+
+        if (request.account) {
+            this.loadIdToken(response.id_token, request.account.homeAccountId, request.account.environment, request.account.tenantId, options);
+            this.loadAccessToken(request, response, options, request.account.homeAccountId, request.account.environment, request.account.tenantId);
```
sameerag, line 61: nit: Can we manage the same order for common parameters? Increases readability.

lib/msal-browser/src/cache/TokenCache.ts
```diff
@@ -0,0 +1,157 @@
...
+
+            const authorityOptions: AuthorityOptions = {
+                protocolMode: this.config.auth.protocolMode,
+                knownAuthorities: this.config.auth.knownAuthorities,
+                cloudDiscoveryMetadata: this.config.auth.cloudDiscoveryMetadata,
+                authorityMetadata: this.config.auth.authorityMetadata
+            };
+            const authority = new Authority(request.authority, this.config.system.networkClient, this.storage, authorityOptions);
+
+            // If provided, clientInfo from options used before clientInfo in response
```
sameerag, line 72:
```suggestion
            // If provided, clientInfo from options takes precedence over. clientInfo in response
```

lib/msal-browser/src/cache/TokenCache.ts
jo-arroyo: Done.

samples/msal-browser-samples/VanillaJSTestApp2.0/app/customizable-e2e-test/auth.js
jo-arroyo: After working through e2e tests with Thomas, we have decided to scope this feature to be browser environments only. E2E tests have been removed.

lib/msal-browser/docs/testing.md
jo-arroyo: Good call-out. Done.

lib/msal-browser/src/cache/ITokenCache.ts
jasonnutter: "loadTokens" is the part I was referring to.

lib/msal-browser/src/cache/TokenCache.ts
```diff
@@ -0,0 +1,149 @@
...
+    // Browser cache storage
+    private storage: BrowserCacheManager;
+    // Logger
+    private logger: Logger;
+    // Crypto class
+    private cryptoObj: ICrypto;
+
+    constructor(configuration: Configuration, storage: BrowserCacheManager, logger: Logger, cryptoObj: ICrypto) {
+        this.isBrowserEnvironment = typeof window !== "undefined";
+        this.config = buildConfiguration(configuration, this.isBrowserEnvironment);
```
tnorling, line 35: I don't think you need to rebuild the configuration, right? The config you pass in should already be type `BrowserConfiguration`

lib/msal-browser/docs/testing.md
```diff
@@ -0,0 +1,138 @@
+# Testing
+
+MSAL Browser starting version 2.17.0 has added the `loadTokens()` API, which allows the loading of id tokens and access tokens to the MSAL cache, which can then be read out using `acquireTokenSilent()`. 
+
+**Note: This is an advanced feature that is intended for testing purposes in the browser environment only. Loading tokens to your application's cache may cause your app to break.**
+
+## The loadTokens() API
```
jasonnutter, line 7: I think this heading can go above the two sentences above.

lib/msal-common/src/response/ServerAuthorizationROPCResponse.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Deserialized response object from server authorization code request in the ROPC flow.
+ * - token_type: Indicates the token type value. The only type that Azure AD supports is Bearer.
+ * - scope: The scopes that the access_token is valid for.
+ * - expires_in: How long the access token is valid (in seconds).
+ * - access_token: The requested access token. The app can use this token to authenticate to the secured resource, such as a web API.
+ * - refresh_token: An OAuth 2.0 refresh token. The app can use this token acquire additional access tokens after the current access token expires.
+ * - id_token: A JSON Web Token (JWT). The app can decode the segments of this token to request information about the user who signed in.
+ * - client_info: Client info object 
+ */
+export type ServerAuthorizationROPCResponse = Pick<ServerAuthorizationTokenResponse, "token_type" | "scope" | "expires_in" | "id_token"> & {
```
tnorling, line 18: Is this type actually being used in our ROPC flow or is this type only used to describe the object shape that should be provided to the new loadToken API?

lib/msal-browser/src/cache/TokenCache.ts
jo-arroyo: You're right. Changed.

lib/msal-browser/docs/testing.md
jo-arroyo: Done.

lib/msal-common/src/response/ServerAuthorizationROPCResponse.ts
jo-arroyo: This type isn't used in the ROPC flow, but object shape to be provided is based on the ROPC flow. Open to suggestions for a better name.

tnorling: Maybe something along the lines of `TestTokenResponse` or `ExternalTokenResponse` given the tokens can come from anywhere, not just ROPC? And drop `refresh_token` if we don't support it?

lib/msal-browser/src/cache/ITokenCache.ts
hectormmg: How about `sideloadTokens`? A bit more accurate and descriptive?

jo-arroyo: Thanks for the input, all. Have changed the API name to `loadExternalTokens` for uniformity with `ExternalTokenResponse`. Let me know if there are any other concerns.

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -496,6 +502,13 @@ export abstract class ClientApplication {
         this.eventHandler.removeEventCallback(callbackId);
     }
 
+    /**
+     * Gets the token cache for the application.
+     */
+    getTokenCache(): TokenCache {
```
jasonnutter, line 508: We should add this to `IPublicClientApplication` as well.

jasonnutter: And I think the return type should `ITokenCache`

jo-arroyo: Done.

lib/msal-browser/docs/testing.md
```diff
@@ -0,0 +1,138 @@
...
+const msalTokenCache = myMSALObj.getTokenCache();
+msalTokenCache.loadExternalTokens(silentRequest, serverResponse, loadTokenOptions);
+```
+
+`loadExternalTokens()` takes in a request of type `SilentRequest`, a response of type `ServerAuthenticationROPCResponse`, and options of type `LoadTokenOptions`.
+
+See the type definitions for each, which can be imported from `@azure/msal-browser`:
+
+- [`SilentRequest`](https://azuread.github.io/microsoft-authentication-library-for-js/ref/modules/_azure_msal_browser.html#silentrequest)
+- [`ServerAuthorizationROPCResponse`](https://azuread.github.io/microsoft-authentication-library-for-js/ref/modules/_azure_msal_common.html#serverauthorizationropcresponse)
```
sameerag, line 21: Are we changing this name? As @tnorling suggested in a different comment.

jo-arroyo: Yes! Changed

lib/msal-common/src/response/ServerAuthorizationROPCResponse.ts
jo-arroyo: Done.

tnorling (approved):

jasonnutter (approved):

sameerag (approved):

PR #3899, opened 2021-07-23
Fix Logger constructor

### LINKED ISSUE #3887: Cannot set log level to Error with configuration
### Core Library

### Core Library Version

2.14.1

### Wrapper Library

Not Applicable

### Wrapper Library Version

2.14.1

### Description

It is not possible to specify ErrorLevel (0) as default log level due to or condition:
`this.level = loggerOptions.logLevel || LogLevel.Info;`
if logLevel is 0 (Error) then it is set to Info instead.

Fix should be simple, just change it to 
`this.level = loggerOptions.logLevel ?? LogLevel.Info;`

https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/2f66302660169c12ee8bb4622ecc0c497e1c7471/lib/msal-common/src/logger/Logger.ts#L65

### Error Message

_No response_

### Msal Logs

_No response_

### MSAL Configuration

```javascript
{
  auth: {
    not important for this case
  },
  system: {
    loggerOptions: {
        if (containsPii) {
          return;
        }
        switch (level) {
          case msal.LogLevel.Error:
            console.error(message);
            return;
          case msal.LogLevel.Info:
            console.info(message);
            return;
          case msal.LogLevel.Verbose:
            console.debug(message);
            return;
          case msal.LogLevel.Warning:
            console.warn(message);
            return;
        }
      },
      piiLoggingEnabled: false,
      logLevel: msal.LogLevel.Error, // cannot be set because it is zero and there is or condition
    },
```

### Relevant Code Snippets

```javascript
just create new myMSALObj and verify level
```

### Reproduction Steps

create new myMSALObj object with the provided configuration

### Expected Behavior

myMSALObj.logger.level should be set to 0 (Error)

### Identity Provider

Azure AD / MSA

### Regression

_No response_

### Source

Internal (Microsoft)

hectormmg:
This PR:

- Fixes logical check in `Logger` constructor that was treating `LogLevel.Error` (with value `0`) as `falsey` and thus always setting `level` to `LogLevel.Info` (with value `2`) whenever `logLevel` was configured to `Error`.
- Fixes #3887


lib/msal-common/src/logger/Logger.ts
```diff
@@ -62,7 +62,7 @@ export class Logger {
         const defaultLoggerCallback = () => {};
         this.localCallback = loggerOptions.loggerCallback || defaultLoggerCallback;
         this.piiLoggingEnabled = loggerOptions.piiLoggingEnabled || false;
-        this.level = loggerOptions.logLevel || LogLevel.Info;
+        this.level = (typeof(loggerOptions.logLevel) === "number") ? loggerOptions.logLevel : LogLevel.Info;
```
tnorling, line 65: Tests?

jasonnutter (commented):
Can you please add a test proving the bug is fixed?

lib/msal-common/src/logger/Logger.ts
hectormmg: On it, just have to get a little creative since the level attribute is private.

hectormmg: Added tests

hectormmg: Missed some edge cases, adding them now

hectormmg: Did a refactor after adding edge-case tests, validating LogLevel was a bit more complicated than it seemed.

lib/msal-common/src/logger/Logger.ts
```diff
@@ -76,6 +76,25 @@ export class Logger {
...
+        // Set default log level
+        let logLevel = LogLevel.Info;
+        const configuredLogLevel = loggerOptions.logLevel;
+        // Check if configured log level is a number
+        if (typeof(configuredLogLevel) === "number") {
+            // Check if configured log level is a supported log level
+            if (Object.keys(LogLevel).indexOf(`${configuredLogLevel}`) !== -1) {
+                logLevel = configuredLogLevel;
+            }
+        }
```
jasonnutter, line 93: This feels a little too complicated, considering in theory we could have this logic for every config enum, which is not preferable. I would be fine with simpler logic to fall back to Info if nothing is passed, and then in `logMessage`, throw an error if an invalid log level is set.

jasonnutter: Either that or do the validation and throw errors in configuration class, where the other defaults are set.

hectormmg: I guess we can just make sure its a number and leave it at that. In the off chance someone inputs a negative log level in config, then they just won't see logs, and if they input a level greater than our current highest, they'll just see all the logs.

So in summary, non-numbers in `logLevel` will default to 2 (Info), and numbers outside the range will either mean no logs (< 0) or all logs (> Trace).

lib/msal-common/src/logger/Logger.ts
```diff
@@ -62,7 +62,7 @@ export class Logger {
         const defaultLoggerCallback = () => {};
         this.localCallback = loggerOptions.loggerCallback || defaultLoggerCallback;
         this.piiLoggingEnabled = loggerOptions.piiLoggingEnabled || false;
-        this.level = loggerOptions.logLevel || LogLevel.Info;
+        this.level = this.validateLogLevel(loggerOptions);
```
tnorling, line 65: Would this work?
```suggestion
        this.level = typeof(loggerOptions.logLevel) === "number" ? loggerOptions.logLevel : LogLevel.Info;
```

hectormmg: It works for non-numbers, but it fails to catch cases in which the logLevel number doesn't represent one of the established levels (i.e. anything under 0 and over 4 at the moment).  For numbers over 4, it would just default to `Trace` since it's the highest level.

jasonnutter: That's fine. We provide typings that codify the accepted values.

lib/msal-common/src/logger/Logger.ts
```diff
@@ -64,7 +64,7 @@ export class Logger {
         };
         this.localCallback = loggerOptions.loggerCallback || defaultLoggerCallback;
         this.piiLoggingEnabled = loggerOptions.piiLoggingEnabled || false;
-        this.level = loggerOptions.logLevel || LogLevel.Info;
+        this.level = typeof(loggerOptions.logLevel) === "number" ? loggerOptions.logLevel : LogLevel.Info;
```
peterzenz, line 67: Do we want to verify it's a valid LogLevel and if not set it to LogLevel.Info as well

peterzenz (approved):

lib/msal-common/src/logger/Logger.ts
hectormmg: That's what I implemented originally, but we agreed that the validation logic added too much complexity. Here's why we think this is the appropriate change:

1. This catches and handles the "non-number" issue
2. Any number higher than the max supported level will result in an equivalent to setting the level to `Trace`
3. Any number lower than the minimum supported (0 for Error) will result in an equivalent to setting the level to `Error`
4. We provide the `LogLevel` enum, so developers generally don't need to set the level numeric value themselves.

jasonnutter (approved):

tnorling (approved):

PR #3903, opened 2021-07-27
State encoding issue

### LINKED ISSUE #3491: URIError on Safari
### Core Library

@azure/msal-browser

### Core Library Version

2.12.1

### Wrapper Library

Not Applicable

### Wrapper Library Version

NA

### Description

When passing "viva-pi%C3%B1ata-tip" as state in loginRedirect, it somehow gets converted into "viva-pi%25u00f1ata-tip". when [StringUtils](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/fee7efbd7616e144918c969af5cce2c927db0ba0/lib/msal-common/src/utils/StringUtils.ts#L66) tries to run that through decodeURIComponent, it fails with URIError

Temporary work around is to URIEncode the state before passing it to loginRedirect using encodeURIComponent.

Could be a [similar issue](https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/2088) 

### Error Message

URIError

### Msal Logs

_No response_

### MSAL Configuration

```javascript
{
  "auth": {
    "clientId": "1f907974-e22b-4810-a9de-d9647380c97e",
    "authority": "https://login.microsoftonline.com/consumers",
    "navigateToLoginRequestUrl": false
  },
  "cache": { "cacheLocation": "localStorage" }
}
```

### Relevant Code Snippets

```javascript
// On app startup
const pca = new PublicClientApplication(msalConfig);
await pca.loginRedirect({
  scopes: [
    "service::http://Passport.NET/purpose::PURPOSE_XBOX_CLOUD_CONSOLE_TRANSFER_TOKEN",
  ],
  state: "viva-pi%C3%B1ata-tip",
});

// Or successful redirect
try {
  const redirectResult = await pca.handleRedirectPromise();
} catch (err) {
  // err = URIError
}
```

### Reproduction Steps

see above code snippet

### Expected Behavior

`handleRedirectPromise` should not throw an error and `viva-pi%C3%B1ata-tip` should be returned as is in the state

### Identity Provider

Azure AD / MSA

Safari

### Regression

_No response_

### Source

Internal (Microsoft)

bmahall:
Fixes #3491 
This PR ->
-Removes nested decodeURIComponent in queryStringToObject{StringUtils.ts}
-Removes the decodeURIComponent in parseRequestState {ProtocolUtils.ts} (which was resulting in an extra decode of the state val)
-Unit tests for each of the above changes
-Removes an existing unit test case in AuthorizationCodeClient.spec.ts that was checking for server response in case of double encoded state val


lib/msal-common/test/utils/ProtocolUtils.spec.ts
```diff
@@ -93,4 +93,10 @@ describe("ProtocolUtils.ts Class Unit Tests", () => {
         const requestState = ProtocolUtils.parseRequestState(cryptoInterface, testState);
         expect(requestState.userRequestState).toBe(userState);
     });
+
+    it("parseRequestState() correctly splits the state by the resource delimiter", () => {
```
tnorling, line 97:
```suggestion
    it("parseRequestState returns user state without decoding", () => {
```

lib/msal-common/test/utils/StringUtils.spec.ts
```diff
@@ -138,6 +138,14 @@ describe("StringUtils.ts Class Unit Tests", () => {
 
     });
 
+    it("queryStringToObject correctly deserializes query string into object", () => {
```
tnorling, line 141:
```suggestion
    it("queryStringToObject correctly deserializes URI encoded query string into decoded object", () => {
```

tnorling (approved):

hectormmg (approved):

pkanher617 (approved):

PR #3910, opened 2021-07-29
Fix AccessToken_With_AuthScheme implementation

hectormmg:
This PR fixes the AccessToken_With_AuthScheme implementation, which was missing an extra field for the auth scheme in the cache key.


lib/msal-common/src/cache/entities/CredentialEntity.ts
```diff
@@ -126,14 +131,19 @@ export class CredentialEntity {
         clientId: string,
         realm?: string,
         target?: string,
-        familyId?: string
+        familyId?: string,
+        tokenType?: string
```
tnorling, line 135: Reason for using type string here instead of the enum?

lib/msal-common/src/cache/entities/CredentialEntity.ts
```diff
@@ -182,4 +192,16 @@ export class CredentialEntity {
     private static generateTargetForCacheKey(scopes?: string): string {
         return (scopes || "").toLowerCase();
     }
+
+    /**
+     * Generates scheme key component as per schema: <-scheme>
+     */
+    private static generateSchemeForCacheKey(credentialType: string, tokenType?: string): string {
```
tnorling, line 199: Any reason tokenType shouldn't be a required param here? Looks like you've already validated that its a non-empty string before this gets called

lib/msal-common/src/cache/entities/CredentialEntity.ts
hectormmg: @tnorling  The AT with AuthScheme spec points out that the `tokenType` value is to be stored as-is from the server response. Would that prevent us from making it an enum? Otherwise I think it's a good idea.

lib/msal-common/src/cache/entities/CredentialEntity.ts
hectormmg: Good catch, initially this was meant to be called for all credentials, most of which wouldn't have a tokenType and we'd just leave an extra trailing dash. After a lot of discussion/investigation we decided to limit the key change to AT with AuthScheme, which always has a tokenType, so now we can make it required.

lib/msal-common/src/cache/entities/CredentialEntity.ts
tnorling: You're comparing it to an enum below. I'm assuming that means if it isn't a pre-defined, known value i.e. "pop" that we're discarding it? Do I have that right?

lib/msal-common/src/cache/entities/CredentialEntity.ts
hectormmg: As I did the update I realized the whole method was unnecessary, we could just push the tokenType directly into the cache key elements array.

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -335,7 +341,8 @@ export abstract class CacheManager implements ICacheManager {
         familyId?: string,
         realm?: string,
         target?: string,
-        oboAssertion?: string
+        oboAssertion?: string,
+        tokenType?: string
```
tnorling, line 345:
```suggestion
        tokenType?: AuthenticationScheme
```

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -820,6 +842,15 @@ export abstract class CacheManager implements ICacheManager {
         return entityScopeSet.containsScopeSet(requestTargetScopeSet);
     }
 
+    /**
+     * Returns true if the credential's tokenType or Authentication Scheme matches the one in the request, false otherwise
+     * @param entity 
+     * @param tokenType 
+     */
+    private matchTokenType(entity: CredentialEntity, tokenType: string): boolean {
```
tnorling, line 850:
```suggestion
    private matchTokenType(entity: CredentialEntity, tokenType: AuthenticationScheme): boolean {
```

lib/msal-common/src/cache/utils/CacheTypes.ts
```diff
@@ -54,6 +54,7 @@ export type CredentialFilter = {
     realm?: string;
     target?: string;
     oboAssertion?: string;
+    tokenType?: string;
```
tnorling, line 57:
```suggestion
    tokenType?: AuthenticationScheme;
```

sameerag (approved):
No new comments apart from what are raised already. Approving.

change/@azure-msal-node-387567d9-d62e-4380-bf2e-b7a4187e9d01.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
tnorling, line 2: Only test changes
```suggestion
  "type": "none",
```

change/@azure-msal-browser-52e7c2a2-f316-463e-a74d-bdd0d0fcb112.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
tnorling, line 2: Only test changes
```suggestion
  "type": "none",
```

tnorling (approved):

PR #3930, opened 2021-08-04
Add correlationId to errors thrown

tnorling:
Changes include:

- Adds `setCorrelationId` method to `AuthError` class
- CorrelationId is added to the error before being thrown from all public APIs
- CorrelationId is generated upfront when instantiating `InteractionClient`
- CorrelationId is stored as a temporary cache item during the redirect flow
- Logger is cloned with new CorrelationId when instantiating `InteractionClient`

Follow up PR will address successful responses

### DIFF
--- a/.github/workflows/beachball-check.yml
+++ b/.github/workflows/beachball-check.yml
@@ -26,14 +26,19 @@ jobs:
         filters: |
           lib:
             - 'lib/**/**.!(md)'
+
     - uses: actions/checkout@v2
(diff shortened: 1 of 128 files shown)


jasonnutter (approved):

jo-arroyo (approved):

sameerag (approved):
Lets get the `interactionHandler` code cleaned up eventually. This looks good. Approved.

PR #3944, opened 2021-08-09
Fixed a bug where msal-node doesn't work behind a proxy

Robbie-Microsoft:


lib/msal-node/src/network/HttpClient.ts
```diff
@@ -23,21 +25,39 @@ export class HttpClient implements INetworkModule {
...
+        proxy?: string
     ): Promise<NetworkResponse<T>> {
-        const request: AxiosRequestConfig = {
-            method: HttpMethod.GET,
-            url: url,
-            headers: options && options.headers,
-            validateStatus: () => true
-        };
+        let response;
+        let responseObject: any = {};
```
jasonnutter, line 33: Prefer to not use `any`, please define a type.

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -23,21 +25,39 @@ export class HttpClient implements INetworkModule {
      */
     async sendGetRequestAsync<T>(
         url: string,
-        options?: NetworkRequestOptions
+        options?: NetworkRequestOptions,
+        cancellationToken?: number,
+        proxy?: string
```
jasonnutter, line 30: Can `proxy` and `cancellationToken` be added to `NetworkRequestOptions` instead? I would like us to get out of the habit of many optional parameters like this.

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -10,6 +10,8 @@ import {
 } from "@azure/msal-common";
 import { HttpMethod } from "../utils/Constants";
 import axios, { AxiosRequestConfig } from "axios";
+import { HttpsProxyAgent } from "https-proxy-agent";
+import fetch from "node-fetch";
```
jasonnutter, line 14: So Axios doesn't support working with proxies? I find that surpising.

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -23,21 +25,39 @@ export class HttpClient implements INetworkModule {
...
+        let responseObject: any = {};
+
+        if (proxy) {
+            const fetchOptions =  {
+                headers: options?.headers,
+                body: options?.body,
+                agent: new HttpsProxyAgent(proxy),
+            };
+    
+            response = await fetch(url, fetchOptions);
```
jasonnutter, line 42: The other option is to document how applications can provide their own http client, which give them complete control over the proxied requests. Did we consider that?

lib/msal-node/src/network/HttpClient.ts
jasonnutter: And `const` is fine, since you are not reassigning the object.

jasonnutter (changes requested):
A few comments. Also, would like to see tests for this feature. Thanks!

lib/msal-node/src/network/HttpClient.ts
sameerag: Yeah. Its took a few escalations before we figured axios is the issue. :)

lib/msal-node/src/network/HttpClient.ts
sameerag: Yes. I am about to suggest that. The alternative I have in mind is provide a sample with a network client that supports proxies and leave it at that with instructions. We are open to suggestions.

Robbie-Microsoft: Per our discussion, we are switching from axios to node-fetch.

tnorling:
Addressed in #4447

PR #3946, opened 2021-08-09
Add configuration for popup window size and placement in msal-browser

### LINKED ISSUE #2603: popup window in a non full screen iframe, the window is not centered.
we are using msal-browser to popup window inside an iframe, but the iframe has a specific small size, not in full screen, this caused the popup window not centered. Could you add a configuration to specify the popup window's position?

jo-arroyo:
Adds configuration for popup window dimension and position to `PopupRequest` and `EndSessionPopupRequest`, and updates `PopupClient` and `PopupUtils` to check for configuration when opening popups.

Addresses #2603.


lib/msal-browser/src/utils/PopupUtils.ts
```diff
@@ -82,6 +104,9 @@ export class PopupUtils {
         const left = Math.max(0, ((width / 2) - (BrowserConstants.POPUP_WIDTH / 2)) + winLeft);
```
peterzenz, line 104: If the top/left are passed in to the popupConfig as zero then we should use the logic you have here to center the popup.

A very small thing to consider is that you shouldn't compute top/left unless it's needed.  So if width/height/top/left are all set, we should skip the math here.

lib/msal-browser/src/interaction_client/PopupClient.ts
```diff
@@ -28,11 +28,15 @@ export class PopupClient extends StandardInteractionClient {
             // asyncPopups flag is true. Acquires token without first opening popup. Popup will be opened later asynchronously.
             if (this.config.system.asyncPopups) {
                 this.logger.verbose("asyncPopups set to true, acquiring token");
+                // Passes on popup position and dimensions if in request
+                if (request.popupConfiguration) {
```
peterzenz, line 32: You use the pattern "if popup config then do with popup config, else do without popup config" throughout the code.  You might want to consider adding a default popup config that's always used to eliminate the if/elses and just use the do with popup config approach.  If the user passes in values, you can overwrite the default popup config.

lib/msal-browser/docs/logout.md
```diff
@@ -82,7 +82,13 @@ const currentAccount = msalInstance.getAccountByHomeId(homeAccountId);
 await msalInstance.logoutPopup({
     account: currentAccount,
     postLogoutRedirectUri: "https://contoso.com/loggedOut",
-    mainWindowRedirectUri: "https://contoso.com/homePage"
+    mainWindowRedirectUri: "https://contoso.com/homePage",
+    popupConfiguration: {
+        height: 100,
```
peterzenz, line 87: Values of 100/100/100/100 are great for a sample or test, but the code should likely enforce some sane defaults.  For instance, 1/0/9999/9999 shouldn't be a valid popup config.  Considerations: popup should be smaller than screen/parent window size, trying to position off-screen should likely be invalid as well.  If we think the config is bad, we can reject it.  Documentation should be updated to reflect the approach we take for this.

lib/msal-browser/src/utils/PopupUtils.ts
```diff
@@ -9,6 +9,16 @@ import { BrowserAuthError } from "../error/BrowserAuthError";
 import { AuthorizationUrlRequest } from "../request/AuthorizationUrlRequest";
 import { BrowserConstants, InteractionType, TemporaryCacheKeys } from "./BrowserConstants";
 
+/**
+ * Popup configurations for setting dimensions and position of popup window
+ */
+export type PopupConfiguration = {
```
tnorling, line 15: Can we call this something like `PopupDimensions` or `PopupWindowAttributes` instead? We should try to avoid creating confusion about whether this references popup api "configuration" or popup window "configuration" or both

lib/msal-browser/docs/logout.md
jasonnutter: We have default height width defined already: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/f9d2bdbab54f39cb32cd7f82acb5647a296db493/lib/msal-browser/src/utils/BrowserConstants.ts#L25-L29

And top/left is dynamically calculated based on the window. I agree there are some non-sensical values we could/should protect against, but I think we want to be careful to be too opinionated (given an application can open its own popup window with whatever dimensions it wants, in theory).

lib/msal-browser/src/utils/PopupUtils.ts
```diff
@@ -9,6 +9,16 @@ import { BrowserAuthError } from "../error/BrowserAuthError";
...
+    height: number;
+    width: number;
+    top: number;
+    left: number;
```
jasonnutter, line 19: One question I have is whether all of these should be required, or if we can make some of them optional. For example, if an app provides just height and width, we can assume they want it centered, and use the logic we have today to center it for them.

jasonnutter: (I see Peter made a similar suggestion below)

tnorling: What if we make this 2 types `PopupSize` (height, width) and `PopupPosition` (top, left) where both are optional but the values inside each are required. I don't think someone would provide just the height but not the width, for example, right?

lib/msal-browser/src/error/BrowserAuthError.ts
```diff
@@ -254,6 +258,16 @@ export class BrowserAuthError extends AuthError {
         return new BrowserAuthError(BrowserAuthErrorMessage.popUpWindowError.code, errorMessage);
     }
 
+    /**
+     * Creates an error thrown when the popup window could not be opened because invalid attributes passed in request.
+     * @param errDetail 
+     */
+    static createPopupWindowAttributeError(errDetail?: string): BrowserAuthError {
```
peterzenz, line 265: I'd recommend against failing in this flow and instead logging an error and using the default size and centered position.  If the developer wants a larger popup than the users screen size then the user would not get a prompt.

lib/msal-browser/docs/logout.md
jo-arroyo: Any suggestions for these defaults? I recognize 1/1/1/1 would be virtually useless, but agree that we don't want to be too opinionated

lib/msal-browser/src/interaction_client/PopupClient.ts
```diff
@@ -72,12 +79,14 @@ export class PopupClient extends StandardInteractionClient {
...
      * Helper which obtains an access_token for your API via opening a popup window in the user's browser
+     * @param validRequest
      * @param popupName
      * @param popup
+     * @param popupConfig
      *
      * @returns A promise that is fulfilled when this function has completed, or rejected if an error was raised.
      */
-    private async acquireTokenPopupAsync(validRequest: AuthorizationUrlRequest, popupName: string, popup?: Window|null): Promise<AuthenticationResult> {
+    private async acquireTokenPopupAsync(validRequest: AuthorizationUrlRequest, popupName: string, popup?: Window|null, popupConfig?: PopupConfiguration|undefined): Promise<AuthenticationResult> {
```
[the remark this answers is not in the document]
jo-arroyo: Yes, thanks for the catch. Changed.

lib/msal-browser/src/utils/PopupUtils.ts
jo-arroyo: I originally considered dimensions but wondered if that would encompass position well. I changed it to `PopupWindowAttributes`, thanks for the suggestion.

lib/msal-browser/src/utils/PopupUtils.ts
jo-arroyo: I like that, I changed it to `PopupWindowAttribute` with the optional `PopupSize` and `PopupPosition`, with values required.

lib/msal-browser/src/interaction_client/PopupClient.ts
jo-arroyo: Thanks for the suggestion. Updated the approach.

lib/msal-browser/src/utils/PopupUtils.ts
jo-arroyo: Done.

lib/msal-browser/docs/logout.md
jo-arroyo: Thanks for the suggestions. Have decided to log errors and use defaults when popup window is out of bounds, non-sensical values are used, or window is larger than screen/parent window size. Have not set other minimums or maximum, as we do not want to be too opinionated, and it should be up to the developer to set a popup window size that is reasonable for their app and users to use.

lib/msal-browser/src/interaction_client/PopupClient.ts
```diff
@@ -72,12 +75,14 @@ export class PopupClient extends StandardInteractionClient {
...
      * Helper which obtains an access_token for your API via opening a popup window in the user's browser
+     * @param validRequest
      * @param popupName
      * @param popup
+     * @param popupWindowAttributes
      *
      * @returns A promise that is fulfilled when this function has completed, or rejected if an error was raised.
      */
-    private async acquireTokenPopupAsync(validRequest: AuthorizationUrlRequest, popupName: string, popup?: Window|null): Promise<AuthenticationResult> {
+    private async acquireTokenPopupAsync(validRequest: AuthorizationUrlRequest, popupName: string, popupWindowAttributes: PopupWindowAttributes, popup?: Window|null, ): Promise<AuthenticationResult> {
```
tnorling, line 85:
```suggestion
    private async acquireTokenPopupAsync(validRequest: AuthorizationUrlRequest, popupName: string, popupWindowAttributes: PopupWindowAttributes, popup?: Window|null): Promise<AuthenticationResult> {
```

lib/msal-browser/src/interaction_handler/PopupHandler.ts
```diff
@@ -41,7 +42,7 @@ export class PopupHandler extends InteractionHandler {
             this.browserStorage.setTemporaryCache(TemporaryCacheKeys.INTERACTION_STATUS_KEY, BrowserConstants.INTERACTION_IN_PROGRESS_VALUE, true);
             this.browserRequestLogger.infoPii(`Navigate to: ${requestUrl}`);
             // Open the popup window to requestUrl.
-            return this.popupUtils.openPopup(requestUrl, params.popupName, params.popup);
+            return this.popupUtils.openPopup(requestUrl, params.popupName, params.popupWindowAttributes, params.popup);
```
tnorling, line 45: can you just pass `params` here instead of each property separately?

lib/msal-browser/src/utils/PopupUtils.ts
```diff
@@ -77,12 +103,39 @@ export class PopupUtils {
...
-        const height = window.innerHeight || document.documentElement.clientHeight || document.body.clientHeight;
-        const left = Math.max(0, ((width / 2) - (BrowserConstants.POPUP_WIDTH / 2)) + winLeft);
-        const top = Math.max(0, ((height / 2) - (BrowserConstants.POPUP_HEIGHT / 2)) + winTop);
+        const winWidth = window.innerWidth || document.documentElement.clientWidth || document.body.clientWidth;
+        const winHeight = window.innerHeight || document.documentElement.clientHeight || document.body.clientHeight;
+
+        if (popupWindowAttributes.popupPosition) {
+            if (popupWindowAttributes.popupPosition.top < 0 || popupWindowAttributes.popupPosition.top > winHeight) {
+                logger.error("Invalid popup window position. Popup window should not be positioned off-screen.");
+                popupWindowAttributes.popupPosition.top = 0;
```
tnorling, line 112: Reassigning input params. Can we move the `validHeight`, `validWidth` etc. vars to the top and overwrite with the defaults if they're not valid?

lib/msal-browser/src/interaction_handler/PopupHandler.ts
jo-arroyo: Good idea. Done.

lib/msal-browser/src/utils/PopupUtils.ts
jo-arroyo: Sure, done.

peterzenz (approved):

jasonnutter (approved):

PR #3947, opened 2021-08-09
Add correlationId to AuthenticationResult

tnorling:
Adds `correlationId` to the `AuthenticationResult` response type returned by all `acquireTokenXXX` APIs


change/@azure-msal-browser-7788dd86-9af5-4992-a2d3-623b1f85ce50.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
```
jasonnutter, line 2: Do we want to mark as patch, so that browser gets a changelog entry when it consumes the updated version of common?

jasonnutter: Ditto with node.

tnorling: Yeah that's not a bad idea. Maybe in the future we should have our release script append "bumped @azure/msal-common to version x.x.x" and the corresponding common changelog to the bottom of the browser/node changelog so this won't be necessary for changes like this.

jasonnutter (approved):

sameerag (approved):
Lgtm

PR #3960, opened 2021-08-10
Add PerformanceManager and PerformanceMeasurement

hectormmg:
First pass for 2.14.1: https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/3965


lib/msal-browser/src/telemetry/PerformanceMeasurement.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    endMeasurement():void {
+        if (PerformanceMeasurement.supportsBrowserPerformance()) {
+            window.performance.mark(this.endMark);
+            window.performance.measure(this.measureName, this.startMark, this.endMark);
+        }
+    }
+
+    flushMeasurement(): number {
+        if (PerformanceMeasurement.supportsBrowserPerformance()) {
```
jasonnutter, line 38: This check can go in the manager class now.

lib/msal-browser/src/telemetry/PerformanceMeasurement.ts
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    endMeasurement():void {
+        if (PerformanceMeasurement.supportsBrowserPerformance()) {
+            window.performance.mark(this.endMark);
+            window.performance.measure(this.measureName, this.startMark, this.endMark);
+        }
+    }
+
+    flushMeasurement(): number {
```
jasonnutter, line 37: These names can just be start, end, flush, etc.

lib/msal-browser/src/telemetry/PerformanceManager.ts
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { Logger } from "@azure/msal-common";
+import { PerformanceMeasurement } from "./PerformanceMeasurement";
+
+export class PerformanceManager {
```
jasonnutter, line 9: This should probably end up being `BrowserPerformanceManager`, with an abstract `PerformanceManager` in common.

lib/msal-browser/src/telemetry/PerformanceMeasurement.ts
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+export class PerformanceMeasurement {
```
jasonnutter, line 6: We'll want an interface for this class, e.g. `IPerformanceMeasurement`

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -84,6 +84,7 @@ export class PublicClientApplication extends ClientApplication implements IPubli
      * @returns {Promise.<AuthenticationResult>} - a promise that is fulfilled when this function has completed, or rejected if an error was raised. Returns the {@link AuthResponse} object
      */
     async acquireTokenSilent(request: SilentRequest): Promise<AuthenticationResult> {
+        const endMeasurement = this.performanceManager.startMeasurement("acquireTokenSilent");
```
jasonnutter, line 87: Let's create an enum for the measurement names, so that when we expose an api, apps can use that enum to filter which values they get back.

jasonnutter (commented):
Thanks for starting this. Posted some immediate thoughts based on our pairing session, don't bother addressing until later though.

lib/msal-browser/src/telemetry/PerformanceManager.ts
jasonnutter: And come up with a name other than "manager"

lib/msal-browser/src/telemetry/PerformanceMeasurement.ts
```diff
@@ -0,0 +1,47 @@
...
+    private measureName: string;
+    private startMark: string;
+    private endMark: string;
+
+    constructor(measureName: string) {
```
jasonnutter, line 11: correlation id should be included as well.

lib/msal-browser/src/telemetry/PerformanceManager.ts
```diff
@@ -0,0 +1,28 @@
...
+    constructor(logger: Logger) {
+        this.logger = logger;
+    }
+
+    startMeasurement(measureName: string): Function {
+        this.logger.trace(`Performance measurement started for ${measureName}`);
+        const performanceMeasure = new PerformanceMeasurement(measureName);
+        performanceMeasure.startMeasurement();
+
+        return () => {
```
jasonnutter, line 21:
```suggestion
        return (): number => {
```

jasonnutter: Actually this should return more than just the number, probably an object with number, name, correlation id, etc.

lib/msal-browser/src/app/PublicClientApplication.ts
sameerag: Agree. And keep it crisp: START, END, FLUSH which is standard across performance metrics.

jasonnutter:
Closing in favor of #4570

PR #3963, opened 2021-08-11
Remove experimental folder from lerna

jasonnutter:
This isn't needed right now, and just slows down `lerna bootstrap`.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -25,8 +25,7 @@
     }
   },
   "packages": [
-    "lib/*",
-    "experimental/*"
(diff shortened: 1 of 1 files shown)


tnorling (approved):

tnorling:
Might be good to also add `lerna.json` to the workflow dependency cache keys so we regenerate node_modules when it changes

jasonnutter:
Do you think that is necessary, given `lerna.json` should not change often?

sameerag (approved):
Approved

PR #3965, opened 2021-08-11
Instrument 2.14 to obtain performance measurements for acquireTokenSilence

hectormmg:


lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -78,6 +78,7 @@ export class PublicClientApplication extends ClientApplication implements IPubli
      * @returns {Promise.<AuthenticationResult>} - a promise that is fulfilled when this function has completed, or rejected if an error was raised. Returns the {@link AuthResponse} object
      */
     async acquireTokenSilent(request: SilentRequest): Promise<AuthenticationResult> {
+        const endMeasurement = this.performanceManager.startMeasurement("acquireTokenSilent");
```
peterzenz, line 81: For these can we pass in the same ApiId we're using for the call?  ApiId.acquireTokenSilent_silentFlow in this case?

peterzenz (approved):
This is a good first pass here.   Next up will be correlating perf data with call results, success vs the various error states we could land on.  For instance, you might find that all of the timeouts have a call time of 30+ seconds.  We could possibly exclude these calls from our analysis by concluding that this happens when our endpoints are unreachable from the client.  Some other error might consistently fall into a higher than normal timeframe to complete.  This gives us different pivots into the data and can help focus perf investigations as well.

lib/msal-browser/src/app/PublicClientApplication.ts
jasonnutter: For now, these labels are just used for label the measurements that show up in the browser performance profile. We can easily include more data once we start sending this data to the server.

jasonnutter: @hectormmg Might be good to show a screenshot of these annotations to provide context.

peterzenz (approved):

tnorling (approved):
This looks fine for the backport. When we implement this in the latest version I wonder if it would be useful to have the ability to make measurements in between start and end. i.e. start the measurement at the start of `acquireTokenSilent` get the time after the cache lookup, RT renewal and iframe renewal and then end it. That way at each step we can see the total time since the start as well as the time since the previous step, without needing to call `startMeasurement` for each individual measurement. Just something to consider

peterzenz:
> This looks fine for the backport. When we implement this in the latest version I wonder if it would be useful to have the ability to make measurements in between start and end. i.e. start the measurement at the start of `acquireTokenSilent` get the time after the cache lookup, RT renewal and iframe renewal and then end it. That way at each step we can see the total time since the start as well as the time since the previous step, without needing to call `startMeasurement` for each individual measurement. Just something to consider

This is a great suggestion and it plays well into long term plans for telemetry around perf.  Thanks!

jasonnutter:
> nd iframe renewal and then end it. That way at each step we can see the total time since the start as well as the time since the previous ste

This can be done fairly easily, as the browser performance apis allow you to take measurements between arbitrary marks.

PR #3974, opened 2021-08-16
[Docs] Clarify post logout redirect and refresh tokens

tnorling:
Quick update to clarify a couple of questions that have come up recently


derisen (approved):

lib/msal-browser/docs/logout.md
```diff
@@ -16,7 +16,13 @@ msalInstance.logoutRedirect();
 msalInstance.logoutPopup();
 ```
 
-These APIs will clear the token cache of any user and session data, then navigate the browser window or popup window to the server's logout page. The server will then prompt the user to select the account they would like to be signed out of and redirect back to the page the url provided as `postLogoutRedirectUri`.
+These APIs will clear the token cache of any user and session data, then navigate the browser window or popup window to the server's logout page. The server will then prompt the user to select the account they would like to be signed out of and redirect back to your `postLogoutRedirectUri` as long as the following conditions are met:
+
+1. The URI is registered as a reply url on the app registration
+1. The URI is provided as the `postLogoutRedirectUri` on either the `PublicClientApplication` config or the logout request
+1. The user has an active session with the identity provider
```
jasonnutter, line 23: I believe the MSA service also requires a front channel logout to be configured (AAD doesn't). We can either call out this nuance, or document that apps should configure front-channel logout (which is a good idea, regardless).

pkanher617 (approved):

PR #3981, opened 2021-08-18
Emit event when account is added or removed in a different tab

### LINKED ISSUE #3969: MsalAuthenticationTemplate is rendering children despite having logged out in other tab
### Core Library

### Core Library Version

2.16.1

### Wrapper Library

### Wrapper Library Version

1.0.1

### Description

MsalAuthenticationTemplate does not get reevaluated upon reoping the browser tab containing that component.
This is a problem for NextJS applications using the SWR-Library with an MSAL-Access-Token to fetch data, as data is refetched upon focus.

### Error Message

_No response_

### Msal Logs

_No response_

### MSAL Configuration

```javascript
{
  auth: {
    clientId: `${process.env.NEXT_PUBLIC_AUTH_CLIENT_ID}`,
    authority: MSAL_AUTHORITY_SIGNIN_SIGNUP,
    knownAuthorities: [`${process.env.NEXT_PUBLIC_AUTH_TENANT_URL}`],
    redirectUri: process.env.NEXT_PUBLIC_AUTH_REDIRECT_URL,
  },
  cache: {
  },
};
```

### Relevant Code Snippets

```javascript
<MsalAuthenticationTemplate
    interactionType={InteractionType.Redirect}
    authenticationRequest={{
        scopes: SCOPES,
        redirectUri: MSAL_CONFIG.auth.redirectUri,
        extraQueryParameters: { ui_locales: locale },
        prompt: 'select_account',
    }}
    errorComponent={x}
    loadingComponent={y}
>
    {children}
</MsalAuthenticationTemplate>
```

### Reproduction Steps

Given a React or Next.JS Application similar to the sample https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev/samples/msal-react-samples/nextjs-sample

1.) Open the browser
2.) Open a page / route containing `MsalAuthenticationTemplate` using `InteractionType.Redirect`
3.) Login & await Login to be successful
4.) Open a new tab and open the same page / route as in the first tab
=> You are immediately logged in in the second tab due to `cacheLocation: 'localStorage'`

5.) Logout in tab 2 & await Logout to be completed
6.) Switch to tab 1
=> MsalAuthenticationTemplate is still rendering the child
=> useIsAuthenticated() hook is still returning true

### Expected Behavior

=> The children of MsalAuthenticationTemplate are not rendered as user logged out in tab 2 and access token is shared through local storage.
=> useIsAuthenticated() hook returns false

### Identity Provider

Azure B2C Custom Policy

Chrome

### Regression

_No response_

### Source

External (Customer)

tnorling:
If you have 2 browser tabs opened to the same app and log in or out on one tab, the other tab does not update until you refresh the page. This PR improves the multi-tab experience by introducing a new event emitted when an account is added to or removed from localStorage in a different browsing context which can be used to update UI. 

`msal-react` has also been updated to update account state when these new events are emitted, enabling hooks and components to re-render with the new state.

Fixes #3969


lib/msal-browser/src/event/EventHandler.ts
```diff
@@ -71,4 +81,31 @@ export class EventHandler {
...
+            }
+            const parsedValue = JSON.parse(cacheValue);
+            if (typeof parsedValue !== "object" || !AccountEntity.isAccountEntity(parsedValue)) {
+                return;
+            }
+            const accountEntity = CacheManager.toObject<AccountEntity>(new AccountEntity(), parsedValue);
+            const accountInfo = accountEntity.getAccountInfo();
+            if (!e.oldValue && e.newValue) {
+                this.logger.info("Account was added to cache in a different window");
+                this.emitEvent(EventType.ACCOUNT_ADDED, undefined, accountInfo);
```
tnorling, line 101: We can also emit this event whenever accounts are added in the same browsing context, though I thought it might be redundant with the `LOGIN_SUCCESS` event. Open to thoughts

lib/msal-browser/src/event/EventHandler.ts
```diff
@@ -26,6 +28,10 @@ export class EventHandler {
      */
     addEventCallback(callback: EventCallbackFunction): string | null {
         if (typeof window !== "undefined") {
+            if (this.eventCallbacks.size === 0) {
```
jasonnutter, line 31: Is there somewhere else we can add the even listener? This feels kinda hacky and brittle.

lib/msal-browser/src/event/EventHandler.ts
tnorling: The event would not be emitted for that use case with the current implementation. But if we think that would be useful we can add it, I've been going back and forth on it myself so I'm very open to arguments for or against

lib/msal-browser/src/event/EventHandler.ts
tnorling: That still doesn't give us a way to remove it without checking the callback length. I'd also be open to just making this opt-in or out as sort of a compromise between the two options

tnorling: @jasonnutter Changed the approach here to be opt-in, putting the onus on the developer to enable/disable these events if and when they want them

lib/msal-browser/src/event/EventHandler.ts
```diff
@@ -13,11 +13,15 @@ export class EventHandler {
...
     private browserCrypto: ICrypto;
+    private listeningToStorageEvents: boolean;
+    private boundStorageListener: (e: StorageEvent) => void;
 
     constructor(logger: Logger, browserCrypto: ICrypto) {
         this.eventCallbacks = new Map();
         this.logger = logger;
         this.browserCrypto = browserCrypto;
+        this.listeningToStorageEvents = false;
+        this.boundStorageListener = this.handleAccountCacheChange.bind(this);
```
jasonnutter, line 24: Can the name for these be closer? It is acceptable/idiomatic in JS to do this:

```suggestion
        this.handleAccountCacheChange = this.handleAccountCacheChange.bind(this);
```

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -505,6 +505,20 @@ export abstract class ClientApplication {
         this.eventHandler.removeEventCallback(callbackId);
     }
 
+    /**
+     * Adds event listener that emits an event when a user account is added or removed from localstorage in a different browser tab or window
+     */
+    addAccountStorageListener(): void {
```
jasonnutter, line 511: This name makes it seem like you should be able to add a callback, which you can't.

lib/msal-browser/src/event/EventHandler.ts
```diff
@@ -45,6 +49,28 @@ export class EventHandler {
...
 
+    /**
+     * Adds event listener that emits an event when a user account is added or removed from localstorage in a different browser tab or window
+     */
+    addAccountStorageListener(): void {
+        if (!this.listeningToStorageEvents) {
+            this.logger.verbose("Adding account storage listener.");
+            this.listeningToStorageEvents = true;
+            window.addEventListener("storage", this.boundStorageListener);
+        }
```
jasonnutter, line 60: Can you add log statements in the `else` for both of these?

lib/msal-browser/src/event/EventHandler.ts
```diff
@@ -45,6 +49,28 @@ export class EventHandler {
...
     }
 
+    /**
+     * Adds event listener that emits an event when a user account is added or removed from localstorage in a different browser tab or window
+     */
+    addAccountStorageListener(): void {
+        if (!this.listeningToStorageEvents) {
+            this.logger.verbose("Adding account storage listener.");
+            this.listeningToStorageEvents = true;
+            window.addEventListener("storage", this.boundStorageListener);
```
jasonnutter, line 59: I think it might be good to check for `window` truthiness here and either do nothing, or return a dedicated error if its not available.

lib/msal-browser/src/app/ClientApplication.ts
tnorling: Changed to `enableAccountStorageEvents` and `disableAccountStorageEvents`

lib/msal-browser/src/event/EventHandler.ts
tnorling: Done

lib/msal-browser/src/event/EventHandler.ts
tnorling: Done

lib/msal-browser/src/event/EventHandler.ts
tnorling: Done

lib/msal-browser/src/event/EventHandler.ts
pkanher617: it sounds fine to add it for every account addition, it allows for more granular control over actions related to account CRUD.

jo-arroyo: Given loading external tokens is a testing-only use case at the moment, we could add this later if it proves necessary.

jo-arroyo (approved):
LGTM!

lib/msal-browser/src/event/EventHandler.ts
sameerag: Where did we land here?

sameerag (approved):
One comment. Approved.

pkanher617 (approved):

lib/msal-browser/src/event/EventHandler.ts
tnorling: We'll add the event for same-context account additions incrementally, if needed. Until then people can achieve the same result by listening to the `LOGIN_SUCCESS` or `LOGOUT_SUCCESS` events.

PR #3982, opened 2021-08-18
Fix front channel logout

### LINKED ISSUE #3913: Azure B2C Single Sign Out/Front channel Logout URL Supported?
### Core Library

### Core Library Version

2.16

### Wrapper Library

Not Applicable

### Wrapper Library Version

None

### Description

For our project, we have made the necessary changes to the Azure B2C policy to support Single Sign Out and setup the Front channel logout URL. When we log out, we can trace and see both our server based (ASP.NET) and SPA (Angular using msal.js) application front channel URL is being called. The end result the user is logged out of the ASP.NET application but does not get logged out of the SPA (Angular) application.

Has anyone gotten Single Sign out working with Azure B2C and successfully gotten msal.js to logout when the Front end logout URL is called?

### MSAL Configuration

_No response_

### Relevant Code Snippets

_No response_

### Identity Provider

Azure B2C Custom Policy

### Source

External (Customer)

tnorling:
Currently when calling logout we create the AuthCodeClient and construct the logout url before clearing the cache. This is problematic for front channel logout scenarios because it means a network request to the `./well-known/openid-configuration` endpoint is made before the cache is cleared, potentially resulting in the front channel logout iframe to close before the cache is cleared. 

This PR moves the cache clearing logic into `BaseInteractionClient` from `AuthorizationCodeClient` so that it can be called before the AuthCodeClient is initialized.

Fixes #3913


change/@azure-msal-common-3f6577bb-a98b-48b2-88bb-28e4bb4bdd9b.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
```
tnorling, line 2: Removed function is new and has not been released yet

jasonnutter (approved):

sameerag (approved):

PR #3983, opened 2021-08-19
Address security and linting for passport-azure-ad

jo-arroyo:
This PR adds linting and updates dependencies for the passport-azure-ad library. 

- Upgrades `grunt-contrib-nodeunit`
- Upgrades `https-proxy-agent`, addressing issue #3914 
- Upgrades `lodash` in the package-lock, addressing security advisory [1673](https://www.npmjs.com/advisories/1673)
- Upgrades `path-parse` in the package-lock, addressing security advisory [1773](https://www.npmjs.com/advisories/1773)
- Upgrades `diff` in the package-lock, addressing security advisory [1631](https://www.npmjs.com/advisories/1631)
- Adds `eslint`, `eslint-plugin-header`, `eslint-plugin-security` and lint rules in line with the rest of MSAL.js libraries. Lint errors and warnings addressed.

PR #3984, opened 2021-08-19
Upgrade https-proxy-agent

bjornharrtell:
This PR upgrades the https-proxy-agent dependency.

The old version of https-proxy-agent brings in an old version of agent-base which has several serious issues that has since been fixed. Some of the most important ones can break built in https API for example, see https://github.com/TooTallNate/node-agent-base/pull/47.


tnorling:
@bjornharrtell Thanks, this will be addressed in #3983. Closing as duplicate

PR #3986, opened 2021-08-19
add distributed token cache to ExpressTestApp sample

### LINKED ISSUE #2828: Redis token cache for msal-node?
## Library
- [x] `@azure/msal-node@1.x.x`

## Description

Does a redis-backed token cache exist for msal-node (for use in confidential client application)? If not, could you point me to an existing implementation in another framework/language that would serve as a good reference? I would try to port it to js. I'm looking at the dotnet library's equivalent as a potential candidate.

## Source

- [x] Internal (Microsoft)

derisen:
This adds distributed caching support to the ExpressTestApp sample using Redis as persistence client  (via [node-redis](https://github.com/NodeRedis/node-redis)).

The custom cache plugin implements the [ICachePlugin](https://azuread.github.io/microsoft-authentication-library-for-js/ref/interfaces/_azure_msal_common.icacheplugin.html). The persistent cache here is a simple key-value store. For each user, there are 2 records:

| Key  | Value |
|------|-------|
| `sessionId`       |    session variable, which contains `homeAccountId`       |
| `homeAccountId`       |    msal's cache blob for that user       |

The way this works is roughly as follows: 

* A custom middleware `initializeTokenCachePlugin` is used pass the session variable to the custom cache plugin. 
* When a user attempts to sign or acquire a token, cache access is triggered:
           * Before cache access, the plugin checks the persistence store for the given `sessionId`. If a record is found, it grabs the `homeAccountId` from the session record, and then gets the relevant cache blob using `homeAccountId` as key. The blob then is deserialized into msal's in-memory cache.
            * After cache access, the plugin checks if the cache has changed (in-memory vs persistence). If so, it serializes the in-memory cache back to persistence store.
* When the user signs-out, the session is destroyed, removing the session data but not token cache for that user.

Couple of observations:

* My intention was to implement this in TypeScript on the wrapper side, as I think this is best abstracted away (ideally, devs should only need supply the persistence client, using whatever persistence solution they like, and the session variable). However, because the initial cachePlugin needs to be re-initialized later on with session variable, and because `tokenCache` is a private member of `ClientApplication`, this leads to compile errors. 
* During the instantiation of msal client, if cachePlugin field is left as null in the configuration object, it cannot be initialized later on i.e. it effectively sets the msal client to use the in-memory cache only. 
* I was expecting `acquireTokenSilent` to trigger a cache access, (so that `beforeCacheAccess`/`afterCacheAccess` runs). This wasn't the case, and I had to call `getAccountByHomeId` to do the job. This seems off, the dev might be storing the account object in a session or etc. so might not necessarily call this API.


samuelkubai (approved):
Looks good to me, sample works as expected and a great starting point for distributed caching.

samples/msal-node-samples/ExpressTestApp/TestApp/App/utils/cachePlugin.js
```diff
@@ -1,38 +1,58 @@
...
+                
+                // express session ids start with the string "sess:"
+                persistenceHelper.get("sess:" + sessionId, (err, sessionData) => {
+                    if (err) {
+                        console.log(err);
+                        reject();
+                    }
+
+                    if (sessionData) {
+                        persistenceHelper.get(JSON.parse(sessionData).account.homeAccountId, (err, cacheData) => {
```
tnorling, line 20: Probably should wrap the `JSON.parse` in a `try/catch`

samples/msal-node-samples/ExpressTestApp/TestApp/App/utils/cachePlugin.js
```diff
@@ -1,38 +1,58 @@
...
+            });
+        },
+        afterCacheAccess: async (cacheContext) => {
+            return new Promise((resolve, reject) => {
+
+                if (cacheContext.cacheHasChanged) {
+                    const kvStore = cacheContext.tokenCache.getKVStore();
+
+                    // getting homeAccountId from account entity in kvStore
+                    const homeAccountId = Object.values(kvStore)[1]["homeAccountId"];
```
tnorling, line 42: We should validate that `kvStore` even has any values and then validate that `homeAccountId` is defined before moving on to the next part

samples/msal-node-samples/ExpressTestApp/package-lock.json
```diff
@@ -1,2260 +0,0 @@
-{
```
tnorling, line 1: You can add an `.npmrc` file with `package-lock=false` to this directory to prevent package-locks from being generated in the future.

samples/msal-node-samples/ExpressTestApp/package.json
```diff
@@ -21,7 +21,7 @@
     "typescript": "^4.0.5"
   },
   "dependencies": {
-    "@azure/msal-node": "^1.1.0",
+    "@azure/msal-node": "file:../../../lib/msal-node",
```
tnorling, line 24: Let's leave this versioned. You can add scripts to help going back and forth between local and published versions. Some of our other samples demonstrate this.

tnorling (commented):
Just a couple nits but otherwise looks good

jasonnutter (approved):

tnorling (approved):

sameerag (approved):

PR #3995, opened 2021-08-20
Fix infinite reload on code route in angular

### LINKED ISSUE #3928: Infinite reload loop when loading old/empty/expired #code
### Core Library

### Core Library Version

2.16.1

### Wrapper Library

### Wrapper Library Version

2.0.2

### Description

When loading my application with a faulty `#code` in the url, some handler in MSAL constantly redirects my application to base url `"/"` without removing `window.location.hash`.

### Error Message

There is no error, but the app redirects constantly, generating a series of NavigateStart, End, etc events.

### Msal Logs

There are no console logs generated.

### MSAL Configuration

```javascript
"msal": {
    "client_id": "f10bf1a3-...",
    "authority": "https://login.microsoftonline.com/<company>.onmicrosoft.com",
    "redirect_uri": "https://localhost:4200",
    "post_logout_redirect_uri": "https://localhost:4200/logged-out",
    "protected_resources": [
      {
        "uri": "https://graph.microsoft.com/v1.0/me",
        "scopes": ["user.read", "openid", "profile"]
      },
      {
        "uri": "https://localhost:5001",
        "scopes": ["api://d895bb87-.../.default"]
      }
    ]
  },
```

### Relevant Code Snippets

```javascript
#Setup MSAL according to documentation including interceptor and automatic login redirect to authenticate.
```

### Reproduction Steps

Setup MSAL according to documentation, then navigate to application with known incorrect `#code` in the url.

### Expected Behavior

App removes #code and silently redirects like normal (normal meaning redirect when authenticating successfully).

### Identity Provider

Azure AD / MSA

### Regression

_No response_

### Source

External (Customer)

tnorling:
When using `msal-angular` if you navigate to a route that is protected by the MsalGuard and `code=` appears in the url it can result in an infinite reload loop. The MsalGuard is invoked before the route is actually changed and currently the logic checks the **current** path rather than the **desired** path for `code=`. This results in an attempted navigation to the current route without the `code=` where the Guard is invoked again with no change to the **current** url resulting in a loop. This PR fixes this behavior by making the code logic check the **destination** path for `code=` instead of current. 

Fixes #3928

### DIFF
--- a/lib/msal-angular/src/msal.guard.ts
+++ b/lib/msal-angular/src/msal.guard.ts
@@ -145,7 +145,7 @@ export class MsalGuard implements CanActivate, CanActivateChild, CanLoad {
                     this.authService.getLogger().verbose("Guard - at least 1 account exists, can activate or load");
 
                     // Prevent navigating the app to /#code= or /code=
-                    if (state && currentPath.indexOf("code=")> -1) {
+                    if (state && state.url.indexOf("code=") > -1) {
                         this.authService.getLogger().info("Guard - Hash contains known code response, stopping navigation.");
                         
                         // Path routing (navigate to current path without hash)
--- a/change/@azure-msal-angular-2b3afef7-5cf3-4ebc-9156-3f18fa1cbbca.json
+++ b/change/@azure-msal-angular-2b3afef7-5cf3-4ebc-9156-3f18fa1cbbca.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix infinite reload on route protected by MsalGuard when '#code=' appears in the url #3995",
+  "packageName": "@azure/msal-angular",
+  "email": "thomas.norling@microsoft.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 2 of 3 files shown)


lib/msal-angular/src/msal.guard.ts
```diff
@@ -145,7 +145,7 @@ export class MsalGuard implements CanActivate, CanActivateChild, CanLoad {
                     this.authService.getLogger().verbose("Guard - at least 1 account exists, can activate or load");
 
                     // Prevent navigating the app to /#code= or /code=
-                    if (state && currentPath.indexOf("code=")> -1) {
+                    if (state && state.url.indexOf("code=") > -1) {
                         this.authService.getLogger().info("Guard - Hash contains known code response, stopping navigation.");
                         
                         // Path routing (navigate to current path without hash)
```
jasonnutter, line 151: Does the line below this also needs to be changed?

tnorling: It can be but doesn't need to be. It's only checking if we're using path or hash routing

jasonnutter (approved):

jo-arroyo (approved):

PR #4003, opened 2021-08-26
Add msal-node token caching guidance

derisen:
This adds a guide on msal-node token cache


lib/msal-node/docs/caching.md
```diff
@@ -0,0 +1,109 @@
+# Token caching in MSAL Node
+
+When MSAL Node acquires a token, it caches it in memory for future usage. MSAL Node manages the token lifetime and refreshing for you. APIs like `acquireTokenSilent()` retrieves the access tokens from the cache for a given account:
+
+```javascript
+const msal = require('@azure/msal-node');
+
+// Create msal application object
+const cca = new msal.ConfidentialClientApplication({
```
peterzenz, line 9: Please add a comment on where a user can go to get their clientId, tenant info, and clientSecret.  Something that hints what the format would look like for the actual data might be useful for users as well.
clientId:  "your client ID goes here",  // Example "00000..."
Gives the reader an idea whether they have the right format for the items in question.

lib/msal-node/docs/caching.md
```diff
@@ -0,0 +1,109 @@
...
+};
+
+cca.acquireTokenSilent(silentTokenRequest).then((response) => {
+    // do something with response
+}).catch((error) => {
+    // catch and handle errors
+});
+```
+
+In-memory token cache is not suitable for production. In production, you should serialize and persist the token cache. Depending on the type of application, you have several alternatives:
```
peterzenz, line 34: Redis was in-memory, no?  Are we really recommending folks persist tokens in a DB?  I wouldn't have suggested that.  Maybe we should state that a single, per-server cache instance isn't suitable for a cloud environment with many servers?  Maybe we should suggest they store cached tokens in an ephemeral, in-memory cache like Redis that can be distributed across all their servers.  @sameerag and @samuelkubai, did you have other thoughts on this?

lib/msal-node/docs/caching.md
derisen: added!

lib/msal-node/docs/caching.md
derisen: Technically yes : ) I've added a qualifier.

It seems using a DB for caching is common ([see](https://docs.microsoft.com/en-us/aspnet/core/performance/caching/distributed?view=aspnetcore-5.0#distributed-memory-cache)). It also seems common to use Redis-like caches and DBs in tandem, as different layers of persistence.

lib/msal-node/docs/caching.md
```diff
@@ -0,0 +1,115 @@
...
+};
+
+cca.acquireTokenSilent(silentTokenRequest).then((response) => {
+    // do something with response
+}).catch((error) => {
+    // catch and handle errors
+});
+```
+
+MSAL's in-memory token cache is not suitable for production. In production, you should serialize and persist the token cache. Depending on the type of application, you have several alternatives:
```
sameerag, line 36: We should call out encryption too with persistence.

sameerag (approved):
Looks good. Minor comments.

lib/msal-node/docs/caching.md
sameerag: Well ... Couple of points:
* We should call out in-memory as not scalable rather than not suitable to production. For public clients, there can be users who do not want to ever persist their cache
* DB discussion depends on who we are talking to. I would highlight encryption or the necessity of it in all these use cases.

lib/msal-node/docs/caching.md
peterzenz: Where was this added?

derisen: weird.. @peterzenz see [latest commit](https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/4003/commits/c47484339c866520193e2d325d38875d17f43ac8)

lib/msal-node/docs/caching.md
derisen: got it, will update

lib/msal-node/docs/caching.md
peterzenz: Helpful for the home account id below, but  I don't see anything for clientID, tenant info and clientSecret.  Am I missing something?

derisen: My bad, i was fixated on home account id and missed your point. Updated now.

lib/msal-node/docs/caching.md
derisen: @sameerag revised

lib/msal-node/docs/caching.md
derisen: added 👍

peterzenz (approved):
I'm still largely against using a persistent cache for token material and would recommend against it.  I thought @sameerag had a comment about using encryption if they want to do this.  If folks really want a persistent cache then they should definitely be using something like BitLocker on the drive or some other full disc encryption.  Otherwise they're storing tokens on disc which is a practice we don't recommend at all.

derisen:
> I'm still largely against using a persistent cache for token material and would recommend against it. I thought @sameerag had a comment about using encryption if they want to do this. If folks really want a persistent cache then they should definitely be using something like BitLocker on the drive or some other full disc encryption. Otherwise they're storing tokens on disc which is a practice we don't recommend at all.

@peterzenz Added a remark, though this doesn't get much into details:

:warning: We strongly recommend to encrypt the token cache when persisting it on disk. Again, for public client apps, this is offered out-of-box with [MSAL Node Extensions](../../../extensions/msal-node-extensions/README.md). For confidential clients however, you are responsible for devising an appropriate encryption solution.

peterzenz:
That sounds good – thank you!

PR #4007, opened 2021-08-27
Add retry to E2E tests

tnorling:
Adds a retry to the E2E tests to help alleviate intermittent timeout issues. Each test can be retried individually if a failure occurs


derisen (approved):

peterzenz (approved):
Changes look fine, but what error is the retry addressing?  Is there a deeper root cause we need to dig into?

tnorling:
> Changes look fine, but what error is the retry addressing? Is there a deeper root cause we need to dig into?

99% of the time when tests fail unexpectedly its due to a timeout when entering credentials on AAD. It can happen if the sign in page takes too long to load or if the test tries to enter the credentials while the animation (when the password box slides in in place of the username) is still happening. Unfortunately, short of putting in hard wait times between steps I haven't found a way to make it succeed 100% of the time :/

jasonnutter (approved):

sameerag (approved):

PR #4014, opened 2021-08-31
Update in-progress temporary cache in v1

tnorling:
Today in both msal.js v1 and v2 the in-progress temporary cache entry is bound to the clientId which invoked the interaction. This means that if your page is running multiple instances of msal interactive requests will not be blocked by ongoing interactive requests by another instance. This is particularly problematic when using redirects as only 1 redirect can happen at a given time.

This PR, along with #4015, makes several changes to make the multi-instance interaction scenario more robust:

- Moves temporary cache to `sessionStorage` for parity with v2 (for now only the interaction in progress cache item is moved)
- Removes `clientId` from the interaction in progress cache key
- Replaces the interaction in progress cache value with the `clientId` that invoked interaction
- Preflight validation checks if **any** clientId has interaction in progress
- Redirect response handling checks if **this** clientId has interaction in progress

The cache key/value will be standard (`msal.interaction.status`) for both v1 and v2 to help with migration efforts.


jasonnutter (approved):

lib/msal-core/src/UserAgentApplication.ts
```diff
@@ -2158,7 +2158,7 @@ export class UserAgentApplication {
      * @returns {boolean} true/false
      */
     public getLoginInProgress(): boolean {
-        return this.cacheStorage.getItem(TemporaryCacheKeys.INTERACTION_STATUS) === Constants.inProgress;
+        return this.cacheStorage.isInteractionInProgress(true);
```
sameerag, line 2161: Can this be param less `isInteractionInProgress()`?

tnorling: No, in some scenarios we need to know if **any** interaction is in progress and in others we need to know if **this clientId** has interaction in progress. This flag denotes which we care about.

sameerag: Okay. The name confused me a little. Makes sense now.

sameerag (approved):
lgtm

PR #4027, opened 2021-09-02
Populate msal v2 loginHint from cached msal v1 id token

jasonnutter:
Today, when `loginHint` isn't set, we'll look for a cached ADAL id token, and use it to populate the `login_hint` parameter if one is found. It has been requested that we always perform similar SSO for MSAL v1 id tokens.

### DIFF
--- a/change/@azure-msal-browser-8f6a38ac-ffe9-4d7d-af8e-873ea9411e80.json
+++ b/change/@azure-msal-browser-8f6a38ac-ffe9-4d7d-af8e-873ea9411e80.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Populate msal v2 loginHint from cached msal v1 id token",
+  "comment": "Populate msal v2 loginHint from cached msal v1 id token #4027",
   "packageName": "@azure/msal-browser",
   "email": "janutter@microsoft.com",
   "dependentChangeType": "patch"
--- a/change/@azure-msal-browser-8f6a38ac-ffe9-4d7d-af8e-873ea9411e80.json
+++ b/change/@azure-msal-browser-8f6a38ac-ffe9-4d7d-af8e-873ea9411e80.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Populate msal v2 loginHint from cached msal v1 id token",
+  "packageName": "@azure/msal-browser",
+  "email": "janutter@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
+++ b/lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
@@ -270,23 +270,35 @@ export abstract class StandardInteractionClient extends BaseInteractionClient {
             validatedRequest.account = account;
         }
 
-        // Check for ADAL SSO
+        // Check for ADAL/MSAL v1 SSO
         if (StringUtils.isEmpty(validatedRequest.loginHint)) {
-            // Only check for adal token if no SSO params are being used
+            // Only check for adal/msal token if no SSO params are being used
             const adalIdTokenString = this.browserStorage.getTemporaryCache(PersistentCacheKeys.ADAL_ID_TOKEN);
             if (adalIdTokenString) {
-                const adalIdToken = new IdToken(adalIdTokenString, this.browserCrypto);
+                this.logger.verbose("Cached ADAL id token retrieved.");
+            }
+
+            // Check for cached MSAL v1 id token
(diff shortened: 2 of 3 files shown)


lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
```diff
@@ -270,23 +270,35 @@ export abstract class StandardInteractionClient extends BaseInteractionClient {
             validatedRequest.account = account;
         }
 
-        // Check for ADAL SSO
+        // Check for ADAL/MSAL v1 SSO
         if (StringUtils.isEmpty(validatedRequest.loginHint)) {
```
tnorling, line 274: Does this need to be called if `account` is provided above?

hectormmg (approved):
Lgtm

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
```diff
@@ -270,23 +270,35 @@ export abstract class StandardInteractionClient extends BaseInteractionClient {
...
+            // Check for cached MSAL v1 id token
+            const msalIdTokenString = this.browserStorage.getTemporaryCache(PersistentCacheKeys.ID_TOKEN, true);
+            if (msalIdTokenString) {
+                this.logger.verbose("Cached MSAL.js v1 id token retrieved");
+            }
+
+            const cachedIdTokenString = msalIdTokenString || adalIdTokenString;
+            if (cachedIdTokenString) {
+                const cachedIdToken = new IdToken(cachedIdTokenString, this.browserCrypto);
                 this.browserStorage.removeItem(PersistentCacheKeys.ADAL_ID_TOKEN);
```
peterzenz, line 290: Maybe only do this if you enter the conditional on line 277 just above.  Possibly the same for line 291 just below.

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
```diff
@@ -270,23 +270,35 @@ export abstract class StandardInteractionClient extends BaseInteractionClient {
...
-                if (adalIdToken.claims && adalIdToken.claims.preferred_username) {
-                    this.logger.verbose("No SSO params used and ADAL token retrieved, setting ADAL preferred_username as loginHint");
-                    validatedRequest.loginHint = adalIdToken.claims.preferred_username;
+                this.browserStorage.removeItem(this.browserStorage.generateCacheKey(PersistentCacheKeys.ID_TOKEN));
```
peterzenz, line 291: Is the result from generateCacheKey a constant in these flows?  Do we need to make this call or can we define a constant?

peterzenz (approved):
Super minor comments.  Consider them nits/optional if you're looking to get this in for the next release.

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
jasonnutter: Good catch, I don't think so? Will check.

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
jasonnutter: It is more complicated, unfortunately. The other browser storage APIs include this logic baked in (configurable via a flag), however, the `removeItem` API does not do that. Probably a fix we should make, but can be done separately, as it would be a larger PR.

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
jasonnutter: Yeah, good idea.

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
jasonnutter: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/de24de1d3a135c106143a6cc9907cc1264f546e0/lib/msal-common/src/client/AuthorizationCodeClient.ts#L322-L354

Looks like we should not be adding this parameter if `account` is set. The downstream will first look for `request.account`, however, it will use the `request.loginHint` before it uses the username from the account for some reason.

I'll update this conditional to check for that, but we should maybe double check the above logic (can be addressed seperately).

change/@azure-msal-browser-8f6a38ac-ffe9-4d7d-af8e-873ea9411e80.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Populate msal v2 loginHint from cached msal v1 id token",
```
jasonnutter, line 3:
```suggestion
  "comment": "Populate msal v2 loginHint from cached msal v1 id token #4027",
```

tnorling (approved):

PR #4049, opened 2021-09-08
Move helper functions to BaseInteractionClient

tnorling:
WAM integration will need to use several helper functions that are currently one level deeper than they need to be. This moves the necessary helpers up to `BaseInteractionClient` so they can be used by the forthcoming `WamInteractionClient`

### DIFF
--- a/lib/msal-browser/src/cache/BrowserCacheManager.ts
+++ b/lib/msal-browser/src/cache/BrowserCacheManager.ts
@@ -3,7 +3,7 @@
  * Licensed under the MIT License.
  */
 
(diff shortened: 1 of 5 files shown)


hectormmg (approved):

lib/msal-browser/src/interaction_client/BaseInteractionClient.ts
```diff
@@ -63,4 +64,65 @@ export abstract class BaseInteractionClient {
...
+
+    /**
+     *
+     * Use to get the redirect uri configured in MSAL or null.
+     * @param requestRedirectUri
+     * @returns Redirect URL
+     *
+     */
+    protected getRedirectUri(requestRedirectUri?: string): string {
+        this.logger.verbose("getRedirectUri called");
```
jasonnutter, line 104: Aside: I think we should start logging these under `trace` instead of `verbose`.

jasonnutter (approved):

lib/msal-browser/src/cache/BrowserCacheManager.ts
```diff
@@ -918,6 +918,43 @@ export class BrowserCacheManager extends CacheManager {
             this.removeItem(key);
         }
     }
+
+    /**
+     * Returns username retrieved from ADAL or MSAL v1 idToken
+     */
+    getLegacyLoginHint(): string | null {
```
jasonnutter, line 925: Should this be added to cachemanager interface?

PR #4052, opened 2021-09-09
[msal-browser] Add support for different key types in IndexedDB storage

hectormmg:
This PR:
- Updates `DatabaseStorage` to handle upgrading tables on a new DB version
- Adds a `KeyStore` type in `CryptoOps` to model a keystore that supports multiple key types using differently-typed `DatabaseStorage` instances.
- Adds `symmetricKeys` and `asymmetricKeys` instances to `CryptoOps.cache` (keystore).
- Adds database deletion logic to `logout` API when no account is specified to allow applications to leverage the `logout` API to remove newer versions of the MSAL keystore IndexedDB database. This is added to enable safe migration to newer MSAL DB versions that can be rolledback by downgrading the MSAL library version itself.
- Adds basic tests to `DatabaseStorage` class.


lib/msal-browser/src/cache/DatabaseStorage.ts
```diff
@@ -60,7 +69,7 @@ export class DatabaseStorage<T>{
      * Retrieves item from IndexedDB instance.
      * @param key 
      */
-    async get(key: string): Promise<T> {
+    async get<T>(tableName: DBTableNames, key: string): Promise<T> {
```
jasonnutter, line 72: If table name is passed in constructor, why does it need to be passed here?

jasonnutter: I would say each databasestorage instance should point to a specific table

hectormmg: The constructor receives an array of (supported) table names to create all the tables MSAL has. The API's receive a single (supported) table name so they know which table to perform the operation on.

hectormmg: >I would say each databasestorage instance should point to a specific table

@jasonnutter what do you think the benefit there would be? As it stands now, CryptoOps has a single DB object that it treats as "cache". I think it's easier to keep track of a single DB with different tables than to have multiple Databases with a single table each.

jasonnutter: Ah right, misread the code and forgot we'll have a db with multiple tables, as opposed to multiple dbs.

hectormmg: Take the Bound RT use as an example to illustrate:

```js
this.cache = new DatabaseStorage(dbName, [DBTableNames.asymmetricKeys, DBTableNames.symmetricKeys], dbVersion);
const stk_jwk = this.cache.get<CryptoKeyPair>(DBTableNames.asymmetricKeys, stkKid);
const sessionKey = this.cache.get<CryptoKey>(DBTableNames.symmetricKeys, sessionKeyKid);

// Clearing IndexedDB:

this.cache.clear();
```

rather than

```typescript
export type KeyStore = {
   asymmetricKeys: DatabaseStorage<CachedKeyPair>;
   symmetricKeys: DatabaseStorage<CryptoKey>;
}

this.cache: KeyStore = {
    asymmetricKeys: new DatabaseStorage<CachedKeyPair>(dbName, DBTableNames.asymmetricKeys, ...),
    symmetricKeys: new DatabaseStorage<CryptoKey>(dbName, DBTableNames.symmetricKeys, ...)
};

const stk_jwk = this.cache.asymmetricKeys.get(stkKid);
const sessionKey = this.cache.symmetricKeys.get(sessionKeyKid);

// Clearing IndexedDB
Object.keys(this.cache).forEach((tableName: string) => this.cache[tableName].clear());
```

With the former, we just update the DatabaseStorage APIs and remove the type restriction on the class. To add support for new tables, we just add them to the enum.

With the latter we need to update the CryptoOps constructor every time to add a new DatabaseStorage instance, have a type (i.e. KeyStore) to keep track of the DbStorage objects and then we need to iterate over these objects to clear each table.


I think these approaches are functionally equivalent, but I think the former is closer to what IndexedDB's intended usage is.

jasonnutter: I personally like different `DataBaseStorage` instances for each table. Especially since we know we will be adding main cache support for IDB in the future.

jasonnutter: > I think these approaches are functionally equivalent, but I think the former is closer to what IndexedDB's intended usage is.

True, but we should build the abstractions we want, not necessarily match what we've been given.

hectormmg: Switched implementation to have `DatabaseStorage` instances represent a single table each, all still inside a single IndexedDB database.

jasonnutter (changes requested):
Can you please add unit tests for `DatabaseStorage`?

lib/msal-browser/test/crypto/CryptoOps.spec.ts
```diff
@@ -4,30 +4,46 @@ import { BrowserCrypto } from "../../src/crypto/BrowserCrypto";
...
 
+let mockDatabase = {
+    asymmetricKeys: {},
+    symmetricKeys: {}
+};
+
+// Mock DatabaseStorage
+jest.mock("../../src/cache/DatabaseStorage", () => {
+    return {
+        DatabaseStorage: jest.fn().mockImplementation((tableName: string) => {
```
jasonnutter, line 19: This is great. Looks like its used in a few places, possible to factor out into a helper?

jasonnutter (approved):
One thing, otherwise looks good. Thanks for addressing feedback!

lib/msal-browser/test/crypto/CryptoOps.spec.ts
hectormmg: I can't figure out how to keep the `mockDatabase` object global without having the mocked `DatabaseStorage` object in the same test file. I've also been wondering why we have things like the mocked crypto interface repeated across a bunch of browser tests, I assume it's a similar problem.

lib/msal-browser/src/crypto/CryptoOps.ts
```diff
@@ -122,14 +134,16 @@ export class CryptoOps implements ICrypto {
...
     }
 
     /**
      * Removes all cryptographic keys from IndexedDB storage
      */
     async clearKeystore(): Promise<boolean> {
-        return this.cache.clear();
+        const dataStoreNames = Object.keys(this.cache);
+        const databaseStorage = this.cache[dataStoreNames[0]];
+        return (databaseStorage) ? await databaseStorage.deleteDatabase() : false;
```
tnorling, line 146: Nit: Does this need to be in parens?
```suggestion
        return databaseStorage ? await databaseStorage.deleteDatabase() : false;
```

tnorling (approved):
One nit, otherwise looks good

change/@azure-msal-browser-b8da8826-00fd-454d-bf25-e92d112713ca.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
jasonnutter, line 2:
```suggestion
  "type": "patch",
```

No public API change.

lib/msal-browser/src/crypto/CryptoOps.ts
hectormmg: It doesn't, I'm just used to formatting ternaries like that. Without a logical expression there I guess the parentheses are superfluous.

PR #4055, opened 2021-09-10
Add WamInteractionClient

tnorling:
Adds `WamInteractionClient` class which will be used to acquire a token from WAM


lib/msal-browser/src/interaction_client/WamInteractionClient.ts
```diff
@@ -0,0 +1,174 @@
...
+
+        const validatedRequest: WamTokenRequest = {
+            ...request,
+            clientId: this.config.auth.clientId,
+            authority: canonicalAuthority.urlString,
+            scopes: scopeSet.printScopes(),
+            redirectUri: this.getRedirectUri(request.redirectUri),
+            correlationId: this.correlationId,
+            prompt: request.prompt || PromptValue.NONE,
+            instanceAware: false,
```
tnorling, line 145: Currently an issue with the WAM implementation blocking this feature. Will make this configurable in a follow up PR when that gets resolved

lib/msal-browser/src/cache/BrowserCacheManager.ts
```diff
@@ -453,6 +453,32 @@ export class BrowserCacheManager extends CacheManager {
         });
     }
 
+    /**
+     * Checks the cache for accounts matching loginHint or SID
+     * @param loginHint 
+     * @param sid 
+     */
+    getAccountInfoByHints(loginHint?: string, sid?: string): AccountInfo | null {
```
sameerag, line 461: This is useful beyond WAM-JS bridge?

lib/msal-browser/src/interaction_client/WamInteractionClient.ts
```diff
@@ -0,0 +1,174 @@
...
+     * @param request 
+     * @param reqTimestamp 
+     */
+    protected async handleWamResponse(response: WamResponse, request: WamTokenRequest, reqTimestamp: number): Promise<AuthenticationResult> {
+        this.logger.trace("WamInteractionClient - handleWamResponse called.");
+        // create an idToken object (not entity)
+        const idTokenObj = new AuthToken(response.id_token || Constants.EMPTY_STRING, this.browserCrypto);
+
+        // Save account in browser storage
+        const homeAccountIdentifier = AccountEntity.generateHomeAccountId(response.client_info || Constants.EMPTY_STRING, AuthorityType.Default, this.logger, this.browserCrypto, idTokenObj);
```
sameerag, line 69: AuthorityType.Default does not cover ADFS. Are we not supporting ADFS for WAM?

lib/msal-browser/src/cache/BrowserCacheManager.ts
sameerag: What if two accounts have the same email id - translating to same `loginHint` and no `sid`?

sameerag (approved):
Approved with comments. Please add the error handling -  for more than one account found - for the corner case where loginHint may be the same for two accounts.

lib/msal-browser/src/interaction_client/WamInteractionClient.ts
tnorling: No, can't support direct connection to ADFS. Federated ADFS, however, would be covered by AuthorityType.Default as it goes through AAD

lib/msal-browser/src/cache/BrowserCacheManager.ts
tnorling: Added an error for this case

pkanher617 (approved):

PR #4064, opened 2021-09-15
Call WamInteractionClient from silent and popup APIs

tnorling:
Enables silent and popup APIs to call WAM for tokens. Logout and Redirect will be added in future PRs.


lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -115,6 +120,17 @@ export abstract class ClientApplication {
         this.tokenCache = new TokenCache(this.config, this.browserStorage, this.logger, this.browserCrypto);
     }
 
+    /**
+     * Initializer function to perform async startup tasks such as connecting to WAM extension
+     */
+    async initialize(): Promise<void> {
```
jasonnutter, line 126: We should align this with what we'll be doing for broker. @pkanher617

pkanher617: Once broker is merged we will reconcile the differences. Since broker is 1p only, we may not include this here.

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -241,9 +264,23 @@ export abstract class ClientApplication {
             this.eventHandler.emitEvent(EventType.LOGIN_START, InteractionType.Popup, request);
         }
 
-        const popupClient = new PopupClient(this.config, this.browserStorage, this.browserCrypto, this.logger, this.eventHandler, this.navigationClient, request.correlationId);
+        let result: Promise<AuthenticationResult>;
+
+        if (this.config.system.platformSSO && this.wamExtensionProvider) {
+            result = this.acquireTokenNative(request).catch((e: AuthError) => {
+                if (e instanceof WamAuthError && e.isFatal()) {
+                    this.wamExtensionProvider = undefined; // If extension gets uninstalled during session prevent future requests from continuing to attempt
```
sameerag, line 272: Can an extension ever be re-installed in the same session? I did not think so, but either way we should disable all future requests in the same session anyway.

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -288,15 +325,30 @@ export abstract class ClientApplication {
...
-        } catch (e) {
+        let result: Promise<AuthenticationResult>;
+
+        if (this.config.system.platformSSO && this.wamExtensionProvider) {
+            result = this.acquireTokenNative(request).catch((e: AuthError) => {
+                // If native token acquisition fails for availability reasons fallback to standard flow
+                if (e instanceof WamAuthError && e.isFatal()) {
+                    this.wamExtensionProvider = undefined; // If extension gets uninstalled during session prevent future requests from continuing to attempt 
+                    const silentIframeClient = this.createSilentIframeClient(request.correlationId);
+                    return silentIframeClient.acquireToken(request);
```
sameerag, line 336: can we move this out of the if-else since this seems to be. the default fallback?

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -288,15 +325,30 @@ export abstract class ClientApplication {
...
+                }
+                throw e;
+            });
+        } else {
+            const silentIframeClient = this.createSilentIframeClient(request.correlationId);
+            result = silentIframeClient.acquireToken(request);
+        }
+
+        return result.then((response) => {
+            this.eventHandler.emitEvent(EventType.SSO_SILENT_SUCCESS, InteractionType.Silent, response);
```
sameerag, line 346: Do we want to distinguish WAM API vs Browser API in the event emission?

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -128,19 +128,36 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
+        let result: Promise<AuthenticationResult>;
+        if (this.config.system.platformSSO && this.wamExtensionProvider) {
+            result = this.acquireTokenNative(request).catch((e: AuthError) => {
+                // If native token acquisition fails for availability reasons fallback to standard flow
+                if (e instanceof WamAuthError && e.isFatal()) {
+                    this.wamExtensionProvider = undefined; // Prevent future requests from continuing to attempt 
+                    const silentCacheClient = this.createSilentCacheClient();
+                    const silentRequest = silentCacheClient.initializeSilentRequest(request, account);
+                    result = silentCacheClient.acquireToken(silentRequest).catch(async () => {
+                        return this.acquireTokenByRefreshToken(silentRequest);
```
sameerag, line 142: Same comment as above. Can we move this out of if-else as we are defaulting to this nevertheless?

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -128,19 +128,36 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
+        } else {
+            const silentCacheClient = this.createSilentCacheClient();
+            const silentRequest = silentCacheClient.initializeSilentRequest(request, account);
+            result = silentCacheClient.acquireToken(silentRequest).catch(async () => {
+                return this.acquireTokenByRefreshToken(silentRequest);
+            });
+        }
+
+        return result.then((response) => {
+            this.eventHandler.emitEvent(EventType.ACQUIRE_TOKEN_SUCCESS, InteractionType.Silent, response);
```
sameerag, line 156: same comment on events. Does it make sense to also emit an event for native flow?

lib/msal-browser/src/app/ClientApplication.ts
tnorling: Nope. It can be re-installed but the script won't be loaded back into the page until the page is refreshed.

lib/msal-browser/src/app/ClientApplication.ts
tnorling: Probably not. These are implementation details, whether the token comes from WAM or elsewhere should be transparent to the user. We can add it later if there's reason to.

lib/msal-browser/src/app/ClientApplication.ts
tnorling: I couldn't find a clean way to do this in only one place since there's no early return and adding one would duplicate different code. Lmk if you had something in mind

sameerag: Let's keep it this way in that case.

lib/msal-browser/src/app/ClientApplication.ts
sameerag: As discussed, may be something we can revisit later based on the usage for event logging.

lib/msal-browser/src/app/PublicClientApplication.ts
sameerag: Closing, seems the most optimal at this point.

lib/msal-browser/src/app/PublicClientApplication.ts
sameerag: Future consideration. Closing this for now.

sameerag (approved):
Approved. LGTM. Good job making this cleaner with all the refactors.

pkanher617 (approved):

PR #4066, opened 2021-09-16
[msal-node]: Add E2E regional authority test

samuelkubai:
This pull request adds the end to end test for the regional authority feature. To add this end to end tests we need to:-

- [x] Modify the client credentials sample
- [x] Add necessary regional authorities end to end tests
- [x] Add AAD credentials for the client credentials automated application and tests


samples/msal-node-samples/client-credentials/config/AAD.json
```diff
@@ -0,0 +1,7 @@
+{
+    "authOptions": {
+        "clientId": "88f91eac-c606-4c67-a0e2-a5e8a186854f",
+        "authority": "https://login.microsoftonline.com/f645ad92-e38d-4d1a-b510-d1b09a74a8ca",
+        "clientSecret": "SBriCVu[bQROmT@_B:z9PGYa5G3h5=6j"
```
sameerag: For e2e, can we not read this from a key vault instead? Is that the problem with MSIDLabs?

jasonnutter: Please remove this secret, invalidate it, and instead load a new secret via environment variables.

jasonnutter: Or through keyvault.

samuelkubai: Okay, will refactor this. However this is the case with the other samples as well, so we might need a whole refactor of the e2e tests in this case.

samuelkubai: I have refactored the application to fetch the credentials from the key vault.

hectormmg (approved):
Lgtm!

sameerag (approved):

jasonnutter (approved):

PR #4077, opened 2021-09-20
Add a note for proxy support in adal-node

sameerag:
Add a note that `axios` does not support `proxies` out of the box and the alternative in documentation.


ramya-rao-a (approved):

maintenance/adal-node/README.md
```diff
@@ -179,8 +179,11 @@ context.acquireTokenWithClientCredentials(resource, applicationId, clientSecret,
 });
 ```
 
+### Note on Proxy support for adal-node
+We have moved to `axios` as the package of choice for httpRequests for adal-node from `request` since the npm support for `request` is discontinued. However we noticed later that `axios` does not support proxies - filed [here](https://github.com/axios/axios/issues/2072#issuecomment-567473812). Since we are making only security changes to `adal-node`, we are refraining from making any changes. However, we do recommend using `https-proxy-agent` as suggested in the issue linked in case you need proxies for your application.
```
derisen, line 183:
```suggestion
We have moved to `axios` as the package of choice for HTTP requests for adal-node from `request` since the npm support for `request` is discontinued. However we noticed later that `axios` does not support proxies -filed [here](https://github.com/axios/axios/issues/2072#issuecomment-567473812). Since we are making only security changes to `adal-node`, we are refraining from making any changes. However, we do recommend using `https-proxy-agent` as suggested in the issue linked in case you need proxies for your application.
```

derisen (approved):

samuelkubai (approved):
LGTM

pkanher617 (approved):

PR #4089, opened 2021-09-24
Make no_tokens_found error InteractionRequired

tnorling:
The `no_tokens_found` error is thrown when no refresh token is found in the cache. This error requires interaction (or ssoSilent) to resolve and should be thrown as such. This PR expands the scope of `InteractionRequiredAuthError` to encompass both ServerErrors that require interaction and SDK errors that require interaction to resolve in order to help simplify the fallback logic we expect developers to implement on their silent requests.


lib/msal-common/src/error/InteractionRequiredAuthError.ts
```diff
@@ -23,9 +23,19 @@ export const InteractionRequiredAuthSubErrorMessage = [
...
 
 /**
- * Error thrown when user interaction is required at the auth server.
+ * Interaction required errors defined by the SDK
  */
-export class InteractionRequiredAuthError extends ServerError {
+export const InteractionRequiredAuthErrorMessage = {
+    noTokensFoundError: {
+        code: "no_tokens_found",
+        desc: "No refresh token found in the cache. Please sign-in."
```
jasonnutter, line 31: Do we want to call out which APIs to use? 
```suggestion
        desc: "No refresh token found in the cache. Please sign-in via ssoSilent, loginRedirect, or loginPopup."
```

jasonnutter (approved):

sameerag (approved):
Approved with comments

lib/msal-common/src/error/InteractionRequiredAuthError.ts
tnorling: Might be worth redesigning our errors for v3+ to allow each lib to give more specific error messages. \

> And is it always for refresh tokens only?
Yes. Access tokens missing throws a different error.

PR #4106, opened 2021-09-28
Add AzureCloudInstance to JS libraries

sameerag:
This PR adds convenience strings to sovereign cloud endpoints to make it easy for developers to onboard to a specific cloud.
This translates to adding enums for:

```
    // Microsoft Azure public cloud. Maps to https://login.microsoftonline.com
    AzurePublic,

    // Microsoft Chinese national cloud. Maps to https://login.chinacloudapi.cn
    AzureChina,

    // Microsoft German national cloud ("Black Forest"). Maps to https://login.microsoftonline.de
    AzureGermany,

    // US Government cloud. Maps to https://login.microsoftonline.us
    AzureUsGovernment,

```


lib/msal-common/src/authority/AuthorityOptions.ts
```diff
@@ -12,4 +12,22 @@ export type AuthorityOptions = {
...
     azureRegionConfiguration?: AzureRegionConfiguration;
+    azureCloudInstance?: AzureCloudInstance;
 };
+
+export enum AzureCloudInstance {
+    // AzureCloudInstance is not specified.
+    None,
+
+    // Microsoft Azure public cloud. Maps to https://login.microsoftonline.com
+    AzurePublic,
```
jasonnutter, line 23: Can these be given the value of the domain, so that the `getAzureCloundInstanceUrl` method isn't needed?
```suggestion
    AzurePublic = "https://login.microsoftonline.com",
```

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -38,6 +38,7 @@ export type BrowserAuthOptions = {
     navigateToLoginRequestUrl?: boolean;
     clientCapabilities?: Array<string>;
     protocolMode?: ProtocolMode;
+    azureCloudInstance?: AzureCloudInstance;
```
tnorling, line 41: I'm starting to think we should consider breaking authority related options into a separate config object for v3, there's quite a few now

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -122,7 +123,8 @@ export function buildConfiguration({ auth: userInputAuth, cache: userInputCache,
         postLogoutRedirectUri: "",
         navigateToLoginRequestUrl: true,
         clientCapabilities: [],
-        protocolMode: ProtocolMode.AAD
+        protocolMode: ProtocolMode.AAD,
+        azureCloudInstance: 1
```
tnorling, line 127: Can we use the enum here rather than the index?

lib/msal-common/src/authority/AuthorityOptions.ts
```diff
@@ -12,4 +12,22 @@ export type AuthorityOptions = {
     cloudDiscoveryMetadata: string;
     authorityMetadata: string;
     azureRegionConfiguration?: AzureRegionConfiguration;
+    azureCloudInstance?: AzureCloudInstance;
 };
+
+export enum AzureCloudInstance {
```
jasonnutter, line 18: I don't see this enum exported from Node or Browser. I assume the intent is that customers would use this enum? If so, please reexport (otherwise this doesn't need to be a minor bump).

lib/msal-common/src/authority/AuthorityOptions.ts
```diff
@@ -12,4 +12,22 @@ export type AuthorityOptions = {
...
+    // AzureCloudInstance is not specified.
+    None,
```
jasonnutter, line 20: Don't think this is needed

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -156,6 +158,11 @@ export function buildConfiguration({ auth: userInputAuth, cache: userInputCache,
         allowRedirectInIframe: false
     };
 
+    // build authority string based on auth params - azureCloudInstance is prioritized if provided
+    if (userInputAuth && userInputAuth.azureCloudInstance) {
+        userInputAuth.authority = `${Authority.getAzureCloudInstanceUrl(userInputAuth.azureCloudInstance)}/${Constants.DEFAULT_AUTHORITY_TENANT}/`;
```
tnorling, line 163: Can we do this transformation in the request generation instead?

tnorling: Also is this only supported for `/common`? If so we should validate that the user provided `authority` is also `/common`

jasonnutter:
Does `cloudInstance` need to be a separate property in teh configuration, or could we have it so that you can pass this enum as `authority` (and then augment the authority type to be `string | AzureCloundInstance`?

lib/msal-common/src/utils/Constants.ts
```diff
@@ -11,6 +11,7 @@ export const Constants = {
     // default authority
     DEFAULT_AUTHORITY: "https://login.microsoftonline.com/common/",
     DEFAULT_AUTHORITY_HOST: "login.microsoftonline.com",
+    DEFAULT_AUTHORITY_TENANT: "common",
```
tnorling, line 14: Can we add "common" somewhere in this variable name?

lib/msal-common/src/authority/AuthorityOptions.ts
```diff
@@ -12,4 +12,22 @@ export type AuthorityOptions = {
...
+
+    // Microsoft Azure public cloud. Maps to https://login.microsoftonline.com
+    AzurePublic,
+
+    // Microsoft Chinese national cloud. Maps to https://login.chinacloudapi.cn
+    AzureChina,
```
jasonnutter, line 26:
```suggestion
    AzureChina = "https://login.chinacloudapi.cn",
```

lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -378,15 +390,24 @@ export abstract class ClientApplication {
...
      * @param authorityString - authority from user configuration
      */
-    private async createAuthority(authorityString: string, azureRegionConfiguration?: AzureRegionConfiguration, requestCorrelationId?: string): Promise<Authority> {
+    private async createAuthority(authorityString: string, azureRegionConfiguration?: AzureRegionConfiguration, requestCorrelationId?: string, azureCloudInstance?: AzureCloudInstance): Promise<Authority> {
         this.logger.verbose("createAuthority called", requestCorrelationId);
+
+        // build authority string based on auth params - azureCloudInstance is prioritized if provided
+        let authorityAzureCloudInstance;
+        if (azureCloudInstance) {
+            authorityAzureCloudInstance = `${Authority.getAzureCloudInstanceUrl(azureCloudInstance)}/${Constants.DEFAULT_AUTHORITY_TENANT}/`;
```
tnorling, line 399: we're doing this in a couple different places, can we design this in a way that this only needs to be done once?

lib/msal-common/src/authority/AuthorityOptions.ts
```diff
@@ -12,4 +12,22 @@ export type AuthorityOptions = {
...
+
+    // Microsoft German national cloud ("Black Forest"). Maps to https://login.microsoftonline.de
+    AzureGermany,
```
jasonnutter, line 29:
```suggestion
    AzureGermany = "https://login.microsoftonline.de",
```

lib/msal-common/src/authority/AuthorityOptions.ts
```diff
@@ -12,4 +12,22 @@ export type AuthorityOptions = {
...
+
+    // US Government cloud. Maps to https://login.microsoftonline.us
+    AzureUsGovernment,
```
jasonnutter, line 32:
```suggestion
    AzureUsGovernment = "https://login.microsoftonline.us",
```

lib/msal-common/src/authority/AuthorityOptions.ts
sameerag: Yes. Let me add this.

lib/msal-browser/src/config/Configuration.ts
sameerag: Yes, I discussed this before and I strongly think we need to make an `authority` specific config in v3. We can do a doc and add an ADO item in the backlog.

lib/msal-browser/src/config/Configuration.ts
sameerag: This is mostly because we may not have tenant info if the user provides only an `enum`. I can check the request generation part.

lib/msal-node/src/client/ClientApplication.ts
sameerag: Only two places: Configuration and Request which cannot be avoided. We can do a utils API may be. I optimized it for `Request` such that all `Request Types` can reuse the code.

lib/msal-common/src/authority/AuthorityOptions.ts
sameerag: Not a bad idea. Will do.

sameerag: Actually no, this will make it a "String" enum and that may change the MSAL consistency definition.

lib/msal-browser/src/config/Configuration.ts
sameerag: Done

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -122,7 +123,8 @@ export function buildConfiguration({ auth: userInputAuth, cache: userInputCache,
         postLogoutRedirectUri: "",
         navigateToLoginRequestUrl: true,
         clientCapabilities: [],
-        protocolMode: ProtocolMode.AAD
+        protocolMode: ProtocolMode.AAD,
+        azureCloudInstance: AzureCloudInstance.None
```
jasonnutter, line 127: I'm still not sure we need a `None` value, as not being defined semantically is the same as `None`.

lib/msal-common/src/authority/AuthorityOptions.ts
jasonnutter: Hmm okay I see this is how dotnet implements it. In practice, in the enum key is for the customer/application, the value is for us, but we can keep this.

lib/msal-browser/src/config/Configuration.ts
jasonnutter: Dotnet has a config option for `tenantId`, which we don't have, and that can be combined with these enum to build the authority: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/9e1a300564efb5ad2d031d5149012c8e47749e3b/src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs#L526-L601

lib/msal-common/src/authority/AuthorityOptions.ts
jasonnutter, line 18: For completeness, should PPE be added to this list?

lib/msal-browser/src/config/Configuration.ts
jasonnutter: Imo, having `authority` and `azureCloudInstance` (and potentially `tenantId`) is strange, and the library would have to be opinionated about either which option takes precedence (or throwing errors if both are provided).

lib/msal-common/src/authority/AuthorityOptions.ts
sameerag: So does iOS. Hence keeping this for consistency.

lib/msal-common/src/authority/AuthorityOptions.ts
sameerag: Same as above, keeping this as is for consistency.

lib/msal-common/src/authority/AuthorityOptions.ts
sameerag: Keeping this as is for MSAL consistency

lib/msal-common/src/authority/AuthorityOptions.ts
sameerag: Not doing this change for MSAL Consistency as discussed above.

lib/msal-common/src/authority/AuthorityOptions.ts
sameerag: PPE is always internal. I do not know if we need to mix official cloud support with Msft PPE.

sameerag: Addressed.

lib/msal-browser/src/config/Configuration.ts
sameerag: This may change the public interface. I have seen changing the type, even if it supports the old type, will break apps, depending on their import of types.

sameerag:
> Does `cloudInstance` need to be a separate property in teh configuration, or could we have it so that you can pass this enum as `authority` (and then augment the authority type to be `string | AzureCloundInstance`?

This may change the public interface. I have seen changing the type, even if it supports the old type, will break apps, depending on their import of types.

lib/msal-browser/src/config/Configuration.ts
sameerag: > Also is this only supported for `/common`? If so we should validate that the user provided `authority` is also `/common`

I do not think so. 

* We do not have a tenant configuration separately in our library.I am not in favor of introducing this in v2 as is since it is over populated already w.r.t configuration options for `authority`. Redesigning `authority interface` in `V3` can be a better option to introduce `tenantId` configurability IMO.  
* User provided `authority` should be able to provide any `tenantId`. Do not intend to change that. 
* `AzureCloudInstance` is intended to just provide convenience for initiating a specific cloud for certain users (especially sovereign clouds) since docs on them seem to be sparse and unclear. Hence made a decision to default the `tenantId` to `common`.

One option is to check if the user provided `authority` matches the user choice for `azureCloudInstance` if provided, and append the `tenantId` if it is provided. This does not fix all use cases and complicated the user interface.

lib/msal-browser/src/config/Configuration.ts
samuelkubai: We should update the in-file documentation of this type too, to include `azureCloudInstance`

samuelkubai (approved):
Left a comment

lib/msal-browser/src/config/Configuration.ts
sameerag: Added a new interface param `AzureAuthOptions` to address the flexibility and restricting `tenant` to `AzureCloudInstance` only.

lib/msal-node/src/client/ClientApplication.ts
sameerag: Addressed this.

lib/msal-common/src/utils/Constants.ts
sameerag: Added

lib/msal-common/src/error/ClientConfigurationError.ts
```diff
@@ -81,6 +81,10 @@ export const ClientConfigurationErrorMessage = {
         code: "untrusted_authority",
         desc: "The provided authority is not a trusted authority. Please include this authority in the knownAuthorities config parameter."
     },
+    invalidAzureCloudInstance: {
+        code: "invalid_azure_cloud_instance",
+        desc: "Invalid AzureCloudInstance provided. Please refer MSAL JS docs for valid values"
```
jasonnutter: This should probably link to relevant docs.

lib/msal-common/src/config/ClientConfiguration.ts
```diff
@@ -121,6 +123,17 @@ export type ClientCredentials = {
...
 
+/**
+ * AzureCloudInstance specific options
+ *
+ * - azureCloudInstance             - string enum providing short notation for soverign and public cloud authorities
+ * - tenant                         - provision to provide the tenant info
+ */
+export type AzureAuthOptions = {
+    azureCloudInstance?: AzureCloudInstance;
+    tenant: string,
```
jasonnutter: I would propose teh following:

```suggestion
    cloudInstance?: AzureCloudInstance;
    tenantId: string,
```

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
```diff
@@ -19,6 +19,7 @@ import { INavigationClient } from "../navigation/INavigationClient";
 import { RedirectRequest } from "../request/RedirectRequest";
 import { PopupRequest } from "../request/PopupRequest";
 import { SsoSilentRequest } from "../request/SsoSilentRequest";
+import { throwStatement } from "@babel/types";
```
jasonnutter:
```suggestion
```

lib/msal-browser/src/cache/TokenCache.ts
```diff
@@ -58,13 +58,21 @@ export class TokenCache implements ITokenCache {
             this.loadAccessToken(request, response, request.account.homeAccountId, request.account.environment, request.account.tenantId, options);
         } else if (request.authority) {
 
+            let authorityAzureCloudInstance;
+            if (request.azureAuthOptions) {
+                const tenant = request.azureAuthOptions.tenant ? request.azureAuthOptions.tenant : Constants.DEFAULT_COMMON_TENANT;
+                authorityAzureCloudInstance = `${request.azureAuthOptions.azureCloudInstance}/${tenant}/`;
+            }
+            // final authority from the request
+            const authorityUrl = authorityAzureCloudInstance ? authorityAzureCloudInstance : request.authority;
```
jasonnutter: Looks like this could go in common, exactly the same as Node.

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -38,14 +38,15 @@ export type BrowserAuthOptions = {
     navigateToLoginRequestUrl?: boolean;
     clientCapabilities?: Array<string>;
     protocolMode?: ProtocolMode;
+    azureAuthOptions?: AzureAuthOptions;
```
jasonnutter: This name feels a bit too generic...is there something more specific? e.g. `azureCloud`

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
sameerag: Yeah, typo. Thanks for noticing this.

lib/msal-browser/src/cache/TokenCache.ts
sameerag: One other option is to make this a part of `authorityOptions` and change the authority url in `Authority.ts`. I added these at the client level because it is easier to make changes and more control to manipulate client specific data regarding authority in future.

Both options have their pros and cons.

lib/msal-browser/src/config/Configuration.ts
sameerag: `azureCloudOptions`?

lib/msal-common/src/config/ClientConfiguration.ts
sameerag: If we plan to add PPE and other clouds, then yes, may be we can make it `cloudInstance`. Let me check with .Net folks if we can expand the definition to give them a heads up (for any future compat asks that can come to them).

lib/msal-browser/src/cache/TokenCache.ts
tnorling: ^ I think doing this in Authority would be cleaner and easier to maintain

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
```diff
@@ -173,22 +173,44 @@ export abstract class StandardInteractionClient extends BaseInteractionClient {
...
+            authorityMetadata: this.config.auth.authorityMetadata,
         };
 
-        if (requestAuthority) {
+        // build authority string based on auth params, precedence - azureCloudInstance + tenant >> authority and request parameters  >> config parameters
+        let userRequestedAuthority;
+        let authorityAzureCloudInstance;
+
+        // Fetch the authority from request if provided
+        if (requestAuthority || requestAzureAuthOptions) {
```
tnorling: All this logic should live in the Authority class imo

tnorling (commented):
Main piece of feedback is that I'd like to see the authority url generation moved to a central place. I'd also like to see docs around this explaining usage, what scenarios are supported and order of precedence when different combinations of options are provided.

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
sameerag: Moved

lib/msal-browser/src/config/Configuration.ts
sameerag: Done

lib/msal-browser/src/config/Configuration.ts
sameerag: Yep. On it.

lib/msal-common/src/config/ClientConfiguration.ts
sameerag: Keeping it `azureCloudInstance` to comply with MSAL compat

lib/msal-browser/src/cache/TokenCache.ts
sameerag: Moved.

change/msal-b56fcb92-f186-43cd-be5f-ed0834c67b12.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add AzureCloudInstance support for JS libraries(#4106)",
```
jasonnutter:
```suggestion
  "comment": "Add AzureCloudInstance support for JS libraries #4106",
```

(for each of the change files)

lib/msal-core/package.json
```diff
@@ -48,6 +48,7 @@
     "prepack": "npm run build"
   },
   "dependencies": {
+    "caniuse-lite": "^1.0.30001285",
```
jasonnutter: Can this be removed?

change/msal-b56fcb92-f186-43cd-be5f-ed0834c67b12.json
jasonnutter: And the only change for `msal-core` in this PR looks to be the `package.json` which can be reverted, so this specific change file can be removed.

lib/msal-core/package.json
sameerag: Done

change/msal-b56fcb92-f186-43cd-be5f-ed0834c67b12.json
sameerag: Agree

lib/msal-browser/src/index.ts
```diff
@@ -72,6 +72,8 @@ export {
     // Utils
     StringUtils,
     UrlString,
+    // AzureCloudInstance enum
+    AzureCloudInstance,
```
tnorling, line 95: Do we want to export `AzureCloudOptions` as well?

lib/msal-common/src/authority/Authority.ts
```diff
@@ -19,6 +19,8 @@ import { CloudInstanceDiscoveryResponse, isCloudInstanceDiscoveryResponse } from
 import { CloudDiscoveryMetadata } from "./CloudDiscoveryMetadata";
 import { RegionDiscovery } from "./RegionDiscovery";
 import { RegionDiscoveryMetadata } from "./RegionDiscoveryMetadata";
+import { AzureCloudOptions } from "../config/ClientConfiguration";
+import { AzureCloudInstance } from "..";
```
tnorling: This will create a circular dependency. Import from the actual file instead

lib/msal-common/src/authority/Authority.ts
```diff
@@ -433,9 +435,25 @@ export class Authority {
...
 
+    /**
+     * helper function to populate the authority based on azureCloudOptions
+     * @param authorityString
+     * @param azureCloudOptions
+     */
+    static generateAuthority(authorityString: string, azureCloudOptions?: AzureCloudOptions): string {
+        let authorityAzureCloudInstance;
+
+        if (azureCloudOptions && azureCloudOptions.azureCloudInstance !== AzureCloudInstance.None) {
```
tnorling, line 445: What if `azureCloudOptions.azureCloudInstance` is not defined?

lib/msal-common/src/authority/AuthorityOptions.ts
```diff
@@ -13,3 +13,20 @@ export type AuthorityOptions = {
     authorityMetadata: string;
     azureRegionConfiguration?: AzureRegionConfiguration;
 };
+
+export enum AzureCloudInstance {
+    // AzureCloudInstance is not specified.
+    None,
```
tnorling, line 19: Does this need a value?

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -122,7 +123,11 @@ export function buildConfiguration({ auth: userInputAuth, cache: userInputCache,
         postLogoutRedirectUri: "",
         navigateToLoginRequestUrl: true,
         clientCapabilities: [],
-        protocolMode: ProtocolMode.AAD
+        protocolMode: ProtocolMode.AAD,
+        azureCloudOptions: {
+            azureCloudInstance: AzureCloudInstance.None,
```
tnorling, line 128: How does the default work if the developer provides `azureCloudOptions` but only provides the `tenant` property? Do we still give `azureCloudInstance` a default value somewhere?

lib/msal-common/src/authority/Authority.ts
sameerag: Typo, correcting

lib/msal-common/src/authority/AuthorityOptions.ts
sameerag: Compliance with MSALs

lib/msal-browser/src/config/Configuration.ts
sameerag: Good point, I don't think so, we default to "none". It should still hold good, as we check for `azureCloudInstance` value instead. Let me add a test though, to check the validity of this.

lib/msal-common/src/authority/Authority.ts
sameerag: It defaults to `None`, so this should be fine.

lib/msal-browser/src/index.ts
sameerag: Yeah, guess we have to. For types. Adding.

lib/msal-common/src/authority/Authority.ts
tnorling: It's an optional parameter. You should either make it required if it always has a value or check that it's defined when you reference it.

sameerag: Guess I can make `azureCloudInstance` mandated and `azureCloudOptions` optional. Will be consistent. Thanks. Good catch.

lib/msal-browser/src/config/Configuration.ts
sameerag: Added the test.

lib/msal-browser/src/index.ts
sameerag: Done, exported in msal-node too.

lib/msal-common/src/authority/Authority.ts
sameerag: Done

lib/msal-common/src/error/ClientConfigurationError.ts
sameerag: Done

lib/msal-browser/FAQ.md
```diff
@@ -239,6 +239,10 @@ For MSAL.js 2.x, please review [this document](https://github.com/AzureAD/micros
 
 The `authority` string that you need to supplant to MSAL app configuration is not explicitly listed among the **Endpoint** links on `Azure Portal/AzureAD/App Registration/Overview` page. It is simply the domain part of a `/token` or `/authorize` endpoint, followed by the tenant name or ID e.g. `https://login.microsoftonline.com/common`.
 
+## What does authority string default to if I provide "authority" and "azureCloudOptions"?
+
+If the developer provides `azureCloudOptions`, we overwrite any value provided in the `authority`. We also make sure we give preference to these options provided in a `request` over what is set when the msal initializes in the `configuration`.
```
tnorling:
```suggestion
If the developer provides `azureCloudOptions`, MSAL.js will overwrite any value provided in the `authority`. MSAL.js will also give preference to parameters provided on a given request over what is set in the `PublicClientApplication` configuration.
```

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
```diff
@@ -211,22 +211,32 @@ export abstract class StandardInteractionClient extends BaseInteractionClient {
...
             protocolMode: this.config.auth.protocolMode,
             knownAuthorities: this.config.auth.knownAuthorities,
             cloudDiscoveryMetadata: this.config.auth.cloudDiscoveryMetadata,
-            authorityMetadata: this.config.auth.authorityMetadata
+            authorityMetadata: this.config.auth.authorityMetadata,
         };
 
-        if (requestAuthority) {
+        // Fetch the authority from request if provided, default if the user authority is not defined
+        const userRequestedAuthority = requestAuthority ? requestAuthority : Constants.DEFAULT_AUTHORITY;
```
tnorling: This should fallback to the config authority (which will be the default authority if that was not provided either)
```suggestion
        const userRequestedAuthority = requestAuthority ? requestAuthority : this.config.auth.authority;
```

sameerag: Edited offline and pushed

lib/msal-browser/FAQ.md
sameerag: Added more changes

tnorling (approved):

PR #4112, opened 2021-09-29
Fix issues with beachball

tnorling:
- Updates beachball to version 2.x (breaking changes are addressed)
- Renames `changelog.md` to `CHANGELOG.md` resolving conflict on release PR
- Fixes issue of msal-node-extensions getting bumped to 1.0.0


.beachballrc
```diff
@@ -2,5 +2,5 @@
     "branch": "dev",
     "access": "public",
     "changehint": "Need to run the following command before merging this PR: npm run beachball:change",
-    "scope": ["lib/**", "extensions/**"]
+    "scope": ["lib/**", "extensions/*"]
```
jasonnutter, line 5: Why can lib use `**` but extensions can only do `*`?

jasonnutter (approved):

.beachballrc
tnorling: We can use `*` for lib too, I'll update it for consistency. I changed extensions because it was picking up the samples as well

lib/msal-browser/README.md
```diff
@@ -9,7 +9,7 @@
 
 1. [About](#about)
 1. [FAQ](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/FAQ.md)
-1. [Changelog](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/changelog.md)
+1. [Changelog](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/CHANGELOG.md)
```
peterzenz, line 12: Is this case sensitive or is there just some convention you're trying to follow?

peterzenz (approved):

lib/msal-browser/README.md
tnorling: It's case sensitive

sameerag (approved):

PR #4118, opened 2021-09-30
Fix node extensions beachball bump

tnorling:
- Disallows change types other than `prerelease` and `none` for msal-node-extensions, preventing unexpected bumps to `1.0.0`
- Disallows major version bumps in all other packages (with the exception of msal-common)


jasonnutter (approved):

jasonnutter:
Would probably be a good idea to add this for every library (namely react ,angular, browser, node), and include "major", so that way major bumps require editing that field as well.

samuelkubai (approved):
LGTM

PR #4123, opened 2021-10-04
Remove user hint requirement on ssoSilent API

pkanher617:
This PR removes the restriction on the ssoSilent API to require an `account`, `login_hint` or `sid` to initiate a request in a hidden iframe and receive a new authorization code. The API will no longer throw an error, but log a warning that there may not be enough information.

### DIFF
--- a/lib/msal-browser/docs/login-user.md
+++ b/lib/msal-browser/docs/login-user.md
@@ -97,7 +97,7 @@ These APIs will return an account object or an array of account objects with the
 
 ## Silent login with ssoSilent()
 
-If you already have a session that exists with the authentication server, you can use the ssoSilent() API to make request for tokens without interaction.
+If you already have a session that exists with the authentication server, you can use the ssoSilent() API to make requests for tokens without interaction.
 
 ### With User Hint
 
--- a/lib/msal-browser/docs/login-user.md
+++ b/lib/msal-browser/docs/login-user.md
@@ -97,7 +97,11 @@ These APIs will return an account object or an array of account objects with the
 
 ## Silent login with ssoSilent()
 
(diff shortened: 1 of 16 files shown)


lib/msal-browser/docs/login-user.md
```diff
@@ -124,6 +128,24 @@ try {
...
+};
+```
+
+However, be aware that if your application has code paths for multiple users in a single browser session, or if the user has multiple accounts for that single browser session, then there is a higher likelihood of silent sign-in errors. You may see the following error show up in the event of multiple account sessions found by the authorization server:
+
+```txt
+InteractionRequiredAuthError: interaction_required: AADSTS16000: Either multiple user identities are available for the current request or selected account is not supported for the scenario.
+```
+
+This indicates that the server could not determine which account to sign into, and will require either one of the parameters above (`account`, `login_hint`, `sid`) or an interactive sign-in to choose the account.
```
jasonnutter, line 147: We need to make this really clear, and should probably show a code sample for how to handle this.

jasonnutter (approved):
One comment, otherwise looks good!

tnorling (approved):

jasonnutter:
There are a couple of docs that should be updated to reflect that we no longer have this restriction:

- https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/926f1c2ba0598575e23dfd8cdd8b79fa3a3d19ff/lib/msal-angular/docs/v2-docs/ssosilent.md
- https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/06351f0cc63f389d9fbb6b973d9f2417c74fd41f/lib/msal-browser/docs/iframe-usage.md
- https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/b091029c3a406342d26a9c99729146c116743676/lib/msal-react/docs/hooks.md#usemsalauthentication-hook

Issue #4122, opened 2021-10-04
msal version as property


PetrSnobelt:
### Core Library

MSAL.js v2 (@azure/msal-browser)

### Wrapper Library

Not Applicable

### Description

Hello, 
for telemetry purposes, it would be nice if MSAL can report its version so we can send it to telemetry etc.

something like `msal.varsion` should be enough.

Thx

### Source

Internal (Microsoft)

tnorling:
Thanks for the suggestion. This just needed to be exported, the PR linked will address this.

PR #4124, opened 2021-10-04
Export library versions

tnorling:
Exports version from each library

Addresses #4122


jasonnutter (approved):

jo-arroyo (approved):

lib/msal-common/src/index.ts
```diff
@@ -93,3 +93,4 @@ export { TimeUtils } from "./utils/TimeUtils";
 export { ServerTelemetryManager } from "./telemetry/server/ServerTelemetryManager";
 export { ServerTelemetryRequest } from "./telemetry/server/ServerTelemetryRequest";
 export { PopTokenGenerator } from "./crypto/PopTokenGenerator";
+export { version } from "./packageMetadata";
```
sameerag, line 96: Do we need this?

tnorling: We might want it later. At worst it's a consistency item, this is a good thing to have available.

sameerag: Alright. Approving.

hectormmg (approved):
Lgtm

PR #4135, opened 2021-10-05
Update cdn-usage.md

AndyGlaister:
Msal will always fail, so pages will always use the fallback. It's 'msal'


jasonnutter:
Good catch, apologies for the confusion!

jasonnutter (approved):

sameerag (approved):

PR #4137, opened 2021-10-05
Automated publish workflow

bmahall:
- npm_publish.yml :- For automation of cdn & npm publish; Github Releases
- typedoc.yml :- Updated


.github/workflows/npm-publish.yml
```diff
@@ -1,27 +1,122 @@
+# NPM Publish is the name for this action
+name: NPM Publish
 
-# This workflow will run tests using node and then publish a package to GitHub Packages when a release is created
-# For more information see: https://help.github.com/actions/language-and-framework-guides/publishing-nodejs-packages
-
-name: npm Publish
-
+# This action will trigger on every successful release
```
tnorling:
```suggestion
# This action will trigger when Release PR from release-staging branch is merged into dev
```

.github/workflows/typedoc.yml
```diff
@@ -1,13 +1,17 @@
...
-  push:
+  pull_request:
+    types: [closed]
     branches:
-      - master  # Set a branch name to trigger deployment
-      - typedoc-action
+      - dev
+  push:
+     branches:
+       - typedoc-action
```
jasonnutter: This was the branch where this action was originally added, so this can be removed. I think any push/PR merge to dev should trigger this action.

tnorling: Would that potentially be confusing if the typedocs reference something that hasn't been released yet?

pkanher617 (approved):

.github/workflows/typedoc.yml
bmahall: Should we update typedocs directly on any push to dev, or could that have other cascading effects as suggested by @tnorling ?

jasonnutter: I'm fine with a push to dev as part of a release (I agree they probably shouldn't be updated until the code is released).

tnorling (approved):

.github/workflows/npm-publish.yml
```diff
@@ -1,27 +1,122 @@
...
+
+      - name: Publish  msal-core
+        if: steps.changes.outputs.core == 'true'
+        working-directory: lib/msal-core
+        env:
+          CDN_EUNO_SAS: ${{ secrets.CDN_EUNO_SAS}}
+          CDN_USWE_SAS: ${{ secrets.CDN_USWE_SAS}}
+        run: |
+          npm install
+          npm run cdn
```
jasonnutter, line 58: The cdn command will modify a couple markdown files that should be committed back to the repo (without trigger another run of pipeliens), is that handled?

jasonnutter: Nevermind, I see the post release step.

.github/workflows/npm-publish.yml
```diff
@@ -1,27 +1,122 @@
...
+      - name: Update package-locks
+        run: npm install
+
+      # Create Pull Request from Post-release to dev
+      - name: Create Pull Request
+        id: pr
+        uses: peter-evans/create-pull-request@v3
+        with:
+          commit-message: Post-release
+          branch: post-release
```
jasonnutter: Can this be dynamic? Reusing the same branch can get weird sometimes.

tnorling: It can be but this should be auto-deleted when the PR is merged and the action also rebases every time it's run so I don't foresee too many issues.

jasonnutter: Gotya...I guess one thing I might be afraid of is someone accidently creating a branch with this name, and so making the branch name dynamic (e.g. post-release-november-2021) might help avoid that.

jasonnutter (approved):

jasonnutter:
Looks good, thanks for addressing feedback, looking forward to this!

PR #4142, opened 2021-10-07
Fix redirect processing when allowRedirectInIframe: true

### LINKED ISSUE #4138: Sign in flow does not complete when running MSAL 2.0 in iframe
### Core Library

### Core Library Version

2.16.1

### Wrapper Library

### Wrapper Library Version

2.0.2

### Description

I am building an Angular based SPA using MSAL-angular 2.0 and integrating with an Azure B2C custom policy.

I have been able to integrate several policies, including sign up, sign in, forgot password, etc. without any issue as long as the angular application is running in it's own browser window.

However, we have a requirement to use 'journey framing' to perform a sign in when this application is loaded in an iframe. Unfortunately, whenever the application is loaded in an iframe, MSAL triggers a sign in flow, completes the B2C portion, but after returning to the SPA, MSAL fails to complete the sign in flow (session storage shows interaction_in_progress and URL bar contains B2C return values).

There is no error output in console or any warning to suggest why MSAL will not proceed to complete the flow using the values passed in the return URL from B2C. This works completely fine and completes the flow if not running in an iframe.

I have noted the requirements and information in this support article: https://docs.microsoft.com/en-us/azure/active-directory-b2c/embedded-login?pivots=b2c-custom-policy

The article does state that "Running MSAL 2.0 in an iframe is not currently supported.". Does this mean that the issue I am observing is considered normal because MSAL is currently not designed to handle this use case? If so, is there any workaround at this stage?

Thanks

### MSAL Configuration

```javascript
{
    auth: {
      clientId: config.clientId,
      authority: config.signInAuthority,
      redirectUri: "/loading",
      postLogoutRedirectUri: "/",
      knownAuthorities: [config.authorityDomain]
    },
    cache: {
      storeAuthStateInCookie: isIE,
    },
    system: {
      allowRedirectInIframe: true,
      loggerOptions: {
        loggerCallback,
        logLevel: LogLevel.Verbose,
        piiLoggingEnabled: true
      }
    }
}
```

### Relevant Code Snippets

_No response_

### Identity Provider

Azure B2C Custom Policy

### Source

External (Customer)

tnorling:
Currently `handleRedirectPromise` verifies that msal is not running in an iframe before navigating to the start page. This does not take into account the `allowRedirectInIframe` flag for certain B2C scenarios, breaking redirect processing when the start page != redirectUri. This PR fixes this behavior.

Also fixes a similar check in the MsalGuard in Angular by exposing the configuration object through a new API on PCA.

Fixes #4138


jasonnutter (approved):

sameerag (approved):

lib/msal-angular/src/msal.service.ts
```diff
@@ -83,5 +84,7 @@ export class MsalService implements IMsalService {
         this.logger = logger.clone(name, version);
         this.instance.setLogger(logger);
     }
-
+    getConfiguration(): BrowserConfiguration {
```
jasonnutter: I'm okay with not including this, given it doesn't add or changing anything, and a component can invoke any of the functions on the instance directly if it wants.

tnorling: Agreed, removed.

jo-arroyo (approved):

PR #4165, opened 2021-10-18
Update axios in samples

jasonnutter:


samples/package.json
```diff
@@ -7,11 +7,11 @@
         "test": "jest"
     },
     "dependencies": {
-        "axios": "^0.21.1",
+        "axios": "^0.21.4",
```
jasonnutter, line 10: @tnorling Is the package-lock needed for this, or can we remove it like we do for other samples?

sameerag: Is this change made only for node in the last build?

tnorling: We use it for dependency caching in the e2e workflows. If we want to remove it we'd have to look into using something else as the cache key

jasonnutter: > Is this change made only for node in the last build?

Not sure what you mean exactly, but msal-node is already on teh right version.

> We use it for dependency caching in the e2e workflows. If we want to remove it we'd have to look into using something else as the cache key

Got it, will keep.

sameerag: Yeah, I remember making this change for msal-node. I did not realize the samples also use `axios`

sameerag (approved):

tnorling (approved):

PR #4172, opened 2021-10-19
Fixed issue with default url

mpminayo:
Changed config file so that it can use default authority when none is provided.


jasonnutter:
Is the issue here when empty string or null is provided as the authority?

jasonnutter:
Also, can you please add a test that demonstrates this is fixed?

lib/msal-node/src/config/Configuration.ts
```diff
@@ -129,7 +129,12 @@ export function buildAppConfiguration({
     system,
 }: Configuration): NodeConfiguration {
     return {
-        auth: { ...DEFAULT_AUTH_OPTIONS, ...auth },
+        auth: { ...DEFAULT_AUTH_OPTIONS, 
+            ...auth , 
+            authority: auth.authority?.trim() 
+                ? auth.authority.trim() 
+                : DEFAULT_AUTH_OPTIONS.authority
```
jasonnutter: Ternary not needed if the condition is also the value:

```suggestion
            authority: auth.authority?.trim() || DEFAULT_AUTH_OPTIONS.authority
```

mpminayo:
> Is the issue here when empty string or null is provided as the authority?

Yes it is.

mpminayo:
> Also, can you please add a test that demonstrates this is fixed?

On it.

samuelkubai (approved):
LGTM

lib/msal-node/test/config/ClientConfiguration.spec.ts
```diff
@@ -69,6 +69,17 @@ describe('ClientConfiguration tests', () => {
         expect(config.auth!.clientId).toEqual(TEST_CONSTANTS.CLIENT_ID);
     });
 
+    test('builds configuration and assigns default authority when empty string provided', () => {
+        const config: Configuration = buildAppConfiguration({
+            auth: {
+                clientId: TEST_CONSTANTS.CLIENT_ID,
+                authority: " "
```
sameerag, line 76: Do we have a test case for `authority=""`?

sameerag (approved):
Approved with comments.

lib/msal-node/test/config/ClientConfiguration.spec.ts
mpminayo: Just added this.

jasonnutter: In common, we have this logic for empty string: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/8c7c0ea26f4f42078e2ef62cebe9a3f6c88dd261/lib/msal-common/src/authority/AuthorityFactory.ts#L50-L52

It seems like today, empty string is treated as invalid use case, as we throw an error. Seems like we should be consistent between all of our libraries about this?

jasonnutter: In other words, it seems like the change to fallback to the default authority should be done in msal-common, as opposed to in MSAL Node?

ghost:

jasonnutter:
Status on this PR? And curious, how often do we get applications accidently providing an empty string for authority?

lib/msal-node/test/config/ClientConfiguration.spec.ts
```diff
@@ -69,6 +69,28 @@ describe('ClientConfiguration tests', () => {
...
```
Robbie-Microsoft, line 76: Should we be creating a constant for " "?

peterzenz:
> Status on this PR? And curious, how often do we get applications accidently providing an empty string for authority?

@mpminayo updates here?

bktowett (approved):

jasonnutter:
Thinking about this, I'm not sure we really need this logic. I'd rather not write code against configuration values that would intuitively not be expected to work.

mpminayo:
From @jasonnutter's comments, I am closing this for now.

PR #4178, opened 2021-10-20
Add support for Ephemeral SSH Certificate Acquisition

hectormmg:
This PR:
- Adds the `SSH` (`ssh-cert`) scheme to the `AuthenticationScheme` enum
- Adds the SSH public key JWK (as `req_cnf`) to the token request body in the `AuthorizationCodeClient` and `RefreshTokenClient`
- Adds validation to make sure SSH JWK and SSH Key ID are present in requests using the `SSH` authentication scheme
- Adds cache matching logic to support silent acquisition of cached SSH certificates based on the request's authentication scheme and the SSH Key ID provided
- Includes unit tests and a new sample for the changes described above


lib/msal-browser/test/utils/StringConstants.ts
```diff
@@ -108,6 +108,15 @@ export const TEST_POP_VALUES = {
     DECODED_REQ_CNF: "{\"kid\":\"NzbLsXh8uDCcd-6MNwXF4W_7noWXFZAfHkxZsRGC9Xs\",\"xms_ksl\":\"sw\"}"
 };
 
+export const TEST_SSH_VALUES = {
```
peterzenz, line 111: Are these fakes, or do they represent valid (but likely unusable) items?

lib/msal-common/test/client/AuthorizationCodeClient.spec.ts
```diff
@@ -1130,6 +1131,209 @@ describe("AuthorizationCodeClient unit tests", () => {
...
+            sinon.stub(AuthorizationCodeClient.prototype, <any>"executePostToTokenEndpoint").resolves(POP_AUTHENTICATION_RESULT);
+            const createTokenRequestBodySpy = sinon.spy(AuthorizationCodeClient.prototype, <any>"createTokenRequestBody");
+
+            const config: ClientConfiguration = await ClientTestUtils.createTestClientConfiguration();
+            if (!config.cryptoInterface) {
+                throw TestError.createTestSetupError("configuration cryptoInterface not initialized correctly.");
+            }
+
+            // Set up required objects and mocked return values
+            const testState = `eyAiaWQiOiAidGVzdGlkIiwgInRzIjogMTU5Mjg0NjQ4MiB9${Constants.RESOURCE_DELIM}userState`;
```
peterzenz, line 1145: Suggest a comment on what the contents of this are and how you got it so that future explorers in the code base might have a better understanding of your intent.  Ditto for 1154 below.

peterzenz: There are a few more below that might be worthwhile to identify the what and how on them.

peterzenz (approved):

lib/msal-browser/docs/ssh-certificates.md
```diff
@@ -0,0 +1,93 @@
...
+> :information_source: SSH Certificate acquisition is an advanced feature that is only enabled a reduced set of applications. Make sure this feature is suited for your use case before attempting to use it.
+
+In addition to `Bearer` and `PoP` access tokens, MSAL Browser supports the acquisition of Ephemeral `SSH` certificates to be used for authorized resource request. This document outlines the recommended usage pattern of MSAL's `acquireToken` APIs when acquiring these ephemeral SSH Certificates.
+
+
+## Acquiring an SSH Certificate
+
+### Pre-requsities
+
+MSAL takes care of acquiring, caching and renewing (or "refreshing") the SSH certificate being requested. The following actions are the cleint application's responsibility:
```
tnorling:
```suggestion
MSAL takes care of acquiring, caching and renewing (or "refreshing") the SSH certificate being requested. The following actions are the client application's responsibility:
```

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -409,8 +411,15 @@ export abstract class CacheManager implements ICacheManager {
                     return;
                 }
 
+                // KeyId (sshKid) in request must match cached SSH certificate keyId
+                if(tokenType === AuthenticationScheme.SSH) {
+                    if(!!keyId && !this.matchKeyId(entity, keyId)) {
```
tnorling: Dont think this is necessary
```suggestion
                    if(keyId && !this.matchKeyId(entity, keyId)) {
```

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -409,8 +411,15 @@ export abstract class CacheManager implements ICacheManager {
...
+                        return;
+                    }
+                }
+
                 // This check avoids matching outdated POP tokens that don't have the <-scheme> in the cache key
-                if(cacheKey.indexOf(AuthenticationScheme.POP) === -1) {
+                if(cacheKey.indexOf(AuthenticationScheme.POP) === -1 && cacheKey.indexOf(AuthenticationScheme.SSH) === -1) {
```
tnorling: Is this needed since this is a brand new feature?

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -582,17 +591,24 @@ export abstract class CacheManager implements ICacheManager {
...
-            const kid = accessTokenWithAuthSchemeEntity.keyId;
-
-            if (kid) {
-                try {
-                    await this.cryptoImpl.removeTokenBindingKey(kid);
-                } catch (error) {
-                    throw ClientAuthError.createBindingKeyNotRemovedError();
-                }
+            // MSAL only stores key for PoP tokens, not for SSH Certificates
+            switch (credential.tokenType) {
```
tnorling: Can this just be an `if` since there's only one condition anyway?

lib/msal-browser/docs/ssh-certificates.md
```diff
@@ -0,0 +1,93 @@
+# Ephemeral SSH Certificates
+
+> :information_source: SSH Certificate acquisition is an advanced feature that is only enabled a reduced set of applications. Make sure this feature is suited for your use case before attempting to use it.
```
tnorling: Can you expand on this a bit more or include a link that does? It's not clear to me from reading this what this flow is or when it should be used. We should explain what applications/use cases this is suitable for, how it's different from the Bearer and Pop flows and what the benefits/drawbacks are.

lib/msal-common/src/cache/CacheManager.ts
hectormmg: It is, this if block catches credentials that have -accesstoken_with_authscheme in the key but no auth scheme  (i.e. pop or ssh-cert) at the end of the key and removes them. If we don't add ssh-cert to the list of exceptions, SSH certs will be deleted even with the proper key format. There may be a more elegant solution, if I can figure it out I'll change this.

lib/msal-common/src/cache/CacheManager.ts
hectormmg: You're right. I initially thought there'd be some special case for SSH but there isn't.

lib/msal-common/src/cache/CacheManager.ts
tnorling: Ok no worries. I thought this was part of the migration strategy going from Pop tokens without scheme to Pop token with scheme.

lib/msal-common/test/client/AuthorizationCodeClient.spec.ts
hectormmg: This is just a test state value that matches the existing tests. This one is the GUID MSAL generates + "userState" which is a placeholder for whatever the dev decides to add to the state.

lib/msal-browser/test/utils/StringConstants.ts
hectormmg: I just used an online JWK generator. This key isn't useable or bound to anything, just to keep the test value as close to a real one as possible (although it doesn't matter because to MSAL the input should be an opaque string).

lib/msal-common/src/cache/CacheManager.ts
hectormmg: I realized this is only relevant if the token type is POP since, as you mention, there shouldn't be any SSH certs stored previous to whatever version this feature is released on. I changed this to a switch statement since there is specific behavior (the migration logic) for PoP tokens and different behavior (matching based on keyId) for SSH certs.

lib/msal-common/src/cache/CacheManager.ts
hectormmg: Changed this to an if statement.

lib/msal-browser/docs/ssh-certificates.md
hectormmg: Replaced the info source block with a general description that outlines the only use case is enabling remote login to an Azure Linux VM through SSH. There don't seem to be any existing docs for the use case yet.

tnorling (approved):

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -320,7 +320,8 @@ export abstract class CacheManager implements ICacheManager {
             filter.realm,
             filter.target,
             filter.oboAssertion,
-            filter.tokenType
+            filter.tokenType,
+            filter.keyId
```
sameerag, line 324: Wasn't this needed for POP also?

sameerag: What I am asking is will this cause confusion with pop `kid` property?

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -696,7 +710,8 @@ export abstract class CacheManager implements ICacheManager {
             clientId,
             realm: account.tenantId,
             target: scopes.printScopesLowerCase(),
-            tokenType: authScheme
+            tokenType: authScheme,
+            keyId: keyId
```
sameerag, line 715: Here: `kid` is specifically `sshKid` and not `popKid` right? If yes, will this be confusing to use the name `kid`

sameerag (approved):
Approved with comments

lib/msal-common/src/cache/CacheManager.ts
hectormmg: It's the same parameter. It wasn't necessary here before because the PoP token kid is one of the secret's claims, so it didn't need to be passed into the method (it was just parsed from the token). For SSH Certs, since we don't manage the key ourselves, we need the keyId to be passed in to reference it in the cache entity. But in short, it doesn't matter if it gets confused because it's technically the same thing.

lib/msal-common/src/cache/CacheManager.ts
hectormmg: Same as above, they're the same thing. The only reason we call it sshKid in the BaseAuthRequest is to make it clearer to whoever uses SSH that that's where they should put it.

pkanher617 (approved):

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -404,16 +406,26 @@ export abstract class CacheManager implements ICacheManager {
...
+                            // AccessToken_With_AuthScheme that doesn't have "-pop" in the key is outdated and needs to be removed
+                            this.removeItem(cacheKey, CacheSchemaType.CREDENTIAL);
+                            return;
+                        }
+                        break;
+                    case AuthenticationScheme.SSH:
+                        // KeyId (sshKid) in request must match cached SSH certificate keyId because SSH cert is bound to a specific key
+                        if(keyId && !this.matchKeyId(entity, keyId)) {
+                            return;
+                        }
```
jasonnutter, line 428: Can you add a `break` here too?

lib/msal-browser/docs/ssh-certificates.md
```diff
@@ -0,0 +1,90 @@
...
+    scopes: ["SAMPLE_SSH_SCOPE"], // Replace with your resource's scope
+    authenticationScheme: msal.AuthenticationScheme.SSH,
+    sshJwk: JSON.stringify(sshKeyData.key),
+    sshKid: sshKeyData.keyId
+}
+```
+
+*Note: Setting the `authenticationScheme` to `AuthenticationScheme.SSH` will make the `sshJwk` and `sshKid` attributes mandatory, causing the request to fail with a `ClientConfigurationError` if either is omitted.*
+
+Once the request has been configured and `SSH` is set as the `authenticationScheme`, it can be sent into the `acquireTokenRedirect` MSAL v2 API.
```
jasonnutter: Can be any of the APIs right?

lib/msal-common/src/response/AuthenticationResult.ts
```diff
@@ -13,7 +13,7 @@ import { AccountInfo } from "../account/AccountInfo";
  * - account                - An account object representation of the currently signed-in user
  * - idToken                - Id token received as part of the response
  * - idTokenClaims          - MSAL-relevant ID token claims
- * - accessToken            - Access token received as part of the response
+ * - accessToken            - Access token or SSH certificate received as part of the response
```
jasonnutter, line 16: Was it considered to expose the cert as its own property?

hectormmg: It was pretty quickly decided to go under `accessToken` with no modifications to `AuthenticationResult` because it worked and this feature isn't meant for a wide audience.

lib/msal-browser/docs/ssh-certificates.md
hectormmg: Yeah, this section is just using redirect as an example. Will make the description more general though.

lib/msal-common/src/cache/CacheManager.ts
hectormmg: Done

PR #4182, opened 2021-10-21
Remove fallback logic for iframe creation in msal-core

jasonnutter:
All of the browsers we support have `document.createElement`, and the fallback logic is potentially insecure. Can be safely removed.


tnorling (approved):

change/msal-05ef66c8-27b4-42f0-960b-aeac9bc5de18.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Remove fallback logic for iframe creation in msal-core",
```
jasonnutter:
```suggestion
  "comment": "Remove fallback logic for iframe creation in msal-core #4182",
```

pkanher617 (approved):

PR #4183, opened 2021-10-22
Add support for in-memory storage of token binding keys

hectormmg:
- Adds the `IAsyncStorage` interface and refactors `DatabaseStorage` to implement it
- Makes `IWindowStorage` generic to enable `MemoryStorage` for different stored object types (i.e. `CryptoKey`)
- Adds the `AsyncMemoryStorage` that leverages `DatabaseStorage` for async/persisted storage and `MemoryStorage` for backup
- Enables the use of the PoP auth scheme even when `IndexedDB` isn't available (i.e. when using Firefox in Private Mode, when 3rd party cookies are blocked, etc.) by allowing crypto keys to be stored in memory.
- Adds `Logger` to `CryptoOps` so it can be passed down to the Storage classes.


lib/msal-browser/src/cache/AsyncMemoryStorage.ts
```diff
@@ -0,0 +1,111 @@
...
+     * storage object if item isn't found in-memory.
+     * @param key 
+     */
+    async getItem(key: string): Promise<T | null> {
+        let item = this.inMemoryCache.getItem(key);
+        if(!item) {
+            try {
+                item = await this.indexedDBCache.getItem(key);
+            } catch (e) {
+                // Do nothing if IndexedDB is unavailable
```
jasonnutter: Can you add error log messages here (and any other errors)?

lib/msal-browser/src/cache/AsyncMemoryStorage.ts
```diff
@@ -0,0 +1,111 @@
...
+    /**
+     * Sets the item in the in-memory cache and then tries to set it in the asynchronous
+     * storage object with the given key.
+     * @param key 
+     * @param value 
+     */
+    async setItem(key: string, value: T): Promise<void> {
+        this.inMemoryCache.setItem(key, value);
+        try {
+            this.indexedDBCache.setItem(key, value);
```
jasonnutter: You should await this to avoid race conditions, I think.

lib/msal-browser/src/cache/AsyncMemoryStorage.ts
hectormmg: I still haven't figured out how to get a Logger all the way down here since CryptoOps doesn't take one.

lib/msal-browser/src/cache/AsyncMemoryStorage.ts
```diff
@@ -0,0 +1,111 @@
...
+    /**
+     * Get the item matching the given key. Tries in-memory cache first, then in the asynchronous
+     * storage object if item isn't found in-memory.
+     * @param key 
+     */
+    async getItem(key: string): Promise<T | null> {
+        let item = this.inMemoryCache.getItem(key);
+        if(!item) {
+            try {
+                item = await this.indexedDBCache.getItem(key);
```
jasonnutter: Prefer early returns instead of `let` and reassignment.

lib/msal-browser/src/cache/AsyncMemoryStorage.ts
```diff
@@ -0,0 +1,111 @@
...
+        return containsKey;
+    }
+
+    /**
+     * Clears in-memory Map and tries to delete the IndexedDB database.
+     */
+    async clear(): Promise<void> {
+        this.inMemoryCache.clear();
+        try {
+            this.indexedDBCache.deleteDatabase();
```
jasonnutter: This should be awaited.

lib/msal-browser/src/cache/AsyncMemoryStorage.ts
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    /**
+     * Removes the item matching the key from the in-memory cache, then tries to remove it from the asynchronous storage object.
+     * @param key 
+     */
+    async removeItem(key: string): Promise<void> {
+        this.inMemoryCache.removeItem(key);
+        try {
+            this.indexedDBCache.removeItem(key);
```
jasonnutter: Should be awaited

hectormmg: Didn't think about race conditions, will await every async call. Thanks!

lib/msal-browser/src/crypto/CryptoOps.ts
```diff
@@ -134,7 +134,8 @@ export class CryptoOps implements ICrypto {
      * @param kid 
      */
     async removeTokenBindingKey(kid: string): Promise<boolean> {
-        return this.cache.asymmetricKeys.delete(kid);
+        this.cache.asymmetricKeys.removeItem(kid);
```
jasonnutter: Should be awaited

lib/msal-browser/src/cache/DatabaseStorage.ts
```diff
@@ -40,50 +41,50 @@ export class DatabaseStorage<T> {
...
             const dbGet = objectStore.get(key);
-            
-            dbGet.addEventListener("success", (e: Event) => {
+            dbGet.onsuccess = (e: Event) => {
                 const event = e as IDBRequestEvent;
                 resolve(event.target.result);
-            });
-            dbGet.addEventListener("error", e => reject(e));
+            };
+            dbGet.onerror = e => reject(e);
```
jasonnutter: `addEventListerner` not work?

lib/msal-browser/src/cache/AsyncMemoryStorage.ts
jasonnutter: Right, you'll need to update CryptoOps accordingly. I started doing that in this PR: https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/3826/files

lib/msal-browser/src/cache/AsyncMemoryStorage.ts
```diff
@@ -0,0 +1,111 @@
...
+    async getItem(key: string): Promise<T | null> {
+        let item = this.inMemoryCache.getItem(key);
+        if(!item) {
+            try {
+                item = await this.indexedDBCache.getItem(key);
+            } catch (e) {
+                // Do nothing if IndexedDB is unavailable
+            }
+        }
+        return item;
```
jasonnutter:
```suggestion
        const item = this.inMemoryCache.getItem(key);
        if (item) {
            return item;
        }
            try {
                return await this.indexedDBCache.getItem(key);
            } catch (e) {
                // Do nothing if IndexedDB is unavailable
                return null;
            }
```

lib/msal-browser/src/cache/AsyncMemoryStorage.ts
jasonnutter: Also if an error is thrown in `removeItem` and the call isnt awaited, it wont be caught by teh catch.

lib/msal-browser/src/cache/DatabaseStorage.ts
hectormmg: It works but the IndexedDB MDN page uses this event format. I can change it back but I think this is the "new" way of doing it.

lib/msal-browser/src/cache/AsyncMemoryStorage.ts
```diff
@@ -0,0 +1,126 @@
...
+     * @param key 
+     */
+    async getItem(key: string): Promise<T | null> {
+        const item =this.inMemoryCache.getItem(key);
+        if(!item) {
+            try {
+                return await this.indexedDBCache.getItem(key);
+            } catch (e) {
+                if (e instanceof BrowserAuthError.createDatabaseUnavailableError) {
+                    this.logger.error(`Could not search for object with key "${key}" in asynchronous storage since IndexedDB is not available in the current browser environment."`);
```
jasonnutter: Key in theory could be pii, and doesnt seem necessary for this error...maybe create a second errorPii message that has the key?

lib/msal-browser/src/cache/DatabaseStorage.ts
jasonnutter: With event listener, generally speaking addEventListener is preferred, so that you don't remove event listeners added by others. Technically may not matter here, but I think good to be consistent to use the best practice.

lib/msal-browser/src/error/BrowserAuthError.ts
```diff
@@ -148,6 +148,10 @@ export const BrowserAuthErrorMessage = {
     signingKeyNotFoundInStorage: {
         code: "crypto_key_not_found",
         desc: "Cryptographic Key or Keypair not found in browser storage."
+    },
+    databaseUnavailable: {
+        code: "database_unavailable",
+        desc: "IndexedDB, which is required for cryptographic key storage, is unavailable. This may happen when browsing in private mode."
```
jasonnutter:
```suggestion
        desc: "IndexedDB, which is required for persistent cryptographic key storage, is unavailable. This may be caused by browser privacy features which block persistent storage in third-party contexts."
```

lib/msal-browser/src/cache/AsyncMemoryStorage.ts
```diff
@@ -0,0 +1,126 @@
...
+        this.logger = logger;
+    }
+
+    /**
+     * Get the item matching the given key. Tries in-memory cache first, then in the asynchronous
+     * storage object if item isn't found in-memory.
+     * @param key 
+     */
+    async getItem(key: string): Promise<T | null> {
+        const item =this.inMemoryCache.getItem(key);
```
jasonnutter:
```suggestion
        const item = this.inMemoryCache.getItem(key);
```

lib/msal-browser/src/cache/AsyncMemoryStorage.ts
```diff
@@ -0,0 +1,126 @@
...
+     */
+    async getKeys(): Promise<string[]> {
+        const cacheKeys = this.inMemoryCache.getKeys();
+        if (cacheKeys.length === 0) {
+            try {
+                return await this.indexedDBCache.getKeys();
+            } catch (e) {
+                if (e instanceof BrowserAuthError.createDatabaseUnavailableError) {
+                    this.logger.error("Could not query keys from asynchronous storage since IndexedDB is not available in the current browser environment.");
+                }
```
jasonnutter: Seems like this check and log message could be a helper function?

lib/msal-browser/src/cache/AsyncMemoryStorage.ts
```diff
@@ -0,0 +1,126 @@
...
+    /**
+     * Get the item matching the given key. Tries in-memory cache first, then in the asynchronous
+     * storage object if item isn't found in-memory.
+     * @param key 
+     */
+    async getItem(key: string): Promise<T | null> {
+        const item =this.inMemoryCache.getItem(key);
+        if(!item) {
+            try {
+                return await this.indexedDBCache.getItem(key);
```
jasonnutter, line 41: Can you add a log message here indicating key was not found in memory storage, and that we are looking in IDB?

lib/msal-browser/src/cache/AsyncMemoryStorage.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    /**
+     * Get all the keys from the in-memory cache as an iterable array of strings. If no keys are found, query the keys in the 
+     * asynchronous storage object.
+     */
+    async getKeys(): Promise<string[]> {
+        const cacheKeys = this.inMemoryCache.getKeys();
+        if (cacheKeys.length === 0) {
+            try {
+                return await this.indexedDBCache.getKeys();
```
jasonnutter, line 86: Ditto log message here

lib/msal-browser/src/cache/DatabaseStorage.ts
hectormmg: Got it, updating.

lib/msal-browser/src/cache/AsyncMemoryStorage.ts
hectormmg: Removing specific errors in favor of a generic "Couldn't access persistent storage" error. Stack trace should tell whether it was a get/set/remove anyway. As you point out, key ID is unecessary.

lib/msal-browser/src/cache/AsyncMemoryStorage.ts
hectormmg: Yes

jasonnutter (approved):
Looks good, thanks for the tests!

pkanher617 (approved):

PR #4188, opened 2021-10-22
Add Vue 3 Sample

tnorling:
Adds a sample demonstrating how to integrate `msal-browser` with Vue 3


samples/msal-browser-samples/vue3-sample-app/package-lock.json
```diff
@@ -0,0 +1,1227 @@
+{
+  "name": "msal-vue-sample",
```
jasonnutter: I would recommend not including a package lock, otherwise we have to keep up with its dependencies getting out of date for component governance alerts. (and instead including an .npmrc with `package-lock=false`

jasonnutter (approved):
Looks good...can you also add a link to this sample from teh msal browser readme?

samples/msal-browser-samples/vue3-sample-app/package-lock.json
tnorling: Good call, forgot to do this. Thanks for the reminder

samples/msal-browser-samples/vue3-sample-app/src/authConfig.ts
```diff
@@ -0,0 +1,52 @@
+import { LogLevel, PublicClientApplication } from '@azure/msal-browser';
+
+// Config object to be passed to Msal on creation
+export const msalConfig = {
+  auth: {
+    clientId: '3fba556e-5d4a-48e3-8e1a-fd57c12cb82e',
```
peterzenz: Do you want to remove the clientId and change it to add your clientId here?  Likewise, do we want to point to a non-PPE authority on the line just below?

samples/msal-browser-samples/vue3-sample-app/README.md
```diff
@@ -0,0 +1,207 @@
+# MSAL.js 2.x + Vue 3 + TypeScript Sample
+
+## About this sample
+
+This sample demonstrates one way you can integrate the `@azure/msal-browser` package into your Vue 3 application using the [composition API](https://v3.vuejs.org/api/composition-api.html). It is not exhaustive and there may simpler or more complex solutions depending on your specific use case.
```
peterzenz: Do we want to add a note that this will help but we're not currently supporting it at this time?  Otherwise, you'll likely have to take the bulk of the incoming questions on Vue.

peterzenz (approved):

samples/msal-browser-samples/vue3-sample-app/src/authConfig.ts
tnorling: We usually leave these in on our dev samples, mostly for testing. Given that we don't have tests for this at the moment, I've removed this for now.

samples/msal-browser-samples/vue3-sample-app/README.md
tnorling: Thanks for the callout! Added.

PR #4199, opened 2021-10-28
Add logger messages for cache misses

tnorling:
- Adds log messages with the reason for a cache miss
- Moves early return for `forceRefresh` and `claims` before the actual cache lookup occurs


change/@azure-msal-common-570c6596-9a99-4972-ab02-9d57da3fa644.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Log reason for cache miss",
```
tnorling:
```suggestion
  "comment": "Log reason for cache miss #4199",
```

jasonnutter (approved):

sameerag (approved):

PR #4234, opened 2021-11-02
Use https to install node-extensions bindings

tnorling:
One of the issues during the November release was that node-extensions [failed to install](https://github.com/AzureAD/microsoft-authentication-library-for-js/runs/4074852910?check_suite_focus=true) due to the use of the unencrypted git protocol which is [set to be deprecated](https://github.blog/2021-09-01-improving-git-protocol-security-github/)

This PR resolves this issue by using `git+https`.

This PR also adds the repository field to the `package.json` file to resolve a warning.


extensions/msal-node-extensions/package.json
```diff
@@ -45,7 +49,7 @@
   },
   "dependencies": {
     "@azure/msal-common": "^5.1.0",
-    "bindings": "git://github.com/samuelkubai/node-bindings.git#v1.6.0",
+    "bindings": "git+https://github.com/samuelkubai/node-bindings.git#v1.6.0",
```
jasonnutter, line 52: I think this could also be this:
```suggestion
    "bindings": "samuelkubai/node-bindings#v1.6.0",
```

See: https://docs.npmjs.com/cli/v6/commands/npm-install

tnorling: Yeah there's a few options, but the docs aren't clear on what protocol they use under the hood, I thought it's best to be explicit unless it's documented somewhere

jasonnutter (approved):

samuelkubai (approved):
LGTM

PR #4235, opened 2021-11-02
Fix circular dependencies in AuthenticationHeaderParser and AsyncMemoryStorage

hectormmg:
This PR:
- Changes the `ClientAuthError` import in `AuthenticationHeaderParser` from "../index.ts" to "../error/ClientAuthError.ts"
- Changes the `BrowserAuthError` import in `AsyncMemoryStorage` from "../index.ts" to "../error/BrowserAuthError"
- Changes the `Logger` import in `AsyncMemoryStorage` from "../index.ts" to "@azure/msal-common"

### DIFF
--- a/change/@azure-msal-browser-91cafee3-de82-436e-968a-0f392b472d2a.json
+++ b/change/@azure-msal-browser-91cafee3-de82-436e-968a-0f392b472d2a.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix circular dependencies in AuthenticationHeaderParser and AsyncMemoryStorage #4235",
+  "packageName": "@azure/msal-browser",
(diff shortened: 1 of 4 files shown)


jasonnutter:
Interesting that typescript doesn't catch this...is this breaking the library right now?

hectormmg:
> Interesting that typescript doesn't catch this...is this breaking the library right now?

@jasonnutter  not just that, the circular dependency was actually added by VS Code's "Quick fix..." import option

jasonnutter (approved):

tnorling (approved):

tnorling:
Looks like there's a lint rule we could use to catch this in the future: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-cycle.md

PR #4244, opened 2021-11-03
Publish workflow updates

tnorling:
This PR makes several changes to the publishing workflow to solve issues observed during the November release.

- Breaks each library's publish into a separate job
image

- Enables manually triggering a publish of any combination of libraries (provided the versions have already been bumped)
image

- Adds a script that checks npm and waits for the recently published version to be available before publishing dependent libraries
- Uploads CDN markdown updates as an artifact so that they can be recovered if a later step fails
- Includes a bootstrap step to symlink local dependencies in order to prevent failures when `npm run build:all` is run in the prepack step
- Fixes the post release branch name by replacing spaces with hyphens
- Include the 'release' label on release PRs


.github/workflows/publish-template.yml
```diff
@@ -0,0 +1,83 @@
...
+        - uses: actions/checkout@v2
+
+        - uses: actions/setup-node@v1
+
+        - name: Install dependencies
+          run: npm install
+
+        # Runs lerna bootstrap scoped to this package and any local dependencies so that npm run build:all commands work
+        - name: Link dependencies
+          run: npm run link:localDeps
```
jasonnutter: FYI you can pass `--if-present` to this command to not error out if a given package doesn't have this script defined.

tnorling: Good to know, thanks! Added.

release-scripts/getReleaseDate.js
```diff
@@ -18,4 +19,7 @@ if (dateOfNextRelease > 7) {
     }
 }
 
-console.log(months[month], year);
\ No newline at end of file
+const delimiter = noSpaces ? '-' : ' ';
+const dateComponents = [months[month], year]
+const formattedDate = noSpaces?dateComponents.join(delimiter).toLowerCase():dateComponents.join(delimiter);
```
jasonnutter:
```suggestion
const formattedDate = noSpaces ? dateComponents.join(delimiter).toLowerCase() : dateComponents.join(delimiter);
```

jasonnutter (approved):

sameerag (approved):

PR #4247, opened 2021-11-04
Ensure correlation id is set at the start of all flows

jasonnutter:


lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -341,8 +344,12 @@ export abstract class ClientApplication {
      * @deprecated
      */
     async logout(logoutRequest?: EndSessionRequest): Promise<void> {
-        this.logger.warning("logout API is deprecated and will be removed in msal-browser v3.0.0. Use logoutRedirect instead.");
-        return this.logoutRedirect(logoutRequest);
+        const correlationId = (logoutRequest && logoutRequest.correlationId) || this.browserCrypto.createNewGuid();
```
DarylThayil, line 347: you are duplicating this logic a lot. 
Maybe a helper function that encapsulates ?
```js
const correlationId = ensureCorrelationId(logoutRequest && logoutRequest.correlationId);
```

DarylThayil (approved):

jasonnutter:
Will be replaced by #4570

PR #4253, opened 2021-11-10
handle request on custom protocol instead of 'will-redirect' event

ki0ki0:
Handling of 'will-redirect' is a dirty hack in the scope of this sample.
No custom protocol is needed for 'will-redirect' handling as the redirect link doesn't make any difference.
A better approach would be to handle requests to our custom protocol to get the code.


hectormmg (approved):
LGTM, thanks for your contribution!

jasonnutter (approved):

PR #4264, opened 2021-11-17
move MsalRedirectComponent to APP_INITIALIZER

### LINKED ISSUE #4036: [Angular] Move MsalRedirectComponent To APP_INITIALIZER
### Core Library

### Wrapper Library

### Description

I've been working with MSAL v2 with the Angular wrapper recently on a project, and for the most part, I like having the MsalRedirectComponent as a way, out of the box to handle the redirect observable. 

Where things started to go awry was where various parts of the app would start requesting tokens while interaction was in progress. I know that in general, I should redirect back to a page that has no authentication required, but other developers aren't so knowledgeable and often added in things to root components (For example, something in the app.component.ts that calls an API etc). 

I solved almost all of my issues instantly by using an APP_INITIALIZER as described here : https://tutorialsforangular.com/2019/12/05/using-the-app_initializer-token-to-bootstrap-your-application/

An APP_INITIALIZER is basically a set of code that you can run *before* any other piece of code in your application, synchronously. Including any components/guards/routing etc. 

By creating an APP_INITIALIZER that simply ran the following (Note in this case, I wanted to return a promise but it also handles observables). 

````
await this.injector.get(MsalService).handleRedirectObservable().toPromise();
````

This ensured that no matter the route I got redirect back to, no matter which guards were in place, no matter which code was in app.component etc, this would *always* run before all of them and synchronously too (Getting rid of having to wait on InProgress being none). 

Anyway, the point is, is that I personally believe this to be a much cleaner and simpler way to get around many gotchas of using the MsalRedirectComponent and I would be happy to create a PR if it's something we should use. 

### Source

External (Customer)

### PR DESCRIPTION
**Pull request by yksht**
Fixes #4036 

Sorry about tests, they are not prepared good enough. Review the idea please

### CODE REVIEW

File: lib/msal-angular/src/public-api.ts [hunk 1]
```diff
@@ -17,6 +17,5 @@ export { MsalInterceptorConfiguration, MsalInterceptorAuthRequest, ProtectedReso
 export { MSAL_INSTANCE, MSAL_GUARD_CONFIG, MSAL_INTERCEPTOR_CONFIG } from "./constants";
 export { MsalBroadcastService } from "./msal.broadcast.service";
 export { MsalModule } from "./msal.module";
-export { MsalRedirectComponent } from "./msal.redirect.component";
```
jasonnutter, line 20: FYI this would be a breaking change, which would be a non-starter for this type of PR.

jasonnutter:
Closing due to inactivity. Please let us know if you would still like to explorer this idea, because I think it is interesting (preferably in a backwards compatible way).

PR #4279, opened 2021-11-24
[msal-node]: Add support for distributed cache

samuelkubai:
Add plug and play support for the redis distributed cache

To-do:

- [x] Add the feature
- [x] Add the necessary tests


lib/msal-node/src/cache/distributed/DistributedCachePlugin.ts
```diff
@@ -0,0 +1,40 @@
...
+            const kvStore = (cacheContext.tokenCache as TokenCache).getKVStore();
+  
+            if (Object.keys(kvStore).length >= 2) {
+                const accountEntity = Object.values(kvStore)[1] as AccountEntity; // the second entity is the account
+                const partitionKey = await this.partitionManager.extractKey(accountEntity);
+                          
+                await this.client.set(partitionKey, cacheContext.tokenCache.serialize());           
+            }
+        }
+    }
```
justinmchase: This strikes me as not a sufficient api for a distributed scenarios. I believe you need to alter this api to include hooks for intercepting when individual keys are accessed rather than the entire cache.

I don't believe that loading and persisting the entire cache before and after every access is scalable. If I have millions of users I cannot have the entire object stored in memory and I can't load it all on every access and the process of just reading and writing that document in a horizontally scaled environment will cause various servers to get out of sync.

Instead I would _add_ additional functions to implement that would let you hook in before/after each individual key is accessed. That would allow me to manage fetching and persisting only individual records used by each server on demand.

For example it might look like this:

```ts
export class DistributedCachePlugin implements ICachePlugin {
  public async beforeCacheKeyAccess(cacheContext: TokenCacheContext, key: string): Promise<void> {
    // if not in memory, or if whats in memory is expried...
    // fetch individual record from database using key
    // if in db and not expired then load it into in memory cache
  }

  public async afterCacheKeyAccess(cacheContext: TokenCacheContext, string key): Promise<void> {
    // grab record from memory by key
    // persist into DB
  }
}
```

And these functions would be called before / after accessing the records stored on the kv store itself. Or optionally, make an IStore interface and allow that kvstore to be a plugin.

I just don't think its reasonable to attempt to serialize and deserialize and entire token cache at all, and you probably need to abstract the store itself to be a database just load it all into memory and persist it into a database.

justinmchase: To elaborate on that last point, probably what you really need is to make the storage module plugins rather than a hard coded instance in the ClientApplication constructor:

https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/src/client/ClientApplication.ts#L77-L87

```ts
export abstract class ClientApplication {
  // private fields...
  protected storage: IStoragePlugin;

  protected constructor(configuration: Configuration) {
    this.config = buildAppConfiguration(configuration);
    this.cryptoProvider = new CryptoProvider();
    this.logger = new Logger(this.config.system.loggerOptions, name, version);
    this.storage = configuration.storagePlugin(this.logger, this.cryptoProvider) ?? new NodeStorage(this.logger, this.config.auth.clientId, this.cryptoProvider);
    this.tokenCache = new TokenCache(
        this.storage,
        this.logger,
        this.config.cache.cachePlugin
    );
  }
}
```

samuelkubai: Hi @justinmchase,

Thank you for taking the time to review the pull request, I really appreciate that.
The partition manager in this case ensures that the specific instance is only pulling and storing into cache the information specific to the user in question. We are not loading the entire cache each time, rather we are loading the user specific information only for this instance. See line 22 and 34 for more on this.

We have also added a sample in `samples/msal-node-samples/ExpressTestApp/RedisTestApp` that shows this in action and from the sample you will not that we are using session data to store the specific key for the user and loading only their information into cache.

lib/msal-node/src/cache/distributed/DistributedCachePlugin.ts
```diff
@@ -0,0 +1,40 @@
...
+            cacheContext.tokenCache.deserialize(cacheData);
+        }
+    }
+  
+    public async afterCacheAccess(cacheContext: TokenCacheContext): Promise<void> {
+        if (cacheContext.cacheHasChanged && this.partitionManager) {
+            const kvStore = (cacheContext.tokenCache as TokenCache).getKVStore();
+  
+            if (Object.keys(kvStore).length >= 2) {
+                const accountEntity = Object.values(kvStore)[1] as AccountEntity; // the second entity is the account
```
justinmchase: This seems very brittle, can you not get the key for the account entity? You're assuming the keys are in a certain ordered which while the order _is_ deterministic in javascript, relies on internal knowledge of private code which may change and break this.

samples/msal-node-samples/ExpressTestApp/RedisTestApp/App/utils/partitionManager.js
```diff
@@ -0,0 +1,30 @@
...
+function partitionManager(redisClient, sessionId) {
+    this.sessionId = sessionId;
+    this.redisClient = redisClient;
+
+    return {
+        setSessionId: (sessionId) => { 
+            this.sessionId = sessionId
+        },
+        getKey: async () => {
+            const sessionData = await this.redisClient.get(`sess:${this.sessionId}`);
```
justinmchase: This doesn't seem safe. Mutating the state on a shared service like this where its being referenced by potentially multiple calls... I don't see how you can guarantee it will be the correct session later.

samples/msal-node-samples/ExpressTestApp/RedisTestApp/App/app.js
```diff
@@ -0,0 +1,83 @@
...
+* then re-initializes msal's token cache plugin.
+*/
+function initializeTokenCachePlugin(req, _, next) {
+    partitionManager.setSessionId(req.session.id);
+    next();
+}
+
+app.use(initializeTokenCachePlugin);
+
+app.use(router(authProvider));
```
justinmchase: You appear to have a race condition here, since the sessionId can be mutated on these providers / plugins which is shared between requests, if you have two or more simultaneous calls then you may have users accessing and caching each others keys. The `beforeCacheAccess` / `afterCacheAccess` functions are both async so if parallel calls are made then you're going to be mutating the state shared between the two.

I believe you'll need to pass the req itself into the `beforeCacheAccess`/`afterCacheAccess` functions or find a way to instance the partionManager per session otherwise you probably are going to be facing situations where users are getting the wrong credentials when load increases.

lib/msal-node/src/cache/distributed/DistributedCachePlugin.ts
```diff
@@ -0,0 +1,40 @@
...
+    constructor(client: ICacheClient, partitionManager?: IPartitionManager) {
+        this.client = client;
+        this.partitionManager = partitionManager;
+    }
+  
+    public async beforeCacheAccess(cacheContext: TokenCacheContext): Promise<void> {
+        if (this.partitionManager) {
+            const partitionKey = await this.partitionManager.getKey();
+            const cacheData = await this.client.get(partitionKey);
+            cacheContext.tokenCache.deserialize(cacheData);
```
justinmchase: I think you can clearly see the race condition potential here in this one function. There entire function is async and it has multiple awaits in it. If you imagine multiple simultaneous calls happening and they're all mutating the state of the partiion manager while blocked in various non-deterministic awaits you're going to be getting the cacheContext, which is also shared mutable state, deserializing random cacheData across requests.

lib/msal-node/src/cache/distributed/DistributedCachePlugin.ts
justinmchase: Thanks for the reply @samuelkubai but I stand by my recommendation after re-reviewing. Your implementation here has likely race conditions which could result in users caching and retrieving each others keys randomly I believe. Also the code that fetches the AccountEntity seems to rely on implementation details of code outside of this plugin.

The recommendation I was making here was specifically to address these issues and I think the current cache plugin api doesn't have sufficient information for these problem to be solved safely. 

Adding a storage plugin would simply expand your storage from in memory to a database without having to worry about race conditions or internal implementation details.

samuelkubai: Hi @justinmchase, I agree that we can go the StoragePlugin route, there are definitely two ways to skin this cat.

However, I want to better understand, why you think the proposed solution will have race conditions. And yes you are right, the access to the cache is being defined by the user of the plugin as they implement the `IPartitionManager` and thus handle how the cache, which ever it maybe is accessed in a safe way.

lib/msal-node/src/cache/distributed/DistributedCachePlugin.ts
```diff
@@ -0,0 +1,40 @@
...
+import { AccountEntity, ICachePlugin, TokenCacheContext } from "@azure/msal-common";
+import { TokenCache } from "../TokenCache";
+import { IPartitionManager } from "./IPartitionManager";
+import { ICacheClient } from "./ICacheClient";
+
+export class DistributedCachePlugin implements ICachePlugin {
+    private client: ICacheClient;
+    private partitionManager?: IPartitionManager;
+
+    constructor(client: ICacheClient, partitionManager?: IPartitionManager) {
```
jasonnutter: When would a partition manager not be provided? Seems like it is needed for both functions.

lib/msal-node/src/cache/distributed/DistributedCachePlugin.ts
jasonnutter: Also not clear to me how we know the second entity is the account, or can be guaranteed of that. It looks like `kvStore` is generating by spreading the `InMemoryCache` [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/e45fd0fe87f955f6289419246248073e3d4996a3/lib/msal-node/src/cache/NodeStorage.ts#L96), but I agree relying on the order is probably not a good idea.

lib/msal-node/src/cache/distributed/DistributedCachePlugin.ts
jasonnutter: When the request is first received, this line is invoked: 

`partitionManager.setSessionId(req.session.id);`

And then downstream from that, the plugin reads/write from the cache, assuming `partitionManager.setSessionId(req.session.id);` hasn't been called again with a different `req.session.id`. However, it you make multiple requests with different users to the same server instance in parallel, it looks like there is a possibility that a request for one user would set the session id on the partition manager (which is a singleton in this sample), but then in between the time that happens and the time the plugin is invoked and the data is retrieved from the cache for the first user, another request for another user could have update that session id to the second user, which would result in the first user getting data for the second user. 

I agree with the feedback that the plugin should get the session id explicitly as a parameter, so that way it doesn't rely upon global state.

lib/msal-node/src/cache/distributed/DistributedCachePlugin.ts
samuelkubai: Yes, you are right, this should be required.

lib/msal-node/src/cache/distributed/DistributedCachePlugin.ts
samuelkubai: Yes, there's definitely a better way to do this, I have refactored the code to fetch the account entity explicitly from the list of values from the kvStore.

samples/msal-node-samples/ExpressTestApp/RedisTestApp/App/utils/partitionManager.js
samuelkubai: You a right, I have refactored the sample to have the sample create a new instance of the `PartitionManager` for every request and not share it across calls.

samples/msal-node-samples/ExpressTestApp/RedisTestApp/App/app.js
samuelkubai: You a right, I have refactored the sample to have the sample create a new instance of the `PartitionManager` for every request and not share it across calls. I would love you review on the refactor.

lib/msal-node/src/cache/distributed/DistributedCachePlugin.ts
samuelkubai: You are both right, I have refactored the sample to not share instances of the `PartitionManager` across calls by creating a new instance for each request as opposed to using the singleton we were using. I would love your review on the refactor.

samples/msal-node-samples/RedisTestApp/TestApp/App/utils/partitionManager.js
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+function partitionManager(redisClient, sessionId) {
+    this.sessionId = sessionId;
+    this.redisClient = redisClient;
```
justinmchase: This looks like a slight refactor bug, I think based on the way you're calling this function the `this.` in these two lines will be the global object which isn't quite right. You can probably just remove these two lines.

samples/msal-node-samples/RedisTestApp/TestApp/App/utils/partitionManager.js
```diff
@@ -0,0 +1,30 @@
...
+
+    return {
+        setSessionId: (sessionId) => { 
+            this.sessionId = sessionId
+        },
```
justinmchase: This seems no longer needed, probably should be removed.

samples/msal-node-samples/RedisTestApp/TestApp/App/utils/partitionManager.js
```diff
@@ -0,0 +1,30 @@
...
+        getKey: async () => {
+            const sessionData = await this.redisClient.get(`sess:${this.sessionId}`);
```
justinmchase: should remove `this.` from both of these references, they're just parameters on the outter function.

samples/msal-node-samples/RedisTestApp/TestApp/App/utils/redisClientWrapper.js
```diff
@@ -0,0 +1,33 @@
...
+* You must have redis installed on your machine and have redis server listening.
+* Note that this is only for illustration, and you'll likely need to consider cache eviction policies and handle cache server connection issues. 
+* For more information, visit: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/docs/caching.md
+*/
+
+module.exports = (client) => {
+    return {
+        get: (key) => {
+            return new Promise((resolve, reject) => {
+                client.get(key, (err, data) => {
```
justinmchase, line 12: This is a minor comment but there is actually function in nodejs which can help with this and make this whole thing simpler [`util.promisify`](https://nodejs.org/docs/latest-v16.x/api/util.html#utilpromisifyoriginal)

This whole file could look like:

```ts
const { promisify } = require('util')

module.exports = (client) => {
  const get = promisify(client.get)
  const set = promisify(client.set)
  return { get, set }
}
```

samples/msal-node-samples/RedisTestApp/src/TokenValidator.ts
```diff
@@ -0,0 +1,164 @@
...
+        // Check if a B2C application i.e. app has b2cPolicies
+        if (this.appSettings.policies) {
+            jwksUri = `${this.msalConfig.auth.authority}/discovery/v2.0/keys`;
+        } else {
+            jwksUri = `https://${Constants.DEFAULT_AUTHORITY_HOST}/${tid}/discovery/v2.0/keys`;
+        }
+
+        const client = jwksClient({
+            jwksUri: jwksUri,
+        });
```
justinmchase, line 120: I don't think you want to be creating this client and throwing it away over and over, I'm pretty sure you should only have one instance of the client per application. This client is going to be doing some caching and rate-limiting etc. and if you create a new one on every call it seems like a problem.

samples/msal-node-samples/RedisTestApp/src/TokenValidator.ts
```diff
@@ -0,0 +1,164 @@
...
+            console.log(ErrorMessages.KEYS_NOT_OBTAINED);
+            console.log(error);
+            return false;
+        }
+
+        // verify the signature at header section using keys
+        let verifiedToken: TokenClaims;
+
+        try {
+            verifiedToken = jwt.verify(authToken, keys);
```
justinmchase, line 78: I don't think this checks expiration, I could be wrong though maybe double check.

samples/msal-node-samples/RedisTestApp/src/AuthProvider.ts
```diff
@@ -0,0 +1,411 @@
...
+     * @param {NextFunction} next: express next 
+     */
+    handleRedirect = async (req: Request, res: Response, next: NextFunction): Promise<any> => {
+
+        if (req.query.state) {
+            const state = JSON.parse(this.cryptoProvider.base64Decode(req.query.state as string));
+
+            // check if nonce matches
+            if (state.nonce === req.session.nonce) {
+                const msalClient = this.initializeMsalClient(req);
```
justinmchase, line 195: So the trick is you're just creating a new ConfidentialClientApplication per request, instead of just once at startup. I agree that this solves the race condition though I'm a little concerned about the overhead of this. It doesn't seem like the way this api was designed to be used... certainly the in memory cache in here will be wiped and you'll have to fetch from redis every time.

But basically this would work, it just seems awfully hacky compared to just implementing the IStoragePlugin which really seems like the right way.

samuelkubai: First thank you for your review on this @justinmchase, really do appreciate the extra set of eyes on this feature. 

We are reviewing this with the team, to discuss whether there's a better way to implement this and not break our current APIs

samples/msal-node-samples/RedisTestApp/TestApp/App/utils/redisClientWrapper.js
samuelkubai: True, just wanted to do it the bare metal way for our sample, just in case someone comes around that is not familiar with the `promisify` library.

samples/msal-node-samples/RedisTestApp/src/TokenValidator.ts
samuelkubai: I agree, refactor this to take advantage of the client caching strategies.

samples/msal-node-samples/RedisTestApp/TestApp/App/utils/partitionManager.js
samuelkubai: Good catch, refactored.

samples/msal-node-samples/RedisTestApp/TestApp/App/utils/partitionManager.js
samuelkubai: Removed, thank you.

samples/msal-node-samples/RedisTestApp/TestApp/App/utils/partitionManager.js
samuelkubai: Yes, you are right, I have fixed it.

sadasant (commented):
This looks amazing! Thank you for doing this 🙏

lib/msal-node/src/cache/distributed/IPartitionManager.ts
```diff
@@ -0,0 +1,29 @@
...
+    /**
+     * This function given the account entity would extract the 
+     * appropriate partition key best suited for the
+     * application.
+     * 
+     * @param accountEntity: AccountEntity
+     * @returns Promise<string>
+     */
+    extractKey(accountEntity: AccountEntity): Promise<string>
+}
```
sameerag, line 39: Can we document the difference between `getKey()` and `extractKey()`?

samples/msal-node-samples/RedisTestApp/src/TokenValidator.ts
sameerag: @samuelkubai Is this intended to check expiration though? Is this done downstream?
@justinmchase is right, as we did not provide the `options` to check expiry in the fn signature, this call is not checking for expiration.

lib/msal-node/src/cache/distributed/IPartitionManager.ts
samuelkubai: @sameerag the documentation has been updated.

samples/msal-node-samples/RedisTestApp/src/TokenValidator.ts
samuelkubai: @sameerag the [jwt.verify](https://www.npmjs.com/package/jsonwebtoken#jwtverifytoken-secretorpublickey-options-callback) method does check for [token expiration](https://github.com/auth0/node-jsonwebtoken/blob/74d5719bd03993fcf71e3b176621f133eb6138c0/verify.js#L147) by default. You need to provide the options if you want to ignore that instead.

sameerag (approved):
Approved

samples/msal-node-samples/ExpressTestApp/package.json
```diff
@@ -23,7 +23,7 @@
     "typescript": "^4.0.5"
   },
   "dependencies": {
-    "@azure/msal-node": "^1.3.1",
+    "@azure/msal-node": "file:../../../lib/msal-node",
```
tnorling: We should probably leave this as-is

samples/msal-node-samples/RedisTestApp/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "@types/express": "^4.17.8",
+    "@types/express-session": "^1.17.3",
+    "@typescript-eslint/eslint-plugin": "^4.7.0",
+    "@typescript-eslint/parser": "^4.7.0",
+    "eslint": "^7.13.0",
+    "tslint": "^6.1.3",
+    "typescript": "^4.0.5"
+  },
+  "dependencies": {
+    "@azure/msal-node": "file:../../../lib/msal-node",
```
tnorling: Let's use the published version here.

tnorling (approved):
Couple minor nits but otherwise looks good

samples/msal-node-samples/ExpressTestApp/package.json
samuelkubai: Sure thing, reverting back.

samples/msal-node-samples/RedisTestApp/package.json
samuelkubai: Sure thing, using the latest version.

PR #4280, opened 2021-11-25
useMsalAuthentication hook acquires a token if user is already signed in

### LINKED ISSUE #3087: Should the contents of MsalAuthenticationTemplate tag be assured that there is a valid user session?
## Library

- [x] `@azure/msal-react@1.x.x`

## Description

Our expectation was that the content of MsalAuthenticationTemplate would only ever be rendered if a user was actually logged in and that they would be prompted to sign in again if their login had expired.

We've seen both errors that we're unable to get a token silently because the token is at its refresh limit and I've seen occasional errors that a token can't be acquired silently because no user is logged in.

In both cases, I would have expected the outer MsalAuthenticationTemplate to have redirected to login again.

Note that these errors are occurring on a fresh page load of the app, not after extended use with it open.

Are we missing something?

```
const AppContent: React.FunctionComponent<{}> = () => {
  const { accounts, instance, inProgress } = useMsal();
  const [token, setToken] = useState('');

  useEffect(() => {
    if (inProgress !== 'none') {
      return;
    }

    const username = accounts[0]?.username;
    if (!username) {
      return;
    }

    const account = instance.getAccountByUsername(username);
    if (!account) {
      return;
    }

    const tokenRequest: SilentRequest = {
      scopes: ['openid'],
      account: account,
    };
    instance
      .acquireTokenSilent(tokenRequest)
      .then((response) => {
        setToken(response.idToken || '');
      })
      .catch((e) => {
        // Something went wrong
        console.error(e);
      });
  }, [accounts, inProgress, instance]);
  
  if (!token) {
    return <span>Not signed in yet</span>;
  }
  
  return <span>Signed in!</span>;
};

export const loginRequest = {
  scopes: ['openid'],
  extraQueryParameters: { domain_hint: 'example.com' },
};
interface Props {
  msalInstance: PublicClientApplication;
}
export const App: React.FunctionComponent<Props> = (props: Props) => {
    return (
        <MsalProvider instance={props.msalInstance}>
          <MsalAuthenticationTemplate interactionType={InteractionType.Redirect} authenticationRequest={loginRequest}>
            <EmployeeAppContent />
          </MsalAuthenticationTemplate>
        </MsalProvider>
    );
}

```

## Source

- [ ] Internal (Microsoft)
- [x] Customer request

### LINKED ISSUE #4232: Easier Access Token Retrieval via React Hook
### Core Library

### Wrapper Library

### TL;DR
I would like to see a better way of retrieving a valid access token from MSAL that doesn't involve taking up half of my file (yes, I'm looking at you `useEffect()` hooks). See proposed changes below.

---

### Scenario
React.js page loads, it needs some data fetched from my API so I use a `useEffect()` hook to make the API call on page mount. Because my API endpoints are protected with bearer token auth, I need to pass along a the access token given to me by msal-react on the frontend.

### Problem
Retrieving the access token on the frontend just so that I can pass it along to my generic API call function is a *MASSIVE* pain. It requires (as per documentation) around [23 lines of code](https://github.com/Azure-Samples/ms-identity-javascript-react-tutorial/blob/main/3-Authorization-II/1-call-api/SPA/src/pages/Hello2.jsx#L21-L43) in every single file just to acquire the token used to make an API call in single line.

### Proposed Changes
I would like to see a more efficient way of retrieving access tokens that abstracts the logic around getting the access token silently or forcing a popup/redirect to login OR improving the documentation to show how this can be done in one place, rather than in every single `useEffect()` hook where data needs to be fetched from an authenticated API (could someone please share/comment if they have managed to do this this neatly).

Auth0 provides a great example of a `getAccessTokenSilently` function as is [documented here](https://auth0.com/docs/quickstart/spa/react/02-calling-an-api#get-an-access-token). Something like this, **with built-in verbose fallback to popup or redirect auth when the silent function fails**, would be a welcome addition to the `useMsal()` hook and allow for a huge reduction of repetitive code inside the `useEffect()` hooks we build today:

```javascript
const accessToken = await getAccessTokenSilently({
    audience: `api://${domain}`,
    scope: "User.Read",
});
```

### Source

External (Customer)

### LINKED ISSUE #4254: no_tokens_found with msal-react when there is a second instance of msal present.
### Core Library

### Core Library Version

2.14.2

### Wrapper Library

### Wrapper Library Version

1.1.1

### Description

We are trying to load our React app using msal-react as a nested application in a portal experience. The portal is also using MSAL and their app reg is on the same tenant. We are trying to use the useMsalAuthentication hook with a silent interaction and providing the account identifiers but keep running into the no_tokens_found error. It seems like the the wrapper recognizes the account logged in on the other instance and never fires the sso for our client. Since they are on the same tenant the local home account ids are the same and the only way to differentiate would be the aud on the token for the account.  I have had some success using msal-browser directly and bypassing the react-wrapper but even going about it that way seems to be a bit unpredictable. 

We were previously running this same app with the react wrapper in an older version of the portal that used ADAL and everything worked fine that way. The only major change is the newer portal is using MSAL. 

Is this a supported use case to have two instances for the same account? If not what is the recommended way to handle a case like this? 

### Error Message

 "no_tokens_found: No tokens were found for the given scopes, and no authorization code was passed to aquireToken"

### Msal Logs

_No response_

### MSAL Configuration

```javascript
auth: {
    authority: 'https://login.microsoftonline.com/microsoft.onmicrosoft.com',
    clientId: '...',
    redirectUri: window.location.origin,
  },
  cache: {
    storeAuthStateInCookie: isIE || isEdge || isFirefox,
  },
```

### Relevant Code Snippets

```javascript
const { login, error } = useMsalAuthentication(InteractionType.Silent, loginRequest, accountIdentifier);
```

### Reproduction Steps

Run a react app nested inside another app that is also using MSAL.
- The "shell" app has an instance of msal client with clientId:  A in configuration.
- The "nested" app has an instance of msal client with clientId: B in configuration.
- Both app registrations are on the same tenant.
- When the nested app uses the useMsalAuthentication hook no login occurs 
- Token calls from "nested" app fail with no_tokens_found

### Expected Behavior

Run a react app nested inside another app that is also using MSAL.
- The "shell" app has an instance of msal client with clientId:  A in configuration.
- The "nested" app has an instance of msal client with clientId: B in configuration.
- Both app registrations are on the same tenant.
- When the nested app uses the useMsalAuthentication hook it can determine that the aud on the id token does not match?
- Token calls from separate instances are successful. 

### Identity Provider

Azure AD / MSA

Chrome, Edge

### Regression

_No response_

### Source

Internal (Microsoft)

### LINKED ISSUE #4416: Include activeAccount in the MSAL Context
### Core Library

### Wrapper Library

### Description

I think it would be useful to include `activeAccount` in the MSAL Context (from MsalProvider). 

When performing token requests and other account operations I'm switching to using `instance.getActiveAccount()` (after setting this on login), instead of `accounts[0]` (after reading the suggestions [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/4403)). This however, won't refresh my hooks. And I'd rather not include `accounts` in the hook dependencies just to refresh.

If I could instead make `activeAccount` a dependency for my hooks my code would be a lot clearer. 

### Source

External (Customer)

tnorling:
Today the `useMsalAuthentication` hook only calls login APIs if no user is authenticated. This poses a few challenges:

1. If a single domain uses more than 1 clientId users can get into a situation where the login is skipped but they don't have any tokens available (as good as not being logged in)
2. If the user logged in a long time ago tokens and/or session may be expired and need to be refreshed

This PR aims to solve these 2 problems by adding functionality to the `useMsalAuthentication` hook that calls `acquireTokenSilent` if the user is already authenticated (and includes fallback logic). This means the hook will always attempt to acquire a token on one of the first renders, regardless of authentication state. The callback is also returned so users can acquire new tokens if needed after the first one.


hectormmg (approved):
LGTM!

jo-arroyo (approved):

lib/msal-react/src/hooks/useMsalAuthentication.ts
```diff
@@ -48,10 +74,60 @@ export function useMsalAuthentication(
...
+        let tokenRequest: SilentRequest;
+
+        if (callbackRequest) {
+            tokenRequest = {
+                ...callbackRequest
+            };
+        } else if (authenticationRequest) {
+            tokenRequest = {
+                ...authenticationRequest,
+                scopes: authenticationRequest.scopes || OIDC_DEFAULT_SCOPES
```
jasonnutter, line 95: Is this only needed because `scopes` are required for login requests?

jasonnutter: If so, that is something we've been meaning to address: #4312

tnorling: It's because `acquireTokenSilent` requires scopes and they may not have been provided since the current `useMsalAuthentication` implementation only calls login

lib/msal-react/src/hooks/useAccount.ts
```diff
@@ -4,42 +4,28 @@
...
  */
-export function useAccount(accountIdentifiers: AccountIdentifiers): AccountInfo | null {
+export function useAccount(accountIdentifiers?: AccountIdentifiers): AccountInfo | null {
     const { instance, inProgress } = useMsal();
 
-    const initialStateValue = inProgress === InteractionStatus.Startup ? null : getAccount(instance, accountIdentifiers);
-    const [account, setAccount] = useState<AccountInfo|null>(initialStateValue);
+    const [account, setAccount] = useState<AccountInfo|null>(() => getAccount(instance, accountIdentifiers));
 
     useEffect(() => {
```
ToshB, line 30: Be careful with having the value you're updating in a hook be a dependency for the hook as well (accounts). It's usually better to use [functional updates](https://reactjs.org/docs/hooks-reference.html#functional-updates) when you need to compare or mutate the current state value. 

I think the current approach will end up re-running the useEffect hook until the value is stable. Also you might encounter cases where the current value is outdated, as setState only queues an update. With functional updates the value will be updated.

lib/msal-react/src/hooks/useAccount.ts
```diff
@@ -4,42 +4,28 @@
...
         const currentAccount = getAccount(instance, accountIdentifiers);
```
ToshB: Be careful with having the value you're updating in a hook be a dependency for the hook as well. It's usually better to use [functional updates](https://reactjs.org/docs/hooks-reference.html#functional-updates) when you need to compare or mutate the current state value. 

I think the current approach will end up re-running the useEffect hook until the value is stable. Also you might encounter cases where the current value is outdated, as setState only queues an update. With functional updates the value will be updated.

Edit: I see the placement of my comment was a bit off, probably because the thing I commented wasn't changed in the PR. But it refers to the setAccount-call inside the useEffect just below.

lib/msal-react/src/hooks/useAccount.ts
```diff
@@ -4,42 +4,28 @@
...
+
+    return getAccountByIdentifiers(instance.getAllAccounts(), accountIdentifiers);
 }
 
 /**
  * Given 1 or more accountIdentifiers, returns the Account object if the user is signed-in
  * @param accountIdentifiers 
  */
-export function useAccount(accountIdentifiers: AccountIdentifiers): AccountInfo | null {
+export function useAccount(accountIdentifiers?: AccountIdentifiers): AccountInfo | null {
```
ToshB, line 25: Is there a reason behind the mixed use of null or undefined in this library? As CommonAuthorizationUrlRequest takes `AccountInfo | undefined` it would be prettier if functions returning accountInfo, like this and instance.getActiveAccount() returned undefined.

lib/msal-react/src/hooks/useAccount.ts
tnorling: Many thanks for the suggestion! I'll take a look to see how I can improve this before merging.

lib/msal-react/src/hooks/useAccount.ts
tnorling: `null` = explicitly none
`undefined` = optional/not provided

Using `null` helps us avoid mistakes when we expect certain variables to be assigned (either to a value or an acknowledgement that it doesnt have one) whereas `undefined` also carries with it the additional possibility that it hasn't been assigned yet, and that may be allowed in some scenarios i.e. optional parameters or it could be a mistake.

To give a more concrete example, in the case of `useAccount` and `getActiveAccount` by using `null` instead of `undefined` we are saying "we checked and we have no account to return to you" vs. "maybe we checked and there was no account or maybe we didnt check at all"

lib/msal-react/src/hooks/useAccount.ts
tnorling: Done

lib/msal-react/src/hooks/useMsalAuthentication.ts
```diff
@@ -48,10 +74,60 @@ export function useMsalAuthentication(
...
+
+        const getToken = async (): Promise<AuthenticationResult|null> => {
+            logger.verbose("useMsalAuthentication - Calling acquireTokenSilent");
+            return instance.acquireTokenSilent(tokenRequest).catch(async (e: AuthError) => {
+                if (e instanceof InteractionRequiredAuthError) {
+                    if (!interactionInProgress.current) {
+                        logger.verbose("useMsalAuthentication - Interaction required, falling back to interaction");
+                        return login(fallbackInteractionType, tokenRequest);
+                    } else {
+                        logger.verbose("useMsalAuthentication - Interaction required but is already in progress. Please try again, if needed, after interaction completes.");
```
jasonnutter: I think these should be `logger.error`, given they are during an error case.

lib/msal-react/src/hooks/useMsalAuthentication.ts
```diff
@@ -80,16 +156,28 @@ export function useMsalAuthentication(
...
-            // Ensure login is only called one time from within this hook, any subsequent login attempts should use the callback returned
-            setHasBeenCalled(true);
-            login().catch(() => {
-                // Errors are handled by the event handler above
-                return;
-            });
+        if (shouldAcquireToken.current && inProgress === InteractionStatus.None) {
+            shouldAcquireToken.current = false;
+            if (!isAuthenticated) {
+                logger.info("useMsalAuthentication - No user is authenticated, attempting to login");
```
jasonnutter, line 166: Potentially `logger.error` here as well?

lib/msal-react/docs/hooks.md
```diff
@@ -121,7 +119,21 @@ Docs for the APIs `PublicClientApplication` exposes can be found in the `@azure/
 
 ## `useMsalAuthentication` hook
 
-The `useMsalAuthentication` hook will initiate a login if a user is not already signed in. It accepts an `interactionType` ("Popup", "Redirect", or "Silent") and optionally accepts a [request object](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/request-response-object.md#request-and-response-objects) and an `accountIdentifiers` object if you would like to ensure a specific user is signed in. The hook will return the `response` or `error` from the login call and the `login` callback which can be used to retry a failed login.
+The `useMsalAuthentication` hook will initiate a login if a user is not already signed in, otherwise it will attempt to acquire a token.
+
+### Input Parameters
+
+- `interactionType` ("Popup", "Redirect", or "Silent") specifies how you would like to acquire tokens or login when interaction is required (note the "Silent" option has some extra considerations explained below)
```
jasonnutter: Should we link to the typedocs here?

jasonnutter: And show using the enum

lib/msal-react/src/hooks/useMsalAuthentication.ts
jasonnutter: Ah, yes makes sense.

lib/msal-react/src/hooks/useMsalAuthentication.ts
```diff
@@ -48,10 +74,60 @@ export function useMsalAuthentication(
...
+        } else if (authenticationRequest) {
+            tokenRequest = {
+                ...authenticationRequest,
+                scopes: authenticationRequest.scopes || OIDC_DEFAULT_SCOPES
+            };
+        } else {
+            tokenRequest = {
+                scopes: OIDC_DEFAULT_SCOPES
+            };
+        }
```
jasonnutter, line 102: Might be helpful to add some log messages in these if/else blocks.

lib/msal-react/src/hooks/useMsalAuthentication.ts
tnorling: Done

lib/msal-react/src/hooks/useMsalAuthentication.ts
tnorling: Not so sure given logging a user in is one of the main purposes of this hook

lib/msal-react/docs/hooks.md
tnorling: Updated with links to the typedocs. The examples below show using the enum

lib/msal-react/src/hooks/useMsalAuthentication.ts
tnorling: Done

lib/msal-react/src/hooks/useMsalAuthentication.ts
jasonnutter: That's fair!

jasonnutter (approved):

tabarra:
Awesome change guys.  
Any ETA for this to be pushed to NPM?

tnorling:
@tabarra We release new versions of the library on the first Monday of every month. You can expect this change to be included in the February release.

PR #4296, opened 2021-12-02
Add support for requested claims in silent token acquisition

hectormmg:
This PR:
- Removes logic that forces a network request if claims are included in a silent token request
- Adds `requestedClaimsHash` optional property to `BaseAuthRequest` so the claims string SHA-256 hash can be passed into the `CacheManager` when matching access tokens
- Adds `requestedClaimsHash` and `requestedClaims` properties to `AccessTokenCredential` type to support the claims hash in the cache key as well as including the claims string in the cached credential for visibility
- Updates `ICrypto` and its implementations to include the `hashString` API, enabling SHA-256 hashing outside of protocol-specific cryptographic steps
- Adds generic migration logic for updating valid credentials with an outdated cache key schema to the updated cache key schema


lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -405,6 +409,14 @@ export abstract class CacheManager implements ICacheManager {
             if (!!target && !this.matchTarget(entity, target)) {
                 return;
             }
+            
+            // If request OR cached entity has requested Claims Hash, check if they match
+            if (requestedClaimsHash || entity.requestedClaimsHash) {
+                // Don't match if either is undefined or they are different
+                if (!this.matchRequestedClaimsHash(entity, requestedClaimsHash)) {
```
tnorling: Does this need a helper?
```suggestion
                if (entity.requestedClaimsHash !== requestedClaimsHash) {
```

lib/msal-common/src/client/SilentFlowClient.ts
```diff
@@ -57,21 +56,20 @@ export class SilentFlowClient extends BaseClient {
...
         }
 
         // We currently do not support silent flow for account === null use cases; This will be revisited for confidential flow usecases
         if (!request.account) {
             throw ClientAuthError.createNoAccountInSilentRequestError();
         }
 
-        const requestScopes = new ScopeSet(request.scopes || []);
+        if (request.claims && !StringUtils.isEmpty(request.claims)) {
+            request.requestedClaimsHash = await this.cryptoUtils.hashString(request.claims);
```
tnorling: We should do this as part of the request generation/validation

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -139,7 +139,11 @@ export class ResponseHandler {
         // Add keyId from request to serverTokenResponse if defined
         serverTokenResponse.key_id = serverTokenResponse.key_id || request.sshKid || undefined;
 
-        const cacheRecord = this.generateCacheRecord(serverTokenResponse, authority, reqTimestamp, idTokenObj, request.scopes, oboAssertion, authCodePayload);
+        if(request.claims && !StringUtils.isEmpty(request.claims)) {
+            request.requestedClaimsHash = await this.cryptoObj.hashString(request.claims);
```
tnorling: Do this at request generation

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -636,10 +648,11 @@ export abstract class CacheManager implements ICacheManager {
      * @param environment
      * @param authScheme
      */
-    readCacheRecord(account: AccountInfo, clientId: string, scopes: ScopeSet, environment: string, authScheme: AuthenticationScheme, keyId?: string): CacheRecord {
+    readCacheRecord(account: AccountInfo, clientId: string, request: BaseAuthRequest, environment: string): CacheRecord {
```
tnorling, line 649: This is public surface and may be a breaking change

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -700,7 +713,9 @@ export abstract class CacheManager implements ICacheManager {
      * @param scopes
      * @param authScheme
      */
-    readAccessTokenFromCache(clientId: string, account: AccountInfo, scopes: ScopeSet, authScheme: AuthenticationScheme, keyId?: string): AccessTokenEntity | null {
+    readAccessTokenFromCache(clientId: string, account: AccountInfo, request: BaseAuthRequest): AccessTokenEntity | null {
```
tnorling, line 714: Same here

tnorling (commented):
Does this need any migration from existing non-claims aware access token entities to the new claims aware one?

lib/msal-common/src/cache/CacheManager.ts
tnorling: What about idTokens?

lib/msal-common/src/request/BaseAuthRequest.ts
```diff
@@ -30,5 +31,6 @@ export type BaseAuthRequest = {
     resourceRequestMethod?: string;
     resourceRequestUri?: string;
     sshJwk?: string,
-    sshKid?: string
+    sshKid?: string,
+    requestedClaimsHash?: string;
```
jasonnutter, line 35: Would applications be expected to pass this in? If so, can you please provide an example of what that would look like?

lib/msal-common/src/cache/CacheManager.ts
hectormmg: It doesn't really, just added it out of habit/consistency. Will apply your suggestion.

lib/msal-common/src/cache/CacheManager.ts
hectormmg: The unified cache schema spec doesn't cover requested claims on the ID token. With these changes, the ID token will match all requests regardless of the requested claims. Should we have a discussion on if we want to have different behavior?

lib/msal-common/src/request/BaseAuthRequest.ts
hectormmg: @jasonnutter  no, this is calculated by MSAL based on the `claims` value. I just added it here so it can be set and retrieved later on. Is this the right place for it?

hectormmg:
> Does this need any migration from existing non-claims aware access token entities to the new claims aware one?
@tnorling 

Regarding this question, as implemented now no migration code is necessary. The reason we needed migration code for the -with_authscheme change was that the initial implementation was wrong, so we needed to remove the previous (wrongly keyed) cache items because they weren't overwritten on their own.

If a token was cached without claims before updating, it will match if no claims are requested in a new request after update.

The one thing I'm still uncertain about is that as implemented, if no claims are requested, the "-" separator in the claims place isn't added to the key. The spec technically requires the separator to be left in for all the optional key fields, even when the value is undefined, but I don't see the point of doing that here. What do you think?

lib/msal-common/src/cache/CacheManager.ts
tnorling: Today all claims requests result in a network request to get new access/id tokens. If we're changing the behavior to use cache we need to make sure we don't ignore id tokens otherwise we may break existing scenarios

lib/msal-common/src/response/ResponseHandler.ts
hectormmg: Moved to `initializeBaseRequest` in `BaseClient`

lib/msal-common/src/cache/CacheManager.ts
hectormmg: @tnorling @jasonnutter do we prefer to add another optional parameter add the end of these method signatures to avoid a breaking change, or would we rather have a breaking change and a cleaner method signature?

I'm leaning towards avoiding a breaking change even if it means having a longer argument list.

tnorling: I'm fine with breaking changes in common in the name of cleaner code, we just need to make the changefiles aware if that's the direction we're taking.

lib/msal-common/src/cache/CacheManager.ts
hectormmg: As discussed offline, pending a more in-depth design discussion, we're leaving this as an access token-only behavior, adhering to the cross-platform cache spec given that we'd need to look into the claims object to distinguish between access token and id token claims requests.

tnorling:
> Regarding this question, as implemented now no migration code is necessary. The reason we needed migration code for the -with_authscheme change was that the initial implementation was wrong, so we needed to remove the previous (wrongly The one thing I'm still uncertain about is that as implemented, if no claims are requested, the "-" separator in the claims place isn't added to the key. The spec technically requires the separator to be left in for all the optional key fields, even when the value is undefined, but I don't see the point of doing that here. What do you think?

This is done so that if some, but not all, optional values are provided you know how to distinguish which optional values were provided. The separators are important and should be included regardless.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -131,7 +131,7 @@ export class PublicClientApplication extends ClientApplication implements IPubli
      */
     private async acquireTokenSilentAsync(request: SilentRequest, account: AccountInfo): Promise<AuthenticationResult>{
         const silentCacheClient = new SilentCacheClient(this.config, this.browserStorage, this.browserCrypto, this.logger, this.eventHandler, this.navigationClient, request.correlationId);
-        const silentRequest = silentCacheClient.initializeSilentRequest(request, account);
+        const silentRequest = await silentCacheClient.initializeSilentRequest(request, account);
         this.eventHandler.emitEvent(EventType.ACQUIRE_TOKEN_START, InteractionType.Silent, request);
```
jasonnutter, line 135: This would result in teh start event not firing at teh correct time. Please either remove this change or put the event first.

jasonnutter: Oh nevermind, misread.

jasonnutter:
> > Regarding this question, as implemented now no migration code is necessary. The reason we needed migration code for the -with_authscheme change was that the initial implementation was wrong, so we needed to remove the previous (wrongly The one thing I'm still uncertain about is that as implemented, if no claims are requested, the "-" separator in the claims place isn't added to the key. The spec technically requires the separator to be left in for all the optional key fields, even when the value is undefined, but I don't see the point of doing that here. What do you think?
> 
> This is done so that if some, but not all, optional values are provided you know how to distinguish which optional values were provided. The separators are important and should be included regardless.

Will this cause cache misses with previous versions of the cache?

lib/msal-common/src/request/BaseAuthRequest.ts
jasonnutter: I would prefer not to expose it in the API if the it is not intended to be something the application can set.

hectormmg:
> > > Regarding this question, as implemented now no migration code is necessary. The reason we needed migration code for the -with_authscheme change was that the initial implementation was wrong, so we needed to remove the previous (wrongly The one thing I'm still uncertain about is that as implemented, if no claims are requested, the "-" separator in the claims place isn't added to the key. The spec technically requires the separator to be left in for all the optional key fields, even when the value is undefined, but I don't see the point of doing that here. What do you think?
> > 
> > 
> > This is done so that if some, but not all, optional values are provided you know how to distinguish which optional values were provided. The separators are important and should be included regardless.
> 
> Will this cause cache misses with previous versions of the cache?

I added migration code that will update an outdated cache key as long as the credential's internals match. So there shouldn't be any cache misses. This makes it backward compatible. In the case of a version downgrade, there's really no way to avoid tokens being duplicated under different keys and then causing a multiple matching tokens error (so it's not forward compatible).

lib/msal-browser/src/interaction_client/BaseInteractionClient.ts
```diff
@@ -91,6 +91,11 @@ export abstract class BaseInteractionClient {
             this.logger.verbose(`Authentication Scheme set to "${request.authenticationScheme}" as configured in Auth request`);
         }
 
+        // Set requested claims hash if claims were requested
+        if (request.claims && !StringUtils.isEmpty(request.claims)) {
+            request.requestedClaimsHash = await this.browserCrypto.hashString(request.claims);
+        } 
+
         const validatedRequest: BaseAuthRequest = {
```
jasonnutter, line 99: Instead of having this method return `BaseAuthRequest`, can we have it return some internal type that includes `requestedClaimsHash`?

jasonnutter: That would be one way to pass this downstream without exposing it upstream.

jasonnutter: The other alternative is to have it omitted from `SilentRequest`, `RedirectRequest`, `PopupRequest`, etc, which there is already precedence for, so I'd be fine with either approach.

hectormmg: Got it, I'll go through both options and see which is cleaner.

jasonnutter (approved):

change/@azure-msal-angular-db641b59-aad7-49f5-8b4f-06625de69bab.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Add support for requested claims in silent token acquisition #4296",
+  "packageName": "@azure/msal-angular",
+  "email": "hectormg@hey.com",
```
tnorling: Double check your git config on a couple of these

samples/package.json
```diff
@@ -15,9 +15,9 @@
         "@types/jest": "^26.0.19",
         "dotenv": "^8.2.0",
         "find-process": "^1.4.4",
-        "jest": "^27.0.4",
-        "puppeteer": "^12.0.1",
-        "ts-jest": "^27.0.3",
+        "jest": "^27.1.1",
+        "puppeteer": "^10.2.1",
```
tnorling: Why were these downgraded?

change/@azure-msal-angular-db641b59-aad7-49f5-8b4f-06625de69bab.json
hectormmg: Thanks, changed the incorrect e-mails.

samples/package.json
hectormmg: I was having problems with a few of the testing deps a couple of weeks ago. Just bumped the versions back up and confirmed tests still work. Thanks!

tnorling (approved):

sameerag (approved):
Approving. This `async` public APIs will probably be useful for async storage APIs.

PR #4302, opened 2021-12-03
Fix React inProgress startup bug

### LINKED ISSUE #4282: MSAL in Storybook
### Core Library

### Core Library Version

2.19.0

### Wrapper Library

### Wrapper Library Version

1.1.1

### Description

I'm Trying to set up MSAL in Storybook using a decorator, https://storybook.js.org/docs/react/writing-stories/decorators

It works fine on pageload and if refresh the page, but if i click between the stories I can see that "inProgress" is "startup" and nothing else happens.

I have tried with different configurations for cache, but to no avail.

Anyone got any tips or have done this themself?

I reckon it is something on Storybooks side and how they handle routing as it the same configration i use for the other apps and they work fine.

I dont want to add MSALProvider for each story

### MSAL Configuration

```javascript
auth: {
    clientId: xxx,
    authority: xxx,
    redirectUri: '/',
  },
  cache: {
    secureCookies: true,
    cacheLocation: 'localStorage', // This configures where your cache will be stored
    storeAuthStateInCookie: true, // Set this to "true" if you are having issues on IE11 or Edge
  },
```

### Relevant Code Snippets

```javascript
./storybook/preview.js

const withMSALProvider = (Story) => {
  return (
    <MsalProvider instance={publicClientApplication}>
      <MsalAuthenticationTemplate
        interactionType={InteractionType.Popup}
        loadingComponent={InProgressComponent}
        errorComponent={ErrorComponent}
      >
        <Story />
      </MsalAuthenticationTemplate>

      <UnauthenticatedTemplate>
        <LoggInn />
      </UnauthenticatedTemplate>
    </MsalProvider>
  );
};

export const decorators = [withMSALProvider];
```

### Identity Provider

Azure AD / MSA

### Source

External (Customer)

tnorling:
If `handleRedirectPromise` is called before `MsalProvider` is rendered, the event callbacks that `MsalProvider` relies on to set `inProgress` state won't be available yet resulting in `inProgress` being forever stuck in `startup`. This PR fixes this by setting `inProgress` to `none` in the `finally` block of its own call to `handleRedirectPromise`

Fixes #4282

### DIFF
--- a/lib/msal-react/src/MsalProvider.tsx
+++ b/lib/msal-react/src/MsalProvider.tsx
@@ -86,6 +86,15 @@ export function MsalProvider({instance, children}: MsalProviderProps): React.Rea
         instance.handleRedirectPromise().catch(() => {
             // Errors should be handled by listening to the LOGIN_FAILURE event
             return;
+        }).finally(() => {
+            /*
+             * If handleRedirectPromise returns a cached promise the necessary events may not be fired
+             * This is a fallback to prevent inProgress from getting stuck in 'startup'
+             */
+            if (inProgressRef.current === InteractionStatus.Startup) {
+                inProgressRef.current = InteractionStatus.None;
+                setInProgress(InteractionStatus.None);
+            }
         });
 
         return () => {
--- a/change/@azure-msal-react-c93b2a90-5aee-462b-a101-6eadcceacfa6.json
+++ b/change/@azure-msal-react-c93b2a90-5aee-462b-a101-6eadcceacfa6.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix inProgress stuck in startup state #4302",
+  "packageName": "@azure/msal-react",
+  "email": "thomas.norling@microsoft.com",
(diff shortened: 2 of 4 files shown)


lib/msal-react/src/MsalProvider.tsx
```diff
@@ -86,6 +86,15 @@ export function MsalProvider({instance, children}: MsalProviderProps): React.Rea
         instance.handleRedirectPromise().catch(() => {
             // Errors should be handled by listening to the LOGIN_FAILURE event
             return;
+        }).finally(() => {
```
jasonnutter, line 89: We've had problems with ".finally" in the past with old browsers/polyfills, FYI.

jasonnutter (approved):
One comment, otherwise looks good!

lib/msal-react/src/MsalProvider.tsx
tnorling: Is it a large enough concern to not use it here or can we address it if and when it comes up?

pkanher617 (approved):

PR #4343, opened 2021-12-30
Workflow update

bmahall:
- Modifies create-releases.js to accept the module parameter
- Creates Github Releases and Discussions as part of publish rather than post-publish script


.github/create-releases.js
```diff
@@ -187,7 +181,11 @@ async function createReleaseForFolder(folderName) {
     }
 }
 
-Promise.all(libFolders.map(createReleaseForFolder))
+if (libName == null) {
```
tnorling:
```suggestion
if (!libName) {
```

.github/create-releases.js
```diff
@@ -187,7 +181,11 @@ async function createReleaseForFolder(folderName) {
...
+    return;
+}
+
+Promise.resolve(createReleaseForFolder(libName))
```
tnorling:
```suggestion
createReleaseForFolder(libName)
```

.github/workflows/publish-template.yml
```diff
@@ -80,4 +80,8 @@ jobs:
         - name: Check npm for package availability
           run: node ../../release-scripts/checkPackageAvailability.js ${{ inputs.path }}/${{ inputs.lib-name }}
 
-          
\ No newline at end of file
+        - name: Create Github releases, discussions and milestones
+          run: node ../../.github/createReleases.js -lib ${{ inputs.lib-name }}
```
tnorling:
```suggestion
          run: node ../../.github/createReleases.js -lib ${{ inputs.path}}/${{ inputs.lib-name }}
```

tnorling: And also remove the `lib/` from line 75 to include node-extensions

.github/create-releases.js
```diff
@@ -26,14 +28,6 @@ const repoMeta = {
     repo: "microsoft-authentication-library-for-js"
 };
 
-const libFolders = [
```
jasonnutter, line 29: Any objections to supporting both this hard coded list and providing the name as an environment variable? It would make it easier if this script needs to be run manually locally, i.e. use the environment variable if provided, fallback to the existing hard-coded list.

jasonnutter: Thanks!

jasonnutter (approved):

tnorling (approved):

PR #4380, opened 2022-01-10
fix: msal-browser not working in non enumerable context (storage)

### LINKED ISSUE #4379: Salesforce integration impossible
### Core Library

### Core Library Version

2.21.0

### Wrapper Library

Not Applicable

### Wrapper Library Version

None

### Description

There is a small failure in the process of retrieving cached data.

On our case, we try to integrate on Salesforce environment. Unfortunately for us, Salesforce first load it's own framework which add a kind of proxy to all window API.

This proxy works well with msal apart from one point, this line :
https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/2f0c4095993f54056acc1b30b52628a9a308cd9e/lib/msal-browser/src/cache/BrowserStorage.ts#L42

It seems that the proxy make the keys of the storage to not be enumerable.

### Error Message

It basically failed to return token using `acquireTokenSilent` with an error of `InteractionRequiredAuthError` which according to documentation should lead to a call to `acquireTokenRedirect` (or Popup).

In this case we have a redirect loop as everytime I try to get the token I got redirected.

### Msal Logs

_No response_

### MSAL Configuration

```javascript
{
    auth: {
        clientId: 'someClientId',
        authority: 'https://login.microsoftonline.com/someTenantId',
    },
}
```

### Relevant Code Snippets

```javascript
A bit hard to explain.
```

### Reproduction Steps

(Might only means something for people with Salesforce knowledge)
1. Create a js file and put the source code of https://alcdn.msauth.net/browser/2.21.0/js/msal-browser.min.js
2. Upload the js file as static resource
3. Create a LWC (Lightning Web Component) and import the static resource with `loadScript`
4. Implement `loginRedirect` and `acquireTokenSilent` flow
5. Insert your LWC on a page
6. You will get continuous redirection

### Expected Behavior

Get the token

### Identity Provider

Azure AD / MSA

Chrome, Firefox, Edge, Safari, Internet Explorer, Other

### Regression

_No response_

### Source

External (Customer)

Bartheleway:
When window storage is being proxied key might not be enumerable anymore. Safe way is to use key() which returns the n-th key of the storage.

This intends to fix #4379.

I did test modifying directly the code into the compiled source and report that it is working as expected.

### DIFF
--- a/lib/msal-browser/src/cache/BrowserStorage.ts
+++ b/lib/msal-browser/src/cache/BrowserStorage.ts
@@ -50,6 +50,7 @@ export class BrowserStorage implements IWindowStorage<string> {
     }
 
     containsKey(key: string): boolean {
-        return this.windowStorage.hasOwnProperty(key);
+        // Use getKeys to properly support environments where localStorage/sessionStorage are not enumerable, e.g. Salesforce
+        return this.getKeys().includes(key);
     }
 }
--- a/lib/msal-browser/src/cache/BrowserStorage.ts
+++ b/lib/msal-browser/src/cache/BrowserStorage.ts
@@ -39,7 +39,14 @@ export class BrowserStorage implements IWindowStorage<string> {
     }
 
     getKeys(): string[] {
-        return Object.keys(this.windowStorage);
+        const keys: string[] = [];
+
+        // Manually iterate to properly support environments where localStorage/sessionStorage are not enumerable, e.g. Salesforce
+        for (let i = 0; i < this.windowStorage.length; ++i) {
+            keys.push(this.windowStorage.key(i) || "");
+        }
+
+        return keys;
     }
 
     containsKey(key: string): boolean {
--- a/change/@azure-msal-browser-e936bf87-f0c4-456a-a850-da2f7fc384a5.json
+++ b/change/@azure-msal-browser-e936bf87-f0c4-456a-a850-da2f7fc384a5.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "fix: BrowserStorage was not working properly one enumerable context (getKeys, containsKey)",
+  "packageName": "@azure/msal-browser",
+  "email": "bartheleway@gmail.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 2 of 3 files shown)


jasonnutter (changes requested):
Thanks for filing this PR. Please include the following:

1. A test demonstrating the fix.
2. A comment in the file explaining why this change was made.
3. Change file, created by running npm run beachball:change in the root of the repo.

Thanks!

Bartheleway:
Hi, thanks for reviewing this quickly

1. I will try to make a test, not sure how I can setup the storageProxy in test as it is code written by Salesforce developers. Do you also want a test that shows that previous version and new works the same when no proxy ?
2. I don't get what you want, the commit message is describing the change and that is the correct place for explaining changes. Code comment should only clarify functionality not change. Do you agree ?
3. Ok, I missed that point, will do

jasonnutter:
> 1. I will try to make a test, not sure how I can setup the storageProxy in test as it is code written by Salesforce developers. Do you also want a test that shows that previous version and new works the same when no proxy ?

A test that shows when `window.localStorage` and `window.sessionStorage`  are not enumerable, that relevant APIs still work.

> 2. I don't get what you want, the commit message is describing the change and that is the correct place for explaining changes. Code comment should only clarify functionality not change. Do you agree ?

I disagree. Comments can be used to detail why a piece of code exists or is written in a certain way, in addition to in the commit message.

> 3. Ok, I missed that point, will do

Thanks!

lib/msal-browser/src/cache/BrowserStorage.ts
```diff
@@ -39,7 +39,13 @@ export class BrowserStorage implements IWindowStorage<string> {
     }
 
     getKeys(): string[] {
-        return Object.keys(this.windowStorage);
+        const keys: string[] = [];
+
+        for (let i = 0; i < this.windowStorage.length; ++i) {
```
jasonnutter, line 45:
```suggestion
        // Manually iterate (instead of using Object.keys()) to properly support environments where localStorage/sessionStorage are not enumerable, e.g. Salesforce
        for (let i = 0; i < this.windowStorage.length; ++i) {
```

ghost:

jasonnutter:
Closing, per comment in https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/4379#issuecomment-1088042572 to not accept fixes for unsupported environments that are not following standards.

PR #4384, opened 2022-01-11
Fix typo in forgot password flow examples

syed-ahmad:


lib/msal-react/FAQ.md
```diff
@@ -114,7 +114,7 @@ function Example() {
     const { result, error, login } = useMsalAuthentication(InteractionType.Popup);
 
     useEffect(() => {
-        if (error && error.errorMessage.indexOf("AADB2C90118") > -1) {
+        if (error && error.message.indexOf("AADB2C90118") > -1) {
```
tnorling: I don't think this is correct. This error should be of type `ServerError` which has [this shape](https://azuread.github.io/microsoft-authentication-library-for-js/ref/classes/_azure_msal_common.servererror.html). Are you having problems with this?

syed-ahmad: Hi @tnorling, yes you are right the error object returned by useMsalAuthentication does have the correct shape `ServerError`. Will update my PR.

syed-ahmad: I have updated this PR @tnorling

lib/msal-react/FAQ.md
```diff
@@ -133,7 +133,7 @@ function Example() {
     useEffect(() => {
         const callbackId = instance.addEventCallback((event) => {
             if (event.eventType === EventType.LOGIN_FAILURE) {
-                if (event.error && event.error.errorMessage.indexOf("AADB2C90118") > -1) {
+                if (event.error && event.error.message.indexOf("AADB2C90118") > -1) {
```
tnorling, line 136: Same comment here. Is this causing issues?

tnorling:
Closing as I believe the docs, as they stand today, are correct. Please let me know if you disagree.

PR #4386, opened 2022-01-11
Node testing mock behaviour updation

bmahall:
- Client : Removed mocking for msal-common module as a whole and updated accordingly. 
- Cache: Added new tests in Storage.spec.ts & TokenCache.spec.ts to improve coverage
- More tests to be added in Cache.


bmahall:
-redundant

PR #4450, opened 2022-01-26
[msal-browser] Add support for logout_hint

### LINKED ISSUE #4335: Silent logout, login without cache token
### Core Library

### Core Library Version

2.20.0

### Wrapper Library

### Wrapper Library Version

2.0.6

### Description

Hi guys, 
i have a little special use case: I have a signed in user in my application, but the app requires also an four eyes principle. So I have to aquire a token(which will be checked in backend) for a second user with prompt login. 
This is where my problems start: 
If i aquire the token for the second user, the second user stays logged in and i have to show the logout page where the second user must interactivate again. 
Currently this is a security issue, because if logout is not properly executed it is possible to sign in to all microsoft services via SSO
So i look for a solution to aquire a token without being cached or an option to logout silent. Can you help me please?

### MSAL Configuration

```javascript
auth: {
    clientId: 'XXX',
    authority: 'https://login.microsoftonline.com/XXX/',
    redirectUri: '/',
    postLogoutRedirectUri: '/logout'
  },

  cache: {
    cacheLocation: BrowserCacheLocation.SessionStorage,
    storeAuthStateInCookie: false
  }
```

### Relevant Code Snippets

```javascript
this.msalSvc.acquireTokenPopup({ loginHint: user.userPrincipal, prompt: 'login', scopes: ['api://XXX/sign']} as PopupRequest).subscribe({
...
this.msalSvc.instance.logoutPopup({ account: response.account, postLogoutRedirectUri: null});
```

### Identity Provider

Azure AD / MSA

### Source

External (Customer)

hectormmg:
This PR:
- Adds `logoutHint` attribute to `CommonEndSessionRequest` type
- Adds logic in `StandardInteractionClient.initializeLogoutRequest` to parse `login_hint` from ID token claims in the account (if) passed in and set it as `logout_hint` in the end session request
- Adds `addLogoutHint` API to `RequestParameterBuilder` and calls it on `AuthorizationCodeClient.getLogoutUri()` to populate the `logout_hint` query parameter of the end session endpoint request
- Adds unit tests that cover these changes and scenarios

### DIFF
--- a/lib/msal-browser/docs/logout.md
+++ b/lib/msal-browser/docs/logout.md
@@ -105,7 +105,7 @@ await msalInstance.logoutPopup({
 
 ## Promptless logout
 
-If your client application has the [login_hint optional claim](https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-optional-claims#v10-and-v20-optional-claims-set) enabled for ID Tokens, you can leverage the ID Token's `login_hint` claim to perform a "silent" or promptless logout while using either `logoutRedirect` or `logoutPopup`. There are two ways to achieve a promptless logout:
+If your client application has the [login_hint optional claim](https://docs.microsoft.com/azure/active-directory/develop/active-directory-optional-claims#v10-and-v20-optional-claims-set) enabled for ID Tokens, you can leverage the ID Token's `login_hint` claim to perform a "silent" or promptless logout while using either `logoutRedirect` or `logoutPopup`. There are two ways to achieve a promptless logout:
 
 ### Option 1: Let MSAL automatically parse the login_hint out of the account's ID token claims
 
--- a/lib/msal-browser/docs/logout.md
+++ b/lib/msal-browser/docs/logout.md
@@ -103,6 +103,34 @@ await msalInstance.logoutPopup({
 });
 ```
 
+## Promptless logout
+
+If your client application has the [login_hint optional claim](https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-optional-claims#v10-and-v20-optional-claims-set) enabled for ID Tokens, you can leverage the ID Token's `login_hint` claim to perform a "silent" or promptless logout while using either `logoutRedirect` or `logoutPopup`. There are two ways to achieve a promptless logout:
+
+### Option 1: Let MSAL automatically parse the login_hint out of the account's ID token claims
+
+The first and simplest option is to provide the account object you want to end the session for to the logout API. MSAL will check to see if the `login_hint` claim is available in the account's ID token and automatically add it to the end session request as `logout_hint` to skip the account picker prompt.
+
+```javascript
+const currentAccount = msalInstance.getAccountByHomeId(homeAccountId);
+// The account's ID Token must contain the login_hint optional claim to avoid the account picker
+await msalInstance.logoutRedirect({ account: currentAccount});
+```
+
+### Option 2: Manually set the logoutHint option in the logout request
+
+Alternatively, if you prefer to manually set the `logoutHint`, you can extract the `login_hint` claim in your app and set it as the `logoutHint` in the logout request: 
+
+```javascript
+const currentAccount = msalInstance.getAccountByHomeId(homeAccountId);
+
+// Extract login hint to use as logout hint
+const logoutHint = currentAccount.idTokenClaims.login_hint;
+await msalInstance.logoutPopup({ logoutHint: logoutHint });
+```
+
+***Note: Depending on the API you choose (redirect/popup), the app will still redirect or open a popup to terminate the server session. The difference is that the user will not see or have to interact with the server's account picker prompt.***
+
 ## Front-channel logout
 
 Azure AD and Azure AD B2C support the [OAuth front-channel logout feature](https://openid.net/specs/openid-connect-frontchannel-1_0.html), which enables single-sign out across all applications when a user initiates logout. To take advantage of this feature with MSAL.js, perform the following steps:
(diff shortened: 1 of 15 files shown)


lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
```diff
@@ -93,6 +116,26 @@ export abstract class StandardInteractionClient extends BaseInteractionClient {
...
 
+    /**
+     * Parses preferred_username ID Token Claim out of AccountInfo object to be used as 
+     * logout_hint in end session request.
+     * @param account 
+     */
+    protected getLogoutHintFromIdTokenClaims(account: AccountInfo): string | null {
+        const idTokenClaims: IdTokenClaims | undefined = account.idTokenClaims;
+        if (idTokenClaims) {
+            if (idTokenClaims.preferred_username) {
```
jasonnutter: We should be using the `login_hint` claim, not preferred_username, right?

jasonnutter: (this is a new optional id token claim that must be enabled by the app first)

hectormmg: Right, fixed, thanks!

lib/msal-browser/docs/logout.md
```diff
@@ -103,6 +103,34 @@ await msalInstance.logoutPopup({
 });
 ```
 
+## Promptless logout
+
+If your client application has the [login_hint optional claim](https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-optional-claims#v10-and-v20-optional-claims-set) enabled for ID Tokens, you can leverage the ID Token's `login_hint` claim to perform a "silent" or promptless logout while using either `logoutRedirect` or `logoutPopup`. There are two ways to achieve a promptless logout:
```
jasonnutter:
```suggestion
If your client application has the [login_hint optional claim](https://docs.microsoft.com/azure/active-directory/develop/active-directory-optional-claims#v10-and-v20-optional-claims-set) enabled for ID Tokens, you can leverage the ID Token's `login_hint` claim to perform a "silent" or promptless logout while using either `logoutRedirect` or `logoutPopup`. There are two ways to achieve a promptless logout:
```

jasonnutter (approved):
This looks great!

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
```diff
@@ -69,6 +69,29 @@ export abstract class StandardInteractionClient extends BaseInteractionClient {
...
         };
 
+        /**
+         * Set logout_hint to be login_hint from ID Token Claims if present
+         * and logoutHint attribute wasn't manually set in logout request
+         */
+        if (logoutRequest) {
+            // If logoutHint isn't set and an account was passed in, try to extract logoutHint from ID Token Claims
+            if (!logoutRequest.logoutHint) {
+                if(logoutRequest.account) {
```
tnorling, line 79: Should we use the `activeAccount` if no account is provided on the request?

hectormmg: It depends, passing an account in is more "intentional". If we default to the active account, then apps can't really "opt into" having the prompt.

jasonnutter: I thought about this as well, but given that currently, if you dont pass an account but do have activeAccount set, MSAL clears the cache for all accounts, not just the active account, I figured we should treat this the same and not only logout the active account from the server.

tnorling (approved):

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
jearle: Flat if statements

```js
getLogoutHint(logoutRequest) {
    if (!logoutRequest) {
      return {
        message: "logoutHint will not be set since no logout request was configured",
        logouthint: null
      }
    }

    if (logoutRequest.logoutHint) {
      return {
        message: "logoutHint has already been set in logoutRequest",
        logoutHint: null
      }
    }

    if (!logoutRequest.account) {
      return {
        message: "logoutHint was not set and account was not passed into logout request, logoutHint will not be set",
        logoutHint: null
      }
    }

    const logoutHint = this.getLogoutHintFromIdTokenClaims(
      logoutRequest.account
    );

    if (!logoutHint) {
      return {
        message: null, // are you missing a message here on purpose?
        logoutHint: null
      }
    }

    return {
      message: "Setting logoutHint to login_hint ID Token Claim value for the account provided",
      logoutHint
    }
  },
  
  ...
  const { message, logoutHint } = this.getLogoutHint(logoutRequest);

  if (message) this.logger.verbose(message);
  if (logoutHint) validLogoutRequest.logoutHint = logoutHint;
  ...
  
```
  if else ?
```ts
if (!logoutRequest) {
  this.logger.verbose("logoutHint will not be set since no logout request was configured");
} else if (logoutRequest.logoutHint) {
  this.logger.verbose("logoutHint has already been set in logoutRequest");
} else if (!logoutRequest.account) {
  this.logger.verbose("logoutHint was not set and account was not passed into logout request, logoutHint will not be set");
} else {
  const logoutHint = this.getLogoutHintFromIdTokenClaims(
    logoutRequest.account
  );

  if (logoutHint) {
    this.logger.verbose("Setting logoutHint to login_hint ID Token Claim value for the account provided");
    validLogoutRequest.logoutHint = logoutHint;
  }
}
 ```

PR #4459, opened 2022-01-28
[msal-browser] Adding exports and other changes for extensibility

pkanher617:
This PR adds additional exported files and other changes to further support extensibility in other libraries.

### DIFF
--- a/lib/msal-browser/src/interaction_client/PopupClient.ts
+++ b/lib/msal-browser/src/interaction_client/PopupClient.ts
@@ -190,12 +190,11 @@ export class PopupClient extends StandardInteractionClient {
                 const absoluteUrl = UrlString.getAbsoluteUrl(mainWindowRedirectUri, BrowserUtils.getCurrentUri());
 
                 this.logger.verbose("Redirecting main window to url specified in the request");
-                this.logger.verbosePii(`Redirecing main window to: ${absoluteUrl}`);
+                this.logger.verbosePii(`Redirecting main window to: ${absoluteUrl}`);
                 this.navigationClient.navigateInternal(absoluteUrl, navigationOptions);
             } else {
                 this.logger.verbose("No main window navigation requested");
             }
-
         } catch (e) {
             if (popup) {
                 // Close the synchronous popup if an error is thrown before the window unload event is registered
--- a/lib/msal-browser/src/interaction_client/PopupClient.ts
+++ b/lib/msal-browser/src/interaction_client/PopupClient.ts
@@ -23,6 +23,7 @@ export class PopupClient extends StandardInteractionClient {
     async acquireToken(request: PopupRequest): Promise<AuthenticationResult> {
         try {
             const validRequest = await this.preflightInteractiveRequest(request, InteractionType.Popup);
+            this.browserStorage.updateCacheEntries(validRequest.state, validRequest.nonce, validRequest.authority, validRequest.loginHint || "", validRequest.account || null);
             const popupName = PopupUtils.generatePopupName(this.config.auth.clientId, validRequest);
             const popupWindowAttributes = request.popupWindowAttributes || {};
 
@@ -82,7 +83,7 @@ export class PopupClient extends StandardInteractionClient {
      *
      * @returns A promise that is fulfilled when this function has completed, or rejected if an error was raised.
      */
-    private async acquireTokenPopupAsync(validRequest: AuthorizationUrlRequest, popupName: string, popupWindowAttributes: PopupWindowAttributes, popup?: Window|null): Promise<AuthenticationResult> {
+    protected async acquireTokenPopupAsync(validRequest: AuthorizationUrlRequest, popupName: string, popupWindowAttributes: PopupWindowAttributes, popup?: Window|null): Promise<AuthenticationResult> {
         this.logger.verbose("acquireTokenPopupAsync called");
         const serverTelemetryManager = this.initializeServerTelemetryManager(ApiId.acquireTokenPopup);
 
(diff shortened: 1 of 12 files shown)


lib/msal-browser/src/interaction_client/PopupClient.ts
```diff
@@ -189,7 +190,7 @@ export class PopupClient extends StandardInteractionClient {
                 const absoluteUrl = UrlString.getAbsoluteUrl(mainWindowRedirectUri, BrowserUtils.getCurrentUri());
 
                 this.logger.verbose("Redirecting main window to url specified in the request");
-                this.logger.verbosePii(`Redirecting main window to: ${absoluteUrl}`);
+                this.logger.verbosePii(`Redirecing main window to: ${absoluteUrl}`);
```
tnorling: Regression?

lib/msal-browser/src/interaction_client/RedirectClient.ts
```diff
@@ -83,6 +84,7 @@ export class RedirectClient extends StandardInteractionClient {
             let state: string;
             try {
                 state = this.validateAndExtractStateFromHash(responseHash, InteractionType.Redirect);
+                BrowserUtils.clearHash(window);
```
tnorling: Also regression?

peterzenz: For this and line 25 above, neither of these fit with the visibility changes (private/protected) that you have in the rest of the PR.  Is this incorporating a change from a different commit/PR that you've brought in?

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
```diff
@@ -250,8 +249,6 @@ export abstract class StandardInteractionClient extends BaseInteractionClient {
             }
         }
 
-        this.browserStorage.updateCacheEntries(validatedRequest.state, validatedRequest.nonce, validatedRequest.authority, validatedRequest.loginHint || "", validatedRequest.account || null);
```
peterzenz, line 273: Changes in this file, like the RedirectClient.ts look unrelated to the intent of the PR, maybe you need to re-sync with other updates.

lib/msal-browser/src/utils/BrowserConstants.ts
```diff
@@ -106,7 +106,8 @@ export enum ApiId {
 export enum InteractionType {
     Redirect = "redirect",
     Popup = "popup",
-    Silent = "silent"
+    Silent = "silent",
+    None = "none"
```
peterzenz, line 110: This change also looks unrelated to the intent of your PR.

peterzenz (approved):
Please confirm the changes that don't match the extensibility changes are not regressions before committing.

lib/msal-browser/src/interaction_client/RedirectClient.ts
pkanher617: These are merge conflicts from PRs that were merged after this branch was created, resolving them now.

pkanher617: Also, the cache changes are meant to be done, as they are part of the "extensibility" changes

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
pkanher617: updating cache entries is part of the "extensiblity" changes, those are now done in their respective interaction handlers

lib/msal-browser/src/utils/BrowserConstants.ts
pkanher617: also related to enums available for downstream packages

tnorling, line 110: Did you get a chance to see if this could be added in the 1p lib instead of here?

lib/msal-browser/src/interaction_client/SilentIframeClient.ts
```diff
@@ -44,6 +44,7 @@ export class SilentIframeClient extends StandardInteractionClient {
             ...request,
             prompt: PromptValue.NONE
         }, InteractionType.Silent);
+        this.browserStorage.updateCacheEntries(silentRequest.state, silentRequest.nonce, silentRequest.authority, silentRequest.loginHint || "", silentRequest.account || null);
```
tnorling, line 47: Can we put this in `initializeAuthorizationRequest`?

Can you also explain what this is for?

pkanher617: This is to update temporary cache values. Using InitializeAuthorizationRequest makes it harder to control when the cache values are updated. For some broker scenarios, we don't want to update cache entries when we initialize the request, but after we send the request up to the broker.

lib/msal-browser/src/utils/BrowserConstants.ts
pkanher617: Yes, you can't extend enums and it wouldn't make sense to add as a separate enum.

lib/msal-browser/src/interaction_client/SilentIframeClient.ts
tnorling: I see now that's where it used to be... Can we add a comment explaining this? This feels like something that may get moved back at some point without realizing what it affects. Not sure I like taking a dependency on implementation details like this

tnorling (approved):

lib/msal-browser/src/interaction_client/SilentIframeClient.ts
sameerag: I have the same inclination that this may get "optimized" at some point. 

Least case add a comment, Best case make a commented helper API with instructions regarding the sequence , 
`processReqBeforeDispatch(browserStorage, request: BaseReqType)`.

May prevent someone moving this later that way and any other functionality we may add before request dispatch for broker. Could be an overkill, will let you do the trade off.

sameerag, line 47: nit: Should we replace "" with `Constants.EMPTY_STRING`?

sameerag (approved):
Approved with comments

PR #4460, opened 2022-01-29
Improve interaction_in_progress reliability

tnorling:
- Throws `interaction_in_progress` error at time of attempting to set the temporary cache value to minimize the gap between checking/setting cache and reduce occurence of race conditions
- Moves this step to the top of interactive APIs so that events are not emitted when this is thrown
- Fixes a recent regression that made the synchronous part of `acquireTokenPopup` asynchronous


peterzenz (approved):

pkanher617 (approved):
looks good, no other comments

PR #4477, opened 2022-02-02
Msal-React updates state values concurrently

### LINKED ISSUE #4402: Unexpected/delayed context data after logout
### Core Library

### Core Library Version

2.20.0

### Wrapper Library

### Wrapper Library Version

1.1.2

### Description

After logging out using logoutPopup i consistently got an error saying no_tokens_found. After digging a bit, this was due to an acquireTokenSilent-call I had in a useEffect, taking both `accounts` and `inProgress` as dependencies. Immediately after logging out I had `inProgress=InteractionStatus.None` and `accounts.length > 0`, so my acquireTokenSilent ran, and immediately after the accounts array was emptied.

While the instance variable is stable, both inProgress and accounts change during logout. And as these are stored in different state variables in MsalProvider, they do not change at the same time (ref https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/master/lib/msal-react/src/MsalProvider.tsx#L57 and https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/master/lib/msal-react/src/MsalProvider.tsx#L81).

I am able to work around this by not using `context.accounts` in my useEffect hooks (while it works fine in most useCallbacks), and instead use `instance.getAllAccounts()`. However, this is confusing behavior, as it works fine in some cases (typically things triggered by user interaction, like useCallback) but not in useEffecs triggered by value change.

I would suggest changing these into useReducer instead of 2x useState, so you can set both at the same time, so state changes are atomic. Perhaps just moving the whole event callback handling into useReducer. 

Another note, I can't see the need for inProgressRef. It seems to be to get the current value of inProgress without requiring another hook dependency, but you should be able to do this with `setInProgress(currentProgress -> { /* get new value */; return status;})`.

### Error Message

Uncaught (in promise) InteractionRequiredAuthError: no_tokens_found: No refresh token found in the cache. Please sign-in.
    at InteractionRequiredAuthError.AuthError [as constructor] (AuthError.js:27:1)
    at new InteractionRequiredAuthError (InteractionRequiredAuthError.js:40:1)
    at Function.InteractionRequiredAuthError.createNoTokensFoundError (InteractionRequiredAuthError.js:63:1)
    at RefreshTokenClient.<anonymous> (RefreshTokenClient.js:99:1)
    at step (_tslib.js:75:1)
    at Object.next (_tslib.js:56:1)
    at _tslib.js:49:1
    at new Promise (<anonymous>)
    at __awaiter (_tslib.js:45:1)
    at RefreshTokenClient.acquireTokenWithCachedRefreshToken (RefreshTokenClient.js:93:1)

### Msal Logs

authConfig.ts:28 [Fri, 14 Jan 2022 10:45:21 GMT] : @azure/msal-browser@2.21.0 : Info - Emitting event: msal:logoutStart
2authConfig.ts:28 [Fri, 14 Jan 2022 10:45:21 GMT] : @azure/msal-react@1.2.0 : Info - MsalProvider - msal:logoutStart results in setting inProgress from none to logout
authConfig.ts:28 [Fri, 14 Jan 2022 10:45:21 GMT] : @azure/msal-browser@2.21.0 : Info - Emitting event: msal:logoutSuccess
authConfig.ts:28 [Fri, 14 Jan 2022 10:45:21 GMT] : @azure/msal-browser@2.21.0 : Info - Emitting event: msal:popupOpened
authConfig.ts:28 [Fri, 14 Jan 2022 10:45:25 GMT] : @azure/msal-browser@2.21.0 : Info - Emitting event: msal:logoutEnd
2authConfig.ts:28 [Fri, 14 Jan 2022 10:45:25 GMT] : @azure/msal-react@1.2.0 : Info - MsalProvider - msal:logoutEnd results in setting inProgress from logout to none
2authConfig.ts:28 [Fri, 14 Jan 2022 10:45:25 GMT] : @azure/msal-react@1.2.0 : Info - MsalProvider - updating account state
authConfig.ts:28 [Fri, 14 Jan 2022 10:45:25 GMT] : @azure/msal-react@1.2.0 : Info - MsalProvider - no account changes
authConfig.ts:28 [Fri, 14 Jan 2022 10:45:25 GMT] : @azure/msal-browser@2.21.0 : Info - Emitting event: msal:acquireTokenStart
authConfig.ts:28 [Fri, 14 Jan 2022 10:45:25 GMT] : [5f8c2d7a-f1b2-4354-b6ea-a05c93874085] : @azure/msal-common@6.0.0 : Info - SilentFlowClient:acquireCachedToken - No access token found in cache for the given properties.
authConfig.ts:28 [Fri, 14 Jan 2022 10:45:25 GMT] : @azure/msal-browser@2.21.0 : Info - Emitting event: msal:acquireTokenFromNetworkStart
authConfig.ts:28 [Fri, 14 Jan 2022 10:45:25 GMT] : @azure/msal-browser@2.21.0 : Info - Emitting event: msal:acquireTokenFailure
authConfig.ts:28 [Fri, 14 Jan 2022 10:45:25 GMT] : @azure/msal-react@1.2.0 : Info - MsalProvider - updating account state
2authConfig.ts:28 [Fri, 14 Jan 2022 10:45:25 GMT] : @azure/msal-react@1.2.0 : Info - MsalProvider - no account changes
-- Followed by the error message above --

### MSAL Configuration

```javascript
{
  auth: {
    clientId: CLIENT_ID,
    authority: AUTHORITY,
    redirectUri: '/',
    postLogoutRedirectUri: '/',
  },
  cache: {
  }
}
```

### Relevant Code Snippets

```javascript
export function useToken() {
  const { accounts, instance, inProgress } = useMsal();
  const [token, setToken] = React.useState<string>();

  React.useEffect(() => {
    async function load() {
      if (inProgress === InteractionStatus.None && accounts.length > 0) {
        const response = await instance.acquireTokenSilent({
          scopes: ['myScope'],
          account: accounts[0],
        });
        setToken(response.accessToken);
      } else {
        setToken(undefined);
      }
    }

    load();
  }, [accounts, inProgress, instance]);

  return [token];
}
```

### Reproduction Steps

1. Log in, and see the token being acquired
2. Log out using instance.logoutPopup()

### Expected Behavior

I would expect inProgress and accounts to change value at the same time, so the useEffect-hook would not receive an intermediate and invalid state.

### Identity Provider

Azure AD / MSA

Chrome, Firefox, Edge

### Regression

_No response_

### Source

External (Customer)

tnorling:
Today msal-react manages the `inProgress` and `accounts` state values separately which results in:
- Two renders on login/logout operations
- The first of the two renders may contain an intermediate state which is incorrect

This PR combines the state management for these two values into a single state object that is updated using a reducer. This results in a single render that contains one or both updated values.

Fixes #4402


jasonnutter (approved):

jo-arroyo (approved):

nguyenquyhy:
Previously, `accounts` state is changed (i.e. `setAccounts` is called) only when `!accountArraysAreEqual(currentAccounts, accounts)` satisfies. This means MsalProvider is not re-rendered for events such as ACQUIRE_TOKEN_START and ACQUIRE_TOKEN_SUCCESS. With this switch to useReducer, the component is re-rendered after every single MSAL event. Is this behavior change really intended?

With this change, MsalProvider would keep re-rendering everytime any component tries to request token via acquireTokenSilent, which can easily lead to a render loop if any child component tries to access the stored token.

Even if this new behavior is intended, I really think this change is large enough to warrant a minor or major version release instead of a patch release.

tnorling:
@nguyenquyhy Unnecessary re-renders are not intentional. If that's happening please open an [issue](https://github.com/AzureAD/microsoft-authentication-library-for-js/issues) or if you have an idea what changes need to be made feel free to open a pull request. Thanks!

nguyenquyhy:
Thanks for the confirmation. I'll collect a bit more details and open an issue.

PR #4478, opened 2022-02-02
Token validation basic tests

jo-arroyo:
Unit tests for the `node-token-validation` library.


lib/node-token-validation/test/config/OpenIdConfigProvider.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+
+            const mockResponse = {
+                headers: { },
+                body: {},
+                status: 200
+            };
+
+            jest.spyOn(HttpClient.prototype, 'sendGetRequestAsync').mockReturnValue(Promise.resolve(mockResponse));
+    
+            provider.fetchJwksUriFromEndpoint()
```
jasonnutter: I think you need to return/await this call (to prevent it from ending the test early)?

jo-arroyo: Yes, good point. Updated!

bmahall (approved):
LGTM

tnorling (approved):
Approving, please make sure CI completes and tests are passing before merging

PR #4505, opened 2022-02-08
Release pipeline fixes

tnorling:
This PR makes several updates to the release workflows to solve a few issues we experienced in the latest release:

- Sets the timezone of the workflow runners to PST (ensures the PR titles don't switch over to the next month premature)
- Moves `create-releases` script to `release-scripts` folder and ensures dependencies are installed before running
- Don't fail the job if releases/milestones fail to create after publish (we can do this manually if needed at the end and shouldn't block publishing of other packages)
- Updates package-locks for each lib as part of the post release PR


.github/workflows/publish-packages.yml
```diff
@@ -225,19 +226,67 @@ jobs:
       - name: Remove artifacts working-directory
         run: rm -rf artifacts
 
-      - name: Update package-locks
+      # Update package-lock files
```
jasonnutter, line 229: What version of npm/node are we running here? Would be nice to update to the latest of each (v8 and v16), if possible.

jasonnutter: Namely to help avoid lockfile churn

jasonnutter (approved):

.github/workflows/publish-packages.yml
tnorling: Should be running latest by default. We should formalize this though across the board so we don't go back and forth

.github/workflows/publish-template.yml
```diff
@@ -77,8 +77,17 @@ jobs:
...
+        run: npm install
+
       - name: Check npm for package availability
-        run: node ../../release-scripts/checkPackageAvailability.js ${{ inputs.path }}/${{ inputs.lib-name }}
+        working-directory: ./release-scripts
+        run: node checkPackageAvailability.js ${{ inputs.path }}/${{ inputs.lib-name }}
 
       - name: Create Github releases, discussions and milestones
-        run: node ../../.github/create-releases.js -lib ${{ inputs.path }}/${{ inputs.lib-name }}
+        working-directory: ./release-scripts
```
bmahall, line 89: Is there an option to set the working-directory across these three steps?

.github/workflows/publish-packages.yml
jasonnutter: Agreed. We can codify in the engines field in the root package.json.

.github/workflows/publish-template.yml
tnorling: Unfortunately no. There's a job level `working-directory` which is already set to something else for the rest of the steps in the job

bmahall (approved):

PR #4513, opened 2022-02-11
Fix broken back/forward cache behavior

### LINKED ISSUE #4171: Login redirect interaction not cleared when clicking browser back
### Core Library

### Core Library Version

2.18.0

### Wrapper Library

### Wrapper Library Version

1.1.0

### Description

I have created a minimal React application using these instructions: https://docs.microsoft.com/en-us/azure/active-directory/develop/tutorial-v2-react I am using Azure B2C tenant and loginRedirect() to log in. Previous interaction is not cleared if user presses browser back -button in the login page. Calling handleRedirectPromise() does not help.

### Error Message

_No response_

### Msal Logs

_No response_

### MSAL Configuration

```javascript
export const msalConfig = {
  auth: {
    clientId: "xxx",
    authority:
      "https://xxx/B2C_1_SignupSignin_v1",
    knownAuthorities: [
      "https://xxx/B2C_1_SignupSignin_v1",
    ],
    redirectUri: "/",
    postLogoutRedirectUri: "/",
  },
};

export const loginRequest = {
  scopes: ["openid", "email", "profile"],
};
```

### Relevant Code Snippets

```javascript
import logo from "./logo.svg";
import "./App.css";

import { useIsAuthenticated, useMsal } from "@azure/msal-react";
import { InteractionStatus } from "@azure/msal-browser";

function App() {
  const isAuthenticated = useIsAuthenticated();
  const { instance, inProgress } = useMsal();

  const logInClicked = async () => {
    console.log("logInClicked");

    if (!isAuthenticated && inProgress === InteractionStatus.None) {
      instance.loginRedirect().then((res) => {
        instance.handleRedirectPromise().catch((error) => {
          console.log(error);
        });
      });
    }
  };

  return (
    <div className="App">
      <header className="App-header">
        <img src={logo} className="App-logo" alt="logo" />
        <p>
          Edit <code>src/App.js</code> and save to reload.
        </p>
        <button onClick={logInClicked}>Login</button>
      </header>
    </div>
  );
}

export default App;
```

### Reproduction Steps

1. Press login button.
2. When you are redirected to the login page, press browser back -button to return to your application. 
3. Press the login -button again and you will not be redirected to the login -page, since previous interaction is ongoing. 

### Expected Behavior

Ongoing session should be cleared and login page should be displayed.

### Identity Provider

Azure B2C Custom Policy

Chrome, Firefox

### Regression

_No response_

### Source

External (Customer)

tnorling:
Browsers implement a performance optimization called [back/forward cache](https://web.dev/bfcache/) which caches the page before navigation and attempts to restore from that cache (including the javascript stack) if the user hits the back button. This breaks MSAL.js because `handleRedirectPromise` will not be re-run after restoring from cache and even if you manually call `handleRedirectPromise` again it will return the memoized response from the first page load, leaving temporary cache intact. Subsequent token requests will then fail with interaction_in_progress. 

In order to address this behavior this PR adds an event listener inside `acquireTokenRedirect` which will clear temporary cache if the page is reloaded as a result of hitting the back button.

Fixes #4171


lib/msal-browser/src/interaction_client/RedirectClient.ts
```diff
@@ -41,6 +41,14 @@ export class RedirectClient extends StandardInteractionClient {
             const redirectStartPage = this.getRedirectStartPage(request.redirectStartPage);
             this.logger.verbosePii(`Redirect start page: ${redirectStartPage}`);
 
+            window.addEventListener("pageshow", (event) => {
```
jasonnutter: Do we need to remove this listener at some point?

tnorling: I suppose we should remove it if an error is thrown before the redirect happens...

tnorling: Done

jasonnutter (approved):

peterzenz (approved):

PR #4514, opened 2022-02-12
[msal-common] [CacheManager] Cache is retrieved even if new claims are provided

sadasant:
This test passes. Should it fail?

In this draft PR I’m showcasing a test which is able to retrieve a token cached that initially did not have claims associated to it even if I send a silent flow request with a claims value that I have not used before in this scenario.

### DIFF
--- a/lib/msal-common/test/cache/CacheManager.spec.ts
+++ b/lib/msal-common/test/cache/CacheManager.spec.ts
@@ -5,12 +5,12 @@ import { CacheRecord } from "../../src/cache/entities/CacheRecord";
 import { AccountFilter, CredentialFilter } from "../../src/cache/utils/CacheTypes";
 import sinon from "sinon";
 import {
-  TEST_CONFIG,
-  TEST_TOKENS,
-  CACHE_MOCKS,
-  TEST_POP_VALUES,
-  TEST_SSH_VALUES,
-  TEST_CRYPTO_VALUES
+    TEST_CONFIG,
+    TEST_TOKENS,
+    CACHE_MOCKS,
+    TEST_POP_VALUES,
+    TEST_SSH_VALUES,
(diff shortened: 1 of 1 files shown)


lib/msal-common/test/cache/CacheManager.spec.ts
```diff
@@ -816,6 +816,45 @@ describe("CacheManager.ts test cases", () => {
...
+            username: "John Doe"
+        };
+
+        const silentFlowRequest: CommonSilentFlowRequest = {
+            scopes: ["user.read"],
+            account: mockedAccountInfo,
+            authority: TEST_CONFIG.validAuthority,
+            correlationId: TEST_CONFIG.CORRELATION_ID,
+            forceRefresh: false,
+            claims: `{"access_token":{"nbf":{"essential":true, "value":"1644629450"}}}`
```
sadasant, line 851: Shouldn’t this cause `readAccessTokenFromCache` to fail?

hectormmg:
A PR to fix the issue described here has been opened #4533, closing so we can track the fix there.

sadasant:
~Turns out that the issue was on the SDK side. I was migrating to the newer version of our recorder. I had forgotten that our previous recorder replaces the id_token of the responses with this: https://github.com/Azure/azure-sdk-for-js/blob/77888f34aec8b7ca3950729a9829a0a7d0d3a914/sdk/test-utils/recorder/src/utils/msalAuth.node.ts#L60~

~Therefore, my implementation was likely flawed and I hadn’t noticed it.~

Correction: I mixed up two different issues. The issue I’m stating in this comment was unrelated to this PR.

sadasant:
After further tests, I have confirmed that https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/4533 fixes this problem!

PR #4533, opened 2022-02-18
[msal-node] Fix support for claims-based token caching

hectormmg:
This PR:
- Addresses #4514 
- Adds missing implementation of `hashString` method in `CryptoProvider`
- Factors `sha256` method out from `PkceGenerator` into a static method in the new `HashUtils` class to be reused in `CryptoProvider.hashString`
- Adds missing `requestedClaims` and `requestedClaimsHash` properties to `SerializedAccessTokenEntity`, and adds their assignment in `Serializer` and `Deserializer` Access Token steps
- Adds a check for `claims` in `ClientApplication.initializeBaseRequest` method, then calculates and sets the requested claims hash if claims are requested. Also makes `initializeBaseRequest` **async** and adds required `await` statements in all it's calls.


sadasant:
I tested this PR manually by copying the resulting bundled .js files, and this PR does fix the problem for me! Thank you @hectormmg 🙏

jasonnutter (approved):

lib/msal-node/src/crypto/HashUtils.ts
```diff
@@ -0,0 +1,20 @@
...
+
+import { Hash } from "../utils/Constants";
+import crypto from "crypto";
+
+export class HashUtils {
+    /**
+     * generate 'SHA256' hash
+     * @param buffer
+     */
+    static sha256(buffer: string): Buffer {
```
jasonnutter: Can we make this a non-static method? Prefer to see us stop using static classes and methods for utility classes.

jasonnutter:
Also, can you make sure tests are added as a fast follow, if not included here?

bmahall:
> https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/4533#issuecomment-1060860619

Yes, am working with Hector on this during this week, will consult with @sameerag in case of blockers.

lib/msal-node/src/crypto/HashUtils.ts
hectormmg: Done, made it into an instantiable class and an instance method.

hectormmg:
> Also, can you make sure tests are added as a fast follow, if not included here?

As @bmahall mentioned, we're looking at adding acquireTokenSilent tests ASAP, will add tests to cover this fix and scenario as soon as we can.

sameerag (approved):

PR #4542, opened 2022-02-23
Air Gapped Cloud Support for MSAL Node E2E Tests

Robbie-Microsoft:


Robbie-Microsoft:
Per advice from my team: I'm going to make a code change where I extract entire URLs to environment variables - not just the .com part.

PR #4576, opened 2022-03-04
Token validation public APIs

jo-arroyo:
This PR adds public APIs to the `TokenValidator` class for validating tokens from requests, responses, and Express-style middleware. 

This PR also includes:
- Unit tests for the new APIs
- **A response sample**: demonstrating using msal-node to sign in and acquire tokens, and the `validateTokenFromResponse` API to validate said tokens.
- **An Express sample with views**: demonstrating using msal-node to sign in, and the `validateTokenMiddleware` API to validate tokens from the authorization header.
- Updates to the basic sample README

Separate PRs will be opened for:
- Integration tests for `validateTokenMiddleware`

**NOTE:** The excessive commits on this PR is a result of the PR it was based on being squashed and merged. Actual changes are reflected in the file changed tab.


lib/node-token-validation/src/api/TokenValidator.ts
```diff
@@ -35,9 +36,9 @@ export class TokenValidator {
      * Constructor for the TokenValidator class
      * @param configuration Object for the TokenValidator instance
      */
-    constructor(configuration: Configuration) {
+    constructor(configuration?: Configuration) {
         // Set the configuration
-        this.config = buildConfiguration(configuration);
+        this.config = configuration ? buildConfiguration(configuration) : buildConfiguration();
```
hectormmg: Is this ternary necessary? It seems the configuration object is also optional in the buildConfiguration method so we could just pass it in even if it's undefined.

lib/node-token-validation/src/api/TokenValidator.ts
```diff
@@ -49,6 +50,125 @@ export class TokenValidator {
...
 
+    /**
+     * Middleware-style function to validate token from request.
+     * @param {@link (TokenValidationParameters:type)}
+     * @param resource Optional resource to retrieve access token from session 
+     * @returns 
+     */
+    validateTokenMiddleware(options: TokenValidationParameters, resource?: string): Function {
+
+        // @ts-ignore
```
tnorling: What is this ignoring? Prefer to either fix the problem or ignore the specific rule if fixing it isn't feasible

lib/node-token-validation/src/api/TokenValidator.ts
jo-arroyo: Thanks for catching it! Forgot to remove this after the configuration change. Updated.

lib/node-token-validation/src/api/TokenValidator.ts
```diff
@@ -49,6 +50,136 @@ export class TokenValidator {
         this.openIdConfigProvider = new OpenIdConfigProvider(this.config, this.networkInterface, this.logger);
     }
 
+    /**
+     * Middleware-style function to validate token from request.
+     * @param {@link (TokenValidationParameters:type)}
+     * @param resource Optional resource to retrieve access token from session 
+     * @returns 
+     */
+    validateTokenMiddleware(options: TokenValidationParameters, resource?: string): Function {
```
jasonnutter: Instead of just `Function`, can we set a type with the expected signature?

lib/node-token-validation/src/api/TokenValidator.ts
```diff
@@ -49,6 +50,136 @@ export class TokenValidator {
...
+            // Adding access token to authorization header from session
+            if (resource && req.session && req.session.protectedResources) {
+                if (!req.session.protectedResources[resource]) {
+                    const error = ValidationConfigurationError.createMissingTokenError(`Resource ${resource} is not a protectedResource in req session. No access token to add to headers.`);
+                    next(error);
+                } else {
+                    const token = req.session.protectedResources[resource].accessToken;
+                    if (!token) {
+                        const error = ValidationConfigurationError.createMissingTokenError(`No access token for resource ${resource} in session. Unable to add to request headers.`);
+                        next(error);
```
jasonnutter, line 73: Could there a valid scenario where we would want this to not return an error?

lib/node-token-validation/src/api/TokenValidator.ts
```diff
@@ -49,6 +50,136 @@ export class TokenValidator {
...
+                this.logger.verbose("Bearer token extracted from request authorization headers");
+                return this.validateToken(token, options);
+            } else {
+                this.logger.verbose("Request authorization headers does not include bearer token");
+                // If not bearer, call CAE/EasyAuth/MISE solution here for more complex handling
+            }
+        } 
+        
+        // Validates token if found in request body
+        if (request.body && request.body.access_token) {
```
jasonnutter:
```suggestion
        if (request.body?.access_token) {
```

lib/node-token-validation/src/api/TokenValidator.ts
```diff
@@ -49,6 +50,136 @@ export class TokenValidator {
...
+     * @param request 
+     * @param {@link (TokenValidationParameters:type)}
+     * @returns {Promise.<TokenValidationResponse>}
+     */
+    async validateTokenFromRequest(request: ExpressRequest, options: TokenValidationParameters): Promise<TokenValidationResponse> {
+        this.logger.trace("TokenValidator.validateTokenFromRequest called");
+
+        // Validates token if found in authorization header
+        if (request.headers && request.headers.authorization) {
+            const authComponents = request.headers.authorization.split(" ");
```
jasonnutter: FYI nitpick:

```suggestion
            const [ scheme, token ] = request.headers.authorization.split(" ");
```

lib/node-token-validation/src/api/TokenValidator.ts
```diff
@@ -49,6 +50,136 @@ export class TokenValidator {
...
+     * Function to validate token from msal-node token acquisition response. Returns array of token validation responses.
+     * @param response 
+     * @param {@link (TokenValidationParameters:type)}
+     * @returns {Promise.<TokenValidationResponse>}
+     */
+    async validateTokenFromResponse(response: AuthenticationResult, idTokenOptions?: TokenValidationParameters, accessTokenOptions?: TokenValidationParameters): Promise<TokenValidationResponse[]> {
+        this.logger.trace("TokenValidator.validateTokenFromResponse called");
+
+        // Only validates token if tokenType from response is Bearer
+        if (response.tokenType === AuthenticationScheme.BEARER) {
```
jasonnutter: Do we need to force casing here?

jo-arroyo: As opposed to case-insensitive?

lib/node-token-validation/src/api/TokenValidator.ts
jo-arroyo: Makes sense. Added response type.

lib/node-token-validation/src/api/TokenValidator.ts
jo-arroyo: Ah, I changed it to match the other bearer comparison earlier in the code. Let me know if that's not what you meant.

lib/node-token-validation/src/api/TokenValidator.ts
jo-arroyo: The assumption is that if someone is passing an (optional) resource into the `validateTokenMiddleware` API, that they're wanting an access token retrieved and validated for that resource. It's currently throwing an error when an access token is not found, but I suppose if someone was using this API on every call there could be times when an access token is not found and they would go fetch one instead. Unsure whether it would still be best to return an error that they would then handle, or pass something else along.

lib/node-token-validation/src/api/TokenValidator.ts
```diff
@@ -49,6 +51,135 @@ export class TokenValidator {
...
+                    next(error);
+                } else {
+                    const token = req.session.protectedResources[resource].accessToken;
+                    if (!token) {
+                        const error = ValidationConfigurationError.createMissingTokenError(`No access token for resource ${resource} in session. Unable to add to request headers.`);
+                        next(error);
+                    }
+    
+                    // Attach bearer token to authorization header
+                    req.headers.authorization = `Bearer ${token}`;
```
tnorling, line 77: Are we only supporting Bearer right now?

lib/node-token-validation/src/config/Configuration.ts
```diff
@@ -55,12 +53,9 @@ const DEFAULT_SYSTEM_OPTIONS = {
...
  */
-export function buildConfiguration({
-    auth,
-    system,
-}: Configuration): TokenValidationConfiguration {
+export function buildConfiguration(config?: Configuration): TokenValidationConfiguration {
     return {
-        auth: { ...DEFAULT_TOKEN_VALIDATION_OPTIONS, ...auth },
-        system: { ...DEFAULT_SYSTEM_OPTIONS, ...system },
+        auth: { ...DEFAULT_TOKEN_VALIDATION_OPTIONS, ...config?.auth },
```
tnorling, line 58: Will this not throw if config is undefined?

lib/node-token-validation/src/api/TokenValidator.ts
jo-arroyo: Fair enough. Fixed. Thanks!

lib/node-token-validation/src/config/Configuration.ts
jo-arroyo: It doesn't seem to throw. Config is an optional parameter.

lib/node-token-validation/src/api/TokenValidator.ts
jo-arroyo: Yes, just bearer at the moment. The current stance is that more complicated scenarios (e.g. PoP) would be handled down the line by other services, so at the moment this library will only deal with Bearer.

lib/node-token-validation/src/config/Configuration.ts
tnorling: Interesting... looks like you have a test for this so that's good enough for me!

tnorling (approved):
LGTM

samples/node-token-validation-samples/response-sample/index.js
```diff
@@ -0,0 +1,205 @@
...
+             * A separate object for access token options can also be passed in to the API to validate access tokens on the response.
+             */
+            const idTokenOptions = {
+                validIssuers: config.validationParams.idTokenOptions.validIssuers,
+                validAudiences: config.validationParams.idTokenOptions.validAudiences,
+                nonce: response.idTokenClaims.nonce
+            };
+
+            /**
+             * Uncomment accessTokenOptions and add to line 138 if wanting to validate access token with custom scopes
```
bmahall, line 121: Curious if this is part of the actual code

pkanher617 (approved):

lib/node-token-validation/src/api/TokenValidator.ts
pkanher617: An error here is fine for now IMO, but we could consider making it an actionable error like InteractionRequired?

samples/node-token-validation-samples/response-sample/index.js
jo-arroyo: Could you clarify your question? These commented out lines are to demonstrate to someone using the sample how they could add `accessTokenOptions` to validate an access token returned on the response.

PR #4467, opened 2022-02-01
Feature/angular 13

sandrooco:
This PR includes updates for msal-angular.

- Angular 13.2+ (#4298)
- RxJS 7.5+ (#4154)

PR #4599, opened 2022-03-11
feat: update @azure/msal-angular to Angular 13 and rxjs 7

### LINKED ISSUE #4154: MSAL Angular support Angular 13 and RxJS v7
### Core Library

### Wrapper Library

### Description

Expecting MSAL.ja can support RxJS v7
The Roadmap of RxJS is already from 7 to 8

### Source

External (Customer)

### LINKED ISSUE #4298: Angular Ivy distribution
### Core Library

### Wrapper Library

### Description

When compiling with Angular 13, the following pops up:
> $ ng serve
> ⠙ Generating browser application bundles (phase: setup)...Processing legacy "View Engine" libraries:
> - @azure/msal-angular [es2015/esm2015] (https://github.com/AzureAD/microsoft-authentication-library-for-js.git)
> Encourage the library authors to publish an Ivy distribution.

### Source

External (Customer)

kyubisation:
This PR updates @azure/msal-angular to Angular 13, which switches the
library/package format from ViewEngine to Ivy. The new package format
also drops UMD bundles and instead provides (F)ESM 2015 and 2020
variants.
This also enables strict mode in the tsconfig.json for
@azure/msal-angular. Due to this the return type of the method
`handleRedirectObservable` in the class `MsalService` was changed
from `Observable<AuthenticationResult>` to
`Observable<AuthenticationResult | null>`, which is the correct return
type according to the interface and implementation.
Further rxjs was updated from version 6 to 7, while retaining backward
compatibility with version 6. The peer dependency now allows either
rxjs v6 or v7.

Closes #4154, closes #4298

### DIFF
--- a/change/@azure-msal-angular-24a121ea-000a-4689-a67f-c31d1e48657c.json
+++ b/change/@azure-msal-angular-24a121ea-000a-4689-a67f-c31d1e48657c.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Update @azure/msal-angular to Angular 13 and rxjs 7 (#4599)",
+  "packageName": "@azure/msal-angular",
+  "email": "lukas.spirig@gmail.com",
+  "dependentChangeType": "patch"
+}
--- a/lib/msal-angular/.gitignore
+++ b/lib/msal-angular/.gitignore
@@ -32,6 +32,7 @@ speed-measure-plugin*.json
 .history/*
 
 # misc
+.angular/cache
 /.sass-cache
 /connect.lock
 /coverage
--- a/lib/msal-angular/README.md
+++ b/lib/msal-angular/README.md
@@ -44,7 +44,7 @@ At a minimum, `@azure/msal-angular` will follow the [support schedule of the mai
 
 | MSAL Angular version | MSAL support status     | Supported Angular versions |
 |----------------------|-------------------------|----------------------------|
-| MSAL Angular v2      | Active development      | 9, 10, 11, 12              |
+| MSAL Angular v2      | Active development      | 9, 10, 11, 12, 13          |
 | MSAL Angular v1      | Active development      | 6, 7, 8, 9                 |
 | MSAL Angular v0      | In maintenance          | 4, 5                       |
 
--- a/lib/msal-angular/angular.json
+++ b/lib/msal-angular/angular.json
@@ -12,14 +12,17 @@
         "build": {
           "builder": "@angular-devkit/build-angular:ng-packagr",
           "options": {
-            "tsConfig": "./tsconfig.lib.json",
             "project": "./ng-package.json"
           },
           "configurations": {
             "production": {
               "tsConfig": "./tsconfig.lib.prod.json"
+            },
+            "development": {
+              "tsConfig": "./tsconfig.lib.json"
             }
-          }
+          },
+          "defaultConfiguration": "production"
         },
         "test": {
           "builder": "@angular-devkit/build-angular:karma",
@@ -28,18 +31,6 @@
             "tsConfig": "./tsconfig.spec.json",
             "karmaConfig": "./karma.conf.js",
             "codeCoverage": true
-          }
-        },
-        "lint": {
-          "builder": "@angular-devkit/build-angular:tslint",
-          "options": {
-            "tsConfig": [
-              "./tsconfig.lib.json",
-              "./tsconfig.spec.json"
-            ],
-            "exclude": [
-              "**/node_modules/**"
-            ]
           }
         }
       }
--- a/lib/msal-angular/ng-package.json
+++ b/lib/msal-angular/ng-package.json
@@ -2,9 +2,6 @@
   "$schema": "./node_modules/ng-packagr/ng-package.schema.json",
   "dest": "./dist",
   "lib": {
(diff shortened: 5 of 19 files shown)


jasonnutter:
@kyubisation Thanks for the contribution. Please note, there is already a PR open with proposed (#4467) changes for updating to support Angular 13 and rxjs v7, which we are currently evaluating. Furthermore, we are looking at if/how we can make supporting rxjs v7 backwards compatible with Angular 9-12 and rxjs v6, given those are the versions we currently support and dropping those would require a major version bump.

kyubisation:
> Further rxjs was updated from version 6 to 7, while retaining backward compatibility with version 6. The peer dependency now allows either rxjs v6 or v7.

I don't particularly mind which PR gets merged. Due to our company's frustrations with @azure/msal-angular (we've been in contact with Microsoft/Azure AD, I'll provide details if desired) I wanted to check how much effort is required to upgrade to Angular 13, since no roadmap was provided for that change.

Also please note that technically a backwards compatible change is not possible with the upgrade to Angular 13, since the package format changed as described.

kyubisation:
@jasonnutter As a side note; I missed the other PR. Great work by @sandrooco.
Unfortunately as it stands it will break consumers with rxjs 6, as the imports were changed in a way that is not backwards compatible. Leaving the imports as they previously were works in both v6 and v7.
Additionally `ng update` introduced backwards compatible test teardown logic, which is not necessary for this project. Destroying the context after each test works fine.
Apart from that, due to configuring strict mode, I noticed an error in the `MsalService` typings as described above and fixed it accordingly.

jasonnutter:
> Leaving the imports as they previously were works in both v6 and v7.

Thanks for the insight on this, was not immediately clear this was the case. We've put a PR to add Angular 13 sample apps and update the peer dependency range for to include version 7. #4605 

> Also please note that technically a backwards compatible change is not possible with the upgrade to Angular 13, since the package format changed as described.

Testing the above PR appears to show that just updating the peer dep to include v7 is enough, and that Angular will support packages packaged in the "legacy" format. Assuming that is the case, we will continue to do that, given we do not want to publish a breaking change of MSAL Angular (given that requires a lot more work on our side outside of just the code, e.g. documentation). Once we get the above PR merged we can incrementally look at supporting an Ivy distribution in a backwards compatible way.

Appreciate the insight and patience, thanks!

kyubisation:
> Testing the above PR appears to show that just updating the peer dep to include v7 is enough, and that Angular will support packages packaged in the "legacy" format. 

Correct, with the ngcc (Angular Compatibility Compiler). This however requires ngcc to run every time dependencies are installed fresh (to convert ViewEngine packages to the Ivy format).

> Assuming that is the case, we will continue to do that, given we do not want to publish a breaking change of MSAL Angular (given that requires a lot more work on our side outside of just the code, e.g. documentation).

Understandable, although from a consumer viewpoint very little changes. The (breaking) changes are only internal i.e. the package format. The API does not change (for the MSAL subset).

> Once we get the above PR merged we can incrementally look at supporting an Ivy distribution in a backwards compatible way.

Again, this is not really possible due to the way the Angular Package Format works. It is either a ViewEngine distribution (umd, es5 (dropped in Angular 10), esm2015, fesm2015), in which case the ngcc converts the package to the Ivy format by changing package.json and the entry files inside node_modules _or_ it is a Ivy distribution (NO umd, fesm2015, esm2020, festm2020), in which case the Angular linker directly consume the package without any changes applied within the package in node_modules.

The Ivy compiler has been the default since version 9 of Angular, so for any consumer using Ivy it should still work, but for consumers using ViewEngine (manually configurable until v12), this would break.

Feel free to reach out to the Angular team, if there is a way to provide both distribution formats, but I'd guess they would recommend just providing the ViewEngine format, since it technically works.

Angular Package Format:
- https://angular.io/guide/angular-package-format
- https://docs.google.com/document/d/1CZC2rcpxffTDfRDs6p1cfbmKNLA6x5O-NtkJglDaBVs/mobilebasic
- https://docs.google.com/document/d/1Vyqu-D2EkRbnfw4YEm4xxUZe_fAOg001nFfuU3pWMMs/edit

jasonnutter:
@kyubisation Thanks for that insight. We'll follow up when we have an update or more questions!

sandrooco:
Maybe another detail, I just stumbled across the Angular release schedule: https://angular.io/guide/releases#support-policy-and-schedule
Angular itself doesn't support 10 anymore...

jasonnutter:
> Maybe another detail, I just stumbled across the Angular release schedule: https://angular.io/guide/releases#support-policy-and-schedule Angular itself doesn't support 10 anymore...

Correct, and as such, Angular 10 will not be supported in the next major version for MSAL Angular, but will continue to be supported by MSAL Angular v2 (as dropping support would technically be a breaking change).

jasonnutter:
@kyubisation We have merged and published #4154 (as `@azure/msal-angular@2.2.0`). Please try that out and let us know if that addresses baseline Angular 13 compatibility. If so, we can then look at incrementally addressing other concerns such as Ivy distribution and the type fixes included in this PR (and I would prefer to have separate PRs for each of those). Thanks!

jasonnutter:
Closing, please open a new PR to incrementally address issues/features not addressed in the latest version of MSAL Angular.

PR #4616, opened 2022-03-21
Add Client Application telemetry parameters to MSAL.js (Node.js and Browser-based Libraries)

pkanher617:
This PR adds two telemetry parameters to the query string of the OAuth URLs generated (`/authorize` and `/token`) - `x-app-name` (Name of the client application, i.e. "MSAL.js Browser Sample") and `x-app-ver` (Version of the application using semver, i.e. `1.0.0`). The data for these parameters are provided in the configuration of the MSAL object and are included in all generated URLs, similar to the other `x-client-...` parameters currently being sent. 

Part of this PR involves a new Configuration object called `telemetry` which will house our telemetry configurations moving forward for both `msal-browser` and `msal-node`.

This PR also includes substitutions of the empty string `""` character with a constant from the Constants export of `@azure/msal-common`.


lib/msal-browser/src/config/Configuration.ts
```diff
@@ -161,10 +167,18 @@ export function buildConfiguration({ auth: userInputAuth, cache: userInputCache,
         allowRedirectInIframe: false
     };
 
+    const DEFAULT_TELEMETRY_OPTIONS: Required<BrowserTelemetryOptions> = {
+        clientTelemetry: {
```
jasonnutter: Instead of `telemetry.clientTelemetry`, I might consider something like `telemetry.application`, e.g. `telemetry.application.name`, etc.

jasonnutter: Then we could have `telemetry.performance`, `telemetry.server`, etc

tnorling (approved):

lib/msal-browser/docs/configuration.md
```diff
@@ -95,9 +101,22 @@ const msalInstance = new PublicClientApplication(msalConfig);
...
+| Option | Description | Format | Default Value |
+| ------ | ----------- | ------ | ------------- |
+| `application` | Telemetry options for applications using MSAL.js | See [below](#application-telemetry) | See [below](#application-telemetry) |
+
+#### Application Telemetry
+
+| Option | Description | Format | Default Value |
+| ------ | ----------- | ------ | ------------- |
+| `appName` | Unique string name of an application | string | Empty string "" |
+| `appVersion` | Version of the application using MSAL | string | Empty string "" |
```
jasonnutter, line 122: I'd like to stop adding to this file, and instead generate proper JS doc comments for each property and have this file link the typedocs.

jasonnutter: Can be done incrementally from this PR.

jasonnutter (approved):

bmahall (approved):

lib/msal-browser/docs/configuration.md
pkanher617: I will update typedocs as well, we should deprecate these docs and ensure our typedocs have enough info in a follow-up PR

lib/msal-browser/src/config/Configuration.ts
pkanher617: done

PR #4617, opened 2022-03-21
[msal-browser] Fix persistent key store deletion on logout

### LINKED ISSUE #4564: logoutRedirect() gives --> "Error - Attempted to clear all MSAL cache items and failed. Local cache unchanged."
### Core Library

### Core Library Version

2.22.0

### Wrapper Library

### Wrapper Library Version

2.1.1

### Description

Everything seems to be working fine, but when you do msalService.logoutRedirect() you will get an error (in Chrome browser console + preserve logs):
**"Error - Attempted to clear all MSAL cache items and failed. Local cache unchanged.".** 

After some research I managed to find out the there it at least a "not a function" exception in CryptoOps.ts when doing databaseStorage.deleteDatabase():

typeerror

### Error Message

1. Error - Attempted to clear all MSAL cache items and failed. Local cache unchanged.
2. databaseStorage.deleteDatabase is not a function

### Msal Logs

_No response_

### MSAL Configuration

```javascript
auth: {
      clientId: <myappid>,
      authority: `https://login.microsoftonline.com/${<mytenantid>}`,
      redirectUri: '<myredirecturl>',
      postLogoutRedirectUri: '<mylogouturl>',
      navigateToLoginRequestUrl: false
    },
    cache: {
    },
    system: {
      loggerOptions: {
          logLevel: LogLevel.Error,
          loggerCallback: (level, message) => {
              console.log(level, message);
          },
          piiLoggingEnabled: false
      }
    }
```

### Relevant Code Snippets

```javascript
msalService.logoutRedirect();
```

### Reproduction Steps

1. Open Chrome developer tools
2. Open console
3. Toggle "preserve logs"
4. msalService.LoginRedirect();
5. msalService.logoutRedirect();
6. See the console output for error.

### Expected Behavior

1. databaseStorage.deleteDatabase() should not cause "is not a function"
2. logoutRedirect() should not cause an error message

### Identity Provider

Azure AD / MSA

Chrome

### Regression

_No response_

### Source

External (Customer)

hectormmg:
This PR:
- Fixes #4564 
- Adds logic to close the IndexedDB connection after every transaction, whether successful or unsuccessful, to prevent `deleteDatabase` calls from being blocked by unclosed connections
- Adds comprehensive logging on key store deletion steps for easier debugging


lib/msal-browser/src/cache/AsyncMemoryStorage.ts
```diff
@@ -112,12 +116,34 @@ export class AsyncMemoryStorage<T> implements IAsyncStorage<T> {
...
+            return false;
         }
     }
+
+    /**
+     * Helper that prints the keystore name in a log-friendly way
+     * @returns 
+     */
+    private logStoreName() {
+        return (this.name) ? `"${this.name}"` : Constants.EMPTY_STRING;
```
jasonnutter: Did you mean to include the quotes in teh message?

jasonnutter: if not, could also be:

```suggestion
        return this.name || Constants.EMPTY_STRING;
```

lib/msal-browser/src/cache/AsyncMemoryStorage.ts
```diff
@@ -112,12 +116,34 @@ export class AsyncMemoryStorage<T> implements IAsyncStorage<T> {
...
+            }
+            return false;
         }
     }
+
+    /**
+     * Helper that prints the keystore name in a log-friendly way
+     * @returns 
+     */
+    private logStoreName() {
```
jasonnutter: Return type

lib/msal-browser/src/cache/AsyncMemoryStorage.ts
```diff
@@ -17,16 +17,20 @@ export class AsyncMemoryStorage<T> implements IAsyncStorage<T> {
     private inMemoryCache: MemoryStorage<T>;
     private indexedDBCache: DatabaseStorage<T>;
     private logger: Logger;
+    private name?: string;
 
-    constructor(logger: Logger) {
+    constructor(logger: Logger, name?: string) {
```
jasonnutter: `name` is a little bit vague, is there a more specific name to use here?

lib/msal-browser/src/interaction_client/BaseInteractionClient.ts
```diff
@@ -54,11 +54,11 @@ export abstract class BaseInteractionClient {
             }
         } else {
             try {
+                this.logger.verbose("No account provided in logout request, clearing all cache items.");
```
jasonnutter: Can you add `this.correlationId` here?

```suggestion
                this.logger.verbose("No account provided in logout request, clearing all cache items.", this.correlationId);
```

lib/msal-browser/src/cache/AsyncMemoryStorage.ts
hectormmg: I did since it's meant to indicate the given name of the key store... Do we prefer to not make the distinction?

Basically: Deleting keystore "asymmetricKeys" / Keystore "asymmetricKeys" deleted

vs

Deleting asymmetricKeys keystore / Keystore asymmetricKeys deleted

lib/msal-browser/src/cache/AsyncMemoryStorage.ts
hectormmg: Oops, adding it

lib/msal-browser/src/cache/AsyncMemoryStorage.ts
hectormmg: storeName?

lib/msal-browser/src/cache/AsyncMemoryStorage.ts
jasonnutter: Yeah not a big deal to me, log message are for us and we know the intent here.

lib/msal-browser/src/crypto/CryptoOps.ts
```diff
@@ -144,9 +149,23 @@ export class CryptoOps implements ICrypto {
...
+            this.logger.verbose("Deleting in-memory and persistent asymmetric key stores");
+            await this.cache.asymmetricKeys.clear();
+            this.logger.verbose("Successfully deleted asymmetric key stores");
+            this.logger.verbose("Deleting in-memory and persistent symmetric key stores");
+            await this.cache.symmetricKeys.clear();
+            this.logger.verbose("Successfully deleted symmetric key stores");
+            return true;
+        } catch (e) {
+            if (e instanceof Error) {
+                this.logger.error("Clearing keystore failed with error: ", e.message);
```
jasonnutter: This passes `e.message` as correlationId, which I don't think you meant. 

```suggestion
                this.logger.error(`Clearing keystore failed with error: ${e.message}`);
```

lib/msal-browser/src/cache/AsyncMemoryStorage.ts
jasonnutter: Also, can this be required?

lib/msal-browser/src/cache/AsyncMemoryStorage.ts
hectormmg: Since we're going with a required storeName attribute, this method is no longer necessary, replacing with the regular property accessor everywhere it's called.

lib/msal-browser/src/cache/AsyncMemoryStorage.ts
hectormmg: Method was removed since storeName isn't optional anymore

change/@azure-msal-browser-6b150027-dba7-407a-a266-afa6dc7fe44f.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
jasonnutter: This seems more like a bug fix than a new feature (esp given public API isnt changing)

hectormmg: Right, I confused myself since we added a required attribute to a "public" (but not public facing) API.

lib/msal-browser/src/cache/AsyncMemoryStorage.ts
```diff
@@ -112,12 +116,26 @@ export class AsyncMemoryStorage<T> implements IAsyncStorage<T> {
     /**
      * Clears in-memory Map and tries to delete the IndexedDB database.
      */
-    async clear(): Promise<void> {
+    async clear(): Promise<boolean> {
+    // InMemory cache is a Map instance, clear is straightforward
```
tnorling: Can you tab this in?

lib/msal-browser/src/cache/AsyncMemoryStorage.ts
```diff
@@ -112,12 +116,26 @@ export class AsyncMemoryStorage<T> implements IAsyncStorage<T> {
...
+            const dbDeleted = await this.indexedDBCache.deleteDatabase();
+            
+            if (dbDeleted) {
+                this.logger.verbose(`Persistent keystore ${this.storeName} deleted`);
+            }
+            
+            return dbDeleted;
         } catch (e) {
-            this.handleDatabaseAccessError(e);
+            if (e instanceof Error) {
```
tnorling: Why is this check needed? Can you add a comment explaining this?

lib/msal-browser/src/cache/DatabaseStorage.ts
```diff
@@ -71,20 +88,26 @@ export class DatabaseStorage<T> implements IAsyncStorage<T> {
      */
     async getItem(key: string): Promise<T | null> {
         await this.validateDbIsOpen();
-
         return new Promise<T>((resolve, reject) => {
             // TODO: Add timeouts?
-            if (!this.db) {
+            const db = this.db;
```
tnorling: Can you please explain this?

change/@azure-msal-common-c0850675-6450-475c-b9fe-8bc4074f8ffd.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
tnorling: Looks like the changes to common are only package-lock updates?

lib/msal-browser/src/cache/DatabaseStorage.ts
jasonnutter: I was also curious as to why this was needed everywhere.

jasonnutter (approved):
Thanks for addressing feedback. Same feedback as Thomas, but looks good otherwise!

lib/msal-browser/src/cache/DatabaseStorage.ts
hectormmg: I've been trying to figure out a way to avoid a race that happens when we try to delete the database when there are pending transactions (and thus database connections). Explicitly assigning the db property to a const works for some reason (I'm assuming it adds the extra few ms required for all connections to close). 

In theory waiting for the database deletion isn't necessary since IDB guarantees it'll be deleted once the pending connections are closed (which happens on redirect), but since I want our logs to properly document the database deletion (to help diagnose bugs like the one this PR fixes), I had to figure out a way to always get a logged confirmation that the database was deleted.

This does the trick and I will add some comments to that effect. It's a very rough patch so I'd be okay with spending more time figuring out other ways to achieve this if we don't like it.

lib/msal-browser/src/cache/AsyncMemoryStorage.ts
hectormmg: It wasn't needed, I think I added it during debugging and didn't remove it.

lib/msal-browser/src/cache/DatabaseStorage.ts
tnorling: We should address the race rather than attempting to avoid it this way

hectormmg: I undid the changes related to the constant assignment and I haven't been able to repro the race that was happening. It seems something else must have been off when I was testing before.

change/@azure-msal-common-c0850675-6450-475c-b9fe-8bc4074f8ffd.json
hectormmg: Reset the package-lock and removed the changefile

tnorling (approved):

PR #4620, opened 2022-03-22
[msal-node-extensions]: Resolve minimist audit error

samuelkubai:
Bumping up the version of minimist to 1.2.6 to resolve the audit error


jasonnutter:
Looks like an unrelated set of changes were pulled in. Can you please reduce to just the package lock change needed for extensions? New pr might be needed. Thanks!

PR #4627, opened 2022-03-23
Add SHR support for JS-WAM bridge

sameerag:
This feature adds the below support to the token binding work on MSAL Browser:

- Adds support for POP tokens
- Generates SHR in the `msal-browser` in case native layer doesn't support it


lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -236,6 +265,22 @@ export class NativeInteractionClient extends BaseInteractionClient {
 
         const instanceAware: boolean = !!(request.extraQueryParameters && request.extraQueryParameters.instance_aware);
 
+        // validate SHR parameters
+        if (request.authenticationScheme === AuthenticationScheme.POP) {
+            if (!request.resourceRequestMethod && !request.resourceRequestUri) {
```
jasonnutter: I thought these were technically optional?

jasonnutter: Or does wam shr require them?

jasonnutter (commented):
Are there tests that be added? Or will those be added later?

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -24,16 +24,18 @@ import { BrowserAuthError } from "../error/BrowserAuthError";
 export class NativeInteractionClient extends BaseInteractionClient {
     protected apiId: ApiId;
     protected nativeMessageHandler: NativeMessageHandler;
+    protected cryptoObj: ICrypto;
 
     constructor(config: BrowserConfiguration, browserStorage: BrowserCacheManager, browserCrypto: ICrypto, logger: Logger, eventHandler: EventHandler, navigationClient: INavigationClient, apiId: ApiId, provider: NativeMessageHandler, correlationId?: string) {
         super(config, browserStorage, browserCrypto, logger, eventHandler, navigationClient, provider, correlationId);
         this.apiId = apiId;
         this.nativeMessageHandler = provider;
+        this.cryptoObj = browserCrypto;
```
tnorling: This is already a part of `BaseInteractionClient` as `this.browserCrypto`

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -404,6 +404,21 @@ export class AuthorizationCodeClient extends BaseClient {
             parameterBuilder.addExtraQueryParameters(request.extraQueryParameters);
         }
 
+        if (request.nativeBroker) {
```
tnorling, line 407: This is also present on line 422 - can we combine?

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
sameerag: SHR always required them as far as I can check. Only `shrClaims` and `shrNonce` are optional.

sameerag: So yes, in future, when WAM will calculate the SHR, we need to mandate them. We never needed to before, as these never left the browser layer.

sameerag:
> Are there tests that be added? Or will those be added later?

Later. Thomas is manually testing this. I will have a cache PR out soon which will include some unit tests for SHR too.

lib/msal-common/src/client/AuthorizationCodeClient.ts
sameerag: Done, merge issue. Thanks for catching this.

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
sameerag: Yes. Edited.

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
sameerag: This is from the docs, the code does not restrict anything as such. @hectormmg let's chat about this. I am also following up with WAM team on what they would like to mandate in the future. For now, my preference is to keep this and do an additional PR later with this data to proceed with a single branch testing for WAM JS Bridge.

sameerag: Reverting this, the decision is pending based on WAM response, for now, keeping the public surface intact.

tnorling (approved):

PR #4680, opened 2022-04-05
fix: Allow direct access to Identity Token Claims

### LINKED ISSUE #4538: Change type of `idTokenClaims` for direct access
### Core Library

### Wrapper Library

### Description

If the `idTokenClaims` are set on an account, it's recognized as an `object`. This means you cannot access it's members (`roles doesn't exist on object`). With the following work-around I'm able to access the properties in the identity token.

```js
// Doesn't work
const roles = account.idTokenClaims.roles;
// or
const roles2 = account.idTokenClaims['roles'];
// this does work
const claims = account.idTokenClaims as { [key: string]: any };
const roles3 = claims.roles;
// and this does also work:
const roles4 = claims['roles'];
```

The type `{ [key: string]: any }` is valid for each json document, and since the idTokenClaims are always represented as json, this would be a better description, and allow access to all the claims without manually changing the type. It would also support future claims that are not known yet.

An alternative would be to create something like this, but that requires more maintenance:

```js
declare type IdentityToken {
  aud?: string;
  exp?: number;
  iss?: string;
  name?: string;
  roles?: string[];
}

// Or supporting both methods
declare type IdentityTokenExtension  = IdentityToken & {  [key: string]: any; }
```

### Source

External (Customer)

svrooij:
Currently you cannot access claims in the Identity token because it's specified as an `object`. I've created a object containing all the common claims in an Identity Token and set it as `& { [key: string]: string | number | string[] | undefined | unknown };` for fallback to claims not defined in the list, which should support anything that matches the definition of any json string.

Fixed: #4538


lib/msal-common/src/account/IdentityToken.ts
```diff
@@ -0,0 +1,42 @@
...
+ * - given_name             - Given name
+ * - idp                    - Identity provider, is the same as the issuer unless it's a guest user.
+ * - ipaddr                 - Users' IP address
+ * - name                   - Mutable username
+ * - oid                    - Immutable object identifier, this ID uniquely identifies the user across applications
+ * - sub                    - Immutable subject identifier, this is a pairwise identifier - it is unique to a particular application ID
+ * - tid                    - The Users' tenant, or '9188040d-6c67-4c5b-b112-36a304b66dad' for personal accounts.
+ * - roles                  - The application roles the user has assigned
+ */
+export type IdentityToken = {
```
tnorling: This is already defined in `TokenClaims` [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-common/src/account/TokenClaims.ts). Not sure if it's exported from msal-common currently, however

svrooij: As discussed with @jasonnutter I was invited to create a PR, see https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/4538#issuecomment-1075465326

This PR would allow access to IdentityToken claims, currently this is exposed as `object` in the **AccountInfo**, with these changes the user has direct access to the identity token claims. as discussed [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/4538#issue-1146825813)

svrooij: Also when checking the [TokenClaims](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-common/src/account/TokenClaims.ts) I see a lot of missing claims and claims that won't be in the token anyway.

tnorling: I'm suggesting to update the existing `TokenClaims` object with your desired behavior rather than creating a duplicate.

svrooij:
The failing job for `node-token-validation` is also failing in the dev branch. Probably has something to do with the build order. It expects the `@azure/msal-common` package which isn't present (yet). That job failing has nothing to do with this change to that package.

lib/msal-common/src/account/AccountInfo.ts
```diff
@@ -21,5 +22,5 @@ export type AccountInfo = {
     username: string;
     localAccountId: string;
     name?: string;
-    idTokenClaims?: object;
+    idTokenClaims?: TokenClaims & { [key: string]: string | number | string[] | undefined | unknown };
```
jasonnutter: `object` is also valid.

```suggestion
    idTokenClaims?: TokenClaims & { [key: string]: string | number | string[] | object | undefined | unknown };
```

change/@azure-msal-common-0a241400-d057-4c0b-8d02-318aebb9421a.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "fix: Allow direct access to Identity Token Claims",
```
jasonnutter:
```suggestion
  "comment": "fix: Allow direct access to Identity Token Claims #4680",
```

jasonnutter:
> The failing job for `node-token-validation` is also failing in the dev branch. Probably has something to do with the build order. It expects the `@azure/msal-common` package which isn't present (yet). That job failing has nothing to do with this change to that package.

Okay thanks we'll take a look. cc @jo-arroyo

jo-arroyo:
> The failing job for `node-token-validation` is also failing in the dev branch. Probably has something to do with the build order. It expects the `@azure/msal-common` package which isn't present (yet). That job failing has nothing to do with this change to that package.

Build has been fixed. Please update with base branch.

jasonnutter (approved):

svrooij:
@sameerag and @tnorling do you get automatic notifications if someone else approves a PR?

tnorling (approved):

jviney:
Can we apply the same interface to `AuthenticationResult.idTokenClaims`? I can do a PR.

svrooij:
> Can we apply the same interface to `AuthenticationResult.idTokenClaims`? I can do a PR.

>
> @jviney 

Sure you can, that is how I started, finally settled on doing it only on the account info. That is because the change you're suggestion will require changes in several places.

I'm not a code owner, so I can't say what will and won't be accepted but I like the suggestion.

jasonnutter:
> Can we apply the same interface to `AuthenticationResult.idTokenClaims`? I can do a PR.

@jviney Sure, feel free to open a PR.

PR #4691, opened 2022-04-07
Fix obo cache filters

### LINKED ISSUE #4486: ClientAuthError: multiple_matching_tokens error when cache kept more than 2 users
### Core Library

MSAL Node (@azure/msal-node)

### Core Library Version

1.5.0

### Wrapper Library

Not Applicable

### Wrapper Library Version

None

### Description

When using a "singleton instance of confidentialClientApplication" we have found the "case that in cache there is two accounts, third account try to make first time a OBO and we get multiple_matching_tokens error."

Originally posted by @CrazyBaran in https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/4158#issuecomment-944152234

With my team, we have found exactly the same issue... Maybe important to say that the cache has exactly 12 items when it happens.

@sameerag We are wondering if the potential solution is still alive, for it seems recently to be more or less without activity: https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/4081

And by the way, the only quick workaround we've found so far is to always clear the cache XD
Have you by chance found another one @CrazyBaran ?

### MSAL Configuration

```javascript
const config: Configuration = {
  auth: {
    clientId: 'our clientId',
    authority: 'https://login.microsoftonline.com/our tenantId',
    clientSecret: 'our clientSecret',
  },
};
this.cca = new ConfidentialClientApplication(config);
```

### Relevant Code Snippets

```javascript
const cachedAccount: AccountInfo = await this.cca
      .getTokenCache()
      .getAccountByLocalId((jwtDecode(token) as UserPayload).oid);
    if (cachedAccount?.localAccountId) {
      return await this.cca.acquireTokenSilent({
        account: cachedAccount,
        scopes: ['https://vault.azure.net/.default'],
        authority: 'https://login.microsoftonline.com/our tenantId',
      });
    }
	return this.cca.acquireTokenOnBehalfOf({
      oboAssertion: token.split(' ').pop(),
      scopes,
      authority: 'https://login.microsoftonline.com/our tenantId',
    });
```

### Identity Provider

Azure AD / MSA

### Source

External (Customer)

sameerag:
Adding `user_assertion_hash` in `access_token` and `refresh_token` instead of `oboAssertion`. This will add two advantages to the OBO flow:

- Shortens the assertion length in cache
- Helps with looking up cached OBO tokens only when the incoming_assertion matches
- Removes assertion in `id_token` as not all scenarios would have this returned from the service and `idToken` is not assertion specific

Please note that this PR scopes `msal-node` OBO flow to short term OBO only. Assertions are valid only for the duration of AT validity and we invoke interaction after that. The RT flow will be added when `msal-node` supports distributed cache natively (since this would mean partitioning the cache based on the assertion hash). We do not want to add this for public clients and cause scalability issues.


andrew-hwahin:
@sameerag @tnorling thanks for the effort, when will the fixes merge into production? 🙏

change/@azure-msal-browser-1b5c87e6-d99e-462a-9597-e8e9c3ee0185.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
```
tnorling: Why are these all change type none?

sameerag: Typo, will update.

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -711,7 +711,7 @@ export abstract class CacheManager implements ICacheManager {
      * @param scopes
      * @param authScheme
      */
-    readAccessTokenFromCache(clientId: string, account: AccountInfo, request: BaseAuthRequest): AccessTokenEntity | null {
+    readAccessTokenFromCache(clientId: string, account: AccountInfo, request: BaseAuthRequest, hash?: string): AccessTokenEntity | null {
```
hectormmg: I think we now have a few different hashes involved, can we name this something more specific?

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -752,14 +753,15 @@ export abstract class CacheManager implements ICacheManager {
      * @param account
      * @param familyRT
      */
-    readRefreshTokenFromCache(clientId: string, account: AccountInfo, familyRT: boolean): RefreshTokenEntity | null {
+    readRefreshTokenFromCache(clientId: string, account: AccountInfo, familyRT: boolean, hash?: string): RefreshTokenEntity | null {
```
hectormmg: Same as above re: multiple hashes

lib/msal-common/src/cache/entities/CredentialEntity.ts
```diff
@@ -81,7 +81,7 @@ export class CredentialEntity {
             this.target,
             this.familyId,
             this.tokenType,
-            this.requestedClaimsHash
+            this.requestedClaimsHash,
```
hectormmg, line 84: Did you mean to add the `userAssertionHash` to the key?

change/@azure-msal-browser-1b5c87e6-d99e-462a-9597-e8e9c3ee0185.json
sameerag: Done

lib/msal-common/src/cache/entities/AccountEntity.ts
```diff
@@ -143,7 +141,6 @@ export class AccountEntity {
         homeAccountId: string,
         idToken: AuthToken,
         authority?: Authority,
-        oboAssertion?: string,
```
tnorling, line 146: This is a breaking change

lib/msal-common/src/cache/CacheManager.ts
sameerag: Ideally this can be avoided I guess, as we have a separate look up added in the OBOClient. We need to have this only in the filter code. I had hopes to reuse the existing look-up but it adds a lot of refactoring which I am not inclined to make at this point. Let me see if I can remove this here.

lib/msal-common/src/cache/entities/AccountEntity.ts
sameerag: I will note that it is a major version for `common`. However public API should not change. `AccountInfo` is exposed to users and not `AccountEntity`.

lib/msal-common/src/cache/CacheManager.ts
sameerag: Removed this, not necessary for this flow.

lib/msal-common/src/cache/CacheManager.ts
sameerag: This will be needed when we start supporting RT flow. Removing as it is not needed now.

lib/msal-common/src/cache/entities/CredentialEntity.ts
sameerag: No. Initially considered it, but since we are supporting only short term OBO, it is not necessary now. Will need it once we add RT flows.

bmahall (approved):
To add additional test cases, rest looks good

lib/msal-common/src/cache/entities/AccountEntity.ts
sameerag: Done

lib/msal-browser/test/cache/TokenCache.spec.ts
```diff
@@ -123,7 +123,8 @@ describe("TokenCache tests", () => {
                 clientInfo: testClientInfo
             };
             const testIdAuthToken = new AuthToken(testIdToken, cryptoObj);
-            const testAccount = AccountEntity.createAccount(testClientInfo, TEST_DATA_CLIENT_INFO.TEST_HOME_ACCOUNT_ID, testIdAuthToken, undefined, undefined, undefined, undefined, testEnvironment);
+            console.log(testEnvironment);
```
jasonnutter:
```suggestion
```

lib/msal-common/test/client/OnBehalfOfClient.spec.ts
```diff
@@ -92,244 +123,48 @@ describe("OnBehalfOf unit tests", () => {
...
-        const client = new OnBehalfOfClient(config);
-        const onBehalfOfRequest: CommonOnBehalfOfRequest = {
-            authority: TEST_CONFIG.validAuthority,
-            correlationId: TEST_CONFIG.CORRELATION_ID,
-            scopes: TEST_CONFIG.DEFAULT_GRAPH_SCOPE,
-            oboAssertion: TEST_TOKENS.ACCESS_TOKEN,
-            skipCache: false
-        };
-
-        client.acquireToken(onBehalfOfRequest);
```
jasonnutter, line 117: Why were these tests deleted?

jasonnutter: Also, is there a test that asserts you dont get a token back for the wrong user?

lib/msal-common/src/client/OnBehalfOfClient.ts
```diff
@@ -147,19 +143,49 @@ export class OnBehalfOfClient extends BaseClient {
...
      */
-    private readAccountFromCache(account: AccountInfo): AccountEntity | null {
-        return this.cacheManager.readAccountFromCache(account);
+    private readAccessTokenFromCacheForOBO(clientId: string, request: CommonOnBehalfOfRequest) {
+        const authScheme = request.authenticationScheme || AuthenticationScheme.BEARER;
+        /*
+         * Distinguish between Bearer and PoP/SSH token cache types
+         * Cast to lowercase to handle "bearer" from ADFS
+         */
+        const credentialType = (authScheme && authScheme.toLowerCase() !== AuthenticationScheme.BEARER.toLowerCase()) ? CredentialType.ACCESS_TOKEN_WITH_AUTH_SCHEME : CredentialType.ACCESS_TOKEN;
```
peterzenz, line 157: is there a reason AuthenticationScheme.BEARER isn't lowercased in the definition and we change the rest of the code base to read 
```ts
(authScheme && authScheme.toLowerCase() !== AuthenticationScheme.Bearer)
```

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -302,10 +300,10 @@ export class ResponseHandler {
      * @param stateString
      */
     static async generateAuthenticationResult(
-        cryptoObj: ICrypto, 
+        cryptoObj: ICrypto,
         authority: Authority,
-        cacheRecord: CacheRecord, 
-        fromTokenCache: boolean,
```
peterzenz, line 308: There's a ton of little cleanup in here like addressing extra whitespace.  Suggest creating a separate PR just for code cleanup/refactoring in the future.  Makes the core functional changes a smaller PR which is often easier to review.

peterzenz (approved):

tnorling (approved):

lib/msal-common/src/response/ResponseHandler.ts
jasonnutter: Guessing the whitespace changes were done automatically by the IDE. FYI you can hide whitespace changes from the diff here: 
image

lib/msal-common/src/client/OnBehalfOfClient.ts
jasonnutter: We've unfortunately been a bit inconsistent with casing for enums.

lib/msal-common/src/response/ResponseHandler.ts
sameerag: Yep. It is my IDE setup. I do think however we need a the same IDE settings as a standard we can import and apply if we want to avoid this in PRs.

lib/msal-browser/test/cache/TokenCache.spec.ts
sameerag: Addressed

lib/msal-common/test/client/OnBehalfOfClient.spec.ts
sameerag: @bmahall reorganized some tests. I added the tests for the `filtering` regarding `access_token`.

PR #4711, opened 2022-04-14
quick fix for timing issue

### LINKED ISSUE #4189: Token expiry not calculated when sideloading token
### Core Library

### Core Library Version

2.18.0

### Wrapper Library

Not Applicable

### Wrapper Library Version

None

### Description

When sideloading a token the `expires_in` value is used for the `expiresOn` field in the token cache. This causes values like `3600` to be checked as a timestamp and be deemed invalid. The value of `extendedExpiresOn` that is also required is ignored completely when the token is eventually validated by `SilentFlowClient`.

### Error Message

_No response_

### Msal Logs

_No response_

### MSAL Configuration

```javascript
{
  auth: {
    clientId: process.env.VUE_APP_AUTH_CLIENT,
    authority: "https://stfs.company.com/adfs",
    knownAuthorities: ["https://stfs.company.com/adfs"],
    redirectUri: process.env.BASE_URL,
    protocolMode: "OIDC"
  },
  cache: {
    storeAuthStateInCookie: true
  }
}
```

### Relevant Code Snippets

```javascript
// https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/7b5f2eaef2eede43f23765b97547922f0024437d/lib/msal-browser/src/cache/TokenCache.ts#L137
const expiresOn = response.expires_in;

// Token expiry is validated in `SilentFlowClient`
// https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/51e8f1bbc286c3ce088ec5659deb186822e18be3/lib/msal-common/src/client/SilentFlowClient.ts#L74
TimeUtils.isTokenExpired(cacheRecord.accessToken.expiresOn, this.config.systemOptions.tokenRenewalOffsetSeconds);
```

### Reproduction Steps

1. Attempt to sideload a token with a response containing `"expires_in": 3600`
2. Set `extendedExpiresOn` to a timestamp in the future

### Expected Behavior

The token should be fetched from cache as valid.

### Identity Provider

ADFS

Firefox

### Regression

_No response_

### Source

External (Customer)

### LINKED ISSUE #4710: loadExternalTokens incorrectly assigns expires_in to expiresOn
### Core Library

### Core Library Version

2.23.0

### Wrapper Library

### Wrapper Library Version

1.3.2

### Description

In the code to side-load an access token the `expires_in` field from the server response is written to the cache as `expiresOn`. This causes all access tokens to fail validation because they appear to be out of date (at a time very close to epoch time).

expiresOn is a date
expires_in is a number of seconds

### Error Message

_No response_

### Msal Logs

_No response_

### MSAL Configuration

```javascript
{ not-relevant }
```

### Relevant Code Snippets

```javascript
https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/f8f7c2068905e9ce2be9046400f808d208bf5573/lib/msal-browser/src/cache/TokenCache.ts#L138
```

### Reproduction Steps

Using loadExternalTokens will usually fail due to a number of errors because of this

### Expected Behavior

expiresOn is written as expires_in + current date as a workaround

### Identity Provider

Azure AD / MSA

Chrome

### Regression

_No response_

### Source

External (Customer)

prsauer:
See https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/4710

This is a quick fix for the issue since this functionality is only needed for E2E mocks/testing


change/@azure-msal-browser-0f2774f8-a6da-47d6-9814-983561d0272e.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fixes assignment of expiresOn inside loadExternalTokens to fix issue where all access tokens appear expired during E2E",
```
jasonnutter:
```suggestion
  "comment": "Fixes assignment of expiresOn inside loadExternalTokens to fix issue where all access tokens appear expired during E2E #4711",
```

jasonnutter (changes requested):
Thanks for contributing! Can you please add a test (or update an existing test) to demonstrate the fix? https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/009baa137cc1efdacf7a4de1f1fbb81a42fe18a6/lib/msal-browser/test/cache/TokenCache.spec.ts#L46

prsauer:
> Thanks for contributing! Can you please add a test (or update an existing test) to demonstrate the fix?
> 
> https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/009baa137cc1efdacf7a4de1f1fbb81a42fe18a6/lib/msal-browser/test/cache/TokenCache.spec.ts#L46

Changed it a bit to make the testing make sense. I think this strategy is cleaner unless the token loader wants to decode the jwt to read the exp claim internally.

I'd be ok making `expiresOn` a required option as well -- the token that's written to the cache isn't very useful with the incorrect value there.

lib/msal-browser/src/cache/TokenCache.ts
```diff
@@ -135,7 +136,7 @@ export class TokenCache implements ITokenCache {
         }
 
         const scopes = new ScopeSet(request.scopes).printScopes();
-        const expiresOn = response.expires_in;
+        const expiresOn = options.expiresOn ? options.expiresOn : response.expires_in;
```
jasonnutter: Nitpick:

```suggestion
        const expiresOn = options.expiresOn || response.expires_in;
```

jasonnutter: Also, @jo-arroyo did we consider setting a default expiration (e.g. 1 hour)?

jasonnutter (approved):

lib/msal-browser/src/cache/TokenCache.ts
```diff
@@ -135,7 +136,7 @@ export class TokenCache implements ITokenCache {
         }
 
         const scopes = new ScopeSet(request.scopes).printScopes();
-        const expiresOn = response.expires_in;
+        const expiresOn = options.expiresOn || response.expires_in;
```
tnorling: Should the right half of this operation be updated to calculate `expiresOn` from `expires_in`?

prsauer: Potentially -- we'd have to add the request timestamp somewhere in scope to make the calculation. This could go into `ServerAuthorizationTokenResponse` (a root for the `ExternalTokenResponse` type) or even `LoadTokenOptions`.

It's worth noting though that the interface creating the "real" tokens is calling `createAccessTokenEntity` and setting expiresOn directly as the callsite has request timing in context already.

I was a bit hesitant to make a change to the real behavior to fix a bug that only occurs in the loading of external tokens; open to suggestions though. Another consideration is potentially making `expiresOn` required or adding `requestTimestamp` into `LoadTokenOptions` to make it clearer to consumers of `loadExternalTokens` that those data are required to get correct timing behavior.

tnorling: Since this is merely for testing purposes, I wonder if we can just use now (or now minus a buffer, say 5 minutes) as the request timestamp. Presumably your test suite would be requesting a token, then immediately loading it, using it once and then moving on to the next test. We don't need to be super accurate for this scenario and for those that prefer a more accurate timestamp they can still provide it explicitly with your change here. Thoughts? @jasonnutter also curious your thoughts here

jasonnutter: I would say default to now and allow an explicit request timestamp to be optionally passed in.

prsauer: So something like `const expiresOn = options.expiresOn || now + response.expires_in;`?

tnorling: Yeah something like that should work

prsauer: @tnorling @jasonnutter Ok finally got around to pushing - https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/4711/commits/9788d91a41c3adcde125045e0b5a8050f12f7635

tnorling (approved):

PR #4717, opened 2022-04-15
Perf telemetry js wam bridge

sameerag:
This PR adds performance telemetry measurements to JS-WAM Bridge code:

- Adds measurements to `NativeInteractionClient`.
- Adds measurements to ests call with `native brokering` on.
- Adds measurements to top level API for `acquireTokenPopup`.


lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -138,10 +152,17 @@ export class NativeInteractionClient extends BaseInteractionClient {
      * @param request
      * @param reqTimestamp
      */
-    protected async handleNativeResponse(response: NativeResponse, request: NativeTokenRequest, reqTimestamp: number): Promise<AuthenticationResult> {
+    protected async handleNativeResponse(response: NativeResponse, request: NativeTokenRequest, reqTimestamp: number, nativeAcquireTokenMeasurement?: InProgressPerformanceEvent): Promise<AuthenticationResult> {
```
tnorling: Since this is just for the error case, can we catch the error above instead of passing the measurement down?

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -246,7 +274,7 @@ export class NativeInteractionClient extends BaseInteractionClient {
      * Translates developer provided request object into NativeRequest object
      * @param request
      */
-    protected async initializeNativeRequest(request: PopupRequest|SsoSilentRequest): Promise<NativeTokenRequest> {
+    protected async initializeNativeRequest(request: PopupRequest|SsoSilentRequest, nativeAcquireTokenMeasurement?: InProgressPerformanceEvent): Promise<NativeTokenRequest> {
```
tnorling: same comment here

lib/msal-browser/src/interaction_client/PopupClient.ts
```diff
@@ -125,6 +132,15 @@ export class PopupClient extends StandardInteractionClient {
 
             if (serverParams.accountId) {
                 this.logger.verbose("Account id found in hash, calling WAM for token");
+                // end measurement for server call with native brokering enabled
+                if (fetchNativeAccIdMeasurement) {
+                    fetchNativeAccIdMeasurement.endMeasurement({
+                        success: true,
+                        isNativeBroker: true
+                    });
+                    fetchNativeAccIdMeasurement.flushMeasurement();
```
tnorling: Do we need to discard this measurement if accountId is **not** in the response?

lib/msal-browser/src/interaction_client/PopupClient.ts
```diff
@@ -97,10 +97,17 @@ export class PopupClient extends StandardInteractionClient {
             const authClient: AuthorizationCodeClient = await this.createAuthCodeClient(serverTelemetryManager, validRequest.authority, validRequest.azureCloudOptions);
             this.logger.verbose("Auth code client created");
 
+            const isNativeBroker = NativeMessageHandler.isNativeAvailable(this.config, this.logger, this.nativeMessageHandler, request.authenticationScheme);
+            // Start measurement for server calls with native brokering enabled
+            let fetchNativeAccIdMeasurement;
```
tnorling: Can we spell out "Account" here?

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -48,8 +51,19 @@ export class NativeInteractionClient extends BaseInteractionClient {
...
+
+        // generate the authentication result
+        const authResult = await this.handleNativeResponse(validatedResponse, nativeRequest, reqTimestamp, nativeATMeasurement);
+
+        // end the perf measurement
+        nativeATMeasurement.endMeasurement({
+            success: true,
+            isNativeBroker: true
+        });
+        nativeATMeasurement.flushMeasurement();
```
jasonnutter: Don't need to call `flushMeasurement` here, only needs to be called at the top level API.

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
jasonnutter: Yes, prefer not to pass measurements done, instead have them be self contained within a given function.

lib/msal-browser/src/interaction_client/PopupClient.ts
jasonnutter: `flushMeasurement` does not need to be called here.

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
sameerag: So that is the q I had. @jasonnutter is there no need to retain the measurement instance (since we started it above) for an error? Should I create a new one before the throw? I was assuming it should be done, I can sync up with you on this.

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
sameerag: What is `top level API`? Should we not flush once we no longer need the measurement? I may be mixing up the definitions, can you please point to the docs?

lib/msal-browser/src/interaction_client/PopupClient.ts
sameerag: Sure. Was trying to save length but it is long anyway. :)

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
jasonnutter: A top level API would be `acquireTokenSilent`, for example (i.e. an API invoked by teh application). Internal measurements should just call start and stop, not flush.

jasonnutter: `flushMeasurement` will emit the measurements to the application, which should not be done until the top level API is about to complete.

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -48,8 +51,19 @@ export class NativeInteractionClient extends BaseInteractionClient {
 
         const reqTimestamp = TimeUtils.nowSeconds();
         const response: object = await this.nativeMessageHandler.sendMessage(messageBody);
-        const validatedResponse: NativeResponse = this.validateNativeResponse(response);
-        return this.handleNativeResponse(validatedResponse, nativeRequest, reqTimestamp);
+        const validatedResponse: NativeResponse = this.validateNativeResponse(response, nativeATMeasurement);
+
+        // generate the authentication result
+        const authResult = await this.handleNativeResponse(validatedResponse, nativeRequest, reqTimestamp, nativeATMeasurement);
```
jasonnutter: Don't pass `nativeATMeasurement` in here. If you want to measure `handeNativeResponse` on its own, start a new measurement in that function.

jasonnutter: It looks like you are doing that beacuse `handleNativeresponse` may throw, in which case here you should catch the error, end the measurement, and then rethrow the error.

jasonnutter: Or, move this measurement inside `handleNativeResponse` completely.

lib/msal-browser/src/interaction_client/PopupClient.ts
```diff
@@ -97,10 +97,17 @@ export class PopupClient extends StandardInteractionClient {
...
+            if (isNativeBroker) {
+                fetchNativeAccIdMeasurement = this.performanceClient.startMeasurement(PerformanceEvents.FetchAccountIdWithNativeBroker, request.correlationId);
```
jasonnutter: What do you want to measure here? Not, we currently aren't capturing top level measurements for popup flows, do we want to do that?

I would prefer that you start a measurement at the beginning of this function, and then set the `isNativeBroker` property when it ends.

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
sameerag: I can modify it to a try/catch ofcourse. I thought this makes it easy to read, no special preference.

lib/msal-browser/src/interaction_client/PopupClient.ts
sameerag: I added it and then deleted since it is conditionally created for native. But you have a point, let me sync up with Jason on "flush" and "discard" definitions and will make the changes.

lib/msal-browser/src/interaction_client/PopupClient.ts
sameerag: Yes, we want to start for popup. I was trying to be least invasive, but I can start a top level measurement as needed. What we are measuring is a webflow to server with native brokering on, which returns us an account id instead of code.

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
sameerag: Got it. I used flush the same way, I wanted to clean it up when NativeMeasurement ends. 

So here will be the flow:
silent flow measurement starts -> native flow measurement starts -> native flow measurement is flushed (emitted to the application) -> silent flow measurement ends (in silentSSO)

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
sameerag: Done.

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
sameerag: Addressed

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
sameerag: Handled with try/catch

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
sameerag: Clarified `flush` functionality, the `clearance` is based on `correlationId` and hence cannot flush early. Added to the top level now.

lib/msal-browser/src/interaction_client/PopupClient.ts
sameerag: Done. Added toplevel and this measurement.

lib/msal-browser/src/interaction_client/PopupClient.ts
sameerag: Addressed

tnorling (approved):

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -49,7 +52,26 @@ export class NativeInteractionClient extends BaseInteractionClient {
...
+        } catch (e) {
+            // end the perf measurement for failure
+            nativeATMeasurement.endMeasurement({
+                success: false,
+                isNativeBroker: true
+            });
+            throw e;
+        }
+
+        return authResult;
```
jasonnutter: Nitpick: it can be a bit cleaner to put this onto the promise chain returned, e.g.: 

```suggestion
        return this.handleNativeResponse(validatedResponse, nativeRequest, reqTimestamp)
          .then((result: AuthenticationResult) => {
            nativeATMeasurement.endMeasurement({
                success: true,
                isNativeBroker: true
            });
            return result;
          })
          .catch((error) => {
            nativeATMeasurement.endMeasurement({
                success: false,
                isNativeBroker: true
            });
            throw e;
          });
```

jasonnutter (approved):

PR #4729, opened 2022-04-21
Add support for RDP using TLS with proof-of-possession

hectormmg:
This PR:

- Adds `kid` claim and value to SHR header to better comply with SHR draft spec and enable new RDP
- Fixes SHR header replacing `type` (wrong) with `typ` (correct)
- Adds JoseHeader class to better manage JSON assertion headers
- Fixes x5c_ca ID token claim type which was inaccurately set to string (a string array is expected since it's a certificate chain).
- Adds tests to cover changes


lib/msal-browser/src/crypto/CryptoOps.ts
```diff
@@ -203,8 +203,10 @@ export class CryptoOps implements ICrypto {
         // Generate header
         const header = {
             alg: publicKeyJwk.alg,
-            type: KEY_FORMAT_JWK
+            type: KEY_FORMAT_JWK,
+            kid: kid
```
jasonnutter: Any tests you add for this?

hectormmg: The spec doesn't really require any of the header attributes as mandatory, so I'm not sure if there's much to test here. Open to suggestions though.

sameerag: I was about to comment the same on tests. Well, can we test if we are setting the header correctly if it is set indeed?

sameerag (approved):

lib/msal-browser/src/crypto/BrowserCrypto.ts
```diff
@@ -254,6 +254,12 @@ export class BrowserCrypto {
      * @param jwk 
      */
     static getJwkString(jwk: JsonWebKey): string {
-        return JSON.stringify(jwk, Object.keys(jwk).sort());
+        const cnf = {
```
hectormmg: This is a temporary patch used to debug a server-side bug. Will be rolled back before merging.

lib/msal-browser/src/crypto/CryptoOps.ts
hectormmg: I wrote a class (JoseHeader) that generates the SHR header string so we can assign the SHR-specific values (like typ = JWT). I made sure the method that generates the SHR header throws if keyId/alg aren't added since those are required for AAD SHRs AFAIK.

As we add other assertion types, we can add errors for when their specific header values aren't provided, this way we don't force the inclusion of a header that isn't required for every JSON assertion. @

lib/msal-browser/src/crypto/BrowserCrypto.ts
hectormmg: Rolled the patch back since the server-side bug has been fixed.

lib/msal-common/src/crypto/JoseHeader.ts
```diff
@@ -0,0 +1,50 @@
...
+     * the stringified header object.
+     * Throws if keyId or algorithm aren't provided.
+     * @param shrHeaderOptions 
+     * @returns 
+     */
+    getShrHeaderString(shrHeaderOptions?: JoseHeaderOptions): string {
+        // SHR assertions must have type "JWT"
+        this.typ = JsonTypes.Jwt;
+
+        this.kid = shrHeaderOptions?.kid || this.kid;
```
jasonnutter: I'm not sure I like the pattern here, considering these are set in the constructor. Maybe make this a static method?

lib/msal-common/src/crypto/JoseHeader.ts
```diff
@@ -0,0 +1,50 @@
...
+        if (!this.kid) {
+            throw JoseHeaderError.createMissingKidError();
+        }
+
+        this.alg = shrHeaderOptions?.alg || this.alg;
+        if (!this.alg) {
+            throw JoseHeaderError.createMissingAlgError();
+        }
+
+        return JSON.stringify(this);
```
jasonnutter: Also would prefer being explicity with the fields, instead of stringifying the entire `this` object.

lib/msal-common/src/crypto/JoseHeader.ts
hectormmg: This pattern actually came up because I was trying to avoid static methods. However, I think in this case it may be better like you said, so I replaced it with a static method.

lib/msal-common/src/crypto/JoseHeader.ts
hectormmg: Still calling stringify on the JoseHeader object but now we explicitly set the specific options through the constructor so there is explicit control over which options make it into the final stirng.

jasonnutter (approved):

PR #4730, opened 2022-04-21
Add PoP support for Encrypted Access Tokens

hectormmg:
This PR:

- For default SPA PoP flow, uses the `keyId` that was stored in the cache record when the token was obtained instead of extracting the `kid` attribute from the `cnf` access token claim by decoding the access token
- For native/WAM PoP flow, uses the `keyId` stored in the `NativeRequest`
- Enables the use of token binding for encrypted access tokens because MSAL no longer needs to decode the access token to find the cached key (which isn't possible with encrypted tokens)
- Updates `PopTokenGenrator.generateCnf` API to return an object containing the `keyId`, `reqCnfString` and `reqCnfStringHash` instead of only the `reqCnfString` to make it easier for different request types to obtain the required attributes from the generated req_cnf


lib/msal-common/src/crypto/PopTokenGenerator.ts
```diff
@@ -49,16 +46,11 @@ export class PopTokenGenerator {
         };
     }
 
-    async signPopToken(accessToken: string, request: SignedHttpRequestParameters): Promise<string> {
-        const tokenClaims: TokenClaims | null = AuthToken.extractTokenClaims(accessToken, this.cryptoUtils);
```
jasonnutter, line 78: To handle a upgrade/downgrade scenario, we should consider still try this as a fallback if it is not found in the cache.

jasonnutter:
Please add tests (e.g. using an encrypted access token), thanks!

lib/msal-common/src/crypto/PopTokenGenerator.ts
hectormmg: Could you elaborate a bit? They `kid` is already being stored in the cache, so a missing `kid` would be an unexpected error even in an upgrade scenario.

Philosophically speaking, I think cracking open the access token was wrong to start with, but what are the advantages to using this as a fallback?

jasonnutter: Ah, nevermine, forgot it was already there.

bmahall (approved):
Looks good, only test file missing.

change/@azure-msal-common-2c376d9d-df7f-464b-9cf4-1b078579288f.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Use keyId from cache to find binding key #4730",
+  "packageName": "@azure/msal-common",
+  "email": "hectormg@hey.com",
```
hectormmg:
```suggestion
  "email": "hemoral@microsoft.com",
```

jasonnutter (approved):

sameerag (approved):
Lgtm. Why was this not added initially? Did we add "kid" later?

hectormmg:
> Lgtm. Why was this not added initially? Did we add "kid" later?

I wasn't here when it was implemented. My 2 theories:

1. It was more complicated to persist the keyId in the request, so we opted to pull the reference from cracking the access token (which I believe was the wrong approach since we shouldn't be decoding access tokens)
2. We didn't want to trust a keyId in a request that was cached locally, so we decided to find the binding key ID on the access token as it came back from ESTS

Either way, we can't crack the token open anymore in order to support PoP for encrypted tokens.

@sameerag note for a follow-up review: I had to rework the Native Client PoP flow because there was no persisted KeyId in the request. In the SPA PoP flow we'd get the keyId from the cached Access Token credential, which we can't do since the credential lives in WAM on the native flow.

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -320,7 +320,13 @@ export class ResponseHandler {
         if (cacheRecord.accessToken) {
             if (cacheRecord.accessToken.tokenType === AuthenticationScheme.POP) {
                 const popTokenGenerator: PopTokenGenerator = new PopTokenGenerator(cryptoObj);
-                accessToken = await popTokenGenerator.signPopToken(cacheRecord.accessToken.secret, request);
+                const { secret, keyId } = cacheRecord.accessToken;
+
+                if (!keyId) {
+                    throw ClientAuthError.createKeyIdMissingError();
```
jasonnutter, line 326: What would be the way an application would handle this? Retry the request?

hectormmg: Yes, I don't expect this error to be hit but since keyId is an optional param there has to be a check before passing it into the signPopToken API. 

In order for this error to throw, either the keyId would have to be deleted from the cached access token credential somehow or (in the native client case) the keyId would have to be removed from the NativeRequest.

Any suggestions on how to avoid a check on an optional attribute that isn't optional down the line?

jasonnutter: No, I think this should be fine.

PR #4755, opened 2022-04-28
Fix for WAM window parenting

tnorling:
Adds `windowTitleSubstring` to native requests which will help WAM find the window making the request to properly paint the prompt over the correct window


lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -299,6 +299,7 @@ export class NativeInteractionClient extends BaseInteractionClient {
             correlationId: this.correlationId,
             instance_aware: instanceAware,
             token_type: request.authenticationScheme,
+            windowTitleSubstring: document.title,
```
jasonnutter, line 302: This is guaranteed to be unique, that okay?

jasonnutter (approved):

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
tnorling: I assume you meant it's *not* guaranteed to be unique? Yes, we're aware of that scenario and we're ok with the parenting potentially being wrong in this case (we don't have a better solution at the moment). eSTS may include some unique characters in their title in the future to mitigate this if it becomes an issue.

jasonnutter: Yeah, sorry for the typo. Cool, resolving.

peterzenz (approved):

PR #4758, opened 2022-04-29
Update passport-azure-ad to cache-manager@3.6.1

kplindegaard:
When upgrading `async` for `passport-azure-ad`, `cache-manager` should be bumped too to stay up to date with its `async` dependency.

```
passport-azure-ad@4.3.2 microsoft-authentication-library-for-js/maintenance/passport-azure-ad
├── async@3.2.3
├─┬ cache-manager@2.10.2
│ └── async@1.5.2  <-- Vulnerable
....
```

Would be mega grateful to have this bumped before v4.3.2 is pushed to npmjs.


hectormmg:
@kplindegaard thanks for the Pull Request, we will review it and provide feedback if necessary!

jasonnutter:
FYI This version of cache-manager previously broke passport-azure-ad: https://github.com/AzureAD/passport-azure-ad/issues/471

jasonnutter:
Closing in favor of #4781 which addresses breaking changes in `cache-manager@3`, thanks!

PR #4803, opened 2022-05-11
[msal-node] export AuthorizationCodePayload type from msal-common

dhensby:
The `ClientApplication.acquireTokenByCode()` method accepts two arguments, the second is an optional argument of type `AuthorizationCodePayload`.

So that consumers of `msal-node` don't need to require `msal-common` themselves to obtain this type definition, it should be exported by `msal-node`.

### DIFF
--- a/change/@azure-msal-node-2c44a8b8-4645-4ad0-8963-8ba5b369789a.json
+++ b/change/@azure-msal-node-2c44a8b8-4645-4ad0-8963-8ba5b369789a.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "export AuthorizationCodePayload type from msal-common #4803",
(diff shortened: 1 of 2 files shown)


change/@azure-msal-node-2c44a8b8-4645-4ad0-8963-8ba5b369789a.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "export AuthorizationCodePayload type from msal-common",
```
tnorling:
```suggestion
  "comment": "export AuthorizationCodePayload type from msal-common #4803",
```

tnorling (approved):
Thanks for catching this and opening a PR to address!

jasonnutter (approved):

PR #4807, opened 2022-05-12
Fix password encoding on UsernamePassword flow

### LINKED ISSUE #4326: Special characters in passwords using PublicClientApplication.acquireTokenByUsernamePassword() in Node JS
### Core Library

MSAL Node (@azure/msal-node)

### Core Library Version

1.4.0

### Wrapper Library

Not Applicable

### Wrapper Library Version

Nest JS

### Description

I have integrated Azure AD authentication into an existing Node JS application built with the Nest JS framework for a Web application used by college students. As we need to collect additional data about students from Azure AD, I implemented authentication via user name and password. This works well for most users. However, some have special characters in their password. So far we have identified the characters & and + as problematic.
I can assure you that these characters are posted intact as far as the initial call to PublicClientApplication.acquireTokenByUsernamePassword() :

The same characters are accepted by the server's local authentication system using the Passport module.
Where two-factor authentication is required, the frontend Web app does provide a client-side solution via a popup triggered by a remote instance of the MSAL library. This does enable students to enter any character in their passwords, but we'd rather avoid this scenario and have the same user journey for all users, whether using Azure AD authentication or other means.
I have run tests to ensure the original string matches the one sent to acquireTokenByUsernamePassword() and it does. I have tried to escape these characters with backslashes to no avail and do not see any hints in your documentation. Please note this is not a browser issue as the same happens via Postman.

I have upgraded to @azure/msal-node to version 1.4.0. We initially deployed version 1.1.0 last year. 

Any help or pointers, e.g. via configuration options, would be much appreciated.

### Error Message

50126 - [2021-12-14 21:37:35Z]: AADSTS50126: Error validating credentials due to invalid username or password.\r\nTrace ID: 885a428b-91b5-4c94-aacc-082bbfdf2000\r\nCorrelation ID: 0567c418-a6cc-4c0e-9994-747317a5a929\r\nTimestamp: 2021-12-14 21:37:35Z - Correlation ID: 0567c418-a6cc-4c0e-9994-747317a5a929 - Trace ID: 885a428b-91b5-4c94-aacc-082bbfdf2000

### Msal Logs

N/A

### MSAL Configuration

```javascript
const cachePlugin = {
  beforeCacheAccess,
  afterCacheAccess,
};

const msalConfig = {
  auth: {
    clientId: azureAd.auth.clientId,
    authority: azureAd.auth.authority,
  },
  cache: {
    cachePlugin,
  },
};
```

### Relevant Code Snippets

```javascript
const usernamePasswordRequest = {
      scopes: ['user.read'],
      username,
      password,
    };
    await pca
      .acquireTokenByUsernamePassword(usernamePasswordRequest)
      .then(response => {
        data.message = 'acquired token by password grant';
        data.result = response;
      }).catch(error =>{ ....})
```

### Reproduction Steps

Enter email and password. 
Post data to the server.
Decide authentication method based on email type.
If using Azure AD, authenticate email and password via MSAL-node library 
Once authenticated, fetch additional data from Azure AD.

### Expected Behavior

Should work in the same way as college portals where the students can log in with their email and password.

### Identity Provider

Azure AD / MSA

None (Server)

### Regression

_No response_

### Source

External (Customer)

hectormmg:
This PR:
- Fixes #4326 
- Wraps password paramenter inside `encodeURIComponent` before adding to the token request in `UsernamePasswordClient`
- Adds tests to cover changes

### DIFF
--- a/lib/msal-common/src/client/UsernamePasswordClient.ts
+++ b/lib/msal-common/src/client/UsernamePasswordClient.ts
@@ -90,7 +90,7 @@ export class UsernamePasswordClient extends BaseClient {
 
         parameterBuilder.addClientId(this.config.authOptions.clientId);
         parameterBuilder.addUsername(request.username);
-        parameterBuilder.addPassword(encodeURIComponent(request.password));
+        parameterBuilder.addPassword(request.password);
 
         parameterBuilder.addScopes(request.scopes);
 
--- a/lib/msal-common/src/client/UsernamePasswordClient.ts
+++ b/lib/msal-common/src/client/UsernamePasswordClient.ts
@@ -90,7 +90,7 @@ export class UsernamePasswordClient extends BaseClient {
 
         parameterBuilder.addClientId(this.config.authOptions.clientId);
         parameterBuilder.addUsername(request.username);
-        parameterBuilder.addPassword(request.password);
+        parameterBuilder.addPassword(encodeURIComponent(request.password));
 
         parameterBuilder.addScopes(request.scopes);
 
--- a/change/@azure-msal-common-00400af1-6f04-4c51-921f-d96e55fa3fe5.json
+++ b/change/@azure-msal-common-00400af1-6f04-4c51-921f-d96e55fa3fe5.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix password encoding on UsernamePassword flow #4807",
+  "packageName": "@azure/msal-common",
+  "email": "hemoral@microsoft.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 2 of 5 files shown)


lib/msal-common/src/client/UsernamePasswordClient.ts
```diff
@@ -90,7 +90,7 @@ export class UsernamePasswordClient extends BaseClient {
 
         parameterBuilder.addClientId(this.config.authOptions.clientId);
         parameterBuilder.addUsername(request.username);
-        parameterBuilder.addPassword(request.password);
+        parameterBuilder.addPassword(encodeURIComponent(request.password));
```
tnorling: URI encoding is usually done in `RequestParameterBuilder`, can we do it there instead? And please check if username needs it as well

sameerag: +1, this will be needed for the usename.

hectormmg: I added it to username as well, moved the encoding to the RequestParamaterBuilder methods.

tnorling (approved):

sameerag (approved):

PR #4819, opened 2022-05-18
ElectronTestApp e2e tests

derisen:
This PR adds e2e tests (AAD, ADFS) to ElectronTestApp sample using playwright. 

* Since the existing test utils are geared toward puppeteer types, I added another set of testUtils, leading to some seemingly duplicated code.
* I was able to run the tests in CI environment for macos and windows OSs, but currently having difficulty with linux systems via xvfb.


samples/msal-node-samples/ElectronTestApp/test/testUtils.ts
```diff
@@ -0,0 +1,114 @@
+import { ElectronApplication, Page } from "playwright";
```
tnorling: Would be great if the contents of this file could live in the e2eTestUtils folder so we can reuse for our other tests

derisen: of course -moved it into e2eTestUtils

hectormmg (approved):
These look great @derisen ! Thanks!

.github/workflows/msal-node-e2e.yml
```diff
@@ -105,3 +105,95 @@ jobs:
...
+
+    name: ${{ matrix.sample }} - ${{ matrix.os }}
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Use Node.js
+      uses: actions/setup-node@v2
+      with:
+        node-version: 14
```
tnorling: Reason why we only need to test against one version of node? If we need to choose one I'd suggest 16 or 18 as we're going to start enforcing 16+ for local development of our repo soon

.github/workflows/msal-node-e2e.yml
```diff
@@ -105,3 +105,95 @@ jobs:
...
+      run: |
+        npm run install:local
+        npm install
+      env:
+        PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD: 1
+
+    - name: Install Playwright
+      working-directory: samples/msal-node-samples/${{ matrix.sample }}
+      run: |
+        npx playwright install --with-deps chromium
```
tnorling: Can we include this as a postinstall step in the package.json instead?

tnorling (approved):
Approved with comments

.github/workflows/msal-node-e2e.yml
derisen: Electron itself is packaged with a particular version of Node, so I didn't see much value in testing over a range of node versions. Let me switch to v16 in that case.

.github/workflows/msal-node-e2e.yml
derisen: will try 👍

Issue #4795, opened 2022-05-06
PublicClientApplication constructor parameters


Deewens:
### Core Library

MSAL.js v2 (@azure/msal-browser)

### Wrapper Library

MSAL Angular (@azure/msal-angular)

### Documentation Location

https://docs.microsoft.com/en-us/azure/active-directory/develop/tutorial-v2-angular-auth-code#configure-the-application

### Description

I am following the documentation linked above to setup SSO with Azure on my Angular App. The documentation tells us to add the MsalModule.forRoot({...}) import, as you can see on this screenshot:

![image](https://user-images.githubusercontent.com/26287270/167115491-7eec66f6-6caa-46c6-97c0-c53c764c069f.png)

However, there are two other parameters to give to the PublicClientApplication constructor, which should be "null" according to the documentation. However, as you can see on the screenshot below, the TS compiler is not happy and is telling me that "null" value is not possible. I don't know if the issue come from the code or if the documentation is wrong, but I am not sure what to put instead.

![image](https://user-images.githubusercontent.com/26287270/167115649-fafe7c50-97c1-4b13-9ec9-5b94ec2e0a35.png)

Thank you.

tnorling:
@Deewens Apologies for the confusion. That tutorial takes you step by step in setting up an Angular app from the ground up. As such, some of the individual code snippets may be incomplete until you get further into the tutorial. If you scroll down you'll find a section that has you replace those `null` placeholders with the `MsalGuard` and `MsalInterceptor` configs. I'll see what we can do to make these docs more intuitive.

Were you just wanting to understand how to get up and running or is your desire to configure an app without the Guard or Interceptor?

Deewens:
Hi!

Thanks for your reply! When I scrolled down I found the other parts about MsalGuard and MsalInterceptor. It is not really intuitive, however. I think it would be best to add the required parameter (if it is really mandatory) and explain with a comment why it is there and that will be described in a later part.

I just wanted to understand how to get up and running, but I am also curious to know how I can configure without the Guard/Interceptor?

Btw, the error arises only for the `MsalGuard` param and not for the `MsalInterceptor `param.

Thanks for your quick response!

tnorling:
@Deewens Thanks, I agree the tutorial could be more clear in this regard. I've noted your feedback and will see what we can do to improve this for future users.

As for not using the Guard or Interceptor, the configs are required but you don't have to protect any routes or APIs.

[...]

PR #4841, opened 2022-05-29
Make guardConfig and interceptorConfig nullable

EmanH:
The [official docs](https://docs.microsoft.com/en-us/azure/active-directory/develop/tutorial-v2-angular-auth-code) provide an example where these values are null:

image

The type definitions, however, do not allow this currently:

image

### DIFF
--- a/change/@azure-msal-angular-78278c74-496e-4083-8956-27811f8b8082.json
+++ b/change/@azure-msal-angular-78278c74-496e-4083-8956-27811f8b8082.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Make guardConfig and interceptorConfig nullable",
+  "packageName": "@azure/msal-angular",
+  "email": "eman@onlythebible.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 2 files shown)


jo-arroyo:
@EmanH Thanks for your contribution. The official docs you linked was intended to be a step-by-step tutorial for setting up an Angular app with msal-angular. If you see later on in the docs, both the msal-guard and the msal-interceptor are configured (instead of left as `null`) and in fact, this is required for msal-angular to run properly. We are aware that the way this tutorial is set up creates unnecessary confusion, and are prioritizing a fix for this tutorial to be more user-friendly.

Linking a previous issue with this same feedback: #4795
@EmLauber @mmacy for docs visibility.

EmanH:
@jo-arroyo Ok I see. Yes, I think the docs are a bit misleading there. [It looks like a working code snippet](https://docs.microsoft.com/en-us/azure/active-directory/develop/tutorial-v2-angular-auth-code#configure-the-application) while not obvious that its not.

gustavoabel (approved):

jo-arroyo:
Closing as change should be to documentation rather than code.

PR #4844, opened 2022-05-31
allow overriding cache behavior in RT flow to support adal-node to msal-node migration

derisen:
Currently, `acquireTokenByRefreshToken` API will not cache the response from service if there exists no corresponding account entity in the cache already. This diminishes the usefulness of `acquireTokenByRefreshToken` for apps migrating from adal-node to msal-node as it prevents them from making use of their still valid refresh tokens without prompting their users.

This PR proposes a change to refresh token flow caching behavior to accommodate migration scenarios, by allowing devs to override the default behavior via an optional `forceCache` property on `RefreshTokenRequest` (msal-node) and `CommonRefreshTokenRequest` (msal-common) types.

See also #4917 for an alternative solution

This also adds a sample to demonstrate usage.


lib/msal-common/src/request/CommonRefreshTokenRequest.ts
```diff
@@ -21,4 +21,5 @@ export type CommonRefreshTokenRequest = BaseAuthRequest & {
     refreshToken: string;
     tokenQueryParameters?: StringDict;
     ccsCredential?: CcsCredential;
+    forceCache?: boolean;
```
tnorling, line 25: Can we add an explanation of this new parameter above?

derisen: For sure. Added.

tnorling:
@derisen I wonder if another solution might be to expose an API that allows developers to pass in an account to be added to the cache and then ask migration scenarios to call this API first. Would that address the gap?

derisen:
> @derisen I wonder if another solution might be to expose an API that allows developers to pass in an account to be added to the cache and then ask migration scenarios to call this API first. Would that address the gap?

@tnorling ah, good idea! There's actually a `setAccount` method, but it's not public. Let me look into this.

derisen:
> ws developers to pass in an account to be added to the cache and then ask migration scenarios to call this API first. Would that address the gap?

@tnorling I've looked into this, but I'm not sure if we can reliably construct an account object (or rather, generate an account cache key) based on the limited information from adal cache (cache schemas are incompatible, but we can retrieve clientId, tenantId, oid, username, authority and issuer). For instance, what should we assign to `environment` parameter for an adal app that acquired tokens from AAD v1 , so that when handling the RT flow response from AAD v2, the account cache keys would match and the account section in cache gets updated. It looks like there's some guesswork which would open up hard to foresee edge cases. @sameerag thoughts?

tnorling:
> > ws developers to pass in an account to be added to the cache and then ask migration scenarios to call this API first. Would that address the gap?
> 
> @tnorling I've looked into this, but I'm not sure if we can reliably construct an account object (or rather, generate an account cache key) based on the limited information from adal cache (cache schemas are incompatible, but we can retrieve clientId, tenantId, oid, username, authority and issuer). For instance, what should we assign to `environment` parameter for an adal app that acquired tokens from AAD v1 , so that when handling the RT flow response from AAD v2, the account cache keys would match and the account section in cache gets updated. It looks like there's some guesswork which would open up hard to foresee edge cases. @sameerag thoughts?

Environment is just the authority domain and shouldn't be affected by v1 vs v2. I think the fields you listed should be enough, is there anything else you think is missing?

derisen:
> > > ws developers to pass in an account to be added to the cache and then ask migration scenarios to call this API first. Would that address the gap?
> > 
> > 
> > @tnorling I've looked into this, but I'm not sure if we can reliably construct an account object (or rather, generate an account cache key) based on the limited information from adal cache (cache schemas are incompatible, but we can retrieve clientId, tenantId, oid, username, authority and issuer). For instance, what should we assign to `environment` parameter for an adal app that acquired tokens from AAD v1 , so that when handling the RT flow response from AAD v2, the account cache keys would match and the account section in cache gets updated. It looks like there's some guesswork which would open up hard to foresee edge cases. @sameerag thoughts?
> 
> Environment is just the authority domain and shouldn't be affected by v1 vs v2. I think the fields you listed should be enough, is there anything else you think is missing?

@tnorling what I'm failing to see is how can we be sure of what environment will be returned from the service

1. adal app acquires an access token from AAD v1 -authority/issuer is "sts.windows.net"
2. msal app uses the proposed `addAccount` API, passes account information for the user (homeAccountId, environment, idToken etc.)
3. `addAccount` API does some mapping under the hood from v1 "sts.windows.net" to one of the v2 authority aliases ("login.microsoftonline.com", "login.windows.net", "login.microsoft.com"), creates the account entity and the corresponding account cache key `<home_account_id>-<environment>-<realm*>`
4. msal app uses `acquireTokenByRefreshToken`, service responds, a cache key is generated in `responseHandler` based on the service response, and this matches with the previously generated cache key, and the cache gets updated.

I'm just missing how in step 3 we can reliably map "sts.windows.net" to the either of ("login.microsoftonline.com", "login.windows.net", "login.microsoft.com") that the server may respond with. Will the authority in response always be exactly what is set in the configuration object?

tnorling:
> > > > ws developers to pass in an account to be added to the cache and then ask migration scenarios to call this API first. Would that address the gap?
> > > 
> > > 
> > > @tnorling I've looked into this, but I'm not sure if we can reliably construct an account object (or rather, generate an account cache key) based on the limited information from adal cache (cache schemas are incompatible, but we can retrieve clientId, tenantId, oid, username, authority and issuer). For instance, what should we assign to `environment` parameter for an adal app that acquired tokens from AAD v1 , so that when handling the RT flow response from AAD v2, the account cache keys would match and the account section in cache gets updated. It looks like there's some guesswork which would open up hard to foresee edge cases. @sameerag thoughts?
> > 
> > 
> > Environment is just the authority domain and shouldn't be affected by v1 vs v2. I think the fields you listed should be enough, is there anything else you think is missing?
> 
> @tnorling what I'm failing to see is how can we be sure of what environment will be returned from the service
> 
> 1. adal app acquires an access token from AAD v1 -authority/issuer is "sts.windows.net"
> 2. msal app uses the proposed `addAccount` API, passes account information for the user (homeAccountId, environment, idToken etc.)
> 3. `addAccount` API does some mapping under the hood from v1 "sts.windows.net" to one of the v2 authority aliases ("login.microsoftonline.com", "login.windows.net", "login.microsoft.com"), creates the account entity and the corresponding account cache key `<home_account_id>-<environment>-<realm*>`
> 4. msal app uses `acquireTokenByRefreshToken`, service responds, a cache key is generated in `responseHandler` based on the service response, and this matches with the previously generated cache key, and the cache gets updated.
> 
> I'm just missing how in step 3 we can reliably map "sts.windows.net" to the either of ("login.microsoftonline.com", "login.windows.net", "login.microsoft.com") that the server may respond with. Will the authority in response always be exactly what is set in the configuration object?

This happens during authority discovery. When you call `acquireTokenByRefreshToken` it will first make a network request to the instance discovery endpoint to get the aliases for the provided authority. It will then do a cache lookup for accounts/tokens and any entries that match one of the aliases are considered valid.

derisen:
> > > > > ws developers to pass in an account to be added to the cache and then ask migration scenarios to call this API first. Would that address the gap?
> > > > 
> > > > 
> > > > @tnorling I've looked into this, but I'm not sure if we can reliably construct an account object (or rather, generate an account cache key) based on the limited information from adal cache (cache schemas are incompatible, but we can retrieve clientId, tenantId, oid, username, authority and issuer). For instance, what should we assign to `environment` parameter for an adal app that acquired tokens from AAD v1 , so that when handling the RT flow response from AAD v2, the account cache keys would match and the account section in cache gets updated. It looks like there's some guesswork which would open up hard to foresee edge cases. @sameerag thoughts?
> > > 
> > > 
> > > Environment is just the authority domain and shouldn't be affected by v1 vs v2. I think the fields you listed should be enough, is there anything else you think is missing?
> > 
> > 
> > @tnorling what I'm failing to see is how can we be sure of what environment will be returned from the service
> > 
> > 1. adal app acquires an access token from AAD v1 -authority/issuer is "sts.windows.net"
> > 2. msal app uses the proposed `addAccount` API, passes account information for the user (homeAccountId, environment, idToken etc.)
> > 3. `addAccount` API does some mapping under the hood from v1 "sts.windows.net" to one of the v2 authority aliases ("login.microsoftonline.com", "login.windows.net", "login.microsoft.com"), creates the account entity and the corresponding account cache key `<home_account_id>-<environment>-<realm*>`
> > 4. msal app uses `acquireTokenByRefreshToken`, service responds, a cache key is generated in `responseHandler` based on the service response, and this matches with the previously generated cache key, and the cache gets updated.
> > 
> > I'm just missing how in step 3 we can reliably map "sts.windows.net" to the either of ("login.microsoftonline.com", "login.windows.net", "login.microsoft.com") that the server may respond with. Will the authority in response always be exactly what is set in the configuration object?
> 
> This happens during authority discovery. When you call `acquireTokenByRefreshToken` it will first make a network request to the instance discovery endpoint to get the aliases for the provided authority. It will then do a cache lookup for accounts/tokens and any entries that match one of the aliases are considered valid.

So basically any one of the aliases will be considered a cache key match during response handling? -but this doesn't seem to be the case currently i.e. I was only able to get a match if I pass exactly "login.windows.net" to `addAccount` API as environment (which is also not what's set for authority on configuration object). The cache won't update for any other aliases :confused:

derisen:
> > > > > > ws developers to pass in an account to be added to the cache and then ask migration scenarios to call this API first. Would that address the gap?
> > > > > 
> > > > > 
> > > > > @tnorling I've looked into this, but I'm not sure if we can reliably construct an account object (or rather, generate an account cache key) based on the limited information from adal cache (cache schemas are incompatible, but we can retrieve clientId, tenantId, oid, username, authority and issuer). For instance, what should we assign to `environment` parameter for an adal app that acquired tokens from AAD v1 , so that when handling the RT flow response from AAD v2, the account cache keys would match and the account section in cache gets updated. It looks like there's some guesswork which would open up hard to foresee edge cases. @sameerag thoughts?
> > > > 
> > > > 
> > > > Environment is just the authority domain and shouldn't be affected by v1 vs v2. I think the fields you listed should be enough, is there anything else you think is missing?
> > > 
> > > 
> > > @tnorling what I'm failing to see is how can we be sure of what environment will be returned from the service
> > > 
> > > 1. adal app acquires an access token from AAD v1 -authority/issuer is "sts.windows.net"
> > > 2. msal app uses the proposed `addAccount` API, passes account information for the user (homeAccountId, environment, idToken etc.)
> > > 3. `addAccount` API does some mapping under the hood from v1 "sts.windows.net" to one of the v2 authority aliases ("login.microsoftonline.com", "login.windows.net", "login.microsoft.com"), creates the account entity and the corresponding account cache key `<home_account_id>-<environment>-<realm*>`
> > > 4. msal app uses `acquireTokenByRefreshToken`, service responds, a cache key is generated in `responseHandler` based on the service response, and this matches with the previously generated cache key, and the cache gets updated.
> > > 
> > > I'm just missing how in step 3 we can reliably map "sts.windows.net" to the either of ("login.microsoftonline.com", "login.windows.net", "login.microsoft.com") that the server may respond with. Will the authority in response always be exactly what is set in the configuration object?
> > 
> > 
> > This happens during authority discovery. When you call `acquireTokenByRefreshToken` it will first make a network request to the instance discovery endpoint to get the aliases for the provided authority. It will then do a cache lookup for accounts/tokens and any entries that match one of the aliases are considered valid.
> 
> So basically any one of the aliases will be considered a cache key match during response handling? -but this doesn't seem to be the case currently i.e. I was only able to get a match if I pass exactly "login.windows.net" to `addAccount` API as environment (which is also not what's set for authority on configuration object). The cache won't update for any other aliases 😕

thanks @tnorling, this approach would work too. I still favour the optional flag approach for the sake of simplicity, but I'll leave the final decision to you @sameerag @tnorling. Will open another PR for adding an `addAccount` API and leave this one as it is for comparison.

samples/msal-node-samples/refresh-token/adalApp.js
```diff
@@ -0,0 +1,99 @@
+// Import dependencies
+var express = require('express');
```
Robbie-Microsoft, line 2: should you be using "const" instead of "var"?

samples/msal-node-samples/refresh-token/adalApp.js
```diff
@@ -0,0 +1,99 @@
...
+    }
+}));
+
+app.get('/', function (req, res) {
+
+    // Create a random string to use against XSRF
+    crypto.randomBytes(48, function (ex, buf) {
+        req.session.state = buf.toString('base64')
+            .replace(/\//g, '_')
+            .replace(/\+/g, '-');
```
Robbie-Microsoft, line 56: should you be defining these with "new Regexp()"?

samples/msal-node-samples/refresh-token/adalApp.js
Robbie-Microsoft, line 2: nitpick: Do we have a style guide? I'm wondering why you used single quotes instead of double quotes.

samples/msal-node-samples/refresh-token/adalApp.js
```diff
@@ -0,0 +1,99 @@
...
+        redirectUri,
+        resource,
+        clientId,
+        clientSecret,
+        function (err, response) {
+            if (err) return next(err);
+
+            // cache the response
+            authenticationContext.cache.add([response], (err, result) => {
+                if (err) return next(err);
```
Robbie-Microsoft: you're using "function (err, response) {" here, but then also using an arrow function "() => {". We should consistently use one or the other.

samples/msal-node-samples/refresh-token/adalCustomCache.js
```diff
@@ -0,0 +1,114 @@
...
+            }
+        });
+    });
+
+    // Add the new entries to the end of the cache.
+    entries = _.compact(entries);
+
+    for (var i = 0; i < entries.length; i++) {
+        cachedEntries.push(entries[i]);
+    }
```
Robbie-Microsoft: use a .forEach here

samples/msal-node-samples/refresh-token/adalCustomCache.js
```diff
@@ -0,0 +1,114 @@
...
+        }
+    });
+}
+
+function entriesHaveEqualHashKeys(a, b) {
+    return _.isEqual(extractCacheKeyFromEntry(a), extractCacheKeyFromEntry(b));
+}
+
+function extractCacheKeyFromEntry(entry) {
+    if (entry) {
```
Robbie-Microsoft: should you be returning something if there is no entry?

samples/msal-node-samples/refresh-token/adalApp.js
derisen: Node.js seems to favour single quotes in their [own documentation](https://nodejs.org/en/docs/guides/getting-started-guide/), although I don't think they have a public style guide

samples/msal-node-samples/refresh-token/adalApp.js
derisen: ah, good catch! will revise

samples/msal-node-samples/refresh-token/adalApp.js
derisen: Just a stylistic choice -ADAL apps tend to be pre-ES6

samples/msal-node-samples/refresh-token/adalCustomCache.js
derisen: that would be the better choice semantically, but `forEach` is also ES6 -so just a stylistic choice (as well as this code was adopted from ADAL-Node [in-memory token cache](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/maintenance/adal-node/lib/memory-cache.js), but I'm happy to reconsider!)

samples/msal-node-samples/refresh-token/adalApp.js
derisen: I think both constructor and literal approaches are valid. Stylistically I've prefer to leave the older literal approach.

samples/msal-node-samples/refresh-token/adalCustomCache.js
derisen: Of course, good point. Revising

sameerag (approved):

samples/msal-node-samples/refresh-token/adalApp.js
```diff
@@ -0,0 +1,99 @@
...
+var config = require('./config/customConfig.json');
+
+var DiskCache = require('./adalCustomCache');
+var diskCache = new DiskCache(config.adalCacheLocation);
+
+// Authentication parameters
+var clientId = config.clientId;
+var clientSecret = config.clientSecret;
+var tenant = config.tenantInfo;
+var authorityUrl = config.authority + '/' + tenant;
```
Robbie-Microsoft, line 16: Should you use a template literal here (and everywhere else in this review where applicable)?

var authorityUrl = `${config.authority}/${tenant}`;

I personally think it makes the code look much nicer. Up to you though.

Edit: Just noticed you are using template literals in other places in this review. Why not here too?

samples/msal-node-samples/refresh-token/adalCustomCache.js
```diff
@@ -0,0 +1,114 @@
...
+
+/**
+ * Removes a collection of entries from the cache in a single batch operation.
+ * @param {Array} entries An array of cache entries to remove.
+ * @param {Function} callback This function is called when the operation is complete. Any error is provided as the first parameter.
+ */
+DiskCache.prototype.remove = function (entries, callback) {
+    const cachedEntries = _readCache.call(this);
+
+    var updatedEntries = _.filter(cachedEntries, function (element) {
```
Robbie-Microsoft: I had previously commented let/const vs var, and you said you wanted to use var every where. Why aren't you using var on line 20?

samples/msal-node-samples/refresh-token/adalCustomCache.js
```diff
@@ -0,0 +1,114 @@
...
+    _.each(cachedEntries, function (element) {
+        _.each(entries, function (addElement, index) {
+            if (entriesHaveEqualHashKeys(element, addElement)) {
+                entries[index] = null;
+            }
+        });
+    });
+
+    // Add the new entries to the end of the cache.
+    entries = _.compact(entries);
```
Robbie-Microsoft: It's a bad idea to reassign function parameters. I think you should you create a new variable here.

https://medium.com/the-andela-way/its-bad-practice-to-reassign-function-params-c37f289a5dc4

var compactedEntries = _.compact(entries);

samples/msal-node-samples/refresh-token/adalCustomCache.js
derisen: ah, my bad, this one slipped through.. updated!

samples/msal-node-samples/refresh-token/adalApp.js
derisen: I love template literals too! It's just in the ADAL sample I'm trying to stick to pre-ES6

samples/msal-node-samples/refresh-token/adalCustomCache.js
derisen: That's good catch -updated!

samples/msal-node-samples/refresh-token/adalApp.js
Robbie-Microsoft: I forgot that we discussed that already. Sorry!

Robbie-Microsoft (approved):

derisen:
Much appreciated @sameerag @Robbie-Microsoft -merging this.

PR #4846, opened 2022-05-31
Fix ClientAssertion configuration typing between common and node

hectormmg:
This PR:

- Updates the clientAssertion type in BaseAuthRequest to support either a string or a ClientAssertionCredential object.
- Adds the `ClientCredentials` module with relevant types in `msal-common`
- Adds step to build a `ClientAssertionCredential` object from an assertion string in the request building process
- Exports the `ClientAssertionCredential` type from `msal-common`


lib/msal-node/src/config/Configuration.ts
```diff
@@ -30,7 +31,7 @@ export type NodeAuthOptions = {
     clientId: string;
     authority?: string;
     clientSecret?: string;
-    clientAssertion?:string;
+    clientAssertion?: ClientAssertionCredential;
```
tnorling: This is a breaking change

tnorling: You could do `string | ClientAssertionCredential` but then you'll need to detect what was provided and do the right thing

hectormmg: Do you think that's the right way to go? If so I'll make that change, I'm not sure if there's any other alternative really. Although it's a breaking change, I don't believe anyone is using this (the client assertion flow simply shouldn't work as the code is right now).

jasonnutter: If the existing functionality does not work with `clientAssertion` as a string, then this would technically be a bug fix and doesn't need to be a breaking change or made backwards compatible.

tnorling: +1 Agreed

peterzenz: How do we know if the existing functionality isn't working and we aren't going to break someone?

peterzenz (approved):
Would like some clarity on if this breaking or not.

lib/msal-node/src/config/Configuration.ts
jasonnutter: A unit test should be able to answer that, in theory.

hectormmg: It seems a bit more complicated than I originally realized (and there are more bugs). There's a few different types of client assertions that are built on the node side which the common client class doesn't support. None of the currently implemented assertion configs in Node should work at all. 

I'll spend the day fixing this and I really hope I can get it merged by Monday.

lib/msal-common/src/request/BaseAuthRequest.ts
```diff
@@ -21,7 +22,7 @@ import { AzureCloudOptions, ClientAssertion } from "../config/ClientConfiguratio
  * - sshKid                  - Key ID that uniquely identifies the SSH public key mentioned above.
  * - azureCloudOptions       - Convenience string enums for users to provide public/sovereign cloud ids
  * - requestedClaimsHash     - SHA 256 hash string of the requested claims string, used as part of an access token cache key so tokens can be filtered by requested claims
- * - clientAssertion         - Client assertion passed by the user for confidential client flows
+ * - clientAssertion         - Client assertion strung passed by the user for confidential client flows which is then processed into an object when request is built
```
hectormmg:
```suggestion
 * - clientAssertion         - Client assertion string passed by the user for confidential client flows which is then processed into an object when request is built
```

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -201,8 +201,8 @@ export class RefreshTokenClient extends BaseClient {
 
         // Use clientAssertion from request, fallback to client assertion in base configuration
         const clientAssertion = request.clientAssertion || this.config.clientCredentials.clientAssertion;
-
-        if (clientAssertion) {
+        
+        if (clientAssertion && typeof clientAssertion === "object") {
```
tnorling: What happens if it's not type object?

hectormmg: I'm thinking of throwing an error, but in theory the check is just for type safety given that the string assertion is converted to the ClientAssertionCredential type  before it reaches this point. I'm unsure if throwing here would be better than simply not adding the assertion and letting the invalid credential error from the server happen on it's own.

I guess if we throw here we do save ourselves a network call.

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -226,10 +226,8 @@ export class AuthorizationCodeClient extends BaseClient {
             parameterBuilder.addClientSecret(this.config.clientCredentials.clientSecret);
         }
 
-        // Use clientAssertion from request, fallback to client assertion in base configuration
-        const clientAssertion = request.clientAssertion || this.config.clientCredentials.clientAssertion;
-
-        if (clientAssertion) {
+        if (this.config.clientCredentials.clientAssertion) {
```
hectormmg, line 229: This has been rolled back to the behavior before adding client assertion to the request since the auth code, OBO, refresh token and UPN flows won't use a client assertion. We should consider removing the client assertion logic from non-client credential clients in the future.

sameerag (approved):
Please remove the unnecessary "omits". Approving.

hectormmg:
> Please remove the unnecessary "omits". Approving.

Done, thanks!

PR #4880, opened 2022-06-09
Return correct fromCache value when tokens are acquired from WAM

tnorling:
WAM exposes whether or not a token was returned from the cache via MATS telemetry. This PR parses MATS to return the correct value in the `fromCache` field of the response.


lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -255,6 +255,7 @@ export class NativeInteractionClient extends BaseInteractionClient {
      * @param response
      */
     private validateNativeResponse(response: object): NativeResponse {
+        console.log(response);
```
jasonnutter:
```suggestion
```

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -269,6 +270,40 @@ export class NativeInteractionClient extends BaseInteractionClient {
         }
     }
 
+    /**
+     * Gets MATS telemetry from native response
+     * @param response 
+     * @returns 
+     */
+    private getMATSFromResponse(response: NativeResponse): object|null {
```
jasonnutter: This object appears to have known properties, can we type this instead?

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -269,6 +270,40 @@ export class NativeInteractionClient extends BaseInteractionClient {
...
+        let mats = null;
+        if (response.properties.hasOwnProperty(NativeConstants.MATS_TELEMETRY)) {
+            try {
+                mats = JSON.parse(response.properties[NativeConstants.MATS_TELEMETRY]);
```
jasonnutter: Would be fine with immediately returning so that you don't need to instantiate `mats` outside of the try.

tnorling: Done

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
tnorling: Done

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -269,6 +269,39 @@ export class NativeInteractionClient extends BaseInteractionClient {
...
+        return null;
+    }
+
+    /**
+     * Returns whether or not response came from native cache
+     * @param response 
+     * @returns 
+     */
+    private isResponseFromCache(response: NativeResponse): boolean {
+        const mats = this.getMATSFromResponse(response);
```
peterzenz: There are a ton of possible metrics in the MATS blob, are there any others we want to include?  
Also suggest refactoring this out of the isResponseFromCache method so this doesn't accidentally get parsed multiple times if we pull additional values from the MATS result.

lib/msal-browser/src/broker/nativeBroker/NativeResponse.ts
```diff
@@ -21,9 +21,23 @@ export type NativeResponse = {
...
     state: string;
     shr?: string;
     extendedLifetimeToken?: boolean;
 };
+
+/**
+ * Properties returned under "properties" of the NativeResponse
+ */
+export type NativeResponseProperties = {
+    MATS?: string;
```
peterzenz, line 35: More definition for what MATS is.  Maybe something like:
"The native token broker on Windows can optionally include additional performance metrics about operations it performs.  If that data is returned, we'll extract certain metrics to include in the library telemetry."

peterzenz (approved):

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
tnorling: Done. I've added additional interesting metrics to the MATS type. In the interest of time, I will open a follow up PR to consume those extra metrics and send to aria

lib/msal-browser/src/broker/nativeBroker/NativeResponse.ts
tnorling: Thanks! Done.

hectormmg (approved):

PR #4892, opened 2022-06-13
setInteractionInProgress should not throw for different client ids

jasonnutter:
For pairwise broker, there are scenarios were the broker and embedded application are in the same frame, and then embedded application can invoke a popup. Today, this throws because the embedded application will call setInteractionInProgress with its client id, then send a message to the broker, which attempts to do the same thing but with its client id, which throws. This call no longer throws, allowing the popup request to go through.

### DIFF
--- a/change/@azure-msal-browser-66879877-e54b-49f6-a42d-445e712e2624.json
+++ b/change/@azure-msal-browser-66879877-e54b-49f6-a42d-445e712e2624.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "setInteractionInProgress should not throw for different client ids",
+  "packageName": "@azure/msal-browser",
+  "email": "janutter@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/lib/msal-browser/src/cache/BrowserCacheManager.ts
+++ b/lib/msal-browser/src/cache/BrowserCacheManager.ts
@@ -962,7 +962,8 @@ export class BrowserCacheManager extends CacheManager {
         // Ensure we don't overwrite interaction in progress for a different clientId
         const key = `${Constants.CACHE_PREFIX}.${TemporaryCacheKeys.INTERACTION_STATUS_KEY}`;
         if (inProgress) {
-            if (this.getInteractionInProgress()) {
+            // For pairwise broker, ensure same frame embedded app can invoke interaction, as both instances will set this flag
+            if (this.isInteractionInProgress(true)) {
                 throw BrowserAuthError.createInteractionInProgressError();
             } else {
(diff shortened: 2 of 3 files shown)


change/@azure-msal-browser-66879877-e54b-49f6-a42d-445e712e2624.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "setInteractionInProgress should not throw for different client ids",
```
jasonnutter, line 3:
```suggestion
  "comment": "setInteractionInProgress should not throw for different client ids #4892",
```

hectormmg (approved):

lib/msal-browser/src/cache/BrowserCacheManager.ts
```diff
@@ -962,7 +962,8 @@ export class BrowserCacheManager extends CacheManager {
         // Ensure we don't overwrite interaction in progress for a different clientId
         const key = `${Constants.CACHE_PREFIX}.${TemporaryCacheKeys.INTERACTION_STATUS_KEY}`;
         if (inProgress) {
-            if (this.getInteractionInProgress()) {
+            // For pairwise broker, ensure same frame embedded app can invoke interaction, as both instances will set this flag
+            if (this.isInteractionInProgress(true)) {
```
tnorling, line 966: We may need to think about this some more. This will break scenarios where an app has multiple clientIds that may be invoking interaction, such as Sharepoint's scenarios.

tnorling: Maybe have the embedded frame either not call `setInteractionInProgress` or have it pass in a flag that indicates we shouldn't set it?

jasonnutter: > We may need to think about this some more. This will break scenarios where an app has multiple clientIds that may be invoking interaction, such as Sharepoint's scenarios.

Break in what way?

jasonnutter: > Maybe have the embedded frame either not call setInteractionInProgress

This could work, but there be side effects as well.

peterzenz: Do we understand what could break here?  Or what the side effects of the alternative approach would be?

jasonnutter: I think the approach of not setting interaction in progress when the application is embedded is better, shouldn't be an issue. PR: https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/4893/files

PR #4893, opened 2022-06-13
preflightBrowserEnvironmentCheck should not always set interaction in progress

jasonnutter:
For pairwise broker scenarios, we do not need the embedded application to set interaction in progress, as the broker will be performing interaction, and will set this value itself. This fixes a bug when the embedded application and the broker are in the same frame, and they both set interaction in progress for the same request.

Replaces #4892


lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -738,7 +738,7 @@ export abstract class ClientApplication {
      * Helper to validate app environment before making an auth request
      * * @param interactionType
      */
-    protected preflightBrowserEnvironmentCheck(interactionType: InteractionType): void {
+    protected preflightBrowserEnvironmentCheck(interactionType: InteractionType, setInteractionInProgress: boolean = true): void {
```
jasonnutter, line 756: This is called from the 1p library for the broker scenarios, so there will be a separate PR in the 1p repo that will consume this change.

tnorling, line 756: Can we add a description to the typedocs above?

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -772,13 +772,15 @@ export abstract class ClientApplication {
      * @param request
      * @param interactionType
      */
-    protected preflightInteractiveRequest(): void {
+    protected preflightInteractiveRequest(setInteractionInProgress: boolean): void {
```
tnorling, line 791: Here too

lib/msal-browser/src/app/ClientApplication.ts
jasonnutter: Done.

lib/msal-browser/src/app/ClientApplication.ts
jasonnutter: Done

tnorling (approved):

sameerag (approved):

PR #4906, opened 2022-06-14
Add node-token-validation cache manager

jo-arroyo:
This PR adds caching to node-token-validation, with unit tests.

Additional signature validation key entities/types and cache methods for those types are pending design and will be in a future PR.


peterzenz (approved):

lib/node-token-validation/src/cache/CacheTypes.ts
```diff
@@ -0,0 +1,12 @@
...
+ * Licensed under the MIT License.
+ */
+
+import { ValidCacheType } from "@azure/msal-common";
+
+/**
+ * Key value store for in-memory cache
+ * @public
+ */
+export type CacheKVStore = Record<string, ValidCacheType>;
```
tnorling: There's been some discussion about moving to `Map` instead of `Record` due to an attack vector with `Record`. With that in mind, I'd suggest using `Map` here instead to get ahead of that.

lib/node-token-validation/src/cache/NodeCacheManager.ts
```diff
@@ -0,0 +1,243 @@
...
+     */
+    emitChange(): void {
+        this.changeEmitters.forEach(func => func.call(null));
+    }
+
+    /**
+     * Gets the current cache key-value store
+     * @returns {CacheKVStore} Key Value store for in-memory storage
+     */
+    getCache(): CacheKVStore {
```
tnorling, line 25: Some of this is also implemented in `msal-node` did you consider moving this to `msal-common` so you can share code?

lib/node-token-validation/src/cache/CacheTypes.ts
jo-arroyo: Good call. Updated.

lib/node-token-validation/src/cache/NodeCacheManager.ts
jo-arroyo: After talking with @sameerag, it seems that having this code in `msal-node` as opposed to `msal-common` was a deliberate choice, so that each platform retains control over storage processes in case it changes. It does mean some of the code would be duplicated in `node-token-validation` since it's also a node library, but the priority is control over code optimization.

tnorling: That may have been the case when it was just browser and node but we may want to revisit that decision with the introduction of another node based library. Are there any known existing or future scenarios where the cache implementation may differ that could justify writing it twice? Potentially a question for design review?

lib/node-token-validation/src/cache/CacheTypes.ts
```diff
@@ -0,0 +1,12 @@
...
+ * Licensed under the MIT License.
+ */
+
+import { ValidCacheType } from "@azure/msal-common";
+
+/**
+ * Key value store for in-memory cache
+ * @public
+ */
+export type CacheKVStore = Map<string, ValidCacheType>;
```
sameerag, line 12: +1. Prefer maps wherever we can

lib/node-token-validation/src/cache/NodeCacheManager.ts
```diff
@@ -0,0 +1,243 @@
...
+    /**
+     * Sets the current cache key-value store
+     * @param {CacheKVStore} cache 
+     */
+    setCache(cache: CacheKVStore): void {
+        this.logger.trace("Setting cache");
+        this.cache = cache;
+
+        // Mark change in cache
+        this.emitChange();
```
sameerag: Why would you need this? Ideally this was set as an event marker to denote change in the cache to enable disk read/write with data integrity for `msal-node-extensions` where we needed to read from an SSO island (where multiple apps have access to the token store - MSAL .Net, Py, Java etc..).

Would we need similar storage choices for `node-token-validation`?

Happy to talk if you need more context.

jo-arroyo: Thanks for the chat today. We concluded that this event marker was not needed for `node-token-validation`, and it has been removed.

sameerag (approved):

PR #4912, opened 2022-06-15
Token validation authority update

jo-arroyo:
This PR:
- Updates `OidcConfigProvider` to use `Authority` from msal-common
- Add `knownAuthorities` to Configuration

Note: `NodeCacheManager` is still needed for the `Authority` class usage, but will not be used for other caching purposes (i.e. signing key caching) at this time.


peterzenz (approved):

lib/node-token-validation/src/config/Configuration.ts
```diff
@@ -14,6 +14,10 @@ export type Configuration = {
 
 export type TokenValidationOptions = {
     authority?: string,
+    /**
+     * An array of URIs that are known to be valid. Used in B2C scenarios.
+     */
+    knownAuthorities?: Array<string>
     protocolMode?: ProtocolMode
```
Robbie-Microsoft: should there be commas at the end of these 2 lines?

lib/node-token-validation/test/config/OpenIdConfigProvider.spec.ts
```diff
@@ -20,52 +24,63 @@ describe("OpenIdConfigProvider", () => {
 
     describe("constructor", () => {
         it("exports a class", () => {
-            const provider = new OpenIdConfigProvider(config, config.system.networkClient, new Logger(config.system.loggerOptions));
+            const logger = new Logger(config.system.loggerOptions);
```
Robbie-Microsoft: could this be defined at the top of the file instead of defining it in each "it"?

Robbie-Microsoft (approved):

lib/node-token-validation/src/config/Configuration.ts
jo-arroyo: Yes! Fixed. Thank you

lib/node-token-validation/test/config/OpenIdConfigProvider.spec.ts
jo-arroyo: Yes, thanks for the catch. Fixed.

PR #4917, opened 2022-06-17
createAccount API to support adal-node to msal-node migration

derisen:
Currently, `acquireTokenByRefreshToken` API will not cache the response from service if there exists no corresponding account entity in the cache already. This diminishes the usefulness of `acquireTokenByRefreshToken` for apps migrating from adal-node to msal-node as it prevents them from making use of their still valid refresh tokens without prompting their users.

This PR proposes a `createAccount` API to msal-node `TokenCache` class, which takes in a raw ID token string and optionally a base64 encoded `clientInfo` object to create an `AccountEntity` and the relevant cache record in the cache. This in turn enables `acquireTokenByRefreshToken` calls to successfully cache RT flow response in migration scenarios.

See also #4844 for an alternative solution.

This also adds a sample to demonstrate usage.

cc @tnorling


ghost:

derisen:
We have discussed this offline and decided that it introduces unnecessary complexity. Closing.

PR #4936, opened 2022-06-24
Add extensiblity point AppTokenCacheProvider

bgavrilMS:
# Spec

Extensibility point that allows Azure SDK to inject tokens obtained through Managed Identity (MI) into MSAL. 

Full spec: https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=/Managed%20Identity%20Support/ManagedIdentity.md&version=GBbogavril/msi&_a=preview

### Samples and docs
This functionality is specifically for Azure SDK to be able to offer a good experience with Managed Identity tokens. There is no need for extra documentation, samples etc.

### Tech details
Note that I did not commit package.lock.json because they were wildly different. Not sure why. Let me know if this a problem.

# Usage 

```javascript
    const clientConfig = {
        auth: {
           // client_id etc.
           // no secret or certificate needed
        },         
        extensibility: {
            appTokenCacheProvider(appTokenProviderParameters) {
                return {
                    accessToken: "foo",
                    expiresInSeconds: 3600,
                    refreshInSeconds: 1800,
                }
            }
        }
         
    };

   
   const confidentialClientApplication = new msal.ConfidentialClientApplication(clientConfig);


 const clientCredentialRequest = {
        scopes: "https://www.graph.microsoft.com/.default",
        skipCache: false,
    };

    return confidentialClientApplication 
        .acquireTokenByClientCredential(clientCredentialRequest)
        .then((response) => {
            // Uncomment to see the successful response logged
             console.log("Response: ", response);
        }).catch((error) => {
            // Uncomment to see the errors logges
             console.log(JSON.stringify(error));
        });
```


bgavrilMS:
Gentle reminder for reviewing this PR. 
It looks like the `auth-code` UI tests are failing because of unrelated reasons. 
However the electron app is failing due to a tsdx failure, which I cannot track down. Please advise. https://github.com/AzureAD/microsoft-authentication-library-for-js/runs/7090361360?check_suite_focus=true

lib/msal-common/src/config/ClientConfiguration.ts
```diff
@@ -61,7 +62,59 @@ export type CommonClientConfiguration = {
...
+    appTokenProvider: IAppTokenProvider | null,
+};
+
+/**
+ * Extensibility interface, which allows the app developer to return a token, based on the passed-in parameters, instead of fetching tokens from
+ * the Identity Provider (AAD).
+ * Developers need to construct and return an AppTokenProviderResult object back to MSAL. MSAL will cache the token response
+ * in the same way it would do if the result were comming from AAD.
+ * This extensibility point is only defined for the client_credential flow, i.e. acquireTokenByClientCredential and
+ * meant for Azure SDK to enhance Managed Identity support.
```
peterzenz: Do we want stronger wording here indicating that unless you're Azure SDK, you shouldn't be looking at this?

lib/msal-common/test/client/ClientCredentialClient.spec.ts
```diff
@@ -359,5 +360,61 @@ describe("ClientCredentialClient unit tests", () => {
...
+        const authResult = await client.acquireToken(clientCredentialRequest) as AuthenticationResult;
+
+        expect(callbackedCalledCount).toEqual(1);
 
+        expect(authResult.scopes).toEqual(expectedScopes);
+        expect(authResult.accessToken).toEqual(accessToken);
+        expect(authResult.state).toHaveLength(0);
+        const dateDiff = (authResult.expiresOn!.valueOf() - Date.now().valueOf()) / 1000;
+        expect(dateDiff).toBeLessThanOrEqual(1900);
+        expect(dateDiff).toBeGreaterThan(1700);
```
peterzenz, line 412: So the test fails if it takes more than 200ms to run?

peterzenz (approved):

lib/msal-common/src/config/ClientConfiguration.ts
bgavrilMS: Feel free to suggest stronger wording.

lib/msal-common/test/client/ClientCredentialClient.spec.ts
bgavrilMS: Those numbers are in seconds, not ms. I wanted a way to test that the expiration is the one set by the test at line 371, i.e. 1800s. I can't use exactly 1800 in the assert because some time will have elapsed. So, I put a generous interval in there.

change/@azure-msal-common-cae9fa3d-96e5-4b30-859c-179e06b9a6be.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add extensiblity point AppTokenCacheProvider",
```
tnorling:
```suggestion
  "comment": "Add extensiblity point AppTokenCacheProvider #4936",
```

change/@azure-msal-node-78f888fd-8751-47ff-8ca7-09ba01e55c95.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add extensiblity point AppTokenCacheProvider",
```
tnorling:
```suggestion
  "comment": "Add extensiblity point AppTokenCacheProvider #4936",
```

contributing.md
```diff
@@ -64,6 +64,11 @@ npm run build
 npm run build -- --scope <package-name>
 ```
 
+### Build all libraries
+
+Link the libraries toghether: from the root of the repo `npx lerna bootstrap`
```
tnorling: This should be handled automatically by a postinstall script when you install from the root. We're also planning to ditch lerna sometime in the near future so we should stick to just `npm install`

lib/msal-common/src/config/ClientConfiguration.ts
```diff
@@ -61,7 +62,59 @@ export type CommonClientConfiguration = {
...
+ */
+export interface IAppTokenProvider {
```
tnorling: Can we put the interface along with its dependent types in its own file?

lib/msal-common/src/config/ClientConfiguration.ts
```diff
@@ -46,7 +46,8 @@ export type ClientConfiguration = {
     telemetry?: TelemetryOptions,
     serverTelemetryManager?: ServerTelemetryManager | null,
     persistencePlugin?: ICachePlugin | null,
-    serializableCache?: ISerializableTokenCache | null
+    serializableCache?: ISerializableTokenCache | null,    
+    appTokenProvider?: IAppTokenProvider | null
```
tnorling: Since this is only relevant to the `ClientCredentialClient` I'd propose including this as a parameter on the `ClientCredentialClient` constructor instead of polluting all of our client classes

lib/msal-node/src/config/Configuration.ts
```diff
@@ -84,6 +94,7 @@ export type Configuration = {
     cache?: CacheOptions;
     system?: NodeSystemOptions;
     telemetry?: NodeTelemetryOptions;
+    extensibility?: NodeExtensibilityOptions;
```
tnorling: Same comment here. I'd like to see if there's a way we can expose this only for the Classes for which this feature is supported.

lib/msal-common/src/config/ClientConfiguration.ts
sameerag: We do not provide "ClientCredentialClient" specific config, we only control requests that way which I guess why Bogdan chose config.

tnorling: I'm suggesting to add it as an optional parameter on [this line](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/867516c11c83639890abe2fcd6fc6e73ed659644/lib/msal-common/src/client/ClientCredentialClient.ts#L29)

lib/msal-common/src/config/ClientConfiguration.ts
peterzenz: I'll leave it to you. If customer questions come in, I'll route them your way.

contributing.md
bgavrilMS: Ok, reverted

lib/msal-common/src/config/ClientConfiguration.ts
bgavrilMS: @tnorling - I am not sure I understand how the app developer can use object? I was working based on this sample: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/42287f399ea292b4b25eb689eb4b304b30646d62/samples/msal-node-samples/client-credentials/index.js#L77

It is important to note that this extensibility point should be set at the application level, because it is used instead of "secret" or "certificate". It is not meant to be defined at the request level.

We could generalize the extensibility point so that's it is a general-purpose "token provider" instead of an "app token provider" (i.e. via renaming). But making it work with user tokens is quite a lot of work and outside the scope of this improvement.

lib/msal-node/src/config/Configuration.ts
bgavrilMS: This is not a request level configuration, it is an extensibility point for the application level. It only works with app tokens (i.e. app 2 app), for which it happens that we only have 1 flow (client_credentials). For now (OBO for service principals could also support this if needed, it's also an app 2 app flow).

bgavrilMS:
@sameerag @tnorling - can you help with the https://github.com/AzureAD/microsoft-authentication-library-for-js/runs/7198576591?check_suite_focus=true error? (the electron app compilation issue). I do not repro this locally and I can't understand the root cause.

lib/msal-node/src/config/Configuration.ts
tnorling: Understood. The point I'm trying to make is that this configuration object is used for _all_ flows, including PublicClient flows, and if this is only relevant to one or two flows we should make it an option for only those flows rather than all flows. It's confusing for our customers to provide options that aren't relevant and ultimately get ignored.

lib/msal-common/src/config/ClientConfiguration.ts
tnorling: App developer does not use this one directly, this Configuration object is used internally across both msal-node and msal-browser/react/angular for all flows. The msal-common clients are instantiated once for each request and msal-node/browser clients are instantiated once per application instance. I'm suggesting that we only expose this for ClientCredentialClients so that only msal-node acquiring tokens using ClientCredential flows needs to understand what this is.

tnorling:
> @sameerag @tnorling - can you help with the https://github.com/AzureAD/microsoft-authentication-library-for-js/runs/7198576591?check_suite_focus=true error? (the electron app compilation issue). I do not repro this locally and I can't understand the root cause.

This is happening because the test suite is restoring dependencies from cache where the local package linking doesn't appear to be working. We need to investigate why this is happening but this PR is not at fault, feel free to ignore for now

lib/msal-node/src/client/ConfidentialClientApplication.ts
```diff
@@ -52,6 +55,17 @@ export class ConfidentialClientApplication extends ClientApplication implements
...
+        this.appTokenProvider = undefined;
+    }
+
+    /**               
+     * This extensibility point only works for the client_credential flow, i.e. acquireTokenByClientCredential and
+     * is meant for Azure SDK to enhance Managed Identity support.
+     * 
+     * @param IAppTokenProvider  - Extensibility interface, which allows the app developer to return a token from a custom source.     
+     */
+    public SetAppTokenProvider(provider: IAppTokenProvider): void {
```
tnorling: nit:
```suggestion
    setAppTokenProvider(provider: IAppTokenProvider): void {
```

tnorling: Please also add this to `IConfidentialClientApplication`

bgavrilMS: Strictly speaking, adding a method to an interface breaks backwards compat...

bgavrilMS:
Folks, can I get some help with this PR please? I see 2 failures remain (which is a lot better than yesterday!):

1. some UI tests are still failing 
2. a `tsdx` failure which I discussed with @tnorling above and looks to be a false positive. 

Please feel free to push changes if you see any issue, would be good to get this one in your next release. 

CC @hectormmg @tnorling @sameerag @peterzenz

lib/msal-common/src/config/ClientConfiguration.ts
```diff
@@ -46,7 +46,7 @@ export type ClientConfiguration = {
     telemetry?: TelemetryOptions,
     serverTelemetryManager?: ServerTelemetryManager | null,
     persistencePlugin?: ICachePlugin | null,
-    serializableCache?: ISerializableTokenCache | null
+    serializableCache?: ISerializableTokenCache | null,
```
Robbie-Microsoft, line 49: nit: can you delete all the extra space you added in here and elsewhere in this review?

lib/msal-common/test/client/ClientCredentialClient.spec.ts
```diff
@@ -360,4 +361,62 @@ describe("ClientCredentialClient unit tests", () => {
...
+    it("Uses the extensibility AppTokenProvider callback to get a token", async () => {
+        sinon.stub(Authority.prototype, <any>"getEndpointMetadataFromNetwork").resolves(DEFAULT_OPENID_CONFIG_RESPONSE.body);
+        // no need to stub out the token response, MSAL will use the AppTokenProvider instead
+
+        const config = await ClientTestUtils.createTestClientConfiguration();
+        const accessToken = "some_token";
+        const appTokenProviderResult: AppTokenProviderResult = {
+            accessToken: accessToken,
+            expiresInSeconds: 1800,
+            refreshInSeconds: 900,
```
Robbie-Microsoft, line 373: should you define these numbers as constants at the top of the file?

lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -347,8 +352,10 @@ export abstract class ClientApplication {
             },
             telemetry: this.config.telemetry,
             persistencePlugin: this.config.cache.cachePlugin,
-            serializableCache: this.tokenCache,
+            serializableCache: this.tokenCache
```
Robbie-Microsoft, line 355: why did you delete the trailing comma?

lib/msal-node/test/client/ConfidentialClientApplication.spec.ts
```diff
@@ -91,8 +91,24 @@ describe('ConfidentialClientApplication', () => {
...
 
-
     test('acquireTokenByClientCredential', async () => {
+
+        const testProvider: msalCommon.IAppTokenProvider = () => {
+            return new Promise<msalCommon.AppTokenProviderResult>(
+                (resolve) => resolve({
+                    accessToken: "accessToken",
+                    expiresInSeconds: 3601,
+                    refreshInSeconds: 1801,
```
Robbie-Microsoft, line 101: should you define these as constants at the top of the file?

Robbie-Microsoft (approved):

sameerag:
@bgavrilMS Do we know why Electron E2E failed?

derisen:
@sameerag see @tnorling's [comment above](https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/4936#issuecomment-1175344580). It only happens when the workflow runs tests on Windows -macOS seems to be fine.

PR #4945, opened 2022-06-28
replace acquireTokenSilent to ssoSilent method

dandresini:
If I edit the information from msal EditProfile flow and then I verify acquire token information the method not refresh the information if i use acquireTokenSilent.

### DIFF
--- a/samples/msal-react-samples/b2c-sample/src/pages/Protected.jsx
+++ b/samples/msal-react-samples/b2c-sample/src/pages/Protected.jsx
@@ -28,7 +28,6 @@ const ProtectedContent = () => {
             const silentRequest = {
                 ...loginRequest,
                 account: account.username,
-                forceRefresh : true
             };
 
             instance.acquireTokenSilent(silentRequest).then((response) => {
--- a/samples/msal-react-samples/b2c-sample/src/pages/Protected.jsx
+++ b/samples/msal-react-samples/b2c-sample/src/pages/Protected.jsx
@@ -20,12 +20,18 @@ const ProtectedContent = () => {
 
     useEffect(() => {
         if (!atsResponse && account && inProgress === InteractionStatus.None) {
-            const request = {
+            const request ={
                 ...loginRequest,
                 loginHint: account.username
+            }
+            
+            const silentRequest = {
+                ...loginRequest,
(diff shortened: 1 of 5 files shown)


samples/msal-react-samples/b2c-sample/src/pages/Protected.jsx
```diff
@@ -22,9 +22,10 @@ const ProtectedContent = () => {
         if (!atsResponse && account && inProgress === InteractionStatus.None) {
             const request = {
                 ...loginRequest,
-                account: account
+                loginHint: account.username
             };
-            instance.acquireTokenSilent(request).then((response) => {
+
+            instance.ssoSilent(request).then((response) => {
```
jasonnutter: I don't think we want to invoke ssoSilent this often. Have you tried invoking `acquireTokenSilent` with `forceRefresh` set to `true`?

dandresini: I'll test with force refresh and I'll send a report of result.

dandresini: I use the button "Edit profile" that call the "_instance.loginPopup(b2cPolicies.authorities.editProfile)_" to change the city from **Bari** to **Torino** as image below (in the claims i pass the city information)
B2c_EditProfile

When I show the information of token, the city claim isn't the last changed (as you said I use _forceRefresh: true_)

result_request

tnorling: B2C will keep returning the same token claims when the RT is used to acquire it so ssoSilent would be the correct way to refresh it silently. However, I agree with @jasonnutter, we shouldn't call ssoSilent every time, only when something has been updated (to ensure the cache has the updated token, and then use acquireTokenSilent for all subsequent requests).

dandresini: I used in editprofile component the method "loginRedirect" and all work fine. With loginPopUp ther'is a problem in refresh token

tnorling:
Demonstration of the edit profile flow will be added in #5015 - closing in favor of that PR

PR #4949, opened 2022-06-29
Fix prompt behavior for native requests

tnorling:
- Fixes bug where calling a silent API with prompt != none throws and instead overwrites prompt to always be none for silent calls
- Fixes bug where prompt=login skips the native broker and goes directly to web

### DIFF
--- a/lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
+++ b/lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
@@ -319,17 +319,29 @@ export class NativeInteractionClient extends BaseInteractionClient {
         const scopeSet = new ScopeSet(scopes);
         scopeSet.appendScopes(OIDC_DEFAULT_SCOPES);
 
-        if (request.prompt) {
+        const prompt = () => {
+            // If request is silent, prompt is always none
+            switch (this.apiId) {
+                case ApiId.ssoSilent:
+                case ApiId.acquireTokenSilent_silentFlow:
+                    this.logger.trace("initializeNativeRequest: silent request sets prompt to none");
+                    return PromptValue.NONE;
+                default:
+                    break;
+            }
+
+            // If request is interactive, check if prompt is allowed to go directly to native broker
             switch (request.prompt) {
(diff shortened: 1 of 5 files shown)


lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -319,17 +319,34 @@ export class NativeInteractionClient extends BaseInteractionClient {
         const scopeSet = new ScopeSet(scopes);
         scopeSet.appendScopes(OIDC_DEFAULT_SCOPES);
 
-        if (request.prompt) {
+        const prompt = () => {
```
jasonnutter:
```suggestion
        const getPrompt = () => {
```

or similar?

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -319,17 +319,34 @@ export class NativeInteractionClient extends BaseInteractionClient {
...
+                case ApiId.ssoSilent:
+                case ApiId.acquireTokenSilent_silentFlow:
+                    this.logger.trace("initializeNativeRequest: silent request sets prompt to none");
+                    return PromptValue.NONE;
+                default:
+                    break;
+            }
+
+            // If request is interactive, check if prompt is allowed to go directly to native broker
+            if (!request.prompt) {
```
jasonnutter, line 334: Not sure I follow this given the comment. Is this saying request.prompt being falsey means the request shuold go to the broker?

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
tnorling: Done

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
tnorling: Yes, rephrased to clarify

jasonnutter (approved):

peterzenz (approved):

PR #4951, opened 2022-06-29
Handle account unavailable error status

tnorling:
If a user has removed their account from WAM and MSAL.js makes a request for that account, WAM will return an ACCOUNT_UNAVAILABLE error. For silent calls this needs to be surfaced as an `InteractionRequired` error and for interactive calls this should be caught and handled by falling back to web.

### DIFF
--- a/lib/msal-browser/src/app/ClientApplication.ts
+++ b/lib/msal-browser/src/app/ClientApplication.ts
@@ -278,8 +278,8 @@ export abstract class ClientApplication {
                     return redirectClient.acquireToken(request);
                 } else if (e instanceof InteractionRequiredAuthError) {
                     this.logger.verbose("acquireTokenRedirect - Resolving interaction required error thrown by native broker by falling back to web flow");
-                    const popupClient = this.createRedirectClient(request.correlationId);
-                    return popupClient.acquireToken(request);
+                    const redirectClient = this.createRedirectClient(request.correlationId);
+                    return redirectClient.acquireToken(request);
                 }
                 this.browserStorage.setInteractionInProgress(false);
                 throw e;
--- a/lib/msal-browser/src/app/ClientApplication.ts
+++ b/lib/msal-browser/src/app/ClientApplication.ts
@@ -276,6 +276,10 @@ export abstract class ClientApplication {
                     this.nativeExtensionProvider = undefined; // If extension gets uninstalled during session prevent future requests from continuing to attempt
                     const redirectClient = this.createRedirectClient(request.correlationId);
                     return redirectClient.acquireToken(request);
+                } else if (e instanceof InteractionRequiredAuthError) {
+                    this.logger.verbose("acquireTokenRedirect - Resolving interaction required error thrown by native broker by falling back to web flow");
+                    const popupClient = this.createRedirectClient(request.correlationId);
+                    return popupClient.acquireToken(request);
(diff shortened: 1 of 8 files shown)


jasonnutter (approved):

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -276,6 +276,10 @@ export abstract class ClientApplication {
                     this.nativeExtensionProvider = undefined; // If extension gets uninstalled during session prevent future requests from continuing to attempt
                     const redirectClient = this.createRedirectClient(request.correlationId);
                     return redirectClient.acquireToken(request);
+                } else if (e instanceof InteractionRequiredAuthError) {
+                    this.logger.verbose("acquireTokenRedirect - Resolving interaction required error thrown by native broker by falling back to web flow");
```
peterzenz, line 280: I think it would be worth reporting this in our telemetry.  A boolean to indicate this has occurred would be enough.

lib/msal-browser/test/error/NativeAuthError.spec.ts
```diff
@@ -58,6 +58,17 @@ describe("NativeAuthError Unit Tests", () => {
                 expect(error.errorCode).toBe("interaction_required");
             });
 
+            it("translates ACCOUNT_UNAVAILABLE status into corresponding InteractionRequiredError", () => {
+                const error = NativeAuthError.createError("interaction_required", "interaction is required", {
+                    error: 1,
```
peterzenz, line 63: What is the "1" here?  Just a boolean true?  Or is this an error code.

peterzenz (approved):
Nothing blocking commit from my comments.

lib/msal-browser/src/app/ClientApplication.ts
tnorling: I believe this will be addressed in the additional Telemetry work I have on my radar for JS-WAM bridge

lib/msal-browser/test/error/NativeAuthError.spec.ts
tnorling: error code

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -276,6 +276,10 @@ export abstract class ClientApplication {
...
+                    const popupClient = this.createRedirectClient(request.correlationId);
```
sameerag: typo: redirecClient??

lib/msal-browser/test/error/NativeAuthError.spec.ts
sameerag, line 63: I thought we wanted enums for "error codes of WAM". Is that never pulled?

sameerag (approved):
Added minor comments, approved

lib/msal-browser/test/error/NativeAuthError.spec.ts
tnorling: Not for codes, that's for _statuses_

lib/msal-browser/src/app/ClientApplication.ts
tnorling: Oops...lazy copy pasting. Thanks for catching this!

PR #4957, opened 2022-06-30
Mark temporary cache cookies as SameSite lax

jasonnutter:
Mark cookies used by MSAL.js for temporary cache as `SameSite=Lax`, which is the default in most browsers. `SameSite` controls whether cookies are sent in cross-site contexts, which shouldn't be needed for temporary cache values. More: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite#lax

Screenshot 2022-06-30 132915

### DIFF
--- a/change/@azure-msal-browser-a0c60e1e-0205-4316-b099-d598f43c0317.json
+++ b/change/@azure-msal-browser-a0c60e1e-0205-4316-b099-d598f43c0317.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Mark temporary cache cookies as SameSite lax",
+  "comment": "Mark temporary cache cookies as SameSite lax #4957",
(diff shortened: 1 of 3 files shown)


change/@azure-msal-browser-a0c60e1e-0205-4316-b099-d598f43c0317.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Mark temporary cache cookies as SameSite lax",
```
jasonnutter:
```suggestion
  "comment": "Mark temporary cache cookies as SameSite lax #4957",
```

peterzenz (approved):

tnorling (approved):

sameerag (approved):

PR #4969, opened 2022-07-05
Update errors.md

danielvoigt:
The documentation code sample for the `Interaction_in_progress` error needs to be fixed.

`InteractionStatus` is exported from `@azure/msal-browser`, it is not exported from `@azure/msal-angular`


derisen (approved):
LGTM, thanks @danielvoigt

tnorling (approved):

PR #4971, opened 2022-07-05
Cache tokens in internal memory for native calls

sameerag:
This is an optimization for the existing native bridge. `idTokens`, `accessTokens` are cached locally in the internal memory to optimize performance for the native calls when possible. This should improve the average performance for cache lookups by avoiding a native call into the WAM cache every time.


bgavrilMS:
nit: auth_code tests are failing on this branch as well. Probably some selenium UI automation hitting a snag because service folks changed the name of some field.

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -45,6 +45,9 @@ export abstract class ClientApplication {
     // Storage interface implementation
     protected readonly browserStorage: BrowserCacheManager;
 
+    // Native Cache in memory storage implementation
+    protected readonly nativeInternalStorage: BrowserCacheManager | undefined;
```
tnorling: Can we rename this to be more general i.e. not specific to native? (`inMemoryStorage` perhaps?) I expect we will want to use this for other things as well.

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -142,6 +145,15 @@ export abstract class ClientApplication {
             new BrowserCacheManager(this.config.auth.clientId, this.config.cache, this.browserCrypto, this.logger) :
             DEFAULT_BROWSER_CACHE_MANAGER(this.config.auth.clientId, this.logger);
 
+        if (this.config.system.allowNativeBroker) {
+            const nativeCacheOptions: Required<CacheOptions> = {
+                cacheLocation: BrowserCacheLocation.MemoryStorage,
+                storeAuthStateInCookie: false,
+                secureCookies: false
+            };
+            this.nativeInternalStorage = new BrowserCacheManager(this.config.auth.clientId, nativeCacheOptions, this.browserCrypto, this.logger);
```
tnorling: Likewise, we can probably initialize this for all scenarios so we don't need to worry about `undefined`

lib/msal-browser/src/error/NativeAuthError.ts
```diff
@@ -31,7 +31,15 @@ export const NativeAuthErrorMessage = {
     userSwitch: {
         code: "user_switch",
         desc: "User attempted to switch accounts in the native broker, which is not allowed. All new accounts must sign-in through the standard web flow first, please try again."
-    }
+    },
+    tokensNotFoundInCache: {
```
tnorling, line 35: We can use the existing no_tokens_found error and leverage logger messages if we need to know which code path led to it

lib/msal-browser/src/error/NativeAuthError.ts
```diff
@@ -31,7 +31,15 @@ export const NativeAuthErrorMessage = {
...
+        code: "tokens_not_found_in_internal_memory_cache",
+        desc: "Tokens not cached in MSAL JS internal memory, please make the WAM request"
+    },
+    nativeInternalCacheNotDefined: {
```
tnorling: Shouldn't be needed if we always make sure to instantiate the in memory storage object

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -72,6 +93,83 @@ export class NativeInteractionClient extends BaseInteractionClient {
...
+    /**
+     *
+     * @param clientId
+     * @param nativeAccountId
+     * @param request
+     * @param reqTimestamp
+     * @param authority
+     * @returns
+     */
+    protected acquireTokensFromInternalCache(clientId: string, nativeAccountId: string, request: PopupRequest | SilentRequest | SsoSilentRequest, reqTimestamp: number, authority: Authority): AuthenticationResult {
```
tnorling: We should be able to use `SilentCacheClient` for this and provide it with the in memory storage instead of the browser storage object.

lib/msal-browser/src/app/ClientApplication.ts
sameerag: I intentionally named this `native...` as we already use `inMemory` in `BrowserCacheManager`(we call it `internalStorage`).
 
 `BrowserCacheManager` has three types of storage:

- `internalStorage` - An instance of `inMemory` inside the `browserStorage` instance, used for telemetry, authority etc
- `browserStorage` - instance of browser storage, can be of type `inMemory` if the user chooses it
- `temporaryStorage` - Same instance of `internalStorage` already instantiated in `browserStorage` 

By limiting the new type to `native` calls only, I think, will make it clear that a new instance of `inMemory` storage is created along with `browserStorage` only for `native` cache purposes. All other `inMemory` needs can be met within `browserStorage` instance.
  
We can eventually unify all the different internal memory references, may be. However, it is a major refactor I do not want to start now. Let us discuss if you still believe the name change is needed.

lib/msal-browser/src/app/ClientApplication.ts
sameerag: I have limited this by design. Explanation added above.

lib/msal-browser/src/error/NativeAuthError.ts
sameerag: Well, thought about it. May be the differentiation is not needed after all. This is to address the scenarios that `nativeInternalStorage` can be `undefined`. I can declare it for all use cases and forget this nuance, functionally it is the same.

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
sameerag: Not a bad idea, let me check if this optimization can be done.

lib/msal-browser/src/app/ClientApplication.ts
tnorling: > By limiting the new type to native calls only, I think, will make it clear that a new instance of inMemory storage is created along with browserStorage only for native cache purposes.

I think this is an artificial constraint. There's no reason this object needs to be scoped only to native caching.

> All other inMemory needs can be met within browserStorage instance.

The `internalStorage` inside `BrowserCacheManager` is not the same as this. It is limited to just an implementation of getters and setters, rather than an implementation of our specific caching structure (which this is).

> We can eventually unify all the different internal memory references, may be. However, it is a major refactor I do not want to start now.

While I do think our cache could use some cleanup or refactor, I think taking an incremental step with this PR is fine, we don't need to unify things now but I also think this is largely a separate conversation. That conversation doesn't need to inform the purpose of this new object.

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
sameerag: `SilentCacheClient` assumes we use RT to renew tokens. It also reads all tokens from cache, and not just `idToken` and `accessToken`. I would rather keep this simple rather than refactor the code in `SilentCacheClient`.

lib/msal-browser/src/app/ClientApplication.ts
sameerag: > > By limiting the new type to native calls only, I think, will make it clear that a new instance of inMemory storage is created along with browserStorage only for native cache purposes.
> 
> I think this is an artificial constraint. There's no reason this object needs to be scoped only to native caching.
> 
> > All other inMemory needs can be met within browserStorage instance.
> 
> The `internalStorage` inside `BrowserCacheManager` is not the same as this. It is limited to just an implementation of getters and setters, rather than an implementation of our specific caching structure (which this is).
> 

 For both these cases, If a user wants to use `internalMemory` for all their token storage, they can specify the `cacheLocation` as `MemoryStorage`. Do we envision any other use case outside of this, to use `browserStorage` (assume `localStorage` or `sessionStorage`) and cannot leverage `internalStorage` within the `browserStorage`?

sameerag: We agreed on creating a `nativeInternalStorage` instance.

lib/msal-browser/src/error/NativeAuthError.ts
sameerag: Agreed to keep the name.

lib/msal-browser/src/error/NativeAuthError.ts
sameerag: Modified

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
sameerag: Refactored this code to call `SilentCacheClient` instead.

lib/msal-browser/src/error/NativeAuthError.ts
```diff
@@ -87,9 +91,17 @@ export class NativeAuthError extends AuthError {
...
-     * @returns 
+     * @returns
      */
     static createUserSwitchError(): NativeAuthError {
         return new NativeAuthError(NativeAuthErrorMessage.userSwitch.code, NativeAuthErrorMessage.userSwitch.desc);
     }
+
+    /**
+     * Creates a tokens not found error when the internal cache look up fails
+     * @returns
```
Robbie-Microsoft: missing return type?

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -42,14 +47,31 @@ export class NativeInteractionClient extends BaseInteractionClient {
...
+            const result = await this.acquireTokensFromCache(this.accountId, nativeRequest);
+            nativeATMeasurement.endMeasurement({
+                success: true,
+                isNativeBroker: true,
+                fromCache: true
+            });
+            return result;
+        } catch (e) {
+            // continue with a native call for any and all errors
+            this.logger.info("MSAL internal Cache does not contain tokens, proceed to make a native call");
```
Robbie-Microsoft, line 66: should you make a new error type for this?

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -72,6 +94,47 @@ export class NativeInteractionClient extends BaseInteractionClient {
             });
     }
 
+    /**
+     * Creates silent flow request
+     * @param request
+     * @param cachedAccount
+     * @returns
```
Robbie-Microsoft: missing return type

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -72,6 +94,47 @@ export class NativeInteractionClient extends BaseInteractionClient {
...
+     */
+    private createSilentCacheRequest(request: NativeTokenRequest, cachedAccount: AccountInfo): CommonSilentFlowRequest {
+        return {
+            authority: request.authority,
+            correlationId: this.correlationId,
+            scopes: ScopeSet.fromString(request.scopes).asArray() ,
```
Robbie-Microsoft: nit: delete the space in front of the "," at the end of the line

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -676,6 +682,27 @@ export abstract class CacheManager implements ICacheManager {
         return this.getAccount(accountKey);
     }
 
+    /**
+     * Retrieve AccountEntity from cache
+     * @param account
+     */
```
Robbie-Microsoft, line 689: missing @returns

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -270,7 +281,7 @@ export abstract class ClientApplication {
         let result: Promise<void>;
 
         if (this.nativeExtensionProvider && this.canUseNative(request)) {
-            const nativeClient = new NativeInteractionClient(this.config, this.browserStorage, this.browserCrypto, this.logger, this.eventHandler, this.navigationClient, ApiId.acquireTokenRedirect, this.performanceClient, this.nativeExtensionProvider, this.getNativeAccountId(request), request.correlationId);
+            const nativeClient = new NativeInteractionClient(this.config, this.browserStorage, this.browserCrypto, this.logger, this.eventHandler, this.navigationClient, ApiId.acquireTokenRedirect, this.performanceClient, this.nativeExtensionProvider, this.getNativeAccountId(request), this.nativeInternalStorage, request.correlationId);
             result = nativeClient.acquireTokenRedirect(request).catch((e: AuthError) => {
                 if (e instanceof NativeAuthError && e.isFatal()) {
                     this.nativeExtensionProvider = undefined; // If extension gets uninstalled during session prevent future requests from continuing to attempt
```
peterzenz, line 287: Might be worth adding a datapoint to our telemetry to let us know that this is happening.

lib/msal-browser/src/error/NativeAuthError.ts
```diff
@@ -52,7 +56,7 @@ export class NativeAuthError extends AuthError {
         if (this.ext && this.ext.status && (this.ext.status === NativeStatusCode.PERSISTENT_ERROR || this.ext.status === NativeStatusCode.DISABLED)) {
             return true;
         }
-
```
peterzenz, line 55: Appreciate the code cleanup, but recommend putting code cleanup in a separate PR, this makes the impacting changes easier to review.

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -246,7 +309,35 @@ export class NativeInteractionClient extends BaseInteractionClient {
...
+            homeAccountIdentifier,
+            request.authority,
+            response.id_token || Constants.EMPTY_STRING,
+            request.clientId,
+            idTokenObj.claims.tid || Constants.EMPTY_STRING,
+        );
+        this.nativeStorageManager.setIdTokenCredential(idTokenEntity);
+
+        // cache accessToken in inmemory storage
+        const expiresIn: number = (responseTokenType === AuthenticationScheme.POP)
```
peterzenz, line 323: Consider reformatting the nested ternary operators here for readability.

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
sameerag: Done

lib/msal-browser/src/error/NativeAuthError.ts
sameerag: Added

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
sameerag: Done

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
sameerag: This is intentional as we anyway fall back to network call and there is nothing to handle. Did not want to add an error type when there is no next step of handling the specific error except convey the message(which logs do here)

lib/msal-common/src/cache/CacheManager.ts
sameerag: Added.

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -366,7 +457,6 @@ export class NativeInteractionClient extends BaseInteractionClient {
                 telemetry: NativeConstants.MATS_TELEMETRY
             },
             extendedExpiryToken: false // Make this configurable?
```
peterzenz, line 463: Please sync with CCS on this to ensure this is the right thing.  They're adding a lot of coverage to web apps and we don't want the WAM side to break this.

lib/msal-common/src/utils/Constants.ts
```diff
@@ -47,7 +47,8 @@ export const Constants = {
     REGIONAL_AUTH_PUBLIC_CLOUD_SUFFIX: "login.microsoft.com",
     KNOWN_PUBLIC_CLOUDS: ["login.microsoftonline.com", "login.windows.net", "login.microsoft.com", "sts.windows.net"],
     TOKEN_RESPONSE_TYPE: "token",
-    ID_TOKEN_RESPONSE_TYPE: "id_token"
+    ID_TOKEN_RESPONSE_TYPE: "id_token",
+    SHR_NONCE_VALIDITY: 240,
```
peterzenz, line 51: Do we have to hardcode this?  It would be great if there was a signal coming in to us on the length in case an RP wanted to alter this.

peterzenz (approved):
Some minor comments, nothing blocking.  Thanks.

lib/msal-browser/test/interaction_client/NativeInteractionClient.spec.ts
```diff
@@ -3,30 +3,70 @@
...
+
+const testAccessTokenEntity: AccessTokenEntity = new AccessTokenEntity();
+testAccessTokenEntity.homeAccountId = `${ID_TOKEN_CLAIMS.oid}.${ID_TOKEN_CLAIMS.tid}`;
+testAccessTokenEntity.clientId = TEST_CONFIG.MSAL_CLIENT_ID;
+testAccessTokenEntity.environment = testAccountEntity.environment;
+testAccessTokenEntity.realm = ID_TOKEN_CLAIMS.tid;
+testAccessTokenEntity.secret = TEST_TOKENS.ACCESS_TOKEN;
+testAccessTokenEntity.target = TEST_CONFIG.DEFAULT_SCOPES.join(" ");
+testAccessTokenEntity.credentialType = CredentialType.ACCESS_TOKEN;
+testAccessTokenEntity.expiresOn = `${TimeUtils.nowSeconds() + 3600}`;
```
Robbie-Microsoft: should you define 3600 as a constant?

Robbie-Microsoft (approved):

lib/msal-browser/src/app/ClientApplication.ts
sameerag: Well we have a perf entry in `nativeInteractionClient` instance, we do not differentiate cache locally vs WAM cache with a separate sub-measurement. However, we can do the mapping indirectly.

lib/msal-browser/src/error/NativeAuthError.ts
sameerag: Was auto corrected using my editor, I usually diff by ignoring white spaces. May be standardizing UX settings across the board will help.

lib/msal-common/src/utils/Constants.ts
sameerag: It is agreed that nonce validity is 5 mins. The only way to find if the SHR is valid is to send the nonce to the RP. I am sure it can be customized, will follow up with a separate PR if there is a reliable method to fetch this info over the wire.

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
sameerag: Sure. Will do.

lib/msal-browser/test/interaction_client/NativeInteractionClient.spec.ts
sameerag: Done, moved to a constant.

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
sameerag: Done

PR #4973, opened 2022-07-05
Token sizes in the perf telemetry data 

bmahall:
This PR:-

Obtains token sizes {Id Token, Access Token} from the appropriate sub measurements into the high level Event data. 

- Adds the properties : IdTokenSize, AccessTokenSize} in Performance Event


lib/msal-common/src/response/AuthenticationResult.ts
```diff
@@ -39,5 +39,6 @@ export type AuthenticationResult = {
     cloudGraphHostName?: string;
     msGraphHost?: string;
     code?: string;
+    tokenSizes?: object;
```
jasonnutter: I don't think we should expose this on the result object. Should just be used internally.

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
```diff
@@ -17,23 +17,26 @@ export abstract class PerformanceClient implements IPerformanceClient {
     protected clientId: string;
     protected logger: Logger;
     protected callbacks: Map<string, PerformanceCallbackFunction>;
-    
+    protected idTokenSize?: number;
+    protected accessTokenSize?: number;
+    protected refreshTokenSize?: number;
```
jasonnutter: Why are these class properties? An application can have multiple tokens at a time, so singletons doesn't really make sense.

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
```diff
@@ -104,7 +110,10 @@ export abstract class PerformanceClient implements IPerformanceClient {
             clientId: this.clientId,
             name: measureName,
             startTimeMs: Date.now(),
-            correlationId: eventCorrelationId
+            correlationId: eventCorrelationId,
+            idTokenSize: this.idTokenSize,
+            accessTokenSize: this.accessTokenSize,
+            refreshTokenSize: this.refreshTokenSize,
```
jasonnutter: These should be passed in on `endMeasurement`, not start.

jasonnutter: I would propose that instead of attaching the sizes as properties on the performanceclient, to instead include them as properties on whatever low-level submeasurement has access to the sizes, and then pull them out and attach them as top-level properties on teh performance event in `flushMeasurement`.

jasonnutter: For example:

```typescript
                const eventToEmit = sortedCompletedEvents.reduce((previous, current) => {
                    if (current.name !== measureName) {
                        this.logger.trace(`PerformanceClient: Complete submeasurement found for ${current.name}`, correlationId);
                        // TODO: Emit additional properties for each subMeasurement
                        const subMeasurementName = `${current.name}DurationMs`;
                        
                        /*
                         * Some code paths, such as resolving an authority, can occur multiple times.
                         * Only take the first measurement, since the second could be read from the cache,
                         * or due to the same correlation id being used for two distinct requests.
                         */
                        if (!previous[subMeasurementName]) {
                            previous[subMeasurementName] = current.durationMs;
                        } else {
                            this.logger.verbose(`PerformanceClient: Submeasurement for ${measureName} already exists for ${current.name}, ignoring`, correlationId);
                        }

                         const accessTokenSize = current.accessTokenSize;
                         if (!previous.accessTokenSize) {
                              previous.accessTokenSize = accessTokenSize;
                         }
                    }
                    return previous;
                }, topLevelEvent);

```

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
peterzenz: It would be nice to support a collection of <tokenType, size> in the library but I'm not sure how that would look on the Aria/Kusto side.  It would be helpful if in a given request, we had an aggregate size of Code/ID/Access/Refresh tokens returned in a given call.  Maybe we can aggregate these so the results are number of tokens per type, and then total size of tokens of that type and that is how we represent this in the telemetry we track.

lib/msal-common/src/response/AuthenticationResult.ts
bmahall: Got it. 
Additionally, is it a good idea to include Refresh token as part of AuthenticationResult to be able to access the size?

change/@azure-msal-browser-0daaf89f-81ba-4902-b92d-14bd6eb15b19.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Added new token size field in the perf telemetry data",
```
tnorling:
```suggestion
  "comment": "Added new token size field in the perf telemetry data #4973",
```

change/@azure-msal-common-50b3a71d-2ba8-4276-a7d0-7bbf1c2d408a.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Added new token size field in perf telemetry data",
```
tnorling:
```suggestion
  "comment": "Added new token size field in perf telemetry data #4973",
```

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -65,7 +65,7 @@ export class RefreshTokenClient extends BaseClient {
      * Gets cached refresh token and attaches to request, then calls acquireToken API
      * @param request
      */
-    public async acquireTokenByRefreshToken(request: CommonSilentFlowRequest): Promise<AuthenticationResult> {
+    public async acquireTokenByRefreshToken(request: CommonSilentFlowRequest): Promise<{ refreshTokenSize: number, result: AuthenticationResult }> {
```
tnorling: This is a breaking change. Can we have this API report a sub measurement instead of changing the return type?

lib/msal-common/src/telemetry/performance/PerformanceEvent.ts
```diff
@@ -94,6 +94,22 @@ export enum PerformanceEvents {
      * Used to acquire a token from Native component when native brokering is enabled.
      */
     NativeInteractionClientAcquireToken = "nativeInteractionClientAcquireToken",
+
+    /**
+     * Used to measure the time taken for completing embedded-broker handshake (PW-Broker).
+     */
+    BrokerHandhshake = "brokerHandshake",
```
sameerag, line 101: I assume the corresponding code is added in the PWB code related PR in the 1P library?

sameerag (approved):

tnorling (approved):

lib/msal-common/src/telemetry/performance/PerformanceEvent.ts
bmahall: Yes !

jasonnutter: Yep.

jasonnutter (approved):

PR #4990, opened 2022-07-06
Send login_hint claim instead of sid if available

jasonnutter:
Similar to #5048, sends `login_hint` claim instead of `sid`, if available.


lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -354,16 +353,29 @@ export class AuthorizationCodeClient extends BaseClient {
...
+                // If login_hint claim is present, use it over sid/username
+                if (accountLoginHintClaim) {
+                    this.logger.verbose("createAuthCodeUrlQueryString: login_hint claim present on account");
+                    parameterBuilder.addLoginHint(accountLoginHintClaim);
+                    try {
+                        const clientInfo = buildClientInfoFromHomeAccountId(request.account.homeAccountId);
+                        parameterBuilder.addCcsOid(clientInfo);
+                    } catch (e) {
+                        this.logger.verbose("createAuthCodeUrlQueryString: Could not parse home account ID for CCS Header");
+                    }
```
tnorling, line 366: Should this move down below the if/else so we don't have to duplicate this logic?

tnorling (approved):
Minor suggestion otherwise looks good

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -354,16 +353,29 @@ export class AuthorizationCodeClient extends BaseClient {
...
+                     * SessionId is only used in silent calls
+                     */
                     this.logger.verbose("createAuthCodeUrlQueryString: Prompt is none, adding sid from account");
                     parameterBuilder.addSid(accountSid);
                     try {
                         const clientInfo = buildClientInfoFromHomeAccountId(request.account.homeAccountId);
                         parameterBuilder.addCcsOid(clientInfo);
                     } catch (e) {
-                        this.logger.verbose("Could not parse home account ID for CCS Header: " + e);
+                        this.logger.verbose("createAuthCodeUrlQueryString: Could not parse home account ID for CCS Header");
```
peterzenz, line 378: Why eat the exception here and on 392 just below?

peterzenz (approved):
Not blocking, just wondering on why we're not logging the exception on those two lines.

lib/msal-common/src/client/AuthorizationCodeClient.ts
jasonnutter: I believe because adding the ccs parameters are a best-effort and it failing shouldn't fail the auth request.

lib/msal-common/src/client/AuthorizationCodeClient.ts
jasonnutter: Hmm not sure it makes it much cleaner, a little more readable this way imo.

lib/msal-browser/docs/login-user.md
```diff
@@ -101,13 +101,15 @@ If you already have a session that exists with the authentication server, you ca
...
+
+It is recommended to leverage the [`login_hint` optional ID token claim](https://docs.microsoft.com/azure/active-directory/develop/active-directory-optional-claims#v10-and-v20-optional-claims-set) (provided to `ssoSilent` as `loginHint`), as it is the most reliable account hint of silent (and interactive) requests.
 
 - `account` (which can be retrieved using on of the [account APIs](./accounts.md))
 - `sid` (which can be retrieved from the `idTokenClaims` of an `account` object)
-- `login_hint` (which can be retrieved from the account object `username` property or the `upn` claim in the ID token)
+- `login_hint` (which can be retrieved from either the account object `login_hint` ID token claim, `username` property, or the `upn` ID token claim)
 
-Passing an account will look for sid in the token claims, then fall back to loginHint (if provided) or account username.
+Passing an account will look for the `login_hint` optional ID token claim (preferred), then the `sid` optional id token claim, then fall back to `loginHint` (if provided) or account username.
```
jasonnutter, line 112: FYI @sameerag ^^

PR #5014, opened 2022-07-18
Implemented Cache Lookup Policy for acquireTokenSilent

### LINKED ISSUE #4104: Can't catch invalid_grant error after refresh token revoke
### Core Library

### Core Library Version

2.14.1

### Wrapper Library

Not Applicable

### Wrapper Library Version

none

### Description

We have an Azure AD B2C with custom policies configured.
By our security policy we are going to revoke all refresh tokens once user changed his password.
The expected behavior is like that;
1. Revoke all refresh tokens for a particular user
2. Access token is expired for that user
3. User initiates any request which under the hood is calling `acquireTokenSilent`
4. Token endpoint returns `invalid_grant` error
5. We catch that error and sign the user out immediately.

Even though Msal.js v2 receives this error under the hood, it doesn't allow developers to detect it and does a silent renew of the access/refresh token pair.

FYI: I've explored the source code a little bit and found that there's no way to abandon silent request execution in the iframe in case of `invalid_grant` error. Does it mean that the new feature/flag is required? What do you think?
Here is a reference to the code for your information:
https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/0e8a18c059bbacbd7f68057d1e760d9e52bdfa02/lib/msal-browser/src/app/ClientApplication.ts#L325

### Error Message

Only network error with `invalid_grant` token endpoint error, which is correct.

### Msal Logs

image

### MSAL Configuration

```javascript
Any relevant B2C config from this repository can be used.
```

### Relevant Code Snippets

```javascript
Any B2C example from this repository can be used
```

### Reproduction Steps

1. Sign in under any Azure AD B2C user. 
2. Wait for access token expiration (I used minimal access token's expiration time = 5min). 
3. In Azure Portal click 'Revoke sessions' button for the user.
4. In the web app put a breakpoint on `catch` clause of for `PublicClientApplication.acquireTokenSilent`
5. Initiate a request in the web application which will cause a new token acquisition.

Problems:
1. Error is not caught (execution flow won't stop on the breakpoint).
2. New refresh token/access token received successfully.

### Expected Behavior

1. I wanna be able to catch the `invalid_grant` (revoked/expired refresh token) exception and sign the user out immediately.

Generally, we are going to revoke all refresh tokens once user's changed his password. That's a part of our security policy.

### Identity Provider

Azure B2C Custom Policy

Chrome, Safari

### Regression

_No response_

### Source

External (Customer)

### LINKED ISSUE #4153: Retrieving Current AccessToken Expiration Extends It
### Core Library

### Core Library Version

2.17.0

### Wrapper Library

### Wrapper Library Version

1.0.2

### Description

I used to retrieve how long user's access token available from 1st approach

I got a feedback from @tnorling that, it's not guaranteed that this would give up to date expiration date for user's access token. So I was recommended to use second approach 

Using 2nd approach, I am able to say 1st approach does bring old expiration time. The problem is in the 2nd approach updating expiration time even though, I am not intending to do that. All I need to do is to know when this user's access token will expire.

My goal explanation, if matters is here. Access token needs to be active only 1 hour from the last time user used it. If user sign-in at 1pm, they take some action at 1:20pm, their access token at that time should expire at 2:20pm. The reason I ask for expiration time of the Access Token is that, if user does not take an action for 1 hour, I need to log them out. If I do not check their access token user stays logged in forever. If someone logs in and leave their PC on for hours someone else can click and load up things through API.

### Error Message

_No response_

### Msal Logs

_No response_

### MSAL Configuration

```javascript
export const b2cPolicies = {
  names: {
    signUpSignIn: "***",
    forgotPassword: "***",
  },
  authorities: {
    signUpSignIn: {
      authority:
        "***",
    },
    forgotPassword: {
      authority:
        "***",
    },
  },
  authorityDomain: "loginservicedev.b2clogin.com",
};
/**
 * Configuration object to be passed to MSAL instance on creation.
 * For a full list of MSAL.js configuration parameters, visit:
 * https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/configuration.md
 */
export const msalConfig = {
  auth: {
    clientId: "****", // This is the ONLY mandatory field that you need to supply.
    authority: b2cPolicies.authorities.signUpSignIn.authority, // Choose SUSI as your default authority.
    knownAuthorities: [b2cPolicies.authorityDomain], // Mark your B2C tenant's domain as trusted.
    redirectUri: "/", // You must register this URI on Azure Portal/App Registration. Defaults to window.location.origin
    postLogoutRedirectUri: "/", // Indicates the page to navigate after logout.
    navigateToLoginRequestUrl: false, // If "true", will navigate back to the original request location before processing the auth code response.
  },
  cache: {
    cacheLocation: "localStorage", // Configures cache location. "sessionStorage" is more secure, but "localStorage" gives you SSO between tabs.
  },
  system: {
    tokenRenewalOffsetSeconds: 1740,
  },
};
```

### Relevant Code Snippets

```javascript
1st approach

export const b2cGetSessionTimeLeft = () => {
  return Math.round(
    (accounts[0]?.idTokenClaims.exp - Math.round(Date.now() / 1000)) / 60
  );
};
```

2nd approach
```
function fetchToken(forceRefresh = false) {
  params.forceRefresh = forceRefresh;
  return instance
    .acquireTokenSilent(params)
    .then((response) => response)
    .catch(() =>
      instance
        .acquireTokenRedirect(params)
        .then((response) => response)
        .catch(() => false)
    );
}
export const b2cTimeLeft = async () => {
  let expDateTime = await fetchToken().then((res) => res.expiresOn);
  return Math.round(
    (Math.round(expDateTime / 1000) - Math.round(Date.now() / 1000)) / 60
  );
};
```
```


### Reproduction Steps

Request token using acquireTokenSilent and it will extend accesstoken expireson.

### Expected Behavior

Do not extend expireson on each acquireTokenSilent call unless forcerefresh is true.

### Identity Provider

Azure B2C Basic Policy

Chrome

### Regression

_No response_

### Source

Internal (Microsoft)

### PR DESCRIPTION
**Pull request by Robbie-Microsoft**

### CODE REVIEW

File: lib/msal-browser/src/app/PublicClientApplication.ts [hunk 1]
```diff
@@ -93,14 +93,22 @@ export class PublicClientApplication extends ClientApplication implements IPubli
      * @returns {Promise.<AuthenticationResult>} - a promise that is fulfilled when this function has completed, or rejected if an error was raised. Returns the {@link AuthResponse} object
      */
     async acquireTokenSilent(request: SilentRequest): Promise<AuthenticationResult> {
+        // check to make sure a valid silent token retrieval stratgey was passed in, if applicable
+        if (request.silentTokenRetrievalStrategy && !Object.values(SilentTokenRetrievalStrategy).includes(request.silentTokenRetrievalStrategy as SilentTokenRetrievalStrategy)) {
```
hectormmg: I suggest we move this check and error into a helper method. Something like `validateSilentTokenRetrievalStrategy()` that checks and throws if the strategy is set but not supported.

lib/msal-browser/src/error/BrowserAuthError.ts
```diff
@@ -184,7 +184,15 @@ export const BrowserAuthErrorMessage = {
     nativePromptNotSupported: {
         code: "native_prompt_not_supported",
         desc: "The provided prompt is not supported by the native platform. This request should be routed to the web based flow."
-    }
+    },
+    accessCodeNotInCache: {
```
hectormmg:
```suggestion
    accessTokenNotInCache: {
```

Nit: For consistency, I suggest "token" instead of "code"

lib/msal-browser/src/request/SilentRequest.ts
```diff
@@ -26,4 +26,5 @@ export type SilentRequest = Omit<CommonSilentFlowRequest, "authority"|"correlati
     account?: AccountInfo;
     correlationId?: string;
     forceRefresh?: boolean;
+    silentTokenRetrievalStrategy?: String;
```
hectormmg: You can use the enum as the type, just make sure you import the enum in this file.

```suggestion
    silentTokenRetrievalStrategy?: SilentTokenRetrievalStrategy;
```

This way the type checker can make sure devs only pass supported types in and we have an extra bug detection layer.

lib/msal-common/src/request/CommonSilentFlowRequest.ts
```diff
@@ -22,4 +22,5 @@ export type CommonSilentFlowRequest = BaseAuthRequest & {
     account: AccountInfo;
     forceRefresh: boolean;
     tokenQueryParameters?: StringDict;
+    silentTokenRetrievalStrategy?: String;
```
hectormmg:
```suggestion
    silentTokenRetrievalStrategy?: SilentTokenRetrievalStrategy;
```

Again, make sure you import the enum into this file as well if you make this change.

lib/msal-browser/src/app/PublicClientApplication.ts
jasonnutter: I don't think we need this check. The enum clearly documents which values are expected, and downstream code can make explicit checks for each of those values, and a value that isn't one of those can be ignored.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -114,35 +122,50 @@ export class PublicClientApplication extends ClientApplication implements IPubli
             sshKid: request.sshKid
         };
         const silentRequestKey = JSON.stringify(thumbprint);
+
         const cachedResponse = this.activeSilentTokenRequests.get(silentRequestKey);
         if (typeof cachedResponse === "undefined") {
+            if (request.silentTokenRetrievalStrategy === SilentTokenRetrievalStrategy.CacheOnly) {
+                throw BrowserAuthError.createAccessTokenNotInCacheError();
```
jasonnutter: `cachedresponse` here is referring to deduplication we do for multiple concurrent ongoing requests, but does not refer to the token cache. This should be removed.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -114,35 +122,50 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
+
+            if (tokenError) {
+                throw tokenError;
+            }
+
+            const tokenPromise = new Promise<AuthenticationResult>((resolve) => {
+                resolve(token);
+            });
+            this.activeSilentTokenRequests.set(silentRequestKey, tokenPromise);
+            return tokenPromise;
```
jasonnutter: Why is this refactored? Seems unrelated (and the usage of .then/.catch was intentional).

jasonnutter: And yeah, this isn't working the way the previous code was, as youll note by failing unit tests. Please revert, thanks!

lib/msal-browser/src/app/PublicClientApplication.ts
jasonnutter: The place where you need to make changes is in `acquireTokenSilentAsync`, as that is where the core high-level business logic is for looking in the cache first, then going to the network.

jasonnutter: For example, you add code in `silentCacheClient.acquireToken` to immediately throw an error (preferably an existing error that is already handled downstream) if the enum value is one of the network only values.

jasonnutter: And then make sure `acquireTokenByRefreshToken` is only called when the value is one of the network only values.

jasonnutter: (and then further logic inside of there based on which network values are provided)

lib/msal-browser/src/app/PublicClientApplication.ts
Robbie-Microsoft: @jasonnutter I don't see any cons of leaving it in. It only helps the developer of an app due to user error (typo).

lib/msal-browser/src/app/PublicClientApplication.ts
Robbie-Microsoft: I refactored it so it would be easier to read for us as developers. It still functions the same way... Why is the .then/.catch intentional?

Robbie-Microsoft: I could update the unit tests...

jasonnutter: It does not function the same. We use promise chain explicitly because we create the promise chain and then memorize it, so that if another request with the same request thumbprint is made, we return the previous promise for that request, instead of making a new request.

jasonnutter: See #3837. Async/await does not work here because `await` executes the promise, and returns the resolved value. We need a reference to the underlying promise, so that we can reuse it.

jasonnutter: We'll make sure there is a comment explaining this for future reference.

Robbie-Microsoft: I see, thanks for explaining.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -188,6 +193,10 @@ export class PublicClientApplication extends ClientApplication implements IPubli
             const silentCacheClient = this.createSilentCacheClient(request.correlationId);
             const silentRequest = await silentCacheClient.initializeSilentRequest(request, account);
             result = silentCacheClient.acquireToken(silentRequest).catch(async () => {
+                if (request.silentTokenRetrievalStrategy === SilentTokenRetrievalStrategy.CacheOnly) {
+                    throw new AuthError("", "Can't make network call when SilentTokenRetrievalStrategy is set to CacheOnly");
```
jasonnutter: Please just rethrow the error thrown by `silentCacheClient.acquireToken`

jasonnutter: (an additional log message would be helpful)

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -188,6 +193,10 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
+                }
```
jasonnutter:
```suggestion
            result = silentCacheClient.acquireToken(silentRequest).catch(async (error) => {
                if (request.silentTokenRetrievalStrategy === SilentTokenRetrievalStrategy.CacheOnly) {
                    this.logger.error("SilentTokenRetrievalStrategy is set to CacheOnly, not attempting to renew token silently.");
                    throw error;
                }
```

lib/msal-browser/src/interaction_client/SilentRefreshClient.ts
```diff
@@ -14,6 +14,10 @@ export class SilentRefreshClient extends StandardInteractionClient {
      * @param request
      */
     async acquireToken(request: CommonSilentFlowRequest): Promise<AuthenticationResult> {
+        if (request.silentTokenRetrievalStrategy === SilentTokenRetrievalStrategy.NetworkOnly) {
+            throw new ServerError(BrowserConstants.INVALID_GRANT_ERROR);
```
jasonnutter: Can you also add a log message here indicating that this if statement was entered?

jasonnutter:
Can you please add a section to this document detailing each of these enumeration options?https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/acquire-token.md

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -614,7 +615,14 @@ export abstract class ClientApplication {
...
                 const isInvalidGrantError = (e.errorCode === BrowserConstants.INVALID_GRANT_ERROR);
-                if (isServerError && isInvalidGrantError && !isInteractionRequiredError) {
+
+                /*
+                 * if RT has expired and SilentTokenRetrievalStrategy is not set to use only existing RT,
+                 * then renew the RT with a hidden iFrame (which uses the network)
+                 */
+                if (isServerError && isInvalidGrantError && !isInteractionRequiredError
+                    && (request.silentTokenRetrievalStrategy !== SilentTokenRetrievalStrategy.NetworkWithExistingRefreshTokenOnly)
+                    && (request.silentTokenRetrievalStrategy !== SilentTokenRetrievalStrategy.CacheOrExistingRefreshToken)) {
```
jasonnutter: Can you add a log statement if either of these retrieval strategies are true, and print a log statement?

For example:

```typescript
if (isServerError && isInvalidGrantError && !isInteractionRequiredError
                    && (request.silentTokenRetrievalStrategy !== SilentTokenRetrievalStrategy.NetworkWithExistingRefreshTokenOnly)
                    && (request.silentTokenRetrievalStrategy !== SilentTokenRetrievalStrategy.CacheOrExistingRefreshToken)) {
    // ...
}

if (request.silentTokenRetrievalStrategy === SilentTokenRetrievalStrategy.NetworkWithExistingRefreshTokenOnly || request.silentTokenRetrievalStrategy === SilentTokenRetrievalStrategy.CacheOrExistingRefreshToken) {
    this.logger.info("SilentTokenRetrievalStrategy set to not renew expired refresh token.");
}

throw e;
````

File: lib/msal-common/src/utils/Constants.ts [hunk 11]
```diff
@@ -401,3 +401,21 @@ export enum JsonTypes {
     Jwt = "JWT",
     Jwk = "JWK"
 }
+
+export enum SilentTokenRetrievalStrategy {
```
jasonnutter: This should be exported from the `index.ts` file for msal-common, and then rexported in msal-browser's `index.ts`.

lib/msal-common/src/utils/Constants.ts
```diff
@@ -401,3 +401,21 @@ export enum JsonTypes {
...
+    /*
+     * look in cache, if not in cache then use existing RT to get AT
+     * if existing RT is not valid, do not attempt to silently renew it
+     */
+    CacheOrExistingRefreshToken = 2,
+    // Only go to network, use existing RT but don't renew it
+    NetworkWithExistingRefreshTokenOnly = 3,
+    // Only go to network, use existing RT and renew it (existing forceRefresh flag equals true)
+    NetworkWithRefreshToken = 4,
+    // Only go to network, don't use existing RT, get a new one
```
jasonnutter: Can you clean up these descriptions? They will be shown to developers in their IDE.

```suggestion
    // acquireTokenSilent will only go to network, and will not use cached access token or refresh token.
```

Robbie-Microsoft:
> Can you please add a section to this document detailing each of these enumeration options?https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/acquire-token.md

How can I go about updating https://azuread.github.io/microsoft-authentication-library-for-js/ref/modules/_azure_msal_browser.html#silentrequest ?

jasonnutter:
> > Can you please add a section to this document detailing each of these enumeration options?https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/acquire-token.md
> 
> How can I go about updating https://azuread.github.io/microsoft-authentication-library-for-js/ref/modules/_azure_msal_browser.html#silentrequest ?

We have putting the comments above the `SilentRequest` type definition itself, but really, we should be adding them in the place where they are defined which in this case would be above the type definition for it in ` lib/msal-common/src/request/CommonSilentFlowRequest.ts`.

lib/msal-browser/docs/acquire-token.md
```diff
@@ -27,6 +27,14 @@ You can read more about using `acquireTokenSilent` [here](./token-lifetimes.md).
 
 You must pass a request object to the acquireToken APIs. This object allows you to use different parameters in the request. See [here](./request-response-object.md) for more information on the request object parameters. Scopes are required for all acquireToken calls.
 
+A Silent Token Retrieval Strategy can be optionally provided to the request. The Retrieval Strategies are:
+- CacheAndNetwork - acquireTokenSilent will first look in the cache before going to the network with the existing refresh token. The refresh token will be renewed if it is expired. This is existing default behavior.
```
tnorling:
```suggestion
- Default - `acquireTokenSilent` will attempt to retrieve an access token from the cache. If the access token is expired or cannot be found the refresh token will be used to acquire a new one. Finally, if the refresh token is expired `acquireTokenSilent` will attempt to acquire new access and refresh tokens.
```

lib/msal-browser/docs/acquire-token.md
```diff
@@ -27,6 +27,14 @@ You can read more about using `acquireTokenSilent` [here](./token-lifetimes.md).
...
+- CacheOnly - acquireTokenSilent will only look in the cache, and will not go to the network with the existing refresh token. The refresh token will not be renewed if it is expired.
```
tnorling:
```suggestion
- CacheOnly - `acquireTokenSilent` will only look for access tokens in the cache. It will not attempt to renew access or refresh tokens.
```

lib/msal-browser/docs/acquire-token.md
```diff
@@ -27,6 +27,14 @@ You can read more about using `acquireTokenSilent` [here](./token-lifetimes.md).
...
+- CacheOrExistingRefreshToken - acquireTokenSilent will first look in the cache before going to the network with the existing refresh token. The refresh token will not be renewed if it is expired.
```
tnorling:
```suggestion
- CacheOrRefreshToken - `acquireTokenSilent` will attempt to retrieve an access token from the cache. If the access token is expired or cannot be found, the refresh token will be used to acquire a new one. If the refresh token is expired, it will not be renewed and `acquireTokenSilent` will fail.
```

lib/msal-browser/docs/acquire-token.md
```diff
@@ -27,6 +27,14 @@ You can read more about using `acquireTokenSilent` [here](./token-lifetimes.md).
...
+- NetworkWithExistingRefreshTokenOnly - acquireTokenSilent will only go to network with the existing refresh token, and will not look in the cache. The refresh token will not be renewed if it is expired.
```
tnorling:
```suggestion
- RefreshTokenOnly - `acquireTokenSilent` will not attempt to retrieve access tokens from the cache and will instead attempt to exchange the cached refresh token for a new access token. If the refresh token is expired, it will not be renewed and `acquireTokenSilent will fail.
```

lib/msal-browser/docs/acquire-token.md
```diff
@@ -27,6 +27,14 @@ You can read more about using `acquireTokenSilent` [here](./token-lifetimes.md).
...
+- NetworkWithRefreshToken - acquireTokenSilent will only go to network with the existing refresh token, and will not look in the cache. The refresh token not be renewed if it is expired. This is existing "`forceRefresh: true`" behavior.
```
tnorling:
```suggestion
- ForceRefresh - `acquireTokenSilent` will not attempt to retrieve access tokens from the cache and will instead attempt to exchange the cached refresh token for a new access token. if the refresh token is expired `acquireTokenSilent` will attempt to acquire new access and refresh tokens. (This is equivalent to setting `forceRefresh: true` on the request)
```

lib/msal-browser/docs/acquire-token.md
```diff
@@ -27,6 +27,14 @@ You can read more about using `acquireTokenSilent` [here](./token-lifetimes.md).
...
+- NetworkOnly - acquireTokenSilent will only go to network, and not look in the cache. It will not use the existing refresh token, regardless of whether it is expired or not.
```
tnorling: IMO this option can be left off as this is the `ssoSilent` behavior

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -638,6 +646,12 @@ export abstract class ClientApplication {
                             throw error;
                         });
                 }
+
+                if (request.silentTokenRetrievalStrategy === SilentTokenRetrievalStrategy.NetworkWithExistingRefreshTokenOnly ||
```
tnorling: Might help readability if you write helper functions that handle this logic for us. i.e. `shouldCheckCache`, `shouldExchangeRefreshToken` (or `shouldRenewAccessToken`), `shouldRenewRefreshToken`

lib/msal-common/src/utils/Constants.ts
```diff
@@ -401,3 +401,38 @@ export enum JsonTypes {
...
     Jwk = "JWK"
 }
+
+export enum SilentTokenRetrievalStrategy {
+    /*
+     * acquireTokenSilent will first look in the cache before going to the network with the existing refresh token.
+     * The refresh token will be renewed if it is expired.
+     * (existing default behavior)
+     */
+    CacheAndNetwork = 0, // 0 is falsy, is equivalent to not passing in a SilentTokenRetrievalStrategy
```
tnorling: See my comments above regarding different name suggestions for these

lib/msal-browser/docs/acquire-token.md
jasonnutter: True, but I'd like to keep the semantics separate and not have apps call ssosilent when they could call acquiretokensilent.

lib/msal-browser/docs/acquire-token.md
jasonnutter: Hmm don't like this as much, as it isn't consistent with the others, and I don't really like "force refresh"

tnorling: `RefreshTokenOrRenewAll`? Which makes the ssoSilent-like behavior `RenewAll`?

lib/msal-browser/docs/acquire-token.md
tnorling: That's fair. FWIW I do think this is the least likely option to be used and if we keep it we should discourage its use as much as possible.

tnorling:
Telemetry should probably also track when these options are used

lib/msal-common/src/utils/Constants.ts
```diff
@@ -401,3 +401,38 @@ export enum JsonTypes {
...
+    /*
+     * acquireTokenSilent will only look in the cache, and will not go to the network with the existing refresh token.
+     * The refresh token will not be renewed if it is expired.
+     */
+    CacheOnly = 1,
+    /*
+     * acquireTokenSilent will first look in the cache before going to the network with the existing refresh token.
+     * The refresh token will not be renewed if it is expired.
```
bgavrilMS: AcquireTokenSilent won't be able to renew the RT if is expired. interaction is required.

lib/msal-common/src/utils/Constants.ts
```diff
@@ -401,3 +401,38 @@ export enum JsonTypes {
...
+export enum SilentTokenRetrievalStrategy {
+    /*
+     * acquireTokenSilent will first look in the cache before going to the network with the existing refresh token.
+     * The refresh token will be renewed if it is expired.
+     * (existing default behavior)
+     */
+    CacheAndNetwork = 0, // 0 is falsy, is equivalent to not passing in a SilentTokenRetrievalStrategy
+    /*
+     * acquireTokenSilent will only look in the cache, and will not go to the network with the existing refresh token.
+     * The refresh token will not be renewed if it is expired.
```
bgavrilMS: For MSAL libraries, RT expiration and other details are opaque. It is an implementation detail of AAD. I don't think we should document anything related expiration of RT.

lib/msal-common/src/utils/Constants.ts
```diff
@@ -401,3 +401,38 @@ export enum JsonTypes {
...
+     */
+    CacheOnly = 1,
```
bgavrilMS: Are there example scenarios for each of these options? I know that there are some scenarios where a developer might want to not use AT in the cache, and so other MSAL libraries have a `bool forceRefresh`, but I can't see when these options would be used.

lib/msal-common/src/utils/Constants.ts
```diff
@@ -401,3 +401,38 @@ export enum JsonTypes {
...
+     */
+    NetworkWithExistingRefreshTokenOnly = 3,
+    /*
+     * acquireTokenSilent will only go to network with the existing refresh token, and will not look in the cache.
+     * The refresh token not be renewed if it is expired.
+     * (existing forceRefresh=true behavior)
+     */
+    NetworkWithRefreshToken = 4,
+    /*
+     * acquireTokenSilent will only go to network, and not look in the cache.
```
bgavrilMS: AcquireTokenSilent is an API name that maps to the `refresh_token` flow. If you don't have a refresh token, what will this do?

lib/msal-common/src/utils/Constants.ts
jasonnutter: On the web, we can attempt to silently renew the rt using a hidden iframe and prompt=none (succeeds if the user still has an active session and 3p cookies aren't blocked).

lib/msal-common/src/utils/Constants.ts
tnorling: We attempt to use the RT and if the server tells us it is expired or invalid we attempt to silently renew it using a hidden iframe. Though this brings up an interesting perspective which is that these options are only relevant to msal-browser's implementation of `acquireTokenSilent` - perhaps this enum should live in msal-browser rather than common

lib/msal-common/src/utils/Constants.ts
tnorling: Refresh token flow is only one of 3 parts that `acquireTokenSilent` performs. 1. Cache 2. refresh_token flow 3. Hidden iframe renewal of all tokens

lib/msal-common/src/utils/Constants.ts
jasonnutter: See the linked github issues for examples of applications wanting to have more control over the caching behavior.

jasonnutter: (those don't demonstrate all of the scenarios, but figured for completeness we would enable applications to use any of there strategies)

bgavrilMS: I did read the associated GH issue and raised some concerns there as well - the scenario described by the user can be handled differently in my opinion.

These extensibility / configuration points increase the maintainability effort in the long run. For example, you'll need to document that these options won't work with WAM - this probably seems obvious to you and me, but not so much to our customers. The ROI seems poor if there aren't user scenarios or if the scenario is obscure, which I think it is. 

In the end, I'm only involved, and you are [committed](https://www.visual-paradigm.com/scrum/scrum-pig-and-chicken/) ... but just thought I'd highlight that it took me a bit to find the user request and that maybe it should have been challenged more.

jasonnutter: Those are fair points, however, if we don't offer this, applications who truly need this will impement hacky workarounds (i.e. reading directly from the cache or manually caching tokens themselves) or not use our library at all. I'm personally an advocate for providing sensible defaults, but also allowing applications to get finer grained control if desired ("make the easy things easy and the hard things possible").

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -187,7 +192,12 @@ export class PublicClientApplication extends ClientApplication implements IPubli
             this.logger.verbose("acquireTokenSilent - attempting to acquire token from web flow");
             const silentCacheClient = this.createSilentCacheClient(request.correlationId);
             const silentRequest = await silentCacheClient.initializeSilentRequest(request, account);
-            result = silentCacheClient.acquireToken(silentRequest).catch(async () => {
+            result = silentCacheClient.acquireToken(silentRequest).catch(async (error) => {
+                if (request.silentTokenRetrievalStrategy === SilentTokenRetrievalStrategy.CacheOnly) {
+                    this.logger.error("SilentTokenRetrievalStrategy is set to CacheOnly, not attempting to renew token silently.");
+                    throw error;
```
peterzenz: Do we want to throw the raw cache exception or map this to something more friendly?  I'm assuming right now if we throw a cache error today we'd catch that (it would never get returned) and if the subsequent network request failed with a network or server error we'd throw that back.

lib/msal-browser/src/interaction_client/SilentRefreshClient.ts
```diff
@@ -14,6 +14,11 @@ export class SilentRefreshClient extends StandardInteractionClient {
      * @param request
      */
     async acquireToken(request: CommonSilentFlowRequest): Promise<AuthenticationResult> {
+        if (request.silentTokenRetrievalStrategy === SilentTokenRetrievalStrategy.NetworkOnly) {
+            this.logger.error("SilentTokenRetrievalStrategy is set to NetworkOnly, not attempting to acquire refresh token");
+            throw new ServerError(BrowserConstants.INVALID_GRANT_ERROR);
```
peterzenz: Why would this be an existing server error and not something new?

lib/msal-browser/test/app/PublicClientApplication.spec.ts
```diff
@@ -2045,6 +2045,104 @@ describe("PublicClientApplication.ts Class Unit Tests", () => {
             expect(silentIframeSpy.called).toBe(false);
         });
 
+        it("Does't call SilentIframeClient.acquireToken if cache lookup throws and refresh token is expired when SilentTokenRetrievalStrategy is set to NetworkWithExistingRefreshTokenOnly", async () => {
```
peterzenz: nit:  Doesn't
Also copied to line 2081 below.

lib/msal-common/src/utils/Constants.ts
peterzenz: Fair statements on both sides here.  There are a few other small differences between the WAM and JS implementations we'll have to sort out as well.  Once we have to enable the JS/WAM flow by default I expect we'll have to address or disable all of these.   @jasonnutter  do you feel strongly about fixing this?  The finer grained control get's really painful if a user had to understand differences in the JS/WAM behavior or JS w/o WAM behavior.

peterzenz (dismissed):
Would like to settle point with Jason and Bogdan before committing this one.

lib/msal-browser/src/app/PublicClientApplication.ts
jasonnutter: I think we would either get [this error](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/eb612862f60dd6b9d28e6bec8e08989890dbf441/lib/msal-common/src/error/ClientAuthError.ts#L161-L164) (the error message for which should get updated to reflect this new scenario @Robbie-Microsoft), or an interaction_required/login_required error from the server, which I think is fine.

lib/msal-browser/src/interaction_client/SilentRefreshClient.ts
jasonnutter: Because there is downstream logic that looks for an invalid grant error already.

jasonnutter: (which could be updated to look for a second type of error, if we wanted.)

lib/msal-common/src/utils/Constants.ts
jasonnutter: Yes, I believe we need to ship it as we have partners asking for it. I think it's fine if the enums don't map perfectly for JS/WAM flow, we can figure out sensible fallbacks for scenarios that don't perfectly map.

lib/msal-browser/docs/acquire-token.md
Robbie-Microsoft: @jasonnutter thoughts about Thomas's comment?

lib/msal-browser/src/app/ClientApplication.ts
Robbie-Microsoft: I'm thinking that adding a comment here will be sufficient. This is the only occurrence of this specific check.

lib/msal-browser/docs/acquire-token.md
jasonnutter: And then `NetworkWithExistingRefreshTokenOnly` -> `RefreshTokenOnly`? Would like some consistency between the network only options.

lib/msal-common/src/utils/Constants.ts
Robbie-Microsoft: The enum IS being in msal-common though. In lib/msal-common/src/client/SilentFlowClient.ts

lib/msal-browser/docs/acquire-token.md
sameerag: In that case shouldn't `CacheOrExistingRefreshToken` be renamed `CacheOrNetworkWithExistingRefreshTokenOnly`?

lib/msal-browser/docs/acquire-token.md
sameerag: Can we refactor the code to call `ssoSilent` when this option is selected for consistency?

lib/msal-browser/docs/acquire-token.md
sameerag: My two cents:  Naming consistency across, `cache`, `networkWithExistingRT` and `networkWithRT` needed. Seems inconsistent and confusing as it stands. It is okay to add comments relating to `default`, `forceRefresh` or `ssoSilent` in the description, however I prefer the naming convention of the options to be uniform. And @Robbie-Microsoft may be a table in docs will help.

Mapping them to existing APIs and what is new makes it more readable IMO for usage docs.

lib/msal-browser/src/interaction_client/SilentRefreshClient.ts
sameerag: The protocol says interactive auth for both invalid_grant and interaction_required errors should be a retry, reference [here](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow). 

I recently had a conversation with someone from CCS concluding:

1. **For requests with authz code:**
a. `interaction_required` will return `InteractionRequiredAuthError` 
b. `invalid_grant` will return errors to client

2. **For requests with RT:**
a. `interaction_required `will return `InteractionRequiredAuthError` 
b.  `invalid_grant` will silently retry /authorize, and will return errors to client in case of errors

There are two things we may need to address (@Robbie-Microsoft  please check this):
- Are we handling 1.b correctly?
- With these new options, should we introduce MSAL JS specific errors that do not fall under the above protocol sub cases due to user preference (not wanting to use RT etc)?

lib/msal-common/src/utils/Constants.ts
sameerag: Catching up. IMHO, we need to advocate these as `advanced options` and have defaults defined in our docs - which could be more uniform across WAM vs non-WAM flows. Otherwise, the fine grained control may be more of a maintenance risk than help in my experience.

lib/msal-browser/docs/acquire-token.md
jasonnutter: It already is (effectively, calls the same internal apis).

jasonnutter: I don't think we should call ssosilent directly because ssosilent is meant as a public api and needs different semantics than an internal api. Also, it would make metrics and telemetry confusing.

lib/msal-browser/docs/acquire-token.md
tnorling: @jasonnutter I'm good with that, I think that's what I had suggested above. @Sameerag `CacheOrExistingRefreshToken` became `CacheOrRefreshToken`

lib/msal-common/src/utils/Constants.ts
tnorling: The new logic you've added there can be moved up in the stack to be executed in msal-browser. If we don't envision this enum being used in msal-node (which I don't) then that is likely the right move.

lib/msal-common/src/utils/Constants.ts
peterzenz: Having the differences in `advanced options` doesn't help when those might behave differently depending on the client configuration which is out of the app developers control.  I've filed a consistency item with WAM so we can support this there as well.

peterzenz:
@Robbie-Microsoft please commit this one you have the requisite signoffs.  Thanks.

lib/msal-browser/docs/acquire-token.md
```diff
@@ -27,6 +27,14 @@ You can read more about using `acquireTokenSilent` [here](./token-lifetimes.md).
...
+- CacheOnly - `acquireTokenSilent` will only look for access tokens in the cache. It will not attempt to renew access or refresh tokens.
+- CacheOrRefreshToken - `acquireTokenSilent` will attempt to retrieve an access token from the cache. If the access token is expired or cannot be found, the refresh token will be used to acquire a new one. If the refresh token is expired, it will not be renewed and `acquireTokenSilent` will fail.
+- RefreshTokenOnly - `acquireTokenSilent` will not attempt to retrieve access tokens from the cache and will instead attempt to exchange the cached refresh token for a new access token. If the refresh token is expired, it will not be renewed and `acquireTokenSilent will fail.
+- NetworkWithRefreshToken - acquireTokenSilent will only go to network with the existing refresh token, and will not look in the cache. The refresh token not be renewed if it is expired. This is existing "`forceRefresh: true`" behavior.
+- NetworkOnly - `acquireTokenSilent` will only go to network and not look in the cache. It will not use the existing refresh token, regardless of whether it is expired or not.
```
tnorling: Thought we changed this to `RefreshTokenOrRenewAll` and `RenewAll`? Don't particularly like the word "network" in this context given all options other than CacheOnly use the network.

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -582,24 +582,23 @@ export abstract class ClientApplication {
...
+     * Attempt to acquire an access token via a refresh token
+     * @param request CommonSilentFlowRequest
+     * @param atbrtMeasurement measurement API
+     * @returns A promise that, when resolved, returns the access token
      */
-    protected async acquireTokenByRefreshToken(request: CommonSilentFlowRequest): Promise<AuthenticationResult> {
+    protected async acquireTokenByRefreshToken(
+        request: CommonSilentFlowRequest,
+        // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/no-explicit-any
+        atbrtMeasurement: any
```
tnorling: We should not be passing performance measurements down through APIs

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -610,39 +609,64 @@ export abstract class ClientApplication {
...
+     * @param request CommonSilentFlowRequest
+     * @param atbrtMeasurement measurement API
+     * @param networkOnly a boolean indicating if this function is being called as a result of an error that may
+     * have occurred when trying to acquire an access token via the refresh token, or if the SilentTokenRetrievalStrategy
+     * is set to NetworkOnly
+     * @returns A promise that, when resolved, returns the access token
+     */
+    protected async acquireTokenBySilentIframe(
+        // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/no-explicit-any
+        error: any,
```
tnorling: Please do error detection in the caller rather than passing the error around

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -610,39 +609,64 @@ export abstract class ClientApplication {
...
+        request: CommonSilentFlowRequest,
+        // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/no-explicit-any
+        atbrtMeasurement: any,
```
tnorling: Same comment here regarding performance measurements

lib/msal-common/src/telemetry/performance/PerformanceEvent.ts
```diff
@@ -291,5 +291,12 @@ export type PerformanceEvent = {
...
      */
-    isNativeBroker?: boolean
+    isNativeBroker?: boolean,
+
+    /**
+     * The Silent Token Retrieval Strategy
+     *
+     * @type {?(number | undefined)}
+     */
+    silentTokenRetrievalStrategy?: number | undefined
```
jasonnutter: You should be able to specify the enum type here instead of number.

lib/msal-browser/test/app/PublicClientApplication.spec.ts
```diff
@@ -2045,6 +2045,106 @@ describe("PublicClientApplication.ts Class Unit Tests", () => {
             expect(silentIframeSpy.called).toBe(false);
         });
 
+        // it("Doesn't call SilentIframeClient.acquireToken if cache lookup throws and refresh token is expired when SilentTokenRetrievalStrategy is set to NetworkWithExistingRefreshTokenOnly", async () => {
```
jasonnutter: Please delete or uncomment

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -187,9 +193,64 @@ export class PublicClientApplication extends ClientApplication implements IPubli
             this.logger.verbose("acquireTokenSilent - attempting to acquire token from web flow");
             const silentCacheClient = this.createSilentCacheClient(request.correlationId);
             const silentRequest = await silentCacheClient.initializeSilentRequest(request, account);
-            result = silentCacheClient.acquireToken(silentRequest).catch(async () => {
-                return this.acquireTokenByRefreshToken(silentRequest);
-            });
+
+            const acquireTokenByRefreshToken = (silentRequest: CommonSilentFlowRequest, silentTokenRetrievalStrategy: SilentTokenRetrievalStrategy | undefined): Promise<AuthenticationResult> => {
```
jasonnutter: This refactor is more difficult to read and understand.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -187,9 +193,64 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
+            /*
+             * do not look in the cache (just go to the network) if the silentTokenRetrievalStrategy
+             * is set to one of the network-only options
+             */
+            if ((request.silentTokenRetrievalStrategy === SilentTokenRetrievalStrategy.RefreshTokenOnly)
+                || (request.silentTokenRetrievalStrategy === SilentTokenRetrievalStrategy.NetworkWithRefreshToken)
+                || (request.silentTokenRetrievalStrategy === SilentTokenRetrievalStrategy.NetworkOnly)) {
+                this.logger.info("SilentCacheClient:acquireToken - Skipping cache because SilentTokenRetrievalStrategy is set to one of the network-only options.");
+                
+                result = acquireTokenByRefreshToken(silentRequest, request.silentTokenRetrievalStrategy);
```
jasonnutter: Why are both parameters needed? `silentRequest` should have `silentTokenRetrievalStrategy` as a property.

lib/msal-browser/src/app/PublicClientApplication.ts
jasonnutter: Same comment here about parameters.

lib/msal-browser/src/app/PublicClientApplication.ts
jasonnutter: Ah, `silentRequest` is `CommonSilentFlowRequest`, which doesn't have `silentTokenRetrievalStrategy`. Will think about how to best handle this.

Robbie-Microsoft: Yep, that's also why I can't use the enum in PerformanceEvent.ts like you suggested above.

lib/msal-browser/docs/acquire-token.md
Robbie-Microsoft: I chatted with @jasonnutter, and we decided to stick with what it currently is, which is what we came up with in the design review.

lib/msal-browser/src/utils/BrowserConstants.ts
```diff
@@ -178,3 +178,38 @@ export enum WrapperSKU {
...
+     */
+    CacheOrRefreshToken = 2,
+    /*
+     * acquireTokenSilent will only go to network with the existing refresh token, and will not look in the cache.
+     * The refresh token will not be renewed if it is expired.
+     */
+    RefreshTokenOnly = 3,
+    /*
+     * acquireTokenSilent will only go to network with the existing refresh token, and will not look in the cache.
+     * The refresh token not be renewed if it is expired.
```
jasonnutter:
```suggestion
     * The refresh token will be renewed if it is expired.
```

lib/msal-browser/src/utils/BrowserConstants.ts
```diff
@@ -178,3 +178,38 @@ export enum WrapperSKU {
...
+     * acquireTokenSilent will first look in the cache before going to the network with the existing refresh token.
+     * The refresh token will be renewed if it is expired.
+     * (existing default behavior)
+     */
+    Default = 0, // 0 is falsy, is equivalent to not passing in a SilentTokenRetrievalStrategy
+    /*
+     * acquireTokenSilent will only look in the cache, and will not go to the network with the existing refresh token.
+     * The refresh token will not be renewed if it is expired.
+     */
+    CacheOnly = 1,
```
jasonnutter:
```suggestion
    AccessToken = 1,
```

lib/msal-browser/src/utils/BrowserConstants.ts
```diff
@@ -178,3 +178,38 @@ export enum WrapperSKU {
...
+    /*
+     * acquireTokenSilent will first look in the cache before going to the network with the existing refresh token.
+     * The refresh token will not be renewed if it is expired.
+     */
+    CacheOrRefreshToken = 2,
```
jasonnutter:
```suggestion
    AccessTokenAndRefreshToken = 2,
```

lib/msal-browser/src/utils/BrowserConstants.ts
```diff
@@ -178,3 +178,38 @@ export enum WrapperSKU {
...
+    /*
+     * acquireTokenSilent will only go to network with the existing refresh token, and will not look in the cache.
+     * The refresh token will not be renewed if it is expired.
+     */
+    RefreshTokenOnly = 3,
```
jasonnutter:
```suggestion
    RefreshToken = 3,
```

lib/msal-browser/src/utils/BrowserConstants.ts
```diff
@@ -178,3 +178,38 @@ export enum WrapperSKU {
...
+    /*
+     * acquireTokenSilent will only go to network with the existing refresh token, and will not look in the cache.
+     * The refresh token not be renewed if it is expired.
+     * (existing forceRefresh=true behavior)
+     */
+    NetworkWithRefreshToken = 4,
```
jasonnutter:
```suggestion
    RefreshTokenAndNetwork = 4,
```

lib/msal-browser/src/utils/BrowserConstants.ts
```diff
@@ -178,3 +178,38 @@ export enum WrapperSKU {
...
+    /*
+     * acquireTokenSilent will only go to network, and not look in the cache.
+     * It will not use the existing refresh token, regardless of whether it is expired or not.
+     */
+    NetworkOnly = 5,
```
jasonnutter:
```suggestion
    Skip = 5,
```

lib/msal-browser/src/utils/BrowserConstants.ts
```diff
@@ -178,3 +178,38 @@ export enum WrapperSKU {
 export const DB_NAME = "msal.db";
 export const DB_VERSION = 1;
 export const DB_TABLE_NAME = `${DB_NAME}.keys`;
+
+export enum SilentTokenRetrievalStrategy {
```
jasonnutter:
```suggestion
export enum CacheLookupStrategy {
```

lib/msal-browser/src/request/SilentRequest.ts
```diff
@@ -26,4 +28,5 @@ export type SilentRequest = Omit<CommonSilentFlowRequest, "authority"|"correlati
     account?: AccountInfo;
     correlationId?: string;
     forceRefresh?: boolean;
+    silentTokenRetrievalStrategy?: SilentTokenRetrievalStrategy;
```
jasonnutter:
```suggestion
    cacheLookupStrategy?: CacheLookupStrategy;
```

lib/msal-common/src/telemetry/performance/PerformanceEvent.ts
jasonnutter:
```suggestion
    cacheLookStrategy?: number | undefined
```

lib/msal-browser/src/utils/BrowserConstants.ts
Robbie-Microsoft: Skip

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -583,67 +583,59 @@ export abstract class ClientApplication {
...
+        commonRequest: CommonSilentFlowRequest,
+        silentRequest: SilentRequest
+    ): Promise<AuthenticationResult> {
+        switch(silentRequest.cacheLookupPolicy) {
+            case CacheLookupPolicy.Default:
+            case CacheLookupPolicy.AccessToken:
+            case CacheLookupPolicy.AccessTokenAndRefreshToken:
+                return silentCacheClient.acquireToken(commonRequest);
+            default:
+                throw ClientAuthError.createRefreshRequiredError();
```
Robbie-Microsoft, line 603: @jasonnutter @tnorling Do you guys think we should add another ClientAuthError here that says something like "Invalid Cache Lookup Policy".

Same question for line 625.

lib/msal-browser/docs/acquire-token.md
```diff
@@ -27,6 +27,14 @@ You can read more about using `acquireTokenSilent` [here](./token-lifetimes.md).
 
 You must pass a request object to the acquireToken APIs. This object allows you to use different parameters in the request. See [here](./request-response-object.md) for more information on the request object parameters. Scopes are required for all acquireToken calls.
 
+A Cache Lookup Policy can be optionally provided to the request. The Cache Lookup Policies are:
+- Default - `acquireTokenSilent` will attempt to retrieve an access token from the cache. If the access token is expired or cannot be found the refresh token will be used to acquire a new one. Finally, if the refresh token is expired `acquireTokenSilent` will attempt to acquire new access and refresh tokens.
```
jasonnutter: Please show entire enum value and wrap in backticks.

```suggestion
- `CacheLookupPolicy.Default` - `acquireTokenSilent` will attempt to retrieve an access token from the cache. If the access token is expired or cannot be found the refresh token will be used to acquire a new one. Finally, if the refresh token is expired `acquireTokenSilent` will attempt to acquire new access and refresh tokens.
```

jasonnutter: Also, please show a code snippet demonstrating basic usage, thanks!

lib/msal-browser/src/app/ClientApplication.ts
jasonnutter: No, don't think so. Enumeration pretty clearly documents expected possible values.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -187,10 +194,40 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
         } else {
             this.logger.verbose("acquireTokenSilent - attempting to acquire token from web flow");
+
             const silentCacheClient = this.createSilentCacheClient(request.correlationId);
             const silentRequest = await silentCacheClient.initializeSilentRequest(request, account);
-            result = silentCacheClient.acquireToken(silentRequest).catch(async () => {
-                return this.acquireTokenByRefreshToken(silentRequest);
+
+            // set the request's CacheLookupPolicy to Default if it was not optionally passed in
+            request.cacheLookupPolicy = (request.cacheLookupPolicy === undefined) ? CacheLookupPolicy.Default : request.cacheLookupPolicy;
```
jasonnutter: This is typically done in `initializeSilentRequest`: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/src/interaction_client/SilentCacheClient.ts#L70

jasonnutter: Also, it should treat any falsey value (null, undefined) as being the default:

```suggestion
            request.cacheLookupPolicy = request.cacheLookupPolicy || CacheLookupPolicy.Default;
```

jasonnutter:
A couple more comments, thanks for addressing feedback!

lib/msal-browser/src/app/PublicClientApplication.ts
Robbie-Microsoft: @jasonnutter `initializeSilentRequest` returns a `Promise<CommonSilentFlowRequest>` (msal-common), not a `Promise<SilentRequest>` (msal-browser). Therefore, I would need to copy/paste the suggested change above to line 66 in SilentCacheClient.ts. If we did that, then we could refactor the code in msal-browser and would no longer need to pass around the silentRequest from PublicClientApplication to ClientApplication. Thoughts?

lib/msal-browser/docs/acquire-token.md
```diff
@@ -27,6 +27,14 @@ You can read more about using `acquireTokenSilent` [here](./token-lifetimes.md).
...
+- AccessToken - `acquireTokenSilent` will only look for access tokens in the cache. It will not attempt to renew access or refresh tokens.
+- AccessTokenAndRefreshToken - `acquireTokenSilent` will attempt to retrieve an access token from the cache. If the access token is expired or cannot be found, the refresh token will be used to acquire a new one. If the refresh token is expired, it will not be renewed and `acquireTokenSilent` will fail.
+- RefreshToken - `acquireTokenSilent` will not attempt to retrieve access tokens from the cache and will instead attempt to exchange the cached refresh token for a new access token. If the refresh token is expired, it will not be renewed and `acquireTokenSilent will fail.
+- RefreshTokenAndNetwork - acquireTokenSilent will only go to network with the existing refresh token, and will not look in the cache. The refresh token will be renewed if it is expired. This is existing "`forceRefresh: true`" behavior.
```
tnorling:
```suggestion
- RefreshTokenAndNetwork - acquireTokenSilent will not look in the cache for the access token. It will directly go to network with the cached refresh token. If the refresh token is expired an attempt will be made to renew it. This is equivalent to setting `forceRefresh: true`.
```

lib/msal-browser/docs/acquire-token.md
```diff
@@ -27,6 +27,14 @@ You can read more about using `acquireTokenSilent` [here](./token-lifetimes.md).
...
+- Skip - `acquireTokenSilent` will only go to network and not look in the cache. It will not use the existing refresh token, regardless of whether it is expired or not.
```
tnorling:
```suggestion
- Skip - `acquireTokenSilent` will attempt to renew both access and refresh tokens. It will not look in the cache. This will always fail if 3rd party cookies are blocked by the browser.
```

lib/msal-browser/src/app/PublicClientApplication.ts
tnorling: Given that, I think this is fine here but please create a new request object rather than modifying the object passed in.

lib/msal-browser/src/utils/BrowserConstants.ts
```diff
@@ -178,3 +178,38 @@ export enum WrapperSKU {
...
+    AccessTokenAndRefreshToken = 2,
+    /*
+     * acquireTokenSilent will only go to network with the existing refresh token, and will not look in the cache.
+     * The refresh token will not be renewed if it is expired.
+     */
+    RefreshToken = 3,
+    /*
+     * acquireTokenSilent will only go to network with the existing refresh token, and will not look in the cache.
+     * The refresh token will be renewed if it is expired.
+     * (existing forceRefresh=true behavior)
```
tnorling: Better yet if you can copy the descriptions from the .md file verbatim.
```suggestion
     * equivalent to forceRefresh=true behavior
```

tnorling (approved):
Looking good with a few more suggestions. Thanks for putting this together!

lib/msal-browser/docs/acquire-token.md
```diff
@@ -27,11 +27,20 @@ You can read more about using `acquireTokenSilent` [here](./token-lifetimes.md).
...
+- `CacheLookupPolicy.RefreshTokenAndNetwork` - `acquireTokenSilent` will not look in the cache for the access token. It will go directly to network with the cached refresh token. If the refresh token is expired an attempt will be made to renew it. This is equivalent to setting `forceRefresh: true`.
+- `CacheLookupPolicy.Skip` - `acquireTokenSilent` will attempt to renew both access and refresh tokens. It will not look in the cache. This will always fail if 3rd party cookies are blocked by the browser.
+
 - Popup
 
 ```javascript
 var request = {
-    scopes: ["Mail.Read"]
+    scopes: ["Mail.Read"],
+    cacheLookupPolicy: 0 // Default
```
hectormmg: I suggest using the enum in the doc since we want to promote readable code.
```suggestion
    cacheLookupPolicy: CacheLookupPolicy.Default
```

lib/msal-browser/docs/acquire-token.md
```diff
@@ -50,7 +59,8 @@ msalInstance.acquireTokenSilent(request).then(tokenResponse => {
...
```
hectormmg: Same here.

hectormmg (approved):
Some suggested changes, looks good to me otherwise!

lib/msal-browser/docs/acquire-token.md
```diff
@@ -27,11 +27,20 @@ You can read more about using `acquireTokenSilent` [here](./token-lifetimes.md).
 
 You must pass a request object to the acquireToken APIs. This object allows you to use different parameters in the request. See [here](./request-response-object.md) for more information on the request object parameters. Scopes are required for all acquireToken calls.
 
+A Cache Lookup Policy can be optionally provided to the request. The Cache Lookup Policies are:
+- `CacheLookupPolicy.Default` - `acquireTokenSilent` will attempt to retrieve an access token from the cache. If the access token is expired or cannot be found the refresh token will be used to acquire a new one. Finally, if the refresh token is expired `acquireTokenSilent` will attempt to acquire new access and refresh tokens.
```
jasonnutter:
```suggestion
- `CacheLookupPolicy.Default` - `acquireTokenSilent` will attempt to retrieve an access token from the cache. If the access token is expired or cannot be found the refresh token will be used to acquire a new one. Finally, if the refresh token is expired, `acquireTokenSilent` will attempt to silently acquire a new access token, id token, and refresh token.
```

jasonnutter (approved):

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -187,10 +194,43 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
-            result = silentCacheClient.acquireToken(silentRequest).catch(async () => {
-                return this.acquireTokenByRefreshToken(silentRequest);
+            
+            const requestWithCLP = {
+                ...request,
+                // set the request's CacheLookupPolicy to Default if it was not optionally passed in
+                cacheLookupPolicy: request.cacheLookupPolicy || CacheLookupPolicy.Default
+            };
+
+            result = this.acquireTokenFromCache(silentCacheClient, silentRequest, requestWithCLP).catch((cacheError: AuthError) => {
```
bmahall, line 207: @Robbie-Microsoft This looks like a breaking change for ExperimentalPublicClientApplication in msal-1p. Are we working on corresponding changes in 1p?
@jasonnutter

Issue #4953, opened 2022-06-30
`useMsalAuthentication` `acquireTokenSilent` logic fails if using `prompt` other than `none`


ckeboss:
### Core Library

MSAL.js v2 (@azure/msal-browser)

### Core Library Version

2.26.0

### Wrapper Library

MSAL React (@azure/msal-react)

### Wrapper Library Version

1.4.2

### Description

This change (https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/4280) introduced logic to the `useMsalAuthentication` hook that tries to call `acquireTokenSilent`.

It takes the `authenticationRequest` passed to the hook, with some minor logic surrounding scopes, and passed the resulting object to `acquireTokenSilent`.

acquireTokenSilent expects a SilentRequest (no prompt).

If a user is not passing in a prompt type of `none`, they will get an error. (https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/msal-browser-v2.26.0/lib/msal-browser/src/interaction_client/SilentIframeClient.ts#L45)

Relevant code here: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/60c4aafa5/lib/msal-react/src/hooks/useMsalAuthentication.ts#L111

My initial thought here is to simply spread `prompt: none` into the argument for `acquireTokenSilent`.

### MSAL Configuration

```javascript
const config = {
  auth: {
    authority: `https://login.microsoftonline.com/${environment.msalTenant}`,
    clientId: environment.msalClientID
  },
  cache: {
    cacheLocation: "localStorage",
    storeAuthStateInCookie: true
  }
};
```

### Relevant Code Snippets

```javascript
const { error, login } = useMsalAuthentication(
    InteractionType.Redirect,
    {
      scopes,
      prompt: 'select_account'
    }
  )
```

### Reproduction Steps

1. Call `useMsalAuthentication` hook with `prompt` not equal to `none` while logged in to an account.

### Expected Behavior

https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/msal-browser-v2.26.0/lib/msal-browser/src/interaction_client/SilentIframeClient.ts#L45 error is not thrown.

### Identity Provider

Azure AD / MSA

### Browsers Affected (Select all that apply)

Chrome, Firefox, Edge, Safari, Internet Explorer

### Source

External (Customer)

ckeboss:
[...]

PR #5029, opened 2022-07-20
don't pass authenticationRequest prompt attribute to acquireTokenSilent

ckeboss:
Should address https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/4953


tnorling:
`prompt` isn't part of the public surface for silent calls, so I think this is more so a bug in msal-browser. It's probably cleaner, and the right thing to do, to just remove the error you're facing and hardcode it to use `prompt: none` under the hood (which it should already be doing)

tnorling:
Closing per my comment above. If this continues to be an issue please feel free to open a new PR or an issue so we can add this to our backlog. Thanks!

PR #5033, opened 2022-07-21
Add error handling in device-code

### LINKED ISSUE #4507: [msal-common] DeviceCodeClient not properly throwing if the server errors during polling
### Core Library

MSAL Node (@azure/msal-node)

### Core Library Version

latest

### Wrapper Library

Not Applicable

### Wrapper Library Version

not applicable

### Description

The DeviceCodeClient [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/6592652877f31405c8ed73a66f03eada90a78c0a/lib/msal-common/src/client/DeviceCodeClient.ts#L186-L200) does not consider scenarios in which the service crashes. The `executePostToTokenEndpoint` method of the base client only adds telemetry of service error cases.

### Error Message

No error message is shown.

### Msal Logs

_No response_

### MSAL Configuration

```javascript
{}
```

### Relevant Code Snippets

```javascript
//
```

### Reproduction Steps

This issue can be reproduced by passing a HTTP client pointing to a proxy, or mock the network requests to return an error during polling.

### Expected Behavior

If the authentication service fails to respond or crashes, or the authorization-pending string is missing, polling should stop and the error should be thrown.

### Identity Provider

Azure AD / MSA

None (Server)

### Regression

_No response_

### Source

Internal (Microsoft)

sameerag:
Add error handling for deviceCodeClient use cases where the server may crash/polling may have failed for unexpected reasons.

Addresses PR #4507


lib/msal-common/src/client/DeviceCodeClient.ts
```diff
@@ -196,17 +197,23 @@ export class DeviceCodeClient extends BaseClient {
...
                 thumbprint);
 
-            if (response.body && response.body.error === Constants.AUTHORIZATION_PENDING) {
+            if (response.body && response.body.error) {
                 // user authorization is pending. Sleep for polling interval and try again
-                this.logger.info(response.body.error_description || "Authorization pending. Continue polling.");
-  
-                await TimeUtils.delay(pollingIntervalMilli);
+                if(response.body.error === Constants.AUTHORIZATION_PENDING) {
+                    this.logger.info(response.body.error_description || "Authorization pending. Continue polling.");
```
tnorling: What's the scenario where the OR condition of this log message gets hit?

sameerag: I guess to accommodate different server end points we may support. AAD always sends one type of error afaik.

lib/msal-common/src/client/DeviceCodeClient.ts
```diff
@@ -196,17 +197,23 @@ export class DeviceCodeClient extends BaseClient {
...
+                    await TimeUtils.delay(pollingIntervalMilli);
+                } else {
+                    // for any other error, throw
+                    this.logger.info("Unexpected error in polling from the server");
+                    throw ServerError.createUnexpectedError(response.body.error);
```
tnorling: Can we throw post_request_failed instead of a generic "unexpected" error? Like we do in msal-browser? This isn't necessarily a server error, is it?

sameerag: Yes. Any thing that fails : 4xx, 5xx etc will come under this + could be a router issues etc. I can add the new error type.

lib/msal-common/src/client/DeviceCodeClient.ts
tnorling: Can we just log the right side then? Not sure I understand why we need both

sameerag: Yes, logger message can be a string. Updated.

peterzenz (approved):

lib/msal-common/src/client/DeviceCodeClient.ts
sameerag: Done.

lib/msal-common/src/client/DeviceCodeClient.ts
sameerag: Added a new error type.

tnorling (approved):

PR #5036, opened 2022-07-21
Exposed Raw ID Token on Account Info objects.

### LINKED ISSUE #3569: MSAL Angular v2: Logout with ID Token
### Core Library

@azure/msal-browser

### Core Library Version

2.14.1

### Wrapper Library

@azure/msal-angular

### Wrapper Library Version

2.0.0-beta.5

### Description

In Azure AD B2C, one can enable the option to require the ID Token for logouts. This option does not seem to be supported out of the box by the MSAL Angular (v2) wrapper.

Are there any plans to support this feature before coming out of beta? I probably could include the ID token myself, if I replicate the whole MSAL Angular Interceptor logic where the `acquireTokenSilent()` call is made, but looking at the code it looks a bit cumbersome. The [EndSessionRequest](https://azuread.github.io/microsoft-authentication-library-for-js/ref/modules/_azure_msal_browser.html#endsessionrequest) looks like it already supports the `idTokenHint`.

### MSAL Configuration

_No response_

### Relevant Code Snippets

```javascript
logout(): Observable<void> {
  return this.msal.logout({idTokenHint: "missing-token-here"});
}
```

### Identity Provider

Azure B2C Basic Policy

### Source

External (Customer)

### LINKED ISSUE #5053: idToken can't be retrieved from cache
### Core Library

### Core Library Version

2.21.1

### Wrapper Library

Not Applicable

### Wrapper Library Version

None

### Description

`acquireTokenSilent` method can be used only for accessToken.
It should also be available for idToken.

My app is using only idToken, accessToken is always empty.

Currently there is no way to get idToken from cache, with the same nice flow as it exist for accessToken.

`msal.acquireTokenSilent({ forceRefresh: false, etc })` will always trigger network call.

https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/master/lib/msal-common/src/client/SilentFlowClient.ts#L68

Is there any configuration where we can set which token is main, eg { authToken: 'idToken'; //  default 'accessToken' }.

Thanks.

### Error Message

_No response_

### Msal Logs

_No response_

### MSAL Configuration

```javascript
auth: {
    clientId: config.clientId,
    authority: config.authority,
    knownAuthorities: config.knownAuthorities,
    redirectUri: config.loginRedirectUrl,
    postLogoutRedirectUri: config.logoutRedirectUrl
  },
  cache: {
    cacheLocation: CONSTANTS.localStorage,
    storeAuthStateInCookie: false
  }
```

### Relevant Code Snippets

```javascript
It's clear from the source code that network call is always required, since `!cacheRecord.accessToken` is true:


       if (request.forceRefresh) {
            // Must refresh due to present force_refresh flag.
            this.serverTelemetryManager?.setCacheOutcome(CacheOutcome.FORCE_REFRESH);
        } else if (!cacheRecord.accessToken) {
            // Must refresh due to non-existent access_token.
            this.serverTelemetryManager?.setCacheOutcome(CacheOutcome.NO_CACHED_ACCESS_TOKEN);
        }
```
```


### Reproduction Steps

1. check Relevant Code Snippets

### Expected Behavior

acquireTokenSilent should work for idToken if accessToken is empty.
There should be some configuration to mark main authToken.

### Identity Provider

Azure B2C Custom Policy

### Regression

_No response_

### Source

External (Customer)

### PR DESCRIPTION
**Pull request by Robbie-Microsoft**
Updated all relevant unit tests.

### CODE REVIEW

File: lib/msal-common/src/account/AccountInfo.ts [hunk 1]
```diff
@@ -23,6 +24,7 @@ export type AccountInfo = {
     username: string;
     localAccountId: string;
     name?: string;
+    idToken?: string;
```
sameerag, line 27: Why is this param optional in `AccountInfo` and mandated in `AccountEntity`?

Robbie-Microsoft: @sameerag I made a mistake - I believe it should be mandated in both. Do you agree?

sameerag (approved):

jasonnutter (dismissed):

jasonnutter (commented):
Looks good, one question: the `setAccount` method in `BrowserCacheManager` will save the entire account object into the cache, and with this change that would mean the entire id token would be saved into the cache twice (once for teh id token itself, and once for the account). It might be good to not do that.

jasonnutter:
Referring to here: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/de5ef1c6219d43dce37aea74fd0fcf48f67c471e/lib/msal-browser/src/cache/BrowserCacheManager.ts#L189

It might be good to do something like this:

```typescript
setAccount(account: AccountEntity): void {
        this.logger.trace("BrowserCacheManager.setAccount called");
        const key = account.generateAccountKey();
       // Don't save idtoken back into cache twice
        const {
            // @ts-ignore
            // eslint-disable-next-line @typescript-eslint/no-unused-vars
            idToken,
            ...accountWithoutIdToken
        } = account;
        this.setItem(key, JSON.stringify(accountWithoutIdToken));
    }

```

tnorling:
> Referring to here:
> 
> https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/de5ef1c6219d43dce37aea74fd0fcf48f67c471e/lib/msal-browser/src/cache/BrowserCacheManager.ts#L189
> 
> It might be good to do something like this:
> 
> ```ts
> setAccount(account: AccountEntity): void {
>         this.logger.trace("BrowserCacheManager.setAccount called");
>         const key = account.generateAccountKey();
>        // Don't save idtoken back into cache twice
>         const {
>             // @ts-ignore
>             // eslint-disable-next-line @typescript-eslint/no-unused-vars
>             idToken,
>             ...accountWithoutIdToken
>         } = account;
>         this.setItem(key, JSON.stringify(accountWithoutIdToken));
>     }
> ```

Alternatively, could we just include idToken on `AccountInfo` but not `AccountEntity`? Since the Entity is the part that's cached and the Info is derived/returned

jasonnutter:
> Alternatively, could we just include idToken on `AccountInfo` but not `AccountEntity`? Since the Entity is the part that's cached and the Info is derived/returned

That would work.

Robbie-Microsoft:
> Alternatively, could we just include idToken on `AccountInfo` but not `AccountEntity`? Since the Entity is the part that's cached and the Info is derived/returned

@tnorling @jasonnutter I don't understand... would I need to modify the "getAccount" function to include the id token (associated with the account) that was retrieved from the cache?

tnorling:
> > Alternatively, could we just include idToken on `AccountInfo` but not `AccountEntity`? Since the Entity is the part that's cached and the Info is derived/returned
> 
> @tnorling @jasonnutter I don't understand... would I need to modify the "getAccount" function to include the id token (associated with the account) that was retrieved from the cache?

Not `getAccount`, but rather when `BrowserCacheManager.getAllAccounts` is called you'll need to do 2 cache lookups - one for the `AccountEntity` and another for the `IdTokenEntity` and then you'll need to create the `AccountInfo` object based on those 2 entities.

tnorling:
Can we add some tests (or assertions to existing tests) that our public `getAccountXXX` and `acquireTokenXXX` APIs now return the idToken as part of the account? Otherwise looks good!

change/@azure-msal-browser-1de1d356-3446-4ed8-bd1d-439763a7abaa.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
tnorling:
```suggestion
  "type": "none",
```

Looks like there's no longer any changes to msal-browser

change/@azure-msal-browser-1de1d356-3446-4ed8-bd1d-439763a7abaa.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Exposed raw id token on account info objects #5036",
```
tnorling, line 3:
```suggestion
  "comment": "Updated unit tests #5036",
```

tnorling (approved):

PR #5039, opened 2022-07-22
Npm workspaces

tnorling:


samples/msal-angular-v2-samples/angular10-sample-app/package.json
```diff
@@ -50,5 +48,9 @@
     "ts-node": "~8.3.0",
     "tslint": "~6.1.0",
     "typescript": "~3.9.5"
-  }
+  },
+  "workspaces": [
+    "../../../lib/msal-browser",
```
peterzenz, line 53: General question, in build/packaging scripts for C++/.Net we'd try to avoid relative paths and something like this would look like "$(SourcePath)/lib/msal-browser"  Is that semantic not available here?  I think that improves readability/maintenance.

peterzenz (approved):

samples/msal-angular-v2-samples/angular10-sample-app/package.json
tnorling: Agree that would help - this isn't a script though, just a static config file. I'm not aware of a way to use variables, at least not on this specific line, but I'll confirm. How/where is "SourcePath" defined in those scripts? Is it defined as a variable in the scripts themselves or defined elsewhere?

PR #5045, opened 2022-07-28
Build msal using strict typescript

snowinmars:
I can't build my own project in strict typescript mode due to msal can't be build in the mode.

So I fixed typings - build and tests passes (except for the one with timeout).

Lint doesn't work, but stylings seems ok - `ESLint couldn't find the config "react-app" to extend from`

Idk should I bump major version or not, so it should be discussed.

### DIFF
--- a/lib/msal-core/src/Account.ts
+++ b/lib/msal-core/src/Account.ts
@@ -20,18 +20,19 @@ import { StringDict } from "./MsalTypes";
  */
 export class Account {
 
-    accountIdentifier: string;
-    homeAccountIdentifier: string;
-    userName: string;
-    name: string;
-    idToken: StringDict; // will be deprecated soon
+    accountIdentifier: string | null;
+    homeAccountIdentifier?: string;
+    userName: string | null;
+    name: string | null;
+    idToken: StringDict ; // will be deprecated soon
     idTokenClaims: StringDict;
-    sid: string;
-    environment: string;
+    sid: string | null;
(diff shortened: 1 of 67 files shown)


jasonnutter:
@snowinmars Unfortunately, it looks like a number of these changes are technically breaking (e.g. turning response properties into nullable), and we won't accept PRs that include breaking changes.

tnorling:
In addition to what Jason said - you can upgrade to MSAL.js v2 (msal-browser) which already has `strict` type checking turned on

snowinmars:
> @snowinmars Unfortunately, it looks like a number of these changes are technically breaking (e.g. turning response properties into nullable), and we won't accept PRs that include breaking changes.

There are no logic breaking changes in pr, only typings - tests prove it.
Current msal typings simply doesn't work, turning it back is a breaking change, but does the change worth the pain? I think kinda yea.

jasonnutter:
> > @snowinmars Unfortunately, it looks like a number of these changes are technically breaking (e.g. turning response properties into nullable), and we won't accept PRs that include breaking changes.
> 
> There are no logic breaking changes in pr, only typings - tests prove it. Current msal typings simply doesn't work, turning it back is a breaking change, but does the change worth the pain? I think kinda yea.

Right, but there are breaking type changes. And given v2 is in strict mode, and we are not prioritizing more investment in v1 and would like everyone to move to v2, I don't think we will accept this change.

PR #5048, opened 2022-07-28
Use login_hint claim over sid/upn for msal v1

jasonnutter:
When the `login_hint` optional id token claim is present in an id token, use it instead of `sid`.


tnorling (approved):

peterzenz (approved):

sameerag (approved):
Can we please update the guidance in docs too?

jasonnutter:
> Can we please update the guidance in docs too?

There are no `sid` references in teh docs for v1, but there are for v2, and those will get updated in the other PR. And docs.microsoft.com already mentions `login_hint`.

PR #5061, opened 2022-08-01
Add release pipeline

hectormmg:
This PR:
- Adds .pipelines directory with a governed build and release workflow and a publish template that matches our current github pre-release and release workflows (post-release workflows missing)
- Adds a prepublish script that filters the libraries to be released based on whether their package.json was updated in the last commit

### DIFF
--- a/.pipelines/OneBranch.PullRequest.yml
+++ b/.pipelines/OneBranch.PullRequest.yml
@@ -1,0 +1,161 @@
+#################################################################################
+#                      OneBranch Pipelines - PR Build                           #
+# This pipeline was created by EasyStart from a sample located at:              #
+#   https://aka.ms/obpipelines/easystart/samples                                #
+# Documentation:  https://aka.ms/obpipelines                                    #
+# Yaml Schema:    https://aka.ms/obpipelines/yaml/schema                        #
+# Retail Tasks:   https://aka.ms/obpipelines/tasks                              #
+# Support:        https://aka.ms/onebranchsup                                   #
+#################################################################################
+
+trigger: none # https://aka.ms/obpipelines/triggers
+
+parameters: # parameters are shown up in ADO UI in a build queue time
+- name: 'debug'
+  displayName: 'Enable debug output'
+  type: boolean
+  default: false
+
+variables:
+  CDP_DEFINITION_BUILD_COUNT: $[counter('', 0)] # needed for onebranch.pipeline.version task https://aka.ms/obpipelines/versioning
+  system.debug: ${{ parameters.debug }}
+  ENABLE_PRS_DELAYSIGN: 0
+  ROOT: $(Build.SourcesDirectory)
+  REPOROOT: $(Build.SourcesDirectory)
+  OUTPUTROOT: $(REPOROOT)\out
+  NUGET_XMLDOC_MODE: none
+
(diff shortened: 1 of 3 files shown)


release-scripts/prepublish.sh
```diff
@@ -0,0 +1,36 @@
...
+publishFlagNames["msal-common"]=publishMsalCommon;
+publishFlagNames["msal-browser"]=publishMsalBrowser;
+publishFlagNames["msal-node"]=publishMsalNode;
+publishFlagNames["msal-angular"]=publishMsalAngular;
+publishFlagNames["msal-react"]=publishMsalReact;
+publishFlagNames["msal-node-extensions"]=publishMsalNodeExtensions;
+
+# Iterate each library directory name
+for i in "${libNames[@]}"; do
+    libPath="../lib/${i}/package.json"
```
jasonnutter, line 17: Note that msal-node-extensions is not in the `lib` folder.

hectormmg: Good catch, I missed it.

peterzenz (approved):

.pipelines/OneBranch.PullRequest.yml
```diff
@@ -0,0 +1,161 @@
...
+        # Release MSAL Browser
+        - template: .pipelines/custom-templates/publish-template.yml@self
+          parameters:
+            jobName: 'publish_msal_browser'
+            path: 'lib'
+            libName: 'msal-browser'
+            cdn: true
+            publishFlagName: 'MsalBrowser'
+      # Release MSAL React
+        - template: .pipelines/custom-templates/publish-template.yml@self
```
tnorling, line 109: react and angular depend on browser - should these be separate stages?

.pipelines/custom-templates/publish-template.yml
```diff
@@ -0,0 +1,127 @@
...
+          workingDir: '${{ parameters.path }}/${{ parameters.libName }}'
+          customCommand: 'run link:localDeps --if-present'
+
+  # Publish to CDN
+      # KeyVault connection
+      - task: AzureKeyVault@2
+        inputs:
+          azureSubscription: 'MSIDLABKeyVault'
+          KeyVaultName: 'ADALTestInfo'
+          SecretsFilter: 'MSALJS-CDN-SAS-EUNO, MSALJS-CDN-SAS-USWE, MSALJSNPMTOKEN'
```
tnorling, line 56: We will need to make sure that the current tokens are marked as such in the key vault - we have been sloppy about this in the past

hectormmg: Not sure exactly what you mean? Do you mean that the names should match?

.pipelines/OneBranch.PullRequest.yml
hectormmg: I've marked dependencies and added a success condition, keeping all the publications on the same stage since it doesn't make much difference and we'll have less artifacts this way.

tnorling (approved):

.pipelines/OneBranch.PullRequest.yml
hectormmg: I've marked dependencies and added a success condition, keeping all the publications on the same stage since it doesn't make much difference and we'll have less artifacts this way.

PR #5062, opened 2022-08-01
AcquireTokenInteractive API

tnorling:
Adds acquireTokenInteractive API to msal-node Public Clients


lib/msal-common/src/url/UrlString.ts
```diff
@@ -192,6 +192,21 @@ export class UrlString {
         return Constants.EMPTY_STRING;
     }
 
+    /**
+     * Parses query string from given string. Returns empty string if no query symbol is found.
+     * @param queryString
+     */
+    static parseQueryString(queryString: string): string {
+        const queryIndex1 = queryString.indexOf("?");
```
jasonnutter, line 200: Possible to use `URL` and/or `URLSearchParams`? https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams#examples

tnorling: The response doesn't come in as a fully formed URL so that API throws

lib/msal-node/src/error/NodeAuthError.ts
```diff
@@ -0,0 +1,45 @@
...
+        desc: "Host header is missing in the request. This is unexpected."
+    }
+};
+
+export class NodeAuthError extends AuthError {
+    constructor(errorCode: string, errorMessage?: string) {
+        super(errorCode, errorMessage);
+        this.name = "NodeAuthError";
+
+        Object.setPrototypeOf(this, NodeAuthError.prototype);
```
bgavrilMS: nit: This is pretty tricky, I'd add a comment about prototype chain here...

tnorling: Thanks for pointing this out...lazy copy/paste from some of our other error classes, don't think it's needed here so I've removed it.

bmahall (approved):

lib/msal-node/src/client/PublicClientApplication.ts
```diff
@@ -74,4 +84,83 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
+                const id = setInterval(() => {
+                    const address = server.address();
+                    if (typeof address === "string") {
+                        clearInterval(id);
+                        reject(NodeAuthError.createAddressWrongTypeError());
+                    } else if (address && address.port) {
+                        clearInterval(id);
+                        resolvePort(address.port);
+                    }
+                }, 100);
```
jasonnutter: Can you create a constant for this value? And do we want this to be configurable?

lib/msal-node/src/client/PublicClientApplication.ts
```diff
@@ -74,4 +84,83 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
+
+    /**
+     * Acquires a token by requesting an Authorization code then exchanging it for a token.
+     */
+    async acquireTokenInteractive(request: InteractiveRequest): Promise<AuthenticationResult> {
+        const { verifier, challenge } = await this.cryptoProvider.generatePkceCodes();
+        const { openBrowser, successTemplate, errorTemplate, ...remainingProperties } = request;
+
+        return new Promise(async (resolve, reject) => {
+            const server = createServer((req: IncomingMessage, res: ServerResponse) => {
```
jasonnutter: I think it would be cleaner if you created a `LoopbackClient` or somthing that hides most the implementation details of how this work.

lib/msal-node/src/client/PublicClientApplication.ts
```diff
@@ -74,4 +84,83 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
+                        server.close();
+                    });
+                } else if (authCodeResponse.error) {
+                    res.end(errorTemplate || `Error occurred: ${authCodeResponse.error}`);
+                } else {
+                    res.end(errorTemplate || "Unknown error occurred");
+                }
+            });
+            server.listen(0); // Listen on any available port
+            const port: number = await new Promise((resolvePort) => {
```
jasonnutter, line 140: Would this never throw?

lib/msal-node/src/client/PublicClientApplication.ts
```diff
@@ -74,4 +84,83 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
+                const id = setInterval(() => {
+                    const address = server.address();
+                    if (typeof address === "string") {
+                        clearInterval(id);
+                        reject(NodeAuthError.createAddressWrongTypeError());
```
jasonnutter: This bit of indirection is confusing. Would rather see this promise throw, and then have code to catch and call `reject`.

lib/msal-node/src/client/PublicClientApplication.ts
tnorling: Created a constant. Don't think this needs to be configurable but we can re-evaluate if someone asks for it

lib/msal-node/src/client/PublicClientApplication.ts
tnorling: Done

lib/msal-node/src/client/PublicClientApplication.ts
tnorling: Done

lib/msal-node/src/client/PublicClientApplication.ts
```diff
@@ -74,4 +84,51 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
+        const { verifier, challenge } = await this.cryptoProvider.generatePkceCodes();
+        const { openBrowser, successTemplate, errorTemplate, ...remainingProperties } = request;
+
+        const loopbackClient = new LoopbackClient();
+        const authCodeListener = loopbackClient.listenForAuthCode(successTemplate, errorTemplate);
+        const port = await loopbackClient.getPort().catch((err) => {
+            // Close server and re-throw any error thrown when getting port
+            loopbackClient.closeServer();
+            throw err;
+        });
```
jasonnutter: Could the implementation detail of the port be hidden behind a `getRedirectUri()` function?

tnorling: Done

lib/msal-common/src/utils/Constants.ts
```diff
@@ -48,7 +48,7 @@ export const Constants = {
     KNOWN_PUBLIC_CLOUDS: ["login.microsoftonline.com", "login.windows.net", "login.microsoft.com", "sts.windows.net"],
     TOKEN_RESPONSE_TYPE: "token",
     ID_TOKEN_RESPONSE_TYPE: "id_token",
-    SHR_NONCE_VALIDITY: 240,
+    SHR_NONCE_VALIDITY: 240
```
peterzenz, line 51: Do you know if the nonce returned includes the validity window?  Would be great to use what is returned to us with the nonce instead of a const.

lib/msal-node/src/network/LoopbackClient.ts
```diff
@@ -0,0 +1,87 @@
...
+     * Get the port that the loopback server is running on
+     * @returns 
+     */
+    async getRedirectUri(): Promise<string> {
+        if (!this.server) {
+            throw NodeAuthError.createNoLoopbackServerExistsError();
+        }
+        const port: number = await new Promise((resolve, reject) => {
+            let ticks = 0;
+            const id = setInterval(() => {
```
peterzenz: This pattern is really curious to me.  Repeatedly (50x over 5 seconds) trying to retrieve a port like this feels like an old bug in the runtime you're trying to work around.  Why are we doing it this way?  Operations like this in native or .net code would either return a port immediately or fail with none available/other error.  We might implement a single retry before giving up.  None of those operations would be async.

samples/msal-node-samples/auth-code-cli-app/src/authConfig.js
```diff
@@ -0,0 +1,26 @@
+const { LogLevel } = require("@azure/msal-node");
+
+const cacheLocation = "./src/data/cache.json";
+const cachePlugin = require('../../cachePlugin')(cacheLocation);
+
+module.exports = {
+    authConfig: {
+        auth: {
+            clientId: "c3a8e9df-f1d4-427d-be73-acab139c40fd",
```
peterzenz: Use a zero guid or some other indicator that they have to enter something here.  Comment that this needs to be updated here would be more useful than what you have on Index.js, line 10

samples/msal-node-samples/auth-code-cli-app/src/index.js
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+const { PublicClientApplication, InteractionRequiredAuthError } = require("@azure/msal-node");
+const open = require("open");
+const { authConfig } = require("./authConfig.js");
+const { callMicrosoftGraph } = require("./graph.js");
+
+// Before running the sample, you will need to replace the values in the config.
```
peterzenz: ...values in src/authConfig.js

peterzenz (approved):
Not blocking, but I'm curious on the setInterval pattern you're using.  Thanks.

lib/msal-common/src/utils/Constants.ts
tnorling: Not sure - nonce is not part of this PR

lib/msal-node/src/network/LoopbackClient.ts
tnorling: In reality this should return on the first or second interval (100ms intervals for 5s were just the defaults I chose arbitrarily, we probably could get away with a much shorter timeout). The reason this is needed is because the server's event listener may not be registered yet by the time this gets called, resulting in a null response. 

An alternative solution (which is perhaps more intuitive) would have been to use the interval to check if the server is listening and then get the port once as you expect. Since the interval is needed either way I took the shortcut - though I see that was maybe not the right choice for readability. I'm open to reconsidering that decision.

jasonnutter: Can you add a comment detailing this implementation detail?

sameerag: > In reality this should return on the first or second interval (100ms intervals for 5s were just the defaults I chose arbitrarily, we probably could get away with a much shorter timeout). The reason this is needed is because the server's event listener may not be registered yet by the time this gets called, resulting in a null response.
> 
> An alternative solution (which is perhaps more intuitive) would have been to use the interval to check if the server is listening and then get the port once as you expect. Since the interval is needed either way I took the shortcut - though I see that was maybe not the right choice for readability. I'm open to reconsidering that decision.

Sounds like polling and I would prefer the separation. Even if technically what you did works, it would be easier to maintain and improve the code if we separate the listening and port acquisition parts, even for performance measurements and telemetry.

lib/msal-common/src/url/UrlString.ts
```diff
@@ -215,6 +230,25 @@ export class UrlString {
...
 
+    /**
+     * Returns URL query string as server auth code response object.
+     */
+    static getDeserializedQueryString(query: string): ServerAuthorizationCodeResponse {
+        // Check if given query is empty
+        if (StringUtils.isEmpty(query)) {
+            return {};
+        }
+        // Strip the # symbol if present
```
sameerag: Is this the right comment? Isn't this for `getDeserializedHash()`?

lib/msal-node/src/error/NodeAuthError.ts
```diff
@@ -0,0 +1,91 @@
...
+ * Licensed under the MIT License.
+ */
+
+import { AuthError } from "@azure/msal-common";
+
+/**
+ * NodeAuthErrorMessage class containing string constants used by error codes and messages.
+ */
+export const NodeAuthErrorMessage = {
+    addressWrongType: {
```
sameerag: `invalidAddressType`?

lib/msal-node/src/network/LoopbackClient.ts
```diff
@@ -0,0 +1,87 @@
...
+                    return;
+                } else if (url === CommonConstants.FORWARD_SLASH) {
+                    res.end(successTemplate || "Auth code was successfully acquired. You can close this window now.");
+                    return;
+                }
+    
+                const authCodeResponse = UrlString.getDeserializedQueryString(url);
+                if (authCodeResponse.code) {
+                    const redirectUri = await this.getRedirectUri();
+                    res.writeHead(302, { location: redirectUri }); // Prevent auth code from being saved in the browser history
```
sameerag: can we have an enum for Http statuses (and hide`302` under the constant)?

sameerag (commented):
Left some comments. Would like to see the separation in the polling code.

samples/msal-node-samples/auth-code-cli-app/README.md
```diff
@@ -0,0 +1,51 @@
+# MSAL Node Standalone Sample: Authorization Code Grant (PKCE) In Typescript
+
+This sample demonstrates an MSAL Node [public client application](../../../lib/msal-node/docs/initialize-public-client-application.md) that lets users authenticate against **Azure AD**.
+
+This sample uses the [OAuth 2.0 Authorization Code Grant](https://oauth.net/2/grant-types/authorization-code/) flow with [Proof-Key For Code Exchange](https://oauth.net/2/pkce/) (PKCE). This flow is particularly suitable for public client applications like desktop and mobile apps.
+
+> This sample demonstrates how to use and setup `msal-node` in a [typescript](https://www.typescriptlang.org) project.
```
derisen:
```suggestion
```

samples/msal-node-samples/auth-code-cli-app/src/index.js
```diff
@@ -0,0 +1,61 @@
...
+
+
+// Get url to sign user in and consent to scopes needed for application
```
derisen: Can we have a more descriptive comment here, and a remark about the **open** dependency?

samples/msal-node-samples/auth-code-cli-app/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Run the app
+
+In the same folder, type:
+
+```console
+    npm start
+```
+
+The server should start at port **3000**. Navigate to `http://localhost:3000` in your browser, which will trigger the token acquisition process.
```
derisen:
```suggestion
The app should attempt to acquire a token interactively and make a call to Microsoft Graph afterwards.
```

samples/msal-node-samples/auth-code-cli-app/README.md
```diff
@@ -0,0 +1,51 @@
...
+```
+
+## Register
+
+1. Navigate to the [Azure portal](https://portal.azure.com) and select the **Azure AD** service.
+1. Select the **App Registrations** blade on the left, then select **New registration**.
+1. In the **Register an application page** that appears, enter your application's registration information:
+   - In the **Name** section, enter a meaningful application name that will be displayed to users of the app, for example `msal-node-webapp`.
+   - Under **Supported account types**, select **Accounts in this organizational directory only**.
+   - In the **Redirect URI (optional)** section, select **Public client/native (mobile & desktop)** in the combo-box and enter the following redirect URI: `http://localhost:3000/redirect`.
```
derisen: Is this redirect URI actually used?

samples/msal-node-samples/auth-code-cli-app/src/authConfig.js
derisen: +1

derisen:
@tnorling can we also have an entry in the docs, perhaps in [request.md](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-common/docs/request.md)? Why/when would a developer want to use the `acquireTokenInteractive` API? Any implications on current Electron samples?

lib/msal-node/src/network/LoopbackClient.ts
peterzenz: Either a comment on the current approach as Jason suggested or a separation as Sameera suggested would be OK here.  I'd lean more towards Sameera's suggestion as that implementation approach feels cleaner.

tnorling: Done - refactored to poll for the listener after creating it.

lib/msal-common/src/url/UrlString.ts
tnorling: Fixed.

lib/msal-node/src/error/NodeAuthError.ts
tnorling: Changed.

lib/msal-node/src/network/LoopbackClient.ts
tnorling: Done

samples/msal-node-samples/auth-code-cli-app/README.md
tnorling: Thanks for the catch!

samples/msal-node-samples/auth-code-cli-app/src/index.js
tnorling: Done

samples/msal-node-samples/auth-code-cli-app/src/index.js
tnorling: Done

samples/msal-node-samples/auth-code-cli-app/src/authConfig.js
tnorling: Done

sameerag (approved):

PR #5063, opened 2022-08-01
RT size in perf

bmahall:
This PR adds Refresh Token size fetched from the RefreshTokenClient at time of RT acquisition in case of both cached/unavailable RT to the performance telemetry data . It adds two new submeasurements RefreshTokenClientAcquireToken, RefreshTokenClientAcquireTokenWithCachedRefreshToken to the PerformanceEvent enum .


change/@azure-msal-browser-883260a8-b8e9-4d81-a005-d5eb07364fbe.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
jasonnutter:
```suggestion
  "type": "patch",
```

Don't consider this worthy of a minor, as it isnt introducing new behavior.

change/@azure-msal-browser-883260a8-b8e9-4d81-a005-d5eb07364fbe.json
```diff
@@ -0,0 +1,7 @@
...
+  "comment": "RT size in telemetry data",
```
jasonnutter:
```suggestion
  "comment": "RT size in telemetry data #5036",
```

lib/msal-browser/src/interaction_client/SilentRefreshClient.ts
```diff
@@ -63,6 +63,6 @@ export class SilentRefreshClient extends StandardInteractionClient {
     protected async createRefreshTokenClient(serverTelemetryManager: ServerTelemetryManager, authorityUrl?: string, azureCloudOptions?: AzureCloudOptions): Promise<RefreshTokenClient> {
         // Create auth module.
         const clientConfig = await this.getClientConfiguration(serverTelemetryManager, authorityUrl, azureCloudOptions);
-        return new RefreshTokenClient(clientConfig, this.performanceClient);
+        return new RefreshTokenClient(clientConfig);
```
jasonnutter: Are we sure this is the right approach? This was an intentional change in the other PR, do the tests need to be updated?

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -50,6 +51,19 @@ export class RefreshTokenClient extends BaseClient {
...
+        if (response.body?.refresh_token) {
+            atsMeasurement?.endMeasurement({
+                success: true,
+                refreshTokenSize: response?.body?.refresh_token.length
+            });
+        }
+        else {
+            atsMeasurement?.endMeasurement({
+                success: true,
+                refreshTokenSize: -1
```
tnorling: What are the pros/cons of using -1 vs simply not providing a value in this case?

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -108,12 +123,20 @@ export class RefreshTokenClient extends BaseClient {
...
 
-        // no refresh Token
         if (!refreshToken) {
+            atsMeasurement?.discardMeasurement();
             throw InteractionRequiredAuthError.createNoTokensFoundError();
         }
+        // attach cached RT size to the current measurement
+        atsMeasurement?.endMeasurement({
+            success: true,
+            refreshTokenSize: refreshToken.secret.length
```
tnorling: Is this measuring the RT size a 2nd time?

tnorling: Should we measure the RT size at time of acquisition rather than at time of use?

lib/msal-common/src/client/RefreshTokenClient.ts
jasonnutter: Would rather this be undefined than -1, as -1 would throw off any calculations.

jasonnutter: If you want to use a number, it should be 0.

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -50,6 +51,19 @@ export class RefreshTokenClient extends BaseClient {
         );
 
         responseHandler.validateTokenResponse(response.body);
+        if (response.body?.refresh_token) {
+            atsMeasurement?.endMeasurement({
+                success: true,
+                refreshTokenSize: response?.body?.refresh_token.length
```
jasonnutter:
```suggestion
                refreshTokenSize: response?.body?.refresh_token.length || 0
```

jasonnutter: Then the else below isn't needed.

lib/msal-common/test/client/RefreshTokenClient.spec.ts
```diff
@@ -139,27 +115,6 @@ describe("RefreshTokenClient unit tests", () => {
...
-        it("Checks whether performance telemetry endMeasurement method is called",async () => {
-            const spy = jest.spyOn(stubPerformanceClient, 'endMeasurement');
-            
-            const client = new RefreshTokenClient(config,stubPerformanceClient);
-            sinon.stub(RefreshTokenClient.prototype, <any>"executePostToTokenEndpoint").resolves(AUTHENTICATION_RESULT);
-            
-            await client.acquireToken(refreshTokenRequest);
-            expect(spy).toHaveBeenCalled();
-            spy.mockClear();
-        });
```
jasonnutter, line 164: Why are these tests deleted?

change/@azure-msal-browser-883260a8-b8e9-4d81-a005-d5eb07364fbe.json
jasonnutter: Also, this can be a patch, no new behavior introduced.

jasonnutter:
```suggestion
  "type": "patch",
```

lib/msal-common/src/client/RefreshTokenClient.ts
peterzenz: This should translate to null/empty in Aria if there is no token.  It would be helpful when looking at the telemetry to infer that a null in the RT size implies no token was returned, where a 0 would indicate a token was returned but it was zero bytes long (a code error somewhere on the client or server)

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -25,18 +25,19 @@ import { UrlString } from "../url/UrlString";
...
  */
 export class RefreshTokenClient extends BaseClient {
     constructor(configuration: ClientConfiguration, performanceClient?: IPerformanceClient) {
         super(configuration,performanceClient);
-    }
 
+    }
     public async acquireToken(request: CommonRefreshTokenRequest): Promise<AuthenticationResult> {
+        // @ts-ignore
+        const atsMeasurement = this.performanceClient?.startMeasurement("RefreshTokenClientAcquireToken", request.correlationId);
```
jasonnutter: Can this be added to the enum?

lib/msal-common/src/client/RefreshTokenClient.ts
jasonnutter: Also, I think we would want this measurement not to end until `handServerTokenResponse` has returned below (i.e. do this logic in the `.then`/`.catch` of `responseHandler.handleServerTokenResponse`).

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -108,12 +123,20 @@ export class RefreshTokenClient extends BaseClient {
      */
     private async acquireTokenWithCachedRefreshToken(request: CommonSilentFlowRequest, foci: boolean) {
         // fetches family RT or application RT based on FOCI value
+
+        // @ts-ignore
+        const atsMeasurement = this.performanceClient?.startMeasurement("RefreshTokenClientAcquireTokenWithCachedRefreshToken", request.correlationId);
```
jasonnutter: This can also be added to the enum (lets always add to the enum except for in the 1p repo).

bmahall: Got it, thanks . Have added both to the enum.

lib/msal-common/src/client/RefreshTokenClient.ts
bmahall: Added

lib/msal-common/test/client/RefreshTokenClient.spec.ts
bmahall: Added back, had removed to make it consistent with earlier made changes (i.e Perf Client a part of Client Config), have undoed those.

lib/msal-common/src/client/RefreshTokenClient.ts
bmahall: Yes, makes sense. Have updated the file.

lib/msal-common/src/client/RefreshTokenClient.ts
bmahall: Thanks for pointing this, great catch! Have removed this as its redundant, only measuring at time of RT acquistion in acquire token()

lib/msal-common/src/client/RefreshTokenClient.ts
bmahall: Using 0 in case of RT length not available (in event of token not returned in response body), and undefined when there is an error in the token acquisition

lib/msal-browser/src/interaction_client/SilentRefreshClient.ts
bmahall: Thanks for adding this , have addressed it

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -25,18 +25,19 @@ import { UrlString } from "../url/UrlString";
...
-
+import { IPerformanceClient } from "../telemetry/performance/IPerformanceClient";
 /**
  * OAuth2.0 refresh token client
  */
 export class RefreshTokenClient extends BaseClient {
     constructor(configuration: ClientConfiguration, performanceClient?: IPerformanceClient) {
-        super(configuration,performanceClient);
-    }
+        super(configuration, performanceClient);
```
bmahall, line 35: Addressed based on previous feedback, maintaining performanceClient as separate parameter to be passed to the RefreshTokenClient and SilentFlowClient

lib/msal-common/src/client/RefreshTokenClient.ts
peterzenz: Please leave this undefined when no RT is returned.  Otherwise, we won't be able to differentiate not getting a token back and getting back a token that is zero bytes in size.

bmahall: There is some issue, the code is not updated . Can you please check https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/5063/files -> Changes from all commits , in case it still shows the older code, I'll need to check the commit sequence on my end, thanks

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -59,7 +61,24 @@ export class RefreshTokenClient extends BaseClient {
...
+                });
+            }
+            return result;
+        })
+            .catch((error) => {
+                atsMeasurement?.endMeasurement({
+                    errorCode: error.errorCode,
+                    subErrorCode: error.subError,
+                    success: false,
+                    refreshTokenSize: undefined
```
bmahall, line 77: @peterzenz Adding here for visibility

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -59,7 +61,24 @@ export class RefreshTokenClient extends BaseClient {
             undefined,
             true,
             request.forceCache
-        );
+        ).then((result: AuthenticationResult) => {
+            if (response.body?.refresh_token) {
+                atsMeasurement?.endMeasurement({
+                    success: true,
+                    refreshTokenSize: response?.body?.refresh_token.length || 0
```
peterzenz: Should this be:
```
response?.body?.refresh_token?.length || undefined
```

peterzenz (changes requested):

lib/msal-common/src/client/RefreshTokenClient.ts
tnorling: FYI that will turn length 0 into undefined. I'm assuming that's not what we want? Should verify the type of length first.

peterzenz (approved):

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -25,18 +25,20 @@ import { UrlString } from "../url/UrlString";
...
  */
 export class RefreshTokenClient extends BaseClient {
     constructor(configuration: ClientConfiguration, performanceClient?: IPerformanceClient) {
-        super(configuration,performanceClient);
-    }
+        super(configuration, performanceClient);
 
+    }
     public async acquireToken(request: CommonRefreshTokenRequest): Promise<AuthenticationResult> {
+        // @ts-ignore
```
tnorling: What is this ignoring?

lib/msal-common/src/client/RefreshTokenClient.ts
jasonnutter: If `response.body?.refresh_token` is truthy, the length will never be 0 (i.e. 0 is falsey). I don't understand why the if statement is needed here?

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -59,7 +62,32 @@ export class RefreshTokenClient extends BaseClient {
...
+                    refreshTokenSize: response?.body?.refresh_token.length || 0
+                });
+            }
+            else{
+                // no refresh token is returned
+                atsMeasurement?.endMeasurement({
+                    success: true,
+                    refreshTokenSize: undefined
+                });
+            }
```
tnorling: Something like this might be more concise
```suggestion
            let refreshTokenSize;
            if (response.body?.refresh_token) {
                refreshTokenSize = response.body.refresh_token.length || 0;
            }
            atsMeasurement?.endMeasurement({
                success: true,
                refreshTokenSize: refreshTokenSize
            });
```

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -96,10 +124,11 @@ export class RefreshTokenClient extends BaseClient {
                     throw e;
                 }
             }
+        } else {
```
tnorling: Reason for this change?

lib/msal-common/src/client/RefreshTokenClient.ts
jasonnutter: Are we wanting to differentiate between not having a refresh token and a refresh token of length 0 (which I don't think is a scenario that can happen, and even if so, should be treated the same semantically imo.

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -108,12 +137,20 @@ export class RefreshTokenClient extends BaseClient {
      */
     private async acquireTokenWithCachedRefreshToken(request: CommonSilentFlowRequest, foci: boolean) {
         // fetches family RT or application RT based on FOCI value
+
+        // @ts-ignore
```
tnorling: What is this ignoring?

lib/msal-common/src/telemetry/performance/PerformanceEvent.ts
```diff
@@ -101,25 +101,25 @@ export enum PerformanceEvents {
...
+     * Time taken for acquiring refresh token , records RT size
      */
-    BaseClientCreateTokenRequestHeaders = "baseClientCreateTokenRequestHeaders",
+    RefreshTokenClientAcquireToken = "refreshTokenClientAcquireToken",
 
     /**
-     * Used to measure the time taken for completing embedded-broker handshake (PW-Broker).
+     * Time taken for acquiring cached refresh token 
      */
-    BrokerHandhshake = "brokerHandshake",
```
tnorling, line 111: Reason for removing this?

lib/msal-common/src/telemetry/performance/PerformanceEvent.ts
```diff
@@ -101,25 +101,25 @@ export enum PerformanceEvents {
...
+    RefreshTokenClientAcquireTokenWithCachedRefreshToken = "refreshTokenClientAcquireTokenWithCachedRefreshToken",
 
     /**
-     * acquireTokenByRefreshToken API in BrokerClientApplication (PW-Broker) .
+     * Time spent creating default headers for requests to token endpoint
      */
-    AcquireTokenByRefreshTokenInBroker = "acquireTokenByRefreshTokenInBroker",
```
tnorling, line 116: Reason for removing this?

lib/msal-common/src/client/RefreshTokenClient.ts
jasonnutter:
```suggestion
                    refreshTokenSize: response?.body?.refresh_token ? response?.body?.refresh_token.length : undefined
```

jasonnutter: And then remove the if statement and just have one endMeasurement call.

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -59,7 +62,32 @@ export class RefreshTokenClient extends BaseClient {
             undefined,
             true,
             request.forceCache
-        );
+        ).then((result: AuthenticationResult) => {
+            if (response.body?.refresh_token) {
```
bmahall: @jasonnutter  Yes, this if-else block is to differentiate between when there is no refresh token in the response (i.e no Refresh token is returned at all) vs when there is a refresh token of length 0 (e.g due to network error, or something else). 

It was desired to make a distinction between these two cases, https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/5063#discussion_r960918678

lib/msal-common/src/client/RefreshTokenClient.ts
bmahall: This sounds feasible, just the distinction between no RT vs size 0 RT would then be treated similarly .

lib/msal-common/src/client/RefreshTokenClient.ts
bmahall: Thanks. This definitely is cleaner, would update. @tnorling

lib/msal-common/src/client/RefreshTokenClient.ts
bmahall: This is unintentional , removing

lib/msal-common/src/client/RefreshTokenClient.ts
bmahall: This is from previous case where didnt have the events added in the Performance Event enum, removing the two ts-ignore statements, thanks for mentioning.

lib/msal-common/src/telemetry/performance/PerformanceEvent.ts
bmahall: These two events are not being used in the PW-Broker flow code in 1p . Were added initially prior to the actual instrumentation.

lib/msal-common/src/client/RefreshTokenClient.ts
jasonnutter: > Please leave this undefined when no RT is returned. Otherwise, we won't be able to differentiate not getting a token back and getting back a token that is zero bytes in size.

Is this something that will actually happen? Seems like were optimizing for a scenario that won't happen in practice (and in my opinion, doesn't need to be treated differently).

lib/msal-common/src/client/RefreshTokenClient.ts
jasonnutter: I disagree that this is something we should differentiate. Comment left below.

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
```diff
@@ -277,6 +277,9 @@ export abstract class PerformanceClient implements IPerformanceClient {
                         if (current.idTokenSize) {
                             previous.idTokenSize = current.idTokenSize;
                         }
+                        if (current.hasOwnProperty("refreshTokenSize")) {
```
jasonnutter: Why `hasOwnProperty` here, versus just:

```suggestion
                        if (currentrefreshTokenSize) {
```

jasonnutter: Would like to be consistent here.

lib/msal-common/src/telemetry/performance/PerformanceEvent.ts
jasonnutter: That seems unrelated to the rest of this PR, please open a separate PR if so.

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -59,7 +61,27 @@ export class RefreshTokenClient extends BaseClient {
...
+        ).then((result: AuthenticationResult) => {
+            let refreshTokenSize;
+            if (response.body?.refresh_token) {
+                refreshTokenSize = response.body.refresh_token.length || 0;
+            }
+            atsMeasurement?.endMeasurement({
+                success: true,
+                refreshTokenSize: refreshTokenSize
+            });
+            return result;
```
jasonnutter: If you are wanting to differentiate, you need to explicitly check the type of `body.refresh_token`. Simply doing a truthy check will result in `undefined` and a string of length 0 being treated the same (which you indicated you don't want to do). Furthermore, this can be simplified (also, please leave comment detailing why this is here, as it is not necessarily obvious):

```suggestion
            // Don't set refreshtokensize if no refresh token
            const refreshTokenSize = typeof response.body?.refresh_token === "undefined" ? undefined : (response.body.refresh_token?.length || 0);
            atsMeasurement?.endMeasurement({
                success: true,
                refreshTokenSize
            });
            return result;
```

jasonnutter: Please add a unit test to confirm this logic.

jasonnutter: But as noted before, I would personally not to differentiate between undefined and length of zero:

```suggestion
            atsMeasurement?.endMeasurement({
                success: true,
                refreshTokenSize: response.body.refresh_token?.length || 0
            });
            return result;
```

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
bmahall: Would this attach the size where it is undefined as well? @jasonnutter

bmahall: Since we don't need to attach the metric when it's undefined, I would revert this. Thanks for the suggestion!

jasonnutter: It would not attach it if `refreshTokenSize` is any falsey value (0, undefined, null, nan, etc)

lib/msal-common/src/client/RefreshTokenClient.ts
peterzenz: We should handle missing data consistently.  If we don't have something from the server or the user we shouldn't log a default there.

lib/msal-common/src/client/BaseClient.ts
```diff
@@ -83,7 +83,7 @@ export abstract class BaseClient {
     /**
      * Creates default headers for requests to token endpoint
      */
-    protected createTokenRequestHeaders(ccsCred?: CcsCredential): Record<string, string> {        
+    protected createTokenRequestHeaders(ccsCred?: CcsCredential): Record<string, string> {
```
bmahall, line 86: no real changes

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -59,7 +61,23 @@ export class RefreshTokenClient extends BaseClient {
             undefined,
             true,
             request.forceCache
-        );
+        ).then((result: AuthenticationResult) => {
+            atsMeasurement?.endMeasurement({
+                success: true,
+                refreshTokenSize: response.body.refresh_token?.length || 0
```
bmahall, line 67: The refresh token size is recorded when a refresh token is returned by acquireToken() and is taken as 0 in all other scenarios for consistency purposes.

lib/msal-common/src/client/SilentFlowClient.ts
```diff
@@ -22,7 +22,7 @@ export class SilentFlowClient extends BaseClient {
     constructor(configuration: ClientConfiguration, performanceClient?: IPerformanceClient) {
         super(configuration,performanceClient);
     }
-
+
```
bmahall, line 25: no real changes

lib/msal-common/src/client/RefreshTokenClient.ts
bmahall: Thanks for your helpful feedback  @jasonnutter ! I have included the code changes based on the second approach(i.e not semantically differentiating between RT of size 0 and undefined). Have included appropriate tests in the RefreshTokenClient. Thanks!

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
bmahall: updated!

lib/msal-common/src/client/RefreshTokenClient.ts
bmahall: Agree, addressed in latest changes .

.github/workflows/metadata-check.yml
```diff
@@ -19,13 +19,13 @@ jobs:
   metadata-checker:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v2
```
jasonnutter, line 22: Is this intentional?

.beachballrc
```diff
@@ -2,6 +2,5 @@
     "branch": "dev",
     "access": "public",
     "changehint": "Need to run the following command before merging this PR: npm run beachball:change",
-    "scope": ["lib/*", "extensions/*"],
-    "ignorePatterns": ["lib/**/**.md", "extensions/**/**.md"]
+    "scope": ["lib/*", "extensions/*"]
```
jasonnutter, line 5: This also doesn't seem right.

jasonnutter:
@bmahall It seems this branch is now in a bad state, as it seems to revert a bunch of recently made changes. Might be best to close this PR and start a new branch from dev.

jasonnutter (changes requested):
Please either open a new PR or make sure not to revert the recently made changes in dev, which this PR now is doing.

jasonnutter:
Closing as #5250 has been merged.

PR #4502, opened 2022-02-08
Update tests and docs for AzureCloudInstance support

sameerag:

PR #5073, opened 2022-08-02
add authority doc to msal-common

### LINKED ISSUE #4966: More Information needed for "authority" option in MSAL js library for Azure AD B2C. It seems like a very important option.
### Core Library

### Wrapper Library

MSAL Angular (@azure/msal-angular), MSAL React (@azure/msal-react), MSAL Node Extensions (@azure/msal-node-extensions), Not Applicable

### Documentation Location

https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-angular/docs/v2-docs/redirects.md

### Description

I'll start with this as an example. 

```
https://login.bbbbbbbb.com/bbb5bbbbbbb01/b2c_1a_signup_scccccn/oauth2/v2.0/authorize?client_id=bbbbbbbbbbbbb&scope=openid%20profile%20offline_access&redirect_uri=http%3A%2F%2Flocalhost%3A6420&client-request-id=fbbbbbbbbbbbc7d5f&response_mode=fragment&response_type=code&x-client-SKU=msal.js.browser&x-client-VER=2.13.1&x-client-OS=&x-client-CPU=&client_info=1&code_challenge=etq0YbbbbbbVxbbbNjuZ-WL3bbbYtw&code_challenge_method=S256&nonce=390bbbb46bc-a46f-bb0167127b4c&state=eyJpZCbbbbbbbFjNiIsIm1ldGEiOnsiaWcccccccccccccccX0%3D
```

This is what I see from above ```<my custom domain>/<my app reg directory id (not tenant id)>/<policy-name>/oauth2/v2.0/authorize? + <clientId> + [[<openid> + <offline_access>]] + <redirect_uri> + <client-request-id> + <response_mode> + <response_type> + <code> + <x-client-sku> + <x-client-VER> + <x-client-OS> + <x-client-CPU> + <client_info> + <code_challenge> + <code_challenge_method> + <nonce> + <state>```

I know about 40% of what's going on up there. It pretty much falls off at client-request-id. Is there an api flow diagram explaining exactly what's going on? 

I'll pick out `nonce` for example. I didn't set that anywhere but yet it's there. Any additional information about that I should know?

***More over, and this is kind of the entire point, what makes this URI `"https://bbbbbbbbb.com/bbbbbb342-bbbd01/B2C_1A_SIGNUP_SIGbbbb"` do/transform the above URI with the following settings:

```
auth: {
      clientId: "............", // This is the ONLY mandatory field; everything else is optional.
      **[[authority]]**: b2cPolicies.authorities.signUpSignIn.authority, // Choose sign-up/sign-in user-flow as your default.
      knownAuthorities: [b2cPolicies.authorityDomain], // You must identify your tenant's domain as a known authority.
      redirectUri: "http://localhost:6420", // You must register this URI on Azure Portal/App Registration. Defaults to "window.location.href".
    },
```

The only thing I can tell that is kicking off the above authority is the `<Endpoints>` section in Azure AD B2C and specifically it is the `Azure AD B2C OAuth 2.0 authorization endpoint (v2)` section. Can I use others? When do I choose token endpoint (v2) versus OpenID Connect metadata document? 

***Which brings me to the sample Angular v2 docs and react for that matter. 

```
export function MSALInstanceFactory(): IPublicClientApplication {
    auth: {
      clientId: 'bbbbbbbbbbb6',
      redirectUri: 'http://localhost:4200',
      postLogoutRedirectUri: 'http://localhost:4200'
    },
``` 

here the authority is not here. Should I add that like I did in the other sample app? Does this MSAL configuration need for me to find the AD B2C authorization endpoint v2 (for PKCE) (???is the token endpoint only for implicit MSAL v1???) flows and therefore that endpoint is the one I should choose? 

Also, would this documentation, I found somewhere, not be entirely correct? 

URI of the tenant to authenticate and authorize with. Usually takes the form of: (because... above)

```
https://{uri}/{tenantid}.
```

It becomes a "how did we get here" exercise that is painful based finding and linking and reading and finding and linking and reading... the current documentation situation. 

As well, I found this in the docs for MSAL JS v2. I am so happy after a week I even found those docs.
 
```
Type declaration
Optional authority?: string
You can configure a specific authority, defaults to " " or "[https://login.microsoftonline.com/common"](https://login.microsoftonline.com/common%22)
```

You see, that doesn't tell me anything at all. If I experiment with it over hours and hours eventually I can figure it out. But on something like the URI/authority why isn't it just fully and completely explained? It literally controls your login. It's not a small thing. 

Also, when you're putting these in wrappers it's even more abstracted away so if you don't explain it in the docs then it's always a pain to figure it out. 

With that said, if you all improved your documenation and created a more comprehensive spec that includes some important parts of the api i.e., like `Authority` you would see so much more wide spread adoption. Even the youtube videos are, in a sense, lacking.

derisen:
* Adds a doc for giving an overview of authority in MSAL and related configuration options
* Adds links to other docs mentioning authority

### DIFF
--- a/lib/msal-common/docs/authority.md
+++ b/lib/msal-common/docs/authority.md
@@ -1,17 +1,15 @@
 # Authority in MSAL
 
-In OAuth 2.0 and OpenID Connect, a client application (aka *relying party*) acquires tokens from an identity provider (IdP), such as Azure AD, to prove authentication (via ID tokens) and/or authorization (via access tokens). When doing so, the client application trusts the *authority* of the IdP.
+In OAuth 2.0 and OpenID Connect, a client application (aka *relying party*) acquires tokens from an identity provider (IdP), such as Azure AD, to prove authentication (via ID tokens) and/or authorization (via access tokens). In this relationship, the IdP has the *authority* to issue tokens to the client application, and the client application trusts the *authority* of the IdP.
 
-MSAL is a token acquisition library. An IdP in the relationship above is denoted as **authority** by MSAL. The **authority** parameter in MSAL configuration indicates the URL to request tokens from. This URL points to the service that issues tokens, also called **security token service** (STS). 
-
-MSAL needs to have accurate information about an STS to be able to requests tokens from it (either directly from the network or from the cache later on). To do so, MSAL performs steps to gather the necessary **metadata** for making a token request (*endpoint discovery*). The [Authority.ts](https://azuread.github.io/microsoft-authentication-library-for-js/ref/classes/_azure_msal_common.authority.html) class encapsulates the methods and data required for endpoint discovery, and is instantiated before each token request.
+MSAL is a token acquisition library. The **authority** parameter in MSAL configuration indicates the URL to request tokens from. This URL points to the service that issues tokens, also called as **security token service** (STS). MSAL needs to have accurate information about an STS to be able to requests tokens from it (either directly from the network or from the cache later on). To do so, MSAL performs steps to gather the necessary **metadata** for making a token request (*endpoint discovery*). The [Authority.ts](https://azuread.github.io/microsoft-authentication-library-for-js/ref/classes/_azure_msal_common.authority.html) class encapsulates the methods and data required for endpoint discovery, and is instantiated before each token request.
 
 ## Authority URL
 
 The default authority for MSAL is `https://login.microsoftonline.com/common`. This string is a URL made of two parts: a **domain** (aka *instance*), and a **tenant identifier** (aka *realm*):
 
 ```console
-    https://<domain-of-the-IdP>/<tenant-identifier>
+    https://<domain-of-the-service>/<tenant-identifier>
 ```
 
 > :warning: The canonical authority URL differs in the case of B2C and ADFS. See [Choosing an authority](#choosing-an-authority) below for more.
@@ -21,7 +19,7 @@ The authority URL guides MSAL where to look for the 3 endpoints that are require
 | Endpoint                | Example     | Description |
 |-------------------------|-------------|-------------|
 | `/openid-configuration` | `https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration` | Contains the metadata required for making a token request e.g. supported scopes, claims, request methods, signing keys etc. |
-| `/authorize`            | `https://login.microsoftonline.com/common/oauth2/v2.0/authorize` | Returns an authorization code in the query parameters via the client app's registered redirect URL. Navigation to this endpoint usually prompts the user with a dialog screen by the IdP. Appropriate query parameters can be used to customize this screen. |
+| `/authorize`            | `https://login.microsoftonline.com/common/oauth2/v2.0/authorize` | Returns an authorization code in the query parameters via the client app's registered redirect URI. Navigation to this endpoint usually prompts the user with a dialog screen by the IdP. Appropriate query parameters can be used to customize this screen. |
 | `/token`                | `https://login.microsoftonline.com/common/oauth2/v2.0/token` | The client app sends the authorizaiton code to this endpoint via a POST request and receives an access and an ID token in return. |
 
 > :bulb: Certain OAuth 2.0 grants may skip the authorize endpoint and go directly for the token endpoint, e.g. [OAuth 2.0 Client Credentials Grant](https://docs.microsoft.com/azure/active-directory/develop/v2-oauth2-client-creds-grant-flow)
--- a/lib/msal-common/docs/authority.md
+++ b/lib/msal-common/docs/authority.md
@@ -1,25 +1,27 @@
 # Authority in MSAL
 
-In OAuth 2.0 and OpenID Connect, a client application (aka *relying party*) acquires tokens from an identity provider (IdP), such as Azure AD, to prove authentication (via ID tokens) and/or authorization (via access tokens). When doing so, the client application *relies* or *trusts* on the authority of the IdP with respect to the claims contained in the said tokens.
+In OAuth 2.0 and OpenID Connect, a client application (aka *relying party*) acquires tokens from an identity provider (IdP), such as Azure AD, to prove authentication (via ID tokens) and/or authorization (via access tokens). When doing so, the client application trusts the *authority* of the IdP.
 
-MSAL is a token acquisition library. An IdP in the relationship above is denoted as **authority** by MSAL. The **authority** parameter in MSAL configuration indicates where the service that provides tokens (*aka* secure token service i.e. STS) is located. As such, MSAL needs to have accurate information about this service to be able to acquire tokens (either from the network or from cache). To do so, MSAL performs a step where it gathers the necessary **metadata** for making a token request (*endpoint discovery*) each time. The [Authority.ts](https://azuread.github.io/microsoft-authentication-library-for-js/ref/classes/_azure_msal_common.authority.html) class encapsulates the methods and data required for endpoint discovery, and is instantiated before each token request.
+MSAL is a token acquisition library. An IdP in the relationship above is denoted as **authority** by MSAL. The **authority** parameter in MSAL configuration indicates the URL to request tokens from. This URL points to the service that issues tokens, also called **security token service** (STS). 
 
-## Authority URI
+MSAL needs to have accurate information about an STS to be able to requests tokens from it (either directly from the network or from the cache later on). To do so, MSAL performs steps to gather the necessary **metadata** for making a token request (*endpoint discovery*). The [Authority.ts](https://azuread.github.io/microsoft-authentication-library-for-js/ref/classes/_azure_msal_common.authority.html) class encapsulates the methods and data required for endpoint discovery, and is instantiated before each token request.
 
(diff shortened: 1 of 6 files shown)


lib/msal-common/docs/authority.md
```diff
@@ -0,0 +1,111 @@
+# Authority in MSAL
+
+In OAuth 2.0 and OpenID Connect, a client application (aka *relying party*) acquires tokens from an identity provider (IdP), such as Azure AD, to prove authentication (via ID tokens) and/or authorization (via access tokens). When doing so, the client application *relies* or *trusts* on the authority of the IdP with respect to the claims contained in the said tokens.
```
sameerag: not: Use `relies on` or `trusts the`. Not both.

sameerag: Btw, we never extract authority from claims. We either go the `discovery` route or trust the `authority` user adds to the `knownAuthorities` array. Hence adding `claims` above (I presume you are referring to `iss` claim) may be misleading here.

lib/msal-common/docs/authority.md
```diff
@@ -0,0 +1,111 @@
...
+
+MSAL is a token acquisition library. An IdP in the relationship above is denoted as **authority** by MSAL. The **authority** parameter in MSAL configuration indicates where the service that provides tokens (*aka* secure token service i.e. STS) is located. As such, MSAL needs to have accurate information about this service to be able to acquire tokens (either from the network or from cache). To do so, MSAL performs a step where it gathers the necessary **metadata** for making a token request (*endpoint discovery*) each time. The [Authority.ts](https://azuread.github.io/microsoft-authentication-library-for-js/ref/classes/_azure_msal_common.authority.html) class encapsulates the methods and data required for endpoint discovery, and is instantiated before each token request.
```
sameerag: instead of saying `where the service is located`, you should probably say `is the URL to request STS tokens` from. The official docs say: `a URL that indicates a directory that MSAL can request tokens from` referenced [here](https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-client-application-configuration#:~:text=using%20the%20library-,Authority,Common%20authority%20URLs).

lib/msal-common/docs/authority.md
```diff
@@ -0,0 +1,111 @@
...
+## Choosing an authority
+
+The correct authority URI that you need pass to MSAL is ultimately determined by the domain of the IdP you register your application with. When registering an application, consider:
+
+- Who are the users you want your app to sign-in?
+- What are the web APIs you want your app to acquire tokens for?
+
+### Azure AD
+
+The authority domain for the global Azure AD instance is `login.microsoftonline.com`. This domain has several aliases (e.g. `login.microsoft.com`), and MSAL keeps a record of these. MSAL trusts authority URIs with any of these aliases by default.
```
sameerag: Should we add a link to the [`discovery endpoint`](https://login.microsoftonline.com/common/discovery/instance?api-version=1.1&authorization_endpoint=https%3A%2F%2Flogin.microsoftonline.com%2Fcommon%2Foauth2%2Fv2.0%2Fauthorize) where these aliases are derived from? Also @samuelkubai did some work to cache these locally etc. Links to them would be helpful.

lib/msal-common/docs/authority.md
```diff
@@ -0,0 +1,111 @@
...
+
+> :warning: The authority domain differs for national Azure deployments, such as Azure China. See [National clouds](https://docs.microsoft.com/azure/active-directory/develop/authentication-national-cloud) for more.
```
sameerag, line 47: We introduced a new `AzureCloudInstance` which also is another i/p for authority. The docs/tests for this are currently work-in-progress here: #4502. Let us make sure I push those changes before this PR, and you can refer them here.

lib/msal-node/docs/configuration.md
```diff
@@ -68,7 +68,7 @@ const msalInstance = new PublicClientApplication(msalConfig);
...
-| `authority` | URI of the tenant to authenticate and authorize with. Usually takes the form of `https://{uri}/{tenantid}`. | String in URI format with tenant - `https://{uri}/{tenantid}` | `https://login.microsoftonline.com/common` |
+| `authority` | URI of the tenant to authenticate and authorize with. Usually takes the form of `https://{uri}/{tenantid}` (see [Authority](../../msal-common/docs/authority.md)) | String in URI format with tenant - `https://{uri}/{tenantid}` | `https://login.microsoftonline.com/common` |
 | `knownAuthorities` | An array of URIs that are known to be valid. Used in B2C scenarios. | Array of strings in URI format | Empty array `[]` |
 | `cloudDiscoveryMetadata` | A string containing the cloud discovery response. Used in AAD scenarios. See performance.md for more info | string | Empty string `""` |
 | `protocolMode` | Enum representing the protocol mode to use. If `"AAD"`, will function on the AAD v2 endpoints; if `"OIDC"`, will function on OIDC-compliant endpoints. | string | `"AAD"` |
```
sameerag, line 74: `AzureCloudInstance` should be added here. Details linked above.

sameerag (commented):
I added a few comments. In addition, we also have a doc MSAL teams are working on to clarify the `authority` definition and initialization across the board. Please check that (sent to you offline) and add any relevant info here.

lib/msal-common/docs/authority.md
derisen: Sure thing, added a remark

lib/msal-common/docs/authority.md
derisen: Sounds good, I'll keep an eye on the PR

lib/msal-common/docs/authority.md
derisen: done, revised to prevent confusion

lib/msal-common/docs/authority.md
derisen: I wasn't aware of this article at all! It seems generic enough to apply to all MSALs with some changes -@mmacy? We could then shorten this doc to discuss MSAL.js internals/technical details only.

tnorling (approved):
Long overdue, thanks for putting this together!

lib/msal-node/docs/configuration.md
mmacy: @derisen Is this information in the TypeDocs?

lib/msal-common/docs/authority.md
mmacy: @derisen Most of this seems like content that could (should) go in the TypeDocs. I won't go so far as to say "Don't publish this," but I'd like to at least explore using this as a real-world use case for showing the team how we can improve our libraries' documentation by putting all (or most) of this stuff in the TypeDocs.

derisen: I'm not sure if we can do without some sort of a wiki in the repo (apart from TypeDocs), but at any rate I can't decide on that effort cc @EmLauber @peterzenz . I think however this doc has a lot of overlap with the [MSAL.NET configuration article on ms.docs](https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-client-application-configuration), and if we can generalize that article to other MSALs (I'll be happy to help with that), we can cut this one much shorter. I'm happy as long as we have something to point to for customers (rightfully) complaining about the `authority` parameter.

lib/msal-node/docs/configuration.md
derisen: There is the [configuration](https://azuread.github.io/microsoft-authentication-library-for-js/ref/modules/_azure_msal_node.html#configuration) entry but I'm afraid it's not as expansive.

lib/msal-common/docs/authority.md
mmacy: > I'm not sure if we can do without some sort of a wiki in the repo (apart from TypeDocs)

OK, disregard. It's pretty clear that skepticism like yours will remain until I dress up a member with full docs to show how it's done.

mmacy: > any rate I can't decide on that effort 

I will point out, however, that if we put it in the TypeDocs, it would be **less** effort, not more.

jasonnutter: We do need to make our typedocs better, and certainly move all property-level docs to those. But long-form documentation will probably need to be separate.

mmacy: Challenge accepted.

peterzenz (approved):

lib/msal-common/docs/authority.md
```diff
@@ -0,0 +1,113 @@
+# Authority in MSAL
+
+In OAuth 2.0 and OpenID Connect, a client application (aka *relying party*) acquires tokens from an identity provider (IdP), such as Azure AD, to prove authentication (via ID tokens) and/or authorization (via access tokens). When doing so, the client application trusts the *authority* of the IdP.
```
mmacy: Recommend not using the term _relying party_ unless you're specifically referring only to OIDC. Since the definition differs between OAuth and OIDC and SAML also refers to a relying party, it's probably safer to stick with "client" and then bring in the _authorization server_.

Maybe lean on the authorization server being the _issuer_ of the tokens--it has the _authority_ to issue those security tokens to client application.

derisen: Removed "relying party" and slightly revised the paragraph, but kept the "IdP" instead of "AS", since we frequently use the term when responding to issues and it's somewhat more commonly used than AS in general.

PR #5084, opened 2022-08-06
Update issue templates & bot to separate public and confidential client flows

tnorling:
- Adds Public/Confidential client dropdown to issue templates
- Adds applicable label to new issues
- Assigns issues to appropriate team
- Updates CODEOWNERS

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -65,6 +65,7 @@ body:
   attributes:
     label: Public or Confidential Client?
     description: Are you using PublicClientApplication or ConfidentialClientApplication? (Issues regarding msal-browser, msal-react and msal-angular are always "Public")
+    multiple: true
     options: 
       - "Public"
       - "Confidential"
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -60,6 +60,16 @@ body:
     placeholder: "2.0.0-beta.0"
   validations:
     required: true
+
+- type: dropdown
+  attributes:
+    label: Public or Confidential Client?
+    description: Are you using PublicClientApplication or ConfidentialClientApplication? (Issues regarding msal-browser, msal-react and msal-angular are always "Public")
+    options: 
+      - "Public"
+      - "Confidential"
+  validations:
+    required: true
     
 - type: textarea
   attributes:
--- a/.github/issue_template_bot.json
+++ b/.github/issue_template_bot.json
@@ -47,9 +47,19 @@
                     "searchStrings": ["ADFS"]
                 }
             }
+        },
+        "Public or Confidential Client?": {
+            "labels": {
+                "public-client": {
+                    "searchStrings": "Public"
+                },
+                "confidential-client": {
+                    "searchStrings": "Confidential"
+                }
+            }
         }
     },
-    "ignoreIssuesOpenedBefore": "2021-04-06",
+    "ignoreIssuesOpenedBefore": "2022-08-05",
     "enforceTemplate": true,
     "templateEnforcementLabel": "more-information-needed",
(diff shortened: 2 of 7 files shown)


.github/issue_template_bot.json
```diff
@@ -47,9 +47,19 @@
...
+                    "searchStrings": "Public"
+                },
+                "confidential-client": {
+                    "searchStrings": "Confidential"
+                }
+            }
         }
     },
-    "ignoreIssuesOpenedBefore": "2021-04-06",
+    "ignoreIssuesOpenedBefore": "2022-08-05",
```
tnorling: Note to update this before merging

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -60,6 +60,16 @@ body:
...
   validations:
     required: true
+
+- type: dropdown
+  attributes:
+    label: Public or Confidential Client?
+    description: Are you using PublicClientApplication or ConfidentialClientApplication? (Issues regarding msal-browser, msal-react and msal-angular are always "Public")
+    options: 
+      - "Public"
+      - "Confidential"
```
jasonnutter, line 71: Can there be a both option? Or n/a?

tnorling: We can do multi-select, which I added to the other templates. I felt that a bug should be more specific to one or the other but I can add multi-select here as well if we think that's valid

peterzenz: Picking one would be helpful as if we allow "both" who would we assign it to for the initial review/investigation?

CODEOWNERS
```diff
@@ -19,23 +19,23 @@
 /samples/msal-core-samples/ @jasonnutter @sameerag @tnorling @peterzenz
 
 # MSAL Node
-/lib/msal-node/ @sameerag @hectormmg @samuelkubai @bmahall @peterzenz
```
peterzenz, line 22: Thanks for the cleanup here as well.

peterzenz (approved):

.github/ISSUE_TEMPLATE/bug_report.yml
tnorling: Assign it to both teams and whoever sees it first can unassign as appropriate?

jasonnutter: I think multi-select (and having it be optional) is best, as there could be issues that affect both or none.

tnorling: Can you provide an example of an issue that would affect neither Public or Confidential? Not a fan of making this optional as it will result in many people not selecting anything even in cases where they should.

jasonnutter: Issue installing MSAL Node itself, for example.

tnorling: IMO that type of issue gets opened so rarely I'm willing to accept the tradeoff of being forced to pick one or both in that scenario in order to ensure all the other issues coming in are being filled out completely.

jasonnutter: Sure, fair enough.

jasonnutter (approved):

CODEOWNERS
```diff
@@ -19,23 +19,23 @@
...
-/samples/msal-node-samples/ @sameerag @hectormmg @samuelkubai @bmahall @peterzenz
-/extensions/msal-node-extensions/ @sameerag @hectormmg @samuelkubai @bmahall @peterzenz
+/lib/msal-node/ @sameerag @hectormmg @bmahall @peterzenz @jennyf19 @jmprieur
+/samples/msal-node-samples/ @sameerag @hectormmg @bmahall @peterzenz @jennyf19 @jmprieur
+/extensions/msal-node-extensions/ @sameerag @hectormmg @bmahall @peterzenz @jennyf19 @jmprieur
```
jasonnutter: @tnorling Might be good to add @bgavrilMS at part of this PR (see #5100)

PR #5098, opened 2022-08-11
Add additional telemetry fields for native broker flows

tnorling:
Adds additional telemetry fields for native broker flows


lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -376,6 +407,7 @@ export class NativeInteractionClient extends BaseInteractionClient {
     private getMATSFromResponse(response: NativeResponse): MATS|null {
         if (response.properties.MATS) {
             try {
+                // start the perf measurement
```
```suggestion
```

lib/msal-common/src/telemetry/performance/PerformanceEvent.ts
```diff
@@ -291,5 +296,24 @@ export type PerformanceEvent = {
      *
      * @type {?boolean}
      */
-    isNativeBroker?: boolean
+    isNativeBroker?: boolean,
+
+    /**
+     * The following are fields that may be emitted in native broker scenarios
```
peterzenz, line 302: Suggest changing the names to hint that they are from mats.

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -88,7 +104,22 @@ export class NativeInteractionClient extends BaseInteractionClient {
                     success: false,
                     errorCode: error.errorCode,
                     subErrorCode: error.subError,
-                    isNativeBroker: true
+                    isNativeBroker: true,
```
peterzenz, line 107: Please ask Anthony to add an indicator as to whether the device is running on battery and adding that to the mats data they present and we log.  Their team did earlier research to indicate token requests were much slower when devices were on battery.

peterzenz (changes requested):

lib/msal-common/src/telemetry/performance/PerformanceEvent.ts
tnorling: Done

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
tnorling: Will add this once it's been added to the mats response

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -88,7 +104,22 @@ export class NativeInteractionClient extends BaseInteractionClient {
...
+                    mats_device_join: mats ? mats.device_join : undefined,
+                    mats_prompt_behavior: mats ? mats.prompt_behavior : undefined,
+                    mats_api_error_code: mats ? mats.api_error_code : undefined,
+                    mats_ui_visible: mats ? mats.ui_visible : undefined,
+                    mats_silent_code: mats ? mats.silent_code : undefined,
+                    mats_silent_bi_sub_code: mats ? mats.silent_bi_sub_code : undefined,
+                    mats_silent_message: mats ? mats.silent_message : undefined,
+                    mats_silent_status: mats ? mats.silent_status : undefined,
+                    mats_http_status: mats ? mats.http_status : undefined,
+                    mats_http_event_count: mats ? mats.http_event_count : undefined
```
jasonnutter, line 122: This looks like a submeasurement, and the only property pulled from submeasurements and included in the top level measurement is `durationMs`. Can this logic be moved up to where acquireTokenSilent, etc is called?

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
```diff
@@ -271,12 +271,14 @@ export abstract class PerformanceClient implements IPerformanceClient {
...
                         }
-                        if (current.accessTokenSize) {
-                            previous.accessTokenSize = current.accessTokenSize;
-                        }
-                        if (current.idTokenSize) {
-                            previous.idTokenSize = current.idTokenSize;
-                        }
+
+                        // Attach any additional submeasurement fields to the event to be emitted
+                        Object.keys(current).forEach(key => {
```
jasonnutter, line 276: While this would make things easier, it could make all events much larger in size, theoritcally.

jasonnutter: Also, when done this way, it is not possible to determine which properties came from which measurements (which is why the durationMs prefixes with the name of the submeasurement).

jasonnutter: I'd rather see an allow list of properties, and when needed, have them prefixed accordingly.

lib/msal-common/src/telemetry/performance/PerformanceEvent.ts
```diff
@@ -291,5 +296,24 @@ export type PerformanceEvent = {
...
+    mats_device_join?: string;
+    mats_prompt_behavior?: string;
+    mats_pi_error_code?: number;
+    mats_ui_visible?: boolean;
+    mats_silent_code?: number;
+    mats_silent_bi_sub_code?: number;
+    mats_silent_message?: string;
+    mats_silent_status?: number;
+    mats_http_status?: number
+    mats_http_event_count?: number;
```
jasonnutter, line 318: Do these need to be snake_cased? Or can they match the casing of the other properties?

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
tnorling: > While this would make things easier, it could make all events much larger in size, theoritcally.

Intuitively speaking I think it makes sense for any fields included on an event to be emitted. If you don't intend for a field to be emitted on the event you shouldn't include it on the submeasurement.

> Also, when done this way, it is not possible to determine which properties came from which measurements (which is why the durationMs prefixes with the name of the submeasurement).

Prefixing makes sense for perf measurements but I'm not so sure the specific submeasurement is important for other data being passed through (and if some sort of distinction is necessary the fields themselves can be named appropriately like these `mats_` fields were)

lib/msal-common/src/telemetry/performance/PerformanceEvent.ts
tnorling: Nope - was originally trying to match the keys coming through from MATS to make implementation cleaner but that's no longer the case so I'll change this

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
jasonnutter: > Prefixing makes sense for perf measurements but I'm not so sure the specific submeasurement is important for other data being passed through (and if some sort of distinction is necessary the fields themselves can be named appropriately like these mats_ fields were)

One example is the `success` property, which for a given request there may be multiple submeasurements that have `success: false`, but the top-level measurement is `success: true` (e.g. `acquireTokenSilent` where the cached RT has expired but we are able successfully acquire a new one).

jasonnutter: The point is that it should be clear for each property which specific operation the value was capture for, and that will not always be the case if all properties are capture without prefixing (or explicit allow-listing for certain well-known properties.

tnorling: So keep the loop but check if each property is in an allow list and if it's not prefix it with the submeasurement name. Does that work?

peterzenz (approved):

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
jasonnutter: I'd rather not add all events by default (even prefixed).

peterzenz: > Intuitively speaking I think it makes sense for any fields included on an event to be emitted. If you don't intend for a field to be emitted on the event you shouldn't include it on the submeasurement.

This feels reasonable to me to add everything that isn't undefined.

PR #5112, opened 2022-08-17
Fix keystore clearing on Firefox

### LINKED ISSUE #5107: Firefox get an error "window.indexedDB.databases is not a function" when user logout
### Core Library

### Core Library Version

2.23.0

### Wrapper Library

### Wrapper Library Version

1.3.2

Public

### Description

When call the ```IPublicClientApplication.logoutRedirect()```, the deletion of MSAL databse https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/be68c190f5561e1a8e7c3144cfee35a5dbc5f109/lib/msal-browser/src/cache/DatabaseStorage.ts#L239 will fail due to ```window.indexedDB.databases is not a function``` in Firefox.

### Error Message

_No response_

### Msal Logs

_No response_

### MSAL Configuration

```javascript
export const msalConfig: Configuration = {
  auth: {
    clientId: IWPAppId,
    redirectUri: `${window.location.origin}/spa-signin-oidc`,
    postLogoutRedirectUri: "/",
  },
  cache: {
  },
  system: {
    // Give more timeout for waiting for a response hash in an iframe or popup other than default value 6s
    iframeHashTimeout: 10000,
    loggerOptions: {
      loggerCallback: (
        level: LogLevel,
        message: string,
        containsPii: boolean
      ): void => {
        if (containsPii) {
          return;
        }

        switch (level) {
          case LogLevel.Error:
            sendConsoleErrorEvent(`@azure/msal-browser ${message}`);
            return;
          case LogLevel.Info:
          case LogLevel.Verbose:
          case LogLevel.Warning:
            sendNormalEvent(
              `@azure/msal-browser ${message}`,
              "web",
              // Use 3 for EventSendType.SyncFetch in const enum isolated module
              shouldSendSync(message) ? 3 : undefined
            );
            return;
        }
      },
      piiLoggingEnabled: false,
    },
  },
};
```

### Relevant Code Snippets

```javascript
instance.logoutRedirect();
```

### Reproduction Steps

1. Signed in user.
2. Call IPublicClientApplication.logoutRedirect() to signout.

### Expected Behavior

The signout will not throw exception and delete all MSAL databses successfully.

### Identity Provider

Azure AD / MSA

Firefox

### Regression

_No response_

### Source

Internal (Microsoft)

hectormmg:
This PR:
- Fixes a bug causing the keystore not to be cleared on logout when no accounts are passed in because the `window.indexedDB.databases()` API hasn't been implemented in Firefox by forcing a deletion of the MSAL IDB database even if it hasn't been confirmed to exist (this may have a minor performance impact on firefox)
- Fixes #5107 
- Reworks the AsyncMemoryStorage clear logic to clear both in-memory stores (asymmetric and symmetric keys) first then both persistent stores (there is only one DB that needs to be deleted)
- Note: These changes cannot be unit tested given there are no ways to properly mock IndexedDB


lib/msal-browser/src/cache/DatabaseStorage.ts
```diff
@@ -234,21 +234,23 @@ export class DatabaseStorage<T> implements IAsyncStorage<T> {
...
-                const deleteDbRequest = window.indexedDB.deleteDatabase(DB_NAME);
-                deleteDbRequest.addEventListener("success", () => resolve(true));
-                deleteDbRequest.addEventListener("error", () => reject(false));
-            });
+
+        try {
+            // @ts-ignore
+            const existingDatabases = await window.indexedDB.databases();
+            const database = existingDatabases.find((database: IDBDatabaseInfo) => database.name === DB_NAME );
+            return (database) ? await this.deleteCurrentDatabase() : true;
```
jasonnutter:
```suggestion
            if (database) {
                return await this.deleteCurrentDatabase();
            }
            return true;
```

jasonnutter: Nitpick, a little simpler to read.

jasonnutter: And a comment explaining why true is being returned as a fallback would be good.

lib/msal-browser/src/cache/DatabaseStorage.ts
```diff
@@ -234,21 +234,23 @@ export class DatabaseStorage<T> implements IAsyncStorage<T> {
...
-            return new Promise<boolean>((resolve: Function, reject: Function) => {
-                const deleteDbRequest = window.indexedDB.deleteDatabase(DB_NAME);
-                deleteDbRequest.addEventListener("success", () => resolve(true));
-                deleteDbRequest.addEventListener("error", () => reject(false));
-            });
+
+        try {
+            // @ts-ignore
+            const existingDatabases = await window.indexedDB.databases();
+            const database = existingDatabases.find((database: IDBDatabaseInfo) => database.name === DB_NAME );
```
jasonnutter: Is this lookup actually needed? It seems like you could blindly invoke `window.indexedDB.deleteDatabase(DB_NAME)` and either the db exists and it gets deleted, or it doesn't exist.

hectormmg: I'll keep fiddling with it to see if there's real perf concerns, I originally did the lookup assuming a write was a significantly more expensive operation than a read and find.

peterzenz: mdn isn't clear on the perf impact either way.  You'd have to inspect the implementations to be certain.  The notes indicate though that the API won't throw in the DB you're trying to delete doesn't exist.  This implies you shouldn't check for existence first.

peterzenz (approved):
Any test changes/impact we should add automation for here?

mchr3k:
I'm interested in this fix being merged. MSAL is still broken on Firefox until this fix gets merged.

jasonnutter (approved):

mchr3k:
@tnorling - Looks like you're the last reviewer outstanding on this PR. Any chance you can take a look at this soon? 🤞🏻

PR #5125, opened 2022-08-22
Added Support for the OIDC max_age parameter

### LINKED ISSUE #4445: Add support for the OIDC `max_age` parameter
### Core Library

### Wrapper Library

Not Applicable

### Description

The ability to force re-authentication by setting the OIDC `max_age` option ([ref](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest)) was added to the MSAL Python library some time ago ([1.15](https://github.com/AzureAD/microsoft-authentication-library-for-python/releases/tag/1.15.0)).

We have a need to FORCE re-authentication on a JS based SPA where a user is carrying out a privileged action (this is highly regulated environment) - could this feature be added to `PublicClientApplication` in parity with MSAL Python?

(BTW see [this article](https://auth0.com/docs/authenticate/login/max-age-reauthentication) to understand why `login=prompt` isn't a good solution).

### Source

External (Customer)

Robbie-Microsoft:
Mirrored existing functionality from MSAL Python: https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/dev/msal/oauth2cli/oidc.py#L215

Issue that spawned this PR: https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/4445


lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -126,6 +126,20 @@ export class ResponseHandler {
...
+            if (request.maxAge) {
+                const authTime = idTokenObj.claims.auth_time;
+
+                if (!authTime) {
+                    throw ClientAuthError.createAuthTimeNotFoundError();
+                }
+
+                const twoMinuteSkew = 120000; // two minutes in milliseconds
+                if ((Date.now() - twoMinuteSkew) > (authTime + request.maxAge)) {
+                    throw ClientAuthError.createMaxAgeTranspiredError();
```
jasonnutter: I wonder if reusing this error would be appropriate? https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dcaef482889239a4fc3da0f0dee56d1551a3ced8/lib/msal-common/src/error/ClientAuthError.ts#L488

Robbie-Microsoft: I see that directly above on line 126, a mismatched nonce has it's own error message. Therefore I believe these max_age errors should have their own error messages as well. Thoughts?

jasonnutter: I would argue that a token needing to be expired to due max age is potentially semantically the same as a natural token expiration error (i.e. they could be handled in the same way), and thus could have the same error (maybe with a different sub error code?).

Robbie-Microsoft: Ok, I can see changing the second error message to the existing tokenRefreshRequired one like you suggested. However, I think I should still have the authTimeNotFoundError message on line 135 in ResponseHandler.ts. Thoughts?

lib/msal-common/src/error/ClientAuthError.ts
```diff
@@ -65,6 +65,14 @@ export const ClientAuthErrorMessage = {
         code: "nonce_not_found",
         desc: "nonce not found"
     },
+    authTimeNotFoundError: {
+        code: "auth_time_not_found",
+        desc: "Max Age was requested, and the ID token is missing the Auth Time variable (auth_time)"
```
jasonnutter: `auth_time` is an optional claim, that is not enabled by default. Can you add a note indicating that this optional claim must be enabled?

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -126,6 +126,20 @@ export class ResponseHandler {
                     throw ClientAuthError.createNonceMismatchError();
                 }
             }
+
+            // token max_age check
+            if (request.maxAge) {
```
jasonnutter: If `maxAge` is `0`, the checks would not run (as `0` is falsey). Is that what we want? Spec is not clear.

lib/msal-common/src/response/ResponseHandler.ts
jasonnutter: Hmm maybe having a different error is fine, curious what others think. And yes, the second error can stay.

jasonnutter: Also, I don't think this is the right place for this logic. This code will only run after a network request, but we would want this to run after retrieving a cache token as well, correct? If so, it should potentially go here? https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dcaef482889239a4fc3da0f0dee56d1551a3ced8/lib/msal-browser/src/app/PublicClientApplication.ts#L195

Robbie-Microsoft: You're right - I forgot about the cache scenario. However the line you suggested still only runs in the case of a network call, inside of acquireSilentTokenAsync (I think). I think the logic should be in two different places. I put it where it currently is in ResponseHandler to mimic the python functionality - I still think it should be there. But I will look into re-using the logic for the cache scenario.

jasonnutter: @Robbie-Microsoft if you renew a token silently using a refresh token, does the `auth_time` reset (you can test by signing in normally, checking the auth_time in the id token, deleting the cached access token from the cache, and calling acquire token silent again, and checking the auth_time in the new id token)? If so, then the logic can be potentially move to here (where we have the existing logic for checking if a cached token has expired): https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dcaef482889239a4fc3da0f0dee56d1551a3ced8/lib/msal-common/src/client/SilentFlowClient.ts#L48-L95

jasonnutter: @Robbie-Microsoft That line runs for both cached and non-cached requests: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dcaef482889239a4fc3da0f0dee56d1551a3ced8/lib/msal-browser/src/app/PublicClientApplication.ts#L190

jasonnutter: And please add tests for cached and non-cached requests.

Robbie-Microsoft: I think we should leave the check where it is in ResponseHandler.ts, but also add it right after [SilentCacheClient.tsL26](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/src/interaction_client/SilentCacheClient.ts#L26). If we move it to where you suggested, then in the case of a network request when the check finally runs, the network request has already been completed. Or has it already been completed at the current point in ResponseHandler.ts anyways? Thoughts?

Robbie-Microsoft: If we take it out of where it's currently located in ResponseHandler.ts, then a CacheRecord will be generated (and other code will run too) even when there is a max_age error.

lib/msal-common/src/response/ResponseHandler.ts
Robbie-Microsoft: According to this: [https://auth0.com/docs/authenticate/login/max-age-reauthentication](https://auth0.com/docs/authenticate/login/max-age-reauthentication)

> To force an immediate re-authentication: If an app requires that a user re-authenticate prior to access, provide a value of 0 for the max_age parameter and the AS will force a fresh login.

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -126,6 +126,20 @@ export class ResponseHandler {
...
+                const authTime = idTokenObj.claims.auth_time;
+
+                if (!authTime) {
+                    throw ClientAuthError.createAuthTimeNotFoundError();
+                }
+
+                const twoMinuteSkew = 120000; // two minutes in milliseconds
```
bgavrilMS: nit: 2 * 60  * 1000 is easier to read, but I'm not sure if the TS transpiler or the JS interpreter will optimize this.

lib/msal-common/src/response/ResponseHandler.ts
jasonnutter: I think thats fine, as long as you create a helper function so that the `auth_time` checking logic isn't duplicated.

jasonnutter: (as a function on the `AuthToken` class could be one place)

lib/msal-common/src/error/ClientAuthError.ts
sameerag: And the aka.ms link for adding optional claims in your token?

lib/msal-common/src/response/ResponseHandler.ts
Robbie-Microsoft: I feel that the comment I left is adequate

Robbie-Microsoft: I can add a comma for readability

lib/msal-common/src/account/AuthToken.ts
```diff
@@ -47,4 +47,27 @@ export class AuthToken {
...
+         */
+        if (maxAge === 0) {
+            throw ClientAuthError.createMaxAgeTranspiredError();
+        }
+
+        if (!authTime) {
+            throw ClientAuthError.createAuthTimeNotFoundError();
+        }
+
+        const twoMinuteSkew = 120000; // two minutes in milliseconds
```
peterzenz: In a number of server flows we've used 5 minute clock skew.  How did we land on 2 minutes here?

lib/msal-common/src/account/AuthToken.ts
```diff
@@ -47,4 +47,27 @@ export class AuthToken {
             throw ClientAuthError.createTokenParsingError(err);
         }
     }
+
+    /**
+     * Determine if the token's max_age has transpired
+     */
+    static checkMaxAge(authTime: number | undefined, maxAge: number): void {
+        /*
+         * per https://auth0.com/docs/authenticate/login/max-age-reauthentication
```
peterzenz: Please link to the official RFC for this and not another vendor's documentation.

lib/msal-common/src/account/AuthToken.ts
```diff
@@ -47,4 +47,27 @@ export class AuthToken {
...
+         * To force an immediate re-authentication: If an app requires that a user re-authenticate prior to access,
+         * provide a value of 0 for the max_age parameter and the AS will force a fresh login.
+         */
+        if (maxAge === 0) {
```
peterzenz: Were there some WAM vs non WAM differences here, @tnorling?

peterzenz (dismissed):

lib/msal-common/src/account/AuthToken.ts
Robbie-Microsoft: I reverse engineered this from MSAL-Python, where a 2 minute clock skew was used. Would you like me to change it to 5 minutes?

lib/msal-common/src/account/AuthToken.ts
tnorling: IIRC WAM did not support max_age=0 - I'll confirm whether that is still the case

lib/msal-common/src/account/AuthToken.ts
Robbie-Microsoft: max_age is not mentioned in the official rfc [here](https://www.ietf.org/rfc/rfc5849.txt), however it is mentioned in the official OIDC spec [here](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest). I will link to the OIDC spec.

lib/msal-common/src/account/AuthToken.ts
tnorling: It has been resolved.

lib/msal-common/src/account/AuthToken.ts
Robbie-Microsoft: The value is 5 minutes in MSAL.NET. I will change it to 5 minutes then.

change/@azure-msal-common-9546872f-4c43-4825-8cfc-58b83d32a2f1.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Added Support for the OIDC max_age parameter",
```
jasonnutter:
```suggestion
  "comment": "Added Support for the OIDC max_age parameter #5125",
```

lib/msal-common/src/account/AuthToken.ts
```diff
@@ -47,4 +47,27 @@ export class AuthToken {
...
+        /*
+         * per https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest
+         * To force an immediate re-authentication: If an app requires that a user re-authenticate prior to access,
+         * provide a value of 0 for the max_age parameter and the AS will force a fresh login.
+         */
+        if (maxAge === 0) {
+            throw ClientAuthError.createMaxAgeTranspiredError();
+        }
+
+        if (!authTime) {
```
tnorling: Feels weird to allow this to be undefined in the function definition but then throw if it's undefined. I'd suggest making this parameter required here and throwing this error at the callsite if it's not defined.

lib/msal-common/src/error/ClientAuthError.ts
```diff
@@ -311,6 +321,23 @@ export class ClientAuthError extends AuthError {
...
+     * Creates an error thrown when max_age was provided in the request, but auth_time is not in the token claims
+     * @param missingNonce
+     */
+    static createAuthTimeNotFoundError(): ClientAuthError {
+        return new ClientAuthError(ClientAuthErrorMessage.authTimeNotFoundError.code,
+            ClientAuthErrorMessage.authTimeNotFoundError.desc);
+    }
+
+    /**
+     * Creates an error thrown when oo much time has elapsed since the last end-user authentication
```
tnorling:
```suggestion
     * Creates an error thrown when too much time has elapsed since the last end-user authentication
```

lib/msal-common/test/client/AuthorizationCodeClient.spec.ts
```diff
@@ -31,6 +31,8 @@ import { ICrypto } from "../../src/crypto/ICrypto";
 import { ClientAuthError } from "../../src/error/ClientAuthError";
 import { CcsCredentialType, ClientConfigurationError } from "../../src";
 
+const ONE_DAY_IN_MILLI = 86400000;
```
tnorling:
```suggestion
const ONE_DAY_IN_MS = 86400000;
```

lib/msal-common/src/account/AuthToken.ts
```diff
@@ -47,4 +47,23 @@ export class AuthToken {
...
+         * per https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest
+         * To force an immediate re-authentication: If an app requires that a user re-authenticate prior to access,
+         * provide a value of 0 for the max_age parameter and the AS will force a fresh login.
+         */
+        if (maxAge === 0) {
+            throw ClientAuthError.createMaxAgeTranspiredError();
+        }
+
+        const twoMinuteSkew = 300000; // five minutes in milliseconds
+        if ((Date.now() - twoMinuteSkew) > (authTime + maxAge)) {
```
tnorling:
```suggestion
        const fiveMinuteSkew = 300000; // five minutes in milliseconds
        if ((Date.now() - fiveMinuteSkew) > (authTime + maxAge)) {
```

tnorling (approved):

lib/msal-common/src/account/AuthToken.ts
Robbie-Microsoft: oops, forgot to change that!

lib/msal-common/src/account/AuthToken.ts
```diff
@@ -47,4 +47,19 @@ export class AuthToken {
...
+    /**
+     * Determine if the token's max_age has transpired
+     */
+    static checkMaxAge(authTime: number, maxAge: number): void {
+        /*
+         * per https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest
+         * To force an immediate re-authentication: If an app requires that a user re-authenticate prior to access,
+         * provide a value of 0 for the max_age parameter and the AS will force a fresh login.
+         */
+        const fiveMinuteSkew = 300000; // five minutes in milliseconds
```
sameerag, line 60: Why is this not a constant?

Robbie-Microsoft: it is? Do you mean named all uppercase?

sameerag (approved):
Approved with a minor comment

PR #5135, opened 2022-08-26
Point to the correct esm file in msal-node-extensions package.json

MilesWellsSVT:
I was using the node package and the extensions for caching in a SvelteKit app. I was attempting to use PKCE to authenticate on the server. Vite's automagic dependency inspector came back with: `Failed to resolve entry for package "@azure/msal-node-extensions". The package may have incorrect main/module/exports specified in its package.json.`

when using the import syntax:

`import { DataProtectionScope, PersistenceCreator, PersistenceCachePlugin } from '@azure/msal-node-extensions';`


jasonnutter (approved):
Good catch, thanks!

jo-arroyo (approved):

tnorling (approved):

Issue #4878, opened 2022-06-08
AAD errors are masked by misleading "network_error" from MSAL node


bgavrilMS:
### Core Library

MSAL Node (@azure/msal-node)

### Core Library Version

Version 1.9.1

### Wrapper Library

Not Applicable

### Wrapper Library Version

none

### Description

This issue is related to MSAL node and can be reproduced with **any** MSAL node sample. I have not tried other MSAL js flavours.

Use for example the Username-Password node sample. 
Configure everything but set **a bad password**. 

## Actual error: 

{"errorCode":"network_error",
"errorMessage":"Network request failed. Please check network trace to determine root cause. | Fetch client threw: Error: HTTP status code 400 | Attempted to reach: https://login.microsoftonline.com/common/oauth2/v2.0/token",
"subError":"",
"name":"ClientAuthError"}

## Expected error: 

Error_code is `invalid_grant`
Error_description is `AADSTS50055: The password is expired`

"{\"error\":\"invalid_grant\",\"error_description\":\"AADSTS50055: The password is expired.\\r\\nTrace ID: e59410c3-5316-43f7-84b9-28e298215600\\r\\nCorrelation ID: f8504f1a-8c13-4894-9af4-b4696246ad92\\r\\nTimestamp: 2022-06-08 13:50:55Z\",\"error_codes\":[50055],\"timestamp\":\"2022-06-08 13:50:55Z\",\"trace_id\":\"e59410c3-5316-43f7-84b9-28e298215600\",\"correlation_id\":\"f8504f1a-8c13-4894-9af4-b4696246ad92\",\"error_uri\":\"https://login.microsoftonline.com/error?code=50055\",\"suberror\":\"user_password_expired\"}"

Note: this message is available in the HttpClient class, but it is wiped out later. 

### Error Message

The network error message is very misleading. It is normal for AAD to reply with 400 Bad Request, but it doesn't indicate a network error. Network errors fall in the 5xx category (and maybe the special 429 Too Many Requests).

### MSAL Configuration

```javascript
just use Username-Password sample with any client_id / tenant_id and set a bad password. Or use the client_credentials sample with a bad scope or a bad secret.
```

### Relevant Code Snippets

```javascript
Username-Password sample
Client_Credentials sample.
```

### Reproduction Steps

See description

### Expected Behavior

1. Error code should be copied from the error from AAD
2. Error message should also be copied from the error from AAD
[...]

PR #4744, opened 2022-04-27
MSAL-Node: Making Network Requests Behind a Proxy

Robbie-Microsoft:
Simulate being behind a corporate proxy:
1. Download and run Fiddler Classic
2. Open Windows Firewall, then Advanced settings -> Properties (in the Actions Pane on the right side).
Block all outbound connections for all profiles you need (domain, private, public) and click OK.
3. Add new outbound firewall rule to allow all access for 8888 port (default Fiddler port) or
"%LOCALAPPDATA%\Programs\Fiddler\Fiddler.exe" app. (I did the latter.
I used "C:\Users\<username>\AppData\Local\Programs\Fiddler\Fiddler.exe")

Source: https://superuser.com/questions/180480/how-to-simulate-corporate-proxy-server-on-my-development-machine

PR #5175, opened 2022-09-06
Node error masking : report underlying AAD error

bmahall:
This PR addresses : #4878 
PR to fix the HttpClient in msal-node , which was masking the underlying AAD server error as a network response error (400), resulting in loss of actual error . 

This was mistakenly introduced in [#4744 ](https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/4744/files).


lib/msal-node/src/network/HttpClient.ts
```diff
@@ -267,7 +272,23 @@ const networkRequestViaHttps = <T>(
                     // do not destroy the request for the device code flow
                     networkResponse.body["error"] !== Constants.AUTHORIZATION_PENDING) {
                     request.destroy();
-                    reject(new Error(`HTTP status code ${statusCode}`));
+                    if (networkResponse.body["error"] === InteractionRequiredAuthErrorMessage.invalid_grant.code) {
```
jasonnutter: I don't think this the right approach. The network client should not have business logic for certain error codes, as that should be handled upstream. The key is that is when the network client receives a server error (e.g. token expired), it should not throw an error, and instead of should return a resolved promise (with the error information in the body of that response) instead of a rejected promise (which is being done here). The network clients should only throw an error when there is a network-level error, not when the server returns an error response.

jasonnutter: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/0a0ca9b006f775309090fafa33fde6a9aae550ad/lib/msal-common/src/client/AuthorizationCodeClient.ts#L71-L83

jasonnutter: i.e. the logic to inspect the error code should likely be removed completely, and only reject when a network connection error is returned (which is what the browser fetch API does).

jasonnutter: And then the upstream code that is calling the network client should be repsonsible for inspecting the response and throwing an error, if needed.

change/@azure-msal-node-b1f6afd4-c1f6-408a-82a9-6d9e8a063035.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Merge branch 'dev' of github.com:AzureAD/microsoft-authentication-library-for-js into node_error_masking",
```
jasonnutter: Please fix with relevant comment about this PR.

jmprieur: I think that the app developers need to have as much information as possible to make decisions about the server errors

bgavrilMS (changes requested):
HttpClient should be responsible for error management.

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -267,7 +272,23 @@ const networkRequestViaHttps = <T>(
...
+                        reject(InteractionRequiredAuthError.createInvalidGrantError(networkResponse.body["error_description"]));
+                    }
+                    else if (networkResponse.body["error"] === InteractionRequiredAuthErrorMessage.noTokensFoundError.code) {
+                        reject(InteractionRequiredAuthError.createNoTokensFoundError());
+                    }
+                    else if (networkResponse.body["error"] === InteractionRequiredAuthErrorMessage.native_account_unavailable.code) {
+                        reject(InteractionRequiredAuthError.createNativeAccountUnavailableError());
+                    }
+                    else if (networkResponse.body["error"] === ClientAuthErrorMessage.invalidRequest.code) {
+                        reject(ClientAuthError.createInvalidRequestError(networkResponse.body["error_description"]));
```
bgavrilMS: These are all ServiceErrors.

lib/msal-common/src/error/ClientAuthError.ts
```diff
@@ -193,6 +193,10 @@ export const ClientAuthErrorMessage = {
     keyIdMissing: {
         code: "key_id_missing",
         desc: "A keyId value is missing from the requested bound token's cache record and is required to match the token to it's stored binding key."
+    },
+    invalidRequest: {
+        code: "invalid_request",
+        desc: "The request body must contain the following parameter: 'username'"
```
bgavrilMS: This is not correct. The description comes from ESTS.

lib/msal-common/src/error/InteractionRequiredAuthError.ts
```diff
@@ -33,6 +33,10 @@ export const InteractionRequiredAuthErrorMessage = {
     native_account_unavailable: {
         code: "native_account_unavailable",
         desc: "The requested account is not available in the native broker. It may have been deleted or logged out. Please sign-in again using an interactive API."
+    },
+    invalid_grant: {
```
bgavrilMS: There is no need to hardcode these in MSAL.

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -267,7 +272,23 @@ const networkRequestViaHttps = <T>(
...
                     networkResponse.body["error"] !== Constants.AUTHORIZATION_PENDING) {
                     request.destroy();
-                    reject(new Error(`HTTP status code ${statusCode}`));
+                    if (networkResponse.body["error"] === InteractionRequiredAuthErrorMessage.invalid_grant.code) {
+                        reject(InteractionRequiredAuthError.createInvalidGrantError(networkResponse.body["error_description"]));
+                    }
+                    else if (networkResponse.body["error"] === InteractionRequiredAuthErrorMessage.noTokensFoundError.code) {
+                        reject(InteractionRequiredAuthError.createNoTokensFoundError());
+                    }
+                    else if (networkResponse.body["error"] === InteractionRequiredAuthErrorMessage.native_account_unavailable.code) {
```
bgavrilMS: These error codes will never come from ESTS. Only `invalid_grant` and `interaction_required` should be transformed into `InteractionRequiredAuthErrorMessage`. I think you have this logic elsewhere in MSAL.

Robbie-Microsoft:
I think that removing the if checks in HttpClient on lines [179](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/src/network/HttpClient.ts#L179) and [266](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/src/network/HttpClient.ts#L266) may be the solution. I had only added those checks because the old axios code (AxiosRequestConfig) had `validateStatus: () => true,`. According to [axios's documentation](https://github.com/axios/axios):

> // `validateStatus` defines whether to resolve or reject the promise for a given
  // HTTP response status code. If `validateStatus` returns `true` (or is set to `null`
  // or `undefined`), the promise will be resolved; otherwise, the promise will be
  // rejected.
  validateStatus: function (status) {
    return status >= 200 && status < 300; // default
  },

Per @jasonnutter's [comment](https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/5175#discussion_r964094154), we are still (and apparently should only be) rejecting the promise for network-level errors on lines [190](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/src/network/HttpClient.ts#L190), [197](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/src/network/HttpClient.ts#L197) and [277](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/src/network/HttpClient.ts#L277).

The original axios code (before I rewrote the HttpClient) can be found [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/4744/files).

lib/msal-node/src/network/HttpClient.ts
bmahall: Thanks for the suggestions @jasonnutter , have updated the HttpClient to not reject based on error code, removing the custom descriptions for the errors;   the second part -"only reject when a network connection error is returned (which is what the browser fetch API does)" is already being handled by the [#3146](https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/3146/files#diff-aa090b4d29f3b8c0c012e0788459864cdc3949a50939fb3da6b93958d473a4c7) as you mentioned as well.

lib/msal-common/src/error/ClientAuthError.ts
bmahall: Thanks for your helpful feedback, removed the custom descriptions @bgavrilMS

lib/msal-common/src/error/ClientAuthError.ts
```diff
@@ -239,7 +239,7 @@ export class ClientAuthError extends AuthError {
      * Creates an error thrown when the id token string is null or empty.
      * @param invalidRawTokenString
      */
-    static createTokenNullOrEmptyError(invalidRawTokenString: string) : ClientAuthError {
+    static createTokenNullOrEmptyError(invalidRawTokenString: string): ClientAuthError {
```
bmahall, line 252: No changes in the file

lib/msal-common/src/error/InteractionRequiredAuthError.ts
```diff
@@ -54,7 +54,7 @@ export class InteractionRequiredAuthError extends AuthError {
      * @param errorString 
      * @param subError 
      */
-    static isInteractionRequiredError(errorCode?: string, errorString?: string, subError?: string) : boolean {
+    static isInteractionRequiredError(errorCode?: string, errorString?: string, subError?: string): boolean {
```
bmahall, line 57: No real changes in this file

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -267,7 +267,7 @@ const networkRequestViaHttps = <T>(
                     // do not destroy the request for the device code flow
                     networkResponse.body["error"] !== Constants.AUTHORIZATION_PENDING) {
                     request.destroy();
-                    reject(new Error(`HTTP status code ${statusCode}`));
+
```
bmahall: removing the early rejection

Robbie-Microsoft: How are you testing this fix? You should also delete line 269, which is why I suggested deleting the entire if statement. Additionally, you should also make the same change to the if statement on line 179.

bmahall: Using the Username-Password sample, as described in the original issue. 
Let's chat further. @Robbie-Microsoft

lib/msal-node/src/network/HttpClient.ts
hectormmg: > I think that removing the if checks in HttpClient on lines [179](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/src/network/HttpClient.ts#L179) and [266](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/src/network/HttpClient.ts#L266) may be the solution. I had only added those checks because the old axios code (AxiosRequestConfig) had `validateStatus: () => true,`. According to [axios's documentation](https://github.com/axios/axios):
> 
> > // `validateStatus` defines whether to resolve or reject the promise for a given
> > // HTTP response status code. If `validateStatus` returns `true` (or is set to `null`
> > // or `undefined`), the promise will be resolved; otherwise, the promise will be
> > // rejected.
> > validateStatus: function (status) {
> > return status >= 200 && status < 300; // default
> > },
> 
> Per @jasonnutter's [comment](https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/5175#discussion_r964094154), we are still (and apparently should only be) rejecting the promise for network-level errors on lines [190](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/src/network/HttpClient.ts#L190), [197](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/src/network/HttpClient.ts#L197) and [277](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/src/network/HttpClient.ts#L277).
> 
> The original axios code (before I rewrote the HttpClient) can be found [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/4744/files).

@Robbie-Microsoft and I discussed this a while back and the issue comes from the added "validation" check that swallows all server errors and reports them as network errors. The confusion came from the axios callback option called "validateStatus" which we set as `validateStatus: () => true` (i.e. hardcoded it to return true so we could resolve with the server error and let the upstream code handle it). This was misinterpreted to mean `validateStatus: true` like we were "turning validation on" so when we replaced the HTTP module logic, a "validation if block" that reported all errors as HTTP errors was added.

As @jasonnutter says, removing that if block should let the service errors bubble-up to the code we already had to handle them.

hectormmg:
> I think that removing the if checks in HttpClient on lines [179](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/src/network/HttpClient.ts#L179) and [266](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/src/network/HttpClient.ts#L266) may be the solution. I had only added those checks because the old axios code (AxiosRequestConfig) had `validateStatus: () => true,`. According to [axios's documentation](https://github.com/axios/axios):
> 
> > // `validateStatus` defines whether to resolve or reject the promise for a given
> > // HTTP response status code. If `validateStatus` returns `true` (or is set to `null`
> > // or `undefined`), the promise will be resolved; otherwise, the promise will be
> > // rejected.
> > validateStatus: function (status) {
> > return status >= 200 && status < 300; // default
> > },
> 
> Per @jasonnutter's [comment](https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/5175#discussion_r964094154), we are still (and apparently should only be) rejecting the promise for network-level errors on lines [190](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/src/network/HttpClient.ts#L190), [197](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/src/network/HttpClient.ts#L197) and [277](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/src/network/HttpClient.ts#L277).
> 
> The original axios code (before I rewrote the HttpClient) can be found [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/4744/files).

To clarify here, the default axios `validateStatus` callback did basic validation (return true if the status was "successful"). Our MSAL axios config effectively "turned off" this validation function by making `validateStatus` always return true (`() => true`), therefore bypassing any form of error validation at the HttpClient level and letting the `ResponseHandler` inspect the response for error codes/messages for handling.

bmahall:
> > I think that removing the if checks in HttpClient on lines [179](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/src/network/HttpClient.ts#L179) and [266](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/src/network/HttpClient.ts#L266) may be the solution. I had only added those checks because the old axios code (AxiosRequestConfig) had `validateStatus: () => true,`. According to [axios's documentation](https://github.com/axios/axios):
> > > // `validateStatus` defines whether to resolve or reject the promise for a given
> > > // HTTP response status code. If `validateStatus` returns `true` (or is set to `null`
> > > // or `undefined`), the promise will be resolved; otherwise, the promise will be
> > > // rejected.
> > > validateStatus: function (status) {
> > > return status >= 200 && status < 300; // default
> > > },
> > 
> > 
> > Per @jasonnutter's [comment](https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/5175#discussion_r964094154), we are still (and apparently should only be) rejecting the promise for network-level errors on lines [190](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/src/network/HttpClient.ts#L190), [197](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/src/network/HttpClient.ts#L197) and [277](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/src/network/HttpClient.ts#L277).
> > The original axios code (before I rewrote the HttpClient) can be found [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/4744/files).
> 
> To clarify here, the default axios `validateStatus` callback did basic validation (return true if the status was "successful"). Our MSAL axios config effectively "turned off" this validation function by making `validateStatus` always return true (`() => true`), therefore bypassing any form of error validation at the HttpClient level and letting the `ResponseHandler` inspect the response for error codes/messages for handling.

AS per our discussion @hectormmg  and the feedback above, the HttpClient is updated.

lib/msal-node/src/network/HttpClient.ts
bmahall: > 

The Username-Password sample is throwing the appropriate server error on commenting out L270 alone. 
Still, I have removed the entire if-block as it was mainly a getaround for validateStatus() and the network errors are already handled upstream. 
We can look at adding a similar if-block for network errors based on the returned error object (Or, status code- 5xx /429) in a separate PR in future , if the distinction is desired. I believe the logic for detecting a network error would be different as compared to Axios client. 
@Robbie-Microsoft @jasonnutter

lib/msal-common/src/error/InteractionRequiredAuthError.ts
```diff
@@ -11,7 +11,8 @@ import { AuthError } from "./AuthError";
 export const InteractionRequiredServerErrorMessage = [
     "interaction_required",
     "consent_required",
-    "login_required"
+    "login_required",
+    "invalid_grant"
```
jasonnutter: This could cause unintended consequences, may want to remove this (from this PR, at least).

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -105,6 +105,7 @@ const networkRequestViaProxy = <T>(
         request.on("connect", (response, socket) => {
             const statusCode = response?.statusCode || 500;
             if (statusCode < 200 || statusCode > 299) {
```
Robbie-Microsoft, line 107: Do we need to also remove this if-statement when connection to the proxy fails? @hectormmg

Robbie-Microsoft: Or maybe just resolve the error?

bmahall: My rationale behind keeping this as such was :  request.on("connect", (response, socket)) is essentially dealing with connection to the proxy alone, all other cases are handled down later, including an error stream on connection to the socket and timeout error. 
Testing with proxy to confirm. 
@Robbie-Microsoft @hectormmg

Robbie-Microsoft: Maybe we can change the error message to include: "Error connecting to proxy: [...]"

lib/msal-common/src/error/InteractionRequiredAuthError.ts
bgavrilMS: It is a fair point though, "invalid_grant" should be translated to `UIRequired` exception. If you remove it, please open a bug / feature request to keep track of it.

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -176,14 +177,6 @@ const networkRequestViaProxy = <T>(
                     status: statusCode as number,
                 };
 
-                if ((statusCode < 200 || statusCode > 299) &&
-                    // do not destroy the request for the device code flow
-                    networkResponse.body["error"] !== Constants.AUTHORIZATION_PENDING) {
```
bgavrilMS, line 181: This condition which relates to Device Code Flow seems to no longer be handled?

context: With device code flow, we do polling until the user completes auth on a different device. I believe AAD returns 400 + authorization_pending error to show that user has not yet completed auth.

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -176,14 +177,6 @@ const networkRequestViaProxy = <T>(
                     status: statusCode as number,
                 };
 
-                if ((statusCode < 200 || statusCode > 299) &&
```
bgavrilMS, line 179: Is it possible to write any unit tests that excercises the code paths in HttpClient?

lib/msal-common/src/error/InteractionRequiredAuthError.ts
bmahall: Will do @bgavrilMS

lib/msal-common/src/error/InteractionRequiredAuthError.ts
bgavrilMS: Thanks for pointing this out, it helps reviewers.

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -103,12 +103,9 @@ const networkRequestViaProxy = <T>(
 
         // establish connection to the proxy
         request.on("connect", (response, socket) => {
+
+            // eslint-disable-next-line @typescript-eslint/no-unused-vars
             const statusCode = response?.statusCode || 500;
```
Robbie-Microsoft: You can delete this statusCode variable since it's not being used anymore.

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -103,13 +103,14 @@ const networkRequestViaProxy = <T>(
 
         // establish connection to the proxy
         request.on("connect", (response, socket) => {
+
```
Robbie-Microsoft: delete this newline

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -103,13 +103,14 @@ const networkRequestViaProxy = <T>(
...
             const statusCode = response?.statusCode || 500;
             if (statusCode < 200 || statusCode > 299) {
                 request.destroy();
                 socket.destroy();
-                reject(new Error(`HTTP status code ${statusCode}`));
+                reject(new Error(response?.statusMessage || "Error in socket connection"));
```
Robbie-Microsoft: I would change this to ``Error connecting to the proxy: ${response?.statusMessage}``, since that's what this is attempting to do.

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -103,13 +103,14 @@ const networkRequestViaProxy = <T>(
...
+
```
Robbie-Microsoft: nit: delete this newline

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -103,13 +103,14 @@ const networkRequestViaProxy = <T>(
...
             const statusCode = response?.statusCode || 500;
             if (statusCode < 200 || statusCode > 299) {
                 request.destroy();
                 socket.destroy();
-                reject(new Error(`HTTP status code ${statusCode}`));
+                reject(new Error(response?.statusMessage || "Error in socket connection"));
+            
             }
-
             if (tunnelRequestOptions.timeout) {
```
Robbie-Microsoft, line 112: nit: add newline here

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -175,15 +176,12 @@ const networkRequestViaProxy = <T>(
                     body: JSON.parse(body) as T,
                     status: statusCode as number,
                 };
-
                 if ((statusCode < 200 || statusCode > 299) &&
```
Robbie-Microsoft, line 177: I thought we agreed to delete this entire if check?

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -267,7 +265,6 @@ const networkRequestViaHttps = <T>(
                     // do not destroy the request for the device code flow
                     networkResponse.body["error"] !== Constants.AUTHORIZATION_PENDING) {
                     request.destroy();
```
Robbie-Microsoft, line 264: I thought we were deleting this if check?

lib/msal-node/test/network/HttpClient.spec.ts
```diff
@@ -15,6 +15,7 @@ jest.mock("https", () => ({
 
 const url: string = "https://www.url.com";
 
+
```
Robbie-Microsoft: nit: delete newline

lib/msal-node/test/network/HttpClient.spec.ts
```diff
@@ -199,16 +200,19 @@ describe("HttpClient", () => {
         });
     });
 
-    describe("Bad Status Code Error", () => {
+
```
Robbie-Microsoft, line 203: nit: delete newline

lib/msal-node/test/network/HttpClient.spec.ts
```diff
@@ -199,16 +200,19 @@ describe("HttpClient", () => {
...
+        const networkResponse: NetworkResponse<T> = getNetworkResponse(mockGetResponseBody, statusCodeError);
 
         describe("Get Request", () => {
             test("Via Https", async () => {
                 (https.request as jest.Mock).mockImplementationOnce(mockHttpsRequest(mockGetResponseBodyBuffer, statusCodeError));
-                await expect(httpClient.sendGetRequestAsync(url)).rejects.toEqual(error);
+                await expect(httpClient.sendGetRequestAsync(url)).resolves.toEqual(networkResponse);
             });
 
+
```
Robbie-Microsoft: nit: delete newline

lib/msal-common/src/error/ClientAuthError.ts
```diff
@@ -554,4 +554,5 @@ export class ClientAuthError extends AuthError {
     static createKeyIdMissingError(): ClientAuthError {
         return new ClientAuthError(ClientAuthErrorMessage.keyIdMissing.code, ClientAuthErrorMessage.keyIdMissing.desc);
     }
+
```
Robbie-Microsoft: nit: delete newline

lib/msal-common/src/error/InteractionRequiredAuthError.ts
```diff
@@ -78,4 +78,5 @@ export class InteractionRequiredAuthError extends AuthError {
     static createNativeAccountUnavailableError(): InteractionRequiredAuthError {
         return new InteractionRequiredAuthError(InteractionRequiredAuthErrorMessage.native_account_unavailable.code, InteractionRequiredAuthErrorMessage.native_account_unavailable.desc);
     }
+
```
Robbie-Microsoft: nit: delete newline

lib/msal-common/src/network/NetworkManager.ts
```diff
@@ -42,6 +42,7 @@ export class NetworkManager {
                 throw e;
             } else {
                 throw ClientAuthError.createNetworkError(tokenEndpoint, e);
+
```
Robbie-Microsoft: nit: delete newline

lib/msal-node/src/network/HttpClient.ts
bmahall: No, this is maintained as the check for the device code flow condition is required, and also since this is at the very end of the logic, it makes sense to maintain this to avoid unintended consequences. 
You can refer to @bgavrilMS 's comment regarding this. Let me know if there are any further questions, thanks

lib/msal-node/src/network/HttpClient.ts
bmahall: @Robbie-Microsoft I believe that only not rejecting while keeping this check intact produces the desired behavior, are there any specific conditions you foresee where this would result in premature termination/ masking of underlying error?

bgavrilMS (approved):

lib/msal-node/test/network/HttpClient.spec.ts
```diff
@@ -199,44 +199,48 @@ describe("HttpClient", () => {
         });
     });
 
-    describe("Bad Status Code Error", () => {
+
+    describe("Bad Status Code Error", <T>() => {
```
jasonnutter:
```suggestion
    describe("Bad Status Code Error", () => {
```

Don't think this change is needed?

change/@azure-msal-node-b1f6afd4-c1f6-408a-82a9-6d9e8a063035.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "This PR modifies HttpClient in Msal-node to return the underlying AAD error in place of wrongfully placed network error, where applicable #5175",
```
jasonnutter: Can this be made shorter?

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -107,9 +107,8 @@ const networkRequestViaProxy = <T>(
             if (statusCode < 200 || statusCode > 299) {
                 request.destroy();
                 socket.destroy();
-                reject(new Error(`HTTP status code ${statusCode}`));
+                reject(new Error(` Error in connection to proxy ${response.statusCode}, ${response?.statusMessage}`));
```
Robbie-Microsoft: nit: `Error connecting to proxy: ${response.statusCode}, ${response?.statusMessage}`

Robbie-Microsoft (approved):

lib/msal-node/test/network/HttpClient.spec.ts
bmahall: Needed for L206

lib/msal-node/test/network/HttpClient.spec.ts
```diff
@@ -199,44 +199,48 @@ describe("HttpClient", () => {
...
         const statusCodeError: number = 500;
-        const error: Error = new Error(`HTTP status code ${statusCodeError}`);
+        const error: Error = new Error("Error in connection to proxy");
+        const networkResponse: NetworkResponse<T> = getNetworkResponse(mockGetResponseBody, statusCodeError);
```
jasonnutter: Create a local type for this payload:

```typescript
export type MockedMetadataResponse = {
    tenant_discovery_endpoint: string,
    token_endpoint: string,
    authorization_endpoint: string,
    device_authorization_endpoint: string
}
```

Then do this:
```suggestion
        const networkResponse: NetworkResponse<MockedMetadataResponse> = getNetworkResponse<MockedMetadataResponse>(mockGetResponseBody, statusCodeError);
```

lib/msal-node/test/network/HttpClient.spec.ts
jasonnutter: That isn't the right way to pass a type for that. See below.

lib/msal-node/test/network/HttpClient.spec.ts
bmahall: Thanks @jasonnutter !

PR #5188, opened 2022-09-07
Extend AuthenticationHeaderParser to parse claims challenge

derisen:
* Adds method for parsing claims challenges to support CAE, auth context and etc
* Changes access modifiers to allow extending the class

### DIFF
--- a/lib/msal-common/src/request/AuthenticationHeaderParser.ts
+++ b/lib/msal-common/src/request/AuthenticationHeaderParser.ts
@@ -56,11 +56,11 @@ export class AuthenticationHeaderParser {
     }
 
     /**
-     * This method parses the claims value out of the WWW-Authenticate authentication header.
+     * This method parses the claims value (if present) out of the WWW-Authenticate authentication header.
      * See: https://docs.microsoft.com/en-us/azure/active-directory/develop/claims-challenge
      * @returns 
      */
-    getClaims(): string {
+    getClaims(): string | null {
         // Attempt to parse claims from WWW-Authenticate
         const wwwAuthenticate = this.headers[HeaderNames.WWWAuthenticate];
         if (wwwAuthenticate) {
@@ -68,11 +68,9 @@ export class AuthenticationHeaderParser {
             if (wwwAuthenticateChallenges.claims){
                 return wwwAuthenticateChallenges.claims;
             }
-            throw ClientConfigurationError.createInvalidAuthenticationHeaderError(HeaderNames.WWWAuthenticate, "claims challenge is missing.");
         }
 
-        // If header is not present, throw missing headers error
-        throw ClientConfigurationError.createMissingClaimsAuthenticationHeadersError();
+        return null;
     }
 
     /**
--- a/lib/msal-common/src/request/AuthenticationHeaderParser.ts
+++ b/lib/msal-common/src/request/AuthenticationHeaderParser.ts
@@ -8,6 +8,7 @@ import { Constants, HeaderNames } from "../utils/Constants";
 
 type WWWAuthenticateChallenges = {
     nonce?: string,
+    claims?: string
 };
 
 type AuthenticationInfoChallenges = {
@@ -19,7 +20,7 @@ type AuthenticationInfoChallenges = {
  * header challenge values that can be used outside the basic authorization flows.
  */
 export class AuthenticationHeaderParser {
-    private headers: Record<string, string>;
+    protected headers: Record<string, string>;
 
     constructor(headers: Record<string, string>) {
         this.headers = headers;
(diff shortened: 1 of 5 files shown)


lib/msal-common/src/request/AuthenticationHeaderParser.ts
```diff
@@ -54,20 +55,40 @@ export class AuthenticationHeaderParser {
...
+     */
+    getClaims(): string {
+        // Attempt to parse claims from WWW-Authenticate
+        const wwwAuthenticate = this.headers[HeaderNames.WWWAuthenticate];
+        if (wwwAuthenticate) {
+            const wwwAuthenticateChallenges = this.parseChallenges<WWWAuthenticateChallenges>(wwwAuthenticate);     
+            if (wwwAuthenticateChallenges.claims){
+                return wwwAuthenticateChallenges.claims;
+            }
+            throw ClientConfigurationError.createInvalidAuthenticationHeaderError(HeaderNames.WWWAuthenticate, "claims challenge is missing.");
```
bgavrilMS: To me it seems strange to throw exceptions. Exceptions means some error happened. But not having claims in a WWW-Authenticate header is normal. In fact in most cases there won't be any claims.

derisen: @bgavrilMS I would say in this particular context (calling `getClaims`) it makes sense, since as a dev I'm expecting it to return a claims string and would probably appreciate knowing it didn't do so before passing it to where I will need it (i.e. `acquireToken*`). Returning null or empty string could easily get overlooked since `acquireToken` will not warn down the line. There's also the matter of consistency with `getShrNonce`. Open to alternatives though: what would be a good way to notify the dev?

```
try {
   const authHeaderParser = new AuthenticationHeaderParser(response.headers);
   const claimsChallenge = authHeaderParser.getClaims(); 
   // ...
} catch {
  // ...
}
```

bgavrilMS: With SHR I kind of understand. Absence of the SHR implies that the RP does not support POP or that the RP is misconfigured.

Claims and all other parameters are optional. Their absence does not indicate a misconfiguration or an error. 
You can always pass `null` claims back to MSAL and MSAL will ignore it (needs verification, but I think it is the case):

```js
const authHeaderParser = new AuthenticationHeaderParser(response.headers);
const claimsChallenge = authHeaderParser.getClaims(); 

confidentialClientApp.AcquireToken(scopes, claimsChallenge); // if claims is null, MSAL will ignore it.
```

There is also a matter of consistency with WWW-Authenticate header parsing in other languages. The spec is still in progress, but pretty much finalized: https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/6315?path=/WWW-Authenticate/WWWAuthenticate.md&_a=files

derisen: > With SHR I kind of understand. Absence of the SHR implies that the RP does not support POP or that the RP is misconfigured.
> 
> Claims and all other parameters are optional. Their absence does not indicate a misconfiguration or an error. You can always pass `null` claims back to MSAL and MSAL will ignore it (needs verification, but I think it is the case):

It will ignore undefined and empty string for sure, null will cause type error in typescript.

> 
> ```js
> const authHeaderParser = new AuthenticationHeaderParser(response.headers);
> const claimsChallenge = authHeaderParser.getClaims(); 
> 
> confidentialClientApp.AcquireToken(scopes, claimsChallenge); // if claims is null, MSAL will ignore it.
> ```
> 
> There is also a matter of consistency with WWW-Authenticate header parsing in other languages. The spec is still in progress, but pretty much finalized: https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/6315?path=/WWW-Authenticate/WWWAuthenticate.md&_a=files

Ah, I see. Should definitely try to align this with the .NET implementation then. Will update.

lib/msal-common/src/request/AuthenticationHeaderParser.ts
```diff
@@ -54,20 +55,40 @@ export class AuthenticationHeaderParser {
...
+    /**
+     * This method parses the claims value out of the WWW-Authenticate authentication header.
+     * See: https://docs.microsoft.com/en-us/azure/active-directory/develop/claims-challenge
+     * @returns 
+     */
+    getClaims(): string {
+        // Attempt to parse claims from WWW-Authenticate
+        const wwwAuthenticate = this.headers[HeaderNames.WWWAuthenticate];
+        if (wwwAuthenticate) {
+            const wwwAuthenticateChallenges = this.parseChallenges<WWWAuthenticateChallenges>(wwwAuthenticate);
```
bgavrilMS, line 67: As per spec, there can be multiple WWW-Authenticate headers. `getShrNonce` should really only look at the WWW-Authenticate header with auth_scheme `POP` (it currently looks at any scheme). 

For claims however, it's trickier. There could be multiple claims parameters associated with various auth schemes. Getting the "first" one seems wrong. 

```
WWW-Authenticate: Bearer authority=https://lmo.com/123 claims=some_claims
WWW-Authenticate: POP authority=https://lmo.com/123 claims=some_other_claims nonce=n123
```

The spec @trwalke wrote has details on how to treat this case.

derisen: @bgavrilMS this seems to be [the case](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate#syntax). Currently we are treating the header as key-value pairs, and JS does not allow duplicate keys, so the last one overrides. I'll convert this to draft for now since the class seems to need an overhaul. cc @hectormmg

hectormmg: Interesting, I wasn't aware there could be multiple headers with the same name. I agree we'll need to update this to handle multiple headers.

derisen:
Closing this for now

PR #5197, opened 2022-09-12
checking whether interaction is in progress inside interceptor

PramodKumarHK89:
checking whether interaction is in progress. issue #5146

### DIFF
--- a/lib/msal-angular/src/msal.interceptor.ts
+++ b/lib/msal-angular/src/msal.interceptor.ts
@@ -10,22 +10,25 @@ import {
     HttpInterceptor
 } from "@angular/common/http";
 import { Location, DOCUMENT } from "@angular/common";
-import { Observable, EMPTY, of } from "rxjs";
-import { switchMap, catchError } from "rxjs/operators";
+import { Subject,Observable, EMPTY, of } from "rxjs";
+import { filter, takeUntil,switchMap, catchError } from "rxjs/operators";
 import { MsalService } from "./msal.service";
-import { AccountInfo, AuthenticationResult, BrowserConfigurationAuthError, InteractionType, StringUtils, UrlString } from "@azure/msal-browser";
+import { AccountInfo, AuthenticationResult, EventMessage,BrowserConfigurationAuthError, InteractionType, StringUtils, UrlString,InteractionStatus } from "@azure/msal-browser";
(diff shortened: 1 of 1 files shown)


jasonnutter:
Apologies for the delay. With this chage, when interaction is in progress, what happens in the interceptor? Nothing?

Also, please add unit tests, and run `npm run beachball:change` from the root of the repo, complete the prompts, and push the generated change files, thanks!

lib/msal-angular/src/msal.interceptor.ts
```diff
@@ -95,15 +98,26 @@ export class MsalInterceptor implements HttpInterceptor {
...
-            this.authService.getLogger().verbose("Interceptor - error acquiring token silently, acquiring by popup");
-            return this.authService.acquireTokenPopup({ ...authRequest, scopes });
-        }
-        this.authService.getLogger().verbose("Interceptor - error acquiring token silently, acquiring by redirect");
-        const redirectStartPage = window.location.href;
-        this.authService.acquireTokenRedirect({...authRequest, scopes, redirectStartPage });
-        return EMPTY;
+    private acquireTokenInteractively(authRequest: MsalInterceptorAuthRequest, scopes: string[]): Observable<AuthenticationResult> { 
+        var subject = new Subject<AuthenticationResult>();
+        this.msalBroadcastService.inProgress$
```
jasonnutter, line 103: I think potentially a better approach is to have the interceptor class start listening for inProgress events in the constuctor, and main an internal field (i.e. this.currentInteractionStatus), instead of listening only when acquireTokenInteractively is called.

tnorling: Agreed. I also think it would be good to have it call acquireTokenSilent again when inProgress turns back to `None`. This would prevent a scenario where many calls are queued up all invoking their own interaction back to back to back.

jasonnutter:
@PramodKumarHK89 We cannot accept this PR in its current form. Please answer the above questions and feedback, or we will close this PR, thanks.

PramodKumarHK89:
@jasonnutter : Unfortunately, I won't be able to spend time on this due to other priorities. Let me go ahead and close this PR for now and later when I have some bandwidth, I'll accommodate all the suggestions that you guys provided. Thank you so much.

PR #5211, opened 2022-09-14
Include http version in performance telemetry data in msal-browser

bmahall:
This PR logs the http version value for perf telemetry for /token and discovery instance network calls for both popup and redirect .


lib/msal-browser/src/interaction_client/BaseInteractionClient.ts
```diff
@@ -163,10 +163,10 @@ export abstract class BaseInteractionClient {
 
         if (requestAuthority) {
             this.logger.verbose("Creating discovered authority with request authority");
-            return await AuthorityFactory.createDiscoveredInstance(requestAuthority, this.config.system.networkClient, this.browserStorage, authorityOptions);
```
bmahall: adds perf client & fixes formatting

lib/msal-common/src/authority/Authority.ts
```diff
@@ -491,6 +500,10 @@ export class Authority {
                     instanceDiscoveryEndpoint,
                     options
                 );
+            const httpVer = response.headers["x-ms-httpver"];
+            if (httpVer) {
+                this.performanceClient?.setHttpVer(httpVer);
```
bmahall: Reads the http version headers and sets the value

jasonnutter: Could this theoretically change between requests? I don't think we should assume that any property of a given request for a user/session will be static (i.e. this should be set on a per request basis, not per session).

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
```diff
@@ -55,6 +56,7 @@ export abstract class PerformanceClient implements IPerformanceClient {
         this.callbacks = new Map();
         this.eventsByCorrelationId = new Map();
         this.measurementsById = new Map();
+        this.httpVer = "1.1";
```
jasonnutter: I assume this is meant to be a default? Can we assume that if we don't receive the header, that it will be 1.1? And even if that is true now, it will likely not be true in the future.

lib/msal-common/src/authority/Authority.ts
jasonnutter: Instead of attaching this as a property on the performance client, how about creating a submeasurement for the instance discovery metadata call (and potentially other network calls, e.g. token endpoint calls), and then adding this as a property on the top-level measurement? Especially since I figure we would want this information not only for the instanceDiscovery call, but also token endpoint calls).

jasonnutter:
Please also include unit tests (at least in performanceclient) regardless of which approach you end up implementing, thanks!

lib/msal-common/src/authority/Authority.ts
bmahall: Great point @jasonnutter , the http version 2 was enabled by Gateway team on login.microsoftonline.com so I used it to test it out and attached the http version for the discovery metadata call . 
Are we intending to have two http version header columns in perf data in this case , one for /token and another for /authorize? If so, it would make sense to include it as part of the submeasurements.

jasonnutter: We only load the authorize endpoint in an iframe, where I don't think we would be able to inspect response headers, so it would be for the metadata endpoints and token endpoint.

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
peterzenz: Anything not explicitly set should default to undefined or should not be reported in our telemetry.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -135,8 +135,8 @@ export class PublicClientApplication extends ClientApplication implements IPubli
                         accessTokenSize: result.accessToken.length,
                         idTokenSize: result.idToken.length,
                         isNativeBroker: result.fromNativeBroker,
-                        cacheLookupPolicy: request.cacheLookupPolicy,
```
jo-arroyo: Is there a reason this is being removed?

lib/msal-common/test/client/AuthorizationCodeClient.spec.ts
```diff
@@ -2012,12 +2107,12 @@ describe("AuthorizationCodeClient unit tests", () => {
 
             const logoutUri = client.getLogoutUri({
                 correlationId: RANDOM_TEST_GUID,
-                postLogoutRedirectUri: 
-                TEST_URIS.TEST_LOGOUT_URI,
+                postLogoutRedirectUri:
+                    TEST_URIS.TEST_LOGOUT_URI,
```
jo-arroyo: Why is this on a new line?

lib/msal-common/src/authority/Authority.ts
```diff
@@ -309,7 +311,7 @@ export class Authority {
 
             metadataEntity.updateEndpointMetadata(metadata, true);
             return AuthorityMetadataSource.NETWORK;
-        }
```
peterzenz: In general refactoring/cleanup is always appreciated, but I'd recommend you create a separate PR just for stylistic or refactoring changes.  It's easier to review changes when they're broken apart and the review can either focus on ensuring there are no functional changes (for cleanup/refactoring) or the specific functional changes otherwise.

lib/msal-browser/src/interaction_client/BaseInteractionClient.ts
```diff
@@ -163,10 +163,10 @@ export abstract class BaseInteractionClient {
...
+            return await AuthorityFactory.createDiscoveredInstance(requestAuthority, this.config.system.networkClient, this.browserStorage, authorityOptions, undefined);
```
jo-arroyo: Why is this `undefined` now needed here?

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -84,17 +84,8 @@ export class AuthorizationCodeClient extends BaseClient {
...
-            reqTimestamp, 
-            request, 
-            authCodePayload,
-            undefined,
-            undefined,
-            undefined,
-            requestId
-        );
+        const httpVer = response.headers?.[HeaderNames.X_MS_HTTP_VERSION] || undefined;
+        return await responseHandler.handleServerTokenResponse(response.body, this.authority, reqTimestamp, request, authCodePayload, undefined, undefined, undefined, httpVer, requestId);
```
peterzenz: With long lists of parameters, the format of one line per parameter is generally more readable.  It also gives some flexibility if you want to comment on parameters.
```
undefined,   // thing with stuff
undefined,   // other thing
undefined,   // yet another thing...
```

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -39,10 +39,10 @@ export class RefreshTokenClient extends BaseClient {
     public async acquireToken(request: CommonRefreshTokenRequest): Promise<AuthenticationResult> {
         const reqTimestamp = TimeUtils.nowSeconds();
         const response = await this.executeTokenRequest(request, this.authority);
-
+        const httpVer= response.headers?.[HeaderNames.X_MS_HTTP_VERSION];
```
peterzenz: Nit:  missing space here `httpVer=`

lib/msal-common/test/client/AuthorizationCodeClient.spec.ts
```diff
@@ -154,7 +154,7 @@ describe("AuthorizationCodeClient unit tests", () => {
             const client = new AuthorizationCodeClient(config);
             const testAccount = TEST_ACCOUNT_INFO;
             // @ts-ignore
-            const testTokenClaims: Required<Omit<TokenClaims, "home_oid"|"upn"|"cloud_instance_host_name"|"cnf"|"emails"|"login_hint">> = {
```
peterzenz: This is a great example of why you should do major cleanup/refactoring in a separate PR, GitHub decided not to render this diff by default as it's too large for this file.

peterzenz (approved):

lib/msal-browser/src/app/PublicClientApplication.ts
bmahall: @jo-arroyo Thanks so much for pointing this out.

lib/msal-browser/src/interaction_client/BaseInteractionClient.ts
bmahall: Not needed, addressed.

lib/msal-common/test/client/AuthorizationCodeClient.spec.ts
```diff
@@ -1943,6 +1943,56 @@ describe("AuthorizationCodeClient unit tests", () => {
...
+                code: authCodeRequest.code,
+                nonce: idTokenClaims.nonce
+            });
+            if (!authenticationResult.expiresOn) {
+                throw TestError.createTestSetupError("mockedAccountInfo does not have a value");
+            }
+
+            expect(authenticationResult.requestId).toBeTruthy;
+            expect(authenticationResult.requestId).toEqual(CORS_RESPONSE_HEADERS.xMsRequestId);
+            console.log(authenticationResult);
```
jasonnutter: Remove
```suggestion
```

jasonnutter:
@bmahall The response handling for `handleRedirectPromise` could also be updated with this value, I don't see that included here.

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -114,6 +114,7 @@ export class ResponseHandler {
         userAssertionHash?: string,
         handlingRefreshTokenResponse?: boolean,
         forceCacheRefreshTokenResponse?: boolean,
+        httpVer ?: string,
```
jasonnutter:
```suggestion
        httpVer?: string,
```

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
```diff
@@ -55,6 +56,7 @@ export abstract class PerformanceClient implements IPerformanceClient {
         this.callbacks = new Map();
         this.eventsByCorrelationId = new Map();
         this.measurementsById = new Map();
+        this.httpVer = "";
```
derisen:
```suggestion
        this.httpVer = "";
```

Should this be an empty string constant instead?

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -189,6 +189,8 @@ export abstract class ClientApplication {
      */
     async handleRedirectPromise(hash?: string): Promise<AuthenticationResult | null> {
         this.logger.verbose("handleRedirectPromise called");
+        // @ts-ignore
+        let handleRedirectPromiseMeasurement:InProgressPerformanceEvent;
```
hectormmg: nit:
```suggestion
        let handleRedirectPromiseMeasurement: InProgressPerformanceEvent;
```

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -242,7 +250,11 @@ export abstract class ClientApplication {
                         this.eventHandler.emitEvent(EventType.LOGIN_FAILURE, InteractionType.Redirect, null, e);
                     }
                     this.eventHandler.emitEvent(EventType.HANDLE_REDIRECT_END, InteractionType.Redirect);
-
+                    handleRedirectPromiseMeasurement?.endMeasurement({
+                        errorCode: e instanceof AuthError && e.errorCode || undefined,
```
hectormmg: Do these result in `e.errorCode` and `e.subError` when they're truthy, or does the `&&` make this statement return `true`?  I would assume by reading this that it would result in `errorCode` being to either `true` or `undefined` instead of the actual error code value.

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
```diff
@@ -3,7 +3,7 @@
  * Licensed under the MIT License.
  */
 
-import { ServerTelemetryManager, CommonAuthorizationCodeRequest, Constants, AuthorizationCodeClient, ClientConfiguration, AuthorityOptions, Authority, AuthorityFactory, ServerAuthorizationCodeResponse, UrlString, CommonEndSessionRequest, ProtocolUtils, ResponseMode, StringUtils, IdTokenClaims, AccountInfo, AzureCloudOptions, PerformanceEvents, AuthError } from "@azure/msal-common";
+import { ServerTelemetryManager, CommonAuthorizationCodeRequest, Constants, AuthorizationCodeClient, ClientConfiguration, AuthorityOptions, Authority, AuthorityFactory, ServerAuthorizationCodeResponse, UrlString, CommonEndSessionRequest, ProtocolUtils, ResponseMode, StringUtils, IdTokenClaims, AccountInfo, AzureCloudOptions, PerformanceEvents, AuthError, IPerformanceClient } from "@azure/msal-common";
```
hectormmg: We should remove the `IPerformanceClient` interface if it's not used.

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
```diff
@@ -199,9 +200,9 @@ export abstract class StandardInteractionClient extends BaseInteractionClient {
      * @param requestAuthority
      * @param requestCorrelationId
      */
-    protected async getDiscoveredAuthority(requestAuthority?: string, requestAzureCloudOptions?: AzureCloudOptions): Promise<Authority> {
+    protected async getDiscoveredAuthority(requestAuthority?: string, requestAzureCloudOptions?: AzureCloudOptions, requestCorrelationId?:string): Promise<Authority> {
```
hectormmg:
```suggestion
    protected async getDiscoveredAuthority(requestAuthority?: string, requestAzureCloudOptions?: AzureCloudOptions, requestCorrelationId?: string): Promise<Authority> {
```

lib/msal-common/src/authority/Authority.ts
```diff
@@ -437,31 +443,52 @@ export class Authority {
      * @param cachedMetadata
      * @param newMetadata
      */
-    private async updateCloudDiscoveryMetadata(metadataEntity: AuthorityMetadataEntity): Promise<AuthorityMetadataSource> {
+    private async updateCloudDiscoveryMetadata(metadataEntity: AuthorityMetadataEntity, correlationId?:string, performanceClient?:IPerformanceClient): Promise<AuthorityMetadataSource> {
```
hectormmg:
```suggestion
    private async updateCloudDiscoveryMetadata(metadataEntity: AuthorityMetadataEntity, correlationId?:string, performanceClient?: IPerformanceClient): Promise<AuthorityMetadataSource> {
```

lib/msal-common/src/authority/Authority.ts
```diff
@@ -530,7 +560,15 @@ export class Authority {
                 metadata,
                 this.hostnameAndPort
             );
+
         } catch (e) {
+            getCloudDiscoveryMetadataFromNetworkMeasurement?.endMeasurement(
+                {
+                    success:false,
```
hectormmg:
```suggestion
                    success: false,
```

lib/msal-common/src/authority/Authority.ts
```diff
@@ -540,6 +578,15 @@ export class Authority {
                 this.hostnameAndPort
             );
         }
+
+        getCloudDiscoveryMetadataFromNetworkMeasurement?.addStaticFields({
+            httpVerCloudMetadata: response?.headers[HeaderNames.X_MS_HTTP_VERSION] || Constants.EMPTY_STRING
+        });
+
+        getCloudDiscoveryMetadataFromNetworkMeasurement?.endMeasurement({
+            success: true,
```
hectormmg: I personally would prefer these to be 1 line if they're only going to have the success attribute. If there's more attributes then multi-line construction does make it more readable.

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
```diff
@@ -197,6 +195,7 @@ export abstract class PerformanceClient implements IPerformanceClient {
      */
     addStaticFields(fields: StaticFields, correlationId: string) : void{
         const existingStaticFields = this.staticFieldsByCorrelationId.get(correlationId);
+        console.log(existingStaticFields);
```
hectormmg:
```suggestion
```

hectormmg (changes requested):

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -189,6 +189,8 @@ export abstract class ClientApplication {
      */
     async handleRedirectPromise(hash?: string): Promise<AuthenticationResult | null> {
         this.logger.verbose("handleRedirectPromise called");
+        // @ts-ignore
```
tnorling: What is this ignoring? Let's try to address the specific warning/error

lib/msal-browser/src/app/ClientApplication.ts
tnorling: This would translate to e.errorCode || undefined. [For more detail on logical operators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_AND#description)

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -140,7 +140,9 @@ export class PublicClientApplication extends ClientApplication implements IPubli
                         success: true,
                         fromCache: result.fromCache,
                         isNativeBroker: result.fromNativeBroker,
-                        requestId: result.requestId
+                        cacheLookupPolicy: request.cacheLookupPolicy,
+                        httpVer: result?.httpVer,
```
tnorling: As discussed please make this a static field

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -209,11 +211,13 @@ export abstract class ClientApplication {
                 let redirectResponse: Promise<AuthenticationResult | null>;
                 if (request && NativeMessageHandler.isNativeAvailable(this.config, this.logger, this.nativeExtensionProvider) && this.nativeExtensionProvider && !hash) {
                     this.logger.trace("handleRedirectPromise - acquiring token from native platform");
+                    handleRedirectPromiseMeasurement = this.performanceClient.startMeasurement(PerformanceEvents.HandleRedirectPromiseMeasurement, request.correlationId);
```
tnorling: Since we aren't already collecting telemetry for redirect calls I'd suggest we discuss whether or not this adds value and consider what else we would want to collect, if any. If we do pick it up let's do so in a separate PR

lib/msal-browser/src/interaction_handler/InteractionHandler.ts
```diff
@@ -122,7 +124,7 @@ export class InteractionHandler {
      */
     protected async updateTokenEndpointAuthority(cloudInstanceHostname: string, authority: Authority, networkModule: INetworkModule): Promise<void> {
         const cloudInstanceAuthorityUri = `https://${cloudInstanceHostname}/${authority.tenant}/`;
-        const cloudInstanceAuthority = await AuthorityFactory.createDiscoveredInstance(cloudInstanceAuthorityUri, networkModule, this.browserStorage, authority.options);
+        const cloudInstanceAuthority = await AuthorityFactory.createDiscoveredInstance(cloudInstanceAuthorityUri, networkModule, this.browserStorage, authority.options, undefined);
```
tnorling: Do we need to include the new params here?

lib/msal-common/src/authority/Authority.ts
```diff
@@ -513,12 +541,14 @@ export class Authority {
          */
 
         let match = null;
+        let response= null;
```
tnorling: Why was this change made? I don't see this being redefined anywhere

lib/msal-common/src/authority/Authority.ts
```diff
@@ -530,7 +560,15 @@ export class Authority {
...
             );
+
         } catch (e) {
+            getCloudDiscoveryMetadataFromNetworkMeasurement?.endMeasurement(
+                {
+                    success:false,
+                }
+              
+            );
+            getCloudDiscoveryMetadataFromNetworkMeasurement?.flushMeasurement();
```
tnorling: We shouldn't flush submeasurements

lib/msal-common/src/authority/Authority.ts
```diff
@@ -540,6 +578,15 @@ export class Authority {
...
         }
+
+        getCloudDiscoveryMetadataFromNetworkMeasurement?.addStaticFields({
+            httpVerCloudMetadata: response?.headers[HeaderNames.X_MS_HTTP_VERSION] || Constants.EMPTY_STRING
+        });
+
+        getCloudDiscoveryMetadataFromNetworkMeasurement?.endMeasurement({
+            success: true,
+        });
+        getCloudDiscoveryMetadataFromNetworkMeasurement?.flushMeasurement();
```
tnorling: Same here

lib/msal-common/src/response/AuthenticationResult.ts
```diff
@@ -42,4 +42,5 @@ export type AuthenticationResult = {
     msGraphHost?: string;
     code?: string;
     fromNativeBroker?: boolean;
+    httpVer?: string;
```
tnorling: Is this something we should be returning or is this just for our Telemetry?

lib/msal-browser/src/app/ClientApplication.ts
sameerag: +1 for separating "redirect" telemetry in a separate PR. I do think it helps to add it, just not here.

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
```diff
@@ -214,17 +215,17 @@ export abstract class StandardInteractionClient extends BaseInteractionClient {
         const userAuthority = requestAuthority ? requestAuthority : this.config.auth.authority;
 
         // fall back to the authority from config
-        const builtAuthority = Authority.generateAuthority( userAuthority, requestAzureCloudOptions || this.config.auth.azureCloudOptions);
+        const builtAuthority = Authority.generateAuthority(userAuthority, requestAzureCloudOptions || this.config.auth.azureCloudOptions);
         this.logger.verbose("Creating discovered authority with configured authority", this.correlationId);
-        return await AuthorityFactory.createDiscoveredInstance(builtAuthority, this.config.system.networkClient, this.browserStorage, authorityOptions)
+        return await AuthorityFactory.createDiscoveredInstance(builtAuthority, this.config.system.networkClient, this.browserStorage, authorityOptions, undefined, requestCorrelationId, this.performanceClient)
             .then((result: Authority) => {
                 getAuthorityMeasurement.endMeasurement({
```
sameerag, line 222: Instead of adding this here, can we rather collect all network related telemetry in the INetworkModule implementation - in this case it will be `XhrClient.ts` and `FetchClient.ts`? Probably will make it centralized and independent of which network call is made.

sameerag (approved):
Left couple comments. 
* Would like to see the "redirect" code handled separately.
* Add telemetry in the `NetworkModule` instead.

sameerag (commented):
Please address the above comments.

sameerag (dismissed):
Updating my approve status - That was a mistake. :)

lib/msal-browser/src/app/ClientApplication.ts
bmahall: Addressed! Thanks

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
bmahall: Do you mean to get rid of StandardInteractionClientGetDiscoveredAuthority submeasurement as a whole from here ? @sameerag

lib/msal-common/src/response/AuthenticationResult.ts
bmahall: this is just for telemetry as discussed.

lalimasharda (approved):

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
bmahall: Keeping it here as discussed.

ghost:

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -72,6 +75,17 @@ export class AuthorizationCodeClient extends BaseClient {
...
         // Retrieve requestId from response headers
         const requestId = response.headers?.[HeaderNames.X_MS_REQUEST_ID];
+        const httpVer = response.headers?.[HeaderNames.X_MS_HTTP_VERSION];
+        atsMeasurement?.addStaticFields({
+            httpVer: httpVer
+        });
+        // cache http version header
+        this.logger.verbose("Cache http version header for auth code ");
+        // CHECK IF cache manager is of type browser cache manager
+        if("setItem" in this.cacheManager){
```
tnorling: hmm not sure I like this...Why do we need to cache the http metadata?

bmahall: 1) for instance discovery call(login.microsoftonline.com):- store the http version 
2) for /token:- check the http version, if different from the instance discovery call, override it. 

Storing it to retain the version across these network calls, and sometimes the http version is not available for /token(due to not being included in ACCESS_CONTROL_EXPOSE_HEADERS). (More:- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers) 
This saved creating the submeasurements solely for the purpose of logging http version, e.g- for redirect flow. 



Alternatively, the http version can be logged as : http_ver_ token (for /token) and http_ver_metadata(for instance discovery metadata call) if cache usage is not desirable/ has performance concerns.

bmahall: Using existing structure & leveraging existing API. 

Side note- Capturing http version for metadata resolution entails either moving the startMeasurement() position for ATS measurement in perf telemetry (or tweaking perf code to allow capturing of initial discovery call -made BEFORE the ats measurement starts, and has a different correlation id), OR using cache/metadata object to store the http version. 
This is outside the scope of this PR.

@tnorling thanks for your helpful feedback !

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
```diff
@@ -199,9 +199,9 @@ export abstract class StandardInteractionClient extends BaseInteractionClient {
      * @param requestAuthority
      * @param requestCorrelationId
      */
-    protected async getDiscoveredAuthority(requestAuthority?: string, requestAzureCloudOptions?: AzureCloudOptions): Promise<Authority> {
+    protected async getDiscoveredAuthority(requestAuthority?: string, requestAzureCloudOptions?: AzureCloudOptions, requestCorrelationId?: string): Promise<Authority> {
         this.logger.verbose("getDiscoveredAuthority called", this.correlationId);
-        const getAuthorityMeasurement = this.performanceClient.startMeasurement(PerformanceEvents.StandardInteractionClientGetDiscoveredAuthority, this.correlationId);
+        const getAuthorityMeasurement = this.performanceClient?.startMeasurement(PerformanceEvents.StandardInteractionClientGetDiscoveredAuthority, requestCorrelationId);
```
tnorling: Why was this changed? What's different between `this.correlationId` and `requestCorrelationId`?

tnorling (commented):
One question otherwise looks good

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
bmahall: no difference

tnorling (approved):

hectormmg (approved):

PR #5233, opened 2022-09-21
Extend msal-browser TokenCache loadExternalTokens to load refresh tokens

lvalentine:
This change extends the `TokenCache.loadExternalTokens` method to support loading refresh tokens. It also makes a few small changes in `TokenCache` to make it easier to use this functionality from an application:
- `TokenCache.loadExternalTokens` now returns a `AuthenticationResult` for the response that was loaded. This allows the calling code to easily get access to the newly loaded account without having to know anything about how msal creates the account id from the Request/Response that was loaded (ie. from `IPublicClientApplication.getAccountByHomeId`).
- Changes the way in which `TokenCache.loadIdToken` generates the `homeAccountId` from the request to match the same method used in `NativeInteractionClient`, using `AccountEntity.generateHomeAccountId`. This change was made so that the tokens that are initially side loaded via the `TokenCache.loadExternalTokens` method can be found in the cache and refreshed when acquiring access tokens via a supported path like `IPublicClientApplication.acquireTokenSilent`.
- Updates to TokenCache tests to account for and cover these changes.

### DIFF
--- a/change/@azure-msal-browser-321fc9fc-e44c-4168-8b90-7b77291857d1.json
+++ b/change/@azure-msal-browser-321fc9fc-e44c-4168-8b90-7b77291857d1.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Extend msal-browser TokenCache to load refresh tokens",
+  "packageName": "@azure/msal-browser",
+  "email": "louisv@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@azure-msal-browser-321fc9fc-e44c-4168-8b90-7b77291857d1.json
+++ b/change/@azure-msal-browser-321fc9fc-e44c-4168-8b90-7b77291857d1.json
@@ -1,6 +1,6 @@
 {
   "type": "minor",
-  "comment": "Extend msal-browser TokenCache to load refresh tokens",
+  "comment": "Extend msal-browser TokenCache to load refresh tokens #5233",
   "packageName": "@azure/msal-browser",
   "email": "louisv@microsoft.com",
   "dependentChangeType": "patch"
--- a/change/@azure-msal-common-4922da53-5379-4871-ab08-329b7ada9d93.json
+++ b/change/@azure-msal-common-4922da53-5379-4871-ab08-329b7ada9d93.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Include refresh token in ExternalTokenResponse",
+  "packageName": "@azure/msal-common",
+  "email": "louisv@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@azure-msal-common-4922da53-5379-4871-ab08-329b7ada9d93.json
+++ b/change/@azure-msal-common-4922da53-5379-4871-ab08-329b7ada9d93.json
@@ -1,6 +1,6 @@
 {
   "type": "minor",
-  "comment": "Include refresh token in ExternalTokenResponse",
+  "comment": "Include refresh token in ExternalTokenResponse #5233",
   "packageName": "@azure/msal-common",
   "email": "louisv@microsoft.com",
   "dependentChangeType": "patch"
--- a/lib/msal-browser/src/cache/ITokenCache.ts
+++ b/lib/msal-browser/src/cache/ITokenCache.ts
@@ -9,6 +9,10 @@ import { LoadTokenOptions } from "./TokenCache";
 
 export interface ITokenCache {
 
-    /** API to side-load tokens to MSAL cache */
-    loadExternalTokens(request: SilentRequest, response: ExternalTokenResponse, options: LoadTokenOptions): void;
+    /**
+     * API to side-load tokens to MSAL cache
+     * @returns The homeAccountId of the account associated with the response.
+     */
+    loadExternalTokens(request: SilentRequest, response: ExternalTokenResponse, options: LoadTokenOptions): string;
+
 }
--- a/lib/msal-browser/src/cache/ITokenCache.ts
+++ b/lib/msal-browser/src/cache/ITokenCache.ts
@@ -3,7 +3,7 @@
  * Licensed under the MIT License.
  */
 
-import { ExternalTokenResponse } from "@azure/msal-common";
+import { ExternalTokenResponse, CacheRecord } from "@azure/msal-common";
 import { SilentRequest } from "../request/SilentRequest";
 import { LoadTokenOptions } from "./TokenCache";
 
@@ -11,8 +11,8 @@ export interface ITokenCache {
 
     /**
      * API to side-load tokens to MSAL cache
-     * @returns The homeAccountId of the account associated with the response.
+     * @returns A `CacheRecord` containing the entities that were loaded.
      */
-    loadExternalTokens(request: SilentRequest, response: ExternalTokenResponse, options: LoadTokenOptions): string;
+    loadExternalTokens(request: SilentRequest, response: ExternalTokenResponse, options: LoadTokenOptions): CacheRecord;
 
 }
--- a/lib/msal-browser/src/cache/ITokenCache.ts
+++ b/lib/msal-browser/src/cache/ITokenCache.ts
@@ -3,7 +3,7 @@
  * Licensed under the MIT License.
  */
 
-import { ExternalTokenResponse, CacheRecord } from "@azure/msal-common";
+import { ExternalTokenResponse, AuthenticationResult } from "@azure/msal-common";
 import { SilentRequest } from "../request/SilentRequest";
 import { LoadTokenOptions } from "./TokenCache";
 
@@ -11,8 +11,8 @@ export interface ITokenCache {
 
     /**
      * API to side-load tokens to MSAL cache
-     * @returns A `CacheRecord` containing the entities that were loaded.
+     * @returns `AuthenticationResult` for the response that was loaded.
      */
-    loadExternalTokens(request: SilentRequest, response: ExternalTokenResponse, options: LoadTokenOptions): CacheRecord;
+    loadExternalTokens(request: SilentRequest, response: ExternalTokenResponse, options: LoadTokenOptions): AuthenticationResult;
 
 }
--- a/lib/msal-browser/src/cache/TokenCache.ts
+++ b/lib/msal-browser/src/cache/TokenCache.ts
@@ -3,7 +3,7 @@
  * Licensed under the MIT License.
  */
 
-import { AccessTokenEntity, ICrypto, IdTokenEntity, Logger, ScopeSet, Authority, AuthorityOptions, ExternalTokenResponse, AccountEntity, AuthToken, RefreshTokenEntity , AuthorityType } from "@azure/msal-common";
+import { AccessTokenEntity, ICrypto, IdTokenEntity, Logger, ScopeSet, Authority, AuthorityOptions, ExternalTokenResponse, AccountEntity, AuthToken, RefreshTokenEntity , AuthorityType, Constants } from "@azure/msal-common";
 import { BrowserConfiguration } from "../config/Configuration";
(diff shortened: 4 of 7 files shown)


lib/msal-browser/src/cache/TokenCache.ts
```diff
@@ -46,17 +46,20 @@ export class TokenCache implements ITokenCache {
      * @param request
      * @param response
      * @param options
+     * @returns The homeAccountId of the account associated with the response.
      */
-    loadExternalTokens(request: SilentRequest, response: ExternalTokenResponse, options: LoadTokenOptions): void {
+    loadExternalTokens(request: SilentRequest, response: ExternalTokenResponse, options: LoadTokenOptions): string {
```
jasonnutter: Instead of just returning a string, I would rather we return an object of some kind (that has homeAccountId as one of its properties, among others), such as `CacheRecord`: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-common/src/cache/entities/CacheRecord.ts#L12

lib/msal-browser/src/cache/TokenCache.ts
```diff
@@ -109,6 +125,8 @@ export class TokenCache implements ITokenCache {
         } else {
             throw BrowserAuthError.createUnableToLoadTokenError("loadExternalTokens is designed to work in browser environments only.");
         }
+
+        return idTokenHomeAccountId;
```
jasonnutter:
```suggestion
        return idTokenEntity;
```

jasonnutter: Then do similar with `loadAccessToken`, `loadRefreshToken`, and then build a `CacheRecord` object from those.

jasonnutter:
@lvalentine Please run `npm run beachball:change` from the root of the repo, complete the prompts, and push up the generated change files, thanks! (this can be marked as a minor bump for msal-browser)

lib/msal-browser/src/cache/TokenCache.ts
lvalentine: Got it

change/@azure-msal-browser-321fc9fc-e44c-4168-8b90-7b77291857d1.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Extend msal-browser TokenCache to load refresh tokens",
```
jasonnutter:
```suggestion
  "comment": "Extend msal-browser TokenCache to load refresh tokens #5233",
```

change/@azure-msal-common-4922da53-5379-4871-ab08-329b7ada9d93.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Include refresh token in ExternalTokenResponse",
```
jasonnutter:
```suggestion
  "comment": "Include refresh token in ExternalTokenResponse #5233",
```

lib/msal-browser/src/cache/TokenCache.ts
```diff
@@ -89,18 +96,27 @@ export class TokenCache implements ITokenCache {
...
-    private loadIdToken(idToken: string, homeAccountId: string, environment: string, tenantId: string, options: LoadTokenOptions): void {
+    private loadIdToken(idToken: string, clientInfo: string, environment: string, tenantId: string, options: LoadTokenOptions, authorityType?: AuthorityType, homeAccountId?: string): CacheRecord {
 
-        const idTokenEntity = IdTokenEntity.createIdTokenEntity(homeAccountId, environment, idToken, this.config.auth.clientId, tenantId);
         const idAuthToken = new AuthToken(idToken, this.cryptoObj);
+        const idTokenHomeAccountId = homeAccountId ?
+            homeAccountId :
+            authorityType !== undefined ?
+                AccountEntity.generateHomeAccountId(clientInfo, authorityType, this.logger, this.cryptoObj, idAuthToken) :
+                clientInfo;
```
jasonnutter: Can this be done without nesting ternaries?

jasonnutter: (I thought we had a lint rule to explicitly disallow this, maybe not)

jasonnutter:
One more comment on style, otherwise looks great, thanks for addressing feedback!

lib/msal-browser/src/cache/TokenCache.ts
lvalentine: Sure, I can write it in a different way.

lib/msal-browser/src/cache/TokenCache.ts
```diff
@@ -72,12 +75,16 @@ export class TokenCache implements ITokenCache {
...
+                const cacheRecord = this.loadIdToken(response.id_token, options.clientInfo, authority.hostnameAndPort, authority.tenant, options, authority.authorityType);
+                this.loadAccessToken(request, response, cacheRecord, authority.hostnameAndPort, authority.tenant, options);
+                this.loadRefreshToken(request, response, cacheRecord, authority.hostnameAndPort);
+                return cacheRecord;
             } else if (response.client_info) {
                 this.logger.trace("TokenCache - homeAccountId from response");
-                this.loadIdToken(response.id_token, response.client_info, authority.hostnameAndPort, authority.tenant, options);
-                this.loadAccessToken(request, response, response.client_info, authority.hostnameAndPort, authority.tenant, options);
+                const cacheRecord = this.loadIdToken(response.id_token, response.client_info, authority.hostnameAndPort, authority.tenant, options, authority.authorityType);
+                this.loadAccessToken(request, response, cacheRecord, authority.hostnameAndPort, authority.tenant, options);
```
hectormmg: Rather than have the `loadIdToken` return a cache record and then mutating the cache record on calls to `loadAccessToken` and `loadRefreshToken`, could you return the individual entities in each API (i.e. IdTokenEntity/AccessTokenEntity/RefreshTokenEntity) and build the cache record once all are returned?

Here's the 2 reasons I think it'd be better:
1. We prefer to avoid methods that mutate their input parameters to avoid unexpected side effects
2.  It seems like initiating the cache record in `loadIdToken` instead of `loadAccessToken` or `loadRefreshToken` is an arbitrary decision that might confuse future maintainers.

lib/msal-browser/src/cache/TokenCache.ts
```diff
@@ -109,19 +129,21 @@ export class TokenCache implements ITokenCache {
...
      * @param response
      * @param options
-     * @param homeAccountId
+     * @param cacheRecord
      * @param environment
      * @param tenantId
      * @returns
      */
-    private loadAccessToken(request: SilentRequest, response: ExternalTokenResponse, homeAccountId: string, environment: string, tenantId: string, options: LoadTokenOptions): void {
+    private loadAccessToken(request: SilentRequest, response: ExternalTokenResponse, cacheRecord: CacheRecord, environment: string, tenantId: string, options: LoadTokenOptions): void {
```
hectormmg: I assume the `SilentRequest` object should contain the account info needed to create the `AccessTokenEntity` later in this method. If not, it'd be better to generate the homeAccountId before generating the entities so one doesn't depend on the other.

hectormmg (changes requested):

lib/msal-browser/src/cache/TokenCache.ts
lvalentine: Yeah I can see that, I was debating doing it that way originally.

lib/msal-browser/src/cache/TokenCache.ts
lvalentine: Right now it generates both the AccountEntity and the IdTokenEntity in the loadIdToken() method. There were already a few different ways for it to be generated, but I didn't fundamentally change any of that.

lvalentine: I created a new `loadAccount` helper function that makes it a bit clearer.

lib/msal-browser/src/cache/TokenCache.ts
lvalentine: I followed this path, let me know how it looks.

lib/msal-browser/src/cache/TokenCache.ts
```diff
@@ -46,17 +46,22 @@ export class TokenCache implements ITokenCache {
...
         if (!response.id_token) {
             throw BrowserAuthError.createUnableToLoadTokenError("Please ensure server response includes id token.");
         }
 
+        const cacheRecord = new CacheRecord();
+
         if (request.account) {
-            this.loadIdToken(response.id_token, request.account.homeAccountId, request.account.environment, request.account.tenantId, options);
-            this.loadAccessToken(request, response, request.account.homeAccountId, request.account.environment, request.account.tenantId, options);
+            cacheRecord.account = this.loadAccount(response.id_token, request.account.environment, undefined, undefined, request.account.homeAccountId);
```
hectormmg: This looks great, if I could offer a final nitpick it'd be to do this in a single statement (or 2 if you need to have the account entity first for the following method calls):

```suggestion
            cacheRecordAccount = this.loadAccount(response.id_token, request.account.environment, undefined, undefined, request.account.homeAccountId);
            
            cacheRecord = {
                        account: cacheRecordAccount,
                        idToken: this.loadIdToken(response.id_token, cacheRecordAccount.homeAccountId, request.account.environment, request.account.tenantId),
                        accessToken: this.loadAccessToken(request, response, cacheRecordAccount.homeAccountId, request.account.environment, request.account.tenantId, options);
                       refreshToken: this.loadRefreshToken(request, response, cacheRecordAccount.homeAccountId, request.account.environment)
                       }
```

Or something similar.

hectormmg (approved):
Looks good to me, just a final optional nit. Thanks for the contribution!

lib/msal-browser/src/cache/TokenCache.ts
lvalentine: updated

jasonnutter (approved):

lib/msal-browser/src/cache/ITokenCache.ts
```diff
@@ -3,12 +3,16 @@
...
 
 export interface ITokenCache {
 
-    /** API to side-load tokens to MSAL cache */
-    loadExternalTokens(request: SilentRequest, response: ExternalTokenResponse, options: LoadTokenOptions): void;
+    /**
+     * API to side-load tokens to MSAL cache
+     * @returns A `CacheRecord` containing the entities that were loaded.
+     */
+    loadExternalTokens(request: SilentRequest, response: ExternalTokenResponse, options: LoadTokenOptions): CacheRecord;
```
tnorling: This makes `CacheRecord` and all of the "Entities" a part of our public API surface, which they were not intended to be. Which particular fields do you need to be returned? I'd be inclined to create a new type which contains what you need rather than exposing internal objects.

tnorling: Perhaps we should use the `AuthenticationResult` type that all of our other APIs return?

lvalentine: For our purposes we want to know which account was used and is encoded in the ExternalTokenResponse, so something like the `AccountInfo` would be enough for us.

jasonnutter: AuthenticationResult (with only relevant fields populated) would work.

lvalentine: Okay I can try that. It looks like the `ResponseHandler.generateAuthenticationResult` helper should populate the correct fields?

lvalentine: Updated. I wasn't able to use `ResponseHandler.generateAuthenticationResult` directly, but I copied the relevant bits for TokenCache to do the same logic.

tnorling (approved):

sameerag (approved):
Approving

PR #5244, opened 2022-09-27
Add x-ms-request-id to perf telemetry logging

jo-arroyo:
This PR:
- Allows retrieval of `x-ms-request-id` (ESTS request ID) from server response headers to be added to `AuthenticationResult` as `requestId`
- These request IDs are then added to existing performance telemetry in msal-browser
- Unit tests for msal-common changes (data point changes added to msal-browser were tested in Kusto)

### DIFF
--- a/lib/msal-common/src/client/AuthorizationCodeClient.ts
+++ b/lib/msal-common/src/client/AuthorizationCodeClient.ts
@@ -71,7 +71,7 @@ export class AuthorizationCodeClient extends BaseClient {
         const response = await this.executeTokenRequest(this.authority, request);
 
         // Retrieve requestId from response headers
-        const requestId = response.headers && response.headers[HeaderNames.X_MS_REQUEST_ID] ? response.headers[HeaderNames.X_MS_REQUEST_ID] : undefined;
+        const requestId = response.headers?[HeaderNames.X_MS_REQUEST_ID];
 
         const responseHandler = new ResponseHandler(
             this.config.authOptions.clientId,
--- a/lib/msal-common/src/client/AuthorizationCodeClient.ts
+++ b/lib/msal-common/src/client/AuthorizationCodeClient.ts
@@ -8,7 +8,7 @@ import { CommonAuthorizationUrlRequest } from "../request/CommonAuthorizationUrl
 import { CommonAuthorizationCodeRequest } from "../request/CommonAuthorizationCodeRequest";
 import { Authority } from "../authority/Authority";
 import { RequestParameterBuilder } from "../request/RequestParameterBuilder";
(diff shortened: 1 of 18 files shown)


peterzenz:
Tests?

jo-arroyo:
> Tests?

Yes! Added.

peterzenz (approved):

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -70,6 +70,9 @@ export class AuthorizationCodeClient extends BaseClient {
         const reqTimestamp = TimeUtils.nowSeconds();
         const response = await this.executeTokenRequest(this.authority, request);
 
+        // Retrieve requestId from response headers
+        const requestId = response.headers && response.headers[HeaderNames.X_MS_REQUEST_ID] ? response.headers[HeaderNames.X_MS_REQUEST_ID] : undefined;
```
hectormmg: I think you can leverage optional chaining here:

```suggestion
        // Retrieve requestId from response headers
        const requestId = response.headers?[HeaderNames.X_MS_REQUEST_ID];
```

If response.headers doesn't exist it'll return `undefined`. Then if it does exist it will either return the `HeaderName.X_MS_REQUEST_ID` header if present or `undefined` if not present.

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -40,6 +40,9 @@ export class RefreshTokenClient extends BaseClient {
         const reqTimestamp = TimeUtils.nowSeconds();
         const response = await this.executeTokenRequest(request, this.authority);
 
+        // Retrieve requestId from response headers
+        const requestId = response.headers && response.headers[HeaderNames.X_MS_REQUEST_ID] ? response.headers[HeaderNames.X_MS_REQUEST_ID] : undefined;
```
hectormmg: Same about optional chaining

hectormmg (approved):
LGTM, just the optional chaining nit

lib/msal-common/src/client/RefreshTokenClient.ts
jasonnutter:
```suggestion
        const requestId = response.headers?[HeaderNames.X_MS_REQUEST_ID];
```

jasonnutter:
> > Tests?
> 
> Yes! Added.

Can you also add an assertion to this test for requestId? https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/test/app/PublicClientApplication.spec.ts#L2633

jo-arroyo:
> Can you also add an assertion to this test for requestId? https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/test/app/PublicClientApplication.spec.ts#L2633

Done!

PR #5246, opened 2022-09-29
Fixed broken homeAccountId check in CacheManager

### LINKED ISSUE #5134: multiple_matching_tokens The cache contains multiple tokens satisfying the requirements. Call AcquireToken again providing more requirements such as authority or account.
### Core Library

MSAL Node (@azure/msal-node)

### Core Library Version

1.12.0

### Wrapper Library

Not Applicable

### Wrapper Library Version

None

Confidential

### Description

After adding 2 new MSAL flows for acquireTokenByRefreshToken and acquireTokenOnBehalfOf we started seeing some errors in existing workflows that had no issues (acquireTokenByClientCredential). The error we are receiving from MSAL is "multiple_matching_tokens". We only see this issue in acquireTokenByClientCredential for our singleton client.

We added `authority` to all of our client codes as suggested by the error message below but we are still receiving these errors. 

We noticed 2 similar incidents that were supposed to fix this issue:
[#4486](https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/4486)
[#4691](https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/4691)

### Error Message

>"stack": "ClientAuthError: multiple_matching_tokens: The cache contains multiple tokens satisfying the requirements. Call AcquireToken again providing more requirements such as authority or account..\n    at ClientAuthError.AuthError [as constructor] (/app/node_modules/@azure/msal-node/node_modules/@azure/msal-common/dist/index.cjs.js:496:24)\n    at new ClientAuthError (/app/node_modules/@azure/msal-node/node_modules/@azure/msal-common/dist/index.cjs.js:790:28)\n    at Function.ClientAuthError.createMultipleMatchingTokensInCacheError (/app/node_modules/@azure/msal-node/node_modules/@azure/msal-common/dist/index.cjs.js:885:16)\n    at ClientCredentialClient.readAccessTokenFromCache (/app/node_modules/@azure/msal-node/node_modules/@azure/msal-common/dist/index.cjs.js:6036:35)\n    at ClientCredentialClient.<anonymous> (/app/node_modules/@azure/msal-node/node_modules/@azure/msal-common/dist/index.cjs.js:5996:50)\n    at step (/app/node_modules/@azure/msal-node/node_modules/@azure/msal-common/dist/index.cjs.js:79:23)\n    at Object.next (/app/node_modules/@azure/msal-node/node_modules/@azure/msal-common/dist/index.cjs.js:60:53)\n    at /app/node_modules/@azure/msal-node/node_modules/@azure/msal-common/dist/index.cjs.js:53:71\n    at new Promise (<anonymous>)\n    at __awaiter (/app/node_modules/@azure/msal-node/node_modules/@azure/msal-common/dist/index.cjs.js:49:12)",
>     "message": "**multiple_matching_tokens**: **The cache contains multiple tokens satisfying the requirements. Call AcquireToken again providing more requirements such as authority or account..**",
>     "errorCode": "multiple_matching_tokens",
>     "errorMessage": "The cache contains multiple tokens satisfying the requirements. Call AcquireToken again providing more requirements such as authority or account..",
>     "subError": "",
>     "name": "ClientAuthError"
>   }

### Msal Logs

_No response_

### MSAL Configuration

```javascript
auth: {
                clientId: "9edfcdd9-0bc5-4bd4-b287-c3afc716aac7"
                authority: "https://login.microsoftonline.com/organizations",
                clientCertificate: {
                    thumbprint: pem.thumbprint,
                    privateKey: pem.privateKey,
                    x5c: pem.certificate,
                },
            },
```

### Relevant Code Snippets

```javascript
this._client.acquireTokenByClientCredential({
  "scopes": [
    "https://management.azure.com/.default"
  ],
  "authority": "https://login.microsoftonline.com/72f988bf-86f1-41af-91ab-2d7cd011db47"
}),
```

### Reproduction Steps

N/A

### Expected Behavior

Expectation would be for `acquireTokenBYClientCredential` to return a valid access token matching the passed in requirements.

### Identity Provider

Azure AD / MSA

None (Server)

### Regression

_No response_

### Source

Internal (Microsoft)

Robbie-Microsoft:
The homeAccountId check in CacheManager.ts did not account for a homeAccountId that was an empty string.

For example, say the following two access tokens exist in the cache for a confidential client application - one for an application and the other for a user account:

```
        "-login.windows.net-accesstoken-[...]-https://graph.microsoft.com/.default--": {
            "home_account_id": "",
            "credential_type": "AccessToken",
            ...
        },
        "[...]-login.windows.net-accesstoken-[...]-profile openid email https://graph.microsoft.com/user.read https://graph.microsoft.com/user.read.all https://graph.microsoft.com/.default--": {
            "home_account_id": "[...]",
            "credential_type": "AccessToken",
            ...
        }
```

Under the old homeAccountId logic, both access tokens were returned when an application called `acquireTokenByClientCredential` because !!("") equates to false, and the rest of the homeAccountId comparison was not being performed. Now, only the matching access token is returned.

### DIFF
--- a/change/@azure-msal-common-f46053c3-36ec-417f-b7a2-250e2a580f76.json
+++ b/change/@azure-msal-common-f46053c3-36ec-417f-b7a2-250e2a580f76.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fixed broken homeAccountId check in CacheManager",
+  "packageName": "@azure/msal-common",
+  "email": "rgins16@gmail.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@azure-msal-common-f46053c3-36ec-417f-b7a2-250e2a580f76.json
+++ b/change/@azure-msal-common-f46053c3-36ec-417f-b7a2-250e2a580f76.json
@@ -2,6 +2,6 @@
   "type": "patch",
   "comment": "Fixed broken homeAccountId check in CacheManager",
   "packageName": "@azure/msal-common",
-  "email": "rgins16@gmail.com",
+  "email": "rginsburg@microsoft.com",
   "dependentChangeType": "patch"
 }
--- a/change/@azure-msal-common-f46053c3-36ec-417f-b7a2-250e2a580f76.json
+++ b/change/@azure-msal-common-f46053c3-36ec-417f-b7a2-250e2a580f76.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Fixed broken homeAccountId check in CacheManager",
+  "comment": "Fixed broken homeAccountId check in CacheManager #5246",
   "packageName": "@azure/msal-common",
   "email": "rginsburg@microsoft.com",
   "dependentChangeType": "patch"
--- a/lib/msal-common/src/cache/CacheManager.ts
+++ b/lib/msal-common/src/cache/CacheManager.ts
@@ -294,7 +294,7 @@ export abstract class CacheManager implements ICacheManager {
             }
 
             // homeAccountId can undefined, and we want to filter out cached items that have a homeAccountId of ""
-            if ((typeof homeAccountId === "string") && !this.matchHomeAccountId(entity, homeAccountId)) {
+            if (!!homeAccountId && !this.matchHomeAccountId(entity, homeAccountId)) {
                 return;
             }
 
--- a/lib/msal-common/src/cache/CacheManager.ts
+++ b/lib/msal-common/src/cache/CacheManager.ts
@@ -293,7 +293,8 @@ export abstract class CacheManager implements ICacheManager {
                 return;
             }
 
-            if (!!homeAccountId && !this.matchHomeAccountId(entity, homeAccountId)) {
+            // homeAccountId can undefined, and we want to filter out cached items that have a homeAccountId of ""
+            if ((typeof homeAccountId === "string") && !this.matchHomeAccountId(entity, homeAccountId)) {
                 return;
             }
 
@@ -390,8 +391,8 @@ export abstract class CacheManager implements ICacheManager {
                 return;
             }
 
-            // homeAccountId can be an empty string, and !!("") = false
-            if (!!(typeof homeAccountId === "string") && !this.matchHomeAccountId(entity, homeAccountId)) {
+            // homeAccountId can undefined, and we want to filter out cached items that have a homeAccountId of ""
+            if ((typeof homeAccountId === "string") && !this.matchHomeAccountId(entity, homeAccountId)) {
                 return;
             }
 
@@ -842,7 +843,7 @@ export abstract class CacheManager implements ICacheManager {
      * @param homeAccountId
      */
     private matchHomeAccountId(entity: AccountEntity | CredentialEntity, homeAccountId: string): boolean {
-        return !!(entity.homeAccountId && homeAccountId === entity.homeAccountId);
+        return !!((typeof entity.homeAccountId === "string") && (homeAccountId === entity.homeAccountId));
     }
 
     /**
--- a/lib/msal-common/src/cache/CacheManager.ts
+++ b/lib/msal-common/src/cache/CacheManager.ts
@@ -293,7 +293,6 @@ export abstract class CacheManager implements ICacheManager {
                 return;
             }
 
-            // homeAccountId can undefined, and we want to filter out cached items that have a homeAccountId of ""
             if (!!homeAccountId && !this.matchHomeAccountId(entity, homeAccountId)) {
                 return;
             }
--- a/lib/msal-common/src/cache/CacheManager.ts
+++ b/lib/msal-common/src/cache/CacheManager.ts
@@ -390,7 +390,8 @@ export abstract class CacheManager implements ICacheManager {
                 return;
             }
 
-            if (!!homeAccountId && !this.matchHomeAccountId(entity, homeAccountId)) {
+            // homeAccountId can be an empty string, and !!("") = false
+            if (!!(typeof homeAccountId === "string") && !this.matchHomeAccountId(entity, homeAccountId)) {
                 return;
             }
 
--- a/lib/msal-common/src/cache/CacheManager.ts
+++ b/lib/msal-common/src/cache/CacheManager.ts
@@ -390,7 +390,10 @@ export abstract class CacheManager implements ICacheManager {
                 return;
             }
 
-            // homeAccountId can undefined, and we want to filter out cached items that have a homeAccountId of ""
+            /*
+             * homeAccountId can undefined, and we want to filter out cached items that have a homeAccountId of ""
+             * because we don't want a client_credential request to return a cached token that has a homeAccountId
+             */
             if ((typeof homeAccountId === "string") && !this.matchHomeAccountId(entity, homeAccountId)) {
                 return;
             }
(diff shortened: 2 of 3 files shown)


lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -390,7 +390,8 @@ export abstract class CacheManager implements ICacheManager {
                 return;
             }
 
-            if (!!homeAccountId && !this.matchHomeAccountId(entity, homeAccountId)) {
+            // homeAccountId can be an empty string, and !!("") = false
+            if (!!(typeof homeAccountId === "string") && !this.matchHomeAccountId(entity, homeAccountId)) {
```
bgavrilMS: Quite subtle, good catch. 

Is it possible to write a regression test for this?

Robbie-Microsoft: I've been working on that

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -390,7 +390,8 @@ export abstract class CacheManager implements ICacheManager {
                 return;
             }
 
-            if (!!homeAccountId && !this.matchHomeAccountId(entity, homeAccountId)) {
+            // homeAccountId can be an empty string, and !!("") = false
```
bgavrilMS: Consider adding a comment more about the WHY and not the WHAT, i.e. 

"In client_credentials, homeAccountId is null, and the logic should filter out tokens with homeAccountId, because these are user tokens"

lib/msal-common/src/cache/CacheManager.ts
jasonnutter: `!!` shouldnt be needed, as `typeof var === "type"` will return a boolean.

change/@azure-msal-common-f46053c3-36ec-417f-b7a2-250e2a580f76.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fixed broken homeAccountId check in CacheManager",
```
jasonnutter:
```suggestion
  "comment": "Fixed broken homeAccountId check in CacheManager #5246",
```

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -390,7 +390,8 @@ export abstract class CacheManager implements ICacheManager {
                 return;
             }
 
-            if (!!homeAccountId && !this.matchHomeAccountId(entity, homeAccountId)) {
+            // homeAccountId can undefined, and we want to filter out cached items that have a homeAccountId of ""
```
sameerag: Cached items that have a homeAccountId as "null" or "empty string"??

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -390,7 +390,8 @@ export abstract class CacheManager implements ICacheManager {
...
+            if ((typeof homeAccountId === "string") && !this.matchHomeAccountId(entity, homeAccountId)) {
```
sameerag, line 397: Why are we checking "type"? Are we expecting data corruption? and should we check for "null" or "empty string" or both?

Robbie-Microsoft: @sameerag homeAccountId can be undefined or empty string. If it's undefined, we're skipping this check. Otherwise we need to account for a client credential application which has a homeAccountId of empty string.

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -390,7 +390,8 @@ export abstract class CacheManager implements ICacheManager {
                 return;
             }
 
-            if (!!homeAccountId && !this.matchHomeAccountId(entity, homeAccountId)) {
```
bgavrilMS, line 393: I don't know JS that well to comment on the topic of "falsy" variables, but the algorithm here should be: 

- if the `homeAccountId` filter is `falsy` (i.e. we're asking for an app token), then only an entity with `homeAccountId` which is `falsy` may be returned.

tnorling: > if the homeAccountId filter is falsy (i.e. we're asking for an app token)

I'm not so sure we can make this assumption. If we want to explicitly ask for a token with no homeAccountId we should explicitly provide `null` or perhaps empty string. This API may be called with any combination of filters and the absence of one just means "don't check".

lib/msal-common/src/cache/CacheManager.ts
Robbie-Microsoft: @sameerag In the [generateHomeAccountId function](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-common/src/cache/entities/AccountEntity.ts#L255) homeAccountId is being set to empty string, in this case for a confidential client application.

lib/msal-common/src/cache/CacheManager.ts
bgavrilMS: Agreed @tnorling. I'm just stating the intent: app tokens do not have home_account_id, user tokens do. A `client_credential` request (i.e. no user is involved) must not retrieve a token from the cache which has a `home_account_id`. I will happily learn how to transform this to js/ts code :)

lib/msal-common/test/client/ClientCredentialClient.spec.ts
```diff
@@ -75,6 +79,45 @@ describe("ClientCredentialClient unit tests", () => {
         expect(returnVal.includes(`${AADServerParamKeys.X_MS_LIB_CAPABILITY}=${ThrottlingConstants.X_MS_LIB_CAPABILITY_VALUE}`)).toBe(true);
     });
 
+    it.only("Multiple access tokens would match, but one of them has a Home Account ID of \"\"", async () => {
```
bgavrilMS: nit: add a link to the bug, saying that it is a regression test for it

bgavrilMS (approved):
LGTM from MSAL logic / E2E test perspective, but please work with @tnorling and @sameerag to get consensus on the "javascriptness"

lib/msal-common/src/cache/CacheManager.ts
sameerag: Can we stick to "null" or "empty_string" - only one of it - and be consistent? I think the check has to be "homeAccountId == <<expected null or empty_string>> for `clientCredential` flow (where we explicitly set it to that value while caching).

lib/msal-common/test/client/ClientCredentialClient.spec.ts
jasonnutter:
```suggestion
    it("Multiple access tokens would match, but one of them has a Home Account ID of \"\"", async () => {
```

Robbie-Microsoft: Yikes! Good catch.

lib/msal-common/src/cache/CacheManager.ts
jasonnutter: I would expand on this, with the details from @bgavrilMS comment above, e.g.:

```suggestion
            // homeAccountId can undefined, and we want to filter out cached items that have a homeAccountId of ""
            // because we don't want a client_credential request to return a cached token that has a home_account_id
```

jasonnutter:
> LGTM from MSAL logic / E2E test perspective, but please work with @tnorling and @sameerag to get consensus on the "javascriptness"

Agree on the need to agree on semantics for `""` versus `null` here.

sameerag (approved):
Synced up offline. The check ensures "empty string" is always matched and "null" isn't, which satisfies the use case here. Approving.

Robbie-Microsoft:
^ Synced with @sameerag and we agreed to leave the check as it is, because we still want to do the comparison even when the homeAccountId is equivalent to empty string. The only time we don't want to do it is when the homeAccountId is null/undefined, and checking for the type accounts for that.

PR #5250, opened 2022-09-30
Include Refresh Token size in perf telemetry data 

bmahall:
This PR adds Refresh Token size fetched from the RefreshTokenClient at time of RT acquisition in case of both cached/unavailable RT to the performance telemetry data . It adds two new submeasurements RefreshTokenClientAcquireToken, RefreshTokenClientAcquireTokenWithCachedRefreshToken to the PerformanceEvent enum .

For discussion , please refer to [#5063](https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/5063). 

Tests included.


lib/msal-common/src/client/BaseClient.ts
```diff
@@ -83,7 +83,7 @@ export abstract class BaseClient {
     /**
      * Creates default headers for requests to token endpoint
      */
-    protected createTokenRequestHeaders(ccsCred?: CcsCredential): Record<string, string> {        
+    protected createTokenRequestHeaders(ccsCred?: CcsCredential): Record<string, string> {
```
bmahall, line 86: No real changes

lib/msal-common/src/client/SilentFlowClient.ts
```diff
@@ -22,7 +22,7 @@ export class SilentFlowClient extends BaseClient {
     constructor(configuration: ClientConfiguration, performanceClient?: IPerformanceClient) {
         super(configuration,performanceClient);
     }
-
+
```
bmahall, line 25: No real changes

jasonnutter (approved):

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -63,7 +65,23 @@ export class RefreshTokenClient extends BaseClient {
             true,
             request.forceCache,
             requestId
-        );
+        ).then((result: AuthenticationResult) => {
+            atsMeasurement?.endMeasurement({
+                success: true,
+                refreshTokenSize: response.body.refresh_token?.length || 0
```
peterzenz, line 71: I thought we had some discussion on cases where no refresh token was returned we'd not return a size of 0 but we wouldn't log that datapoint at all.

peterzenz (approved):
If we get a lot of results with 0 length RTs then we'll likely take the time to address how that data is sent.

hectormmg (approved):

lib/msal-common/src/client/RefreshTokenClient.ts
bmahall: @peterzenz Not logging in either case. 
From the testing point of view, the case where no refresh token is returned is semantically treated as same as having a token of length 0.

PR #5262, opened 2022-10-03
Post release pipeline

hectormmg:
This PR:
- Adds post-release automation for committing release artifacts and creating a GitHub PR
- Updates release pipeline to fix logical bugs in conditional release of dependent packages
- Fixes msal-angular release command
- Fixes typo in CDN secret name


peterzenz (approved):

.pipelines/custom-templates/post-publish-template.yml
```diff
@@ -0,0 +1,53 @@
...
+    inputs:
+      script: |
+        cp $(Pipeline.Workspace)/${{ parameters.libName }}-artifacts/README.md ${{ parameters.path}}/msal-${{ parameters.libName }}/README.md
+        cp $(Pipeline.Workspace)/${{ parameters.libName }}-artifacts/docs/cdn-usage.md ${{ parameters.path}}/msal-${{ parameters.libName }}/docs/cdn-usage.md
+  - task: CmdLine@2
+    displayName: 'Save package.json'
+    condition: succeeded()
+    inputs:
+      script: | 
+        cat $(Pipeline.Workspace)/${{ parameters.libName }}-artifacts/package.json
```
tnorling: Shouldn't this be package-lock?

tnorling: Oh I see you're updating package-lock below. What is this step doing?

.pipelines/custom-templates/publish-template.yml
```diff
@@ -124,9 +131,9 @@ jobs:
           customCommand: 'pack'
 
       # Remove .npmrc file
-      - script: rm .npmrc
+      - script: rm ${{ parameters.path }}/${{ parameters.libName }}/.npmrc
         displayName: Remove .npmrc file
-        workingDirectory: '${{ parameters.path }}/${{ parameters.libName }}'
+        condition: eq('${{ parameters.libName }}', 'msal-angular')
```
tnorling, line 137: Do we want to do this for all libs?

.pipelines/custom-templates/post-publish-template.yml
```diff
@@ -0,0 +1,53 @@
...
+  - name: cdn
+    type: boolean
+    default: false
+
+steps:
+  - task: DownloadPipelineArtifact@2
+    condition: always()
+    inputs:
+      buildType: 'specific'
+      project: 'a7934fdd-dcde-4492-a406-7fad6ac00e17'
```
tnorling, line 18: Is there a variable we can use here instead?

hectormmg: This was auto-populated by the ADO UI when selecting the specific pipeline from a drop-down. I haven't been able to find the place where the project and build reference actually live so I don't think we should mess with the value.

.pipelines/custom-templates/publish-template.yml
hectormmg: Other libs use the NPM task which uses a service connection instead of the .npmrc for authentication. We only do it for angular because its deploy script has other steps and isn't published directly through the npm service connection (that's only configurable when using the NPM publish ADO task)

.pipelines/custom-templates/post-publish-template.yml
hectormmg: This is a good catch of something left over from testing, thanks!

hectormmg: I added a save step for the package.json because it was easier to compare and see if the changes were properly reflected on a diff because it has the version at the top. Removed this step.

.pipelines/custom-templates/publish-template.yml
hectormmg: I did have to add the check to .npmrc write step so it also only runs for angular though. Thanks!

jo-arroyo (approved):

PR #5272, opened 2022-10-03
October Post Release

hectormmg:
This PR contains package lock updates & cdn README updates for msal-browser and msal-core.


jasonnutter (approved):

jasonnutter:
Note, if this needs to be done manually, if possible it would be great to do this step first, so that the updated readme gets included with the version published on npm (and thus appears on npmjs.com listing for the respective packages).

hectormmg:
> Note, if this needs to be done manually, if possible it would be great to do this step first, so that the updated readme gets included with the version published on npm (and thus appears on npmjs.com listing for the respective packages).

@jasonnutter It's not manual, we're just using a PAT for now (until I figure out how to use GitHub apps in this context)

hectormmg:
I guess the CDN step did happen manually, hopefully by next release it won't be the case and the browser/core READMEs will be updated before releasing to npm.

tnorling (approved):

PR #5280, opened 2022-10-06
Migrate node-extensions from nan to node api

tnorling:
- Migrates `msal-node-extensions` C++ addon to node-api (a.k.a. napi)
- Removes dependency on custom bindings package


jasonnutter (approved):
Might be good to get someone from the c++ to review that code, but otherwise looks good!

hectormmg (approved):

peterzenz (approved):

PR #5299, opened 2022-10-13
Remove jasonnutter from CODEOWNERS

jasonnutter:
Thank you all for the last three and half years! Proud of what we have accomplished, and looking forward to following the team's work in the future!


sameerag (approved):
We will miss you man! Signing off.

tnorling (approved):

PR #5342, opened 2022-11-01
Fixed MSAL-Node HttpClient AAD-Server-Error-500 Bug

Robbie-Microsoft:
Previous refactor of MSAL-Node's HttpClient was not tested with an actual 500 error from an AAD server. After using Fiddler Everywhere to further test HttpClient and simulate a 500 error from an AAD server, it was discovered that a 500 error returned from an AAD server would break the application. This PR covers additional functionality in HttpClient to address a 500 error returned from an AAD server.


lib/msal-node/src/network/HttpClient.ts
```diff
@@ -274,3 +276,36 @@ const networkRequestViaHttps = <T>(
...
+     * If the server encounters an error (500 <= statusCode <=599),
+     * the parsed body will be a string error message instead of the expected ServerAuthorizationTokenResponse.
+     * In this case, extra formatting will need to be performed on the error (500 <= statusCode <=599).
+     */
+    let parsedBody;
+    try {
+        parsedBody = JSON.parse(body);
+    } catch (e) {
+        let error;
+        if ((statusCode >= 500) && (statusCode <= 599)) {
```
bgavrilMS: I think semantically 429 also falls in "server_unavailable" - https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429 Typically it has empty body.

lib/msal-node/test/network/HttpClient.spec.ts
```diff
@@ -75,7 +75,23 @@ const mockPostResponseBody: {
 };
 const mockPostResponseBodyBuffer: Buffer = Buffer.from(JSON.stringify(mockPostResponseBody));
 
-const statusCodeOk: number = 200;
+const mockServerErrorResponseBody = "Server Error 500";
+const mockServerErrorResponseBodyBuffer: Buffer = Buffer.from("Server Error 500");
+
+const mockServerErrorResponse: {
```
bgavrilMS: nit: `mockServer500ErrorResponse`

lib/msal-node/test/network/HttpClient.spec.ts
```diff
@@ -195,53 +210,52 @@ describe("HttpClient", () => {
...
-                    await expect(httpClient.sendGetRequestAsync(url, getNetworkRequestOptionsWithProxyUrl)).rejects.toEqual(expect.objectContaining(error));
+                test("Proxy Connection Status Code - Proxy Connection Error 500", async () => {
+                    (http.request as jest.Mock).mockImplementationOnce(mockHttpRequest(mockGetResponseBody, proxyStatusCodeFailure, httpsStatusCodeFailure));
+                    await expect(httpClient.sendGetRequestAsync(url, getNetworkRequestOptionsWithProxyUrl)).rejects.toEqual(proxyError);
                 });
 
-                test("Socket Status Code", async () => {
-                    (http.request as jest.Mock).mockImplementationOnce(mockHttpRequest(mockGetResponseBody, statusCodeOk, statusCodeError));
-                    await expect(httpClient.sendGetRequestAsync(url, getNetworkRequestOptionsWithProxyUrl)).rejects.toEqual(expect.objectContaining(error));
+                test("Socket (Http) Status Code - Server Error 500", async () => {
```
bgavrilMS: Please add test that covers the ` error = "unknown_error";` branch, for example the server returns a 408 + empty body.

lib/msal-node/test/network/HttpClient.spec.ts
```diff
@@ -195,53 +210,52 @@ describe("HttpClient", () => {
...
-        const statusCodeError: number = 500;
-        const error: Error = new Error("Error connecting to proxy");
-        const networkResponse: NetworkResponse<MockedMetadataResponse> = getNetworkResponse<MockedMetadataResponse>(mockGetResponseBody, statusCodeError);
+        const proxyError: Error = new Error("Error connecting to proxy. Status Code: 500. Status Message: Unknown");
 
         describe("Get Request", () => {
-            test("Via Https", async () => {
-                (https.request as jest.Mock).mockImplementationOnce(mockHttpsRequest(mockGetResponseBodyBuffer, statusCodeError));
-                await expect(httpClient.sendGetRequestAsync(url)).resolves.toEqual(networkResponse);
+            test("Via Https - Server Error 500", async () => {
```
bgavrilMS: Are there any tests that cover the AAD error path, where error code is 400 or 401 and  `parsedBody = JSON.parse(body);` succeeds?

change/@azure-msal-node-30bda6e0-6f74-44e0-981f-a7a155ef23ad.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fixed msal-node HttpClient server error bug",
```
bgavrilMS: Can you please log a GitHub issue for this (and an equivalent issue in ADO - this is the JS team's work item pattern and you can continue to follow it - CC @peterzenz ). Having a tracking issue will ensure that the next release explicitly mentions this fix and make this searchable. 

Also please check and see if other HttpClient implementations (browser, angular etc.) suffer from this issue, so that we can discuss with JS team.

tnorling: We do not need a Github Issue for internal work or if it's tracked in ADO. The release notes will mention this change due to the changefile here and link the PR (provided you include the PR number in the comment field)

Robbie-Microsoft: I forgot to add the PR number to the comment field. I'll create an ADO item for the JS team as well.

lib/msal-node/test/network/HttpClient.spec.ts
Robbie-Microsoft: There are not - the existing Http(s) 200 tests cover this. Even in the case you described above, the Http(s) request is successful (which is the purpose of the test). 

Consider an example where the user enters an incorrect client secret. The http(s) request will be successful, and the response's body from the AAD server will look like:
`{
  error: 'invalid_client',
  error_description: "AADSTS7000215: Invalid client secret provided. Ensure the secret being sent in the request is the client secret value, not the client secret ID, for a secret added to app '...'.\r\n" +
    'Trace ID: ...' +
    'Correlation ID: ...' +
    'Timestamp: ...',
  error_codes: [ 7000215 ],
  timestamp: '.',
  trace_id: '.',
  correlation_id: '.',
  error_uri: '...'
}`

The only difference between this and the existing Http(s) 200 tests is the status code and body... but we define those as constants.

lib/msal-node/test/network/HttpClient.spec.ts
Robbie-Microsoft: After implementing your other feedback, I don't think this is necessary. Please take another look at the newest commit and let me know your thoughts.

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -274,3 +279,45 @@ const networkRequestViaHttps = <T>(
...
+     * Server error responses (500 – 599)
+     */
+    
+    let parsedBody;
+    try {
+        parsedBody = JSON.parse(body);
+    } catch (error) {
+        let errorType;
+        let errorDescriptionHelper;
+        if ((statusCode >= 400) && (statusCode <= 499)) {
```
bgavrilMS, line 305: THis is not correct, please remove the 400-499 status codes. These errors are already handled upstream.

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -274,3 +279,45 @@ const networkRequestViaHttps = <T>(
...
+            errorType = "client_error";
+            errorDescriptionHelper = "A client";
+        } else if ((statusCode >= 500) && (statusCode <= 599)) {
```
bgavrilMS, line 308:
```suggestion
        } else if ((statusCode==429) || ((statusCode >= 500) && (statusCode <= 599))) {
```

lib/msal-node/test/network/HttpClient.spec.ts
bgavrilMS: Can you please add a simple unit test for this. Status code 400 and the body being `{ error: 'invalid_client', error_description: "AADSTS7000215: Invalid client secret provided. Ensure the secret being sent in the request is the client secret value, not the client secret ID, for a secret added to app '...'.\r\n" + 'Trace ID: ...' + 'Correlation ID: ...' + 'Timestamp: ...', error_codes: [ 7000215 ], timestamp: '.', trace_id: '.', correlation_id: '.', error_uri: '...' }`

lib/msal-node/test/network/HttpClient.spec.ts
bgavrilMS: From a code coverage perspective, it would be good to cover the "unknown_error" and the "client_error" cases - just add 2 simple unit tests for that. High code coverage is not necessarily a testament to well test code, but it's a good proxy for it and there's no effort (except copy-pasting 2 tests) to get that coverage.

bgavrilMS (approved):

bgavrilMS (approved):

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -107,7 +107,7 @@ const networkRequestViaProxy = <T>(
             if (statusCode < 200 || statusCode > 299) {
                 request.destroy();
                 socket.destroy();
-                reject(new Error(` Error connecting to proxy: ${response.statusCode}, ${response?.statusMessage}`));
+                reject(new Error(`Error connecting to proxy. Status Code: ${response.statusCode}. Status Message: ${response?.statusMessage || "Unknown"}`));
```
sameerag: nit: `proxy.StatusCode` instead of `proxy. Status Code`?

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -169,9 +171,10 @@ const networkRequestViaProxy = <T>(
                 });
                 const headers = Object.fromEntries(entries);
 
+                const parsedHeaders = headers as Record<string, string>;
                 const networkResponse: NetworkResponse<T> = {
-                    headers: headers as Record<string, string>,
-                    body: JSON.parse(body) as T,
+                    headers: parsedHeaders,
```
sameerag: nit: A util API to parse the response into the `networkResponse` format as this code is repeated?

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -274,3 +279,45 @@ const networkRequestViaHttps = <T>(
...
+     * Successful responses (200 – 299)
+     * Redirection messages (300 – 399)
+     * Client error responses (400 – 499)
+     * Server error responses (500 – 599)
+     */
+    
+    let parsedBody;
+    try {
+        parsedBody = JSON.parse(body);
+    } catch (error) {
```
sameerag, line 304: switch instead for readability?

sameerag: I would like ranges and errors in "Constants" and used here as these are pretty much industry standard.

sameerag (commented):
Added a few comments, mostly around code structure and readability.

lib/msal-node/src/network/HttpClient.ts
bgavrilMS: The "." there represents the end of a sentence, not object.method :)

bgavrilMS: Maybe use "Error connecting to proxy. Http status code: ..."

sameerag: :-D. I meant can we have it more like `code:Code` but your suggestion is fine too.

lib/msal-node/src/network/HttpClient.ts
Robbie-Microsoft: I had it as a switch originally... but there's only 2 options before the else. I think an if/else is better for readability. I will define those numbers as constants though.

lib/msal-node/src/utils/Constants.ts
```diff
@@ -13,7 +13,18 @@ export enum HttpMethod {
 
 export enum HttpStatus {
     OK = 200,
-    REDIRECT = 302
+    OK_UPPER_LIMIT = 299,
+    REDIRECT = 302,
+    CLIENT_ERROR_LOWER_LIMIT = 400,
```
sameerag: comments:
* nit: CLIENT_ERROR_RANGE_START and CLIENT_ERROR_RANGE_END? Same pattern for others. 
* nit: SUCCESS instead of OK?
* Why are we addressing only `302`? Do we not need to handle any other REDIRECTxx?

Robbie-Microsoft: The 302 is only being used in LoopbackClient, to "Prevent auth code from being saved in the browser history". I'll implement your syntax suggestions.

lib/msal-node/src/utils/Constants.ts
```diff
@@ -12,8 +12,19 @@ export enum HttpMethod {
 }
 
 export enum HttpStatus {
-    OK = 200,
```
Robbie-Microsoft, line 15: This was not being used anywhere.

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -274,3 +281,45 @@ const networkRequestViaHttps = <T>(
...
+        let errorType;
+        let errorDescriptionHelper;
+        if ((statusCode >= HttpStatus.CLIENT_ERROR_RANGE_START) && (statusCode <= HttpStatus.CLIENT_ERROR_RANGE_END)) {
```
sameerag, line 307: Here is where I think we should address `HttpStatus.REDIRECT_ERROR_RANGE_START` && .... to classify it as a certain type of error.

sameerag (approved):
Added one final comment to complete the status checks and error assignments. Approving though.

lib/msal-node/src/network/HttpClient.ts
Robbie-Microsoft: @bgavrilMS what are your thoughts on this? After some research, it looks like redirections (http status 300-399) can be sent without a body, which would trigger this error. However, redirections aren't errors. I don't think the response body would be used in the case of a redirection anyways... should we just send back an empty one in the case of a redirect?

bgavrilMS: We don't expect 3xx errors from the /token endpoint. But the HTTP stack should handle redirects automatically in your code, which in this case it does not :(

https://stackoverflow.com/questions/7323932/how-do-you-follow-an-http-redirect-in-node-js

I recommend leaving it as an "unknown error" and we'll fix it if it ever comes up.

PR #5352, opened 2022-11-01
JS queue measurements POC

jo-arroyo:
Proof of concept for measuring the time spent in the JS queue.

### Explanation
Instrumentation has been added to each async function called in `acquireTokenSilent`, to see how long between the time the function is called and when it is executed. The time and count are accumulated and sent when telemetry is flushed. 

### To see JS queue measurements in Kusto
- Use the `PerformanceTelemetrySample` in the msal-javascript-1p repo
   - To acquire token by refresh token, add `forceRefresh: true` to the `authRequest` in Profile.tsx
   - To acquire token by Iframe, add `cacheLookupPolicy: CacheLookupPolicy.Skip` to the `loginRequest` in authConfig.ts
- Run the sample, making sure to request profile information
- Use the MSALJS Tenant in Kusto
   - Query for `acquiretokensilent` with `queueTimeMs` and `queueCount`, or follow the example below:
       ```
      acquiretokensilent
      | where EventInfo_Time > ago(1h)
      | where endPageVisibility == startPageVisibility
      | project EventInfo_Name, fromCache, correlationId, queuedCount, queuedTimeMs, acquireTokenSilentAsyncDurationMs, durationMs
      ```
   - Sort by `queueTimeMs` to see queue times and count for each correlation Id
   
### Included in this PR
- Changes to PerformanceClient and BrowserPerformanceClient to allow JS queue measuring
- Additional events added to PerformanceEvents
- Instrumentation in msal-common and msal-browser
- Unit tests


lib/msal-common/src/telemetry/performance/PerformanceClient.ts
```diff
@@ -82,6 +91,76 @@ export abstract class PerformanceClient implements IPerformanceClient {
      */
     abstract generateId(): string;
 
+    /**
+     * Returns current time in milliseconds. 
+     * 
+     * @returns {number}
+     */
+    abstract getCurrentTime(): number;
```
tnorling: Does this have to be abstract or can we define it here?

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
```diff
@@ -82,6 +91,76 @@ export abstract class PerformanceClient implements IPerformanceClient {
...
+     * Calculates the difference between current time and time when function was queued.
+     * Note: It is possible to have 0 as the queue time if the current time and the queued time was the same.
+     * 
+     * @param {number} preQueueTime 
+     * @param {number} currentTime 
+     * @returns {number}
+     */
+    calculateQueuedTime(preQueueTime: number, currentTime: number): number {        
+        // More number calculations here?
+        if (Number.isInteger(preQueueTime) && preQueueTime < 1) {
```
tnorling: Do we have to validate that what was passed in is in fact a number if we've typed it as such in the argument list?

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
```diff
@@ -82,6 +91,76 @@ export abstract class PerformanceClient implements IPerformanceClient {
...
+    /**
+     * Adds queue measurement time to QueueMeasurements array for given correlation ID.
+     * 
+     * @param {PerformanceEvents} name 
+     * @param {?string} correlationId 
+     * @param {?number} time 
+     * @returns 
+     */
+    addQueueMeasurement(name: PerformanceEvents, correlationId?: string, time?: number): void {
+        if (!correlationId) {
```
tnorling, line 192: Can we make correlationId required? At this point all of our APIs should be generating correlationId as a first step, if not one of the first steps, if it's not defined it's likely a bug.

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -620,15 +622,19 @@ export abstract class ClientApplication {
...
+        this.performanceClient.addQueueMeasurement(PerformanceEvents.AcquireTokenByRefreshToken, commonRequest.correlationId, preQueueTime);
         switch(silentRequest.cacheLookupPolicy) {
             case CacheLookupPolicy.Default:
             case CacheLookupPolicy.AccessTokenAndRefreshToken:
             case CacheLookupPolicy.RefreshToken:
             case CacheLookupPolicy.RefreshTokenAndNetwork:
                 const silentRefreshClient = this.createSilentRefreshClient(commonRequest.correlationId);
-                return silentRefreshClient.acquireToken(commonRequest);
+
+                const preAcquireTokenTime = this.performanceClient.getCurrentTime();
```
tnorling: Rather than passing these around, can we utilize the in-memory cache which should be accessible from each of these methods?

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
```diff
@@ -217,6 +226,8 @@ export abstract class StandardInteractionClient extends BaseInteractionClient {
         // fall back to the authority from config
         const builtAuthority = Authority.generateAuthority(userAuthority, requestAzureCloudOptions || this.config.auth.azureCloudOptions);
         this.logger.verbose("Creating discovered authority with configured authority", this.correlationId);
+        
+        // TODO: Need to measure here? Relation to correlationId?
```
peterzenz: Why would we skip this one?

lib/msal-browser/src/interaction_handler/InteractionHandler.ts
```diff
@@ -120,9 +128,10 @@ export class InteractionHandler {
      * @param authority 
      * @param networkModule 
      */
-    protected async updateTokenEndpointAuthority(cloudInstanceHostname: string, authority: Authority, networkModule: INetworkModule): Promise<void> {
+    protected async updateTokenEndpointAuthority(cloudInstanceHostname: string, authority: Authority, networkModule: INetworkModule, preQueueTime?: number): Promise<void> {
+        this.performanceClient.addQueueMeasurement(PerformanceEvents.UpdateTokenEndpointAuthority, this.authCodeRequest.correlationId, preQueueTime);
         const cloudInstanceAuthorityUri = `https://${cloudInstanceHostname}/${authority.tenant}/`;
-        const cloudInstanceAuthority = await AuthorityFactory.createDiscoveredInstance(cloudInstanceAuthorityUri, networkModule, this.browserStorage, authority.options, this.logger);
+        const cloudInstanceAuthority = await AuthorityFactory.createDiscoveredInstance(cloudInstanceAuthorityUri, networkModule, this.browserStorage, authority.options, this.logger); // TODO: AuthorityFactory calculation here? msal-common
```
peterzenz: Why would we skip this one?

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -187,7 +203,7 @@ export class RefreshTokenClient extends BaseClient {
         };
 
         const endpoint = UrlString.appendQueryString(authority.tokenEndpoint, queryParameters);
-        return this.executePostToTokenEndpoint(endpoint, requestBody, headers, thumbprint)
+        return this.executePostToTokenEndpoint(endpoint, requestBody, headers, thumbprint) // TODO: not doing executePostRequest because no correlationId available?
```
peterzenz: As I'm reading through this I'm wondering why we're correlating this with the correlation ID, do we need to do this?

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -620,15 +622,19 @@ export abstract class ClientApplication {
...
+                return silentRefreshClient.acquireToken(commonRequest, preAcquireTokenTime);
```
peterzenz: The general pattern you have here is
`currTime = getCurrentTime();`
`method(request, currTime);`


If you added a timestamp or map of timestamps to the request object could we eliminate all the prototype changes to not add the currTime parameter?  Code would look a bit more like this:
`request.currentTime = getCurrentTime();`
`method(request);`

peterzenz (approved):

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -602,6 +602,8 @@ export abstract class ClientApplication {
         commonRequest: CommonSilentFlowRequest,
         silentRequest: SilentRequest
     ): Promise<AuthenticationResult> {
+        const preQueueTime = this.browserStorage.getPreQueueTime(PerformanceEvents.AcquireTokenFromCache, commonRequest.correlationId);
```
konstantin-msft: Would it be possible to combine `getPreQueueTime` and `addQueueMeasurement` into a single method? These methods are always being invoked one after another.

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
jo-arroyo: No longer skipped, added queue measurements here.

lib/msal-browser/src/interaction_handler/InteractionHandler.ts
jo-arroyo: No longer skipped, added queue measurements here.

lib/msal-browser/src/app/ClientApplication.ts
jo-arroyo: Updated timestamp method to store current time on the Performance client. Please see updated pattern.

lib/msal-common/src/client/RefreshTokenClient.ts
jo-arroyo: Offline discussion with Peter about correlating queue times with correlation Id, as well as the existing pattern for telemetry measurements. Leaving as is for now.

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
jo-arroyo: Outdated. Updated pattern for getting timestamp.

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
jo-arroyo: Good point. Removed typecheck.

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
jo-arroyo: I looked into this, but did not make correlation Id required for the time being, because there are times where we are adding queue measurements before our APIs are generating correlation Id as one of the first steps.

lib/msal-browser/src/app/ClientApplication.ts
jo-arroyo: Thanks for this idea. Updated to store the pre-queue times on the PerformanceClient.

lib/msal-browser/src/app/ClientApplication.ts
jo-arroyo: Thanks for this feedback. Updated the pattern and combined these two into one method.

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
```diff
@@ -104,6 +122,102 @@ export abstract class PerformanceClient implements IPerformanceClient {
...
+            this.logger.trace(`PerformanceClient: currentTime should be a positive integer and not ${currentTime}`);
+            return 0;
+        }
+
+        if (currentTime < preQueueTime) {
+            this.logger.trace("PerformanceClient: currentTime is less than preQueueTime, check how time is being retrieved");
+            return 0;
+        }
+
+        return currentTime-preQueueTime;
```
peterzenz, line 180: Please ensure that this is not negative.  Yes, that can happen.  Might be worth tracing and setting this to 0 in that case.

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
```diff
@@ -104,6 +122,102 @@ export abstract class PerformanceClient implements IPerformanceClient {
...
+        if (!correlationId) {
+            this.logger.trace(`PerformanceClient.addQueueMeasurement: correlationId not provided for ${eventName}, cannot add queue measurement`);
+            return;
+        }
+
+        if (queueTime === 0) {
+            // Possible for there to be no queue time after calculation
+            this.logger.trace(`PerformanceClient.addQueueMeasurement: queue time provided for ${eventName} is ${queueTime}`);
+        } else if (!queueTime) {
+            this.logger.trace(`PerformanceClient.addQueueMeasurement: no queue time provided for ${eventName}`);
```
peterzenz, line 201: Maybe a future PBI (let's complete this PR) is when we have weirdness in our handling, like queue times are negative, or things are missing but expected we log a datapoint indicating something is off here with our timing and that data might be suspect.  Can you track an item for this?

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
```diff
@@ -299,6 +413,30 @@ export abstract class PerformanceClient implements IPerformanceClient {
...
+
+        const queueMeasurementForCorrelationId = this.queueMeasurements.get(correlationId);
+        if (!queueMeasurementForCorrelationId) {
+            this.logger.trace(`PerformanceClient: no queue measurements found for for correlationId: ${correlationId}`);
+        }
+
+        let totalTime = 0;
+        let totalCount = 0;
+        queueMeasurementForCorrelationId?.forEach((measurement) => {
+            if (typeof measurement.queueTime === "number") {
```
peterzenz: Is this likely to happen where the queueTime changes type?

peterzenz (approved):

konstantin-msft (approved):

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
jo-arroyo: Good point. I think this was leftover from an earlier version of QueueMeasurement. Removed. Thanks!

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
jo-arroyo: I think I addressed this in lines 175-178 above, but let me know if you have a better way of handling it.

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
jo-arroyo: Got it. Work item created and tracked.

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
```diff
@@ -104,6 +122,101 @@ export abstract class PerformanceClient implements IPerformanceClient {
...
+            this.logger.trace(`PerformanceClient.addQueueMeasurement: no queue time provided for ${eventName}`);
+            return;
+        }
+
+        const queueMeasurement = {eventName, queueTime} as QueueMeasurement;
+
+        // Adds to existing correlation Id if present in queueMeasurements
+        const existingMeasurements = this.queueMeasurements.get(correlationId);
+        if (existingMeasurements) {
+            this.queueMeasurements.set(correlationId, [...existingMeasurements, queueMeasurement]);
```
tnorling: nit:
```suggestion
            this.queueMeasurements.set(correlationId, existingMeasurements.push(queueMeasurement));
```

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
```diff
@@ -104,6 +122,101 @@ export abstract class PerformanceClient implements IPerformanceClient {
...
+        } else {
+            // Sets new correlation Id if not present in queueMeasurements
+            this.logger.trace(`PerformanceClient.addQueueMeasurement: adding correlationId ${correlationId} to queue measurements`);
+            const measurementArray = [];
+            measurementArray.push(queueMeasurement);
```
tnorling: nit:
```suggestion
            const measurementArray = [queueMeasurement];
```

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -63,15 +67,18 @@ export class AuthorizationCodeClient extends BaseClient {
      * @param request
      */
     async acquireToken(request: CommonAuthorizationCodeRequest, authCodePayload?: AuthorizationCodePayload): Promise<AuthenticationResult> {
+        if (!request || StringUtils.isEmpty(request.code)) {
```
tnorling: nit: (never been a fan of this "helper")
```suggestion
        if (!request || !request.code) {
```

lib/msal-browser/src/telemetry/BrowserPerformanceClient.ts
```diff
@@ -56,4 +56,53 @@ export class BrowserPerformanceClient extends PerformanceClient implements IPerf
...
+    setPreQueueTime(eventName: PerformanceEvents, correlationId?: string): void {
+        if (!correlationId) {
+            this.logger.trace(`BrowserPerformanceClient: correlationId for ${eventName} not provided, unable to set telemetry queue time`);
+            return;
+        }
+
+        const preQueueTimesByEvents = this.preQueueTimeByCorrelationId.get(correlationId);
+
+        if (preQueueTimesByEvents){
+            preQueueTimesByEvents.set(eventName, window.performance.now());
```
tnorling, line 86: Do we need to check that `window` and `window.performance` are available? Or is that done somewhere else?

lib/msal-browser/src/telemetry/BrowserPerformanceClient.ts
```diff
@@ -56,4 +56,53 @@ export class BrowserPerformanceClient extends PerformanceClient implements IPerf
...
+            this.logger.trace(`BrowserPerformanceClient: correlationId for ${eventName} not provided, unable to add queue measurement`);
+            return;
+        }
+
+        const preQueueTime = super.getPreQueueTime(eventName, correlationId);
+        if (!preQueueTime) {
+            return;
+        }
+        
+        const currentTime = window.performance.now();
```
tnorling, line 119: Same question here

tnorling (approved):
A couple of non-blocking nits and a question about API availability. Otherwise LGTM

lib/msal-browser/src/telemetry/BrowserPerformanceClient.ts
jo-arroyo: Thanks for pointing that out. Added checks for window performance API.

lib/msal-browser/src/telemetry/BrowserPerformanceClient.ts
jo-arroyo: Added above.

PR #5355, opened 2022-11-02
Fix logger undefined

### LINKED ISSUE #4412: TypeError: Cannot read properties of undefined (reading 'loggerCallback')
### Core Library

### Core Library Version

2.21.0

### Wrapper Library

### Wrapper Library Version

2.1.0

### Description

The application breaks if `loggerOptions` is set to `undefined` explicitly.

### Error Message

```plaintext
TypeError: Cannot read properties of undefined (reading 'loggerCallback')
```

Minified vendor chunk (error reference)

```javascript
...
function y(m, v, E) {
  this.level = o.Info,
  this.localCallback = m.loggerCallback || function() {}, // <-- this line, according to types the variable m can be undefined
  this.piiLoggingEnabled = m.piiLoggingEnabled || !1,
  this.level = "number" == typeof m.logLevel ? m.logLevel : o.Info,
  this.correlationId = m.correlationId || "",
  this.packageName = v || s.gT.EMPTY_STRING,
  this.packageVersion = E || s.gT.EMPTY_STRING
}
...
```

### Msal Logs

_No response_

### MSAL Configuration

```typescript
const msalInstanceFactory = (environment?: IWebClientEnvironment): IPublicClientApplication => {
    auth: {
      clientId: environment?.msal.clientId ?? 'xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx',
      authority: environment?.msal.authority ?? 'https://login.microsoftonline.com/common',
      redirectUri: environment?.msal.redirectUri ?? 'http://localhost:4200',
      postLogoutRedirectUri: '/',
    },
    cache: {
      storeAuthStateInCookie: isIE,
    },
    system: {
      tokenRenewalOffsetSeconds: 300,
      loggerOptions:
        environment?.production === true
          ? void 0 // <-- This breaks the application
          : {
              loggerCallback,
              logLevel: LogLevel.Info,
              piiLoggingEnabled: false,
            },
    },
  });
};
```

### Relevant Code Snippets

Borwser system options type declaration in the library
```typescript
export declare type BrowserSystemOptions = SystemOptions & {
    loggerOptions?: LoggerOptions; // <-- seems like it can be undefined
    navigationClient?: INavigationClient;
    windowHashTimeout?: number;
    iframeHashTimeout?: number;
    loadFrameTimeout?: number;
    navigateFrameWait?: number;
    redirectNavigationTimeout?: number;
    asyncPopups?: boolean;
    allowRedirectInIframe?: boolean;
};
```

Msal instance factory with loggerOptions set to undefined for the production environment
```typescript
const msalInstanceFactory = (environment?: IWebClientEnvironment): IPublicClientApplication => {
    auth: {
      clientId: environment?.msal.clientId ?? 'xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx',
      authority: environment?.msal.authority ?? 'https://login.microsoftonline.com/common',
      redirectUri: environment?.msal.redirectUri ?? 'http://localhost:4200',
      postLogoutRedirectUri: '/',
    },
    cache: {
      storeAuthStateInCookie: isIE,
    },
    system: {
      tokenRenewalOffsetSeconds: 300,
      loggerOptions:
        environment?.production === true
          ? void 0 // <-- This breaks the application
          : {
              loggerCallback,
              logLevel: LogLevel.Info,
              piiLoggingEnabled: false,
            },
    },
  });
};
```

### Reproduction Steps

1. configure application to use msal-angular;
2. set system.loggerOptions to `undefined` in the configuration explicitly;
3. start the application locally;
4. open a browser and the browser devtools (console);
5. load the application url;
6. the error should be there;

### Expected Behavior

According to existing types shipped with the library, nothing should be broken when `loggerOptions` is set to undefined.

Seems like either types shipped with the library should be fixed, or there should be explicit checks in the library code like

```typescript
if (typeof system.loggerOptions !== 'undefined') {
  // do something
}
```

### Identity Provider

Azure AD / MSA

Chrome, and possibly all other browsers, but I have tested only in Chrome.

I am sorry, I don't have bandwidth to test in all available browsers.

### Regression

_No response_

### Source

External (Customer)

bmahall:
Fixes #4412 

When loggerOptions is explicitly set to undefined/void 0, it leads to "Cannot read options of undefined" error which breaks the application. This PR makes loggerOptions optional to address this. 

Tests included for msal-browser and msal-node.
Covers: msal-browser,msal-angular,msal-node samples.


lib/msal-common/src/logger/Logger.ts
```diff
@@ -62,10 +62,16 @@ export class Logger {
         const defaultLoggerCallback = () => {
             return;
         };
-        this.localCallback = loggerOptions.loggerCallback || defaultLoggerCallback;
-        this.piiLoggingEnabled = loggerOptions.piiLoggingEnabled || false;
-        this.level = typeof(loggerOptions.logLevel) === "number" ? loggerOptions.logLevel : LogLevel.Info;
-        this.correlationId = loggerOptions.correlationId || Constants.EMPTY_STRING;
+        const defaultLoggerOptions = {};
+        if(!loggerOptions)
```
bmahall: this is added for coherence with defaultLoggerCallback, can be removed based on the feedback as it doesnt impact the working/tests.

change/msal-6c65b933-30cb-4cc9-b6ae-4b890ca4b848.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "fix logger undefined error",
+  "packageName": "msal",
```
tnorling: Please remove as it doesn't appear any v1 changes were made

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
```diff
@@ -152,9 +152,9 @@ export abstract class StandardInteractionClient extends BaseInteractionClient {
                 preventCorsPreflight: true
             },
             loggerOptions: {
-                loggerCallback: this.config.system.loggerOptions.loggerCallback,
-                piiLoggingEnabled: this.config.system.loggerOptions.piiLoggingEnabled,
-                logLevel: this.config.system.loggerOptions.logLevel,
+                loggerCallback: this.config.system?.loggerOptions?.loggerCallback,
```
tnorling: These should all be defined by this point in the flow (with defaults assigned if no logger was provided by the developer), so these '?' should not be needed

lib/msal-common/src/logger/Logger.ts
tnorling: Logger options is a required parameter so this should never be hit

peterzenz (approved):

lib/msal-common/src/logger/Logger.ts
```diff
@@ -62,10 +62,16 @@ export class Logger {
...
+        {
+        // eslint-disable-next-line no-param-reassign
```
tnorling: This lint rule is important, please refactor rather than ignoring

lib/msal-common/src/logger/Logger.ts
```diff
@@ -62,10 +62,16 @@ export class Logger {
...
+            loggerOptions= defaultLoggerOptions;
+        }
+        this.localCallback = loggerOptions?.loggerCallback || defaultLoggerCallback;
```
tnorling: Logger options should be defined by now, '?' should not be needed

lib/msal-common/test/logger/Logger.spec.ts
```diff
@@ -187,6 +187,18 @@ describe("Logger.ts Class Unit Tests", () => {
             expect(logStore[LogLevel.Verbose]).toBe(undefined);
             expect(logStore[LogLevel.Trace]).toBe(undefined);
         });
+
+        it("Doesn't throw if loggerOptions explicitly set to undefined", () => {
+            //@ts-ignore
+            loggerOptions=void 0 || undefined || null;
```
tnorling: This will always evaluate to `null`, is the intent here to test `undefined`?

tnorling (changes requested):

lib/msal-common/test/logger/Logger.spec.ts
bmahall: thanks, will update

lib/msal-common/src/logger/Logger.ts
bmahall: When the user is setting the logger options in this way:-
{
loggerOptions:
           void 0
}
the sample throws the undefined error, its not using defaultCallback. Hence , had to make these as optional.

bmahall: In conjunction with L69, it need not be defined , got you. 
Do you prefer using defaultLoggerOptions or using '?' approach here?

tnorling: void 0 is not an accepted type therefore it's expected that this would fail

tnorling: By the time the Logger is initialized we should have a defined logger options object that can be passed in (whether it is user provided or defaults) but that logic lives further up the stack (in msal-browser/node) which is why it is a required parameter here

lib/msal-common/src/logger/Logger.ts
bmahall: Have refactored to use a default setter method

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
bmahall: have refactored and removed '?' , this PR is to cover when loggerOptions explicitly set to undefined/ void 0 to prevent breaking. when no logger is provided, I agree this is not needed at all.

lib/msal-common/test/logger/Logger.spec.ts
bmahall: updated to use undefined options only, thanks for pointing this out! :)

sameerag (approved):

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
```diff
@@ -171,6 +172,16 @@ export abstract class StandardInteractionClient extends BaseInteractionClient {
         };
     }
 
+    private static createDefaultLoggerOptions(): LoggerOptions {
```
tnorling: This should be done in `Configuration.ts` instead

lib/msal-common/src/logger/Logger.ts
```diff
@@ -62,14 +62,25 @@ export class Logger {
         const defaultLoggerCallback = () => {
             return;
         };
-        this.localCallback = loggerOptions.loggerCallback || defaultLoggerCallback;
-        this.piiLoggingEnabled = loggerOptions.piiLoggingEnabled || false;
-        this.level = typeof(loggerOptions.logLevel) === "number" ? loggerOptions.logLevel : LogLevel.Info;
-        this.correlationId = loggerOptions.correlationId || Constants.EMPTY_STRING;
+        const setLoggerOptions = loggerOptions || Logger.createDefaultLoggerOptions();
```
tnorling, line 65: Let's change `LoggerOptions` in the constructor to `Required<LoggerOptions>` and then you don't need to repeat the default logic here

lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -357,11 +357,9 @@ export abstract class ClientApplication {
                 proxyUrl: this.config.system.proxyUrl,
             },
             loggerOptions: {
-                logLevel: this.config.system.loggerOptions.logLevel,
-                loggerCallback: this.config.system.loggerOptions
-                    .loggerCallback,
-                piiLoggingEnabled: this.config.system.loggerOptions
-                    .piiLoggingEnabled,
+                logLevel: this.config.system.loggerOptions?.logLevel || LogLevel.Info,
```
tnorling: Like in browser, do this in `Configuration.ts`

bmahall: @tnorling , This error is occurring due to this bit in the Configuration.ts:-

`    auth,
    cache,
    system,
    telemetry
    return {
        auth: { ...DEFAULT_AUTH_OPTIONS, ...auth },
        cache: { ...DEFAULT_CACHE_OPTIONS, ...cache },
        system: { ...DEFAULT_SYSTEM_OPTIONS, ...fixSystem },
        telemetry: { ...DEFAULT_TELEMETRY_OPTIONS, ...telemetry }
    };
}`


Either this can be modified so that app doesnt break, for the sake of simplicity of the code change involved, making a change in the ClientApplication was preferred.

tnorling: Not sure I understand - all top level configuration overrides should be done in `Configuration.ts`. We shouldn't be modifying the configuration anymore after that.

bmahall: Addressed

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
bmahall: Addressed

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
bmahall: Addressed & resolving

lib/msal-common/src/logger/Logger.ts
bmahall: This will entail type changes in ClientApplication.ts for msal-node. @tnorling

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -270,6 +270,11 @@ export function buildConfiguration({ auth: userInputAuth, cache: userInputCache,
         }
     };
 
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const providedSystemOptions:any = {...userInputSystem};
```
tnorling, line 274:
```suggestion
    const providedSystemOptions: SystemOptions = {...userInputSystem};
```

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -270,6 +270,11 @@ export function buildConfiguration({ auth: userInputAuth, cache: userInputCache,
...
+    if (!providedSystemOptions.loggerOptions) {
+        providedSystemOptions.loggerOptions = DEFAULT_LOGGER_OPTIONS;
+    }
```
tnorling:
```suggestion
    const providedSystemOptions: BrowserSystemOptions = {
        ...userInputSystem,
        loggerOptions: userInputSystem.loggerOptions || DEFAULT_LOGGER_OPTIONS
    };
```

lib/msal-node/src/config/Configuration.ts
```diff
@@ -158,10 +158,14 @@ export function buildAppConfiguration({
     telemetry
 }: Configuration): NodeConfiguration {
 
+    const providedSystemOptions = {...system};
+    if (!providedSystemOptions.loggerOptions) {
+        providedSystemOptions.loggerOptions = DEFAULT_LOGGER_OPTIONS;
+    }
```
tnorling:
```suggestion
    const providedSystemOptions = {
        ...system,
        loggerOptions: system.loggerOptions || DEFAULT_LOGGER_OPTIONS
    };
```

tnorling (commented):
Looks good, can you add some tests?

lib/msal-browser/src/config/Configuration.ts
bmahall: Addressed.

lib/msal-node/src/config/Configuration.ts
bmahall: Addressed.

tnorling (approved):

PR #5378, opened 2022-11-09
Update acquireToken docs with best practices

### LINKED ISSUE #5337: acquiretokenredirect best practices guidance
### Core Library

### Core Library Version

2.21.0

### Wrapper Library

### Wrapper Library Version

1.4.0

Public

### Description

Hi! I am just looking for guidance on best practice with acquireTokenRedirect.

We are using msal-react, but we have reusable functions we use to acquire tokens with msal-browser (not in any react component where we pass in the instance and context).

We have several parallel requests which run, and we are not sure what the best practice is for acquiring tokens by redirect.

It seems this is called several times on each failed acquireTokenSilent request when tokens have expired, causing a redirect loop of about 4 loops until the page loads. I'm wondering what guidance would be to fix our issue and improve how we are using msal.

Thanks so much!

### MSAL Configuration

```javascript
export const msalConfig = {
    auth: {
        clientId: config.auth.clientId,
        authority: 'https://login.microsoftonline.com/consumers',
        navigateToLoginRequestUrl: false,
        postLogoutRedirectUri: `${config.auth.redirectUrl || config.sasUrl}`,
        redirectUri: getRedirectUri()
    },
    cache: {
    },
};
```

### Relevant Code Snippets

```javascript
get token function: const getToken = async (msalContext: IMsalContext, scopes: string[], telemetryService: ITelemetryService): Promise<AuthenticationResult> => {
    const msalInstance = msalContext.instance;
    const accounts = msalInstance.getAllAccounts();
    if (accounts.length > 0) {
        const silentRequest = {
            scopes,
            account: msalInstance.getActiveAccount() || accounts[0]
        };
        const redirectRequest = {
            scopes,
        };
        try {
            return await msalInstance.acquireTokenSilent(silentRequest);
        }catch (error) {
            if (error instanceof (InteractionRequiredAuthError)){
                try {
                    // fallback to interaction when silent call fails
                    if (msalContext.inProgress === InteractionStatus.None){
                        await msalInstance.acquireTokenRedirect(redirectRequest);
                    }
                }catch (e) {
                    telemetryService.trackError(e, TelemetryEventName.GetMsalTokenRedirect);
                    return null;
                }
            }
            telemetryService.trackError(error, TelemetryEventName.GetMsalTokenSilent);
            return null;
        }
    }
};




where we fetch token in axios interceptor (abstracted some away from here  as it's not really needed)

  constructor(msalContext: IMsalContext, telemetryService: ITelemetryService) {
        this.axiosInstance = axios.create();
        this.telemetryService = telemetryService;
        this.axiosInstance.interceptors.request.use(
            (config: ExtendedAxiosRequestConfig) => {
                config.requestId = createGuid();
                config.startTimeMs = performance.now();
                return new Promise((resolve, reject) => {
                    getToken(msalContext, telemetryService).then(tokenResult => {
                        if (!tokenResult) {
                            throw new AxiosError("No token");
                        }
                        const { accessToken } = tokenResult;
                        config.headers['Authorization'] = `Bearer ${accessToken}`;
                        config.headers['Content-Type'] = 'application/json';
                        config.headers['ms-cv'] = telemetryService?.getCorrelationVector(true);
                        if (config.telemetryEventName) {
                            config.headers['x-ms-operation-name'] = config.telemetryEventName;
                        }
                        resolve(config);
                    }, reject);
                });
            }
        );
```

### Identity Provider

Azure AD / MSA

### Source

Internal (Microsoft)

derisen:
Updates the current docs with best practices to follow in order to minimize interactions and avoid concurrent interactive requests. Looking forward to feedback and/or more cases that should be covered in these.

In response to #5337 

p.s. moved the existing content around a little bit for better flow & readibility


lib/msal-browser/docs/acquire-token.md
```diff
@@ -9,38 +9,27 @@ In MSAL, you can get access tokens for the APIs your app needs to call using the
...
-- `CacheLookupPolicy.RefreshTokenAndNetwork` - `acquireTokenSilent` will not look in the cache for the access token. It will go directly to network with the cached refresh token. If the refresh token is expired an attempt will be made to renew it. This is equivalent to setting `forceRefresh: true`.
-- `CacheLookupPolicy.Skip` - `acquireTokenSilent` will attempt to renew both access and refresh tokens. It will not look in the cache. This will always fail if 3rd party cookies are blocked by the browser.
+It is best practice to attempt an `acquireTokenSilent` call before using the interactive APIs if you have already logged in. This allows you to prevent unnecessary user interactions. `acquireTokenSilent` will look for a valid token in the cache, and if it is close to expiring or does not exist, will automatically try to refresh it for you. You should use a `login*` or `acquireToken*` (interactive) API before this to establish a session with the server. You can read more about using `acquireTokenSilent` [here](./token-lifetimes.md#token-renewal).
```
tnorling: The first sentence and 2nd to last sentence may confuse people who don't pick up on the "already logged in" bit. I'd try rephrasing this, perhaps something like this:
```suggestion
After you've logged in with one of the `ssoSilent` or `login*` APIs the cache will contain a set of id, access and refresh tokens. Every time you need an access token you should call `acquireTokenSilent` and if this fails call an interactive API instead. `acquireTokenSilent` will look for a valid token in the cache, and if it is close to expiring or does not exist, will automatically try to refresh it for you using the cached refresh token. You can read more about using `acquireTokenSilent` [here](./token-lifetimes.md#token-renewal).
```

lib/msal-browser/docs/acquire-token.md
```diff
@@ -89,10 +79,42 @@ var graphEndpoint = "https://graph.microsoft.com/v1.0/me";
...
+
+### Attempt silent request first, then interactive
+
+When requesting tokens, always use `acquireTokenSilent` first, falling back to interactive token acquisition if needed. Exception to this is when doing token acquisition outside of an app context that does not have access to interaction status or cannot ensure other components making interactive requests. For instance, if you are making use of `acquireTokenSilent` in a utility function,
+
+Multiple concurrent silent requests are permitted. If two or more of silent requests are made at the same time, only one would go to the network (if needed), but all would receive the response, as long as those requests are for the same requests parameters (e.g. scopes).
+
+### Make one token request per resource
+
+You can only request access tokens for one resource at a time (see [resources and scopes](resources-and-scopes.md)). If needed, you can ask user consent to scopes (permissions) required by more than one resource by using the `extraScopesToConsent` parameter in the request object. Access tokens for previously consented scopes can be acquired silently.
```
1hw7, line 108: Hi Derisen! thanks for doing this! 
I have some questions remaining around best practices for concurrency. Given that it's common for apps to have multiple data fetches concurrently running which may require different tokens for different resources/apis, how should this be handled?

Our logic to fetch tokens existed in an axios request interceptor. The problem was when tokens became stale, and the silent token fetch failed, given the concurrent nature of data fetch requests, we had several requests being fired off for msal tokens, causing many redirects (and throwing msal errors). I'm wondering how a user of msal could get around this without implementing their own cache for their tokens (the problem being, we don't find out any tokens are stale in the msal cache before firing off these api requests, at which point now many requests have been fired off and we will have too many redirects). I'm wondering if this is a problem other people will run into, and if we could illustrate how people can avoid these issues

lib/msal-browser/docs/acquire-token.md
```diff
@@ -89,10 +79,42 @@ var graphEndpoint = "https://graph.microsoft.com/v1.0/me";
...
+
+The following are best practices to acquire tokens with MSAL for avoiding errors, performance hits and usability issues. Certain scenarios may provide exceptions to these.
+
+### Use a single PublicClientApplication instance
+
+Instantiate one `PublicClientApplication` per application and use the same instance throughout your app. This ensures that there is a single source of truth for what MSAL is performing at any given time (see: [MSAL events](events.md)) and eliminates the chance of distinct app objects making parallel interactive requests or potential cache conflicts, which might break apps, reduce performance or hinder user experience.
+
+### Always wait for promises to resolve
+
+All MSAL `acquireToken*` as well as `login*` APIs (which uses `acquireToken*` underneath) perform asynchronous operations and return promises. You should always wait for these promises to resolve, even if you do not need the promise payload.
```
tnorling: Don't need to leak implementation details
```suggestion
All MSAL `acquireToken*` as well as `login*` APIs perform asynchronous operations and return promises. You should always wait for these promises to resolve, even if you do not need the promise payload.
```

lib/msal-browser/docs/acquire-token.md
```diff
@@ -89,10 +79,42 @@ var graphEndpoint = "https://graph.microsoft.com/v1.0/me";
...
+
+### Check for interaction status before interactive requests
+
+You should always check for [InteractionStatus](https://azuread.github.io/microsoft-authentication-library-for-js/ref/enums/_azure_msal_browser.interactionstatus.html) being equal to **None** before making an interactive request. `InteractionStatus` will be emitted through [MsalBroadcastService $inProgress observable](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-angular/docs/v2-docs/events.md#the-inprogress-observable) in MSAL Angular and [MsalProvider useMsal() hook](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-react/docs/hooks.md#usemsal-hook) in MSAL React.
```
tnorling: This is an Angular and React concept - I'd keep this in the docs for those respective libraries

lib/msal-browser/docs/acquire-token.md
```diff
@@ -89,10 +79,42 @@ var graphEndpoint = "https://graph.microsoft.com/v1.0/me";
...
+
+In case of concurrent interactive requests, only the first one will initiate an interaction, while the rest will fail with [interaction_in_progress](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/errors.md#interaction_in_progress) error.
+
+### Attempt silent request first, then interactive
+
+When requesting tokens, always use `acquireTokenSilent` first, falling back to interactive token acquisition if needed. Exception to this is when doing token acquisition outside of an app context that does not have access to interaction status or cannot ensure other components making interactive requests. For instance, if you are making use of `acquireTokenSilent` in a utility function in a React application outside of **MsalProvider context**, you shouldn't attempt to handle *InteractionRequiredAuthError* with an interactive request, but instead route that error to a component that can has access to MSAL context (see: [What can I do outside of msal-react context?](../../msal-react/FAQ.md#what-can-i-do-outside-of-azuremsal-react-context)).
```
tnorling: Same thing here - let's keep this in the React docs

lib/msal-browser/docs/acquire-token.md
derisen: looks better!

lib/msal-browser/docs/acquire-token.md
derisen: Hmm.. I'm thinking this doc as a required reading before msal-angular/react docs, or at least there's benefit in getting familiar with msal-browser token acquisiton APIs even if folks will be working primarily with msal-angular/react. 

Alternatively, do you think we should mention `EventMessageUtils.getInteractionStatusFromEvent` API for obtaining the interaction status (if not using angular or react)?

lib/msal-browser/docs/acquire-token.md
derisen: @1hw7 no worries! From the top of my head, I'm thinking if the first silent request fails and triggers interaction, the newly obtained refresh token would allow other silent requests to succeed... if only the other interceptors would wait for it. Is it not possible to access msal events or interaction status inside the interceptor and have a logic for that? cc @tnorling @sameerag would have better ideas

1hw7: Thanks @derisen for the suggestion and your help! the issue we encountered was that the first request would fail, so we would fall into the catch block to move to an interactive fetch for a token (as we thought was correct from the docs). The part i wasn't clear about in the docs was that users would want to ensure this logic is done in such a place and way that even if multiple requests are going on at the same time, you only want to fallback to this interactive method once so it should ideally be in a component rather than somewhere like an axios interceptor. For that reason, @sameerag had suggested we pull the logic out of the axios interceptor.

We ended up implementing our own token cache to ensure we could safely just do this in one spot in the app, but hope there is a better way we could do that

lib/msal-browser/docs/acquire-token.md
derisen: done -removed this here

lib/msal-browser/docs/acquire-token.md
derisen: Removed this section and added a remark to **Attempt silent request first, then interactive** section for wrapper libraries. Hopefully this would help keeping folks from running into the [issue](https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/5378#discussion_r1019503526) that @1hw7 had

lib/msal-browser/docs/acquire-token.md
```diff
@@ -89,10 +79,89 @@ var graphEndpoint = "https://graph.microsoft.com/v1.0/me";
...
+
+The following are best practices to acquire tokens with MSAL for avoiding errors, performance hits and usability issues. Certain scenarios may provide exceptions to these.
+
+### Use a single PublicClientApplication instance
+
+Instantiate one `PublicClientApplication` per application and use the same instance throughout your app. This ensures that there is a single source of truth for what MSAL is performing at any given time (see: [MSAL events](events.md)) and eliminates the chance of distinct app objects making parallel interactive requests or potential cache conflicts, which might break apps, reduce performance or hinder user experience.
+
+### Always wait for promises to resolve
+
+All MSAL `acquireToken*` as well as `login*` APIs perform asynchronous operations and return promises. You should always wait for these promises to resolve, even if you do not need the promise payload.
```
tnorling: Let's clarify that you should wait for promises to resolve before doing any other tasks that depend on authentication state or tokens such as rendering user information, calling a protected API or calling other MSAL APIs. We don't want to give the impression that these async functions should in fact be synchronous.

lib/msal-browser/docs/acquire-token.md
```diff
@@ -89,10 +79,89 @@ var graphEndpoint = "https://graph.microsoft.com/v1.0/me";
...
+>
+> - [msal-angular interaction_in_progress error](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-angular/docs/v2-docs/errors.md#interaction_in_progress)
+> - [msal-react interaction_in_progress error](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-react/docs/errors.md#interaction_in_progress)
+>
+> In general, your token acquisition method should check if any other interaction is in progress prior to invoking an interaction. If you are not using wrapper libraries, you can achieve this via a global application state or a broadcast service etc. that emits the current MSAL interaction status via [MSAL Events API](./events.md). See for instance: [custom MsalPlugin in a Vue.js application](../../../samples/msal-browser-samples/vue3-sample-app/src/plugins/msalPlugin.ts).
```
tnorling: I'm not sure we should link to the Vue sample here as it's not really supported. If there's a VanillaJS sample we can link to instead let's use that, otherwise the event docs should be sufficient.

lib/msal-browser/docs/acquire-token.md
```diff
@@ -89,10 +79,89 @@ var graphEndpoint = "https://graph.microsoft.com/v1.0/me";
...
+
+### Attempt silent request first, then interactive
+
+When requesting tokens, always use `acquireTokenSilent` first, falling back to interactive token acquisition if needed (e.g., when the `InteractionRequiredAuthError` is thrown).
+
+Concurrent silent requests are permitted. If two or more silent requests are made concurrently, only one would go to the network (if needed), but all would receive the response, as long as those requests are for the same request parameters (e.g. scopes).
+
+Concurrent interactive requests are **not** permitted. If two or more silent requests are made concurrently, only the first one will start an interaction, while the rest will fail with [interaction_in_progress](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/errors.md#interaction_in_progress) error.
+
+> **Avoiding *interaction_in_progress* errors**
```
tnorling: We have a whole [doc](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/errors.md#interaction_in_progress) dedicated to errors and how to resolve them. Can we link to that doc instead and update, if anything is missing? Keeping all the error information in one place is useful because the error messages themselves link to this doc

derisen: good point! will move this to error doc instead then

lib/msal-browser/docs/acquire-token.md
derisen: Good point -revised this bit

lib/msal-browser/docs/acquire-token.md
derisen: Ok, I've removed the link Vue.js example, and instead added a section to events doc about using the getInteractionStatusFromEvent API (doesn't look like we have a vanilla js sample making use of that)

lib/msal-browser/docs/acquire-token.md
derisen: Done -moved most of this content to the error doc

lib/msal-browser/docs/errors.md
```diff
@@ -114,8 +114,85 @@ msalInstance.handleRedirectPromise()
...
+
+    let tokenResponse;
+
+    try {
+        // attempt silent acquisition first
+        tokenResponse = await msalInstance.acquireTokenSilent(request);
+    } catch (error) {
+        if (error instanceof InteractionRequiredAuthError) {
+            try {
+                tokenResponse = await msalInstance.acquireTokenPopup(request);
```
tnorling: Would be good if we can somehow show an example which calls `acquireTokenPopup` twice to really drive this home. Perhaps at the bottom we can call `myAcquireToken` twice without waiting?

lib/msal-browser/docs/errors.md
```diff
@@ -114,8 +114,85 @@ msalInstance.handleRedirectPromise()
...
+        // attempt silent acquisition first
+        tokenResponse = await msalInstance.acquireTokenSilent(request);
+    } catch (error) {
+        if (error instanceof InteractionRequiredAuthError) {
+            try {
+                /**
+                 * "myWaitFor" method polls the interaction status via getInteractionStatus() from
+                 * the application state and resolves when it's equal to "None".
+                 */
+                await myWaitFor(() => myGlobalState.getInteractionStatus() === InteractionStatus.None);
```
tnorling: Rather than delaying execution of acquireTokenPopup, resulting in potentially multiple popups back to back, if interaction is in progress currently, we should wait and then go back to the top (to call ATS again which should now succeed in theory) else if interaction is not already in progress call popup.

derisen: ah, good idea! will revise

lib/msal-browser/docs/errors.md
derisen: makes sense 👍

derisen: done

lib/msal-browser/docs/errors.md
derisen: revised -let me know what you think

lib/msal-browser/docs/errors.md
```diff
@@ -114,8 +114,106 @@ msalInstance.handleRedirectPromise()
...
+                if (myGlobalState.getInteractionStatus() !== InteractionStatus.None) {
+
+                    /**
+                     * "myWaitFor" method polls the interaction status via getInteractionStatus() from
+                     * the application state and resolves when it's equal to "None".
+                     */
+                    await myWaitFor(() => myGlobalState.getInteractionStatus() === InteractionStatus.None);
+
+                    // wait is over, call myAcquireToken again to re-try acquireTokenSilent
+                    return (await myAcquireToken(tokenRequest));
```
tnorling: Rather than making this recursive, it's safer to just throw/return here and handle waiting and retry in the caller. That way it will retry, at most once
```suggestion
                if (myGlobalState.getInteractionStatus() !== InteractionStatus.None) {

                 throw new Error("interaction_in_progress");
```

lib/msal-browser/docs/errors.md
```diff
@@ -114,8 +114,106 @@ msalInstance.handleRedirectPromise()
...
+
+    return tokenResponse;
+}
+
+const request = {
+    scopes: ["User.Read"]
+};
+
+myAcquireToken(request);
+myAcquireToken(request);
```
tnorling: Feel free to refactor to clean this up if needed:
```suggestion
myAcquireToken(request).catch((e) => {
  /**
                     * "myWaitFor" method polls the interaction status via getInteractionStatus() from
                     * the application state and resolves when it's equal to "None".
                     */
                    await myWaitFor(() => myGlobalState.getInteractionStatus() === InteractionStatus.None);
                    // wait is over, call myAcquireToken again to re-try acquireTokenSilent
                    return (await myAcquireToken(tokenRequest));
});
myAcquireToken(request).catch((e) => {
  /**
                     * "myWaitFor" method polls the interaction status via getInteractionStatus() from
                     * the application state and resolves when it's equal to "None".
                     */
                    await myWaitFor(() => myGlobalState.getInteractionStatus() === InteractionStatus.None);
                    // wait is over, call myAcquireToken again to re-try acquireTokenSilent
                    return (await myAcquireToken(tokenRequest));
});
```

derisen: done! I've just put it in a function to avoid repetition

derisen:
@sameerag could you take a look at this again when u have time?

tnorling (approved):

hectormmg (approved):

PR #5384, opened 2022-11-10
Updated Custom INetworkModule Sample

Robbie-Microsoft:


samples/msal-node-samples/custom-INetworkModule-and-network-tracing/package.json
```diff
@@ -4,14 +4,17 @@
...
+    "install:node": "npm install ../../../lib/msal-node"
   },
   "type": "module",
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "@azure/msal-common": "^7.6.0",
-    "@azure/msal-node": "^1.14.2",
+    "@azure/msal-common": "file:../../../lib/msal-common",
+    "@azure/msal-node": "file:../../../lib/msal-node",
```
derisen: Shouldn't we use a specific msal-node version here?

derisen (approved):
LGTM! One comment. 

Also, we probably don't want to commit `package-lock.json` in samples. You can add a `.npmrc` file to say you don't want them generated (can check other samples for reference)

samples/msal-node-samples/custom-INetworkModule-and-network-tracing/HttpClient.ts
```diff
@@ -102,11 +110,11 @@ const networkRequestViaProxy = <T>(
 
         // establish connection to the proxy
         request.on("connect", (response, socket) => {
-            const statusCode = response?.statusCode || 500;
-            if (statusCode < 200 || statusCode > 299) {
+            const proxyStatusCode = response?.statusCode || ProxyStatus.SERVER_ERROR;
```
bgavrilMS, line 144: nit: If the purpose of the same is to keep it simple, I'd remove the proxy stuff from here. If the purpose of the same is for MSAL devs to use it for testing etc, keep it. Not a blocking comment.

Robbie-Microsoft: I'd say the purpose is to be used for testing. I make it clear in the README that this code is copied directly from msal-node's HttpClient. Additionally, if a developer is behind a corporate firewall (at work) they'll need the proxy code.

samples/msal-node-samples/custom-INetworkModule-and-network-tracing/HttpClient.ts
```diff
@@ -8,13 +8,21 @@ import {
     NetworkRequestOptions,
     NetworkResponse
 } from "@azure/msal-common";
+
+/**
+ * After "npx tsc" is executed via the "npm run start" script, Constants.ts and NetworkUtils.ts are compiled to .js and stored in the /dist/utils folder
```
sameerag: We may change this tooling later. Can we leave the tooling out and make the language more generic so we need not have this comment obsolete when the tool changes?

sameerag: Also you can create local constants? I think depending on both the `ts` and compiled `js` files isn't a good programming model.

Robbie-Microsoft: In that case, I think we should update the comments when we update the tooling. Additionally, we're importing the compiled js file, not the ts file.

tnorling: Agree with Sameera. This is tech debt which we'd like to avoid. And we also should not be importing anything that is not exported i.e. don't import from individual files.

sameerag: @Robbie-Microsoft I may in general agree that comments should be updated as needed and not future proof, in this case, I do not think importing library compiled samples for constants is a good programming practice. I would prefer to create a local constant and handle this or have a JS types import from the MSAL JS binary.

bgavrilMS: What do customers do today ? I assume that by referencing MSAL from npm, you don't have this problem? 

IMO, the "samples" from msal repositories are not really "samples", they are "dev apps" used by MSAL developers to test their work. How about

1. Enhance existing official sample (@derisen for thoughts), but this functionality is probably too complex / rarely used to justify the effort
2. Just use documentation (e.g. wiki) 

As for the dev app, agreed on minimizing tech debt.

samples/msal-node-samples/custom-INetworkModule-and-network-tracing/README.md
```diff
@@ -46,7 +46,7 @@ const config = {
 ## Implement your own custom INetworkModule
 
 There are two approaches to implementing a custom INetworkModule in this sample.
-1. The default msal-node (as of v1.14.2) INetworkModule has been copied to HttpClient.ts and can be imported to app.ts to be used as custom INetworkModule. You can edit HttpClient.ts to include console.log()'s to see how network traffic is processed.
+1. The default msal-node (as of v1.14.3) INetworkModule has been copied to HttpClient.ts and can be imported to app.ts to be used as custom INetworkModule. You can edit HttpClient.ts to include console.log()'s to see how network traffic is processed.
```
derisen, line 49: Can we also update the **Run the app** section below, since just `npm start` suffices to run the sample now

samples/msal-node-samples/custom-INetworkModule-and-network-tracing/package.json
```diff
@@ -4,14 +4,17 @@
   "description": "Custom INetworkModule Sample & Network Tracing via Fiddler Everywhere",
   "scripts": {
     "build": "npx tsc",
-    "start": "npm run build && node dist/app.js"
+    "start": "npm run build && node dist/app.js",
+    "build:package": "cd ../../../lib/msal-common && npm run build && cd ../msal-node && npm run build",
+    "install:common": "npm install ../../../lib/msal-common",
+    "install:node": "npm install ../../../lib/msal-node"
```
derisen, line 10:
```suggestion
    "install:common": "npm install ../../../lib/msal-common",
    "install:node": "npm install ../../../lib/msal-node"
```

nit: might be handier to join these two scripts, i.e. `install:local`

samples/msal-node-samples/custom-INetworkModule-and-network-tracing/app.ts
```diff
@@ -4,7 +4,7 @@ import * as msal from "@azure/msal-node";
  * After "npx tsc" is executed via the "npm run start" script, app.ts and HttpClient.ts are compiled to .js and stored in the /dist folder
  * The app is run via "node dist/app.js", hence the .js import of the HttpClient
  */
-// import HttpClient from "./HttpClient.js";
+// import { HttpClient } from "./HttpClient.js";
 
 const clientConfig: msal.Configuration = {
     auth: {
```
derisen, line 10: line 41 below:

`networkClient: new class CustomHttpClient implements INetworkModule`

should be 

`networkClient: new class CustomHttpClient implements msal.INetworkModule`

samples/msal-node-samples/custom-INetworkModule-and-network-tracing/app.ts
```diff
@@ -4,7 +4,7 @@ import * as msal from "@azure/msal-node";
  * After "npx tsc" is executed via the "npm run start" script, app.ts and HttpClient.ts are compiled to .js and stored in the /dist folder
  * The app is run via "node dist/app.js", hence the .js import of the HttpClient
  */
-// import HttpClient from "./HttpClient.js";
+// import { HttpClient } from "./HttpClient.js";
```
derisen, line 7: If we would like to avoid `.js` imports in `.ts` file without much hassle, I think we could look into using [ts-node](https://www.npmjs.com/package/ts-node)

bgavrilMS:
Is this needed / required by any customer or by us (the app developers)? If it's for app developers, it should not matter, although I'd prefer to have everything in TS from the start. 

For official samples, can we have a work item and track this ?

@sameerag for perspective

sameerag (approved):

sameerag:
One observation: If someone includes MSAL JS and runs this sample, please check for name conflicts in the final binary. May be a non-issue, some JS environments do not account for name spaces.

samples/msal-node-samples/custom-INetworkModule-and-network-tracing/README.md
Robbie-Microsoft: `npm start` also builds the app. The user may want to just start it.

PR #5386, opened 2022-11-11
Added logging to Authority class

Robbie-Microsoft:


lib/msal-common/src/authority/Authority.ts
```diff
@@ -47,13 +47,17 @@ export class Authority {
     public regionDiscoveryMetadata: RegionDiscoveryMetadata;
     // Proxy url string
     private proxyUrl: string;
+    // Logger object
+    public logger: Logger;
 
     constructor(
+        logger: Logger,
```
bgavrilMS: I'm not super familiar with NodeJS concept of public API, so please do educate me, but... If `Authority` is part of public API, can we add a new non-null param to the ctor? Isn't it a breaking change?

I would recommend adding it as the last parameter an making is optional if this is indeed the case.

lib/msal-common/src/authority/Authority.ts
```diff
@@ -47,13 +47,17 @@ export class Authority {
     public regionDiscoveryMetadata: RegionDiscoveryMetadata;
     // Proxy url string
     private proxyUrl: string;
+    // Logger object
+    public logger: Logger;
```
bgavrilMS: private?

lib/msal-common/src/authority/Authority.ts
```diff
@@ -438,33 +443,46 @@ export class Authority {
...
+            this.logger.verbose(`Found the following cloud discovery metata in the config: ${JSON.stringify(metadata)}`);
             metadataEntity.updateCloudDiscoveryMetadata(metadata, false);
             return AuthorityMetadataSource.CONFIG;
         }
+        this.logger.verbose("Did not find cloud discovery metadata in the config");
 
-        // If The cached metadata came from config but that config was not passed to this instance, we must go to the network
+        // If the cached metadata came from config but that config was not passed to this instance, we must go to the network
+        this.logger.verbose("Attempting to get cloud discovery metadata in the cache");
         if (this.isAuthoritySameType(metadataEntity) && metadataEntity.aliasesFromNetwork && !metadataEntity.isExpired()) {
```
bgavrilMS: Is it expensive to log `metadataEntity.isExpired()` ? This should be interesting.

lib/msal-common/src/authority/Authority.ts
```diff
@@ -438,33 +443,46 @@ export class Authority {
...
+            this.logger.verbose(`Found the following metadata in the cache: ${JSON.stringify(metadata)}`);
```
bgavrilMS: Since this is on the hot path, we should probably not create `${JSON.stringify(metadata)}` every time. Is there a way to do something like: 

```
if (this.logger.LogLevel == LogLevel.Verbose)
{
    this.logger.verbose(`Found the following metadata in the cache: ${JSON.stringify(metadata)}`);
}
```

lib/msal-common/src/authority/Authority.ts
```diff
@@ -438,33 +443,46 @@ export class Authority {
...
+            this.logger.verbose(`Found the following cloud discovery metata from the network: ${JSON.stringify(metadata)}`);
             metadataEntity.updateCloudDiscoveryMetadata(metadata, true);
             return AuthorityMetadataSource.NETWORK;
         }
+        this.logger.verbose("Did not find cloud discovery metadata from the network");
         
+        this.logger.verbose("Attempting to get cloud discovery metadata from hardcoded values");
+        const harcodedMetadata = this.getCloudDiscoveryMetadataFromHarcodedValues();
         if (harcodedMetadata && !this.options.skipAuthorityMetadataCache) {
+            this.logger.verbose(`Found the following cloud discovery metata from hardcoded values: ${JSON.stringify(metadata)}`);
```
bgavrilMS: Same observation as above - if verbose ...

lib/msal-common/src/authority/Authority.ts
```diff
@@ -438,33 +443,46 @@ export class Authority {
...
+            this.logger.verbose(`Found the following metadata in the cache: ${JSON.stringify(metadata)}`);
             // No need to update
             return AuthorityMetadataSource.CACHE;
         }
+        this.logger.verbose("Did not find cloud discovery metadata in the cache");
 
-        const harcodedMetadata = this.getCloudDiscoveryMetadataFromHarcodedValues();
-
+        this.logger.verbose("Attempting to get cloud discovery metadata from the network");
         metadata = await this.getCloudDiscoveryMetadataFromNetwork();
```
bgavrilMS, line 469: Is it worth adding some more logging statements inside `getCloudDiscoveryMetadataFromNetwork` ?

bgavrilMS (approved):
Approved with comments.

lib/msal-common/src/authority/Authority.ts
tnorling: That's right this would be breaking, however, this is not exposed to our end users and we've traditionally been ok with making breaking changes in msal-common as long as the changefile for msal-common signifies this contains a "major" change, preventing older versions of dependent libraries from picking up these changes.

So @Robbie-Microsoft please either make this optional at the end as Bogdan suggested, or make the change type "major"

lib/msal-browser/test/cache/BrowserCacheManager.spec.ts
```diff
@@ -7,13 +7,12 @@ import sinon from "sinon";
 import { BrowserAuthErrorMessage } from "../../src/error/BrowserAuthError";
 import { TEST_CONFIG, TEST_TOKENS, TEST_DATA_CLIENT_INFO, RANDOM_TEST_GUID, TEST_URIS, TEST_STATE_VALUES, DEFAULT_OPENID_CONFIG_RESPONSE } from "../utils/StringConstants";
 import { CacheOptions } from "../../src/config/Configuration";
-import { Constants, PersistentCacheKeys, CommonAuthorizationCodeRequest as AuthorizationCodeRequest, ProtocolUtils, Logger, LogLevel, AuthenticationScheme, AuthorityMetadataEntity, AccountEntity, Authority, StubbedNetworkModule, IdToken, IdTokenEntity, AccessTokenEntity, RefreshTokenEntity, AppMetadataEntity, ServerTelemetryEntity, ThrottlingEntity, CredentialType, ProtocolMode, AccountInfo, ClientAuthError, AuthError } from "@azure/msal-common";
```
tnorling: Why was this import removed?

lib/msal-common/src/authority/Authority.ts
tnorling: We'd also need to wrap the stringify operation in a try/catch. I would prefer to omit objects from logger messages for now (as they can also be quite long and make logs hard to read). If there's a desire to log objects (which I think in some cases like this may be warranted) we should take a separate action item to add a mechanism to the logger to print objects directly rather than stringifying them.

lib/msal-common/src/authority/Authority.ts
```diff
@@ -438,33 +443,46 @@ export class Authority {
      * @param newMetadata
      */
     private async updateCloudDiscoveryMetadata(metadataEntity: AuthorityMetadataEntity): Promise<AuthorityMetadataSource> {
+        this.logger.verbose("Attempting to get cloud discovery metadata in the config");
```
bgavrilMS, line 446: Can we log the user's authority somewhere here?

lib/msal-common/src/authority/Authority.ts
hectormmg: I know we added a trace logger for this purpose in browser. I think we can do an equivalent here. In Node it'll be a bit more cumbersome since you can't collapse printed logs in the console like you can with browser dev tools.

lib/msal-common/src/authority/Authority.ts
Robbie-Microsoft: the logger already functions like this. There are different tiers of logging: info, trace, verbose, etc. The logger can be found in /lib/msal-common/src/logger/Logger.ts. By using .verbose here, this message will only print when the developer has specified "verbose" in the logging configuration. 

For example:
```
const clientConfig: msal.Configuration = {
    system: {
        loggerOptions: {
                console.log(message);
            },
            piiLoggingEnabled: false,
            logLevel: msal.LogLevel.Verbose,
        },
```

lib/msal-browser/test/cache/BrowserCacheManager.spec.ts
Robbie-Microsoft: It was grayed out - not being used anywhere.

lib/msal-common/src/authority/Authority.ts
Robbie-Microsoft: it's just a boolean check. I'll display it with the logger.

lib/msal-common/src/authority/Authority.ts
Robbie-Microsoft: I don't understand - what specifically do you want to log?

lib/msal-common/src/authority/Authority.ts
Robbie-Microsoft: yes - added

lib/msal-common/src/authority/Authority.ts
```diff
@@ -47,13 +47,17 @@ export class Authority {
     public regionDiscoveryMetadata: RegionDiscoveryMetadata;
     // Proxy url string
     private proxyUrl: string;
+    // Logger object
+    private logger: Logger;
 
     constructor(
+        logger: Logger,
```
pmaytak: nit: I would add new parameters last + order by most important params (like authority string) first.

lib/msal-common/src/authority/Authority.ts
```diff
@@ -438,33 +443,48 @@ export class Authority {
      * @param newMetadata
      */
     private async updateCloudDiscoveryMetadata(metadataEntity: AuthorityMetadataEntity): Promise<AuthorityMetadataSource> {
+        this.logger.verbose("Attempting to get cloud discovery metadata in the config");
         let metadata = this.getCloudDiscoveryMetadataFromConfig();
         if (metadata) {
+            this.logger.verbose("Found cloud discovery metata in the config.");
```
pmaytak:
```suggestion
            this.logger.verbose("Found cloud discovery metadata in the config.");
```
Similar typo in few places below.

lib/msal-common/src/authority/Authority.ts
```diff
@@ -438,33 +443,48 @@ export class Authority {
...
             metadataEntity.updateCloudDiscoveryMetadata(metadata, false);
             return AuthorityMetadataSource.CONFIG;
         }
+        this.logger.verbose("Did not find cloud discovery metadata in the config");
```
pmaytak: nit: On this line and similar ones below, I would remove this log since if we move to the next search method, we already know the previous one failed.
Also in general I'd combine the logger calls, which are one after the other, into one string and one call to improve perf.

pmaytak (approved):

lib/msal-common/src/authority/Authority.ts
pmaytak: Is this method, `updateCloudDiscoveryMetadata`, really in a hot path? Does it get called for every request? (I'd assume instance discovery would be done once per app.)
If it is in the hot path, we could log only something like canonical authority from the metadata.

lib/msal-common/src/authority/Authority.ts
pmaytak: I think the point here is that `stringify` will always get called first to create the message string to pass into the `verbose` regardless whether verbose logging is enabled or not (since that's checked later).

pmaytak (approved):

bgavrilMS (approved):

lib/msal-common/src/authority/Authority.ts
```diff
@@ -438,33 +443,47 @@ export class Authority {
      * @param newMetadata
      */
     private async updateCloudDiscoveryMetadata(metadataEntity: AuthorityMetadataEntity): Promise<AuthorityMetadataSource> {
+        this.logger.verbose("Attempting to get cloud discovery metadata in the config");
+        this.logger.verbose(`Known Authorities: ${this.authorityOptions.knownAuthorities || Constants.NOT_APPLICABLE}`);
```
tnorling: These should be logger PII

change/@azure-msal-browser-927c443a-ae1c-4393-8209-6cc3dde65e47.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
tnorling: This will require changes in msal-browser-1p so let's make this minor and please make sure to make the required changes in 1p as well
```suggestion
  "type": "minor",
```

tnorling (commented):
Last couple nits but looks good!

tnorling (approved):

PR #5432, opened 2022-12-01
Added additional Authority logs

Robbie-Microsoft:


lib/msal-common/src/authority/Authority.ts
```diff
@@ -552,13 +552,15 @@ export class Authority {
...
                 // If no metadata is returned, authority is untrusted
+                this.logger.verbose("The network request completed successfully, but no metadata was returned.");
                 return null;
             }
             match = Authority.getCloudDiscoveryMetadataFromNetworkResponse(
                 metadata,
                 this.hostnameAndPort
             );
         } catch (e) {
+            this.logger.verbose("There was a network error while attempting to get the cloud discovery metadata.");
```
bgavrilMS: Can you please log this error, otherwise there is no way to understand what's happening.

lib/msal-common/src/authority/Authority.ts
```diff
@@ -552,13 +552,15 @@ export class Authority {
                 : [];
             if (metadata.length === 0) {
                 // If no metadata is returned, authority is untrusted
+                this.logger.verbose("The network request completed successfully, but no metadata was returned.");
```
bgavrilMS: Please log the response.body and response.status code here.

lib/msal-common/src/authority/Authority.ts
```diff
@@ -552,13 +552,15 @@ export class Authority {
...
                 return null;
             }
             match = Authority.getCloudDiscoveryMetadataFromNetworkResponse(
```
bgavrilMS, line 563: I think you should add some logging in this method as well.

lib/msal-common/src/authority/Authority.ts
bgavrilMS: Alos, use Error verbosity i.e. `this.logger.error(...)`

lib/msal-common/src/authority/Authority.ts
bgavrilMS: I'd think this is warning verbosity.

lib/msal-common/src/authority/Authority.ts
```diff
@@ -552,13 +552,15 @@ export class Authority {
...
                 metadata,
                 this.hostnameAndPort
             );
         } catch (e) {
+            this.logger.verbose("There was a network error while attempting to get the cloud discovery metadata.");
             return null;
         }
```
bgavrilMS, line 571: Please add a `log.Warning` in the `if (!match)` statement, as this will tell us if instance discoevry failed.

bgavrilMS (changes requested):
more logging statements pls

lib/msal-common/src/authority/Authority.ts
```diff
@@ -444,8 +444,8 @@ export class Authority {
      */
     private async updateCloudDiscoveryMetadata(metadataEntity: AuthorityMetadataEntity): Promise<AuthorityMetadataSource> {
         this.logger.verbose("Attempting to get cloud discovery metadata in the config");
-        this.logger.verbosePii(`Known Authorities: ${this.authorityOptions.knownAuthorities || Constants.NOT_APPLICABLE}`);
-        this.logger.verbosePii(`Authority Metadata: ${this.authorityOptions.authorityMetadata || Constants.NOT_APPLICABLE}`);
+        this.logger.verbose(`Known Authorities: ${this.authorityOptions.knownAuthorities || Constants.NOT_APPLICABLE}`);
```
tnorling: PII

lib/msal-common/src/authority/Authority.ts
```diff
@@ -444,8 +444,8 @@ export class Authority {
...
+        this.logger.verbose(`Authority Metadata: ${this.authorityOptions.authorityMetadata || Constants.NOT_APPLICABLE}`);
```
tnorling: PII

lib/msal-common/src/authority/Authority.ts
tnorling: Any reason we can't just look at the network logs to determine what was returned? I'd rather not output response body here as it could be very large and make the logs hard to read.

lib/msal-common/src/authority/Authority.ts
Robbie-Microsoft: I was thinking about using .error, but we already use a .error when we throw the untrusted authority error. Do you think we should use it here too?

lib/msal-common/src/authority/Authority.ts
Robbie-Microsoft: Bogdan, remember we talked about not logging an object (response.body)?

lib/msal-common/src/authority/Authority.ts
Robbie-Microsoft: I thought we decided that getCloudDiscoveryMetadataFromNetworkResponse doesn't need logging since it just contains a for loop?

lib/msal-common/src/authority/Authority.ts
Robbie-Microsoft: @bgavrilMS and I had come to the conclusion that knownAuthorities and authorityMetadata aren't PII, and that only canonical_authority is.

tnorling: Both could contain tenant information which we've been treating as PII

bgavrilMS: @tnorling - I think there is a naming confusion here. "known authorities" are just hosts, e.g. "login.microsoft.us", they do not have tenants/

lib/msal-common/src/authority/Authority.ts
bgavrilMS: @Robbie-Microsoft - response.body is not an object, it's a string. It is in JSON format, so it can be deserialized into an object, but it is a string. Also, we should log it at least at warning level.
@tnorling - this only gets logged in error cases, where we're going to be throwing an authority is invalid exception. So it's likely that the app developer will open a bug / ICM and we need the logs :) .See the `if` condition above. Capturing network logs with Node is a big pain, @Robbie-Microsoft couldn't get this to work on mac without paid software.

lib/msal-common/src/authority/Authority.ts
bgavrilMS: This block swallows the original exception (e), making the root cause impossible to determine. The root cause should be logged at error level.

lib/msal-common/src/authority/Authority.ts
tnorling: B2C authorities, which is the specific use case knownAuthorities is for, could include company names. In any case, it's supposed to be hosts only but there's nothing stopping people from providing the full authority (in fact I've seen people do this). We sanitize the provided values and pull out the hosts before using them but the raw provided value used here might still include data customers don't want to log.

tnorling: My general opinion on PII is that if we didn't create the data ourselves (user provided, server provided, etc) we should assume it contains PII

lib/msal-common/src/authority/Authority.ts
Robbie-Microsoft: @bgavrilMS I thought of a good solution. The reponse.body only has two properties: tenant_discovery_endpoint and metadata. If we reach this point, it'e because there is no metadata. Therefore, we only need to display the tenant_discovery_endpoint here.

lib/msal-common/src/authority/Authority.ts
sameerag: There is an Enterprise data taxonomy that helps clarify this officially. I am sharing it offline and please make a decision accordingly. My personal take is to err on the side of caution if we are doubtful reg classification.

bgavrilMS: Thanks for sharing @sameerag. Note that this document is Microsoft confidential, so let's continue the discussion on Teams.

lib/msal-common/src/authority/Authority.ts
```diff
@@ -552,17 +552,25 @@ export class Authority {
                 : [];
             if (metadata.length === 0) {
                 // If no metadata is returned, authority is untrusted
+                this.logger.warning(`The cloud instance discovery network request's status code is: ${response.status}`);
+                this.logger.warning("metadata is an empty list.");
+                
+                const tenantDiscoveryEnpoint = response.body.tenant_discovery_endpoint;
```
bgavrilMS, line 558: Possible null ref, response can be null, body can be null...

lib/msal-common/src/authority/Authority.ts
```diff
@@ -552,17 +552,25 @@ export class Authority {
                 : [];
             if (metadata.length === 0) {
                 // If no metadata is returned, authority is untrusted
+                this.logger.warning(`The cloud instance discovery network request's status code is: ${response.status}`);
```
bgavrilMS, line 555: Ultimately, go through all the logs that you just added and ensure we have a clear picture of what is happening. We already asked the affected partner to provide logs twice, let's not make him do it the thrid time. Please use log.verbose to log as much data is needed. The goal of verbose logs is to provide sufficient information to troubleshoot the issue, not to be succint or to provide good readability. That is secondary.

lib/msal-common/src/authority/Authority.ts
bgavrilMS, line 558: Not a problem to log this, but I don't think it'll help with the issue we're fixing.

bgavrilMS (approved):

lib/msal-common/src/authority/Authority.ts
```diff
@@ -552,17 +552,25 @@ export class Authority {
...
+                this.logger.warning("metadata is an empty list.");
+                
+                const tenantDiscoveryEnpoint = response.body.tenant_discovery_endpoint;
+                const tenantDiscoveryEnpointLoggerHelper = tenantDiscoveryEnpoint.length ? `: ${tenantDiscoveryEnpoint.length}` : " an empty string";
+                this.logger.warning(`tenant_discovery_endpoint is${tenantDiscoveryEnpointLoggerHelper}`);
```
bgavrilMS: The pattern is to log 2 strings: 

- the actual tenant id as pii log statement
- a placeholder as non-pii log statement

lib/msal-common/src/authority/Authority.ts
Robbie-Microsoft: How could these be null? This is in a try/catch, so if the network request fails, it will never get to this point. A status and body will always be returned for any completed network request, per HttpClient.ts.

Robbie-Microsoft:
@bgavrilMS I think we should hold off this PR until the [Cloud Instance Discovery Improvement PR](https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/5448) gets merged in to dev.

lib/msal-common/src/authority/Authority.ts
```diff
@@ -552,17 +552,26 @@ export class Authority {
...
+                const tenantDiscoveryEnpointLoggerHelper = tenantDiscoveryEnpoint.length ? `: ${tenantDiscoveryEnpoint.length}` : " an empty string";
+                this.logger.warningPii(`tenant_discovery_endpoint is${tenantDiscoveryEnpointLoggerHelper}`);
                 return null;
             }
             match = Authority.getCloudDiscoveryMetadataFromNetworkResponse(
                 metadata,
                 this.hostnameAndPort
             );
         } catch (e) {
+            this.logger.error(`There was a network error while attempting to get the cloud discovery instance metadata: ${e}`);
```
pmaytak, line 568: nit: not necessarily a network error. Could be because of a null pointer ex somewhere above.
```suggestion
            this.logger.error(`There was an error while attempting to get the cloud discovery instance metadata: ${e}`);
```

pmaytak, line 568: Will `${e}` printout the full exception with stack trace or just something short like error message? Want to make sure descriptive error is logged.

lib/msal-common/src/authority/Authority.ts
```diff
@@ -552,17 +552,26 @@ export class Authority {
                 : [];
             if (metadata.length === 0) {
                 // If no metadata is returned, authority is untrusted
+                this.logger.warning(`The cloud instance discovery network request's status code is: ${response.status}`);
+                this.logger.warning("There weren't any cloud instances returned.");
+                
+                const tenantDiscoveryEnpoint = response.body.tenant_discovery_endpoint;
+                const tenantDiscoveryEnpointLoggerHelper = tenantDiscoveryEnpoint.length ? `: ${tenantDiscoveryEnpoint.length}` : " an empty string";
```
tnorling, line 559:
```suggestion
                const tenantDiscoveryEnpointLoggerHelper = tenantDiscoveryEnpoint || " an empty string";
```

tnorling: I'm assuming you meant to use the actual endpoint and not the length of the string.

lib/msal-common/src/authority/Authority.ts
tnorling, line 568: `e` is an object - Try printing specific properties instead (e.g. code or message)

Robbie-Microsoft:
Combined this PR into https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/5448

PR #5448, opened 2022-12-02
Cloud Instance Discovery Improvement

Robbie-Microsoft:
In the event of a successful network request to "getCloudDiscoveryMetadataFromNetwork" but there is no metadata, only return null if the response's error message is "invalid_instance".

This makes MSALJS's cloud instance discovery more consistent with MSAL.NET


lib/msal-common/src/authority/Authority.ts
```diff
@@ -543,17 +544,25 @@ export class Authority {
...
+            if (isCloudInstanceDiscoveryResponse(response.body)) {
+                typedResponseBody = response.body as CloudInstanceDiscoveryResponse;
+                metadata = typedResponseBody.metadata;
+            } else {
+                typedResponseBody = response.body as CloudInstanceDiscoveryErrorResponse;
+                if (isCloudInstanceInvalid(typedResponseBody.error)) {
+                    return null;
+                }
+                
+                metadata = [];
```
tnorling, line 573: I'm not sure I understand, what is the difference here between null and empty array?

tnorling:
> In the event of a successful network request to "getCloudDiscoveryMetadataFromNetwork" but there is no metadata, only return null if the response's error message is "invalid_instance".

Can you give an example of when the network request would be successful, with no metadata and without this error message?

lib/msal-common/src/authority/Authority.ts
Robbie-Microsoft: On line 566, an array of metadata (cloud instances, can be empty) is needed to see if the developer's authority is within the cloud instances returned from the network request. And if there's no match... then apparently we make a new Authority for the purpose of a custom domain scenario, on line 576.

Robbie-Microsoft:
> > In the event of a successful network request to "getCloudDiscoveryMetadataFromNetwork" but there is no metadata, only return null if the response's error message is "invalid_instance".
> 
> Can you give an example of when the network request would be successful, with no metadata and without this error message?

@bgavrilMS can you help me here? Is there an example we can use from the .net library?

lib/msal-common/src/authority/Authority.ts
```diff
@@ -543,17 +544,25 @@ export class Authority {
...
                     options
                 );
-            const metadata = isCloudInstanceDiscoveryResponse(response.body)
-                ? response.body.metadata
-                : [];
-            if (metadata.length === 0) {
-                // If no metadata is returned, authority is untrusted
-                return null;
+            
+            let typedResponseBody: CloudInstanceDiscoveryResponse | CloudInstanceDiscoveryErrorResponse;
```
bgavrilMS, line 553: This block needs more logs.

lib/msal-common/src/authority/Authority.ts
bgavrilMS: It's probably to prserve behavior from 

```ts
const metadata = isCloudInstanceDiscoveryResponse(response.body)
                ? response.body.metadata
                : [];
```

lib/msal-common/src/authority/Authority.ts
```diff
@@ -543,17 +544,25 @@ export class Authority {
...
+            let metadata: Array<CloudDiscoveryMetadata>;
+            if (isCloudInstanceDiscoveryResponse(response.body)) {
+                typedResponseBody = response.body as CloudInstanceDiscoveryResponse;
+                metadata = typedResponseBody.metadata;
```
bgavrilMS, line 557: Can `typedResponseBody ` be null? Is `typedResponseBody?.metadata` safer?

lib/msal-common/test/authority/Authority.spec.ts
```diff
@@ -911,7 +911,7 @@ describe("Authority.ts Class Unit Tests", () => {
                 });
             });
 
-            it("throws untrustedAuthority error if host is not part of knownAuthorities, cloudDiscoveryMetadata and instance discovery network call doesn't return metadata", (done) => {
+            it("throws untrustedAuthority error if host is not part of knownAuthorities, cloudDiscoveryMetadata and instance discovery network call doesn't return metadata, and the error returned from the network is 'invalid_instance'", (done) => {
```
bgavrilMS: Can we have the "opposite" test as well? I.e. "if the error returned from AAD is not `invalid_instance` then ... "

bgavrilMS:
```suggestion
            it("throws untrustedAuthority error if host is not part of knownAuthorities, cloudDiscoveryMetadata and instance discovery network call doesn't return metadata, and the error returned from AAD is 'invalid_instance'", (done) => {
```

bgavrilMS (approved):
Approved with suggestions - more logging and extra unit test

bgavrilMS:
Authority validation is said to be failed only if AAD responds with HTTP 400 (which MSALs interpret as Service Error) with error code "invalid_instance". 

Any other issue, such as transient HTTP 5xx errors, should not be considered authority validation exceptions. 

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs

lib/msal-common/src/authority/Authority.ts
```diff
@@ -543,17 +544,25 @@ export class Authority {
...
+                if (isCloudInstanceInvalid(typedResponseBody.error)) {
+                    return null;
+                }
+                
+                metadata = [];
             }
+
             match = Authority.getCloudDiscoveryMetadataFromNetworkResponse(
                 metadata,
                 this.hostnameAndPort
```
pmaytak, line 582: Log the error inside the catch on Line 570, otherwise, if we return null, we may not know why.

lib/msal-common/src/authority/Authority.ts
Robbie-Microsoft: I've added more logs in the other PR that is specifically for logging.

lib/msal-common/src/authority/Authority.ts
Robbie-Microsoft: It will never be null thanks to the new functionality we implemented in HttpClient

lib/msal-common/src/authority/Authority.ts
```diff
@@ -543,17 +544,25 @@ export class Authority {
...
+            
+            let typedResponseBody: CloudInstanceDiscoveryResponse | CloudInstanceDiscoveryErrorResponse;
+            let metadata: Array<CloudDiscoveryMetadata>;
+            if (isCloudInstanceDiscoveryResponse(response.body)) {
+                typedResponseBody = response.body as CloudInstanceDiscoveryResponse;
+                metadata = typedResponseBody.metadata;
+            } else {
+                typedResponseBody = response.body as CloudInstanceDiscoveryErrorResponse;
+                if (isCloudInstanceInvalid(typedResponseBody.error)) {
+                    return null;
```
pmaytak, line 566: Add log here. Otherwise, if we return null, we may not know why.

lib/msal-common/src/authority/Authority.ts
pmaytak: +1. The logs should cover all the scenarios in this class - why certain code path was taken (if it's not already logged in other classes like Authority). For ex: log why we return null.

lib/msal-common/src/authority/Authority.ts
Robbie-Microsoft: I have another PR open for logging (specifically this block of code). This PR is just for the improvement.

lib/msal-common/src/authority/Authority.ts
Robbie-Microsoft: I have another PR open for logging (specifically this block of code). This PR is just for the improvement.

bgavrilMS (approved):

pmaytak (approved):

lib/msal-common/src/authority/CloudInstanceDiscoveryErrorResponse.ts
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+/**
+ * The OpenID Configuration Endpoint Response type. Used by the authority class to get relevant OAuth endpoints.
+ */
+export type CloudInstanceDiscoveryErrorResponse = {
+    error: String;
```
tnorling, line 10: These should all be optional or write a validation function which verifies required params are present in the response.body

lib/msal-common/src/authority/CloudInstanceDiscoveryErrorResponse.ts
```diff
@@ -0,0 +1,21 @@
...
+    error_description: String;
+    error_codes: Array<Number>;
+    timestamp: String;
+    trace_id: String;
+    correlation_id: String;
+    error_uri: String;
+};
+
+export function isCloudInstanceInvalid(error: String): boolean {
+    return error === "invalid_instance";
```
tnorling: Let's make this a constant instead of a helper function

lib/msal-common/src/authority/Authority.ts
```diff
@@ -543,27 +544,49 @@ export class Authority {
...
+            let metadata: Array<CloudDiscoveryMetadata>;
+            if (isCloudInstanceDiscoveryResponse(response.body)) {
+                typedResponseBody = response.body as CloudInstanceDiscoveryResponse;
+                metadata = typedResponseBody.metadata;
+
+                this.logger.verbosePii(`tenant_discovery_endpoint is: ${typedResponseBody.tenant_discovery_endpoint}`);
+            } else {
+                this.logger.warning(`A CloudInstanceDiscoveryErrorResponse was returned. The cloud instance discovery network request's status code is: ${response.status}`);
+
+                typedResponseBody = response.body as CloudInstanceDiscoveryErrorResponse;
```
tnorling, line 563: Validate the type matches before casting

lib/msal-common/src/authority/Authority.ts
```diff
@@ -543,27 +545,58 @@ export class Authority {
...
+                if (typedResponseBody.error === Constants.CLOUD_INSTANCE_DISCOVERY_INVALID_INSTANCE_ERROR) {
+                    this.logger.error("The CloudInstanceDiscoveryErrorResponse error is invalid_instance.");
+                    return null;
+                }
+
+                this.logger.warning(`The CloudInstanceDiscoveryErrorResponse error is ${typedResponseBody.error}`);
+                this.logger.warning(`The CloudInstanceDiscoveryErrorResponse error description is ${typedResponseBody.error_description}`);
+                
+                this.logger.warning("Setting the value of the CloudInstanceDiscoveryMetadata (returned form the network) to []");
```
tnorling:
```suggestion
                this.logger.warning("Setting the value of the CloudInstanceDiscoveryMetadata (returned from the network) to []");
```

lib/msal-common/src/utils/Constants.ts
```diff
@@ -51,7 +51,8 @@ export const Constants = {
     KNOWN_PUBLIC_CLOUDS: ["login.microsoftonline.com", "login.windows.net", "login.microsoft.com", "sts.windows.net"],
     TOKEN_RESPONSE_TYPE: "token",
     ID_TOKEN_RESPONSE_TYPE: "id_token",
-    SHR_NONCE_VALIDITY: 240
+    SHR_NONCE_VALIDITY: 240,
+    CLOUD_INSTANCE_DISCOVERY_INVALID_INSTANCE_ERROR: "invalid_instance",
```
tnorling:
```suggestion
    INVALID_INSTANCE: "invalid_instance",
```

tnorling (approved):
Couple last nits but looks good!

PR #5454, opened 2022-12-05
Update engine.io to resolve dependabot alert

hectormmg:


lib/msal-angular/package.json
```diff
@@ -66,7 +66,8 @@
     "ts-node": "~8.3.0",
     "tslib": "^2.0.0",
     "typescript": "~4.1.5",
-    "zone.js": "~0.10.3"
+    "zone.js": "~0.10.3",
+    "engine.io": ">=6.2.1"
```
derisen, line 70: just curious, isn't this a nested dependency? Shouldn't we use [overrides](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#overrides) for this?

tnorling, line 70: Don't think we should add a dependency to resolve this. I don't even see this dep in the package-lock so I'm not sure I understand where the alert came from. Maybe try re-running?

PR #5469, opened 2022-12-08
Bump express in samples to resolve dependabot alert

hectormmg:


derisen (approved):

samples/msal-browser-samples/MsrCryptoSample/package.json
```diff
@@ -8,7 +8,7 @@
   "dependencies": {
     "cookie-parser": "~1.4.4",
     "debug": "~2.6.9",
-    "express": "~4.16.1",
+    "express": "~4.17.3",
```
tnorling, line 11: Does this have to be ~ version? Can we make it ^ so we don't have to make this kind of change again in the future?

tnorling:
This was addressed in #5503

PR #5472, opened 2022-12-09
Added functionality to MSAL-Node's HttpClient that supports custom http(s) agents

### LINKED ISSUE #5295: Setting custom https.Agent for msal-node
### Core Library

MSAL Node (@azure/msal-node)

### Core Library Version

1.14.2

### Wrapper Library

Not Applicable

### Wrapper Library Version

None

Confidential

### Description

Is it possible to set a custom https.Agent for msal-node confidential client? I see there is a way to set networkClient as INetworkModule. Can the default INetworkModule be extended to use custom https.Agent?

### MSAL Configuration

_No response_

### Relevant Code Snippets

_No response_

### Identity Provider

_No response_

### Source

Internal (Microsoft)

Robbie-Microsoft:


lib/msal-node/src/config/Configuration.ts
```diff
@@ -67,6 +68,7 @@ export type NodeSystemOptions = {
     loggerOptions?: LoggerOptions;
     networkClient?: INetworkModule;
     proxyUrl?: string;
+    customAgentOptions?: CustomAgentOptions;
```
bgavrilMS: 1. Not sure about the prefix "custom". Maybe "httpAgentOptions" ?
~2. Should we group proxyUrl and this field in an object, similar to "LoggerOptions" ?~

lib/msal-common/src/authority/Authority.ts
```diff
@@ -51,14 +51,17 @@ export class Authority {
...
+    private customAgentOptions: CustomAgentOptions;
 
     constructor(
         authority: string,
         networkInterface: INetworkModule,
         cacheManager: ICacheManager,
         authorityOptions: AuthorityOptions,
         logger: Logger,
-        proxyUrl?: string
+        proxyUrl?: string,
```
bgavrilMS: Seems a bit odd that Authority class has to change every time we add some field related to HTTP. This violates the S and O principles of SOLID (https://en.wikipedia.org/wiki/SOLID).

I would expect that all these details, including proxyUrl and agent options to be encapsulated inside `networkInterface` (without changing  INetworkModule interface, as that's a breaking change).

lib/msal-node/src/config/Configuration.ts
Robbie-Microsoft: Regarding #2, I like that idea. I'm curious as to what @tnorling and @sameerag think about that.

lib/msal-common/src/authority/Authority.ts
Robbie-Microsoft: I'm not sure I understand the second part of your comment. The network interface is the (optionally) user-custom-defined HttpClient. The Authority class has changed because we need to pass the proxyUrl/customAgentOptions from the (optional) user defined system config to the HttpClient.

change/@azure-msal-common-23d68d34-05ea-4e90-9a2c-8e1b2774e6b2.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Added functionality to MSAL-Node's HttpClient that supports custom http(s) agents",
```
tnorling: To make this a little less verbose:
```suggestion
  "comment": "Add support for custom http(s) agents #5472",
```

change/@azure-msal-node-8199266d-e536-4ec5-8e6a-161cb4f53f59.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Added functionality to MSAL-Node's HttpClient that supports custom http(s) agents",
```
tnorling:
```suggestion
  "comment": "Add support for custom http(s) agents #5472",
```

lib/msal-common/src/authority/Authority.ts
tnorling: Tend to agree here - these sorts of details should be handled by the network interface rather than the caller. Ideally the caller (Authority in this case) shouldn't be exposed to this or expected to understand what to do with this @Robbie-Microsoft The network interface *can* be a custom defined implementation of the interface, however, if it is not provided a default (our implementation of HttpClient) is passed in instead.

lib/msal-node/src/config/Configuration.ts
tnorling: While I am in favor of better organization here, unfortunately `proxyUrl` is public at this point so we can't move it without a breaking change. If that's the direction we want to go we have two options:

1. Create a backlog item for Msal-Node v2 as a reminder to make this change when we're ready for v2
2. Create the new property bucket now in addition to the existing parameter, marking the existing parameter deprecated while maintaining the functionality until v2 when we can remove the old param.

I'm generally not in favor of double dipping on params though if we think there will be more options like this added in the near future it may be worth creating that new bucket now to get ahead of that

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -75,6 +75,12 @@ const networkRequestViaProxy = <T>(
         tunnelRequestOptions.timeout = timeout;
     }
 
+    if (options?.customAgentOptions && Object.keys(options.customAgentOptions).length) {
+        tunnelRequestOptions.agent = new http.Agent({
```
tnorling: Generally param reassignment like this is a no-no. You should create a new local variable if you need to modify properties

lib/msal-node/src/config/Configuration.ts
bgavrilMS: I think we only expose `NetworkRequestOptions` for implementers of `INetworkModule`. But we do not expect ppl to set proxyUrl or agentOptions here. 

We expect customers to set proxyUrl via `SystemOptions`. And this is where we should expose httpsAgentOptions as well.

Architecturally, proxyUrl / httpAgentOptions should only be exposed on `SystemOptions` and injected in implementations of `INetworkModule` via the constuctor. They are not part of NetworkOptions. 

@tnorling @sameerag - what deprecation options do we have here? Can we add a comment that "proxyUrl" from `NetworkOptions` will not be used?

bgavrilMS (changes requested):
1. new option should be on SystemOptions
2. do not expose it on NetoworkOptions

lib/msal-node/src/config/Configuration.ts
tnorling: `NetworkRequestOptions` is set by `msal-node`, not by developers. We only expose the type so that implementers of `INetworkModule` know what the object looks like when they need to use it but the contents of the object are controlled by us. All configuration values controlled by the developer go though the objects defined in this file.

lib/msal-node/src/network/HttpClient.ts
Robbie-Microsoft: I'm modifying a new local variable that I created.

lib/msal-node/src/config/Configuration.ts
Robbie-Microsoft: @bgavrilMS regarding #1, the agent options can be http or https, so I don't think it makes sense to call it httpAgentOptions.

Robbie-Microsoft: @bgavrilMS regarding #1, the agent options don't get converted into http vs https until the httpClient where we determine if the proxyUrl was passed in. Therefore, I don't think it makes sense to name it "httpAgentOptions" - it's only http when a proxy url has been specified.

lib/msal-node/src/config/Configuration.ts
```diff
@@ -67,6 +69,7 @@ export type NodeSystemOptions = {
     loggerOptions?: LoggerOptions;
     networkClient?: INetworkModule;
     proxyUrl?: string;
+    customAgentOptions?: http.AgentOptions | https.AgentOptions;
```
bgavrilMS: Is "http" supported / needed? MSAL should only use https.

lib/msal-node/src/config/Configuration.ts
```diff
@@ -157,11 +161,15 @@ export function buildAppConfiguration({
     system,
     telemetry
 }: Configuration): NodeConfiguration {
+    const systemOptions: Required<NodeSystemOptions> = {
+        ...DEFAULT_SYSTEM_OPTIONS,
+        networkClient: system?.customAgentOptions ? new HttpClient(system.customAgentOptions as http.AgentOptions | https.AgentOptions) : new HttpClient(),
```
bgavrilMS: If the app developer configures their own HttpClient, will that scenario still work?

lib/msal-node/src/config/Configuration.ts
bgavrilMS: Oh I see, it's for the proxy stuff

lib/msal-node/src/config/Configuration.ts
bgavrilMS: If I inject a custom HTTP client AND customAgentOptions, what will happen? Is `customAgentOptions` ignored ( I assume so).

bgavrilMS: I think this aspect needs to be covered by a test: 

-  If I inject a custom HTTP client, will MSAL still use the custom HTTP client?

Robbie-Microsoft: > If the app developer configures their own HttpClient, will that scenario still work?

What scenario?

Robbie-Microsoft: > If I inject a custom HTTP client AND customAgentOptions, what will happen? Is `customAgentOptions` ignored ( I assume so).

If the developer injects their own custom HttpClient, then the customAgentOptions are ignored/never used. In that case, the developer can implement their own HttpAgent.

Robbie-Microsoft: I can confirm that a developer's custom HttpClient is still used even when CustomAgentOptions are included in the system config.

Robbie-Microsoft: I'm confused. You're asking for a test that confirms a custom INetworkModule is used when provided? That should already exist.

bgavrilMS (approved):

lib/msal-node/docs/configuration.md
```diff
@@ -86,6 +87,7 @@ const msalInstance = new PublicClientApplication(msalConfig);
 | `loggerOptions` | Config object for logger. | See [below](#logger-config-options). | See [below](#logger-config-options). |
 | `NetworkClient` | Custom HTTP implementation | INetworkModule | [HttpClient.ts](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/src/network/HttpClient.ts) |
 | `proxyUrl` | The URL of the proxy the app is running behind | string | Empty string `""` |
+| `customAgentOptions` | Set of configurable options to set on a http(s) agent | Object - [NodeJS documentation on alloweable options](https://nodejs.org/docs/latest-v16.x/api/http.html#new-agentoptions) | Empty Object `{}` |
```
sameerag, line 90: Why isn't this named as `httpAgentCustomOptions` or `httpAgentOptions` as suggested elsewhere?

lib/msal-node/docs/faq.md
```diff
@@ -55,6 +55,9 @@ MSAL Node supports self-service sign-up in the auth code flow. Please see our do
 ### Why doesn't my app function correctly when it's running behind a proxy?
 Developers can provide a `proxyUrl` string in the system config options as detailed [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/docs/configuration.md#system-config-options). Developers can also implement their own NetworkManager by instantiating an [INetworkModule](https://azuread.github.io/microsoft-authentication-library-for-js/ref/interfaces/_azure_msal_common.inetworkmodule.html) and building proxy support in it.
 
+### How do I implement a custom http(s) agent in MSAL Node?
+Developers can use a custom http(s) agent by providing a `customAgentOptions` object in the system config options as detailed [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/docs/configuration.md#system-config-options). Developers can also implement their own NetworkManager by instantiating an [INetworkModule](https://azuread.github.io/microsoft-authentication-library-for-js/ref/interfaces/_azure_msal_common.inetworkmodule.html) and building custom http(s) agent support in it.
```
sameerag, line 59: Can we also add a little description: say, this option is a mechanism for developers to control the shape of the `http(s)Agent` that would be built by the library?

sameerag (approved):
Minor naming related comments, approved.

lib/msal-node/docs/configuration.md
Robbie-Microsoft: because the agent can be http or https

PR #5476, opened 2022-12-10
Add event vis change listener in PopupClient

bmahall:
This PR-

Adds a boolean(visChange) to performance data to indicate when Page Visibility changes during the AcquireTokenPopup flow.
This covers the AcquireTokenPopup event only.


tnorling:
I think this should be implemented for silent calls, rather than interactive calls

bmahall:
Closing as is redundant based on feedback, and alternative covered by #5488

PR #5484, opened 2022-12-14
2130019. Fix cache credential look-up.

### LINKED ISSUE #5313: Azure AD B2C user flow name can break token cache
### Core Library

### Core Library Version

2.30.0

### Wrapper Library

### Wrapper Library Version

2.4.5

Public

### Description

The implementation of [`getCredentialType`](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-common/src/cache/entities/CredentialEntity.ts#L110) is ambiguous and can result in false positives.

The reason for this is that the `key` that is passed to the function contains the string an Azure user inputs for its user flow. When the user names his user flow for example `B2C_1_AccessToken`, this string will be part of the `key` (in lowercase), i.e. the key will look something like `...-b2c_1_accesstoken-...`. Thus the `getCredentialType` function will always return `CredentialType.ACCESS_TOKEN` as type of the key, as it just checks if the string `'accesstoken'` is present in `key`.

This actually happened to me as my admin used such a name for the user flow. I was implementing a token refresh in my application using `MsalService.acquireTokenSilent`.

Some more details: According to my knowledge, msal stores the idtoken, access token and refresh token in its cache (in my case in the browsers local storage). As the cache is a <key,value>-store, it needs a `key` to identify the objects that represent those tokens. The key composed of:
`[SOME-UUID]-[NAME_OF_AZURE_USER_FLOW].[AZURE_DIRECTORY_TENANT_ID]-[AZURE_AD_B2C_HOSTNAME]-[TYPE_OF_TOKEN]-[AZURE_APPLICATION_CLIENT_ID]----`
Where TYPE_OF_TOKEN is one of `idtoken`, `accesstoken` or `refreshtoken`. The current implementation of `getCredentialType` assumes that all other parts of the key do not contain one of those strings, which is actually not correct, as the Azure user flow name is set by the user and may contain those strings.

### Error Message

_No response_

### Msal Logs

_No response_

### MSAL Configuration

```javascript
{
        auth: {
            clientId: environment.AZURE_APPLICATION_CLIENT_ID,
            authority,
            redirectUri,
            postLogoutRedirectUri,
        },
        cache: {
            storeAuthStateInCookie: false, // Set to true for Internet Explorer 11
        },
        system: {
            loggerOptions: {
                logLevel: LogLevel.Verbose,
                piiLoggingEnabled: false,
            },
        },
    });
}
```

### Relevant Code Snippets

```javascript
-
```

### Reproduction Steps

1. Setup an Azure user flow with the name `B2C_1_AccessToken`
2. Implement a token refresh using `MsalService.acquireTokenSilent`

### Expected Behavior

It will refresh the token.

### Identity Provider

Azure B2C Basic Policy

Chrome

### Regression

_No response_

### Source

External (Customer)

konstantin-msft:
Changes:
 - Fix cache credential look-up.
 - Add unit tests.

### DIFF
--- a/lib/msal-common/src/cache/entities/CredentialEntity.ts
+++ b/lib/msal-common/src/cache/entities/CredentialEntity.ts
@@ -109,13 +109,12 @@ export class CredentialEntity {
      */
     static getCredentialType(key: string): string {
         const separator = Separators.CACHE_KEY_SEPARATOR;
+        // Match host names like "login.microsoftonline.com", "https://accounts.google.com:4000", etc.
         const domainRe = "(https?:\\/\\/)?([\\w-]+\\.)*([\\w-]{1,63})(\\.(\\w{2,3}))(\\:[0-9]{4,5})?";
 
-        for (const credType of Object.entries(CredentialType)) {
-            const credKey = credType[0];
-            const credVal = credType[1].toLowerCase();
+        for (const credKey of Object.keys(CredentialType)) {
+            const credVal = CredentialType[credKey].toLowerCase();
             // Verify credential type is preceded by a valid host name (environment)
-            // eslint-disable-next-line security/detect-non-literal-regexp
             if (key.toLowerCase().search(`(?<=${separator}${domainRe})${separator}${credVal}${separator}`) !== -1) {
                 return CredentialType[credKey];
             }
--- a/lib/msal-common/src/cache/entities/CredentialEntity.ts
+++ b/lib/msal-common/src/cache/entities/CredentialEntity.ts
@@ -108,17 +108,17 @@ export class CredentialEntity {
      * @param key
      */
     static getCredentialType(key: string): string {
-        // First keyword search will match all "AccessToken" and "AccessToken_With_AuthScheme" credentials
-        if (key.indexOf(CredentialType.ACCESS_TOKEN.toLowerCase()) !== -1) {
-            // Perform second search to differentiate between "AccessToken" and "AccessToken_With_AuthScheme" credential types
-            if (key.indexOf(CredentialType.ACCESS_TOKEN_WITH_AUTH_SCHEME.toLowerCase()) !== -1) {
-                return CredentialType.ACCESS_TOKEN_WITH_AUTH_SCHEME;
+        const separator = Separators.CACHE_KEY_SEPARATOR;
+        const domainRe = "(https?:\\/\\/)?([\\w-]+\\.)*([\\w-]{1,63})(\\.(\\w{2,3}))(\\:[0-9]{4,5})?";
+
+        for (const credType of Object.entries(CredentialType)) {
+            const credKey = credType[0];
+            const credVal = credType[1].toLowerCase();
+            // Verify credential type is preceded by a valid host name (environment)
+            // eslint-disable-next-line security/detect-non-literal-regexp
+            if (key.toLowerCase().search(`(?<=${separator}${domainRe})${separator}${credVal}${separator}`) !== -1) {
+                return CredentialType[credKey];
             }
-            return CredentialType.ACCESS_TOKEN;
-        } else if (key.indexOf(CredentialType.ID_TOKEN.toLowerCase()) !== -1) {
-            return CredentialType.ID_TOKEN;
-        } else if (key.indexOf(CredentialType.REFRESH_TOKEN.toLowerCase()) !== -1) {
-            return CredentialType.REFRESH_TOKEN;
         }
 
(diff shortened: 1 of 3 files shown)


lib/msal-common/src/cache/entities/CredentialEntity.ts
```diff
@@ -108,17 +108,17 @@ export class CredentialEntity {
      * @param key
      */
     static getCredentialType(key: string): string {
-        // First keyword search will match all "AccessToken" and "AccessToken_With_AuthScheme" credentials
-        if (key.indexOf(CredentialType.ACCESS_TOKEN.toLowerCase()) !== -1) {
-            // Perform second search to differentiate between "AccessToken" and "AccessToken_With_AuthScheme" credential types
-            if (key.indexOf(CredentialType.ACCESS_TOKEN_WITH_AUTH_SCHEME.toLowerCase()) !== -1) {
-                return CredentialType.ACCESS_TOKEN_WITH_AUTH_SCHEME;
+        const separator = Separators.CACHE_KEY_SEPARATOR;
+        const domainRe = "(https?:\\/\\/)?([\\w-]+\\.)*([\\w-]{1,63})(\\.(\\w{2,3}))(\\:[0-9]{4,5})?";
```
tnorling, line 113: Can you include a comment with an example or two of what this is trying to match?

lib/msal-common/src/cache/entities/CredentialEntity.ts
```diff
@@ -108,17 +108,17 @@ export class CredentialEntity {
...
+
+        for (const credType of Object.entries(CredentialType)) {
```
tnorling: I don't think Object.entries is supported in IE11/Legacy Edge which we, unfortunately, still need to support until msal-browser v3 to avoid breaking changes. Can we use Object.keys instead?

lib/msal-common/src/cache/entities/CredentialEntity.ts
```diff
@@ -108,17 +108,17 @@ export class CredentialEntity {
...
+            const credKey = credType[0];
+            const credVal = credType[1].toLowerCase();
+            // Verify credential type is preceded by a valid host name (environment)
+            // eslint-disable-next-line security/detect-non-literal-regexp
```
tnorling: This is a rule we might not get away with disabling as the code which violates this rule will likely get flagged in the CodeQL scans. Is there a different way we can do this?

lib/msal-common/src/cache/entities/CredentialEntity.ts
konstantin-msft: Good point. Replaced with `Object.keys()`.

lib/msal-common/src/cache/entities/CredentialEntity.ts
konstantin-msft: Fixed.

lib/msal-common/src/cache/entities/CredentialEntity.ts
konstantin-msft: Added a few samples.

tnorling (approved):

derisen (approved):
LGTM! I think this fixes #5313, so you might wanna link that issue

lib/msal-common/src/cache/entities/CredentialEntity.ts
```diff
@@ -108,17 +108,16 @@ export class CredentialEntity {
...
      */
     static getCredentialType(key: string): string {
-        // First keyword search will match all "AccessToken" and "AccessToken_With_AuthScheme" credentials
-        if (key.indexOf(CredentialType.ACCESS_TOKEN.toLowerCase()) !== -1) {
-            // Perform second search to differentiate between "AccessToken" and "AccessToken_With_AuthScheme" credential types
-            if (key.indexOf(CredentialType.ACCESS_TOKEN_WITH_AUTH_SCHEME.toLowerCase()) !== -1) {
-                return CredentialType.ACCESS_TOKEN_WITH_AUTH_SCHEME;
+        const separator = Separators.CACHE_KEY_SEPARATOR;
+        // Match host names like "login.microsoftonline.com", "https://accounts.google.com:4000", etc.
+        const domainRe = "(https?:\\/\\/)?([\\w-]+\\.)*([\\w-]{1,63})(\\.(\\w{2,3}))(\\:[0-9]{4,5})?";
```
adamblok, line 113: This pattern produces error in Safari browsers, see: https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/5561

PR #5492, opened 2022-12-20
Migrate `msal-browser` VanillaJS tests from `mocha` to `jest`

konstantin-msft:
Changes:
- Migrate VanillaJS tests from `mocha` to `jest`
- Bind ports dynamically for `SAMPLE_HOME_URL`.
- Derive instance of `browser` from `e2eTestUtils` instead of instantiating it manually.
- Remove `@babel/*` and `mocha` dev dependencies.
- Remove mocha `testRunner`.


tnorling:
> Split customizable-e2e-test test suite to avoid concurrent updates on testConfig.json file.

Regarding this point: the "customizable" aspect was that it was 1 sample that could test many different scenarios. The benefit being when we needed to make an update to the sample we only had to make that update in one place rather than for each individual scenario. I'd prefer to keep that pattern, in fact I'd love it if we could roll some of the other samples into this one as well (not in this PR necessarily but that's my personal sample app north star).

The idea I had to make this work was to serve each "scenario" on a separate route that could be used in the relevant test (i.e. http://localhost:3600/localStorage). This is roughly what we did in the React samples and it's worked pretty well. Given that's a bit more work than what this PR aims to address, an interim solution to your concurrency problem would be to configure Jest to not run tests concurrently for now. Thoughts?

samples/msal-browser-samples/VanillaJSTestApp2.0/app/customizable-e2e-aad/test/browser.spec.ts
```diff
@@ -0,0 +1,307 @@
+import puppeteer from "puppeteer";
+
+import { Screenshot, createFolder, setupCredentials, enterCredentials, ONE_SECOND_IN_MS } from "../../../../../e2eTestUtils/TestUtils";
```
peterzenz: Can we configure baseurl/paths to make the ../../../../..... a little more readable?

konstantin-msft:
> > Split customizable-e2e-test test suite to avoid concurrent updates on testConfig.json file.
> 
> Regarding this point: the "customizable" aspect was that it was 1 sample that could test many different scenarios. The benefit being when we needed to make an update to the sample we only had to make that update in one place rather than for each individual scenario. I'd prefer to keep that pattern, in fact I'd love it if we could roll some of the other samples into this one as well (not in this PR necessarily but that's my personal sample app north star).
> 
> The idea I had to make this work was to serve each "scenario" on a separate route that could be used in the relevant test (i.e. http://localhost:3600/localStorage). This is roughly what we did in the React samples and it's worked pretty well. Given that's a bit more work than what this PR aims to address, an interim solution to your concurrency problem would be to configure Jest to not run tests concurrently for now. Thoughts?

Makes sense to me. I am fine with keeping `customizable` test suite intact and running the whole test suite with `--runInBand` flag enabled. That was my initial approach actually. However, the huge downside would be that tests will take 3+ times longer to complete: ~90 seconds against ~300 seconds. Can make a huge difference considering some of those tests are flaky.

tnorling:
> Makes sense to me. I am fine with keeping `customizable` test suite intact and running the whole test suite with `--runInBand` flag enabled. That was my initial approach actually. However, the huge downside would be that tests will take 3+ times longer to complete: ~90 seconds against ~300 seconds. Can make a huge difference considering some of those tests are flaky.

I think I'm ok with that for now until we update the sample to properly support concurrency. Curious if anyone else has opinions

samples/msal-browser-samples/VanillaJSTestApp2.0/app/customizable-e2e-local-storage/auth.js
```diff
@@ -0,0 +1,94 @@
...
+
+function handleResponse(resp) {
+    if (resp !== null) {
+        homeAccountId = resp.account.homeAccountId;
+        showWelcomeMessage(resp.account);
+        if (resp.accessToken) {
+            updateUI(resp);
+        }
+    } else {
+        // need to call getAccount here?
```
peterzenz: Yes.

peterzenz (approved):

samples/msal-browser-samples/VanillaJSTestApp2.0/app/customizable-e2e-aad/test/browser.spec.ts
konstantin-msft: We can. However, since `jest` transforms `ts` files into `js`, it requires maintaining `moduleNameMapper` for each innermost `jest.config.js`.  [tsconfig-paths](https://www.npmjs.com/package/tsconfig-paths) does not work with `jest` either. I would suggest sticking to relative paths for simplicity.

konstantin-msft:
> > Makes sense to me. I am fine with keeping `customizable` test suite intact and running the whole test suite with `--runInBand` flag enabled. That was my initial approach actually. However, the huge downside would be that tests will take 3+ times longer to complete: ~90 seconds against ~300 seconds. Can make a huge difference considering some of those tests are flaky.
> 
> I think I'm ok with that for now until we update the sample to properly support concurrency. Curious if anyone else has opinions

Merged `customizable-e2e-test` back together since there were no additional comments from anyone else.

tnorling (approved):

PR #5497, opened 2022-12-21
Handle storage errors gracefully

bmahall:
Adds error handling to setItem()


change/@azure-msal-browser-03d446bd-1961-4473-b890-89912abebee8.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
tnorling: Adding the logger to the CacheManager makes this minor and you'll need to update the 1P lib as well
```suggestion
  "type": "minor",
```

lib/msal-browser/src/cache/BrowserStorage.ts
```diff
@@ -31,7 +35,17 @@ export class BrowserStorage implements IWindowStorage<string> {
     }
 
     setItem(key: string, value: string): void {
-        this.windowStorage.setItem(key, value);
+        try{
+            this.windowStorage.setItem(key, value);
+        }
+        catch(e){
+            if(e.code === BrowserAuthErrorMessage.chromiumStorageException.code || e.code === BrowserAuthErrorMessage.firefoxStorageException.code)
```
tnorling: What about other browsers? Is there no standard error?

lib/msal-browser/src/cache/BrowserStorage.ts
```diff
@@ -6,14 +6,18 @@
 import { BrowserConfigurationAuthError } from "../error/BrowserConfigurationAuthError";
 import { BrowserCacheLocation } from "../utils/BrowserConstants";
 import { IWindowStorage } from "./IWindowStorage";
+import { BrowserAuthErrorMessage } from "../error/BrowserAuthError";
+import { Logger } from "@azure/msal-common";
 
 export class BrowserStorage implements IWindowStorage<string> {
 
     private windowStorage: Storage;
+    private logger:Logger;
```
tnorling: Can you check the spacing here and in a couple spots below please

lib/msal-browser/src/error/BrowserAuthError.ts
```diff
@@ -184,6 +184,14 @@ export const BrowserAuthErrorMessage = {
     nativePromptNotSupported: {
         code: "native_prompt_not_supported",
         desc: "The provided prompt is not supported by the native platform. This request should be routed to the web based flow."
+    },
+    chromiumStorageException: {
+        code: 22,
```
tnorling, line 189: Do we have to define this ourselves or is there some other variable we can pull from? Are we confident this is stable?

lib/msal-browser/src/cache/BrowserStorage.ts
```diff
@@ -31,7 +35,17 @@ export class BrowserStorage implements IWindowStorage<string> {
...
 
     setItem(key: string, value: string): void {
-        this.windowStorage.setItem(key, value);
+        try{
+            this.windowStorage.setItem(key, value);
+        }
+        catch(e){
+            if(e.code === BrowserAuthErrorMessage.chromiumStorageException.code || e.code === BrowserAuthErrorMessage.firefoxStorageException.code)
+            {
+                this.logger.error("Could not access browser storage. This may be caused by exceeding the quota.");
```
tnorling: Should we attempt to clean up some of our own entries here and retry?

tnorling: In fact I kind of think we should. Not doing so may result in other unintended errors or behaviors, for example temporary cache entries being missing at the end of the flow when they're needed or `getAllAccounts` returning an empty array after a  successful login. I also wonder if there might be a way we can signal to the caller that storage was full so they can do further cleanup on their own, perhaps by using the event API?

lib/msal-browser/src/cache/BrowserStorage.ts
bmahall: There are browser specific error codes , and error most common being QuotaExceededError and QUOTA_EXCEEDED_ERR. Alternatively , to avoid using this if check, any exception thrown during setItem() can be regarded as a storage error since there is already a method call to validate the existence of storage in the constructor. 
I couldnt find a standard error name or code that is applicable across all browser environments.

lib/msal-browser/src/cache/BrowserStorage.ts
bmahall: Have modified to include event handler to emit event which can signal the user to do clean on their own, as for the first part, I am now calling clear() to clear all the msal items in storage, do you have any suggestions on how to retry after clearing the storage?

bmahall: Can we prompt the user to use a custom storage option in this scenario?

lib/msal-browser/src/cache/BrowserStorage.ts
bmahall: [storage api example reference](https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API)

lib/msal-browser/src/cache/BrowserCacheManager.ts
```diff
@@ -157,7 +160,30 @@ export class BrowserCacheManager extends CacheManager {
...
     setItem(key: string, value: string): void {
-        this.browserStorage.setItem(key, value);
+        try{
+            this.browserStorage.setItem(key, value);
+        }
+        catch(e){
+            // check for quota exceeding error 
+            if(e.code === BrowserAuthErrorMessage.chromiumStorageException.code || e.code === BrowserAuthErrorMessage.firefoxStorageException.code || e.name === "QuotaExceededError" || e.name === "NS_ERROR_DOM_QUOTA_REACHED")
+            {
+                this.logger.error("Could not access browser storage. This may be caused by exceeding the quota.");
```
konstantin-msft, line 170: Optional. Since this message suggests that an error "may be caused by exceeding the quota" it could be easier just to check if an error message contains "quota" to avoid referencing to 3rd party error codes that can change in the future (`22` for chrome and `1014` for FF).

lib/msal-browser/src/cache/BrowserStorage.ts
konstantin-msft: I think it would be great if we could maintain cache eviction policy for our own keys. Cleaning up all of them at once in the middle of the flow may produce unexpected results. Please correct me if I am wrong.

bmahall: @konstantin-msft I agree, based on discussion with @tnorling yesterday, it would be better to have a design document to discuss the eviction policies and implications on all flows.Thanks guys for feedback!

tnorling:
Closing for now

PR #5499, opened 2022-12-22
Updating the package

mantupani:
Updated the jsonwebtoken package as per the security issue.

### DIFF
--- a/lib/msal-node/package.json
+++ b/lib/msal-node/package.json
@@ -59,7 +59,7 @@
   },
   "dependencies": {
     "@azure/msal-common": "^9.0.1",
-    "jsonwebtoken": "9.0.0",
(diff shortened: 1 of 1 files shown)


lib/msal-node/package.json
```diff
@@ -59,7 +59,7 @@
   },
   "dependencies": {
     "@azure/msal-common": "^9.0.1",
-    "jsonwebtoken": "^8.5.1",
+    "jsonwebtoken": "9.0.0",
```
rparpa:
```suggestion
    "jsonwebtoken": "^9.0.0",
```

tnorling:
Thanks for opening this! Closing in favor of #5503 to ensure our full test suite is run against this change.

PR #5518, opened 2023-01-02
Use async context in msal interceptor specs

bjorn-einar-bjartnes-4ss:
Using timeouts, I find, makes the tests a bit hard to reason about. Angular provides a method for making async code synchronous, and one can tick to next or a number of millisecond https://angular.io/api/core/testing/fakeAsync.

Using done() can the be removed, also from the synchronous calls.

Before it takes approximately 4 seconds to run tests
image

After the changes it runs in approximately 0.15 seconds.
msal after

The reason I care is mostly that we have made or own customization as suggested in https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/contributing.md#build-library 
```Please note that the MsalInterceptor is provided for your convenience and may not fit all use cases. We encourage you to write your own interceptor if you have specific needs that are not addressed by the MsalInterceptor.
```
and by starting with the same spec file, I found the tests to be a bit slower than I liked.

It doesn't make a huge different when running from the command line, but when running the test runner continuously it is much snappier 
image


lib/msal-angular/src/msal.interceptor.spec.ts
```diff
@@ -93,7 +92,7 @@ function initializeMsal() {
     ],
   });
 
-  interceptor = TestBed.inject(MsalInterceptor);
```
bjorn-einar-bjartnes-4ss, line 96: This was never read and gave a warning in VS Code.

bmahall:
@bjorn-einar-bjartnes-4ss thanks for the contribution! While it's under review, can you please run "npm run beachball:change" in the root to add a changefile as well?

bjorn-einar-bjartnes-4ss:
@bmahall - I ran the beachball command after the build step failed. I might have done it incorrectly, but when I run it now it does say 
```
❯ npm run beachball:change

> beachball:change
> beachball change --branch origin/dev

Checking for changes against "origin/dev"
fetching latest from remotes "origin/dev"
Your local repository already has change files for these packages:
  @azure/msal-angular
"msal" has a "none" change type, no version bump is required.
"@azure/node-token-validation" has a "none" change type, no version bump is required.
Skipping package dep validation - package @azure/node-token-validation has change type as none
Validating no private package among package dependencies
 OK!

No change files are needed
```
image

lib/msal-angular/src/msal.interceptor.spec.ts
```diff
@@ -301,37 +295,34 @@ describe('MsalInterceptor', () => {
...
+    
+    tick();
+    const request = httpMock.expectOne("https://api.test.com");
+    request.flush({ data: "test" });
+    expect(request.request.headers.get("Authorization")).toEqual("Bearer access-token");
+    expect(spy).toHaveBeenCalledWith({account: sampleAccountInfo, scopes: ["default.scope1"]});
+    httpMock.verify();
+  }));
 
-  it("does not attach authorization header when scopes set to null, and resource is before any base url or wildcards", done => {
```
bjorn-einar-bjartnes-4ss, line 314: Here it seems easier to run without the callback to signal done.

bjorn-einar-bjartnes-4ss:
> 

change/@azure-msal-angular-83b38025-f900-496f-b6ff-02044ebd08c3.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Replace setTimeout with fakeAsync in specs for msal.interceptor",
```
bmahall:
```suggestion
  "comment": "Replace setTimeout with fakeAsync in specs for msal.interceptor #5518",
```

bmahall:
> @bmahall - I ran the beachball command after the build step failed. I might have done it incorrectly, but when I run it now it does say
> 
> ```
> ❯ npm run beachball:change
> 
> > beachball:change
> > beachball change --branch origin/dev
> 
> Checking for changes against "origin/dev"
> fetching latest from remotes "origin/dev"
> Your local repository already has change files for these packages:
>   @azure/msal-angular
> "msal" has a "none" change type, no version bump is required.
> "@azure/node-token-validation" has a "none" change type, no version bump is required.
> Skipping package dep validation - package @azure/node-token-validation has change type as none
> Validating no private package among package dependencies
>  OK!
> 
> No change files are needed
> ```
> 
> image

Yes, there is a changefile now, thanks!

lib/msal-angular/src/msal.interceptor.spec.ts
```diff
@@ -105,41 +104,43 @@ describe('MsalInterceptor', () => {
...
-        expect(error.errorMessage).toBe("Invalid interaction type provided to MSAL Interceptor. InteractionType.Popup, InteractionType.Redirect must be provided in the msalInterceptorConfiguration");
-        testInteractionType = InteractionType.Popup;
-        done();
+        response = error;
       }
     });
-  });
 
-  it("does not attach authorization header for unprotected resource", (done) => {
+    tick();
```
bmahall, line 119: using fakeAsync() might lead to some complications with code refactoring down the line, using async() in place of done()  might be more consistent with rest of the test base.

bjorn-einar-bjartnes-4ss: Yes, in this case I was a bit in doubt. It is fine, I guess, to use async as this doesn't require these artificial setTimeouts to block and this runs just as fast with a done().

lib/msal-angular/src/msal.interceptor.spec.ts
```diff
@@ -152,17 +153,15 @@ describe('MsalInterceptor', () => {
...
 
     httpClient.get("https://graph.microsoft.com/v1.0/me").subscribe();
-    setTimeout(() => {
-      const request = httpMock.expectOne("https://graph.microsoft.com/v1.0/me");
-      request.flush({ data: "test" });
-      expect(request.request.headers.get("Authorization")).toEqual("Bearer access-token");
-      expect(spy).toHaveBeenCalledWith({account: sampleAccountInfo, scopes: ["user.read"]});
-      httpMock.verify();
-      done();
-    }, 200);
```
bjorn-einar-bjartnes-4ss, line 162: It is mainly these timeouts I wanted to get rid of, as they slow down the tests. If it is possible to just wait 1ms, that would be good, but I have experienced issues with the test runners. I think we just want one tick of the eventloop, but it is hard to get right with timers in tests.

bjorn-einar-bjartnes-4ss:
Not that important for me tbh. Closing to avoid bot comments.

PR #5550, opened 2023-01-10
Node broker implementation

tnorling:
Exposes nativeBrokerPlugin to enable token acquisition through the system broker via msal-node-extensions and msal-runtime


lib/msal-node/src/client/PublicClientApplication.ts
```diff
@@ -89,15 +104,36 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
     async acquireTokenInteractive(request: InteractiveRequest): Promise<AuthenticationResult> {
+        const correlationId = request.correlationId || this.cryptoProvider.createNewGuid();
+        this.logger.trace("acquireTokenInteractive called", correlationId);
+        const { openBrowser, successTemplate, errorTemplate, windowHandle, ...remainingProperties } = request;
+
+        if (this.nativeBrokerPlugin) {
+            const brokerRequest: NativeRequest = {
+                ...remainingProperties,
+                clientId: this.config.auth.clientId,
+                scopes: request.scopes || OIDC_DEFAULT_SCOPES,
```
peterzenz, line 116: Do we need to add the default scopes here, or should we only add the requested ones and let the broker add any required defaults?

samples/msal-node-samples/auth-code-cli-brokered-app/README.md
```diff
@@ -0,0 +1,47 @@
...
+1. Select **Register** to create the application.
+1. In the app's registration screen, find and note the **Application (client) ID** and **Directory (Tenant) ID**. You use these values in your app's configuration file(s) later.
+
+Before running the sample, you will need to replace the values in the configuration object:
+
+```javascript
+const config = {
+    auth: {
+        clientId: "ENTER_CLIENT_ID",
+        authority: "https://login.microsoftonline.com/ENTER_TENANT_ID",
```
peterzenz: Can they use the organization/consumer/common endpoints here?  Does it have to be tenant id?

samples/msal-node-samples/auth-code-cli-brokered-app/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "msal-node-auth-code-cli-brokered-app",
+  "version": "1.0.0",
```
peterzenz, line 3: Do we want to start with 0.0.1 vs 1.0.0?  1.0.0 implies we can't change any interfaces and we might want to prior to release.

peterzenz (approved):

samples/msal-node-samples/auth-code-cli-brokered-app/package.json
tnorling: This is a sample app, versioning here doesn't really mean anything but a value is required.

samples/msal-node-samples/auth-code-cli-brokered-app/README.md
tnorling: Yes, they can use any of those but this placeholder is the standard we use across our samples

lib/msal-node/src/client/PublicClientApplication.ts
tnorling: This is an OR condition, defaults are used if no scopes were requested explicitly and this keeps the behavior consistent with our non-brokered flows.

lib/msal-node/src/client/PublicClientApplication.ts
```diff
@@ -126,4 +162,62 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
+
+        return super.acquireTokenSilent(request);
+    }
+
+    /**
+     * Removes cache artifacts associated with the given account
+     * @param request 
+     * @returns 
+     */
+    async signOut(request: SignOutRequest): Promise<void> {
```
derisen, line 200: Slightly worried if this could be mistaken for an API doing a service signout (end session). I suppose the name was chosen to align with X.. (likely lacking context, apologies)

samples/msal-node-samples/auth-code-cli-brokered-app/README.md
```diff
@@ -0,0 +1,47 @@
...
+```console
+    npm install
+```
+
+## Register
+
+1. Navigate to the [Azure portal](https://portal.azure.com) and select the **Azure AD** service.
+1. Select the **App Registrations** blade on the left, then select **New registration**.
+1. In the **Register an application page** that appears, enter your application's registration information:
+   - In the **Name** section, enter a meaningful application name that will be displayed to users of the app, for example `msal-node-webapp`.
```
derisen:
```suggestion
   - In the **Name** section, enter a meaningful application name that will be displayed to users of the app, for example `msal-node-cliapp`.
```

samples/msal-node-samples/auth-code-cli-brokered-app/README.md
derisen: +1 for id -probably shouldn't encourage multitenancy by default (though I tend to use `ENTER_TENANT_INFO` to be pedantic)

samples/msal-node-samples/auth-code-cli-brokered-app/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "scripts": {
+    "build:package": "cd ../../../lib/msal-common && npm run build && cd ../msal-node && npm run build",
+    "start": "node .",
+    "start:build": "npm run build:package && npm start"
+  },
+  "author": "Microsoft",
+  "license": "MIT",
+  "dependencies": {
+    "@azure/msal-node": "file:../../../lib/msal-node",
+    "@azure/msal-node-extensions": "file:../../../extensions/msal-node-extensions",
```
derisen, line 16: can we use published versions as default? (though I suppose relevant changes are not published yet)

samples/msal-node-samples/auth-code-cli-brokered-app/src/index.js
```diff
@@ -0,0 +1,64 @@
...
+    openBrowser,
+    successTemplate: "Successfully signed in! You can close this window now."
+};
+
+// Create msal application object
+const pca = new PublicClientApplication(authConfig);
+
+const acquireToken = async () => {
+    const accounts = await pca.getAllAccounts();
+    if (accounts.length == 1) {
```
derisen: >     if (accounts.length == 1) {

nit: can we use strict equality?

derisen (approved):
LGTM -minor comments

lib/msal-node/src/client/IPublicClientApplication.ts
```diff
@@ -51,4 +52,10 @@ export interface IPublicClientApplication {
 
     /** Clear the cache */
     clearCache(): void;
+
+    /** Gets all cached accounts */
+    getAllAccounts(): Promise<AccountInfo[]>;
+
+    /** Removes cache artifacts associated with the given account */
+    signOut(request: SignOutRequest): Promise<void>;
```
sameerag, line 60: For later reference, should we call the `logout` API in `msal-browser` also `signOut`?

lib/msal-node/src/client/PublicClientApplication.ts
peterzenz: Same question whether it's a conditional or a or'ing the collections together.  Do we really need to add these, or should we just user the broker defaults here if none are specified?  How does us not leaving this empty impact what the broker does?

tnorling: 2 reasons to do this:
1. Broker requests _require_ a value for scopes, our APIs do not
2. We provide defaults in non-brokered flows and our APIs should behave the same whether the broker is used or not. Relying on the broker's defaults instead of our own may result in different behaviors on different platforms

lib/msal-node/src/client/PublicClientApplication.ts
```diff
@@ -89,15 +104,36 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
+        const correlationId = request.correlationId || this.cryptoProvider.createNewGuid();
+        this.logger.trace("acquireTokenInteractive called", correlationId);
+        const { openBrowser, successTemplate, errorTemplate, windowHandle, ...remainingProperties } = request;
+
+        if (this.nativeBrokerPlugin) {
+            const brokerRequest: NativeRequest = {
+                ...remainingProperties,
+                clientId: this.config.auth.clientId,
+                scopes: request.scopes || OIDC_DEFAULT_SCOPES,
+                redirectUri: `${Constants.HTTP_PROTOCOL}${Constants.LOCALHOST}`,
```
sameerag, line 117: Basic q: Is it set this way as we do not need it for the native calls?

lib/msal-node/src/client/IPublicClientApplication.ts
tnorling: Let's consider that separately. For the purposes of this PR 'signOut' was chosen for consistency with MsalRuntime

samples/msal-node-samples/auth-code-cli-brokered-app/package.json
tnorling: Yes will be updated when those versions are available. Will leave this open until then.

lib/msal-node/src/client/PublicClientApplication.ts
sameerag: `signOut` API has the same name across all MSAL APIs except for`msal-browser`. What do you have in mind?

lib/msal-node/src/client/IPublicClientApplication.ts
sameerag: Sure. Called it out for vFuture. Not a blocker for this PR.

lib/msal-node/src/client/PublicClientApplication.ts
derisen: @sameerag didn't know that! My worry is, this API only does a "local" signout, it doesn't end the session with IdP. We probably should emphasize this in API description as it's reasonable to expect this to be misunderstood.

sameerag (approved):

hectormmg (approved):

PR #5488, opened 2022-12-17
Add visibility change listener for performance telemetry

bmahall:
This PR-

Adds a boolean(visibilityChange) to performance data to indicate when Page Visibility changes during the AcquireTokenSilent, AcquireTokenByCode, SsoSilent calls. 
Tests included.

PR #5555, opened 2023-01-11
Add visibility change listener for performance telemetry

konstantin-msft:
- Adds counters to `PerformanceEvent`.
- Adds a method to increment counters to `PerformanceEvent`. 
- Adds a `visibilityChangeCount` to performance data to count Page Visibility changes during the AcquireTokenSilent, AcquireTokenByCode, SsoSilent calls. 
- Tests included.

### DIFF
--- a/lib/msal-browser/src/app/ClientApplication.ts
+++ b/lib/msal-browser/src/app/ClientApplication.ts
@@ -4,7 +4,7 @@
  */
 
 import { CryptoOps } from "../crypto/CryptoOps";
-import { StringUtils, InteractionRequiredAuthError, AccountInfo, Constants, INetworkModule, AuthenticationResult, Logger, CommonSilentFlowRequest, ICrypto, DEFAULT_CRYPTO_IMPLEMENTATION, AuthError, PerformanceEvents, PerformanceCallbackFunction, StubPerformanceClient, IPerformanceClient, BaseAuthRequest, PromptValue, ClientAuthError } from "@azure/msal-common";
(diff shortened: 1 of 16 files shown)


konstantin-msft:
Copy of https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/5488. All code review comments are addressed.

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -415,6 +419,20 @@ export abstract class ClientApplication {
...
 
+    private trackPageVisibilityWithMeasurement():void {
+        const measurement = this.ssoSilentMeasurement || this.acquireTokenByCodeAsyncMeasurement;
+        if(!measurement) {
+            return;
+        }
+
+        this.logger.info("Perf: Visibility change detected in ", measurement.event.name);
+        measurement.addStaticFields({
+            visibilityChange: true,
```
peterzenz: I think the visibility change will be really diagnostic but as I'm looking at it, does anyone have any thoughts on whether we should make this a count vs a flag?  For instance, in the course of this call, the visibility changed 3x.

tnorling: I don't have a strong opinion but my thoughts are "why not"

konstantin-msft: I think we can. Do you want me to make a change in this pr or make a follow-up?

peterzenz: I'm thinking of how I use the browser and I often ctrl+click on links which tends to open them in a "hidden" tab.

Let's change it to a counter.  I'm ok with either going in this PR or the next one.  If you want to close this one out as is that's OK with me.

peterzenz (approved):

lib/msal-browser/src/app/ClientApplication.ts
sameerag: I am leaning towards the counter, which gives a more accurate status.

konstantin-msft: Done.

hectormmg (approved):
LGTM!

PR #5557, opened 2023-01-11
Refactoring sample

shoatman:
- Eliminated base class for Public client application
- Created IController interface to extend IPublicClientApplication
- Moved current PCA implementation to StandardController
- Added OperatingContext and specific operating contexts for Standards and a dummy one for MetaOS
- Added and async factory method for construction of PCA and made the existing constructor private.  OperatingContext selection and construction/inialization of implementation happens here.
- Dynamic imports are being used
- Output modules by default are code split currently using the dynamic import statement as effectively 2 different entry points... only for ES module build via rollup
- Modified rollup config to match tsconfig outDir to output directory of rollup build type, by overriding the tsconfig value via compilerOptions passed to the @rollup/typescript-plugin
- I made the constructor on public client application public again; however we need to remove this when we actually ship MetaOS... this will be a breaking change though and we said no breaking changes for v3


shoatman:
NO REVIEW NEEDED YET.  Need to fix unit tests to account for internal changes... sinon is currently being used to mock methods of public client application which in this PR have been moved to standardcontroller.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -3,281 +3,150 @@
...
-        const correlationId: string = this.getRequestCorrelationId(request);
-        this.logger.verbose("loginRedirect called", correlationId);
-        return this.acquireTokenRedirect({
-            correlationId,
-            ...(request || DEFAULT_REQUEST)
-        });
+    public constructor(configuration: Configuration, controller?: IController){
+        if(controller){
+            this.controller = controller;
+        }else{
```
hectormmg: Nit: 
```suggestion
        } else {
```

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -3,281 +3,150 @@
...
-     */
-    async loginRedirect(request?: RedirectRequest): Promise<void> {
-        const correlationId: string = this.getRequestCorrelationId(request);
-        this.logger.verbose("loginRedirect called", correlationId);
-        return this.acquireTokenRedirect({
-            correlationId,
-            ...(request || DEFAULT_REQUEST)
-        });
+    public constructor(configuration: Configuration, controller?: IController){
+        if(controller){
```
hectormmg: nit
```suggestion
        if (controller) {
```

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -3,281 +3,150 @@
...
-     * @returns A promise that is fulfilled when this function has completed, or rejected if an error was raised.
-     */
-    loginPopup(request?: PopupRequest): Promise<AuthenticationResult> {
-        const correlationId: string = this.getRequestCorrelationId(request);
-        this.logger.verbose("loginPopup called", correlationId);
-        return this.acquireTokenPopup({
-            correlationId,
-            ...(request || DEFAULT_REQUEST)
-        });
+    initialize(): Promise<void> {
```
hectormmg: Do we want to annotate as async for consistency?

```suggestion
   async initialize(): Promise<void> {
```

lib/msal-browser/src/index.ts
```diff
@@ -12,14 +12,21 @@
  * Warning: This set of exports is purely intended to be used by other MSAL libraries, and should be considered potentially unstable. We strongly discourage using them directly, you do so at your own risk.
  * Breaking changes to these APIs will be shipped under a minor version, instead of a major version.
  */
-import * as internals from "./internals";
-export { internals };
+/*
```
hectormmg, line 15: Are these commented imports removable at this point?

lib/msal-browser/src/operatingcontext/BaseOperatingContext.ts
```diff
@@ -0,0 +1,63 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { Logger } from "@azure/msal-common";
+import { BrowserConfiguration } from "../config/Configuration";
+
+/**
+ * Base class for operating context
+ * Operating contexts are contexts in which MSAL.js is being run
```
hectormmg, line 12: Nit: Should we give examples here (say: i.e. Embedded app or directly on the browser)?

lib/msal-browser/src/operatingcontext/MetaOSOperatingContext.ts
```diff
@@ -0,0 +1,50 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { BaseOperatingContext } from "./BaseOperatingContext";
+
+export class MetaOSOperatingContext extends BaseOperatingContext {
+    
+    /*
+     * TODO: Once we have determine the bundling code return here to specify the name of the bundle
```
hectormmg: Will this stay as a TODO until we ship MetaOS? Or is the TODO for this PR?

hectormmg:
Re:
> I made the constructor on public client application public again; however we need to remove this when we actually ship MetaOS... this will be a breaking change though and we said no breaking changes for v3

@shoatman @peterzenz based on the previous discussion when initialize became a required call given the changes to make JS/WAM the default option, is this sort of breaking change something we can justify for v3? Particularly considering the instantiation of a PCA is the same block of code that devs will already need to update.

lib/msal-browser/tsconfig.json
```diff
@@ -2,8 +2,8 @@
     "extends": "./tsconfig.base.json",
     "compilerOptions": {
         "moduleResolution": "node",
-        "module": "es6",
-        "target": "es5",
+        "module": "ESNext",
```
konstantin-msft, line 5: We have a separate item to bump module/target for all modules. Would you mind reverting these changes?

lib/msal-browser/package.json
```diff
@@ -22,7 +22,7 @@
     "oauth"
   ],
   "sideEffects": false,
-  "main": "./dist/index.cjs.js",
+  "main": "./lib/index.js",
```
tnorling, line 25: Why the change from `dist` to `lib`?

sameerag: same q. I thought `dist` is preferred as it will be the final export?

lib/msal-browser/src/index.ts
shoatman: yes.  They are.... I'll do that

lib/msal-browser/src/operatingcontext/BaseOperatingContext.ts
shoatman: Good suggestion.  I'll do that.

lib/msal-browser/src/operatingcontext/MetaOSOperatingContext.ts
shoatman: Until we have an implementation

lib/msal-browser/package.json
shoatman: 2 output formats.... common js and es.... put them in 2 separate folders, since that's what rollup appears to require.

lib/msal-browser/tsconfig.json
shoatman: I think i have to leave this for now... or the import expression will not be understood.

lib/msal-browser/src/operatingcontext/MetaOSOperatingContext.ts
```diff
@@ -0,0 +1,50 @@
...
+     * containing the implementation for this operating context
+     */
+    static readonly MODULE_NAME: string = "";
+
+    /**
+     * Unique identifier for the operating context
+     */
+    static readonly ID: string = "MetaOSOperatingContext";
```
tnorling: Should we create an enum for these IDs?

lib/msal-browser/src/operatingcontext/MetaOSOperatingContext.ts
```diff
@@ -0,0 +1,50 @@
...
+     * @returns 
+     */
+    getModuleName(): string {
+        return MetaOSOperatingContext.MODULE_NAME;
+    }
+    /**
+     * Returns the unique identifier for this operating context
+     * @returns string
+     */
+    getId(): string {
```
tnorling: Implementations for getModuleName and getId look the same in both Operating Contexts. Any reason they're not defined in the Base Context?

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -3,281 +3,150 @@
...
+             * If loaded in an environment where window is not available,
+             * set internal flag to false so that further requests fail.
+             * This is to support server-side rendering environments.
+             */
+            const isBrowserEnvironment = typeof window !== "undefined";
+
+            const config = buildConfiguration(configuration, isBrowserEnvironment);
+            const logger = new Logger(config.system.loggerOptions, name, version);
+            const standardOperatingContext = new StandardOperatingContext(logger, config);
+            this.controller = new StandardController(standardOperatingContext, standardOperatingContext.getConfig());
```
tnorling: Can the controller get the config from the operating context itself rather than passing it in?

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -3,281 +3,150 @@
...
+    addEventCallback(callback: Function): string | null {
+        return this.controller.addEventCallback(callback);
+    }
+    removeEventCallback(callbackId: string): void {
+        return this.controller.removeEventCallback(callbackId);
+    }
+    addPerformanceCallback(callback: PerformanceCallbackFunction): string {
+        return this.controller.addPerformanceCallback(callback);
+    }
+    removePerformanceCallback(callbackId: string): boolean {
```
tnorling: nit: new line
```suggestion
    }
    
    removePerformanceCallback(callbackId: string): boolean {
```

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -3,281 +3,150 @@
...
-        }).finally(() => {
-            document.removeEventListener("visibilitychange",this.trackPageVisibility);
-        });
+    acquireTokenSilent(silentRequest: SilentRequest): Promise<AuthenticationResult> {
+        return this.controller.acquireTokenSilent(silentRequest);
+    }
+    acquireTokenByCode(request: AuthorizationCodeRequest): Promise<AuthenticationResult> {
+        return this.controller.acquireTokenByCode(request);
+    }
+    addEventCallback(callback: Function): string | null {
```
tnorling, line 130: Can we keep the block comments above each of these APIs? That's used for documentation

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -3,281 +3,150 @@
...
+    loginRedirect(request?: RedirectRequest | undefined): Promise<void> {
+        return this.controller.loginRedirect(request);
+    }
+    logout(logoutRequest?: EndSessionRequest): Promise<void> {
+        return this.controller.logout(logoutRequest);
+    }
+    logoutRedirect(logoutRequest?: EndSessionRequest): Promise<void> {
+        return this.controller.logoutRedirect(logoutRequest);
+    }
+    logoutPopup(logoutRequest?: EndSessionRequest): Promise<void> {
```
tnorling, line 277: Is there a corresponding 1P PR to ensure relevant changes are made there as well?

lib/msal-browser/src/controllers/ControllerFactory.ts
```diff
@@ -0,0 +1,52 @@
...
+
+            if(metaOS.getAvailable()){
+                /*
+                 * pull down metaos module
+                 * create associated controller
+                 */
+                // return await StandardController.createController(standard);
+                const controller = await import("./StandardController");
+                return await controller.StandardController.createController(standard);
+            }else if(standard.getAvailable()){
```
tnorling: nit can we call this `isAvailable` instead of `getAvailable`?

lib/msal-browser/src/index.ts
tnorling: Let's make sure the 1P library is set up to handle this

lib/msal-browser/src/operatingcontext/BaseOperatingContext.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * Returns the MSAL Logger
+     * @returns Logger
+     */
+    getLogger():Logger {
+        return this.logger;
+    }
+
+    getAvailable():boolean {
```
tnorling: Difference between getAvailable and isAvailable?

lib/msal-browser/src/app/PublicClientApplication.ts
konstantin-msft: +1 I think passing single `standardOperatingContext` param should be enough. `standardOperatingContext.getConfig()` does not do any manipulations with the config. Thus, we don't need an additional param for DI.

lib/msal-browser/src/operatingcontext/BaseOperatingContext.ts
konstantin-msft: Looks like `isAvailable()` acts as both setter and getter, while `getAvailable()` is just a getter function.

lib/msal-browser/src/operatingcontext/MetaOSOperatingContext.ts
shoatman: That makes sense.  I'll take a look

lib/msal-browser/src/app/PublicClientApplication.ts
shoatman: Sure.  Don't recall why i had this way to start....

lib/msal-browser/src/app/PublicClientApplication.ts
shoatman: Will do.

lib/msal-browser/src/app/PublicClientApplication.ts
shoatman: Amarpreet Singh is going to pick up that work.

lib/msal-browser/src/operatingcontext/BaseOperatingContext.ts
shoatman: I'll take a look.

lib/msal-browser/src/controllers/ControllerFactory.ts
shoatman: I think this is async... but i'll doublecheck.

lib/msal-browser/src/operatingcontext/BaseOperatingContext.ts
shoatman: getAvailable is sync and isAvailable is Async.  Do we have an approach that we're using for this pattern?

lib/msal-browser/src/controllers/ControllerFactory.ts
shoatman: isAvailable is async, getAvailable is sync.... see comment below.

lib/msal-browser/src/app/PublicClientApplication.ts
shoatman: applied code formatter

lib/msal-browser/src/operatingcontext/MetaOSOperatingContext.ts
shoatman: Ok.  I looked at this... the implementations could be the same; however, I'd prefer to keep flexibility here.

lib/msal-browser/src/operatingcontext/MetaOSOperatingContext.ts
shoatman: I don't really want to create an enum of these identifiers, what would I use them for?

lib/msal-browser/src/controllers/ControllerFactory.ts
shoatman: Changed this to initialize and isAvailable.

lib/msal-browser/src/operatingcontext/BaseOperatingContext.ts
shoatman: Changed this to initialize and isAvailable

lib/msal-browser/src/app/PublicClientApplication.ts
shoatman: Ok.  This is necessary when initializing the controllers from the controller factory... not needed when creating directly in PCA.

lib/msal-browser/src/operatingcontext/MetaOSOperatingContext.ts
shoatman: Let me know if you feel strongly about this.

lib/msal-browser/src/operatingcontext/MetaOSOperatingContext.ts
shoatman: Let me know if you feel strongly about this

lib/msal-browser/src/app/PublicClientApplication.ts
shoatman: fixed.

lib/msal-browser/src/app/PublicClientApplication.ts
shoatman: Fixed.

lib/msal-browser/src/controllers/ControllerFactory.ts
shoatman: Fixed.  Hopefully more readable now.

lib/msal-browser/src/operatingcontext/BaseOperatingContext.ts
shoatman: fixed.  Hopefully more readable now.

lib/msal-browser/src/index.ts
shoatman: Is there a standard approach to synchronizing changes between the 2 repos?  Or to verify that changes to public will not break 1P?

lib/msal-browser/src/app/PublicClientApplication.ts
shoatman: fixed.

lib/msal-browser/src/app/PublicClientApplication.ts
shoatman: fixed.

lib/msal-browser/src/app/PublicClientApplication.ts
shoatman: fixed.

lib/msal-browser/src/operatingcontext/BaseOperatingContext.ts
shoatman: I'll add this in a future PR>

lib/msal-browser/src/operatingcontext/MetaOSOperatingContext.ts
tnorling: Don't feel strongly, just a suggestion. If it's not useful we can resolve.

lib/msal-browser/src/operatingcontext/MetaOSOperatingContext.ts
tnorling: I'd prefer to move them if they are the same. Flexibility can still be achieved, if needed later, by extending or overriding the function.

shoatman: These are not actually the same though...

lib/msal-browser/src/index.ts
tnorling: Clone the 1P repo, pull in this branch and then attempt to build

lib/msal-browser/src/operatingcontext/MetaOSOperatingContext.ts
tnorling: Ah I see, this can't be moved because the Id and ModuleName are static, is that right? Do they have to be static? If so feel free to resolve.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -3,29 +3,40 @@
...
+export class PublicClientApplication implements IPublicClientApplication {
+
+    private controller: IController;
 
-    // Active requests
-    private activeSilentTokenRequests: Map<string, Promise<AuthenticationResult>>;
-    private astsAsyncMeasurement?: InProgressPerformanceEvent = undefined;
+    public static async createPublicClientApplication(configuration: Configuration): Promise<IPublicClientApplication> {
+
+        const factory = new ControllerFactory(configuration);
```
sameerag, line 33: Can we add some context here on what `StandardController` vs `MetaOSController` would mean? Also will these be added to our type docs?

lib/msal-browser/src/index.ts
shoatman: Should I add an Azure DevOps pipeline for this?  Or trigger a 1P build?

konstantin-msft (approved):

lib/msal-browser/src/app/PublicClientApplication.ts
shoatman: This is internal.  I told Hector I'd add some additional comments in the PRs actually adding support for Teams Apps.

lib/msal-browser/src/index.ts
tnorling: I believe that's something @sameerag is working on this quarter. We've been validating manually up to this point.

lib/msal-browser/src/app/PublicClientApplication.ts
tnorling: Not sure I understand this. Are you saying the config in the OperatingContext may be different from the config passed into the Controller constructor? Is there a reason for that?

lib/msal-browser/src/controllers/ControllerFactory.ts
```diff
@@ -0,0 +1,52 @@
...
+
+            if(metaOS.isAvailable()){
+                /*
+                 * pull down metaos module
+                 * create associated controller
+                 */
+                // return await StandardController.createController(standard);
+                const controller = await import("./StandardController");
+                return await controller.StandardController.createController(standard);
+            }else if(standard.isAvailable()){
```
sameerag, line 47:
```suggestion
            } else if(standard.isAvailable()) {
```

sameerag: Assume this is placeholder until we have metaOS code. Why didn't we leave it undefined rather?

lib/msal-browser/src/app/PublicClientApplication.ts
shoatman: I vaguely remember that there was a problem in the intialization of the logger... I think it was something to do with ensuring that the name and version of the library was actually set.  I want the logger for sure to be initialized correctly before we attempt to construct the objects involved in effectively downloading the implementation of the library.

tnorling: The config only contains the config for the logger not the logger itself. The name/version is passed into the logger constructor.

shoatman: Yes I know..... this code is only here because I was told that we were not going to have any breaking changes in v3... it's going to be removed in a future PR (I think from Konstantin) who's adding an aysnc factory method and making the constructor private.  In the controllerfactory.... current OperatingContexts will need a valid logger reference even if everything else in the config is broken.  I'm actually not sure what happens when you provide an invalid configuration.... but hopefully we log when that happens....

shoatman: At the moment, initialize of the operating contexts are not doing anything.  That will change in the future.

tnorling: Sure I think we're in agreement but I don't feel like my question is being answered...I'm confused why we're passing in `operatingContext` and then also passing in a config which is retrieved by calling `operatingContext.getConfig()`. Why can the controller not call `operatingContext.getConfig()` internally? What am I missing?

shoatman: I"m pretty sure to minimize changes to the existing implementation of Client/PublicClientApplication which moved to StandardController.... I'm not msal.js expert in terms of the implementation there.  If you think it's safe to do so, I can make the larger change.  Was trying to minimize the impact on the existing implementation wherever possible.

tnorling: I see...`PublicClientApplication` is still the public API surface. The controllers are, not only new, but implementation details, so we have the freedom to design those however we want as long as `PublicClientApplication` doesn't change.

lib/msal-browser/src/controllers/StandardController.ts
```diff
@@ -106,7 +116,9 @@ export abstract class ClientApplication {
      *
      * @param configuration Object for the MSAL PublicClientApplication instance
      */
-    constructor(configuration: Configuration) {
+    constructor(operatingContext: StandardOperatingContext, configuration: Configuration) {
```
hectormmg: I think @tnorling is suggesting we remove configuration from this API signature (it's new so not breaking) and call operatingContext.getConfig()... on line 129. Right?

tnorling: Right

lib/msal-browser/src/app/PublicClientApplication.ts
shoatman: Ok.  I'll make this change then.  I think i'm familiar enough with the code to feel comfortable doing that.

lib/msal-browser/src/controllers/StandardController.ts
shoatman: Yep.  I'll do this.

lib/msal-browser/src/app/PublicClientApplication.ts
shoatman: I made some larger changes here since I was going back in anyway... let me know what you think.

lib/msal-browser/src/controllers/StandardController.ts
shoatman: Made bigger initialization change.  Let me know what you think.

tnorling (approved):
LGTM - please review the linter errors before merging. Thanks for addressing feedback! 😄

lib/msal-browser/src/index.ts
sameerag: @shoatman We need to track this. For now, at least, we need to verify compilation before we merge, else 1P will be blocked from release.

hectormmg (approved):
Looks great, thanks Shane!

lib/msal-browser/src/index.ts
sameerag: I misspoke. As this is going in v3, should be good for now.

PR #5003, opened 2022-07-11
Update distributed caching guidance

derisen:
* removes caching from ExpressTestApp, updates readme
* updates caching doc for msal-node, links to RedisTestApp sample
* minor code improvements (refactoring, error handling)

in response to #4632

PR #5559, opened 2023-01-12
Update msal-node caching doc for distributed cache

derisen:
Carrying work from #5003 , as it got hard to review due to PR size. Most comments there are addressed. To sum up:

* Remove custom distributed cache impl. snippet and refer to the actual class
* Remove cache schema as impl. detail.
* Link to the correct sample
* Expand section on obtaining performance metrics

cc @bgavrilMS @Robbie-Microsoft

### DIFF
--- a/lib/msal-node/docs/caching.md
+++ b/lib/msal-node/docs/caching.md
@@ -85,7 +85,7 @@ MSAL Node provides the [DistributedCachePlugin](https://azuread.github.io/micros
 * a **client interface**, which implements `get` and `set` operations on the persistence server (Redis, MySQL etc.).
 * a **partition manager**, for reading from and writing to cache with respect to a given **partition key**.
 
-Please refer to the [RedisTestApp](../../../samples/msal-node-samples/RedisTestApp/README.md) for an implementation.
+Please refer to the [RedisTestApp](../../../samples/msal-node-samples/RedisTestApp/README.md) for a sample implementation.
 
 > :bulb: We recommend you to obtain performance metrics for cache operations, such as **cache retrieval duration** and **cache hit ratios** in your applications. This would give you a reliable picture of how your app scales over time. For instance, to measure how much time MSAL spends in `acquireToken*` calls, you can use Node's [performance measurement API](https://nodejs.org/api/perf_hooks.html)s along with the `fromCache` property in [AuthenticationResult](https://azuread.github.io/microsoft-authentication-library-for-js/ref/modules/_azure_msal_common.html#authenticationresult) class:
 >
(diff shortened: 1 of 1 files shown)


lib/msal-node/docs/caching.md
```diff
@@ -36,30 +36,16 @@ cca.acquireTokenSilent(silentTokenRequest).then((response) => {
...
-    "IdToken": {},
-    "AccessToken": {},
-    "RefreshToken": {},
-    "AppMetadata": {}
-}
-```
-
 ## Cache in memory
-If a user chooses not to persist cache, the `TokenCache` interface is still available to access the tokens cached in memory. The life time of in memory cache is the same as MSAL instance. If the MSAL instance  restarts, the cache is erased when the process lifecycle finishes. We recommend persisting the cache with encryption for all real life applications both for security and desired cache longevity.
+
```
bgavrilMS, line 47: Please add a note that the default in memory cache is not scalable and performance will degrade after holding a few 100s tokens. For web site / web api scenarios this means a few 100 users. For app 2 app scenario, this means a few 100 tenants.

bgavrilMS (approved):

lib/msal-node/docs/caching.md
derisen: done!

lib/msal-node/docs/caching.md
```diff
@@ -92,37 +80,44 @@ On confidential client applications that handle users (web apps that sign in use
...
-        // if in-memory cache has changed
-            // serialize and persist the cache using the same partition key
-    }
-}
-```
-
-It is also a good idea to obtain performance metrics for persistent cache operations, such as **cache hit ratios** from your persistence store. This would give you a reliable picture of how your app scales over time.
+MSAL Node provides the [DistributedCachePlugin](https://azuread.github.io/microsoft-authentication-library-for-js/ref/classes/_azure_msal_node.distributedcacheplugin.html) class for confidential clients, which implements the [ICachePlugin](https://azuread.github.io/microsoft-authentication-library-for-js/ref/interfaces/_azure_msal_common.icacheplugin.html). An instance of `DistributedCachePlugin` requires:
+
+* a client interface, which implements `get` and `set` operations on the persistence server (Redis, MySQL etc.).
```
sameerag:
```suggestion
* a **client interface**, which implements `get` and `set` operations on the persistence server (Redis, MySQL etc.).
```

lib/msal-node/docs/caching.md
```diff
@@ -92,37 +80,44 @@ On confidential client applications that handle users (web apps that sign in use
...
+* a partition manager, for reading from and writing to cache with respect to a given **partition key**.
```
sameerag:
```suggestion
* a **partition manager**, for reading from and writing to cache with respect to a given **partition key**.
```

lib/msal-node/docs/caching.md
```diff
@@ -92,37 +80,44 @@ On confidential client applications that handle users (web apps that sign in use
...
+
+Please refer to the [RedisTestApp](../../../samples/msal-node-samples/RedisTestApp/README.md) for an implementation.
```
sameerag:
```suggestion
Please refer to the [RedisTestApp](../../../samples/msal-node-samples/RedisTestApp/README.md) for a sample implementation.
```

sameerag (approved):
Few minor suggestions, looks good.

PR #5560, opened 2023-01-12
Remove custom ID token validation  in sample

derisen:
We have decided to remove all custom token validation implementations in samples some time ago, unless absolutely necessary (e.g. obo sample). This one apparently slipped by -this PR removes it.

cc @peterzenz @jo-arroyo


bgavrilMS (approved):

jo-arroyo (approved):
LGTM!

samples/msal-node-samples/RedisTestApp/src/AuthProvider.ts
```diff
@@ -205,24 +202,10 @@ export class AuthProvider {
                             const tokenResponse = await msalClient.acquireTokenByCode(req.session.tokenRequest)
                             console.log("\nResponse: \n:", tokenResponse);
 
-                            try {
-                                const isIdTokenValid = await this.tokenValidator.validateIdToken(tokenResponse.idToken);
-
-                                if (isIdTokenValid) {
+                            req.session.account = tokenResponse.account;
+                            req.session.isAuthenticated = true;
```
peterzenz: What we're doing here is removing the validation and implying that the token response alone is enough to ensure correctness.  I don't think this is accurate.  We should keep something in here that says you need to validate the token.

peterzenz (changes requested):

samples/msal-node-samples/RedisTestApp/src/AuthProvider.ts
derisen: If I'm not mistaken, tokens obtained from the `/token` endpoint directly (as opposed to `/authorize` in implicit or hybrid flow) using HTTPS do not really benefit from signature validation (though we don't set up a HTTPS server in samples for simplicity, assuming this will be there when deployed). The [spec](https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation) sort of acknowledges this:

> If the ID Token is received via direct communication between the Client and the Token Endpoint (which it is in this flow), the TLS server validation MAY be used to validate the issuer in place of checking the token signature.

Happy to reconsider though (or I could add a comment to draw attention) @bgavrilMS @jmprieur any thoughts?

sameerag:
@derisen Why have we decided to remove the validation code? To keep the samples lean?

derisen:
> @derisen Why have we decided to remove the validation code? To keep the samples lean?

I think it was because it's a makeshift implementation: doesn't do key rotation and caching, using competitor libraries and etc. but @jo-arroyo can say more

samples/msal-node-samples/RedisTestApp/src/AuthProvider.ts
peterzenz: There are a dozen other things in the spec you reference that need to be validated, no?  I'd feel better about this if we left a breadcrumb for the user to do validation and illustrate that you 401 on a failure.

peterzenz (approved):

derisen:
@peterzenz @sameerag I've restored the token validation code, with appropriate comments for clarifying conditions for requiring signature validation. I had to reflect these changes on the other express web app sample, so apologies for duplication (will work on refactoring these samples to merge them into one and reduce clutter)

sameerag (approved):

peterzenz (approved):

PR #5566, opened 2023-01-12
Added missing fields to InteractionRequiredAuthError

### LINKED ISSUE #5491: InteractionRequiredAuthError doesn't include the claims when implementing conditional access
### Core Library

MSAL Node (@azure/msal-node)

### Core Library Version

1.14.5

### Wrapper Library

N/A

### Wrapper Library Version

0

Confidential

### Description

When trying to Implement [conditional access](https://learn.microsoft.com/en-us/azure/active-directory/conditional-access/overview) (Require MFA) in a `ConfidentialClientApplication` using the `acquireTokenOnBehalfOf` API, we are not receiving the claims in the error object to acquire a new access token with the required claims.

```json
{
  "errorCode": "invalid_grant",
  "errorMessage": "AADSTS50076: Due to a configuration change made by your administrator, or because you moved to a new location, you must use multi-factor authentication to access ''<recourse-id>.\r\nTrace ID: <app-id>\r\nCorrelation ID: <Correlation-ID>\r\nTimestamp: 2022-12-19 22:37:41Z",
  "subError": "basic_action",
  "name": "InteractionRequiredAuthError"
}
```
When looking at the raw response on fiddler, we receive the claims from the token endpoint as shown below:

```json
{
    "error": "invalid_grant",
    "error_description": "AADSTS50076: Due to a configuration change made by your administrator, or because you moved to a new location, you must use multi-factor authentication to access '<recourse-id>'.\r\nTrace ID: <Trace-id>\r\nCorrelation ID: <Correlation-id>\r\nTimestamp: 2022-12-19 22:36:37Z",
    "error_codes": [
        50076
    ],
    "timestamp": "2022-12-19 22:36:37Z",
    "trace_id": "trace_id",
    "correlation_id": "correlation_id",
    "error_uri": "https://login.microsoftonline.com/error?code=50076",
    "suberror": "basic_action",
    "claims": "{\"access_token\":{\"capolids\":{\"essential\":true,\"values\":[\"GUID\"]}}}"
}
```
The issue is when instantiating the [InteractionRequiredAuthError](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-common/src/response/ResponseHandler.ts#L95) class, the claims are not passed to the class. Is there a way to receive the complete error response from msal?

### Error Message

```json
{
  "errorCode": "invalid_grant",
  "errorMessage": "AADSTS50076: Due to a configuration change made by your administrator, or because you moved to a new location, you must use multi-factor authentication to access ''<recourse-id>.\r\nTrace ID: <app-id>\r\nCorrelation ID: <Correlation-ID>\r\nTimestamp: 2022-12-19 22:37:41Z",
  "subError": "basic_action",
  "name": "InteractionRequiredAuthError"
}
```

### Msal Logs

[Tue, 20 Dec 2022 01:28:53 GMT] : @azure/msal-node@1.14.5 : Info - acquireTokenOnBehalfOf called
[Tue, 20 Dec 2022 01:28:53 GMT] : [ed732fe7-13fd-4e60-a6e4-719fd554968f] : @azure/msal-common@9.0.1 : Info - SilentFlowClient:acquireCachedToken - No access token found in cache for the given properties.

### MSAL Configuration

```javascript
const msalConfig = {
    auth: {
        clientId: config.credentials.clientID,
        authority: `https://${config.metadata.authority}/${config.credentials.tenantID}`,
        clientSecret: config.credentials.clientSecret,
        clientCapabilities: ["CP1"],
    },
    system: {
        proxyUrl: "http://localhost:8866",
        loggerOptions: {
                console.log(message);
            },
            piiLoggingEnabled: false,
            logLevel: msal.LogLevel.Info,
        },
    },
};
```

### Relevant Code Snippets

```javascript
const msal = require('@azure/msal-node');
const config = require('./authConfig');

const msalConfig = {
    auth: {
        clientId: config.credentials.clientID,
        authority: `https://${config.metadata.authority}/${config.credentials.tenantID}`,
        clientSecret: config.credentials.clientSecret,
        clientCapabilities: ["CP1"],
    },
    system: {
        proxyUrl: "http://localhost:8866",
        loggerOptions: {
                console.log(message);
            },
            piiLoggingEnabled: false,
            logLevel: msal.LogLevel.Info,
        },
    },
};

// Create msal application object
const cca = new msal.ConfidentialClientApplication(msalConfig);

const getOboToken = async (oboAssertion) => {
    const oboRequest = {
        oboAssertion: oboAssertion,
        scopes: config.resources.downstreamAPI.scopes,
    };

    try {
        const response = await cca.acquireTokenOnBehalfOf(oboRequest);
        return response.accessToken;
    } catch (error) {
        console.log(error)
        throw error;
    }
};
```

### Reproduction Steps

1. Create a ConfidentialClientApplication.
2. Use acquireTokenOnBehalfOf API to get an access token for a recourse.
3. Create a CA policy to enable MFA for the recourse.
4. Check the InteractionRequiredAuthError.

### Expected Behavior

Receiving the claims with the InteractionRequiredAuthError to acquire an access token.

### Identity Provider

Azure AD / MSA

Chrome

### Regression

_No response_

### Source

Internal (Microsoft)

Robbie-Microsoft:
Added missing fields to the InteractionRequiredAuthError as described in the Error response sample [here](https://learn.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-on-behalf-of-flow)

timestamp
trace_id
correlation_id
claims

### DIFF
--- a/lib/msal-common/src/error/InteractionRequiredAuthError.ts
+++ b/lib/msal-common/src/error/InteractionRequiredAuthError.ts
@@ -52,7 +52,13 @@ export class InteractionRequiredAuthError extends AuthError {
     traceId: string;
 
     /**
-     * Claims associated with the error that can be used to get a token
+     * https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-common/docs/claims-challenge.md
+     * 
+     * A string with extra claims needed for the token request to succeed
+     * web site: redirect the user to the authorization page and set the extra claims
+     * web api: include the claims in the WWW-Authenticate header that are sent back to the client so that it knows to request a token with the extra claims
+     * desktop application or browser context: include the claims when acquiring the token interactively
+     * app to app context (client_credentials): include the claims in the AcquireTokenByClientCredential request
      */
     claims: string;
 
--- a/lib/msal-common/src/error/InteractionRequiredAuthError.ts
+++ b/lib/msal-common/src/error/InteractionRequiredAuthError.ts
@@ -41,19 +41,30 @@ export const InteractionRequiredAuthErrorMessage = {
  * Error thrown when user interaction is required.
  */
 export class InteractionRequiredAuthError extends AuthError {
+    /**
+     * The time the error occured at
+     */
     timestamp: string;
+
+    /**
+     * TraceId associated with the error
+     */
     traceId: string;
+
+    /**
+     * Claims associated with the error that can be used to get a token
+     */
     claims: string;
 
     constructor(errorCode?: string, errorMessage?: string, subError?: string, timestamp?: string, traceId?: string, correlationId?: string, claims?: string) {
         super(errorCode, errorMessage, subError);
-        this.name = "InteractionRequiredAuthError";
-        this.timestamp = timestamp || Constants.NOT_APPLICABLE;
-        this.traceId = traceId || Constants.NOT_APPLICABLE;
-        this.correlationId = correlationId || Constants.NOT_APPLICABLE;
-        this.claims = claims || Constants.NOT_APPLICABLE;
-
         Object.setPrototypeOf(this, InteractionRequiredAuthError.prototype);
+        
+        this.timestamp = timestamp || Constants.EMPTY_STRING;
+        this.traceId = traceId || Constants.EMPTY_STRING;
+        this.correlationId = correlationId || Constants.EMPTY_STRING;
+        this.claims = claims || Constants.EMPTY_STRING;
+        this.name = "InteractionRequiredAuthError";
     }
 
     /**
--- a/lib/msal-common/src/error/InteractionRequiredAuthError.ts
+++ b/lib/msal-common/src/error/InteractionRequiredAuthError.ts
@@ -3,6 +3,7 @@
  * Licensed under the MIT License.
  */
 
+import { Constants } from "../utils/Constants";
 import { AuthError } from "./AuthError";
 
 /**
@@ -40,10 +41,17 @@ export const InteractionRequiredAuthErrorMessage = {
  * Error thrown when user interaction is required.
  */
 export class InteractionRequiredAuthError extends AuthError {
+    timestamp: string;
+    traceId: string;
+    claims: string;
 
-    constructor(errorCode?: string, errorMessage?: string, subError?: string) {
+    constructor(errorCode?: string, errorMessage?: string, subError?: string, timestamp?: string, traceId?: string, correlationId?: string, claims?: string) {
         super(errorCode, errorMessage, subError);
         this.name = "InteractionRequiredAuthError";
+        this.timestamp = timestamp || Constants.NOT_APPLICABLE;
+        this.traceId = traceId || Constants.NOT_APPLICABLE;
+        this.correlationId = correlationId || Constants.NOT_APPLICABLE;
+        this.claims = claims || Constants.NOT_APPLICABLE;
 
         Object.setPrototypeOf(this, InteractionRequiredAuthError.prototype);
     }
--- a/lib/msal-common/src/response/ResponseHandler.ts
+++ b/lib/msal-common/src/response/ResponseHandler.ts
@@ -73,7 +73,15 @@ export class ResponseHandler {
         // Check for error
         if (serverResponseHash.error || serverResponseHash.error_description || serverResponseHash.suberror) {
             if (InteractionRequiredAuthError.isInteractionRequiredError(serverResponseHash.error, serverResponseHash.error_description, serverResponseHash.suberror)) {
-                throw new InteractionRequiredAuthError(serverResponseHash.error || Constants.EMPTY_STRING, serverResponseHash.error_description, serverResponseHash.suberror);
+                throw new InteractionRequiredAuthError(
+                    serverResponseHash.error || Constants.EMPTY_STRING,
+                    serverResponseHash.error_description,
+                    serverResponseHash.suberror,
+                    serverResponseHash.timestamp || Constants.NOT_APPLICABLE,
+                    serverResponseHash.trace_id || Constants.NOT_APPLICABLE,
+                    serverResponseHash.correlation_id || Constants.NOT_APPLICABLE,
+                    serverResponseHash.claims || Constants.NOT_APPLICABLE,
+                );
             }
 
             throw new ServerError(serverResponseHash.error || Constants.EMPTY_STRING, serverResponseHash.error_description, serverResponseHash.suberror);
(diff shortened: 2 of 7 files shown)


lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -73,7 +73,15 @@ export class ResponseHandler {
         // Check for error
         if (serverResponseHash.error || serverResponseHash.error_description || serverResponseHash.suberror) {
             if (InteractionRequiredAuthError.isInteractionRequiredError(serverResponseHash.error, serverResponseHash.error_description, serverResponseHash.suberror)) {
-                throw new InteractionRequiredAuthError(serverResponseHash.error || Constants.EMPTY_STRING, serverResponseHash.error_description, serverResponseHash.suberror);
+                throw new InteractionRequiredAuthError(
+                    serverResponseHash.error || Constants.EMPTY_STRING,
+                    serverResponseHash.error_description,
+                    serverResponseHash.suberror,
+                    serverResponseHash.timestamp || Constants.NOT_APPLICABLE,
```
derisen: just curios, why do we have `Constants.EMPTY_STRING` as fallback for `error` but `Constants.NOT_APPLICABLE` for others?

lib/msal-common/src/error/InteractionRequiredAuthError.ts
```diff
@@ -40,10 +41,17 @@ export const InteractionRequiredAuthErrorMessage = {
  * Error thrown when user interaction is required.
  */
 export class InteractionRequiredAuthError extends AuthError {
+    timestamp: string;
```
bgavrilMS, line 47: 1. Please add access modified (public) 
2. Please add comments on all public API.

lib/msal-common/src/error/InteractionRequiredAuthError.ts
```diff
@@ -40,10 +41,17 @@ export const InteractionRequiredAuthErrorMessage = {
...
+    traceId: string;
```
bgavrilMS, line 45: Let's not expose this. It confuses the user as we also have correlation id

lib/msal-common/src/response/ResponseHandler.ts
bgavrilMS: Agreed, empty string or null is the default.

lib/msal-common/test/error/InteractionRequiredAuthError.spec.ts
```diff
@@ -7,7 +7,11 @@ describe("InteractionRequiredAuthError.ts Class Unit Tests", () => {
     it("InteractionRequiredAuthError object can be created", () => {
```
bgavrilMS, line 7: I don't think this test is enough. 

I recommend an E2E test: 

1. copy and modify an existing test which tests for example AcquireTokenForCredential or any other `AcqurieToken`
2. Inject an HTTP response with http status 400 and the body is the JSON with claims
3. Wait for the exception to be thrown and catch it
4. Assert the exception type and the claims

bgavrilMS (changes requested):
Needs a better unit test

lib/msal-common/src/error/InteractionRequiredAuthError.ts
Robbie-Microsoft: They are public by default - I don't need to include a keyword.

bgavrilMS: Please discuss with JS team about code styles used, as I have seen this used elsewhere in our code. Would like to understand the code style / linting, which governs this.

Adding "public" makes the code more readable, particularly if you don't know what the default is. This is part of the .NET code style btw.

derisen: [TypeScript repo's style guide](https://github.com/microsoft/TypeScript/wiki/Coding-guidelines) does not make a mention of this, while [Google's typescript style guide](https://google.github.io/styleguide/jsguide.html) says explicit public should be avoided. Our codebase doesn't seem to be consistent so far, but there's a setting in eslint for enforcing this if going explicit is decided.

lib/msal-common/src/response/ResponseHandler.ts
derisen: is this actually resolved?

bgavrilMS: Re-opening, as it doesn't look resolved.

Robbie-Microsoft: This is resolved, I just hadn't pushed my changes yet.

derisen (approved):
LGTM, thanks!

lib/msal-common/src/error/InteractionRequiredAuthError.ts
```diff
@@ -40,12 +41,30 @@ export const InteractionRequiredAuthErrorMessage = {
...
-    constructor(errorCode?: string, errorMessage?: string, subError?: string) {
-        super(errorCode, errorMessage, subError);
-        this.name = "InteractionRequiredAuthError";
+    /**
+     * TraceId associated with the error
+     */
+    traceId: string;
 
+    /**
+     * Claims associated with the error that can be used to get a token
```
bgavrilMS: Please add a more descriptive exception of what the developer should do with claims, somethinlike 

`A string with extra claims needed for the token request to succeed. If in a web site, redirect the user to the authorization page and set the extra claims. If in a web api, make sure to include the claims in the WWW-Authenticate header that you send back to the client, so that the client knows to request a token with the extra claims. If in a desktop application or in a browser context, include the claims when acquiring the token interactively.  If in an app to app context (client_credentials), include the claims in the AcquireTokenByClientCredential request`

CC @derisen for the comment.

lib/msal-common/test/client/ClientCredentialClient.spec.ts
```diff
@@ -79,6 +79,44 @@ describe("ClientCredentialClient unit tests", () => {
         expect(returnVal.includes(`${AADServerParamKeys.X_MS_LIB_CAPABILITY}=${ThrottlingConstants.X_MS_LIB_CAPABILITY_VALUE}`)).toBe(true);
     });
 
+    it("acquireToken's interactionRequiredAuthError error contains claims", async () => {
+        const errorResponse = {
```
bgavrilMS, line 83: nice

bgavrilMS (approved):
I left one suggestion about the public comment. Signing off. No need for me to re-review this

lib/msal-common/src/error/InteractionRequiredAuthError.ts
Robbie-Microsoft: @bgavrilMS Is there documentation where I can find more info on this? Where did you pull all those examples from?

derisen: That looks pretty explanatory to me! We might also place a link to [our docs](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-common/docs/claims-challenge.md), which has further pointers (and can be updated later easily)

derisen: Taking another look at that doc, it needs to be a little more generalized to cover CA properly (challenge coming from STS vs resource server etc.). I made a note of that and will update, but we can still place the link here I think.

PR #5573, opened 2023-01-13
tokenQueryParameters are now sent to the /token endpoint for all Confidential Client flows

### LINKED ISSUE #5405: Extra Query Parameters are not sent to the /token endpoint
### Core Library

MSAL Node (@azure/msal-node)

### Core Library Version

1.14.3

### Wrapper Library

Not Applicable

### Wrapper Library Version

n/a

Confidential

### Description

1. Use the client_credentials sample
2. Try to set query parameters `foo=bar`. It cannot be done.

Actual: token request is made on <env>/token endpoint
Expected: token request is made on <env>/token?foo=bar endpoint

Impact: unable to target AAD test slices

### Error Message

_No response_

### Msal Logs

_No response_

### MSAL Configuration

```javascript
client_creds with extra query parameter configuration is not possible today, the config does not accept it. This prevents testing on AAD test slices.
```

### Relevant Code Snippets

```javascript
n/a
```

### Reproduction Steps

see description

### Expected Behavior

see description

### Identity Provider

Azure AD / MSA

None (Server)

### Regression

_No response_

### Source

Internal (Microsoft)

Robbie-Microsoft:


lib/msal-common/src/request/CommonClientCredentialRequest.ts
```diff
@@ -19,4 +20,5 @@ export type CommonClientCredentialRequest = BaseAuthRequest & {
     skipCache?: boolean;
     azureRegion?: AzureRegion;
     clientAssertion?: ClientAssertion;
+    tokenQueryParameters?: StringDict;
```
bgavrilMS: I think these should go to all flows. At least all confidential client flows, but I think all public client flows would benefit as well. This allows app developers to select an AAD test slice by adding a special param as query param to the token call (which is a POST request, but it still works). 

@tnorling @peterzenz @sameerag - are you ok with adding this for all flows, i.e in `BaseAuthRequest` ?

lib/msal-common/src/client/ClientCredentialClient.ts
```diff
@@ -141,6 +141,8 @@ export class ClientCredentialClient extends BaseClient {
             };
         } else {
             const requestBody = this.createTokenRequestBody(request);
+            const queryParameters = this.createTokenQueryParameters(request);
```
bgavrilMS: Should be some common logic, this is not specific for client credentials but to all confidential client flows (or all flows, see question above).

lib/msal-common/src/client/ClientCredentialClient.ts
```diff
@@ -141,6 +141,8 @@ export class ClientCredentialClient extends BaseClient {
...
+            const endpoint = StringUtils.isEmpty(queryParameters) ? authority.tokenEndpoint : `${authority.tokenEndpoint}?${queryParameters}`;
```
bgavrilMS: There is a `UrlString` class where the logic of computing the new URL is better suited.

lib/msal-common/test/client/ClientCredentialClient.spec.ts
```diff
@@ -79,6 +83,28 @@ describe("ClientCredentialClient unit tests", () => {
...
+            done();
+        });
+
+        const client = new ClientCredentialClient(config);
+        const clientCredentialRequest: CommonClientCredentialRequest = {
+            authority: TEST_CONFIG.validAuthority,
+            correlationId: TEST_CONFIG.CORRELATION_ID,
+            scopes: TEST_CONFIG.DEFAULT_GRAPH_SCOPE,
+            tokenQueryParameters: {
+                testParam: "testValue"
```
bgavrilMS: Add a few more params please.

bgavrilMS: Also ensure one of the params is empty: 

`testParam2 = ""` in which case you'd want to ignore it, i.e. the url should not contain it.

lib/msal-common/test/client/ClientCredentialClient.spec.ts
```diff
@@ -79,6 +83,28 @@ describe("ClientCredentialClient unit tests", () => {
         expect(returnVal.includes(`${AADServerParamKeys.X_MS_LIB_CAPABILITY}=${ThrottlingConstants.X_MS_LIB_CAPABILITY_VALUE}`)).toBe(true);
     });
 
+    it("Adds tokenQueryParameters to the /token request", (done) => {
+        sinon.stub(ClientCredentialClient.prototype, <any>"executePostToTokenEndpoint").callsFake(async (url: string) => {
+            console.log(url);
```
bgavrilMS: remove console.log

lib/msal-common/test/client/ClientCredentialClient.spec.ts
```diff
@@ -24,9 +24,15 @@ import { CredentialCache } from "../../src/cache/utils/CacheTypes";
 import { CacheManager } from "../../src/cache/CacheManager";
 import { ClientAuthError } from "../../src/error/ClientAuthError";
 import { AuthenticationResult } from "../../src/response/AuthenticationResult";
-import { AppTokenProviderResult, AuthToken, IAppTokenProvider } from "../../src";
+import { AppTokenProviderResult, AuthToken, ClientConfiguration, IAppTokenProvider } from "../../src";
 
 describe("ClientCredentialClient unit tests", () => {
+    let config: ClientConfiguration;
+
+    beforeEach(async () => {
```
bgavrilMS, line 32: nice little refactoring

bgavrilMS (changes requested):
Logic needs to be set on all flows (At least all CCA flows)

lib/msal-common/src/client/ClientCredentialClient.ts
derisen: [+1](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/9384b584a3383732d9070bd7ed345589f8969c3f/lib/msal-common/src/url/UrlString.ts#L88)

change/@azure-msal-common-47c38479-2f76-4779-b095-fd50e112915f.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "ExtraTokenParameters are now sent to the /token endpoint when using acquireTokenByClientCredential #5573",
```
derisen, line 3:
```suggestion
  "comment": "tokenQueryParameters are now sent to the /token endpoint when using acquireTokenByClientCredential #5573",
```

derisen:
```suggestion
  "comment": "tokenQueryParameters are now sent to the /token endpoint when using acquireTokenByClientCredential #5573",
```

change/@azure-msal-node-a37dcd93-77ce-4c8e-bdfb-6437145819be.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "ExtraTokenParameters are now sent to the /token endpoint when using acquireTokenByClientCredential #5573",
```
derisen:
```suggestion
  "comment": "tokenQueryParameters are now sent to the /token endpoint when using acquireTokenByClientCredential #5573",
```

change/@azure-msal-common-47c38479-2f76-4779-b095-fd50e112915f.json
derisen: qol nit in case someone searches the release notes for the param

change/@azure-msal-node-a37dcd93-77ce-4c8e-bdfb-6437145819be.json
derisen: also, seems to me the only change to msal-node is the added comment -would this actually constitute a minor?

lib/msal-common/src/request/CommonClientCredentialRequest.ts
tnorling: This is already implemented for PublicClient flows but I have no objections to adding it where it's missing or moving it to be more centrally available. For reference [this is where it currently lives](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/9384b584a3383732d9070bd7ed345589f8969c3f/lib/msal-common/src/request/CommonAuthorizationUrlRequest.ts#L49) - all of our public client APIs extend this request type

bgavrilMS: @tnorling - no objection here, but please help me understand the inheritance diagram. I see that [DeviceCode](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/9384b584a3383732d9070bd7ed345589f8969c3f/lib/msal-common/src/request/CommonDeviceCodeRequest.ts), [ROPC](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/9384b584a3383732d9070bd7ed345589f8969c3f/lib/msal-common/src/request/CommonUsernamePasswordRequest.ts) and [silent](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/9384b584a3383732d9070bd7ed345589f8969c3f/lib/msal-common/src/request/CommonSilentFlowRequest.ts) all extend `BaseAuthRequest` and not `CommonAuthorizationUrlRequest`

tnorling: You're right, I should have clarified that all of our flows that use _at least_ the `/authorize` endpoint (which is indeed most, but not all PCA APIs) use this type. This will indeed need to be extended at least to Silent. DeviceCode may need its own parameter given it doesn't use `/token`. I'm less certain about the value of adding this for ROPC, but no objections if you'd like to do it anyway. 

One thing I'm remembering now as we're talking about this is that we didn't add `tokenQueryParameters` (or `extraQueryParameters`) to `BaseAuthRequest` because not all request types ultimately make a `/token` request. To keep the request type surface clean I'd ask that you use one of two approaches:

1. Move the params to `BaseAuthRequest` but then explicitly omit them from downstream request types where they are not relevant
or
2. Only include the params on the types which will actually use them

We initially went with option 2 but it appears we either didn't do it completely or some of these types didn't exist yet at that time. If you want to switch to approach 1 I'm ok with that.

lib/msal-common/src/request/CommonAuthorizationCodeRequest.ts
```diff
@@ -17,15 +17,15 @@ import { CcsCredential } from "../account/CcsCredential";
...
+ * - resourceRequestMethod   - HTTP Request type used to request data from the resource (i.e. "GET", "POST", etc.).  Used for proof-of-possession flows.
+ * - resourceRequestUri      - URI that token will be used for. Used for proof-of-possession flows.
+ * - enableSpaAuthCode       - Enables the acqusition of a spa authorization code (confidential clients only)
+ * - tokenQueryParameters    - String to string map of custom query parameters added to the /token call
  */
 export type CommonAuthorizationCodeRequest = BaseAuthRequest & {
     code: string;
     redirectUri: string;
     codeVerifier?: string;
-    tokenQueryParameters?: StringDict;
```
bgavrilMS, line 28: You can't remove a type from a public ctor. This is a breaking change.

lib/msal-common/src/client/BaseClient.ts
```diff
@@ -136,4 +138,22 @@ export abstract class BaseClient {
         }
         this.authority = updatedAuthority;
     }
+
+    /**
+     * Creates query string for the /token request
+     * @param request
+     */
+    createTokenQueryParameters(request:
```
bgavrilMS: Can't this accept a param `request` of type `BaseAuthRequest` ?

This design is a code smell - a base class should not know about derived classes. Otherwise, the design becomes `brittle`, i.e. when a new subclass is defined, you must also update the base class. 

See the O and L principles from SOLID - https://www.baeldung.com/solid-principles

bgavrilMS (changes requested):
No breaking changes in public APIs

lib/msal-common/src/request/CommonAuthorizationCodeRequest.ts
tnorling: In this case it's fine because it was moved to `BaseAuthRequest` which this inherits

lib/msal-common/src/client/BaseClient.ts
Robbie-Microsoft: good point, thanks

lib/msal-browser/src/request/AuthorizationCodeRequest.ts
```diff
@@ -5,7 +5,7 @@
 
 import { CommonAuthorizationCodeRequest } from "@azure/msal-common";
 
-export type AuthorizationCodeRequest = Partial<Omit<CommonAuthorizationCodeRequest, "code"|"enableSpaAuthorizationCode"|"requestedClaimsHash">> & {
+export type AuthorizationCodeRequest = Partial<Omit<CommonAuthorizationCodeRequest, "code"|"enableSpaAuthorizationCode"|"requestedClaimsHash"|"tokenQueryParameters">> & {
```
tnorling: Why is this omitted here? It should not be.

lib/msal-browser/src/request/AuthorizationUrlRequest.ts
```diff
@@ -8,7 +8,7 @@ import { CommonAuthorizationUrlRequest } from "@azure/msal-common";
 /**
  * This type is deprecated and will be removed on the next major version update
  */
-export type AuthorizationUrlRequest = Omit<CommonAuthorizationUrlRequest, "state"|"nonce"|"requestedClaimsHash"|"nativeBroker"> & {
+export type AuthorizationUrlRequest = Omit<CommonAuthorizationUrlRequest, "state"|"nonce"|"requestedClaimsHash"|"nativeBroker"|"tokenQueryParameters"> & {
```
tnorling: Same question here. This should not be omitted.

lib/msal-common/src/request/CommonDeviceCodeRequest.ts
```diff
@@ -16,8 +16,9 @@ import { BaseAuthRequest } from "./BaseAuthRequest";
  * - resourceRequestMethod      - HTTP Request type used to request data from the resource (i.e. "GET", "POST", etc.).  Used for proof-of-possession flows.
  * - resourceRequestUri         - URI that token will be used for. Used for proof-of-possession flows.
  * - timeout                    - Timeout period in seconds which the user explicitly configures for the polling of the device code endpoint. At the end of this period; assuming the device code has not expired yet; the device code polling is stopped and the request cancelled. The device code expiration window will always take precedence over this set period.
+ * - tokenQueryParameters       - String to string map of custom query parameters added to the /token call
```
tnorling: Since this does not use the /token endpoint, consider omitting this here and exposing a more appropriately named parameter instead. Perhaps `extraQueryParameters` would work?

lib/msal-common/src/request/RequestValidator.ts
```diff
@@ -7,6 +7,7 @@ import { StringUtils } from "../utils/StringUtils";
 import { ClientConfigurationError } from "../error/ClientConfigurationError";
 import { PromptValue, CodeChallengeMethodValues} from "../utils/Constants";
 import { StringDict } from "../utils/MsalTypes";
+import { omitBy, isEmpty } from "lodash";
```
tnorling: Prefer not to introduce a new dependency if it can be avoided. I think what you're using this for can be accomplished without this lib

tnorling (changes requested):

lib/msal-common/src/request/RequestValidator.ts
Robbie-Microsoft: Ah, I see now that the existing lodash package is located in devDepencies instead of dependencies.

lib/msal-browser/src/request/AuthorizationCodeRequest.ts
Robbie-Microsoft: I think I got mixed up in determining which public flows allow tokenQueryParameters. Is there a set list I can look at? Or can you tell me here?

lib/msal-browser/src/request/EndSessionPopupRequest.ts
```diff
@@ -17,7 +17,7 @@ import { PopupWindowAttributes } from "./PopupWindowAttributes";
  * - popupWindowAttributes  - Optional popup window attributes. popupSize with height and width, and popupPosition with top and left can be set.
  * - logoutHint             - A string that specifies the account that is being logged out in order to skip the server account picker on logout
  */
-export type EndSessionPopupRequest = Partial<CommonEndSessionRequest> & {
+export type EndSessionPopupRequest = Partial<Omit<CommonEndSessionRequest, "tokenQueryParameters">> & {
```
bgavrilMS, line 20: @Robbie-Microsoft - the 2 session requests should have this omitted (you have already done it)

bgavrilMS (approved):

lib/msal-common/src/request/CommonDeviceCodeRequest.ts
Robbie-Microsoft: Unless we keep it named tokenQueryParameters, I think this would cause too many complications. We'd have to add extraQueryParameters to BaseAuthRequest just so that it can be used in BaseClient's createTokenQueryParameters function. Then we'd have to omit extraQueryParameters in all the other requests. We'd also have to add support/documentation that the tokenQueryParameters should be passed in extraQueryParameters in the app's config file. Leaving it named tokenQueryParameters would eliminate these problems, I think. Thoughts?

tnorling: > We'd also have to add support/documentation that the tokenQueryParameters should be passed in extraQueryParameters in the app's config file.

I'm not sure I understand this concern. Can you please elaborate.

In any event the query string is constructed [in DeviceCodeClient](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/58e57222cfde9ccd20dd839785f09bb4a07c3a06/lib/msal-common/src/client/DeviceCodeClient.ts#L127) I don't think you need to do anything to `BaseAuthRequest`

Robbie-Microsoft: I understand now, thanks

change/@azure-msal-browser-09b253d8-3271-4ccb-aa4c-834e213cbfe2.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "tokenQueryParameters are now sent to the /token endpoint for all Confidential Client flows. They're also explicitly excluded from Public Client flows that don't use them. #5573",
```
tnorling: Please update this comment to be more descriptive of what changed in msal-browser (CCA isn't relevant to browser so should not be mentioned)

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -183,11 +183,12 @@ export class RefreshTokenClient extends BaseClient {
     private async executeTokenRequest(request: CommonRefreshTokenRequest, authority: Authority)
         : Promise<NetworkResponse<ServerAuthorizationTokenResponse>> {
         this.performanceClient?.addQueueMeasurement(PerformanceEvents.RefreshTokenClientExecuteTokenRequest, request.correlationId);
-        const acquireTokenMeasurement = this.performanceClient?.startMeasurement(PerformanceEvents.RefreshTokenClientExecuteTokenRequest, request.correlationId);
```
tnorling, line 186: Why was this moved?

tnorling (approved):
Approved with comments

lib/msal-common/src/client/RefreshTokenClient.ts
Robbie-Microsoft: Not sure! I'll move it back.

PR #5575, opened 2023-01-14
Allow setting a preferred port in acquireTokenInteractive

### LINKED ISSUE #5442: Problem with ElectronSystemBrowserTestApp deep linking with acquireTokenInteractive()
### Core Library

MSAL Node (@azure/msal-node)

### Core Library Version

1.14.2

### Wrapper Library

Not Applicable

### Wrapper Library Version

none

Public

### Description

I have the [ElectronSystemBrowserTestApp ](https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev/samples/msal-node-samples/ElectronSystemBrowserTestApp) sample app setup and running, and can sign and the deeplink calls back to the app. The problem is how to improve the experience with the browser popup to allow the redirect. Since acquireTokenInteractive chooses a random port on localhost each time, the user will always have the browser popup, even if they check the box to allow to open link, because the next time it will be a different port on localhost.
image

### Error Message

_No response_

### Msal Logs

_No response_

### MSAL Configuration

```javascript
{
    "authOptions":
        {
            "clientId": "Enter_the_Application_Id_Here",
            "authority": "https://login.microsoftonline.com/organizations"
        },
    "resourceApi":
    {
        "endpoint": "Enter_the_Graph_Endpoint_Here",
        "scopes": ["User.Read", "Mail.Read"]
    },
    "customProtocol":
    {
        "name": "msal{Enter_the_Application_Id_Here}"
    },
    "cache": {
        "cacheLocation": "./data/cache.json"
    }
}
```

### Relevant Code Snippets

```javascript
ElectronSystemBrowserTestApp: https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev/samples/msal-node-samples/ElectronSystemBrowserTestApp
```

### Reproduction Steps

1. Download the [ElectronSystemBrowserTestApp ](https://github.dev/AzureAD/microsoft-authentication-library-for-js/blob/dev/samples/msal-node-samples/ElectronSystemBrowserTestApp/src/AuthProvider.ts) and configure the application in customConfig.js and enable deeplinking
2. Run the app and sign in > browser will open and after sign in is complete will try to deeplink back to the app, but browser will popup to allow to open this applicaiton.
3. The problem is the browser will popup every time, even if the users checks always allow, because acquireTokenInteractive uses a different port every time.

### Expected Behavior

Expected behavior would be some way to specify a port number or a way to have acquireTokenInteractive use a consistent URL so that the user would not always get the popup consent in browser for the deep link back to the electron app****

### Identity Provider

Azure AD / MSA

### Regression

_No response_

### Source

External (Customer)

derisen:
This PR updates the default loopback client to attempt listening at a preferred port set by the consumer of `acquireTokenInteractive` API, falling back to any available port if preferred port is not available. See also #5578


derisen:
Decided to drop this PR, which seems redundant given #5578 and also introduces a somewhat complex logic into the API, which is probably not a good idea.

PR #5580, opened 2023-01-17
Add broker timeouts

konstantin-msft:


change/@azure-msal-common-558c780e-3387-4832-8c07-b5449309b5dc.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Add broker timeouts #5580",
+  "packageName": "@azure/msal-common",
+  "email": "sh.konst89@gmail.com",
```
tnorling: Please update to your Microsoft email

lib/msal-common/src/telemetry/performance/PerformanceEvent.ts
```diff
@@ -221,6 +221,12 @@ export type StaticFields = {
     matsHttpEventCount?: number;
     httpVerToken?: string;
     httpVerAuthority?: string;
+
+    // Broker timeouts
```
tnorling: Can you add a bit more context here? i.e. "This is used by the 1P browser lib". I do wonder though if there's a way we can avoid exposing these types of details to the 3P/common libs at all?

change/@azure-msal-common-558c780e-3387-4832-8c07-b5449309b5dc.json
konstantin-msft: Good catch, thank you! Fixed

lib/msal-common/src/telemetry/performance/PerformanceEvent.ts
konstantin-msft: Not at the moment unfortunately, but we should be able in the future by extending 3p telemetry classes in 1p. Added context.

lib/msal-common/src/telemetry/performance/PerformanceEvent.ts
```diff
@@ -221,6 +221,12 @@ export type StaticFields = {
     matsHttpEventCount?: number;
     httpVerToken?: string;
     httpVerAuthority?: string;
+
+    // Broker timeouts used by 1p browser lib
+    brokerInteractionTimeout?: number;
```
peterzenz: I'm assuming these are all in ms?  Suggest renaming them to brokerInteractionTimeoutMs, etc.

peterzenz (approved):

lib/msal-common/src/telemetry/performance/PerformanceEvent.ts
konstantin-msft: Good point, thank you. Done.

jo-arroyo (approved):

tnorling (approved):

PR #5593, opened 2023-01-19
install local build in angular b2c e2e

derisen:


samples/msal-angular-v2-samples/angular-b2c-sample-app/package.json
```diff
@@ -23,8 +23,8 @@
     "@angular/platform-browser": "^14.0.0-next.0",
     "@angular/platform-browser-dynamic": "^14.0.0-next.0",
     "@angular/router": "^14.0.0-next.0",
-    "@azure/msal-angular": "^2.4.1",
-    "@azure/msal-browser": "^2.28.1",
+    "@azure/msal-angular": "file:../../../lib/msal-angular/dist",
```
jo-arroyo: Just the `install:local` needs to be changed right? The workflow will run `install:local` when running the tests.

derisen: that's right, only change needed is install:local script (reverted the change in dependencies)

jo-arroyo (approved):
LGTM! Thanks for getting this out.

peterzenz (approved):

PR #5609, opened 2023-01-24
MissingTenantId error is now thrown in acquireTokenByClientCredential if tenantId is missing

Robbie-Microsoft:


bgavrilMS (changes requested):
This is a pretty big breaking change, so we won't be doing this now. The associated issue was for tracking purposes, for when a new major change will be added. 

From my point of view, we should avoid breaking changes. But if the new release schedule of the JS team will mean creating a new major version, we should take this fix.

PR #5614, opened 2023-01-27
Track manually completed sub-measurements

konstantin-msft:
- Track manually completed sub-measurements.
- Add unit tests.

### DIFF
--- a/lib/msal-common/src/telemetry/performance/PerformanceClient.ts
+++ b/lib/msal-common/src/telemetry/performance/PerformanceClient.ts
@@ -451,6 +451,7 @@ export abstract class PerformanceClient implements IPerformanceClient {
 
                     const completedEvent = this.endMeasurement(event);
                     if (completedEvent) {
+                        completedEvent.isIncomplete = true;
                         completedEvents.push(completedEvent);
                     }
                 }
@@ -487,6 +488,9 @@ export abstract class PerformanceClient implements IPerformanceClient {
                          */
                         if (!previous[subMeasurementName]) {
                             previous[subMeasurementName] = current.durationMs;
+                            if (current.isIncomplete) {
+                                previous[`${current.name}Incomplete`] = true;
+                            }
                         } else {
                             this.logger.verbose(`PerformanceClient: Submeasurement for ${measureName} already exists for ${current.name}, ignoring`, correlationId);
                         }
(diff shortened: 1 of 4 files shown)


lib/msal-common/src/telemetry/performance/PerformanceClient.ts
```diff
@@ -318,6 +318,7 @@ export abstract class PerformanceClient implements IPerformanceClient {
 
                     const completedEvent = this.endMeasurement(event);
                     if (completedEvent) {
+                        completedEvent.isIncomplete = true;
```
peterzenz: I'd expect that an incomplete measurement represents either a bug, a timeout or crash somewhere.  Please add a datapoint to log whether we had incomplete measurements on a given call.  Just one datapoint for now, if we find there are tons of them we'll look at something more comprehensive if needed so we can debug/understand.

peterzenz (commented):
I've asked for an additional datapoint. You can complete this without it, but please follow up with another change if you choose to go that route.

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
konstantin-msft: Would a single data point with an array of manually completed events work? This should be both inexpensive to store and easy to debug.

peterzenz: Very easy to debug, expensive to store in telemetry.  For telemetry a boolean just to see how prevalent it is, would be great. 
For tracing/logging, an array of manually completed events would be fine and that would help with one-off debugging in the near term.

konstantin-msft: Replaced the previous approach with a counter for incomplete sub-measurements.

lib/msal-common/test/telemetry/PerformanceClient.spec.ts
```diff
@@ -3,9 +3,16 @@
  * Licensed under the MIT License.
  */
 
-import { ApplicationTelemetry, IGuidGenerator, IPerformanceMeasurement, Logger, PerformanceEvents } from "../../src";
-import { IPerformanceClient } from "../../src/telemetry/performance/IPerformanceClient";
-import { PerformanceClient } from "../../src/telemetry/performance/PerformanceClient";
+import {
```
tnorling, line 6: Import from the relevant file(s), not src

tnorling (approved):

lib/msal-common/test/telemetry/PerformanceClient.spec.ts
konstantin-msft: Good catch, thank you Thomas. Fixed.

peterzenz (approved):

PR #5615, opened 2023-01-28
Update caching.md

saudmish:
Adding additional details on the need to use sessionStorage to ensure compatibility with 'Browser CA' in AAD CA.

### DIFF
--- a/lib/msal-browser/docs/caching.md
+++ b/lib/msal-browser/docs/caching.md
@@ -23,13 +23,15 @@ const pca = new PublicClientApplication({
 
 By default, MSAL stores the various authentication artifacts it obtains from the IdP in browser storage using the [Web Storage API](https://developer.mozilla.org/docs/Web/API/Web_Storage_API) supported by all modern browsers. Accordingly, MSAL offers two methods of persistent storage: `sessionStorage` (default) and `localStorage`. In addition, MSAL provides `memoryStorage` option which allows you to opt-out of storing the cache in browser storage.
 
-| Cache Location   | Cleared on              | Shared between windows/tabs | Redirect flow supported |
-|------------------|-------------------------|-----------------------------|-------------------------|
-| `sessionStorage` | window/tab close        | No                          | Yes                     |
-| `localStorage`   | user logout             | Yes                         | Yes                     |
-| `memoryStorage`  | page refresh/navigation | No                          | No                      |
+> :bulb: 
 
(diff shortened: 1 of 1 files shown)


lib/msal-browser/docs/caching.md
```diff
@@ -23,13 +23,15 @@ const pca = new PublicClientApplication({
...
+| `memoryStorage`  | page refresh/navigation | No                          | No                      | No                                              |
+
+> :bulb: While the authentication state may be lost in session and memory storage due to window/tab close or page refresh/navigation, respectively, users will still have an active session with the IdP as long as the session cookie is not expired and might be able to re-authenticate without any prompts. Additionally - Azure Active Directory Conditional Access (AAD CA) supports the ability for AAD Premium customers to define Browser Session controls at the authentication layer based on that session cookie. To ensure your app supports this feature, ensure you use `sessionStorage` so that AAD can evaluate if re-prompting the user is required based on IT security policies defined AAD CA policy. For more information, see [Persistent Browser Session in AAD CA controls documention](https://learn.microsoft.com/azure/active-directory/conditional-access/howto-conditional-access-session-lifetime#persistence-of-browsing-sessions).
```
derisen, line 34: @saudmish can you tell us more how this feature relates to MSAL browser storage? If I'm understanding correctly, this mainly affects the AAD session cookie. Is it because when using local storage, MSAL cache will persist after closing the browser, so the application auth state will be preserved when a new browser session starts and the app won't necessarily attempt to reauthenticate the user? Not sure about this, since application auth state is mainly a UI/UX thing, which doesn't necessarily represent the auth state with Azure AD -@sameerag @tnorling any thoughts?

tnorling: Agreed. I don't see these things as necessarily related. CA policies will be evaluated each time a request is made against AAD regardless of which storage option is used for the tokens

derisen:
@saudmish unfortunately we won't be able to accept this, but I've made a note update either this doc or the FAQ to clarify the relationship between MSAL cache and Azure AD session. Thanks again for bringing this up. Closing.

lib/msal-browser/docs/caching.md
saudmish: @derisen thanks for reviewing. Yup we have an internal thread on the topic that @tnorling and @sameerag is on. Looking forward to other updates!

PR #5627, opened 2023-01-31
Truncate integral telemetry fields

konstantin-msft:
- Truncate integral telemetry fields.
- Remove broker timeouts (declared in 1p).

### DIFF
--- a/lib/msal-common/src/telemetry/performance/PerformanceClient.ts
+++ b/lib/msal-common/src/telemetry/performance/PerformanceClient.ts
@@ -7,7 +7,14 @@ import { ApplicationTelemetry } from "../../config/ClientConfiguration";
 import { Logger } from "../../logger/Logger";
 import { InProgressPerformanceEvent, IPerformanceClient, PerformanceCallbackFunction, QueueMeasurement } from "./IPerformanceClient";
 import { IPerformanceMeasurement } from "./IPerformanceMeasurement";
-import { Counters, PerformanceEvent, PerformanceEvents, PerformanceEventStatus, StaticFields } from "./PerformanceEvent";
+import {
+    Counters,
+    PerformanceEvent,
+    IntFields,
+    PerformanceEvents,
+    PerformanceEventStatus,
+    StaticFields
+} from "./PerformanceEvent";
 
 export abstract class PerformanceClient implements IPerformanceClient {
     protected authority: string;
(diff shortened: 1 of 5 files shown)


lib/msal-common/src/telemetry/performance/PerformanceClient.ts
```diff
@@ -132,6 +139,13 @@ export abstract class PerformanceClient implements IPerformanceClient {
      */
     abstract setPreQueueTime(eventName: PerformanceEvents, correlationId?: string): void;
 
+    /**
+     * Get integral fields.
+     */
+    getIntFields(): ReadonlySet<string> {
```
tnorling, line 146: Do we need this function or can we just directly use `IntFields`?

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
```diff
@@ -506,6 +520,7 @@ export abstract class PerformanceClient implements IPerformanceClient {
                     queuedCount: totalQueueCount,
                     incompleteSubsCount
                 };
+                this.truncateIntegralFields(finalEvent, this.getIntFields());
```
tnorling, line 524: Then you don't need to pass anything in here...

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
konstantin-msft: I have added it on purpose to allow overriding integral fields in 1p repo.

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
konstantin-msft: Please see above comment.

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
peterzenz: Maybe add a comment to the code itself so it's clear to the reader.

peterzenz (approved):

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
konstantin-msft: Done

sameerag (approved):
Looks good.

PR #5649, opened 2023-02-06
Increase the number of retries for e2e tests

konstantin-msft:
- Increase the number of retries for e2e tests to mitigate VM resource contention.

### DIFF
--- a/samples/e2eTestUtils/TestUtils.ts
+++ b/samples/e2eTestUtils/TestUtils.ts
@@ -4,6 +4,7 @@ import { LabConfig } from "./LabConfig";
 import { LabClient } from "./LabClient";
 
 export const ONE_SECOND_IN_MS = 1000;
+export const RETRY_TIMES = 5;
 
 export class Screenshot {
     private folderName: string;
--- a/samples/msal-angular-v2-samples/angular-b2c-sample-app/test/local-account.spec.ts
+++ b/samples/msal-angular-v2-samples/angular-b2c-sample-app/test/local-account.spec.ts
@@ -1,5 +1,5 @@
 import * as puppeteer from "puppeteer";
-import { Screenshot, setupCredentials, b2cLocalAccountEnterCredentials } from "../../../e2eTestUtils/TestUtils";
+import {Screenshot, setupCredentials, b2cLocalAccountEnterCredentials, RETRY_TIMES} from "../../../e2eTestUtils/TestUtils";
 import { LabClient } from "../../../e2eTestUtils/LabClient";
 import { LabApiQueryParams } from "../../../e2eTestUtils/LabApiQueryParams";
 import { UserTypes, B2cProviders } from "../../../e2eTestUtils/Constants";
@@ -8,7 +8,7 @@ import { BrowserCacheUtils } from "../../../e2eTestUtils/BrowserCacheTestUtils";
 const SCREENSHOT_BASE_FOLDER_NAME = `${__dirname}/screenshots/local-account-tests`;
 
 describe('B2C user-flow tests (local account)', () => {
-    jest.retryTimes(1);
+    jest.retryTimes(RETRY_TIMES);
     let browser: puppeteer.Browser;
     let context: puppeteer.BrowserContext;
     let page: puppeteer.Page;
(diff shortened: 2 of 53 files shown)


hectormmg:
Lgtm, just curious, how easy is it to see how many retries actually had to be done on a test run?

hectormmg (approved):

hectormmg:
Also, this PR is a great reminder of why we should try to leverage constants like RETRY_TIMES everywhere. If we'd had a constant there from the beginning this PR could've had 50 less files.

derisen (approved):
LGTM -thanks for doing this!

sameerag (approved):
Does this increase our CI/CD time? Approving, but would like to see if the effect if this helped with the success rate vs time.

tnorling:
> Does this increase our CI/CD time? Approving, but would like to see if the effect if this helped with the success rate vs time.

Reliability > Perf in a vacuum but I'm actually more concerned about the impact to queue time for other PRs when one PR is hogging the VMs our CI runs on. I've asked Konstantin to look into ways to minimize that impact (such as skipping E2E until the PR has been approved, making E2E a last step before merging)

hectormmg:
> > Does this increase our CI/CD time? Approving, but would like to see if the effect if this helped with the success rate vs time.
> 
> Reliability > Perf in a vacuum but I'm actually more concerned about the impact to queue time for other PRs when one PR is hogging the VMs our CI runs on. I've asked Konstantin to look into ways to minimize that impact (such as skipping E2E until the PR has been approved, making E2E a last step before merging)

We can certainly take this offline, I will say one downside is that we'd need to be running the entire e2e test suite locally until the PR is approved for it to start catching any bugs the PR introduces.

tnorling:
> We can certainly take this offline, I will say one downside is that we'd need to be running the entire e2e test suite locally until the PR is approved for it to start catching any bugs the PR introduces.

You can run workflows manually in Github as well if you're unsure and just want to do a quick check. I'm just worried that PRs that are still in active development will kick off a bunch of E2E tests on every commit which will fail 5 times and hold up the line. That said, if we can't find a middle ground right now I'm open to trying it and revisiting if/when it does become a problem.

hectormmg:
> > We can certainly take this offline, I will say one downside is that we'd need to be running the entire e2e test suite locally until the PR is approved for it to start catching any bugs the PR introduces.
> 
> You can run workflows manually in Github as well if you're unsure and just want to do a quick check. I'm just worried that PRs that are still in active development will kick off a bunch of E2E tests on every commit which will fail 5 times and hold up the line. That said, if we can't find a middle ground right now I'm open to trying it and revisiting if/when it does become a problem.

No, I share that concern, I know it's the main reason we haven't done the retries yet. Maybe leaving E2E as a final step is the best compromise that still gives the team a better experience.

konstantin-msft:
> Lgtm, just curious, how easy is it to see how many retries actually had to be done on a test run?

Analyzing logs for repetitive test names seems like the only option.

konstantin-msft:
> > We can certainly take this offline, I will say one downside is that we'd need to be running the entire e2e test suite locally until the PR is approved for it to start catching any bugs the PR introduces.
> 
> You can run workflows manually in Github as well if you're unsure and just want to do a quick check. I'm just worried that PRs that are still in active development will kick off a bunch of E2E tests on every commit which will fail 5 times and hold up the line. That said, if we can't find a middle ground right now I'm open to trying it and revisiting if/when it does become a problem.

I would avoid running e2e jobs manually cause people can just forget about it. I would rather prefer kicking off e2e jobs when PR is ready for review. Please take a look at a follow-up PR https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/5662.

tnorling:
> I would avoid running e2e jobs manually cause people can just forget about it. I would rather prefer kicking off e2e jobs when PR is ready for review. Please take a look at a follow-up PR https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/5662.

I agree, my suggestion to run the tests manually was more around if you want to verify the tests will pass _before_ you've gotten to the stage where tests will be auto-run by CI. It is an option in addition to, not instead of, auto-run CI tests which are required to merge the PR.

konstantin-msft:
> Does this increase our CI/CD time? Approving, but would like to see if the effect if this helped with the success rate vs time.

It increases CI/CD time in case of a breaking change and/or VM resource contention. I am not sure I am aware of the way to capture CI/CD metrics you are referring to.

hectormmg:
@konstantin-msft is there any way to limit the retries to tests that timeout? Other than the fact that there's less chance of ending up with long queuing times, I think it may be dangerous to have retries on all the time since it would prevent us from catching intermittent bugs.

peterzenz (approved):

konstantin-msft:
> @konstantin-msft is there any way to limit the retries to tests that timeout? Other than the fact that there's less chance of ending up with long queuing times, I think it may be dangerous to have retries on all the time since it would prevent us from catching intermittent bugs.

You can do that locally by changing https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/5649/files#diff-31e14e167d5f2968c7f301ff3197d4ff257f3f9bce6eeffd0fbc472784b861d0R11 param.

Issue #4879, opened 2022-06-09
Getting ClientAuthError: endpoints_resolution_error: Error: could not resolve endpoints with @azure/msal-node 1.9.1


awanlin:
### Core Library

MSAL Node (@azure/msal-node)

### Core Library Version

1.9.1

### Wrapper Library

Not Applicable

### Wrapper Library Version

None

### Description

Since upgrading from 1.9.0 to 1.9.1 we are running into the error listed in this issue below. We have a node application that runs on an interval to update profile pictures in [Backstage a Developer Portal](https://backstage.io/) we are using. It's very simple.

Rolling back to 1.9.0 the error goes away. I'm not sure what specifically is causing this issue but I feel like it is something in PR #4744. That being said there might be simply something that we need to include to resolve this like some additional configuration.

Any help with this would be greatly appreciated!

### Error Message

ClientAuthError: endpoints_resolution_error: Error: could not resolve endpoints. Please check network and try again. Detail: ClientAuthError: openid_config_error: Could not retrieve endpoints. Check your authority and verify the .well-known/openid-configuration endpoint returns the required endpoints. Attempted to retrieve endpoints from: https://login.microsoftonline.com/{tenantId}/v2.0/.well-known/openid-configuration

Note that `{tenantId}` in the URL above is our actual tenantId, just removed for security reasons.

### MSAL Configuration

```javascript
const clientConfig = {
      auth: {
        clientId,
        clientSecret,
        authority: `https://login.microsoftonline.com/${tenantId}`,
      },
    };
```

### Relevant Code Snippets

```javascript
const clientConfig = {
      auth: {
        clientId,
        clientSecret,
        authority: `https://login.microsoftonline.com/${tenantId}`,
      },
    };

    const authClient = new ConfidentialClientApplication(clientConfig);

    const tokens = await authClient.acquireTokenByClientCredential({
      scopes: ['https://graph.microsoft.com/.default'],
    });
```

### Reproduction Steps

I'm not totally sure how to describe this:

1. Upgrade to @azure/msal-node 1.9.1
2. Run the application

Notice in the logs the error listed in this issue occurs

### Expected Behavior

There should be no error and the process should continue to work as it did with version 1.9.0

### Identity Provider

Azure AD / MSA

### Browsers Affected (Select all that apply)

None (Server)

### Regression

@azure/msal-node 1.9.0

### Source

External (Customer)

awanlin:
[...]

PR #5658, opened 2023-02-07
Custom-INetworkModule-and-Network-Tracing sample now supports an axios HttpClient as well as requests running on an express server.

Robbie-Microsoft:
This sample has been updated to help debug issues in [GitHub Issue 4879](https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/4879) - one user is having difficulties with ConfidentialClient in conjunction with an express server. There are two ways to run the sample now: one is as-is, the other via an express server.

Additionally, in order to prepare for potentially adding axios functionality back to msal-node's HttpClient in the case where the developer isn't using a proxyUrl, a HttpClientAxios was added to this sample to give the developer the option to use as a networkClient instead of the current functionality. The developer now has the option to use either networkClient (or neither) in this sample.

### DIFF
--- a/samples/msal-node-samples/custom-INetworkModule-and-network-tracing/HttpClientCurrent.ts
+++ b/samples/msal-node-samples/custom-INetworkModule-and-network-tracing/HttpClientCurrent.ts
@@ -53,7 +53,17 @@ class NetworkUtils {
 /**
  * This class implements the API for network requests.
  */
-export class HttpClient implements INetworkModule {
+export class HttpClientCurrent implements INetworkModule {
+    private proxyUrl: string;
+    private customAgentOptions: http.AgentOptions | https.AgentOptions;
+
+    constructor(
+        proxyUrl?: string,
+        customAgentOptions?: http.AgentOptions | https.AgentOptions,
+    ) {
+        this.proxyUrl = proxyUrl || "";
+        this.customAgentOptions = customAgentOptions || {};
+    }
 
     /**
      * Http Get request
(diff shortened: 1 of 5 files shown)


samples/msal-node-samples/custom-INetworkModule-and-network-tracing/HttpClientCurrent.ts
```diff
@@ -53,7 +53,17 @@ class NetworkUtils {
 /**
  * This class implements the API for network requests.
  */
-export class HttpClient implements INetworkModule {
+export class HttpClientCurrent implements INetworkModule {
```
bgavrilMS, line 56: naming? Maybe `AxiosHttpClient` ?

bgavrilMS (approved):

samples/msal-node-samples/custom-INetworkModule-and-network-tracing/HttpClientCurrent.ts
derisen: Maybe `HttpClientNative` would be more apt?

samples/msal-node-samples/custom-INetworkModule-and-network-tracing/express.ts
```diff
@@ -0,0 +1,68 @@
...
+
+const request: msal.ClientCredentialRequest = {
+    scopes: ["https://graph.microsoft.com/.default"],
+};
+
+const app = express();
+
+// type "http://localhost:3000" into the browser to trigger this route
+app.get("/", async (req, res) => {
+    console.log("Hello World");
```
derisen: Was the intention here to signal that the request is received? Maybe something like `console.log('Request recevied - ', new Date()) `

Robbie-Microsoft: Yep - I can change that.

derisen (approved):
LGTM -thanks @Robbie-Microsoft

PR #5662, opened 2023-02-07
Trigger e2e tests only for PRs that are ready for review

konstantin-msft:
- Trigger e2e tests only for PRs that are ready for review.

A follow-up for https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/5649.

### DIFF
--- a/.github/workflows/msal-angular-e2e.yml
+++ b/.github/workflows/msal-angular-e2e.yml
@@ -1,13 +1,19 @@
 # This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
 # For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
+# Will not trigger unless PR is ready for review.
 
 name: msal-angular E2E Tests
 
 on:
+  push:
+    branches:
+      - dev
+      - master
   pull_request:
-    paths: 
+    types: [opened, reopened, synchronize, ready_for_review]
+    paths:
       - 'lib/msal-angular/**/*'
-      - 'lib/msal-browser/**/*' 
+      - 'lib/msal-browser/**/*'
       - 'lib/msal-common/**/*'
       - 'samples/msal-angular-v2-samples/**/*'
       - 'samples/e2eTestUtils/**/*'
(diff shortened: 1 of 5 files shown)


.github/workflows/msal-angular-e2e.yml
```diff
@@ -28,7 +30,7 @@ permissions:
 
 jobs:
   run-e2e:
-    if: (github.repository == 'AzureAD/microsoft-authentication-library-for-js') && (github.actor != 'dependabot[bot]') && ((github.event.pull_request.head.repo.full_name == github.repository) || (github.event_name == 'push'))
+    if: (github.repository == 'AzureAD/microsoft-authentication-library-for-js') && (github.actor != 'dependabot[bot]') && ((github.event.pull_request.head.repo.full_name == github.repository) || (github.event_name == 'push')) && !github.event.pull_request.draft
```
tnorling: Big fan 👍

.github/workflows/msal-angular-e2e.yml
```diff
@@ -1,17 +1,19 @@
...
 
 on:
   push:
-    branches: 
+    branches:
       - dev
       - master
   pull_request:
-    paths: 
+    types: [opened, reopened, synchronize, ready_for_review]
```
tnorling, line 9: `synchronize` triggers on every commit so the rest become redundant here. I think what you're trying to do here is ready_for_review && any of the previous 3 triggers? The ready for review check would need to go below along with the draft check instead to achieve that.

I think I'd also be fine with only the draft check below for now if we can all be diligent about using draft PRs when still in progress.

konstantin-msft: We have to keep at least "synchronize" and "ready_for_review" to kick off e2e workflow once PR is ready for review without pushing new commits to the branch.

tnorling: Ah, my apologies, I misunderstood (was still thinking "approved"), thanks for clarifying.

tnorling (approved):

hectormmg (approved):

.github/workflows/msal-angular-e2e.yml
sameerag: Can we have draft as the default setting when we open a PR for the repo?

sameerag (approved):
Minor comment on Draft PRs, lgtm otherwise.

.github/workflows/msal-angular-e2e.yml
tnorling: Github defaults to whatever you personally used last. Don't think there is a repo wide setting.

peterzenz (approved):

PR #5665, opened 2023-02-07
Update rollup and associated plugins to latest versions

shoatman:
- Updated first to 2.79.1 and added --strictDeprecations: No dependencies on deprecated features identified.
- Updated to latest rollup and updated all plugins to latest versions
- prior terser plugin was deprecated in favor of @rollup/plugin-terser
- Updated to official typescript plugin
- rollup config for msal-browser updated to import the new version of terser, official typescript plugin
- package.json files for common and msal-browser updated to use the --bundleConfigAsCjs flag.... rather than updating config to ES Modules
- verified build of common and msal-browser locally


lib/msal-browser/package.json
```diff
@@ -90,6 +90,6 @@
     "typescript": "^3.8.3"
   },
   "dependencies": {
-    "@azure/msal-common": "^10.0.0"
+    "@azure/msal-common": "file:../msal-common/dist"
```
tnorling: We do this so that people can take advantage of new versions of common as they come out without needing to update to the latest msal-browser

lib/msal-browser/package.json
```diff
@@ -78,9 +78,9 @@
     "jsdom": "^20.0.0",
     "jsdom-global": "^3.0.2",
     "rimraf": "^3.0.0",
-    "rollup": "^2.46.0",
-    "rollup-plugin-terser": "^7.0.2",
-    "rollup-plugin-typescript2": "^0.29.0",
+    "rollup": "^3.14.0",
+    "@rollup/plugin-terser": "^0.4.0",
+    "rollup-plugin-typescript2": "^0.34.1",
```
tnorling: Did you consider switching to the [official typescript plugin](https://www.npmjs.com/package/@rollup/plugin-typescript)?

shoatman: Nice.  I didn't know that existed.  In the case of terser the original plugin was deprecated.  I'll give this a try.  Thanks!

lib/msal-browser/package.json
shoatman: Ok.  I'll rollback this change and see if things build.... i think we'll get a conflict in node_modules, but we'll see.

lib/msal-common/package.json
```diff
@@ -63,8 +63,8 @@
     "jest": "^27.0.4",
     "lodash": "^4.17.21",
     "rimraf": "^3.0.2",
-    "rollup": "^2.46.0",
-    "rollup-plugin-typescript2": "^0.29.0",
+    "rollup": "^3.14.0",
```
tnorling, line 66: Looks like this requires Node 14+, which becomes a problem for our Node 10/12 pipelines

shoatman: Good a forcing function for our discussion about what the plan is for common :-)

lib/msal-browser/package.json
shoatman: Fixed.

lib/msal-common/package.json
shoatman: @sameerag - What's the plan for Node vNext?  @hectormmg - FYI

shoatman: Are those pipelines checks for PRs?  I didn't see a link to those.

tnorling: Hmm interesting...looks like [they were disabled](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/f4678bff7332890e0f3e680244a9f6557ef6015b/.github/workflows/msal-node-e2e.yml#L126) for some reason...@hectormmg any idea why?

sameerag: Looks like this: Some tooling no longer supports old versions of node - this job will downgrade certain dependencies for testing

sameerag: @shoatman msal-node v2 looks like in a quarter as per the plan (provided we have resourcing - min case we can drop old node versions). Let's take this offline, cc @bgavrilMS

lib/msal-browser/package.json
```diff
@@ -55,7 +55,7 @@
     "test:coverage:only": "npm run clean:coverage && npm run test:coverage",
     "build:all": "npm run build:common && npm run build",
     "build:common": "cd ../msal-common && npm run build",
-    "build:modules": "rollup -c",
+    "build:modules": "rollup -c --strictDeprecations --bundleConfigAsCjs",
```
sameerag, line 58: why common-js bundling?

lib/msal-common/package.json
hectormmg: @tnorling I vaguely remember us making the decision to temporarily disable them because they were breaking without a clear cause. We just never came back around to it until now.

@shoatman I haven't had a Node v2 discussion with anyone. I'm not strictly opposed to releasing Node at the same time as the browser libraries, granting that we'll lag behind (or skip ahead) the Node.js major version release schedule by 6mos.

hectormmg: @tnorling https://github.com/AzureAD/microsoft-authentication-library-for-js/actions/runs/4127019461/jobs/7129701390

Lerna started throwing "Unexpected token . " on Node 10 and 12, which seems like a breaking change in lerna.

hectormmg: Per this thread https://github.com/lerna/lerna/issues/3368 lerna doesn't support either Node 10 nor 12 (as expected, honestly)

lib/msal-browser/package.json
shoatman: Just avoiding making more invasive changes to the current configuration: https://rollupjs.org/command-line-interface/#bundleconfigascjs

sameerag (approved):
Lgtm. Please merge after we sort the Node 10, 12 support. Approving.

lib/msal-common/package.json
sameerag: So the plan is to leave it as is and accelerate dropping these for msal-node?

hectormmg: @sameerag I'm not sure. We will eventually move to workspaces and drop lerna (I believe) but dropping unsupported versions of Node ASAP also has to happen. I don't see a way to keep 10 and 12 in our CI/CD without a lerna fix.

PR #5666, opened 2023-02-08
Enable Node v10 and 12 on E2E pipeline

hectormmg:
This PR uncomments MSAL Node E2E tests for versions 10 and 12 of Node.js


sameerag:
It is failing tests?

hectormmg:
@sameerag yeah this is not going to work as long as we use lerna

PR #5676, opened 2023-02-10
Reduce telemetry RAM footprint and improve usability/readability

konstantin-msft:
RAM footprint reduction:
- Reduce `eventsByCorrelationId` map RAM usage by ~88%. We store 8 sub-measurements on average as separate events in `eventsByCorrelationId` map that are rolled-up into root event on ending.
- Reduce `measurementsByCorrelationId` map RAM usage by 100%. Measurements are not stored in map anymore. 

Code changes:
- Store a single root event per correlation id. Sub-measurements are populated to root event on ending.
- Store static fields and counters in root event instead of caching them.
- Don't cache measurements. Measurement is an attribute of instance of `InProgressPerformanceEvent`. 
- Deprecate `flushMeasurement` event and move all logic to `endMeasurement`. This allows to end measurement with a single method.
- Add a method to flush incomplete browser measurements.

Notes:
- `PerformanceClient.flushMeasurement()` will be removed completely right after i remove all its` references in 1p repo.

### DIFF
--- a/lib/msal-browser/src/app/ClientApplication.ts
+++ b/lib/msal-browser/src/app/ClientApplication.ts
@@ -558,9 +558,6 @@ export abstract class ClientApplication {
                     this.hybridAuthCodeResponses.set(hybridAuthCode, response);
                 } else {
                     this.logger.verbose("Existing acquireTokenByCode request found", request.correlationId);
-                    atbcMeasurement.endMeasurement({
-                        success: true
-                    });
                     atbcMeasurement.discardMeasurement();
                 }
                 return response;
--- a/lib/msal-browser/src/app/ClientApplication.ts
+++ b/lib/msal-browser/src/app/ClientApplication.ts
@@ -357,7 +357,6 @@ export abstract class ClientApplication {
                     isNativeBroker: true,
                     requestId: response.requestId
                 });
-                atPopupMeasurement.flushMeasurement();
                 return response;
             }).catch((e: AuthError) => {
                 if (e instanceof NativeAuthError && e.isFatal()) {
@@ -397,8 +396,6 @@ export abstract class ClientApplication {
                 success: true,
                 requestId: result.requestId
             });
-
-            atPopupMeasurement.flushMeasurement();
             return result;
         }).catch((e: AuthError) => {
             if (loggedInAccounts.length > 0) {
@@ -412,8 +409,6 @@ export abstract class ClientApplication {
                 subErrorCode: e.subError,
                 success: false
             });
-            atPopupMeasurement.flushMeasurement();
-
             // Since this function is syncronous we need to reject
             return Promise.reject(e);
         });
@@ -494,7 +489,6 @@ export abstract class ClientApplication {
                 isNativeBroker: response.fromNativeBroker,
                 requestId: response.requestId
             });
-            this.ssoSilentMeasurement?.flushMeasurement();
             return response;
         }).catch((e: AuthError) => {
             this.eventHandler.emitEvent(EventType.SSO_SILENT_FAILURE, InteractionType.Silent, null, e);
@@ -503,7 +497,6 @@ export abstract class ClientApplication {
                 subErrorCode: e.subError,
                 success: false
             });
-            this.ssoSilentMeasurement?.flushMeasurement();
             throw e;
         }).finally(() => {
             document.removeEventListener("visibilitychange",this.trackPageVisibilityWithMeasurement);
@@ -550,7 +543,6 @@ export abstract class ClientApplication {
                                 isNativeBroker: result.fromNativeBroker,
                                 requestId: result.requestId
                             });
-                            atbcMeasurement.flushMeasurement();
                             return result;
                         })
                         .catch((error: AuthError) => {
@@ -561,7 +553,6 @@ export abstract class ClientApplication {
                                 subErrorCode: error.subError,
                                 success: false
                             });
-                            atbcMeasurement.flushMeasurement();
                             throw error;
                         });
                     this.hybridAuthCodeResponses.set(hybridAuthCode, response);
--- a/lib/msal-browser/src/app/PublicClientApplication.ts
+++ b/lib/msal-browser/src/app/PublicClientApplication.ts
@@ -147,7 +147,6 @@ export class PublicClientApplication extends ClientApplication implements IPubli
                         cacheLookupPolicy: request.cacheLookupPolicy,
                         requestId: result.requestId,
                     });
-                    atsMeasurement.flushMeasurement();
                     return result;
                 })
                 .catch((error: AuthError) => {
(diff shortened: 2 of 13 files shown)


lib/msal-common/src/telemetry/performance/IPerformanceClient.ts
```diff
@@ -21,6 +21,7 @@ export type InProgressPerformanceEvent = {
 export interface IPerformanceClient {
     startMeasurement(measureName: PerformanceEvents, correlationId?: string): InProgressPerformanceEvent;
     endMeasurement(event: PerformanceEvent): PerformanceEvent | null;
+    // @deprecated use "endMeasurement" instead
```
peterzenz, line 24: This is OK for a patch version?

peterzenz: Looks like it isn't going away so this is likely OK

peterzenz (approved):

lib/msal-common/src/telemetry/performance/IPerformanceClient.ts
konstantin-msft: I will mark a consequent PR that removes `flushMeasurement` method as minor.

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -561,15 +553,11 @@ export abstract class ClientApplication {
                                 subErrorCode: error.subError,
                                 success: false
                             });
-                            atbcMeasurement.flushMeasurement();
                             throw error;
                         });
                     this.hybridAuthCodeResponses.set(hybridAuthCode, response);
                 } else {
                     this.logger.verbose("Existing acquireTokenByCode request found", request.correlationId);
-                    atbcMeasurement.endMeasurement({
```
tnorling, line 570: Why was this removed?

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -157,16 +156,12 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
                         success: false
                     });
-                    atsMeasurement.flushMeasurement();
                     throw error;
                 });
             this.activeSilentTokenRequests.set(silentRequestKey, response);
             return response;
         } else {
             this.logger.verbose("acquireTokenSilent has been called previously, returning the result from the first call", correlationId);
-            atsMeasurement.endMeasurement({
```
tnorling, line 167: Why was this removed?

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
```diff
@@ -269,34 +243,27 @@ export abstract class PerformanceClient implements IPerformanceClient {
...
         // Store in progress events so they can be discarded if not ended properly
         this.cacheEventByCorrelationId(inProgressEvent);
 
-        const staticFields: StaticFields = {
-            appName: this.applicationTelemetry?.appName,
-            appVersion: this.applicationTelemetry?.appVersion,
-        };
-        this.addStaticFields(staticFields, eventCorrelationId);
-        this.cacheMeasurement(inProgressEvent, validMeasurement);
+        // this.cacheMeasurement(inProgressEvent, validMeasurement);
```
tnorling: Should this be uncommented or removed?

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
```diff
@@ -27,30 +32,10 @@ export abstract class PerformanceClient implements IPerformanceClient {
...
-     * @type {Map<string, Map<string, PerformanceEvent>>}
-     */
-    protected eventsByCorrelationId: Map<string, Map<string, PerformanceEvent>>;
-
-    /**
-     * Fields to be emitted which are scoped to the top level request and whose value will not change in submeasurements
-     * For example: App name, version, etc.
+     * @type {Map<string, PerformanceEvent>}
      */
-    protected staticFieldsByCorrelationId: Map<string, StaticFields>;
```
tnorling, line 40: Removing these is a breaking change. This should be a major bump of common

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
```diff
@@ -318,51 +285,80 @@ export abstract class PerformanceClient implements IPerformanceClient {
...
      * @param fields
      * @param correlationId
      */
-    addStaticFields(fields: StaticFields, correlationId: string) : void{
-        const existingStaticFields = this.staticFieldsByCorrelationId.get(correlationId);
-        if (existingStaticFields) {
-            this.logger.trace("PerformanceClient: Updating static fields");
-            this.staticFieldsByCorrelationId.set(correlationId, {...existingStaticFields, ...fields});
+    addStaticFields(fields: StaticFields, correlationId: string) : void {
+        this.logger.trace("PerformanceClient: Upserting static fields");
```
tnorling: typo:
```suggestion
        this.logger.trace("PerformanceClient: Updating static fields");
```

lib/msal-browser/src/app/ClientApplication.ts
konstantin-msft: We don't need to end measurements before discarding them because no events are emitted in that case.

lib/msal-browser/src/app/PublicClientApplication.ts
konstantin-msft: We don't need to end measurements before discarding them because no events are emitted in that case.

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
konstantin-msft: I would rather disagree because:
1. It's does not change API signature
2. We don't have any classes inheriting base `PerformanceClient` class that access `staticFieldsByCorrelationId` directly.

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
konstantin-msft: fixed

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
konstantin-msft: fixed

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
tnorling: `PerformanceClient` is exported from common therefore removing anything public or protected is breaking. It's not a big deal though, we just need to mark this as a major change in the msal-common changefile. Please do make sure anything that is not used outside of this class is marked `private` to avoid this churn in the future though.

konstantin-msft: Sounds good. Changed to major version bump.

tnorling (approved):

PR #5677, opened 2023-02-13
Address comp governance issues

sameerag:
Addressing existing CVEs on the latest branch:

* [CVE-2022-37611](https://identitydivision.visualstudio.com/Engineering/_componentGovernance/132772/alert/7855035?action=ContributedHub&controller=Apps&typeId=10692653)
* CVE-2022-21680
* CVE-2022-21681

### DIFF
--- a/lib/msal-browser/package.json
+++ b/lib/msal-browser/package.json
@@ -87,7 +87,7 @@
     "ts-jest": "^27.0.2",
     "tslib": "^1.10.0",
     "tslint": "^5.20.0",
-    "typescript": "^3.8.3"
+    "typescript": "^4.9.5"
   },
   "dependencies": {
     "@azure/msal-common": "^10.0.0"
--- a/lib/msal-common/package.json
+++ b/lib/msal-common/package.json
@@ -70,7 +70,7 @@
     "ts-jest": "^27.1.5",
     "tslib": "^1.10.0",
     "tslint": "^5.20.0",
-    "typescript": "^3.7.5",
+    "typescript": "^4.9.5",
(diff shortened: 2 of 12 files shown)


peterzenz (approved):

bgavrilMS (approved):

lib/msal-browser/package.json
```diff
@@ -87,7 +87,7 @@
     "ts-jest": "^27.0.2",
     "tslib": "^1.10.0",
     "tslint": "^5.20.0",
-    "typescript": "^3.8.3"
+    "typescript": "^4.9.5"
```
tnorling, line 90: This will most certainly contain breaking changes that will affect our build output. Needs to wait until v3

lib/msal-common/package.json
```diff
@@ -70,7 +70,7 @@
     "ts-jest": "^27.1.5",
     "tslib": "^1.10.0",
     "tslint": "^5.20.0",
-    "typescript": "^3.7.5",
+    "typescript": "^4.9.5",
```
tnorling, line 73: Applies to all of these

lib/msal-browser/package.json
tnorling: And we should be using ~ versions as typescript doesn't follow semver

sameerag: cc @hectormmg @konstantin-msft Let me check the release notes before we make the call if we need to make this a part of v3.

sameerag:
Abandoning as V3 is addressing these (typescript update etc).

PR #5678, opened 2023-02-13
msal-node restores cache from persistence

### LINKED ISSUE #5502: OBO and Silent token acquisition does not read from disk cache when persistence is enabled
### Core Library

MSAL Node (@azure/msal-node)

### Core Library Version

1.14.5

### Wrapper Library

Not Applicable

### Wrapper Library Version

N/A

Public, Confidential

### Description

Silent token acquisition with MSAL Node (either via `acquireTokenSilent` or one of the internal methods that make use of the cache e.g. `acquireCachedToken`, `readAccessTokenFromCacheForOBO` etc.) does not trigger a cache read from disk when cache persistence is enabled. To trigger it, the developer must call one of the `getAccount*` APIs ([see](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/eb76935b373a0638e6b5c2afd8bb3c1830c8b080/lib/msal-node/src/cache/TokenCache.ts#L110)). While this may not be a bug per se, the usage is unintuitive and potentially misleading, as not all token acquisition APIs require or make use of the account object (e.g. `acquireTokenOnBehalfOf`). Even in the case of using `acquireTokenSilent` directly where expecting `getAccount*` APIs to be called beforehand is reasonable, developers may still find other (perhaps cheaper) means of retrieving and supplying an account object to the relevant API.

### Error Message

N/A

### Msal Logs

[Thu, 22 Dec 2022 18:29:26 GMT] : @azure/msal-node@1.14.5 : Info - acquireTokenOnBehalfOf called
[Thu, 22 Dec 2022 18:29:26 GMT] : @azure/msal-node@1.14.5 : Verbose - initializeRequestScopes called
[Thu, 22 Dec 2022 18:29:26 GMT] : [da92d27e-d8a5-4dc9-a192-e6e73b59d5e4] : @azure/msal-node@1.14.5 : Verbose - buildOauthClientConfiguration called
[Thu, 22 Dec 2022 18:29:26 GMT] : [da92d27e-d8a5-4dc9-a192-e6e73b59d5e4] : @azure/msal-node@1.14.5 : Verbose - building oauth client configuration with the authority: https://login.microsoftonline.com/TENANT_ID
[Thu, 22 Dec 2022 18:29:26 GMT] : [da92d27e-d8a5-4dc9-a192-e6e73b59d5e4] : @azure/msal-node@1.14.5 : Verbose - createAuthority called
[Thu, 22 Dec 2022 18:29:26 GMT] : @azure/msal-node@1.14.5 : Trace - Retrieving all cache keys
[Thu, 22 Dec 2022 18:29:26 GMT] : @azure/msal-node@1.14.5 : Trace - Getting cache key-value store
[Thu, 22 Dec 2022 18:29:26 GMT] : @azure/msal-node@1.14.5 : Verbose - Attempting to get cloud discovery metadata in the config  
[Thu, 22 Dec 2022 18:29:26 GMT] : @azure/msal-node@1.14.5 : Verbose - Did not find cloud discovery metadata in the config... Attempting to get cloud discovery metadata from the cache.
[Thu, 22 Dec 2022 18:29:26 GMT] : @azure/msal-node@1.14.5 : Verbose - Did not find cloud discovery metadata in the cache... Attempting to get cloud discovery metadata from the network.        
[Thu, 22 Dec 2022 18:29:26 GMT] : @azure/msal-node@1.14.5 : Verbose - Attempting to find a match between the developer's authority and the CloudInstanceDiscoveryMetadata returned from the network request.
[Thu, 22 Dec 2022 18:29:26 GMT] : @azure/msal-node@1.14.5 : Verbose - cloud discovery metadata was successfully returned from getCloudDiscoveryMetadataFromNetwork()
[Thu, 22 Dec 2022 18:29:26 GMT] : @azure/msal-node@1.14.5 : Trace - Getting cache key-value store
[Thu, 22 Dec 2022 18:29:26 GMT] : @azure/msal-node@1.14.5 : Trace - Setting cache key value store
[Thu, 22 Dec 2022 18:29:26 GMT] : [da92d27e-d8a5-4dc9-a192-e6e73b59d5e4] : @azure/msal-node@1.14.5 : Verbose - On behalf of client created
[Thu, 22 Dec 2022 18:29:26 GMT] : @azure/msal-node@1.14.5 : Trace - Retrieving all cache keys
[Thu, 22 Dec 2022 18:29:26 GMT] : @azure/msal-node@1.14.5 : Trace - Getting cache key-value store
[Thu, 22 Dec 2022 18:29:26 GMT] : [da92d27e-d8a5-4dc9-a192-e6e73b59d5e4] : @azure/msal-common@9.0.1 : Info - SilentFlowClient:acquireCachedToken - No access token found in cache for the given 
properties.
[Thu, 22 Dec 2022 18:29:26 GMT] : @azure/msal-node@1.14.5 : Trace - Getting cache key-value store
[Thu, 22 Dec 2022 18:29:27 GMT] : [da92d27e-d8a5-4dc9-a192-e6e73b59d5e4] : @azure/msal-common@9.0.1 : Verbose - Persistence enabled, calling beforeCacheAccess
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Deserializing JSON to in-memory cache
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Reading cache snapshot from disk
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Overlaying input cache with the default cache
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Setting in-memory cache
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Getting cache key-value store
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Setting cache key value store
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Getting cache key-value store
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Setting cache key value store
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Getting cache key-value store
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Setting cache key value store
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Retrieving all cache keys
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Getting cache key-value store
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Getting cache key-value store
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Retrieving all cache keys
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Getting cache key-value store
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Getting cache key-value store
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Getting cache key-value store
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Retrieving all cache keys
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Getting cache key-value store
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Getting cache key-value store
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Getting cache key-value store
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Retrieving all cache keys
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Getting cache key-value store
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Getting cache key-value store
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Getting cache key-value store
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Setting cache key value store
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Getting cache key-value store
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Setting cache key value store
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Getting cache key-value store
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Setting cache key value store
[Thu, 22 Dec 2022 18:29:27 GMT] : [da92d27e-d8a5-4dc9-a192-e6e73b59d5e4] : @azure/msal-common@9.0.1 : Verbose - Persistence enabled, calling afterCacheAccess
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Serializing in-memory cache
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Getting in-memory cache
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Getting cache key-value store
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Reading cache snapshot from disk
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Merging in-memory cache with cache snapshot
[Thu, 22 Dec 2022 18:29:27 GMT] : @azure/msal-node@1.14.5 : Trace - Remove updated entries in cache

### MSAL Configuration

```javascript
const cca = new msal.ConfidentialClientApplication({
    auth: {
        clientId: "CLIENT_ID",
        authority: "https://login.microsoftonline.com/TENANT_ID",
        clientSecret: "CLIENT_SECRET",
    },
    cache: {
        cachePlugin: require('./cachePlugin')('./cache.json')
    },
    system: {
        loggerOptions: {
                console.log(message);
            },
            piiLoggingEnabled: false,
            logLevel: msal.LogLevel.Trace,
        }
    }
});
```

### Relevant Code Snippets

```javascript
app.get('/getOboToken', async (req, res, next) => {
    try {
        const oboRequest = {
            oboAssertion: req.headers.authorization.split(' ')[1],
            scopes: ["user.read", "offline_access"],
        };

        const response = await cca.acquireTokenOnBehalfOf(oboRequest);

        res.send(response);
    } catch(error) {
        next(error)
    }
});
```

### Reproduction Steps

For `acquireTokenOnBehalfOf`:

1. Grab the msal-node [on-behalf-of sample](https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev/samples/msal-node-samples/on-behalf-of)
2. Configure cache persistence via a cache plugin.
3. Configure the client app to use `acquireTokenSilent`
4. Run the flow end to end once.
5. Restart the service app.
6. Run the flow again, expect the service app to retrieve the token from cache (i.e. `response.fromCache: true`)

For `acquireTokenSilent`:

1. Grab the msal-node [silent-flow sample](https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev/samples/msal-node-samples/silent-flow)
2. Configure cache persistence via a cache plugin.
7. Configure session persistence via a session middleware (e.g. express-session)
8. Save the account object in session after `acquireTokenByCode`
9. Next time when calling `acquireTokenSilent`, use the account object in the session (comment out any lines where `getAccount*` is called)
10. Restart the app, and attempt `acquireTokenSilent` again
11. Expect the token to be retrieved from cache (i.e. `response.fromCache: true`)

### Expected Behavior

When cache persistence is enabled, retrieving tokens from cache should read from disk beforehand by default

### Identity Provider

Azure AD / MSA

None (Server)

### Regression

_No response_

### Source

Internal (Microsoft)

derisen:
This PR fixes #5502 cc @sameerag 

**Problem:**
Currently, msal-node `acquireToken*` APIs do not read from disk cache by default. This breaks cache persistence use cases unless devs call `getAccount*` APIs explicitly before `acquireToken*`.

**Implementation**
Restore cache from disk within msal-node `ClientApplication.buildOauthClientConfiguration`. Call site is chosen as such since (1) cache read is an async operation, (2) applies to all `acquireToken*` APIs and (3) a populated cache is conceptually part of the configuration for a client object.

**Alternatives**
A similar effect can be achieved on the msal-common side, within each `*Client`'s `acquireToken `method. However this would lead to code duplication as each client has to perform this within.

**Breaking changes:**
None

_work-in-progress:_

* decision on whether to expose `restoreCacheContext ` as public API


lib/msal-node/src/cache/TokenCache.ts
```diff
@@ -97,6 +97,25 @@ export class TokenCache implements ISerializableTokenCache, ITokenCache {
         return this.storage.getCache();
     }
 
+    /**
+     * API that restores cache context from persistence
+     */
+    async restoreCacheContext(): Promise<void> {
+        this.logger.trace("refreshCacheContext called");
+        if (this.persistence && StringUtils.isEmpty(this.cacheSnapshot)) {
```
sameerag: `CacheSnapShot` not being empty does not signify it being current, does it? If that is the case, we might as well just check for persistence. Else we may have a cache hit but will not have accuracy.

derisen: Ok, I think I see your point. My intention was to avoid running `beforeCacheAccess` here unnecessarily (with the assumption that this is only needed when in-memory cache is clean i.e. a new app instance) since it will have _some_ performance impact, but in a distributed scenario we cannot guarantee cacheSnapshot accuracy. In that case, I suppose we should not do this check and accept the perf hit :/

lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -346,16 +347,21 @@ export abstract class ClientApplication {
 
         serverTelemetryManager?.updateRegionDiscoveryMetadata(discoveredAuthority.regionDiscoveryMetadata);
 
+        // if cache is persisted, restore cache context if needed before building client configuration
+        if (this.config.cache.cachePlugin) {
```
sameerag, line 351: Does setting a `cachePlugin` mean it is `shared`? I understand that is how we have it today, but do we have a case where we can different `persistent` and `shared` cache where we need not read/write to the disk ALWAYS if the cache isn't shared? I think we should separate them with a config param. This will be a feature though.

Can be a separate PR and also point to note for web apps and cache, cc @bgavrilMS @EmLauber

sameerag: Same thing will help with not sticking this code in `getAllAccounts()` every time it is called and the user chose a Disk cache option.

sameerag (changes requested):
Few comments. You can postpone addressing one but can we look into the `cacheSnapshot` assumption please?

lib/msal-node/src/cache/TokenCache.ts
derisen: Sorry for the late reply. I've thought more about this -you're right, we cannot make this assumption i.e. in-memo cache needs to be synced with disk cache before and after each `acquireToken` call to ensure consistency (e.g. app1 and app2 shares cache, later app2 logs out the user, removes tokens from the shared cache, but app1 still has tokens in the in-memory cache. expectation would be that app1 brings its in-memo cache to sync before making another token request). As such, I've removed the check.
 
From what I see, performance hit seems to be negligible with local cache -I'll measure again with a remote location and update just in case. 

One potential inefficiency is due to usage: if a dev is calling `getAccount*` right before `acquireToken*`, `beforeCacheAccess`/`afterCacheAccess` pair will be called twice in a row. Any thoughts about this? @sameerag @bgavrilMS

lib/msal-node/src/cache/TokenCache.ts
```diff
@@ -97,6 +97,24 @@ export class TokenCache implements ISerializableTokenCache, ITokenCache {
         return this.storage.getCache();
     }
 
+    /**
+     * API that restores cache context from persistence
+     */
+    async restoreCacheContext(): Promise<void> {
```
derisen, line 103: This code overlaps a lot with `getAllAccounts` and `removeAccount` now. I propose we refactor them:

```typescript
    private async handleCacheAccess<T>(action: () => Promise<T>, hasCacheChanged: boolean): Promise<T> {
        let cacheContext;
        try {
            if (this.persistence) {
                cacheContext = new TokenCacheContext(this, hasCacheChanged);
                await this.persistence.beforeCacheAccess(cacheContext);
            }
            return await action();
        } finally {
            if (this.persistence && cacheContext) {
                await this.persistence.afterCacheAccess(cacheContext);
            }
        }
    }
    
    async restoreCacheContext(): Promise<void> {
        this.logger.trace("refreshCacheContext called");
        return this.handleCacheAccess(async () => { }, false);
    }

    async getAllAccounts(): Promise<AccountInfo[]> {
        this.logger.trace("getAllAccounts called");
        return this.handleCacheAccess(async () => this.storage.getAllAccounts(), false);
    }

    async removeAccount(account: AccountInfo): Promise<void> {
        this.logger.trace("removeAccount called");

        return this.handleCacheAccess(async () => {
            await this.storage.removeAccount(AccountEntity.generateAccountCacheKey(account));
        }, true);
    }
```

what do you think? @sameerag @bgavrilMS

lib/msal-node/src/client/ClientApplication.ts
derisen: I suppose this could be an option, in particular for performance. It would be equivalent to saying "hydrate the in-memo cache with persistent cache when its empty, but then let in-memo cache be the single source of truth after". In that case, if the persistent cache is modified outside of msal, msal's in-memo cache will diverge from the persistent cache until tokens in the in-memo cache expires (after that, msal acquires tokens from network, then hydrates the in-memo cache again with persistent cache in `handleServerTokenResponse`)

lib/msal-node/src/cache/TokenCache.ts
derisen: following up on this -a cloud cache service like Azure Redis seems to incur about a 30-40ms latency on average to token requests (note that this would be the same if you were calling `getAccounts*` explicitly to trigger cache read)

sameerag: This is probably too high per token fetch. Let us discuss once, if we have a better solution.

lib/msal-node/src/cache/TokenCache.ts
sameerag: Looks good for me. Please test the non-persistent cases for `removeAccount` carefully. Also some refactor of cache is happening now, may be it will impact you/may be not. Please watch out `msal-common` code that could change.

sameerag: @derisen @tnorling made changes on cache lookups (includes accounts and tokens). Can you please update this PR with the latest branch and I can sign off?

derisen:
We are discussing a better alternative to handle this, so converting this to draft for the moment.

derisen:
Closing this PR to in favor of an alternative approach.

PR #5398, opened 2022-11-16
msal-node serializes/deserializes userAssertionHash

derisen:
In response to #5397 

p.s. will update the [obo sample](https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev/samples/msal-node-samples/on-behalf-of) separately for manual testing. will also check if MSIDLabs support OBO flow for E2E testing.

Issue #5397, opened 2022-11-16
MSAL Node does not serialize/desiralize userAssertionHash in OBO


derisen:
### Core Library

MSAL Node (@azure/msal-node)

### Core Library Version

1.14.3

### Wrapper Library

Not Applicable

### Wrapper Library Version

N/A

### Public or Confidential Client?

Confidential

### Description

At the moment msal-node `Serializer` and `Deserializer` classes do not process `userAssertionHash` property in `AccessTokenEntity`. This means when cache is persisted via a cachePlugin, msal-node will not be able to read `userAssertionHash` from the serialized cache entry. Since `userAssertionHash` is a filter property for retrieving a cached AT in OBO flow, this causes the client object's `getCachedAuthenticationResult` method to return null when in memory cache is empty (e.g. server restarts etc.)

### Error Message

N/A

### Msal Logs

[Wed, 16 Nov 2022 20:19:53 GMT] : @azure/msal-node@1.14.3 : Info - getTokenCache called
[Wed, 16 Nov 2022 20:19:51 GMT] : @azure/msal-node@1.14.3 : Info - acquireTokenOnBehalfOf called
[Wed, 16 Nov 2022 20:19:51 GMT] : @azure/msal-node@1.14.3 : Verbose - initializeRequestScopes called
[Wed, 16 Nov 2022 20:19:51 GMT] : [a325b97b-eb28-49de-ae45-bd0d141b7c57] : @azure/msal-node@1.14.3 : Verbose - buildOauthClientConfiguration called
[Wed, 16 Nov 2022 20:19:51 GMT] : [a325b97b-eb28-49de-ae45-bd0d141b7c57] : @azure/msal-node@1.14.3 : Verbose - building oauth client configuration with the authority: https://login.microsoftonline.com/cbaf2168-de14-4c72-9d88-f5f05366dbef
[Wed, 16 Nov 2022 20:19:51 GMT] : [a325b97b-eb28-49de-ae45-bd0d141b7c57] : @azure/msal-node@1.14.3 : Verbose - createAuthority called
[Wed, 16 Nov 2022 20:19:52 GMT] : [a325b97b-eb28-49de-ae45-bd0d141b7c57] : @azure/msal-node@1.14.3 : Verbose - On behalf of client created
[...]

PR #5680, opened 2023-02-13
OBO E2E test

Robbie-Microsoft:


.github/workflows/msal-node-e2e.yml
```diff
@@ -104,7 +105,11 @@ jobs:
       env:
         AZURE_CLIENT_SECRET: ${{ secrets.AZURE_CLIENT_SECRET }}
         AZURE_TENANT_ID: ${{ secrets.AZURE_TENANT_ID }}
-        AZURE_CLIENT_ID: ${{ secrets.AZURE_CLIENT_ID }}
```
derisen, line 109: is this accidental?

Robbie-Microsoft: Yep! I just changed it back. Was wondering why every e2e test was failing...

samples/msal-node-samples/on-behalf-of/test/obo-aad.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+    webApiConfig.authOptions.clientId = process.env.OBO_WEB_API_CLIENT_ID;
+    webApiConfig.webApiUrl = `api://${process.env.OBO_WEB_API_CLIENT_ID}/access_as_user`;
+}
+if (process.env.OBO_WEB_API_CLIENT_SECRET) {
+    webApiConfig.authOptions.clientSecret = process.env.OBO_WEB_API_CLIENT_SECRET;
+}
+if (process.env.OBO_WEB_API_TENANT_ID) {
+    webApiConfig.authOptions.authority = `https://login.microsoftonline.com/${process.env.OBO_WEB_API_TENANT_ID}`;
+    webApiConfig.discoveryKeysEndpoint = `https://login.microsoftonline.com/${process.env.OBO_WEB_API_TENANT_ID}/discovery/v2.0/keys`;
+}
```
derisen: curios why we need these from the environment -can we not query the lab api to get retrieve them at test run time? ([e.g.](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/2e4df196d41db10216889d0b360b9abc6c0cee7b/samples/msal-node-samples/client-credentials/test/client-credentials-aad.spec.ts#L37))

samples/msal-node-samples/on-behalf-of/test/obo-aad.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+
+const TEST_CACHE_LOCATION = `${__dirname}/data/cache.json`;
+const cachePlugin = require("../../cachePlugin.js")(TEST_CACHE_LOCATION);
+
+const HOME_ROUTE = `http://localhost:${webAppConfig.serverPort}`;
+
+describe("OBO AAD PPE Tests", () => {
```
derisen:
```suggestion
describe("OBO AAD Tests", () => {
```

samples/msal-node-samples/on-behalf-of/index.js
```diff
@@ -0,0 +1,211 @@
...
+    webApiConfig.authOptions.clientId = process.env.OBO_WEB_API_CLIENT_ID;
+    webApiConfig.webApiUrl = `api://${process.env.OBO_WEB_API_CLIENT_ID}/access_as_user`;
+}
+if (process.env.OBO_WEB_API_CLIENT_SECRET) {
+    webApiConfig.authOptions.clientSecret = process.env.OBO_WEB_API_CLIENT_SECRET;
+}
+if (process.env.OBO_WEB_API_TENANT_ID) {
+    webApiConfig.authOptions.authority = `https://login.microsoftonline.com/${process.env.OBO_WEB_API_TENANT_ID}`;
+    webApiConfig.discoveryKeysEndpoint = `https://login.microsoftonline.com/${process.env.OBO_WEB_API_TENANT_ID}/discovery/v2.0/keys`;
+}
```
derisen: probably don't need these here (if someone is manually running the app, they would be setting the params in `config/*.json` files)

samples/msal-node-samples/on-behalf-of/jest.config.js
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+module.exports = {
+    displayName: "Device Code",
```
derisen:
```suggestion
    displayName: "On behalf of",
```

samples/msal-node-samples/on-behalf-of/test/obo-aad.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+            const cachedTokens = await NodeCacheTestUtils.waitForTokens(TEST_CACHE_LOCATION, 2000);
+            expect(cachedTokens.accessTokens.length).toBe(2);
+            expect(cachedTokens.idTokens.length).toBe(2);
+            expect(cachedTokens.refreshTokens.length).toBe(2);
+
+            const accounts = await NodeCacheTestUtils.getAccounts(TEST_CACHE_LOCATION);
+            expect(Object.keys(accounts).length).toBe(1);
+            const account = Object.values(accounts)[0];
+            expect(account.username).toEqual(username);
+        });
```
derisen, line 141: Can we add another test case to check if obo retrieves the cached access token (if any) instead of going to the network (essentially covering this [issue](https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/5397) and [PR here](https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/5398))

samples/msal-node-samples/on-behalf-of/index.js
```diff
@@ -0,0 +1,211 @@
+/*
```
derisen: we probably need to update the README as well if we'll use a single project instead of 2.

However I would much prefer maintaining the separate web app and web api projects if possible (at least separate .js files for each under the same project).

samples/msal-node-samples/on-behalf-of/test/obo-aad.spec.ts
Robbie-Microsoft: I'm using the same web app and web api apps as the dotnet obo e2e tests... they're not queryable in the lab.

samples/msal-node-samples/on-behalf-of/index.js
Robbie-Microsoft: I didn't touch the original folders for the web app and web api in here, so the readme is still accurate. Perhaps I could add info to the readme that further explains how to run the sample via index.js. Thoughts?

Robbie-Microsoft: Or we could exclude it from the readme - however the functionality inside index.js is needed for the jest tests.

.github/workflows/msal-node-e2e.yml
```diff
@@ -33,12 +33,10 @@ jobs:
...
           client-id: ${{ secrets.TEST_CLIENT_ID }}
           tenant-id: ${{ secrets.TEST_TENANT_ID }}
-          subscription-id: ${{ secrets.TEST_SUBSCRIPTION_ID }}
+          allow-no-subscriptions: true
   
       - name: 'Run az commands'
-        run: |
-          az account show
-          az group list
+        run: az keyvault secret show --name ${{ secrets.TEST_SECRET_NAME }} --vault-name ${{ secrets.TEST_VAULT_NAME }}
```
bgavrilMS: I suggest doing this in a separate PR. The OBO test with secret should be ready to go and provides good value as is.

samples/msal-node-samples/on-behalf-of/index.js
derisen: OK, in that case can we move index.js (and others related to the test) under the test/ folder then?

samples/msal-node-samples/on-behalf-of/test/obo-aad.spec.ts
Robbie-Microsoft: Looks like that test case is already covered here: [lib/msal-common/test/client/OnBehalfOfClient.spec.ts](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-common/test/client/OnBehalfOfClient.spec.ts#L235)

samples/msal-node-samples/on-behalf-of/test/index.js
```diff
@@ -0,0 +1,191 @@
...
+        const authCodeUrlParameters = {
+            scopes: [webApiUrl],
+            redirectUri: redirectUri,
+        };
+    
+        // get url to sign user in and consent to scopes needed for application
+        cca.getAuthCodeUrl(authCodeUrlParameters).then((response) => {
+            res.redirect(response);
+        }).catch((error) => {
+            console.log(JSON.stringify(error))
```
bgavrilMS: Shouldn't we throw and fail the test here?

samples/msal-node-samples/on-behalf-of/test/index.js
```diff
@@ -0,0 +1,191 @@
+/*
```
bgavrilMS: nit: why are all these samples / tests written in JS and not in TS?

bgavrilMS (approved):

samples/msal-node-samples/on-behalf-of/test/obo-aad.spec.ts
derisen: No this is only testing if the token is retrieved from in-memory cache. The issue comes up when restoring the cache from disk to in-memory (its mainly a serialization issue). To test that we would have to restart the server or instantiate a new CCA and load an already populated cache file, but I think we're fine without this -it's only a nice-to-have

samples/msal-node-samples/on-behalf-of/test/index.js
derisen, line 1: I'm still uneasy about this file. We are duplicating code unnecessarily and testing something other than what we offer to public (web-app/web-api folders). Why not just update the `web-api/index.js` and `web-app/index.js` in the same way i.e wrap the code in a function like you did with `acquireTokenObo` in `test/index.js` and import it in the test file and pass the relevant parameters.

samples/msal-node-samples/on-behalf-of/config/WEB-API.json
derisen, line 1: these two json files also seem unnecessary to me, since you have to fill the credentials during the test (and web-api/index.js, web-api/index.js define the credentials in the code file). I'd say we remove these to reduce clutter and prevent confusion (or alternatively, actually make use of them in the sample code).

samples/msal-node-samples/on-behalf-of/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "build:package": "cd ../../../lib/msal-common && npm run build && cd ../msal-node && npm run build",
+    "start:build": "npm run build:package && npm start",
+    "install:local": "npm install ../../../lib/msal-node"
+  },
+  "author": "Microsoft",
+  "license": "MIT",
+  "dependencies": {
+    "@azure/identity": "^3.1.3",
+    "@azure/keyvault-certificates": "^4.6.0",
+    "@azure/keyvault-secrets": "^4.6.0",
```
derisen:
```suggestion
```

derisen: wasn't exactly sure but it looks like we don't use these here

.github/workflows/msal-node-e2e.yml
```diff
@@ -107,6 +108,12 @@ jobs:
         AZURE_CLIENT_SECRET: ${{ secrets.AZURE_CLIENT_SECRET }}
         AZURE_TENANT_ID: ${{ secrets.AZURE_TENANT_ID }}
         AZURE_CLIENT_ID: ${{ secrets.AZURE_CLIENT_ID }}
+        OBO_WEB_APP_CLIENT_ID: ${{ secrets.OBO_WEB_APP_CLIENT_ID }}
+        OBO_WEB_APP_CLIENT_SECRET: ${{ secrets.OBO_WEB_APP_CLIENT_SECRET }}
+        OBO_WEB_API_CLIENT_ID: ${{ secrets.OBO_WEB_API_CLIENT_ID }}
+        OBO_WEB_API_CLIENT_SECRET: ${{ secrets.OBO_WEB_API_CLIENT_SECRET }}
+        OBO_WEB_API_URL: ${{ secrets.OBO_WEB_API_URL }}
```
derisen: > OBO_WEB_API_URL: ${{ secrets.OBO_WEB_API_URL }}

nit: can we call this APP ID URI or just scope?

samples/msal-node-samples/on-behalf-of/config/WEB-API.json
Robbie-Microsoft: I added them because every other e2e test uses them. Additionally, the default values (server port, redirectUri, authority) are the same between the index.js and the .spec.ts file, so instead of defining them twice, we're importing them. I can remove those files and hardcode all the values in the index.js and spec.ts if you still think that would be better.

Robbie-Microsoft: I'm passing the web_api server port and scope to the web app to be used in the OBO request, so these files are necessary.

derisen (approved):
LGTM!

.github/workflows/msal-node-e2e.yml
Robbie-Microsoft: OBO_WEB_APP_ID_URI or APP_ID_URI?

derisen: I've realized yesterday that it's the full scope (app id uri + .default), so something like `WEB_API_SCOPE`  would be best for clarity.

PR #5681, opened 2023-02-13
Reduce `preQueueTimeByCorrelationId` RAM footprint

konstantin-msft:
Changes:
- Reduce `preQueueTimeByCorrelationId` RAM footprint. Maintain only the latest pre-queue event instead of the whole stack.
- Capture number of manually completed queue events.

Notes:
- We don't need to cache timestamps for all pre-queue events. Once new event comes in - we can safely remove the previous one from `preQueueTimeByCorrelationId` map as already processed.

### DIFF
--- a/lib/msal-browser/src/telemetry/BrowserPerformanceClient.ts
+++ b/lib/msal-browser/src/telemetry/BrowserPerformanceClient.ts
@@ -12,7 +12,8 @@ import {
     IPerformanceMeasurement,
     InProgressPerformanceEvent,
     ApplicationTelemetry,
-    SubMeasurement
+    SubMeasurement,
+    PreQueueEvent
 } from "@azure/msal-common";
 import { CryptoOptions } from "../config/Configuration";
 import { BrowserCrypto } from "../crypto/BrowserCrypto";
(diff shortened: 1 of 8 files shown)


lib/msal-browser/src/telemetry/BrowserPerformanceClient.ts
```diff
@@ -80,25 +90,24 @@ export class BrowserPerformanceClient extends PerformanceClient implements IPerf
...
-            preQueueTimesByEvents.set(eventName, window.performance.now());
-            this.preQueueTimeByCorrelationId.set(correlationId, preQueueTimesByEvents);
-        } else {
-            const preQueueTimes = new Map();
-            preQueueTimes.set(eventName, window.performance.now());
-            this.preQueueTimeByCorrelationId.set(correlationId, preQueueTimes);
+        const preQueueEvent: PreQueueEvent | undefined = this.preQueueTimeByCorrelationId.get(correlationId);
+        /**
+         * Manually complete queue measurement if there is an incomplete pre-queue event.
+         * Incomplete pre-queue events are instrumentation bugs that should be fixed.
```
peterzenz, line 120: Would be great if you can author a kusto query to call out where these bugs are occurring.  We can then run that periodically or as part of a dashboard to find these issues.

peterzenz (approved):

lib/msal-browser/src/telemetry/BrowserPerformanceClient.ts
```diff
@@ -48,7 +58,7 @@ export class BrowserPerformanceClient extends PerformanceClient implements IPerf
     startMeasurement(measureName: PerformanceEvents, correlationId?: string): InProgressPerformanceEvent {
         // Capture page visibilityState and then invoke start/end measurement
         const startPageVisibility = this.getPageVisibility();
-        
+
```
tnorling: For future PRs formatting changes should be done in a separate PR

tnorling (approved):

lib/msal-browser/src/telemetry/BrowserPerformanceClient.ts
jo-arroyo: +1 on finding where these bugs are occurring.

jo-arroyo (approved):

sameerag (approved):

lib/msal-browser/src/telemetry/BrowserPerformanceClient.ts
konstantin-msft: Added new `queuedManuallyCompletedCount` telemetry data point to calculate manually completed queue measurements. We won't be able to tell which ones are closed manually by looking at telemetry though, only the total number. To get more details we will have to investigate logs.

konstantin-msft: Will add a Kusto query later on once changes go live.

PR #5688, opened 2023-02-15
E2E ConfigUtils for config param replacement in samples

derisen:
This PR adds an E2E utility class to replace and restore configuration parameters in samples. This allows dynamically setting params during test runs and help us avoid hardcoding clientIDs and etc.


samples/e2eTestUtils/ConfigUtils.ts
```diff
@@ -0,0 +1,71 @@
...
+    constructor(filePath: string) {
+        this.filePath = filePath;
+        this.mappings = {};
+    }
+
+    async replace(map: Record<string, string>): Promise<void> {
+        this.mappings = map;
+
+        return new Promise((resolve, reject) => {
+            fs.readFile(this.filePath, "utf-8", (err, data) => {
```
shoatman: Looks like we could use the sync version of this function instead of the async one.

samples/e2eTestUtils/ConfigUtils.ts
```diff
@@ -0,0 +1,71 @@
...
+                }
+
+                let updatedContent = data;
+
+                for (const [key, value] of Object.entries(map)) {
+                    const regex = new RegExp(key, "g");
+                    updatedContent = updatedContent.replace(regex, value);
+                }
+
+                fs.writeFile(this.filePath, updatedContent, (err) => {
```
shoatman: sync version?

samples/e2eTestUtils/ConfigUtils.ts
```diff
@@ -0,0 +1,71 @@
+import fs from 'fs';
+
+export class StringReplacer {
```
shoatman, line 3: suggest you add some comments describing what this does... I think I know based on description of PR.  Are there tests for this code?

samples/e2eTestUtils/ConfigUtils.ts
derisen: fair! just switched to sync

samples/e2eTestUtils/ConfigUtils.ts
derisen: Comments added. We currently don't run tests for e2eUtils code itself -not quite sure where/when we would run them if we added so cc @tnorling

samples/e2eTestUtils/ConfigUtils.ts
derisen: done

samples/e2eTestUtils/ConfigUtils.ts
tnorling: Given this is itself only used in tests and those tests can't function without this working properly I don't think this needs special coverage. It should be very apparent if there's an issue.

samples/e2eTestUtils/ConfigUtils.ts
```diff
@@ -0,0 +1,55 @@
...
+    constructor(filePath: string) {
+        this.filePath = filePath;
+        this.mappings = {};
+    }
+
+    /**
+     * Updates the file by replacing all instances of keys with values in the map
+     * @param map: key-value pairs of strings to replace
+     */
+    async replace(map: Record<string, string>) {
```
tnorling, line 16: This all feels a bit overkill to me. Did you consider putting the config in a JSON file instead? That would make it much simpler to update values in place or even just replace the whole file.

derisen: What I had in mind was a solution that satisfies these: 

1) Doesn't require us hardcode things: so IDs don't get leaked, people don't try to sign-in to these apps, if things change on the lab side we don't have to manually update configs and etc.
2) Doesn't sacrifice presentation: while samples here are primarily for testing, they are still the first introduction for a lot of people, and we also want refer to them for minimal reproduction of issues. Putting configs in js files like authConfig.js allows us to place comments, import types, constants, enums and etc.
3) Doesn't require change to the existing code: loading configs from json means we'll have to async initialize msal, and/or have to configure framework settings to play nice with json imports and etc.

I'm happy to reconsider though. We're already doing jsons in node samples and a few js ones, so there's precedent. I just think this is more flexible going forward.

tnorling: I think I prefer the static file approach. It feels cleaner and less brittle to me. Something else to keep in mind is that we use these samples locally to test our changes. Hiding the details behind the lab API will make it more cumbersome for us to get up and running. 

1. To address the leaking of details, they're not secret, I think as long as they're not the default and we clearly document they are for internal use only I don't personally have a problem with them being visible. To address the maintenance concern we could perhaps take a hybrid approach where the lab API is used in the E2E environment but the static config is still available for us to use locally. But how often do we think the lab API changes? Is it worth the extra effort/cost?
2. To address json vs js, I don't have a strong opinion. If you want to define the config in js files it just means we replace (move/rename) the whole file with the alternate config rather than replacing individual values. That's fine too.
3. Sounds like using the js files from point 2 would address these concerns as well

derisen: Hmm, alright then. Looks like this is a larger discussion -my bad, I should've tried to align on this beforehand. @peterzenz @sameerag what are your thoughts?

In the meantime, I've made another PR to just replace ppe apps and endpoints with prod ones [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/5720). We can look into merging that and keep this on the side.

sameerag: I would like the "edit-ability" that a static file provides. If the `lab` goes down, a static file can be replaced easily with any config for our local testing and makes debug easy too. That is my only issue with this approach.

derisen:
Closing this as it seems we have some good reasons against it. Thanks everyone for weighing in.

PR #5687, opened 2023-02-15
Add Lab API param appPlatform to e2eTestUtils

derisen:
MSIDLabs API supports the parameter `appPlatform`, which helps with querying apps supporting SPA redirect URI type. This PR updates the E2ETestUtils/LabClient to make use of the parameter.

PR #5689, opened 2023-02-15
remove hardcoded params in msal-angular samples

derisen:
This PR removes credentials in msal-angular samples and replaces ppe endpoints with prod endpoints. 

See also #5687 #5688


samples/msal-angular-v2-samples/angular10-sample-app/src/app/app.module.ts
```diff
@@ -48,8 +46,7 @@ export function MSALInstanceFactory(): IPublicClientApplication {
 
 export function MSALInterceptorConfigFactory(): MsalInterceptorConfiguration {
   const protectedResourceMap = new Map<string, Array<string>>();
-  // protectedResourceMap.set('https://graph.microsoft.com/v1.0/me', ['user.read']); // Prod environment. Uncomment to use.
-  protectedResourceMap.set('https://graph.microsoft-ppe.com/v1.0/me', ['user.read']);
+  protectedResourceMap.set('https://graph.microsoft.com/v1.0/me', ['user.read']);
```
shoatman: I know this was already here, but suggest we follow the casing used in the documentation: https://learn.microsoft.com/en-us/graph/permissions-reference

samples/msal-angular-v2-samples/angular10-sample-app/src/app/app.module.ts
```diff
@@ -25,10 +25,8 @@ export function loggerCallback(logLevel: LogLevel, message: string) {
 export function MSALInstanceFactory(): IPublicClientApplication {
   return new PublicClientApplication({
     auth: {
-      // clientId: '6226576d-37e9-49eb-b201-ec1eeb0029b6', // Prod enviroment. Uncomment to use. 
-      clientId: '3fba556e-5d4a-48e3-8e1a-fd57c12cb82e', // PPE testing environment
-      // authority: 'https://login.microsoftonline.com/common', // Prod environment. Uncomment to use.
-      authority: 'https://login.windows-ppe.net/common', // PPE testing environment.
+      clientId: 'ENTER_CLIENT_ID_HERE',
```
shoatman, line 33: What will the developer see if they don't update this value when they attempt to run the sample?

samples/msal-angular-v2-samples/angular10-sample-app/src/app/app.module.ts
```diff
@@ -25,10 +25,8 @@ export function loggerCallback(logLevel: LogLevel, message: string) {
...
+      authority: 'https://login.microsoftonline.com/common',
```
shoatman: We may want to add a comment here pointing to app registration.  Not sure that common will work for apps that are registered as single tenant.  I don't recall though.

samples/msal-angular-v2-samples/angular10-sample-app/src/app/profile/profile.component.ts
```diff
@@ -26,7 +23,7 @@ export class ProfileComponent implements OnInit {
   }
 
   getProfile() {
-    this.http.get(GRAPH_ENDPOINT)
+    this.http.get('https://graph.microsoft.com/v1.0/me')
```
shoatman, line 26: I'd keep the constant.  Curious: Should all of our samples being using the official MS Graph sdk to make requests to MS Graph: https://github.com/microsoftgraph/msgraph-sdk-javascript

shoatman:
> This PR removes credentials in msal-angular samples and replaces ppe endpoints with prod endpoints.

We don't have credentials in the samples... we have app identifiers.

samples/msal-angular-v2-samples/angular10-sample-app/test/home.spec.ts
```diff
@@ -35,14 +41,25 @@ describe('/ (Home Page)', () => {
...
+            azureEnvironment: AzureEnvironments.CLOUD,
+            appType: AppTypes.CLOUD,
+            userType: UserTypes.CLOUD,
+            appPlatform: AppPlatforms.SPA,
         };
 
         const labClient = new LabClient();
         const envResponse = await labClient.getVarsByCloudEnvironment(labApiParams);
 
         [username, accountPwd] = await setupCredentials(envResponse[0], labClient);
```
shoatman, line 53: There are no credentials being used here since javascript apps are either user agent or public client applications.

samples/msal-angular-v2-samples/angular11-sample-app/package.json
```diff
@@ -7,7 +7,7 @@
     "build": "ng build",
     "test": "ng test",
     "lint": "ng lint",
-    "e2e": "jest",
+    "e2e": "jest --runInBand",
```
shoatman, line 10: Will this parameter be removed when the PR is complete: https://jestjs.io/docs/cli#--runinband

samples/msal-angular-v2-samples/angular10-sample-app/src/app/profile/profile.component.ts
derisen: We use Graph SDK whenever possible in actual sample repos ([e.g.](https://github.com/Azure-Samples/ms-identity-javascript-angular-tutorial/tree/main/2-Authorization-I/1-call-graph)). Since these are samples for testing primarily, I'm inclined to keep these lean, but we can certainly work on this in another PR if decided cc @peterzenz @EmLauber

samples/msal-angular-v2-samples/angular10-sample-app/test/home.spec.ts
derisen: Maybe a naming confusion -these are username and password for the test account used in automation

samples/msal-angular-v2-samples/angular10-sample-app/src/app/app.module.ts
derisen: msal will throw an endpoint_resolution error in this case. Good point though, I'll add a comment above that points to the registration instructions.

samples/msal-angular-v2-samples/angular10-sample-app/src/app/app.module.ts
derisen: fair point -just added

samples/msal-angular-v2-samples/angular10-sample-app/src/app/app.module.ts
derisen: done!

samples/msal-angular-v2-samples/angular10-sample-app/src/app/profile/profile.component.ts
derisen: coming back to the constant, it doesn't really have a use here -ideally we would centralize these and import from one location, but due to the angular module load ordering it would run into circular dependency unless we add a separate config file. I'm just shy of doing this in this PR due to its size.

samples/msal-angular-v2-samples/angular10-sample-app/src/app/app.module.ts
derisen: done!

samples/msal-angular-v2-samples/angular11-sample-app/package.json
derisen: this is to run tests files for a given sample sequentially. We've been using this for some samples when it has multiple test files that share the same config file, but have different configuration parameters depending on the scenario. In that case running them in parallel would lead to conflicts.

derisen:
> > This PR removes credentials in msal-angular samples and replaces ppe endpoints with prod endpoints.
> 
> We don't have credentials in the samples... we have app identifiers.

Well true, client ID is not top secret. In principle it can still be used in a phishing attack, but more relevantly here, folks just end up using (or trying to use) the samples without registering a new app, which then leads to issues and/or creates confusion.

samples/msal-angular-v2-samples/angular10-sample-app/src/app/profile/profile.component.ts
peterzenz: This seems OK to me then.  Thanks.

peterzenz (approved):

samples/msal-angular-v2-samples/angular10-sample-app/test/home.spec.ts
sameerag: Let us rename them later in a separate PR - `setupTestAccCredentials()`?

samples/msal-angular-v2-samples/angular10-sample-app/src/app/app.module.ts
sameerag: Can we do a custom error in the sample to say their authority is not setup when `authority` matches `ENTER_CLIENT_ID_HERE`?  May be for a later PR, but `endpoint_resolution_error` is confusing and this is special case of folks running the sample as is. Thoughts??

sameerag (approved):
Added a few comments which can be addressed in a separate PR.

samples/msal-angular-v2-samples/angular10-sample-app/test/home.spec.ts
derisen: Sounds good, I made a note of that

samples/msal-angular-v2-samples/angular10-sample-app/src/app/app.module.ts
derisen: Sorry I missed this! Yes I think we could add a simple check before msal instantiation -I'm noting this down, best to tackle in another PR

sameerag: Works, I have already approved, so this should not block you.

PR #5692, opened 2023-02-16
Export internal classes

lalimasharda:
- export native classes as internals
- breakdown native telemetry method


lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -232,24 +232,7 @@ export class NativeInteractionClient extends BaseInteractionClient {
...
-            matsApiErrorCode: mats ? mats.api_error_code : undefined,
-            matsUiVisible: mats ? mats.ui_visible : undefined,
-            matsSilentCode: mats ? mats.silent_code : undefined,
-            matsSilentBiSubCode: mats ? mats.silent_bi_sub_code : undefined,
-            matsSilentMessage: mats ? mats.silent_message : undefined,
-            matsSilentStatus: mats ? mats.silent_status : undefined,
-            matsHttpStatus: mats ? mats.http_status : undefined,
-            matsHttpEventCount: mats ? mats.http_event_count : undefined
-        }, this.correlationId);
+        const mats = this.addNativeBrokerFieldsToTelemetry(response);
```
sameerag: Can we rename this as: `this.addTelemetryFromNativeResponse()`?

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -374,6 +357,30 @@ export class NativeInteractionClient extends BaseInteractionClient {
         return result;
     }
 
+    protected addNativeBrokerFieldsToTelemetry(response: NativeResponse): MATS | null {
```
sameerag: Same as above - Can we please rename this as: this.addTelemetryFromNativeResponse()?

sameerag (approved):
Lgtm, added a comment on naming.

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
lalimasharda: Fixed

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
lalimasharda: Done

change/@azure-msal-browser-1d161695-5cf6-4f1b-83cd-5dfaa0de0457.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
```
tnorling:
```suggestion
  "type": "patch",
```

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -374,6 +357,30 @@ export class NativeInteractionClient extends BaseInteractionClient {
         return result;
     }
 
+    protected addTelemetryFromNativeResponse(response: NativeResponse): MATS | null {
+
+        const mats = this.getMATSFromResponse(response);
+        this.performanceClient.addStaticFields({
+            extensionId: this.nativeMessageHandler.getExtensionId(),
+            extensionVersion: this.nativeMessageHandler.getExtensionVersion(),
+            matsBrokerVersion: mats ? mats.broker_version : undefined,
```
tnorling: I know this code was simply moved from somewhere else but can we do an early return if `mats` is undefined instead of doing `mats ? mats.param : undefined` everywhere?

lalimasharda: Do we just want to return null or do we want to return performanceClient with static fields that don't need mats?

sameerag: What did we decide here?

tnorling (approved):

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
lalimasharda: @sameerag  we are returning null if mats is null

PR #5695, opened 2023-02-17
fixing get account entity from cache using native account id

lalimasharda:
Fixed:
- readAccountFromCacheWithNativeAccountId() not returning correct AccountEntity object.
- Unit test for readAccountFromCacheWithNativeAccountId() to call the correct method.


tnorling (approved):
Thanks!

sameerag (approved):

PR #5697, opened 2023-02-17
remove hardcoded params in msal-react samples

derisen:
This PR removes hardcoded IDs in msal-react samples and replaces ppe endpoints with prod endpoints. 

See also #5687 #5688

### DIFF
--- a/samples/msal-react-samples/gatsby-sample/jest.config.js
+++ b/samples/msal-react-samples/gatsby-sample/jest.config.js
@@ -2,7 +2,7 @@ module.exports = {
     displayName: "Gatsby",
     globals: {
         __PORT__: 3001,
-        __STARTCMD__: "npm run develop -- -p 3001"
+        __STARTCMD__: "NODE_ENV=development npm run develop -- -p 3001"
     },
     preset: "../../e2eTestUtils/jest-puppeteer-utils/jest-preset.js"
 };
--- a/samples/msal-react-samples/gatsby-sample/jest.config.js
+++ b/samples/msal-react-samples/gatsby-sample/jest.config.js
@@ -2,7 +2,7 @@ module.exports = {
     displayName: "Gatsby",
     globals: {
         __PORT__: 3001,
-        __STARTCMD__: "npm run serve -- -p 3001"
+        __STARTCMD__: "npm run develop -- -p 3001"
     },
     preset: "../../e2eTestUtils/jest-puppeteer-utils/jest-preset.js"
 };
--- a/samples/msal-react-samples/react-router-sample/test/home.spec.ts
+++ b/samples/msal-react-samples/react-router-sample/test/home.spec.ts
@@ -1,11 +1,15 @@
+import path from "path";
 import puppeteer from "puppeteer";
-import {Screenshot, setupCredentials, enterCredentials, RETRY_TIMES} from "../../../e2eTestUtils/TestUtils";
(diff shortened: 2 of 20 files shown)


samples/msal-react-samples/react-router-sample/test/home.spec.ts
```diff
@@ -35,14 +41,26 @@ describe('/ (Home Page)', () => {
...
+            appType: AppTypes.CLOUD,
+            userType: UserTypes.CLOUD,
+            appPlatform: AppPlatforms.SPA,
         };
 
         const labClient = new LabClient();
         const envResponse = await labClient.getVarsByCloudEnvironment(labApiParams);
 
         [username, accountPwd] = await setupCredentials(envResponse[0], labClient);
+        [clientID, , authority] = await retrieveAppConfiguration(envResponse[0], labClient, false);
```
peterzenz, line 54: I commented on this in your other PR, but will mention it here too.  Suggest something like `[clientID, unused, authority]` for readability.

peterzenz (approved):

samples/msal-react-samples/gatsby-sample/jest.config.js
```diff
@@ -2,7 +2,7 @@ module.exports = {
     displayName: "Gatsby",
     globals: {
         __PORT__: 3001,
-        __STARTCMD__: "npm run serve -- -p 3001"
+        __STARTCMD__: "NODE_ENV=development npm run develop -- -p 3001"
```
tnorling, line 5: Why the change here from serving the production version to development?

samples/msal-react-samples/typescript-sample/package.json
```diff
@@ -23,7 +23,7 @@
         "start": "react-scripts start",
         "build": "react-scripts build",
         "build:package": "cd ../../../lib/msal-react && npm run build:all",
-        "test": "jest",
+        "test": "jest --runInBand",
```
tnorling, line 26: Why the change to runInBand? Won't this slow down our tests?

derisen: You're right, it would (maybe not much at the moment, since there are 2 test files in each sample). I needed this now that I'm setting the configuration parameters at test runtime, so in principle different tests can have different parameters and running them in parallel would lead to conflict. The inefficiency here is that both files use the same parameters in these tests -I'm looking into doing this as an [environment setup step with jest](https://jestjs.io/docs/configuration#setupfilesafterenv-array), so that we can get rid of `runInBand` afterwards

samples/msal-react-samples/gatsby-sample/jest.config.js
derisen: the build step is happening before the test files run to replace the configuration parameters. So if we run serve here we won't be not running the current state of the project files. I didn't realize it matters though (isn't this what we do with other react samples [e.g.](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/f3f54cbc2faf30894852a98585e22c5869277858/samples/msal-react-samples/typescript-sample/jest.config.js#L5)?)

samples/msal-react-samples/typescript-sample/package.json
tnorling: Alternative approach: If a different config is needed we can do it on a per-route basis in the app itself rather than swapping it out in the test. Not only does that allow us to run the tests in parallel but also allows us to easily test the scenario manually or our customers to see a feature demonstrated

Anyway, I think since both files are using the same config right now (and I don't see that changing anytime soon) we can add this flag if and when it becomes needed.

PR #5702, opened 2023-02-20
remove hardcoded params in msal-browser samples

derisen:
This PR removes hardcoded IDs in msal-browser samples and replaces ppe endpoints with prod endpoints. 

See also #5687 #5688


samples/msal-browser-samples/VanillaJSTestApp2.0/app/client-capabilities/test/browser.spec.ts
```diff
@@ -27,6 +36,16 @@ describe("Browser tests", function () {
         const labClient = new LabClient();
         const envResponse = await labClient.getVarsByCloudEnvironment(labApiParams);
         [username, accountPwd] = await setupCredentials(envResponse[0], labClient);
+        [clientID, , authority] = await retrieveAppConfiguration(envResponse[0], labClient, false);
```
peterzenz, line 39: Rather than a `[clientId, , authority]`  suggest you rewrite this to something more like `[clientID, unused, authority]` to improve readability.  You have this pattern in several places in this PR.

peterzenz (approved):

samples/msal-browser-samples/VanillaJSTestApp2.0/app/client-capabilities/test/browser.spec.ts
derisen: ah, fair! -will replace them all

samples/msal-browser-samples/VanillaJSTestApp2.0/app/customizable-e2e-test/test/browserAAD.spec.ts
```diff
@@ -25,49 +24,72 @@ async function verifyTokenStore(BrowserCache: BrowserCacheUtils, scopes: string[
...
+            appType: AppTypes.CLOUD,
+            userType: UserTypes.CLOUD,
+            appPlatform: AppPlatforms.SPA,
         };
 
         const labClient = new LabClient();
         const envResponse = await labClient.getVarsByCloudEnvironment(labApiParams);
 
         [username, accountPwd] = await setupCredentials(envResponse[0], labClient);
+        [clientID, , authority] = await retrieveAppConfiguration(envResponse[0], labClient, false);
```
sameerag, line 55: Wasn't the whole idea of a separate json file is to create the flexibility to change the config as needed in unit tests? Why is this code refactored now? cc @hectormmg  who made the initial changes to review this.

sameerag (changes requested):
I want to check why we are getting rid of the .json files for config. They served a purpose.

samples/msal-browser-samples/VanillaJSTestApp2.0/app/client-capabilities/test/browser.spec.ts
derisen: @peterzenz well it turned out to be a little more tricky -since we are doing initalization/reassignment here, I would have to declare `unused` somewhere above (which also feels weird). Instead, I'd like to refactor `retrieveAppConfiguration` to return an object instead of array (so order doesn't matter), and match the lab api [schema](https://msidlab.com/swagger/v1/swagger.json) a bit more.

samples/msal-browser-samples/VanillaJSTestApp2.0/app/customizable-e2e-test/test/browserAAD.spec.ts
derisen: Well I think the util `retrieveAppConfiguration` was added more recently (perhaps it wasn't quite needed until cca testing), but now that we have it, its safer to not depend on hardcoded ids and just retrieve as much as possible from the lab at runtime (there's still a json file, its just being populated at runtime)

PR #5703, opened 2023-02-21
remove hardcoded params in msal-node samples

derisen:
This PR removes hardcoded IDs in msal-node samples and replaces ppe endpoints with prod endpoints. 

See also #5688


peterzenz (approved):

samples/msal-node-samples/ElectronTestApp/test/electron-code-adfs.spec.ts
```diff
@@ -47,6 +47,33 @@ test.describe('Electron Auth Code ADFS 2019 PPE Tests ', () => {
         const envResponse = await labClient.getVarsByCloudEnvironment(labApiParams);
 
         [username, accountPwd] = await setupCredentials(envResponse[0], labClient);
+        [clientID, , authority] = await retrieveAppConfiguration(envResponse[0], labClient, false);
+
+        config = {
```
tnorling, line 52: This seems to be repeated in each test file. Any way we can write a helper to build this?

tnorling (approved):
One comment about code reuse but otherwise looks good

samples/msal-node-samples/ElectronTestApp/test/electron-code-adfs.spec.ts
derisen: good idea! I'm intending to refactor `retrieveAppConfiguration` to return more from lab in another PR, could do this as part of that.

PR #5720, opened 2023-02-24
Remove ppe endpoint in samples

derisen:
Replaces ppe apps and endpoints with prod ones in samples

### DIFF
--- a/samples/msal-angular-v2-samples/angular10-sample-app/src/app/app.module.ts
+++ b/samples/msal-angular-v2-samples/angular10-sample-app/src/app/app.module.ts
@@ -26,9 +26,9 @@ export function MSALInstanceFactory(): IPublicClientApplication {
   return new PublicClientApplication({
     auth: {
       clientId: "b5c2e510-4a17-4feb-b219-e55aa5b74144",
-      // clientId: '3fba556e-5d4a-48e3-8e1a-fd57c12cb82e', // PPE testing environment. Uncomment to use.
+      // clientId: '3fba556e-5d4a-48e3-8e1a-fd57c12cb82e', // PPE testing environment
       authority: "https://login.microsoftonline.com/common",
-      // authority: 'https://login.windows-ppe.net/common', // PPE testing environment. Uncomment to use.
+      // authority: 'https://login.windows-ppe.net/common', // PPE testing environment
       redirectUri: '/',
       postLogoutRedirectUri: '/'
     },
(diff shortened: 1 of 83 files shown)


samples/msal-angular-v2-samples/angular10-sample-app/src/app/app.module.ts
```diff
@@ -48,8 +48,8 @@ export function MSALInstanceFactory(): IPublicClientApplication {
 
 export function MSALInterceptorConfigFactory(): MsalInterceptorConfiguration {
   const protectedResourceMap = new Map<string, Array<string>>();
-  // protectedResourceMap.set('https://graph.microsoft.com/v1.0/me', ['user.read']); // Prod environment. Uncomment to use.
-  protectedResourceMap.set('https://graph.microsoft-ppe.com/v1.0/me', ['user.read']);
+  protectedResourceMap.set('https://graph.microsoft.com/v1.0/me', ['user.read']);
+  // protectedResourceMap.set('https://graph.microsoft-ppe.com/v1.0/me', ['user.read']); // PPE testing environment. Uncomment to use.
```
peterzenz: It's more than uncomment here no?  There are a couple things they'd have to change to use PPE again.  Maybe remove the comments or just note that PPE endpoints are more for reference or something?

samples/msal-node-samples/device-code/test/device-code-adfs.spec.ts
```diff
@@ -30,7 +30,7 @@ const cachePlugin = require("../../cachePlugin.js")(TEST_CACHE_LOCATION);
 // Load scenario configuration
 const config = require("../config/ADFS.json");
 
-describe('Device Code ADFS PPE Tests', () => {
+describe('Device Code ADFS 2019 Tests', () => {
```
peterzenz, line 33: Was it 2019?  Do we need to call it out explicitly here rather than ADFS Prod Tests?

peterzenz (approved):

samples/msal-node-samples/device-code/test/device-code-adfs.spec.ts
derisen: [We only support ADFS 2019](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-common/docs/ADFS.md) (latest) as the earlier versions are not oidc complaint. Do you think we should still not call it out?

samples/msal-angular-v2-samples/angular10-sample-app/src/app/app.module.ts
derisen: True, they would have to uncomment all commented out lines to make it work. Let me just remove the "uncomment to use" instructions then -whoever wants to use ppe would probably know what they are doing.

samples/msal-node-samples/device-code/test/device-code-adfs.spec.ts
peterzenz: It's good with it. I'm just surprised the change wasn't ADFS PPE -> ADFS Prod

derisen: makes sense, I think "ADFS PPE" was misleading to begin with -that distinction doesn't really apply to ADFS server

hectormmg (approved):

PR #5722, opened 2023-02-24
Fixed bug in msal-node's httpClient

### LINKED ISSUE #4879: Getting ClientAuthError: endpoints_resolution_error: Error: could not resolve endpoints with @azure/msal-node 1.9.1

Robbie-Microsoft:
We've been receiving reports of users receiving the error:

`Error: TypeError
Error Description: The "listener" argument must be of type function. Received an instance of Object - []`

I wasn't able to reproduce the issue from any user reports, however, a fellow Microsoft employee who works on the VSCode team received the same error in a VSCode extension she was building. She walked me through the build steps for her VSCode extension and reproduction steps, and then I was finally able to reproduce the issue.

As it relates to msal-node's HttpClient: I'm not sure why this error was being thrown, because the nodejs code where the error occurs adheres to the documentation on NodeJS's website. Specifically, I'm talking about [https.request](https://nodejs.org/docs/latest-v16.x/api/http.html#httprequesturl-options-callback).

The relevant code in msal-node's code base is [located here](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/src/network/HttpClient.ts#L249). According to the docs, we can make a https request this way:
 ```https.request(url, options)```.

However, I needed to change that linked code to the following to resolve the error:
```
const customOptions: https.RequestOptions = {
        hostname: url.hostname, // <-- new code
        path: url.pathname, // <-- new code
        method: httpMethod,
        headers: options?.headers || emptyHeaders,
    };
...
https.request(customOptions)
```

Update: I have been looking into this for a while, and the only conclusion I can draw is that maybe the NodeJS docs are incorrect, or this is a NodeJS bug.

### DIFF
--- a/lib/msal-node/src/network/HttpClient.ts
+++ b/lib/msal-node/src/network/HttpClient.ts
@@ -8,6 +8,7 @@ import { HttpMethod, Constants, HttpStatus, ProxyStatus } from "../utils/Constan
 import { NetworkUtils } from "../utils/NetworkUtils";
 import http from "http";
 import https from "https";
+import { urlToHttpOptions } from "node:url";
 
 /**
  * This class implements the API for network requests.
@@ -224,16 +225,12 @@ const networkRequestViaHttps = <T>(
     const body: string = options?.body || "";
 
     const url = new URL(urlString);
-    const emptyHeaders: Record<string, string> = {};
+    const headers = options?.headers || {} as Record<string, string>;
     const customOptions: https.RequestOptions = {
-        hostname: url.hostname,
-        path: url.pathname,
         method: httpMethod,
-        headers: options?.headers || emptyHeaders,
+        headers: headers,
+        ...urlToHttpOptions(url),
     };
-    if (url.search) {
-        customOptions.path += url.search;
-    }
 
     if (timeout) {
         customOptions.timeout = timeout;
--- a/lib/msal-node/src/network/HttpClient.ts
+++ b/lib/msal-node/src/network/HttpClient.ts
@@ -59,18 +59,18 @@ export class HttpClient implements INetworkModule {
 }
 
 const networkRequestViaProxy = <T>(
-    url: string,
+    destinationUrlString: string,
     proxyUrlString: string,
     httpMethod: string,
     options?: NetworkRequestOptions,
     agentOptions?: http.AgentOptions,
     timeout?: number,
 ): Promise<NetworkResponse<T>> => {
-    const headers = options?.headers || {} as Record<string, string>;
+    const destinationUrl = new URL(destinationUrlString);
     const proxyUrl = new URL(proxyUrlString);
-    const destinationUrl = new URL(url);
 
     // "method: connect" must be used to establish a connection to the proxy
+    const headers = options?.headers || {} as Record<string, string>;
     const tunnelRequestOptions: https.RequestOptions = {
         host: proxyUrl.hostname,
         port: proxyUrl.port,
@@ -78,6 +78,9 @@ const networkRequestViaProxy = <T>(
         path: destinationUrl.hostname,
         headers: headers,
     };
+    if (destinationUrl.searchParams) {
+        tunnelRequestOptions.path += `?${destinationUrl.searchParams}`;
+    }
 
     if (timeout) {
         tunnelRequestOptions.timeout = timeout;
@@ -214,7 +217,7 @@ const networkRequestViaProxy = <T>(
 };
 
 const networkRequestViaHttps = <T>(
-    url: string,
+    urlString: string,
     httpMethod: string,
     options?: NetworkRequestOptions,
     agentOptions?: https.AgentOptions,
@@ -223,11 +226,17 @@ const networkRequestViaHttps = <T>(
     const isPostRequest = httpMethod === HttpMethod.POST;
     const body: string = options?.body || "";
 
+    const url = new URL(urlString);
     const emptyHeaders: Record<string, string> = {};
     const customOptions: https.RequestOptions = {
+        hostname: url.hostname,
+        path: url.pathname,
         method: httpMethod,
         headers: options?.headers || emptyHeaders,
     };
+    if (url.searchParams) {
+        customOptions.path += `?${url.searchParams}`;
+    }
 
     if (timeout) {
         customOptions.timeout = timeout;
@@ -246,7 +255,7 @@ const networkRequestViaHttps = <T>(
     }
 
     return new Promise<NetworkResponse<T>>((resolve, reject) => {
-        const request = https.request(url, customOptions);
+        const request = https.request(customOptions);
 
         if (timeout) {
             request.on("timeout", () => {
--- a/lib/msal-node/src/network/HttpClient.ts
+++ b/lib/msal-node/src/network/HttpClient.ts
@@ -8,7 +8,6 @@ import { HttpMethod, Constants, HttpStatus, ProxyStatus } from "../utils/Constan
 import { NetworkUtils } from "../utils/NetworkUtils";
 import http from "http";
 import https from "https";
-import { urlToHttpOptions } from "node:url";
 
 /**
  * This class implements the API for network requests.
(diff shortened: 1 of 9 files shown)


samples/msal-node-samples/custom-INetworkModule-and-network-tracing/HttpClientCurrent.ts
```diff
@@ -264,11 +267,17 @@ const networkRequestViaHttps = <T>(
     const isPostRequest = httpMethod === HttpMethod.POST;
     const body: string = options?.body || "";
 
+    const url = new URL(urlString);
     const emptyHeaders: Record<string, string> = {};
     const customOptions: https.RequestOptions = {
+        hostname: url.hostname,
```
bgavrilMS: We should not be writing this type of code, you are bound to get it wrong. There seems to be a library function for this `https://nodejs.org/api/url.html#urlurltohttpoptionsurl`

(for example, you are missing the port)

samples/msal-node-samples/custom-INetworkModule-and-network-tracing/HttpClientCurrent.ts
```diff
@@ -287,7 +296,7 @@ const networkRequestViaHttps = <T>(
     }
 
     return new Promise<NetworkResponse<T>>((resolve, reject) => {
-        const request = https.request(url, customOptions);
+        const request = https.request(customOptions);
```
bgavrilMS, line 317: Can you please create a minimal repro and open a bug on NodeJS repo and see what they say ? It looks to me that the issue is with them. You can then reference the bug in you "workaround" code. 

https://github.com/nodejs/node/blob/v13.14.0/lib/_http_client.js#L86

samples/msal-node-samples/custom-INetworkModule-and-network-tracing/HttpClientCurrent.ts
```diff
@@ -255,7 +258,7 @@ const networkRequestViaProxy = <T>(
 };
 
 const networkRequestViaHttps = <T>(
```
bgavrilMS, line 282: Would this bug happen if we were using an HttpClient based on Axios?

samples/msal-node-samples/custom-INetworkModule-and-network-tracing/HttpClientCurrent.ts
```diff
@@ -287,7 +296,7 @@ const networkRequestViaHttps = <T>(
     }
 
     return new Promise<NetworkResponse<T>>((resolve, reject) => {
-        const request = https.request(url, customOptions);
```
bgavrilMS, line 290: Does the bug still happen if you use `https.request(url: Url, customOptions: Options)` instead of `https.request(url: string, customOptions: Options)` ?

Robbie-Microsoft: Yep

bgavrilMS (changes requested):
See comments.

samples/msal-node-samples/custom-INetworkModule-and-network-tracing/HttpClientCurrent.ts
Robbie-Microsoft: What do you mean by "We should not be writing this type of code"? What "type of code" are you referring to? What am I bound to get wrong? I'm using the URL as it's defined on the MDN web docs [here](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL). I'm following the example RequestOptions as defined in the NodeJS docs [here](https://nodejs.org/api/https.html#httpsrequesturl-options-callback), and it does appear that I forgot the port.

It looks like:
```
const options = {
  hostname: 'encrypted.google.com',
  port: 443,
  path: '/',
  method: 'GET',
};
```

Robbie-Microsoft: Actually, because this is an htpps request, the port is automatically assumed to be 443.

samples/msal-node-samples/custom-INetworkModule-and-network-tracing/HttpClientCurrent.ts
sameerag: Yes, this issue is definitely on Node.js. I tested it the other day and had the same issue.

samples/msal-node-samples/custom-INetworkModule-and-network-tracing/HttpClientCurrent.ts
Robbie-Microsoft: I can confirm that this bug does not occur in the old implementation of HttpClient (axios).

Robbie-Microsoft: I'll be making a new PR adding back the old implementation for cases when proxyUrl isn't used.

samples/msal-node-samples/custom-INetworkModule-and-network-tracing/HttpClientCurrent.ts
Robbie-Microsoft: @bgavrilMS I'm unable to create a minimal repro that reproduces the bug.

sameerag (approved):
Approved. As discussed, lets work on Axios based client as an option in a separate PR.

bgavrilMS (approved):

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -223,10 +224,12 @@ const networkRequestViaHttps = <T>(
...
     const body: string = options?.body || "";
 
-    const emptyHeaders: Record<string, string> = {};
+    const url = new URL(urlString);
+    const headers = options?.headers || {} as Record<string, string>;
     const customOptions: https.RequestOptions = {
         method: httpMethod,
-        headers: options?.headers || emptyHeaders,
+        headers: headers,
+        ...urlToHttpOptions(url),
```
bgavrilMS: if you can't use this method `urlToHttpOptions` from the library, copy it's source code (it's on GH, see in Teams where I provided a link) and also the unit test associated.

PR #5727, opened 2023-02-27
Require application invoke initialize() function

konstantin-msft:
- Make `PublicClientApplication` constructor protected.
- Add `PublicClientApplication.getInstance()` as a constructor replacement to invoke `PublicClientApplication.initialize()` seamlessly.
- Remove `PublicClientApplication.initialized` as redundant.

### DIFF
--- a/lib/msal-browser/src/app/PublicClientApplication.ts
+++ b/lib/msal-browser/src/app/PublicClientApplication.ts
@@ -29,7 +29,21 @@ export class PublicClientApplication extends ClientApplication implements IPubli
 
     /**
      * @constructor
-     * Constructor for the PublicClientApplication used to instantiate the PublicClientApplication object
+     * Constructor for the PublicClientApplication used to instantiate the PublicClientApplication object. To be used
+     * internally within the class or any class that extends it but not externally
+     *
+     * @param configuration object for the MSAL PublicClientApplication instance
+     */
+    protected constructor(configuration: Configuration) {
+        super(configuration);
+
+        this.activeSilentTokenRequests = new Map();
+        // Register listener functions
+        this.trackPageVisibility = this.trackPageVisibility.bind(this);
+    }
+
+    /**
+     * Static method to instantiate the PublicClientApplication object
      *
      * Important attributes in the Configuration object for auth are:
      * - clientID: the application ID of your application. You can obtain one by registering your application with our Application registration portal : https://portal.azure.com/#blade/Microsoft_AAD_IAM/ActiveDirectoryMenuBlade/RegisteredAppsPreview
@@ -46,14 +60,12 @@ export class PublicClientApplication extends ClientApplication implements IPubli
      * In Azure B2C, authority is of the form https://{instance}/tfp/{tenant}/{policyName}/
      * Full B2C functionality will be available in this library in future versions.
      *
-     * @param configuration object for the MSAL PublicClientApplication instance
+     * @param configuration {Configuration}
      */
-    constructor(configuration: Configuration) {
-        super(configuration);
-
-        this.activeSilentTokenRequests = new Map();
-        // Register listener functions
-        this.trackPageVisibility = this.trackPageVisibility.bind(this);
+    public static async getInstance(configuration: Configuration): Promise<PublicClientApplication> {
+        const app: PublicClientApplication = new PublicClientApplication(configuration);
+        await app.initialize();
+        return app;
     }
 
     /**
--- a/lib/msal-browser/src/app/PublicClientApplication.ts
+++ b/lib/msal-browser/src/app/PublicClientApplication.ts
@@ -29,21 +29,7 @@ export class PublicClientApplication extends ClientApplication implements IPubli
 
     /**
      * @constructor
-     * Constructor for the PublicClientApplication used to instantiate the PublicClientApplication object. To be used
-     * internally within the class or any class that extends it but not externally
-     *
-     * @param configuration object for the MSAL PublicClientApplication instance
-     */
-    protected constructor(configuration: Configuration) {
-        super(configuration);
-
-        this.activeSilentTokenRequests = new Map();
-        // Register listener functions
-        this.trackPageVisibility = this.trackPageVisibility.bind(this);
-    }
-
-    /**
-     * Static method to instantiate the PublicClientApplication object
+     * Constructor for the PublicClientApplication used to instantiate the PublicClientApplication object
      *
      * Important attributes in the Configuration object for auth are:
      * - clientID: the application ID of your application. You can obtain one by registering your application with our Application registration portal : https://portal.azure.com/#blade/Microsoft_AAD_IAM/ActiveDirectoryMenuBlade/RegisteredAppsPreview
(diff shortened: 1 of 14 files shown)


lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -178,7 +170,6 @@ export abstract class ClientApplication {
                 this.logger.verbose(e);
             }
         }
-        this.initialized = true;
```
peterzenz: Maybe we should think about this in a different way.  We can leave the `protected initialized: boolean;` in place on line 84 above and keep the `this.initialized = true` here when the initialize call is made.  We can then check if initilized != true and fail library calls.  I'm not sure what the behavior is if someone fails to call initialize and calls our other APIs.  We might be able to clarify it today but as we go further along, we might find that some APIs that were once working start to fail in weird ways in cases where Initialize was not called.  Thoughts?

tnorling: That approach is what I initially had in mind as well but I actually don't mind the approach in this PR, the developer experience feels a bit cleaner/more intuitive to me. I do think it's a good idea to throw if initialize hasn't been called though regardless of which approach we take.

konstantin-msft: I am fine with keeping `initliazed` variable though i think it would be redundant since we enforce to instantiate an application with a static method that always invokes `initliaze()` method. Thus, it's either everything or nothing,

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -29,7 +29,21 @@ export class PublicClientApplication extends ClientApplication implements IPubli
 
     /**
      * @constructor
-     * Constructor for the PublicClientApplication used to instantiate the PublicClientApplication object
+     * Constructor for the PublicClientApplication used to instantiate the PublicClientApplication object. To be used
+     * internally within the class or any class that extends it but not externally
+     *
+     * @param configuration object for the MSAL PublicClientApplication instance
+     */
+    protected constructor(configuration: Configuration) {
```
shoatman: breaking change?

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -46,14 +60,12 @@ export class PublicClientApplication extends ClientApplication implements IPubli
...
-     * @param configuration object for the MSAL PublicClientApplication instance
+     * @param configuration {Configuration}
      */
-    constructor(configuration: Configuration) {
-        super(configuration);
-
-        this.activeSilentTokenRequests = new Map();
-        // Register listener functions
-        this.trackPageVisibility = this.trackPageVisibility.bind(this);
+    public static async getInstance(configuration: Configuration): Promise<PublicClientApplication> {
```
shoatman: getInstance is commonly used with getting an instance of a singleton.  This doesn't appear to be a singleton though.  I have a similar method in my PR... you may want to take a look.

shoatman (changes requested):
I don't think we want any breaking changes in v3....

lib/msal-browser/src/app/PublicClientApplication.ts
konstantin-msft: Yes, it's a breaking change. Why do we want to avoid breaking changes in major releases?

lib/msal-browser/src/app/PublicClientApplication.ts
konstantin-msft: I am totally fine renaming it to `createPublicClientApplication()`

lib/msal-browser/src/app/ClientApplication.ts
tnorling: Unfortunately `protected` is a TypeScript construct and will be removed at transpile time so we can't prevent people from calling the PCA constructor directly. We'll of course say that's unsupported but throwing an error here is still a good idea IMO

lib/msal-browser/src/app/PublicClientApplication.ts
hectormmg: @konstantin-msft we decided not to ship any API surface changes for v3 to make migration easier. Only breaking changes we're shipping are behavior changes.

konstantin-msft: I wasn't aware of that. Thanks @hectormmg .

peterzenz: I think this one needs to be in for the JS/WAM work.  It's the only API breaking change I'm willing to take and throwing if we're not initialized (suggestion I had above) should help folks detect that this change is needed when they run their build tests vs when they are in prod.

hectormmg: @peterzenz agreed, it's actually not something we can circumvent with the JS/WAM as default change. It will also mostly be a single-place update for most developers, so the surface change delta is minimal anyway.

tnorling: We can keep the API surface the same but also start requiring people to call initialize (by throwing if it hasn't been called). I do like the changes proposed in this PR but if the desire is to make migration as easy as possible we can table the static `getInstance` pattern until v4. 

I don't really feel strongly about surface changes in v3 so either approach is good for me, just wanted to throw that option out there for those who do

konstantin-msft: From my perspective requiring people to call `initialize()` is still a breaking change. I would really prefer to use a static method. It seems like an easy migration to me because it's a one-line code change for customers.

tnorling: Agreed it's still breaking. The difference is: people who are calling initialize today don't have to make any changes at all.

Because the static method turns PCA construction from sync to async it could be much more than a one-line change for people depending on how their apps are designed today. Don't get me wrong, I like the approach but the more I think about it the more I think it may be better suited for v4 given the desire to keep v3 light on changes.

konstantin-msft: Well, in that case I can leave `createPublicClientApplication()` static method as an alternative approach to get initialized app instance.

tnorling: Sure that's a good idea, I'm ok with that.

konstantin-msft: Wrapping up:
1. Public client `initialize()` remains public. We mark it as mandatory in v3.
2. New method `createPublicClientApplication()` is added as an alternative to get initialized app client.
3. Condition checks to validate that app is initialized are already in place, so nothing needs to be changed here.

@tnorling @peterzenz @hectormmg please let me know if you are OK with that.

tnorling: I'm ok with that, however, minor clarification regarding point 3. I believe the checks are only done today when allowBrokering is true. We'll want to make sure this is always checked regardless of config.

konstantin-msft:
Abandoning this PR as a duplicate of https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/5557/files.

lib/msal-browser/src/app/PublicClientApplication.ts
sameerag: Late to the party but adding the comments. Can we not "initialize" if "allow brokering" is true? I do not like the idea of two ways to initialize the PCA. I would rather leverage what @shoatman did in his PR and have this as an "operating context" based on "allow brokering" set up.

PR #5729, opened 2023-02-27
Remove dropped version checks from v3 workflows

hectormmg:
This PR:

- Removes e2e tests for Angular v9-12 since they won't be supported in msal-angular v3
- Removes e2e tests for Node v10-12 since they won't be supported in msal-node v2
- Adds `--bundleConfigAsCjs` flag to msal-browser's watch script


.github/workflows/msal-angular-e2e.yml
```diff
@@ -33,10 +33,6 @@ jobs:
       fail-fast: false
       matrix:
         sample:
-          - 'angular9-v2-sample-app'
```
peterzenz: @jo-arroyo do we need to create new/specific tests for the more recent angular releases?

peterzenz: @hectormmg are there files we want to remove as well?  Or is that in a later PR?

peterzenz (approved):

tnorling (approved):

.github/workflows/msal-angular-e2e.yml
jo-arroyo: I think tests for the more recent angular releases can be done when we do msal-angular v3

hectormmg: @hectormmg we could remove samples for unsupported versions and keep them in the v2 branch once we ship v3. Having said that, when we eventually deprecate v2 the samples will be permanently deleted. Does that sound okay?

sameerag (approved):

sameerag (approved):

PR #5733, opened 2023-02-28
Ignore tests, samples in CodeQL Scan

tnorling:
Adds CodeQL config file and ignores tests and samples from scan

### DIFF
--- a/.github/codeql-config.yml
+++ b/.github/codeql-config.yml
@@ -1,0 +1,5 @@
+name: "CodeQL config"
+
+paths-ignore:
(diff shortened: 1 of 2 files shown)


peterzenz (approved):
Thank you!

konstantin-msft (approved):

jo-arroyo (approved):
Note that the msal-angular spec files are not grouped in a `test` folder, so if we continue seeing CodeQL scan errors involving msal-angular, we may need to change that/add a path here.

PR #5734, opened 2023-03-01
Hardcoded values as default metadata source

hectormmg:
This PR:

- Changes authority metadata source precedence to be Hardcoded values > Configuration > Network
- Updates test suite to reflect behavior change


lib/msal-common/src/authority/Authority.ts
```diff
@@ -306,7 +306,28 @@ export class Authority {
      */
     private async updateEndpointMetadata(metadataEntity: AuthorityMetadataEntity): Promise<AuthorityMetadataSource> {
         this.performanceClient?.addQueueMeasurement(PerformanceEvents.AuthorityUpdateEndpointMetadata, this.correlationId);
+        /*
+         * First attempt to get metadata from hardcoded values
+         * If there is no hardcoded metadata for under the configured canonical authority, fall back to configured metadata
+         * If hardcoded and configured metadata are not present, acquire metadata from network
+         */
 
+        // Get metadata from hardcoded values
```
sameerag, line 315: Can we have switch code here may be for readability?

Based on the config, we can define:
```
const metaDataFetchPolicy = 


switch(metaDataFetchPolicy):
    case hardCodedMetaData: { ... return if present}; // else the switch block will continue
    case configMetaData: { ...return in present};
    .... 
}
```

What do you think?

sameerag (approved):

lib/msal-common/src/authority/Authority.ts
```diff
@@ -306,7 +306,28 @@ export class Authority {
      */
     private async updateEndpointMetadata(metadataEntity: AuthorityMetadataEntity): Promise<AuthorityMetadataSource> {
         this.performanceClient?.addQueueMeasurement(PerformanceEvents.AuthorityUpdateEndpointMetadata, this.correlationId);
+        /*
+         * First attempt to get metadata from hardcoded values
```
tnorling, line 310: Can you tell me a little more about this change? It doesn't seem right to me to ignore configuration options...

ghost:

PR #5750, opened 2023-03-02
Bump typescript to `4.9.5` and build target to `es2020`

konstantin-msft:
- Bump `typescript` to 4.9.5.
- Bump target/module versions to `es2020` when applicable.
- Replace calls to deprecated `mocked()` from `ts-jest/utils` with `jest.spyOn()`.
- Fix types' coercions to facilitate compatibility with the new ts version.
- Make generic type extend `ServerAuthorizationTokenResponse` for `NetworkManager.sendPostRequest()` in `msal-common` to handle post processing properly.
- Make `serverTelemetryManager.cacheFailedRequest()` accept any types of errors.


lib/msal-common/src/account/AuthToken.ts
```diff
@@ -44,7 +44,7 @@ export class AuthToken {
             const base64Decoded = crypto.base64Decode(base64TokenPayload);
             return JSON.parse(base64Decoded) as TokenClaims;
         } catch (err) {
-            throw ClientAuthError.createTokenParsingError(err);
+            throw ClientAuthError.createTokenParsingError(err as string);
```
konstantin-msft, line 47: Type coercions are prohibited in ts v4.

lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -162,8 +162,8 @@ export abstract class ClientApplication {
         } catch (e) {
             if (e instanceof AuthError) {
                 e.setCorrelationId(validRequest.correlationId);
+                serverTelemetryManager.cacheFailedRequest(e);
```
konstantin-msft: `serverTelemetryManager.cacheFailedRequest(e)` expects an instance of `AuthError`. We can't pass an arbitrary error.

lib/msal-node/test/client/ClientAssertion.spec.ts
```diff
@@ -1,11 +1,11 @@
 import { ClientAssertion } from "../../src/client/ClientAssertion";
 import { TEST_CONSTANTS } from "../utils/TestConstants";
 import { CryptoProvider } from "../../src/crypto/CryptoProvider";
-import { sign } from "jsonwebtoken";
-import { mocked } from "ts-jest/utils";
```
konstantin-msft, line 5: Deprecated

lib/msal-common/src/network/NetworkManager.ts
```diff
@@ -31,7 +32,7 @@ export class NetworkManager {
      * @param tokenEndpoint
      * @param options
      */
-    async sendPostRequest<T>(thumbprint: RequestThumbprint, tokenEndpoint: string, options: NetworkRequestOptions): Promise<NetworkResponse<T>> {
+    async sendPostRequest<T extends ServerAuthorizationTokenResponse>(thumbprint: RequestThumbprint, tokenEndpoint: string, options: NetworkRequestOptions): Promise<NetworkResponse<T>> {
```
konstantin-msft, line 35: `T` must extend `ServerAuthorizationTokenResponse` because `ThrottlingUtils.postProcess()` expects it as an input.

lib/msal-react/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "include": ["src", "types", "test"],
   "compilerOptions": {
     "module": "esnext",
+    "target": "esnext",
```
konstantin-msft: Required for proper class transpiling.

change/msal-1a69cf50-db5e-4aaf-ab86-32f98e28bec4.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Bump typescript to 4.9.5 and target/module to es2020 #5750",
+  "packageName": "msal",
```
tnorling, line 4: Shouldn't be making any changes to msal-core

change/@azure-msal-node-extensions-8fbfdb4b-83ee-4c5f-a036-594a7aeb4f06.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "Bump typescript version to 4.9.5 #5750",
+  "packageName": "@azure/msal-node-extensions",
```
tnorling, line 4: Changefiles for browser, common and node?

change/msal-1a69cf50-db5e-4aaf-ab86-32f98e28bec4.json
konstantin-msft: May I ask you why not?

change/@azure-msal-node-extensions-8fbfdb4b-83ee-4c5f-a036-594a7aeb4f06.json
konstantin-msft: Beachball root scope for v3 is `["lib/*", "extensions/*"]`. That's why there are no dedicated files for the packages you listed above.
@hectormmg

change/msal-1a69cf50-db5e-4aaf-ab86-32f98e28bec4.json
tnorling: That's MSAL.js v1 which is no longer receiving updates (and this would need to be a major change there)

change/@azure-msal-node-extensions-8fbfdb4b-83ee-4c5f-a036-594a7aeb4f06.json
tnorling: More likely this is happening because the v3 branch already has changefiles and beachball is configured to use dev as the root, not the v3 branch. We should likely update the beachball config in the v3 branch to use the v3 branch as root.

change/msal-1a69cf50-db5e-4aaf-ab86-32f98e28bec4.json
sameerag: If we are supporting security issues (which I recall this is), we may have to push it to V1 too. What is the cost in terms of breaking change?

tnorling: Let's discuss the security aspect offline. The cost is that we already have an MSAL.js v2, we shouldn't add confusion and the maintenance cost of supporting multiple older versions unless absolutely necessary.

konstantin-msft: Reverted `msal-core` changes. Thanks for bringing the context Thomas, that's very helpful.

lib/msal-browser/src/interaction_client/SilentRefreshClient.ts
```diff
@@ -41,8 +41,8 @@ export class SilentRefreshClient extends StandardInteractionClient {
             .catch((e: AuthError) => {
                 if (e instanceof AuthError) {
                     (e as AuthError).setCorrelationId(this.correlationId);
+                    serverTelemetryManager.cacheFailedRequest(e);
```
sameerag: Was there a possibility `e` is not of the type `AuthError`? If NO, we can get rid of the `if` and if `YES`, we still need to cache it outside of the condition right?

konstantin-msft: Good catch, thank you. `if (e instanceof AuthError)` is redundant here. Fixed.

change/@azure-msal-node-extensions-8fbfdb4b-83ee-4c5f-a036-594a7aeb4f06.json
konstantin-msft: Beachball doesn't require changefiles for browser, common, etc even after updating the root branch https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/5780.

lib/msal-browser/src/interaction_client/PopupClient.ts
```diff
@@ -190,9 +191,9 @@ export class PopupClient extends StandardInteractionClient {
 
             if (e instanceof AuthError) {
                 (e as AuthError).setCorrelationId(this.correlationId);
+                serverTelemetryManager.cacheFailedRequest(e);
```
peterzenz: This looks like a functional change and not a change related to the TS/ES intent of this PR. You mention type coercion changes are not supported in the new TS and I'm seeing this pattern throughout this change.  What happened in the original code if we called `.cacheFailedRequest(e)` and it was not of type `AuthError`?  Did the failure cascade out from `cacheFailedRequest` or was it able to coerce whatever error type that was thrown into something usable?  If it's not of type `AuthError` we might still want to cache a failed response (failure is failure...).  I just don't know what other types we might get in this flow.

lib/msal-browser/src/utils/BrowserUtils.ts
```diff
@@ -73,9 +73,10 @@ export class BrowserUtils {
     }
 
     /**
-     * Returns best compatible network client object. 
+     * Returns best compatible network client object.
      */
     static getBrowserNetworkClient(): INetworkModule {
+        // @ts-ignore TS2774
```
peterzenz, line 79: Maybe a comment on why we're adding the ts-ignore here?  Is there a future fix we're holding off on just to unblock folks?

peterzenz: As I look at the other ts-ignores around window I'm thrown back to some earlier discussions on web workers not supporting window references.  @sameerag or @tnorling I thought there were some alternates here that might be more friendly.  I'll admit though I'm not up to speed on the new TS/ES changes that might impact here.

lib/msal-browser/test/interaction_client/RedirectClient.spec.ts
```diff
@@ -6,7 +6,33 @@
 import sinon from "sinon";
 import { PublicClientApplication } from "../../src/app/PublicClientApplication";
 import { TEST_CONFIG, TEST_URIS, TEST_HASHES, TEST_TOKENS, TEST_DATA_CLIENT_INFO, TEST_TOKEN_LIFETIMES, RANDOM_TEST_GUID, DEFAULT_OPENID_CONFIG_RESPONSE, testNavUrl, TEST_STATE_VALUES, DEFAULT_TENANT_DISCOVERY_RESPONSE, testLogoutUrl, TEST_SSH_VALUES } from "../utils/StringConstants";
-import { ServerError, Constants, AccountInfo, TokenClaims, AuthenticationResult, CommonAuthorizationCodeRequest, CommonAuthorizationUrlRequest, AuthToken, PersistentCacheKeys, AuthorizationCodeClient, ResponseMode, ProtocolUtils, AuthenticationScheme, Logger, ServerTelemetryEntity, LogLevel, NetworkResponse, ServerAuthorizationTokenResponse, CcsCredential, CcsCredentialType, CommonEndSessionRequest, ServerTelemetryManager, AccountEntity, ClientConfigurationError } from "@azure/msal-common";
+import {
```
peterzenz, line 9: Not as terse, but so much more readable.  :)

peterzenz (approved):
I have the questions I mentioned in teams about the explicit type changes for TS4 and how those resolve.  I'd also like some clarity on the ts-ignores as to what it would take to address those correctly.  If we want this in though now to unblock I'm OK, but please track bugs/work items around these things if they are unresolved.

lib/msal-browser/src/interaction_client/PopupClient.ts
sameerag: Same comment, similar pattern below.

lib/msal-browser/test/interaction_client/RedirectClient.spec.ts
sameerag: At some point I had all files like this. Can we make this a standard?

lib/msal-browser/src/utils/BrowserUtils.ts
sameerag: Reading [this](https://github.com/microsoft/TypeScript/issues/36048), I get that we are checking if `window.fetch` and `window.headers` are defined, what is the alternative here?  

May be a `try .. catch` where if `window.fetch` isn't defined, we cannot call the `fetchClient` anyway. I am not sure if that actually throws though but worth a try.

lib/msal-browser/src/interaction_client/PopupClient.ts
konstantin-msft: That's a very good point. I have made `serverTelemetryManager.cacheFailedRequest()` to accept any types of errors to make sure we cache all of them.

lib/msal-browser/src/utils/BrowserUtils.ts
konstantin-msft: Unfortunately, `try .. catch` won't help here.
image

change/msal-1a69cf50-db5e-4aaf-ab86-32f98e28bec4.json
jo-arroyo: If changes reverted, shouldn't this file be deleted?

extensions/msal-node-extensions/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "include": ["src", "test"],
   "compilerOptions": {
     "module": "esnext",
+    "target": "esnext",
```
jo-arroyo: The others are being set to ES2020. Is there a reason why we're setting this to esnext?

peterzenz (approved):

lib/msal-node/src/client/ClientApplication.ts
jo-arroyo: If `serverTelemetryManager.cacheFailedRequest()` is able to accept any types of error, do these need to be reverted? (and below)

extensions/msal-node-extensions/tsconfig.json
konstantin-msft: Good catch, thank you. Fixed.

lib/msal-node/src/client/ClientApplication.ts
konstantin-msft: I have already reverted it but forgot to resolve the comment. Sorry about that.

jo-arroyo: But `cacheFailedRequest` is still being called inside the `if (e instanceof AuthError)` block

change/msal-1a69cf50-db5e-4aaf-ab86-32f98e28bec4.json
konstantin-msft: We still need this. It's a common file for changes in lib/** modules.

lib/msal-node/src/client/ClientApplication.ts
konstantin-msft: Completely forgot about `msal-node` module. Good catch, thank you. Fixed.

jo-arroyo (approved):
Thanks for addressing comments!

change/@azure-msal-node-extensions-8fbfdb4b-83ee-4c5f-a036-594a7aeb4f06.json
tnorling: That's because the [command](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/112abff69ebb68d4c8a88e86af6d056cf650e492/package.json#L63) specifically references the branch

change/msal-1a69cf50-db5e-4aaf-ab86-32f98e28bec4.json
tnorling: Not sure I understand this... msal == msal-core. Changefile should only be needed if you're making msal-core changes, which we shouldn't be.

PR #5751, opened 2023-03-03
Fix SHR typ header to pop value

hectormmg:
This PR replaces the default value of the typ header in SHRs from "JWT" to "pop" to match the SHR specification.


sameerag (approved):
Thanks for the quick turnaround @hectormmg

hectormmg:
> Thanks for the quick turnaround @hectormmg

Small change, no problem!

peterzenz (approved):

PR #5753, opened 2023-03-03
Upgrade msal-angular to Angular 15

jo-arroyo:
This PR upgrades the msal-angular library to use Angular 15.
- Upgrades `typescript` and `zone.js` dependencies
- Updates to ES2020
- Changes peer dependencie `rxjs` to `^7.0.0`
- Removes deprecated Angular features
- Adds `initialize()` to MSAL Guard for WAM-Broker flow
- Removes and reorders imports for legibility 
- Updates unit tests
- Removes angular 13-14 samples from e2e matrix

This also upgrades dependencies in the MSAL.js repo root.

Angular 15 sample app and E2E tests will be in subsequent PRs.


lib/msal-angular/src/msal.guard.spec.ts
```diff
@@ -450,7 +450,7 @@ describe('MsalGuard', () => {
       username: "test"
     }]);
 
-    guard.canLoad()
+    guard.canMatch()
```
konstantin-msft, line 453: May i ask you why are we replacing `canLoad()` with `canMatch()`?

lib/msal-angular/src/msal.guard.ts
```diff
@@ -129,8 +129,11 @@ export class MsalGuard implements CanActivate, CanActivateChild, CanLoad {
         // Capture current path before it gets changed by handleRedirectObservable
         const currentPath = this.location.path(true);
 
-        return this.authService.handleRedirectObservable()
+        return this.authService.initialize()
```
konstantin-msft, line 132: Same question here. Why are we replacing `handleRedirectObservable()` with `initialize()`?

lib/msal-angular/src/msal.guard.spec.ts
jo-arroyo: `canLoad` has been deprecated and replaced with `canMatch`. https://angular.io/api/router/CanLoad

lib/msal-angular/src/msal.guard.ts
jo-arroyo: We're not replacing, we're just calling `initialize()` before `handleRedirectObservable()` (see it being called below). This is because `allowNativeBroker` is set to true by default in v3, and we need to account for [this error](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/errors.md#native_broker_called_before_initialize). This is already being done in the MsalRedirectComponent, but needed to be updated here.

lib/msal-angular/package.json
```diff
@@ -58,16 +58,15 @@
...
-    "ng-packagr": "^11.2.4",
-    "protractor": "~7.0.0",
-    "rxjs": "~6.5.5",
+    "ng-packagr": "^15.1.2",
+    "rxjs": "^7.0.0",
     "ts-node": "~8.3.0",
     "tslib": "^2.0.0",
-    "typescript": "~4.1.5",
-    "zone.js": "~0.10.3"
+    "typescript": "^4.8.4",
```
tnorling: This should probably use ~ version

lib/msal-angular/src/msal.broadcast.service.spec.ts
```diff
@@ -291,9 +290,9 @@ describe('MsalBroadcastService', () => {
     });
 
     // @ts-ignore
-    msalInstance.eventHandler.emitEvent(EventType.HANDLE_REDIRECT_START, InteractionType.Redirect);
+    msalInstance.controller.getEventHandler().emitEvent(EventType.HANDLE_REDIRECT_START, InteractionType.Redirect);
     // @ts-ignore
```
tnorling, line 294: remind me what the ts-ignore was here for?

lib/msal-angular/src/packageMetadata.ts
```diff
@@ -1,3 +1,3 @@
 /* eslint-disable header/header */
 export const name = "@azure/msal-angular";
-export const version = "2.5.3";
+export const version = "3.0.0-alpha";
```
tnorling: Should this be alpha.0? And do we need to update the package.json version to match, or will that be done at release time?

tnorling (approved):
A couple non-blocking questions. Looks good!

lib/msal-angular/src/msal.guard.spec.ts
konstantin-msft: Makes sense, thank you!

lib/msal-angular/src/msal.guard.ts
konstantin-msft: Gotcha, thanks!

konstantin-msft (approved):

lib/msal-angular/src/msal.broadcast.service.spec.ts
jo-arroyo: It is because `controller` is a protected property on `PublicClientApplication`, but it seems to work here for testing purposes.

tnorling: Got it thanks! It works because protected only exists in TypeScript-world, which you're ignoring

lib/msal-angular/package.json
jo-arroyo: Good catch, thanks!

lib/msal-angular/src/packageMetadata.ts
jo-arroyo: Yes it should be alpha.0. Thanks! 
Not 100% sure about the package.json. Will check with the team and update that too if needed.

blogcraft:
Is this upgrade compatible with upcoming Angular 16?

jo-arroyo:
@blogcraft MSAL Angular v3 will be tested against Angular 16 for compatibility and the README will be updated once compatibility is confirmed.

PR #5766, opened 2023-03-07
web api (on-behalf-of) sample with distributed cache

derisen:
Adds a sample demonstrating the distributed token caching pattern for on-behalf-of flow. Includes collecting basic performance metrics.


samples/msal-node-samples/on-behalf-of-distributed-cache/README.md
```diff
@@ -0,0 +1,88 @@
...
+        - For **User consent display name** type `Access msal-node-webapi`.
+        - For **User consent description** type `Allow the application to access msal-node-webapi on your behalf.`
+        - Keep **State** as **Enabled**.
+        - Select the **Add scope** button on the bottom to save this scope.
+1. On the left hand side menu, select the `Manifest` blade.
+   - Set `accessTokenAcceptedVersion` property to **2**.
+   - Modify the `"knownClientApplications": []` array to contain the client ID/app ID of the client application(S) that will call this web API (for example `knownClientApplications": ["APP_ID_OF_THE_CLIENT_APP"]`)
+   - Click on **Save**.
+
+Before running the sample, you will need to replace the values in the configuration object (see [index.ts](./src/index.ts)):
```
salman90:
```suggestion
Before running the sample, you will need to replace the values in the configuration object (see [app.ts](./src/app.ts)):
```

samples/msal-node-samples/on-behalf-of-distributed-cache/src/app.ts
```diff
@@ -0,0 +1,125 @@
...
+
+        if (tokenSource) {
+            const results = {
+                tokenSource,
+                durationTotalInMs,
+                durationInCacheInMs,
+                durationInHttpInMs: tokenSource === "network" ? durationTotalInMs - durationInCacheInMs : 0,
+            };
+
+            console.log(results);
```
salman90, line 101:
```suggestion
            
```

salman90 (approved):

samples/msal-node-samples/on-behalf-of-distributed-cache/src/app.ts
derisen: will keep this one as well

samples/msal-node-samples/on-behalf-of-distributed-cache/src/AuthProvider.ts
```diff
@@ -0,0 +1,154 @@
...
+            },
+            system: {
+                loggerOptions: {
+                    loggerCallback(loglevel, message, containsPii) {
+                        console.log(message);
+                    },
+                    logLevel: LogLevel.Trace,
+                    piiLoggingEnabled: false
+                },
+                // proxyUrl: "http://localhost:8888" // uncomment to capture traffic with Fiddler
```
bgavrilMS, line 55: +1

samples/msal-node-samples/on-behalf-of-distributed-cache/src/AuthProvider.ts
```diff
@@ -0,0 +1,154 @@
...
+                clientId: appConfig.clientId,
+                authority: `${appConfig.instance}/${appConfig.tenantId}`,
+                clientSecret: appConfig.clientSecret,
+            },
+            system: {
+                loggerOptions: {
+                    loggerCallback(loglevel, message, containsPii) {
+                        console.log(message);
+                    },
+                    logLevel: LogLevel.Trace,
```
bgavrilMS, line 52: Info by default?

samples/msal-node-samples/on-behalf-of-distributed-cache/src/AuthProvider.ts
```diff
@@ -0,0 +1,154 @@
...
+    private static async getMetadata(msalConfig: Configuration, cacheClient: RedisClientType): Promise<Configuration> {
+        const msalConfigWithMetadata = msalConfig;
+
+        const clientId = msalConfigWithMetadata.auth.clientId;
+        const tenantId = msalConfigWithMetadata.auth.authority!.split("/").pop()!;
+
+        try {
+            let [cloudDiscoveryMetadata, authorityMetadata] = await Promise.all([
+                cacheClient.get(`${clientId}.${tenantId}.discovery-metadata`),
+                cacheClient.get(`${clientId}.${tenantId}.authority-metadata`)
```
bgavrilMS, line 101: Interesting, so you're caching the metadata in Redis. 

1. I would add a potential improvement if possible - set the TTL for these entries, I believe the AAD endpoints suggest how long to cache this for (24h?). 
2. the key should be `{authority}.discovery-metadata` ? It's not tied to clientid, but it is tied to the cloud. `{tenand_id}.discoevry_metadata` will probably work as well due to GUIDs being unique and all.

samples/msal-node-samples/on-behalf-of-distributed-cache/src/AxiosHelper.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+import axios from "axios";
+
+class AxiosHelper {
+
+    /**
+     * Makes an Authorization "Bearer" request with the given accessToken to the given endpoint.
+     * @param endpoint
+     * @param accessToken
```
bgavrilMS: params?: Record<string, string> is not documented and it's unclear what it a does.

samples/msal-node-samples/on-behalf-of-distributed-cache/src/AxiosHelper.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import axios from "axios";
+
+class AxiosHelper {
+
+    /**
+     * Makes an Authorization "Bearer" request with the given accessToken to the given endpoint.
```
bgavrilMS:
```suggestion
     * Makes an HTTP GET to the endpoint uri. If an access token exists, it adds an Authroization header to the request, where it adds the Bearer token.
```

samples/msal-node-samples/on-behalf-of-distributed-cache/README.md
```diff
@@ -0,0 +1,90 @@
...
+        - Select the **Add scope** button on the bottom to save this scope.
+1. On the left hand side menu, select the `Manifest` blade.
+   - Set `accessTokenAcceptedVersion` property to **2**.
+   - Modify the `"knownClientApplications": []` array to contain the client ID/app ID of the client application(S) that will call this web API (for example `knownClientApplications": ["APP_ID_OF_THE_CLIENT_APP"]`)
+   - Click on **Save**.
+
+Before running the sample, you will need to replace the values in the configuration object (see [app.ts](./src/app.ts)):
+
+```typescript
+const appConfig: AppConfig = {
```
bgavrilMS, line 58: Consider asking the app developers to also configure Redis.

samples/msal-node-samples/on-behalf-of-distributed-cache/src/AuthProvider.ts
```diff
@@ -0,0 +1,154 @@
...
+            performance.mark("acquireTokenOnBehalfOf-start");
+            await cca.getTokenCache().getAllAccounts(); // required for cache read
+            tokenResponse = await cca.acquireTokenOnBehalfOf(tokenRequest);
+            performance.mark("acquireTokenOnBehalfOf-end");
+            performance.measure(
+                tokenResponse?.fromCache ? "acquireTokenOnBehalfOf-fromCache" : "acquireTokenOnBehalfOf-fromNetwork",
+                "acquireTokenOnBehalfOf-start",
+                "acquireTokenOnBehalfOf-end"
+            );
+        } catch (error) {
```
bgavrilMS: catch / throw is not needed.

samples/msal-node-samples/on-behalf-of-distributed-cache/src/CustomCachePlugin.ts
```diff
@@ -0,0 +1,42 @@
...
+    public async beforeCacheAccess(cacheContext: TokenCacheContext): Promise<void> {
+        performance.mark("beforeCacheAccess-start");
+        const cacheData = await this.client.get(this.partitionKey);
+        cacheContext.tokenCache.deserialize(cacheData);
+        performance.mark("beforeCacheAccess-end");
+        performance.measure("beforeCacheAccess", "beforeCacheAccess-start", "beforeCacheAccess-end");
+    }
+
+    public async afterCacheAccess(cacheContext: TokenCacheContext): Promise<void> {
+        if (cacheContext.cacheHasChanged) {
```
bgavrilMS, line 33: Is there any information about the tokens within? You can set the TTL in Redis to `access_token.expiration` here.

samples/msal-node-samples/on-behalf-of-distributed-cache/src/RedisClientWrapper.ts
```diff
@@ -0,0 +1,69 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { ICacheClient } from "@azure/msal-node";
+import { RedisClientType } from "redis";
+
+const CACHE_TTL = 60 * 60 * 24; // 24 hours
```
bgavrilMS: We should ask developers to set this to whatever token lifetime they get from the client app. Ideally MSAL would help with this, but we don't have that implemented :(

bgavrilMS (approved):

samples/msal-node-samples/on-behalf-of-distributed-cache/.gitignore
```diff
@@ -0,0 +1 @@
+build/
```
Robbie-Microsoft, line 1: node_modules folder?

samples/msal-node-samples/on-behalf-of-distributed-cache/src/AuthProvider.ts
```diff
@@ -0,0 +1,154 @@
...
+                    AuthProvider.fetchOIDCMetadata(tenantId)
+                ]);
+
+                if (cloudDiscoveryMetadata && authorityMetadata) {
+                    await cacheClient.set(`${clientId}.${tenantId}.discovery-metadata`, JSON.stringify(cloudDiscoveryMetadata));
+                    await cacheClient.set(`${clientId}.${tenantId}.authority-metadata`, JSON.stringify(authorityMetadata));
+                }
+            }
+
+            msalConfigWithMetadata.auth.cloudDiscoveryMetadata = typeof cloudDiscoveryMetadata === 'string' ? cloudDiscoveryMetadata : JSON.stringify(cloudDiscoveryMetadata);
```
Robbie-Microsoft: super nit: why did you switch to using single quotes? You used double quotes earlier in the file.

samples/msal-node-samples/on-behalf-of-distributed-cache/src/AuthProvider.ts
derisen: Since this is a caching focused sample, I think Trace would be better leave as default -we would have to ask customers to switch to it when asking them to provide MSAL logs for debugging later on

samples/msal-node-samples/on-behalf-of-distributed-cache/src/AxiosHelper.ts
derisen: added a comment for the argument -it's an Axios request parameter object, which apparently has the type `any` in Axios. Not sure why this is so though, in practice it will be always be a JS object (hence key-value pairs) as far as I can see

samples/msal-node-samples/on-behalf-of-distributed-cache/README.md
derisen: done!

samples/msal-node-samples/on-behalf-of-distributed-cache/src/AuthProvider.ts
derisen: My bad -it's not doing anything here: errors thrown here will be caught in the call site already (app.ts)

samples/msal-node-samples/on-behalf-of-distributed-cache/src/AuthProvider.ts
derisen: 1. sure thing, I think we can do this
2. I see. if tenantId alone is ok, would be fine to leave it as it is i.e. clientId.tenantId (asking for consisentency, otherwise I have to go back and update the other samples :D )

samples/msal-node-samples/on-behalf-of-distributed-cache/src/RedisClientWrapper.ts
derisen: The access token lifetime (at least today) is varied though, no? This is a general default setting, but I think we can set the TTL to what comes from the token request response (as you mentioned in your comment above), and fallback to this as default just in case

samples/msal-node-samples/on-behalf-of-distributed-cache/src/CustomCachePlugin.ts
derisen: looking into it

samples/msal-node-samples/on-behalf-of-distributed-cache/src/AuthProvider.ts
derisen: my bad -just fixed

samples/msal-node-samples/on-behalf-of-distributed-cache/.gitignore
derisen: done!

samples/msal-node-samples/on-behalf-of-distributed-cache/src/CustomCachePlugin.ts
derisen: @bgavrilMS coming back to this. We are caching the entire cache blob for this user, so the cache contains id, access and refresh tokens. And potentially there could be a number of access tokens for different targets/resources. The TTL here however is for the entire cache blob -so I'm not sure how we could be accurate given that we have different auth artifacts. Any thoughts?

derisen: @bgavrilMS friendly ping

bgavrilMS: In MSAL.NET we suggest the TTL as being:

- max (expiry (AT)) if only access tokens are in teh cache
- no suggestion if at least one RT is in the cache

bgavrilMS: But anyway, I think this is difficult without MSAL help.

derisen: Ah, ok. All of our user flows (if not mistaken) includes a refresh token (due to default offline_access scope). Leaving this out for now then -we can revisit later.

samples/msal-node-samples/on-behalf-of-distributed-cache/src/AuthProvider.ts
bgavrilMS: For #2 yes, it's fine. Redis will not be shared cross-cloud anyway.

samples/msal-node-samples/on-behalf-of-distributed-cache/src/AuthProvider.ts
bgavrilMS: The concern here is that ppl just copy the sample and go with it. Trace level logging can have a perf impact.

samples/msal-node-samples/on-behalf-of-distributed-cache/src/CustomCachePlugin.ts
bgavrilMS: JFYI,  client_credentials does not use RT. 
OBO in it's current form brings back an RT, but does not use it. Ideally it should not even serialize it. But I agree with you, we can revisit this.

derisen: Gotcha! -I made a note for client credentials, we can update the sample with the max (expiry (AT)) logic for TTL that you suggested.

PR #5770, opened 2023-03-07
Update app initialization doc

konstantin-msft:


lib/msal-browser/docs/initialization.md
```diff
@@ -15,7 +15,12 @@ In this document:
 
 ## Initializing the PublicClientApplication object
 
-In order to use MSAL.js, you need to instantiate a `PublicClientApplication` object. You must provide the `client id` (`appId`) of your application. 
+In order to use MSAL.js, you need to instantiate a `PublicClientApplication` object. You must provide the `client id` (`appId`) of your application.
+
+### Option 1
+
+Instantiate an application object and initialize it afterwards.
```
tnorling:
```suggestion
Instantiate a `PublicClientApplication` object and initialize it afterwards. The `initialize` function is asynchronous and must resolve before invoking other MSAL.js APIs.
```

lib/msal-browser/docs/initialization.md
```diff
@@ -26,6 +31,23 @@ const msalConfig = {
 };
 
 const msalInstance = new msal.PublicClientApplication(msalConfig);
+await msalInstance.initialize();
+```
+
+### Option 2
+
+Invoke a static method to get initialized application object.
```
tnorling:
```suggestion
Invoke the `createPublicClientApplication` static method which returns an initialized `PublicClientApplication` object. Note that this function is asynchronous.
```

tnorling (approved):
A couple suggestions to add clarity. Would also be good to start a migration doc where we can enumerate all the changes between v2 and v3.

lib/msal-browser/docs/initialization.md
konstantin-msft: Updated

lib/msal-browser/docs/initialization.md
konstantin-msft: Updated

konstantin-msft:
> A couple suggestions to add clarity. Would also be good to start a migration doc where we can enumerate all the changes between v2 and v3.

Good idea. I will add a migration doc in a separate PR. Thanks for review!

derisen (approved):

PR #5784, opened 2023-03-10
Update caching and performance docs

### LINKED ISSUE #5748: Cache usage and documentation is confusing
### Core Library

MSAL Node (@azure/msal-node)

### Wrapper Library

Not Applicable

Confidential

### Documentation Location

https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/docs/caching.md

### Description

Over the past few days I've been trying to integrate MSAL into our application. Our system allows for users to add microsoft accounts to read some basic information off graph api. We use SQL database to persist tokens. We also integrate with other providers. After struggling for some time with MSAL I gave up on using the library and simply hit oauth endpoints directly. I was disappointed that the library wasn't more straightforward to use.

A lot of my confusion stemmed from the cache usage. In working with authentication platforms before I handle token refresh on my own. However MSAL looks to handle this for you.

The documentation on MSAL gives the option to persist using ICachePlugin, but doesn't explain the CacheContext object. The linked documentation about cache serialization sends me to .NET examples. I had to look outside microsoft documentation on stackoverflow to determine I need to call ` cacheContext.tokenCache.deserialize` and ` cacheContext.tokenCache.serialize` to write and read to MSAL memory cache respectively. However I still am unsure what is contained in tokenCache itself.

The documentation further suggestions `serialize one cache blob per user` and recommends using DistributedCachePlugin. But does not fully explain why (the reason is perhaps implied, but not explicit). Again I found my answer outside this repo that it was due to each user being added to in memory cache context which quickly gets bloated. Something that does not feel particularly relevant for a serverless application.

Even after a lot of reading and research I'm not sure I understand the cache correctly and I don't want to commit the time to determining if my assumptions are correct. That's why I gave up on the library.

Because MSAL forces the user to use it's cache strategy and abstracts away token refreshing I feel like the cache is one of the most important aspects of the library, but the documentation is very short. I would really appreciate more detail in the cache documentation, not just on what needs to be done, but why and how.

Alternatively I would desire a library that comes without the cache altogether and allows us to create our own cache and refresh strategy.

derisen:
* Updates the caching doc for more clarity and links to new distributed cache samples
* Adds a performance doc to provide suggestions for improvements, monitoring etc.


lib/msal-node/docs/caching.md
```diff
@@ -38,25 +38,25 @@ cca.acquireTokenSilent(silentTokenRequest).then((response) => {
...
 
-## Cache in memory
+## In-memory cache
 
-If you choose not to persist the cache, the `TokenCache` interface is still available to access the tokens cached in memory. The lifetime of in-memory cache is the same as MSAL instance. If the MSAL instance restarts, the cache is erased when the process lifecycle finishes. We recommend persisting the cache with encryption for all real life applications both for security and desired cache longevity.
+By default, MSAL has an in-memory cache. The in-memory cache is the single source of truth for all MSAL activities involving tokens and accounts. The lifetime of in-memory cache is the same as the MSAL instance. If the MSAL instance restarts, the cache is erased when the process lifecycle finishes. We recommend persisting the cache with encryption for all production applications both for security and desired cache longevity. If you choose not to persist the cache, the [TokenCache](https://azuread.github.io/microsoft-authentication-library-for-js/ref/classes/_azure_msal_node.tokencache.html) interface is still available to access the cached entities.
```
bgavrilMS:
```suggestion
MSAL maintains an in-memory cache. The in-memory cache is the single source of truth for all MSAL activities involving tokens and accounts. The lifetime of in-memory cache is the same as the MSAL application object. If the process using MSAL restarts, the cache is erased when the process lifecycle finishes. This affects user scenarios as users will have to re-authenticate, and although the browser will do most some of the work, the end user experience suffers. Service to service scenarios suffer because getting a token from AAD involves HTTP requests, and is much slower than getting a token from a cache.

 We recommend persisting the cache with encryption for all production applications both for security and desired cache longevity. If you choose not to persist the cache, the [TokenCache](https://azuread.github.io/microsoft-authentication-library-for-js/ref/classes/_azure_msal_node.tokencache.html) interface is still available to access the cached entities.
```

bgavrilMS (approved):

lib/msal-node/docs/caching.md
```diff
@@ -38,25 +38,27 @@ cca.acquireTokenSilent(silentTokenRequest).then((response) => {
...
 
-## Cache in memory
+## In-memory cache
 
-If you choose not to persist the cache, the `TokenCache` interface is still available to access the tokens cached in memory. The lifetime of in-memory cache is the same as MSAL instance. If the MSAL instance restarts, the cache is erased when the process lifecycle finishes. We recommend persisting the cache with encryption for all real life applications both for security and desired cache longevity.
+MSAL maintains an in-memory cache. The in-memory cache is the single source of truth for all MSAL activities involving tokens and accounts. The lifetime of in-memory cache is the same as the MSAL application object. If the process using MSAL restarts, the cache is erased when the process lifecycle finishes. This affects user scenarios as users will have to re-authenticate, and although the browser will do most some of the work, the end user experience suffers. Service to service scenarios also suffer because getting a token from AAD involves HTTP requests, and is much slower than getting a token from a cache.
```
sameerag: Is it really "single source of truth"? It is representative of the application cache state isn't it?

lib/msal-node/docs/caching.md
```diff
@@ -38,25 +38,27 @@ cca.acquireTokenSilent(silentTokenRequest).then((response) => {
...
+Note that the in-memory cache is not scalable for server-side applications and performance will degrade after holding a few 100 tokens in cache. For web app and web API scenarios, this approximates to serving a few 100 users. For daemon app scenarios using client credentials grant to call other apps, this means a few 100 tenants. See [performance](#performance-and-security) below for more.
 
-## Persistence
+> :warning: We recommend persisting the cache with encryption for all production applications both for security and desired cache longevity. If you choose not to persist the cache, the [TokenCache](https://azuread.github.io/microsoft-authentication-library-for-js/ref/classes/_azure_msal_node.tokencache.html) interface is still available to access the cached entities.
```
sameerag: `encryption` is lost in the message. Can we make sure it is highlighted and both `persistence` and `encryption` are handled in the following sentence?

lib/msal-node/docs/caching.md
```diff
@@ -38,25 +38,27 @@ cca.acquireTokenSilent(silentTokenRequest).then((response) => {
...
 
-MSAL Node fires events when the cache is accessed, apps can choose whether to serialize or deserialize the cache. This often constitutes two actions:
+## Persistent cache
 
-1. Deserialize the cache from disk to MSAL's memory before accessing the cache
-2. If the cache in memory has changed, serialize the cache back
+MSAL Node fires events when the in-memory cache is accessed and apps can choose whether to serialize or deserialize the cache (see: [TokenCacheContext](https://azuread.github.io/microsoft-authentication-library-for-js/ref/classes/_azure_msal_common.tokencachecontext.html)) to persistence (e.g. a file, a SQL database and etc.). This constitutes two actions:
```
sameerag: Do we need to mention serialization? It is just a technique to ensure cache integrity. We can just say `persist` here. If you believe you need the advanced users to know the details for performance reasons, I prefer to separate the description and expand on why we settled on a `json` format to persist cache and we apply serialization and de-serialization to achieve the same.

sameerag: I see why you may have done this, as the API name consists of it. However, the above still applies.

lib/msal-node/docs/caching.md
```diff
@@ -65,59 +67,61 @@ interface ICachePlugin {
...
->        scopes: ["User.Read"],
->    });
->    performance.mark("acquireTokenSilent-end");
->
->    const measurementName = tokenResponse.fromCache ? "acquireTokenSilent-fromCache" : "acquireTokenSilent-fromNetwork";
->    performance.measure(measurementName, "acquireTokenSilent-start", "acquireTokenSilent-end");
-> ```
->
-> You should collect these along with the performance metrics generated by your storage solution (e.g. Redis) used for persistence.
+On confidential client apps that handle users (web apps that sign in users and call web APIs, and web APIs calling downstream web APIs), there can be many users and the users are processed in parallel. Our recommendation is to serialize one cache blob (see [CacheRecord](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-common/src/cache/entities/CacheRecord.ts)) per user. Use a key for partitioning the cache (*i.e.* **partition key**), such as:
```
sameerag:
```suggestion
On confidential client apps that handle users (web apps that sign in users and call web APIs, and web APIs calling downstream web APIs), there can be many users active concurrently for a given application. Our recommendation is to serialize one cache blob (see [CacheRecord](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-common/src/cache/entities/CacheRecord.ts)) per user. This would help with scaling the cache across a distributed system. Use a key for partitioning the cache (*i.e.* **partition key**), such as:
```

sameerag (approved):
Added some comments. Looks good otherwise.

lib/msal-node/docs/caching.md
derisen: Sorry for the late response. What I'm after here is to prevent any confusion about which cache the acquireToken* APIs primarily rely on. I think "representative of the application cache state" sounds more accurate though -will revise

lib/msal-node/docs/caching.md
derisen: done!

lib/msal-node/docs/caching.md
derisen: Sure thing, used save/load instead to be agnostic about this -there's enough hints elsewhere about using the serialization technique

lib/msal-node/docs/caching.md
```diff
@@ -38,25 +38,27 @@ cca.acquireTokenSilent(silentTokenRequest).then((response) => {
...
 
-## Cache in memory
+## In-memory cache
 
-If you choose not to persist the cache, the `TokenCache` interface is still available to access the tokens cached in memory. The lifetime of in-memory cache is the same as MSAL instance. If the MSAL instance restarts, the cache is erased when the process lifecycle finishes. We recommend persisting the cache with encryption for all real life applications both for security and desired cache longevity.
+MSAL maintains an in-memory cache. The in-memory cache is representative of the application cache state. The lifetime of in-memory cache is the same as the MSAL application object. If the process using MSAL restarts, the cache is erased when the process lifecycle finishes. This affects user scenarios as users will have to re-authenticate, and although the browser will do most some of the work, the end user experience suffers. Service to service scenarios also suffer because getting a token from AAD involves HTTP requests, and is much slower than getting a token from a cache.
```
Lawton:
```suggestion
MSAL maintains an in-memory cache. The in-memory cache is representative of the application cache state. The lifetime of in-memory cache is the same as the MSAL application object. If the process using MSAL restarts, the cache is erased when the process lifecycle finishes. This affects user scenarios as users will have to re-authenticate, and although the browser will do most of the work, the end user experience suffers. Service to service scenarios also suffer because getting a token from AAD involves HTTP requests, and is much slower than getting a token from a cache.
```

Lawton: I'm by no means a wordsmith but the last section is a bit confusing. Might be worth condensing to clarify what is relevant for browser or backend service
```suggestion
MSAL maintains an in-memory cache. The in-memory cache is representative of the application cache state. The lifetime of in-memory cache is the same as the MSAL application object. If the process using MSAL restarts, the cache is erased when the process lifecycle finishes. This affects user scenarios as users will have to re-authenticate. This means a worse user experience in browser or an expensive HTTP request in service to service scenarios. 
```

lib/msal-node/docs/caching.md
```diff
@@ -38,25 +38,27 @@ cca.acquireTokenSilent(silentTokenRequest).then((response) => {
...
 
-> :warning: Please note that the default in-memory cache is not scalable for server-side applications and performance will degrade after holding a few 100 tokens in cache. For web app and web API scenarios, this approximates to serving a few 100 users. For daemon app scenarios using client credentials grant to call other apps, this means a few 100 tenants. See [performance](#performance-and-security) below for more.
+Note that the in-memory cache is not scalable for server-side applications and performance will degrade after holding a few 100 tokens in cache. For web app and web API scenarios, this approximates to serving a few 100 users. For daemon app scenarios using client credentials grant to call other apps, this means a few 100 tenants. See [performance](#performance-and-security) below for more.
```
Lawton, line 52:
```suggestion
Note because this cache strategy is stored in-memory, it is not scalable for server-side applications and performance will degrade after holding a few 100 tokens in cache. For web app and web API scenarios, this approximates to serving a few 100 users. For daemon app scenarios using client credentials grant to call other apps, this means a few 100 tenants. See [performance](#performance-and-security) below for more.
```

lib/msal-node/docs/caching.md
Lawton, line 1: When working through MSAL it was not initially clear to me that I would need to use the integrated cache as refreshToken is not exposed. I think in this document it would be really helpful to document this at the top or even in the README.
1. Explicitly state what the MSAL cache schema format is. I assume it's similar to the exposed token information, but also includes the refreshToken.
2. Explain within this document that the refreshToken is intentionally not included in msal responses and the reasoning behind it.

lib/msal-node/docs/caching.md
```diff
@@ -65,59 +67,61 @@ interface ICachePlugin {
...
+```typescript
+class MyCachePlugin implements ICachePlugin {
+    private client: ICacheClient;
+
+    constructor(client: ICacheClient) {
+        this.client = client; // client object to access the persistent cache
+    }
+
+    public async beforeCacheAccess(cacheContext: TokenCacheContext): Promise<void> {
+        const cacheData = await this.client.get(); // get the cache from persistence
```
Lawton, line 83: Shouldn't there be a key passed in get/set for client?

Lawton (commented):
Adding thoughts based on request. Focused on cache document

lib/msal-node/docs/caching.md
derisen: the basic use cases (for instance, in a desktop or console app), you may not necessarily need a key, but I added a remark to read further for server side scenarios (and I think link to the samples cover this)

lib/msal-node/docs/caching.md
derisen: Sounds good, added a remark about refresh token usage. I think we intentionally didn't expose the cache schema itself, as its an implementation detail and might be subject to change in future. In general, developers shouldn't need to take a dependency on the cache schema, and if the available APIs don't cover every need there, that's likely a problem or bug (fyi, cache schema is [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/cache.json))

lib/msal-node/docs/caching.md
derisen: that's fair. I tried to reword this, hope its a bit more clear now

lib/msal-node/docs/caching.md
sameerag: I also think the reasoning behind `RT not being shared to the users` and `cache schema` are two distinct things and are orthogonal. We can take the RT q in an FAQ and may be link here if that is more appropriate.

PR #5788, opened 2023-03-13
Move node clients from `msal-common` to `msal-node`

konstantin-msft:
1. Move 4 clients used only by `msal-node` from `msal-common` to `msal-node`.
    - ClientCredentialClient
    - DeviceCodeClient
    - OnBehalfOfClient
    - UsernamePasswordClient
2. Move `request` and part of `response` docs from `msal-common` to `msal-node`.
3. Update references.

### DIFF
--- a/lib/msal-node/docs/response.md
+++ b/lib/msal-node/docs/response.md
@@ -1,3 +1,5 @@
+# Response
+
 MSAL will return an `AuthenticationResult.ts` object as a response to all acquire token APIs:
 
 #### `msal-browser` public APIs for token acquisition:
--- a/lib/msal-node/docs/response.md
+++ b/lib/msal-node/docs/response.md
@@ -1,0 +1,6 @@
+MSAL will return an `AuthenticationResult.ts` object as a response to all acquire token APIs:
+
+#### `msal-browser` public APIs for token acquisition:
+`loginPopup`, `acquireTokenPopup`, `acquireTokenSilent` or `handleRedirectPromise`
+
+Reference docs for `AuthenticationResult` expanding on each parameter can be found [here](https://azuread.github.io/microsoft-authentication-library-for-js/ref/msal-common/classes/_src_response_authenticationresult_.authenticationresult.html).
--- a/lib/msal-node/src/client/ClientCredentialClient.ts
+++ b/lib/msal-node/src/client/ClientCredentialClient.ts
@@ -3,24 +3,14 @@
  * Licensed under the MIT License.
  */
 
-import { ClientConfiguration } from "@azure/msal-common/src/config/ClientConfiguration";
-import { BaseClient } from "@azure/msal-common/src/client/BaseClient";
-import { Authority } from "@azure/msal-common/src/authority/Authority";
-import { RequestParameterBuilder } from "@azure/msal-common/src/request/RequestParameterBuilder";
-import { ScopeSet } from "@azure/msal-common/src/request/ScopeSet";
-import { GrantType , CredentialType, CacheOutcome, Constants, AuthenticationScheme } from "@azure/msal-common/src/utils/Constants";
-import { ResponseHandler } from "@azure/msal-common/src/response/ResponseHandler";
-import { AuthenticationResult } from "@azure/msal-common/src/response/AuthenticationResult";
-import { CommonClientCredentialRequest } from "@azure/msal-common/src/request/CommonClientCredentialRequest";
-import { CredentialFilter, CredentialCache } from "@azure/msal-common/src/cache/utils/CacheTypes";
-import { AccessTokenEntity } from "@azure/msal-common/src/cache/entities/AccessTokenEntity";
-import { TimeUtils } from "@azure/msal-common/src/utils/TimeUtils";
-import { StringUtils } from "@azure/msal-common/src/utils/StringUtils";
-import { RequestThumbprint } from "@azure/msal-common/src/network/RequestThumbprint";
-import { ClientAuthError } from "@azure/msal-common/src/error/ClientAuthError";
-import { ServerAuthorizationTokenResponse } from "@azure/msal-common/src/response/ServerAuthorizationTokenResponse";
-import { IAppTokenProvider } from "@azure/msal-common/src/config/AppTokenProvider";
-import { UrlString } from "@azure/msal-common/src/url/UrlString";
+import {
+    AccessTokenEntity,
+    AuthenticationResult, AuthenticationScheme, Authority, BaseClient, CacheOutcome, ClientAuthError,
+    ClientConfiguration,
+    CommonClientCredentialRequest, Constants, CredentialCache, CredentialFilter, CredentialType, GrantType,
+    IAppTokenProvider, RequestParameterBuilder, RequestThumbprint, ResponseHandler,
+    ScopeSet, ServerAuthorizationTokenResponse, StringUtils, TimeUtils, UrlString
+} from "@azure/msal-common";
 
 /**
  * OAuth2.0 client credential grant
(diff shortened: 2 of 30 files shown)


lib/msal-node/src/client/ClientCredentialClient.ts
```diff
@@ -3,24 +3,24 @@
...
-import { CredentialFilter, CredentialCache } from "../cache/utils/CacheTypes";
-import { AccessTokenEntity } from "../cache/entities/AccessTokenEntity";
-import { TimeUtils } from "../utils/TimeUtils";
-import { StringUtils } from "../utils/StringUtils";
-import { RequestThumbprint } from "../network/RequestThumbprint";
-import { ClientAuthError } from "../error/ClientAuthError";
-import { ServerAuthorizationTokenResponse } from "../response/ServerAuthorizationTokenResponse";
-import { IAppTokenProvider } from "../config/AppTokenProvider";
-import { UrlString } from "../url/UrlString";
+import { ClientConfiguration } from "@azure/msal-common/src/config/ClientConfiguration";
```
sameerag: Are we doing `package`  based imports now?

tnorling: No we should not do this, these things should be exported from msal-common if not already.

sameerag: That is what I thought. And this is common pattern across the board, to export the files from `msal-common` and not the pre-compiled package. @konstantin-msft why the change??

sameerag (dismissed):
Please update the `import` paths

derisen:
@konstantin-msft think we should probably relocate [this doc](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-common/docs/request.md) to msal-node side as well (and maybe rename it to `acquire-token` while at it?)

lib/msal-node/src/client/ClientCredentialClient.ts
hectormmg: +1, I think the reason most of these aren't exported already is because none of the client classes were implemented at the wrapper library level. Having to export all of these common classes may be an argument against factoring the confidential clients out.

ghost:

konstantin-msft:
> @konstantin-msft think we should probably relocate [this doc](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-common/docs/request.md) to msal-node side as well (and maybe rename it to `acquire-token` while at it?)

That's a very good point. Thank you Dogan!
I have moved `request` and part of `response` docs to `msal-node`. I would suggest keeping naming the same to avoid confusion.

lib/msal-node/src/client/ClientCredentialClient.ts
konstantin-msft: My bad. Fixed imports, thank you.

konstantin-msft:
> Please update the `import` paths

Done.

lib/msal-node/docs/response.md
```diff
@@ -0,0 +1,6 @@
+MSAL will return an `AuthenticationResult.ts` object as a response to all acquire token APIs:
```
jo-arroyo, line 3: Add heading here?

lib/msal-node/test/client/ConfidentialClientApplication.spec.ts
```diff
@@ -1,19 +1,36 @@
-import { ConfidentialClientApplication } from './../../src/client/ConfidentialClientApplication';
-import { ClientConfiguration, AuthorizationCodeClient, RefreshTokenClient, AuthenticationResult, ClientCredentialClient, OnBehalfOfClient, UsernamePasswordClient } from '@azure/msal-common';
+import {
```
jo-arroyo, line 1: Combine imports from src with imports below?

tnorling (approved):
Looks good. Just as a general comment; we've been making a lot of major bumps of common lately (which is fine!) and it seems like with this change msal-node will be taking on an even heavier dependency on msal-common APIs, which were previously abstracted away from it. We recently discussed our versioning strategy with regards to common and it may be time to revisit that and seriously consider some of the ideas folks had (i.e. pin dependencies to a specific version or stop publishing it as a separate package and bundle it in with the public packages)

lib/msal-node/test/client/PublicClientApplication.spec.ts
```diff
@@ -1,28 +1,45 @@
-import { PublicClientApplication } from './../../src/client/PublicClientApplication';
-import { Configuration, ILoopbackClient, InteractiveRequest } from './../../src/index';
+import { PublicClientApplication } from '../../src';
+import { Configuration, DeviceCodeClient, ILoopbackClient, InteractiveRequest } from '../../src';
```
jo-arroyo: Combine `src` imports?

lib/msal-node/test/client/PublicClientApplication.spec.ts
```diff
@@ -1,28 +1,45 @@
...
 import { ID_TOKEN_CLAIMS, mockAuthenticationResult, TEST_CONSTANTS, TEST_DATA_CLIENT_INFO } from '../utils/TestConstants';
 import {
-    ClientConfiguration, AuthenticationResult, AuthorizationCodeClient, RefreshTokenClient, UsernamePasswordClient,
-    SilentFlowClient, ProtocolMode, Logger, LogLevel, ClientAuthError, AccountInfo, ServerAuthorizationCodeResponse
+    ClientConfiguration,
```
jo-arroyo, line 3: Same as above

lib/msal-node/test/client/PublicClientApplication.spec.ts
```diff
@@ -1,28 +1,45 @@
...
+    ServerAuthorizationCodeResponse
 } from '@azure/msal-common';
-import { CryptoProvider } from '../../src/crypto/CryptoProvider';
-import { DeviceCodeRequest } from '../../src/request/DeviceCodeRequest';
-import { AuthorizationCodeRequest } from '../../src/request/AuthorizationCodeRequest';
-import { RefreshTokenRequest } from '../../src/request/RefreshTokenRequest';
-import { AuthorizationUrlRequest } from "../../src/request/AuthorizationUrlRequest";
-import { UsernamePasswordRequest } from '../../src/request/UsernamePasswordRequest';
-import { SilentFlowRequest } from '../../src/request/SilentFlowRequest';
+import {
```
jo-arroyo, line 15: Same as above

jo-arroyo (approved):
Approved, with a few formatting comments

konstantin-msft:
> Looks good. Just as a general comment; we've been making a lot of major bumps of common lately (which is fine!) and it seems like with this change msal-node will be taking on an even heavier dependency on msal-common APIs, which were previously abstracted away from it. We recently discussed our versioning strategy with regards to common and it may be time to revisit that and seriously consider some of the ideas folks had (i.e. pin dependencies to a specific version or stop publishing it as a separate package and bundle it in with the public packages)

I would say with this PR `msal-node` will be taking on additional request/response/cache on `msal-common` types rather than APIs. Other than that, I totally agree - pinning deps to a specific version sounds like a good idea to me.

lib/msal-node/docs/response.md
konstantin-msft: Good catch, done

lib/msal-node/test/client/ConfidentialClientApplication.spec.ts
konstantin-msft: fixed.

lib/msal-node/test/client/PublicClientApplication.spec.ts
konstantin-msft: fixed.

lib/msal-node/test/client/PublicClientApplication.spec.ts
konstantin-msft: fixed.

lib/msal-node/test/client/PublicClientApplication.spec.ts
konstantin-msft: fixed.

PR #5789, opened 2023-03-13
exposing missing token claims parameters

lalimasharda:
Bug fix:
- Added missing `tenant_region_scope` and `tenant_region_sub_scope` claims from Token Claims

### DIFF
--- a/lib/msal-common/src/account/TokenClaims.ts
+++ b/lib/msal-common/src/account/TokenClaims.ts
@@ -62,4 +62,6 @@ export type TokenClaims = {
     amr?: string[],
     idp?: string,
     auth_time?: number,
+    tenant_region_scope?: string,
+    tenant_region_sub_scope?: string
 };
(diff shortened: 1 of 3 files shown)


lib/msal-common/src/account/TokenClaims.ts
```diff
@@ -62,4 +62,6 @@ export type TokenClaims = {
     amr?: string[],
     idp?: string,
     auth_time?: number,
+    tenant_region_scope?: string,
```
derisen, line 68: maybe a one liner to describe which scenarios this is relevant to?

derisen (approved):

tnorling (approved):

lib/msal-common/src/account/TokenClaims.ts
lalimasharda: Added a comment on claim usage

PR #5792, opened 2023-03-13
Account cache optimization

tnorling:
Today on every lookup of an account in the cache we iterate through every entry in the cache (including cache entries not belonging to MSAL), attempt to parse it into JSON and then validate that it matches our account shape. This is incredibly inefficient and becomes more inefficient with each additional thing stored in browser storage.

This PR adds a new cache entry (with a static key) which contains an array of keys to all of the accounts known to MSAL. Now when the accounts need to be enumerated the worst case number of cache entries touched and parsed is 1 + number of accounts.

A new configuration option is added which determines whether or not this new cache entry should be created on initialization. This is to help migrate apps using localStorage from versions before this change.

Follow up PR will address the same issue for id, access and refresh tokens.


lib/msal-browser/src/cache/BrowserCacheManager.ts
```diff
@@ -124,6 +127,36 @@ export class BrowserCacheManager extends CacheManager {
...
+     * @returns 
+     */
+    private createAccountKeyMap(): void {
+        const accountKeys = this.getItem(StaticCacheKeys.ACCOUNT_KEYS);
+        if (accountKeys) {
+            // Account key map already exists, no need to iterate through cache
+            return;
+        }
+
+        const allKeys = this.browserStorage.getKeys();
```
peterzenz, line 142: Add config to skip this step if the developer knows they don't have old versions in the wild that they care about.

lib/msal-browser/src/cache/BrowserCacheManager.ts
```diff
@@ -124,6 +127,36 @@ export class BrowserCacheManager extends CacheManager {
...
 
+    /**
+     * Searches all cache entries for MSAL accounts and creates the account key map
+     * This is used to migrate users from older versions of MSAL which did not create the map.
+     * @returns 
+     */
+    private createAccountKeyMap(): void {
+        const accountKeys = this.getItem(StaticCacheKeys.ACCOUNT_KEYS);
+        if (accountKeys) {
+            // Account key map already exists, no need to iterate through cache
```
peterzenz, line 138: Please add telemetry so we know whether partners are using this flow vs the legacy scan everything approach.  This will let us pivot perf queries around this for other perf investigations.

peterzenz (commented):
This PR is still marked as draft, I'll re-review it once it's marked ready for review.  Thanks!

lib/msal-browser/src/cache/BrowserCacheManager.ts
tnorling: That's already part of this PR, please see line 47 :)

lib/msal-browser/src/cache/BrowserCacheManager.ts
tnorling: Just to clear up some confusion on the approach here, it's not an either or. This new flow is always used for cache lookups regardless of the migration setting. The "legacy scan everything approach" is a one time fix up, if enabled and if needed, on initialization.

I'll see what I can do about adding telemetry here. We currently don't instrument anything in the CacheManager and don't instrument the PCA constructor so this may need to be tackled separately

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -236,7 +240,9 @@ export function buildConfiguration({ auth: userInputAuth, cache: userInputCache,
     const DEFAULT_CACHE_OPTIONS: Required<CacheOptions> = {
         cacheLocation: BrowserCacheLocation.SessionStorage,
         storeAuthStateInCookie: false,
-        secureCookies: false
+        secureCookies: false,
+        // Default cache migration to true if cache location is localStorage since entries are preserved across tabs/windows. Migration has little to no benefit in sessionStorage and memoryStorage
```
peterzenz, line 244: We had discussed this morning that if using memory storage (session as well?) that we can just store objects and skip the serialization overhead there.  Can you file a task for this as a future optimization.

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -261,60 +291,78 @@ export abstract class CacheManager implements ICacheManager {
...
+    /**
+     * Returns true if the given key matches our account key schema. Also matches homeAccountId and/or tenantId if provided
+     * @param key 
+     * @param homeAccountId 
+     * @param tenantId 
+     * @returns 
+     */
+    isAccountKey(key: string, homeAccountId?: string, tenantId?: string): boolean {
+        if (key.split(Separators.CACHE_KEY_SEPARATOR).length < 3) {
+            // Account cache keys contain 3 items separated by '-' (each item may also contain '-')
```
peterzenz, line 349: Sorry, the separator is '-', but each item can also contain '-' in it?

peterzenz (approved):
Just the one comment on the '-' separator.  I'm likely missing something here. Thanks.

lib/msal-browser/src/config/Configuration.ts
tnorling: Done. This would just be for memory storage. SessionStorage will still need to be serialized

lib/msal-common/src/cache/CacheManager.ts
tnorling: Yes...an unfortunate decision

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -555,39 +603,35 @@ export abstract class CacheManager implements ICacheManager {
...
+        allAccountKeys.forEach((cacheKey) => {
             removedAccounts.push(this.removeAccount(cacheKey));
         });
 
         await Promise.all(removedAccounts);
-        return true;
     }
 
     /**
      * returns a boolean if the given account is removed
```
derisen:
```suggestion
     * Removes the account and related tokens for a given account key
```

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -555,39 +603,35 @@ export abstract class CacheManager implements ICacheManager {
...
      * @param account
      */
-    async removeAccount(accountKey: string): Promise<boolean> {
+    async removeAccount(accountKey: string): Promise<void> {
```
derisen, line 621: Curious why the decision to no longer return a boolean

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -85,6 +85,10 @@ export type CacheOptions = {
      * If set, MSAL sets the "Secure" flag on cookies so they can only be sent over HTTPS. By default this flag is set to false.
      */
     secureCookies?: boolean;
+    /**
+     * If set, MSAL will attempt to migrate cache entries from older versions on initialization. By default this flag is set to true if cacheLocation is localStorage, otherwise false.
+     */
+    cacheMigrationEnabled?: boolean;
```
derisen, line 91: Can we update the configuration doc as well?

lib/msal-common/src/cache/CacheManager.ts
tnorling: We were always returning true so it wasn't meaningful

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -151,7 +151,8 @@ export abstract class ClientApplication {
         const nativeCacheOptions: Required<CacheOptions> = {
             cacheLocation: BrowserCacheLocation.MemoryStorage,
             storeAuthStateInCookie: false,
-            secureCookies: false
+            secureCookies: false,
+            cacheMigrationEnabled: false
```
sameerag, line 155: There is another API called `migrateCacheEntries` in `BrowserCacheManager` (for ADAL to MSAL). How do we distinguish these cases? Or does this boolean always translate to whatever the latest cache look up is? I am fine with the name if we stick to the later logic.

lib/msal-browser/src/cache/BrowserCacheManager.ts
```diff
@@ -187,6 +223,67 @@ export class BrowserCacheManager extends CacheManager {
...
+     * @param key 
+     */
+    addAccountKeyToMap(key: string): void {
+        this.logger.trace("BrowserCacheManager.addAccountKeyToMap called");
+        this.logger.tracePii(`BrowserCacheManager.addAccountKeyToMap called with key: ${key}`);
+        const accountKeys = this.getAccountKeys();
+        if (accountKeys.indexOf(key) === -1) {
+            // Only add key if it does not already exist in the map
+            accountKeys.push(key);
+            this.setItem(StaticCacheKeys.ACCOUNT_KEYS, JSON.stringify(accountKeys));
```
sameerag, line 255: Does this hold if the number of accounts exceed a given limit? `localStorage` or `sessionStorage` entries do have a size limit I assume?

lib/msal-browser/src/cache/BrowserCacheManager.ts
```diff
@@ -124,6 +127,36 @@ export class BrowserCacheManager extends CacheManager {
         }
     }
 
+    /**
+     * Searches all cache entries for MSAL accounts and creates the account key map
+     * This is used to migrate users from older versions of MSAL which did not create the map.
+     * @returns 
+     */
+    private createAccountKeyMap(): void {
+        const accountKeys = this.getItem(StaticCacheKeys.ACCOUNT_KEYS);
```
sameerag, line 136: This defaults to whatever memory option is set, which implies we need to serialize and deserialize the accountKeys every time we load the page. Same q as above, concerned if we ever cross size limits, other wise looks good.

lib/msal-browser/src/config/Configuration.ts
sameerag: I was wondering if there is a way we can avoid de-serialization and serialization unless we reload the page. I think the tricky part is that the user can close the web page/reload at any point and we will lose the in-memory data structure if we do not do it per account creation and deletion. 

Are there any DOM APIs we can leverage? I do not think this is a blocker and may be the perf gains are not so significant. We can circle back on this.

sameerag (approved):
Lgtm. Have some qs on size.

lib/msal-browser/src/app/ClientApplication.ts
tnorling: This boolean controls all cache migrations, including the ADAL to MSAL one you mentioned. We don't need to distinguish the cases. Essentially, if you have recently upgraded to a new version of MSAL or migrated from ADAL to MSAL you should leave this flag on (Default). Then once your application has been using the latest version of MSAL long enough that most users' caches have been migrated you can feel free to turn this off and skip the migration step.

lib/msal-browser/src/cache/BrowserCacheManager.ts
tnorling: Should be fine, there's a size limit for the entire local/session storage but I'm not aware of a limit for individual entries. In any case, I don't imagine a scenario where anyone would have more than a handful of accounts signed in.

lib/msal-browser/src/config/Configuration.ts
tnorling: Done

ghost:
:tada:`@azure/msal-common@v12.0.0` has been released which incorporates this pull request.:tada:

We recommend upgrading to the latest version of `@azure/msal-browser` or `@azure/msal-node` to take advantage of this change.

Handy links:
* [Release Notes](https://github.com/AzureAD/microsoft-authentication-library-for-js/releases/tag/msal-common-v12.0.0)
* [Official NPM Release](https://www.npmjs.com/package/@azure/msal-common/v/12.0.0)

ghost:
:tada:`@azure/msal-browser@v2.35.0` has been released which incorporates this pull request.:tada:

Handy links:
* [Release Notes](https://github.com/AzureAD/microsoft-authentication-library-for-js/releases/tag/msal-browser-v2.35.0)
* [Official NPM Release](https://www.npmjs.com/package/@azure/msal-browser/v/2.35.0)

PR #5794, opened 2023-03-14
Expose controller artifacts for 1p lib

konstantin-msft:
- Export `msal-browser` internals for 1p lib.
- Expose controller artifacts for 1p lib.
- Make eligible fields `readonly` for `StandardController`.

Note: controller artifacts exposed by this PR are required for 1p.

### DIFF
--- a/lib/msal-browser/src/controllers/IController.ts
+++ b/lib/msal-browser/src/controllers/IController.ts
@@ -3,8 +3,6 @@
  * Licensed under the MIT License.
  */
 
-import { IPublicClientApplication } from "../app/IPublicClientApplication";
-
 import {
     AuthenticationResult,
     AccountInfo,
--- a/lib/msal-browser/src/controllers/IController.ts
+++ b/lib/msal-browser/src/controllers/IController.ts
@@ -31,50 +31,96 @@ import { SilentIframeClient } from "../interaction_client/SilentIframeClient";
 
 export interface IController {
     initialize(): Promise<void>;
+
     acquireTokenPopup(request: PopupRequest): Promise<AuthenticationResult>;
+
     acquireTokenRedirect(request: RedirectRequest): Promise<void>;
+
     acquireTokenSilent(silentRequest: SilentRequest): Promise<AuthenticationResult>;
+
     acquireTokenByCode(request: AuthorizationCodeRequest): Promise<AuthenticationResult>;
+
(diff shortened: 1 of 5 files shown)


peterzenz (approved):

lib/msal-browser/src/controllers/IController.ts
```diff
@@ -3,6 +3,78 @@
...
+import { AuthorizationCodeRequest } from "../request/AuthorizationCodeRequest";
+import { BrowserConfiguration } from "../config/Configuration";
+import { BrowserCacheManager } from "../cache/BrowserCacheManager";
+import { NativeMessageHandler } from "../broker/nativeBroker/NativeMessageHandler";
+import { EventHandler } from "../event/EventHandler";
+import { PopupClient } from "../interaction_client/PopupClient";
+import { SilentIframeClient } from "../interaction_client/SilentIframeClient";
 
-export type IController = IPublicClientApplication;
+export interface IController {
```
sameerag, line 32: nit: Can we please have spaces between APIs for readability? What do we do today across headers in our library?

lib/msal-browser/src/controllers/StandardController.ts
```diff
@@ -1109,10 +1109,47 @@ export class StandardController implements IController {
...
 
+    public getPerformanceClient(): IPerformanceClient {
+        return this.performanceClient;
+    }
+
+    public getBrowserStorage(): BrowserCacheManager  {
+        return this.browserStorage;
+    }
+
+    public getNativeInternalStorage(): BrowserCacheManager  {
```
sameerag, line 1133: Why do we want to expose these?

konstantin-msft: native internal storage is used by 1p lib.

lib/msal-browser/src/controllers/IController.ts
konstantin-msft: done.

lib/msal-browser/src/controllers/StandardController.ts
sameerag: I meant, why `public` as `protected` works for 1P today since we derive it? Is this for future proofing where we want `1P` to be based on the `public interface` of 3P?

konstantin-msft: We extend `PublicClientApplication`, not a controller. That's why I have made those methods public.

sameerag: Ah! Alright. Thanks.

sameerag (approved):

PR #5805, opened 2023-03-20
Exception is thrown in acquireTokenByClientCredential if tenantId is missing

### LINKED ISSUE #5567: Exception is thrown in acquireTokenByClientCredential if tenantId is missing
### Core Library

MSAL Node (@azure/msal-node)

### Core Library Version

2.14.1

### Wrapper Library

Not Applicable

### Wrapper Library Version

N/A

Confidential

### Description

An exception should be thrown in AcquireTokenByCredential if "common" or "organization" are used in lieu of the tenant_id in the authority in the auth configuration. As it stands, using "common" or "organization" will work, but the behavior is undefined. AAD will try to guess the correct tenant_id, which is undesired behavior.

### Error Message

_No response_

### Msal Logs

_No response_

### MSAL Configuration

```javascript
auth: {
    clientId: "<ENTER_CLIENT_ID>",
    authority: "https://login.microsoftonline.com/<ENTER_TENANT_ID>",
    clientSecret: "<ENTER_CLIENT_SECRET>",
},
```

### Relevant Code Snippets

```javascript
N/A
```

### Reproduction Steps

N/A

### Expected Behavior

An exception should be thrown when using AcquireTokenByCredential if "common" or "organization" are used in the authority instead of the tenant_id.

### Identity Provider

Azure AD / MSA

None (Server)

### Regression

_No response_

### Source

Internal (Microsoft)

Robbie-Microsoft:


lib/msal-common/src/error/ClientAuthError.ts
```diff
@@ -211,6 +211,10 @@ export const ClientAuthErrorMessage = {
     userCanceledError: {
         code: "user_canceled",
         desc: "User canceled the flow."
+    },
+    missingTenantIdError: {
+        code: "missing_tenant_id_error",
+        desc: "TenantId was set to  \"common\" or \"organization\". AAD will try to guess the correct TenantId, which is undesired behavior."
```
bgavrilMS:
```suggestion
        desc: "A tenant id - not common / organizations - must be specified when using the client_credentials flow. "
```

bgavrilMS (approved):
LGTM. This is for the new major version ?

lib/msal-common/src/error/ClientAuthError.ts
derisen: +1

lib/msal-common/src/error/ClientAuthError.ts
```diff
@@ -576,4 +580,11 @@ export class ClientAuthError extends AuthError {
     static createUserCanceledError(): ClientAuthError {
         return new ClientAuthError(ClientAuthErrorMessage.userCanceledError.code, ClientAuthErrorMessage.userCanceledError.desc);
     }
+
+    /**
+     * Creates an error for during acquireTokenByClientCredential when TenantId is set to "common" or "organization"
```
derisen:
```suggestion
     * Creates an error for during acquireTokenByClientCredential when TenantId is set to "common" or "organizations"
```

lib/msal-node/src/client/ConfidentialClientApplication.ts
```diff
@@ -97,6 +97,15 @@ export class ConfidentialClientApplication extends ClientApplication implements
             clientAssertion
         };
 
+        /*
+         * valid request should not have "common" or "organization" in lieu of the tenant_id in the authority in the auth configuration
+         * example authority: "https://login.microsoftonline.com/TenantId",
+         */
+        const tenantId = validRequest.authority.split("/")[3];
+        if ((tenantId === "common") || (tenantId === "organization")) {
```
derisen:
```suggestion
        if ((tenantId === "common") || (tenantId === "organizations")) {
```

derisen: @Robbie-Microsoft @bgavrilMS I know it doesn't make sense with client credentials, but do we know what is the behavior if someone uses `consumers`? Should we check for this as well?

lib/msal-node/src/client/ConfidentialClientApplication.ts
```diff
@@ -97,6 +97,15 @@ export class ConfidentialClientApplication extends ClientApplication implements
             clientAssertion
         };
 
+        /*
+         * valid request should not have "common" or "organization" in lieu of the tenant_id in the authority in the auth configuration
+         * example authority: "https://login.microsoftonline.com/TenantId",
+         */
+        const tenantId = validRequest.authority.split("/")[3];
```
derisen: > validRequest.authority.split("/")[3];

Slightly worried if this could fail to retrieve the id for certain authority strings/formats. Is there an existing method for stripping the id in the library perhaps? or should we make one and write a unit test for it?

lib/msal-node/src/client/ConfidentialClientApplication.ts
bgavrilMS: I think if you try to use consumers you'll get a good error message from the STS

lib/msal-node/src/client/ConfidentialClientApplication.ts
bgavrilMS: Agreed, this is not great - there must be other places where we extract tenant from the authority string @Robbie-Microsoft ?

bgavrilMS: For reference, the .NET algorithm is: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs#L367

Robbie-Microsoft: All this does is remove the "/"s from the authority URL by splitting the string at each occurrence of it. In this case, we want the very last chunk of string. Perhaps I could change this to: 

```
const segments = validRequest.authority.split("/").filter(segment => segment.trim());
const tenantId = segments[segments.length - 1];
```

it appears to me that this is similar to what .NET does.

Robbie-Microsoft: I don't see anything in the common or node utils that does this already.

lib/msal-common/src/error/ClientAuthError.ts
sameerag: +1. Can we please make this change?

lib/msal-node/src/client/ConfidentialClientApplication.ts
sameerag: `Authority.ts` has this code:
```
    /**
     * Get tenant for authority.
     */
    public get tenant(): string {
        return this.canonicalAuthorityUrlComponents.PathSegments[0];
    }
```

can we reuse it?

lib/msal-node/src/client/ConfidentialClientApplication.ts
sameerag: Can we test it?

lib/msal-node/src/client/ConfidentialClientApplication.ts
Robbie-Microsoft: @sameerag I can't reuse that method because it's part of the Authority class, and here we only have an authority string. However, that code uses the method getUrlComponents from [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-common/src/url/UrlString.ts#L147). I can import that function to msal-node from msal-common I suppose.

lib/msal-node/src/client/ConfidentialClientApplication.ts
derisen: Ok, looks like it doesn't give an error actually. It's the same behavior as if using common or organizations -it returns a token without the necessary permissions, which results in 401 when calling Graph. I suppose we should check it.

change/@azure-msal-node-8b9bf4a1-ea08-4013-8a4d-a6deb79b04ce.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
hectormmg: I think this needs to be a major change type since it breaks v1 behavior

lib/msal-node/src/client/ConfidentialClientApplication.ts
Robbie-Microsoft: @derisen are you saying I should add it to this if statement and the error message?

derisen: Yes, I think we should -it's less likely for someone to use consumers here but if they do, it's hard to figure out what's wrong from the response. We should also use the AADAuthorityConstants here instead of hardcoding it.

change/@azure-msal-node-8b9bf4a1-ea08-4013-8a4d-a6deb79b04ce.json
sameerag: This is going to v3 right?

sameerag (approved):

change/@azure-msal-node-8b9bf4a1-ea08-4013-8a4d-a6deb79b04ce.json
Robbie-Microsoft: @sameerag yep

PR #5806, opened 2023-03-20
Token cache optimization

tnorling:
Follow up to #5792, addressing the same problem but for token lookups.

Today on every lookup of tokens in the cache we iterate through every entry in the cache (including cache entries not belonging to MSAL), attempt to parse it into JSON and then validate that it matches our credential shape. This is incredibly inefficient and becomes more inefficient with each additional thing stored in browser storage.

This PR adds a new cache entry which contains an object of keys to all of the tokens known to MSAL for a given clientId. Now when the tokens need to be enumerated the worst case number of cache entries touched and parsed is 1 + number of tokens of the given type.

Example:

```js
key: msal.token.keys.<clientId>
value: {
  idToken: ["idToken-key-1", "idToken-key-2"],
  accessToken: ["accessToken-key-1", "accessToken-key-2"],
  refreshToken: ["refreshToken-key-1"]
}
```


lib/msal-common/test/cache/MockCache.ts
```diff
@@ -124,6 +124,23 @@ export class MockCache {
         const atThree = CacheManager.toObject(new AccessTokenEntity(), atThreeData);
         this.cacheManager.setAccessTokenCredential(atThree);
 
+        // BEARER with AuthScheme Token
```
hectormmg, line 127: Is this name right? I'd assume bearer tokens still have no auth scheme in the cache key to prevent backwards incompatible changes?

tnorling: hmm I thought everything should be with auth scheme explicitly now even if we still support and know how to read old entries without. Is that not the case?

In any case, nothing changed functionally around this in the library, this code here is just testing all combinations and scenarios.

hectormmg: No, bearer tokens still use the regular -accesstoken- key signature. We could make that change for v3 since it's a breaking change.

hectormmg (approved):
Looks great!

lib/msal-common/test/cache/MockCache.ts
hectormmg: I guess "should" is the right word to use here, it's not an extensive change, I'll create a ticket.

lib/msal-browser/src/cache/BrowserCacheManager.ts
```diff
@@ -132,25 +131,71 @@ export class BrowserCacheManager extends CacheManager {
...
                 if (value) {
-                    try {
-                        const accountObj = JSON.parse(value);
-                        if (AccountEntity.isAccountEntity(accountObj)) {
-                            this.addAccountKeyToMap(key);
+                    const credObj = this.validateAndParseJson(value);
+                    if (credObj && credObj.hasOwnProperty("credentialType")) {
+                        switch (credObj["credentialType"]) {
+                            case CredentialType.ID_TOKEN:
+                                if (IdTokenEntity.isIdTokenEntity(credObj)) {
```
sameerag, line 154: This means that there is a cache item that has "idtoken" in its key but not a valid one. Can we log a warning? Same applies below for other cred types.

lib/msal-browser/src/cache/BrowserCacheManager.ts
```diff
@@ -286,6 +331,135 @@ export class BrowserCacheManager extends CacheManager {
...
+     * Gets the keys for the cached tokens associated with this clientId
+     * @returns 
+     */
+    getTokenKeys(): TokenKeys {
+        this.logger.trace("BrowserCacheManager.getTokenKeys called");
+        const item = this.getItem(`${StaticCacheKeys.TOKEN_KEYS}.${this.clientId}`);
+        if (item) {
+            const tokenKeys = this.validateAndParseJson(item);
+            if (tokenKeys && 
+                tokenKeys.hasOwnProperty("idToken") &&
```
sameerag, line 380: I do not think this will ever happen, but if there is ever a key with "idtoken" property in it added to the set, we need to validate the value when we return it. I think we do, just putting this here to confirm it.

lib/msal-browser/src/cache/BrowserCacheManager.ts
```diff
@@ -286,6 +331,135 @@ export class BrowserCacheManager extends CacheManager {
...
+     * @param type 
+     */
+    removeTokenKey(key: string, type: CredentialType): void {
+        this.logger.trace("BrowserCacheManager removeTokenKey called");
+        const tokenKeys = this.getTokenKeys();
+
+        switch (type) {
+            case CredentialType.ID_TOKEN:
+                const idRemoval = tokenKeys.idToken.indexOf(key);
+                if (idRemoval > -1) {
```
sameerag, line 448: Are we swallowing the errors if the given key is not present in the map? Should we at least log it? Same applies below!

lib/msal-browser/src/cache/BrowserCacheManager.ts
tnorling: Are you asking about validating the key map or validating the actual credential?

lib/msal-common/src/cache/entities/CredentialEntity.ts
```diff
@@ -42,37 +42,6 @@ export class CredentialEntity {
...
-            } catch (err) {
-                // Lookbehind is not supported (Safari or older versions of IE) - removing it
-                CredentialEntity.credentialRegexMap.set(
-                    CredentialType[credKey],
-                    new RegExp(`${separator}${CredentialEntity.credentialDomainRegex}${separator}${credVal}${separator}`));
-            }
-        }
-    }
-
     /**
```
sameerag, line 45: I understand why you removed it, but this was fixing one of the existing bugs. I think we can mitigate this by validating the value when we return the cache item to the user + making sure we delete the "not properly formatted" keys. Just calling out the impact of removing this.

However, I think we cannot control who stores what in localStorage/sessionStorage and we cannot validate every entry. As long as we handle the cache we store and parse error free, this should be fine. cc @konstantin-msft

tnorling: The bug this was addressing is also fixed by the new model so this became redundant. I don't have concerns around this.

lib/msal-browser/src/cache/BrowserCacheManager.ts
tnorling: Logging added

lib/msal-browser/src/cache/BrowserCacheManager.ts
tnorling: Logging added

lib/msal-browser/src/cache/BrowserCacheManager.ts
sameerag: Actual credential

tnorling: Yes, value validation is done before returning or using any tokens

lib/msal-common/src/cache/entities/CredentialEntity.ts
peterzenz: Sameera - was there a broader test added to cover the code that Thomas was removing here?  Something deeper than is the regex doing what's expected.  If so, then that test should still pass.

lib/msal-browser/src/cache/BrowserCacheManager.ts
```diff
@@ -132,25 +131,80 @@ export class BrowserCacheManager extends CacheManager {
...
+                            case CredentialType.ID_TOKEN:
+                                if (IdTokenEntity.isIdTokenEntity(credObj)) {
+                                    this.logger.trace("BrowserCacheManager:createKeyMaps - idToken found, saving key to token key map");
+                                    this.logger.tracePii(`BrowserCacheManager:createKeyMaps - idToken with key: ${key} found, saving key to token key map`);
+                                    const idTokenEntity = CacheManager.toObject(new IdTokenEntity(), credObj);
+                                    const newKey = this.updateCredentialCacheKey(key, idTokenEntity);
+                                    this.addTokenKey(newKey, CredentialType.ID_TOKEN);
+                                    return;
+                                } else {
+                                    this.logger.trace("BrowserCacheManager:createKeyMaps - key found matching idToken schema with value containing idToken credentialType field but value failed IdTokenEntity validation, skipping.");
```
peterzenz, line 162: I'd like to see telemetry that logs a datapoint indicating we have this happening in production.  Similarly for the other types we handle below.

lib/msal-browser/src/cache/BrowserCacheManager.ts
```diff
@@ -286,6 +340,144 @@ export class BrowserCacheManager extends CacheManager {
...
+        if (item) {
+            const tokenKeys = this.validateAndParseJson(item);
+            if (tokenKeys && 
+                tokenKeys.hasOwnProperty("idToken") &&
+                tokenKeys.hasOwnProperty("accessToken") &&
+                tokenKeys.hasOwnProperty("refreshToken")
+            ) {
+                return tokenKeys as TokenKeys;
+            } else {
+                this.logger.error("BrowserCacheManager.getTokenKeys - Token keys found but in an unknown format. Returning empty key map.");
```
peterzenz, line 386: This might be another good telemetry datapoint to add.

lib/msal-browser/src/cache/BrowserCacheManager.ts
```diff
@@ -286,6 +340,144 @@ export class BrowserCacheManager extends CacheManager {
...
+                }
+                break;
+            case CredentialType.REFRESH_TOKEN:
+                if (tokenKeys.refreshToken.indexOf(key) === -1) {
+                    this.logger.info("BrowserCacheManager: addTokenKey - refreshToken added to map");
+                    tokenKeys.refreshToken.push(key);
+                }
+                break;
+            default:
+                this.logger.error("BrowserCacheManager:addTokenKey - CredentialType provided invalid. Doing nothing.");
```
peterzenz: If we hit this, it's a programmer error, we should likely throw here.

lib/msal-browser/src/cache/BrowserCacheManager.ts
```diff
@@ -286,6 +340,144 @@ export class BrowserCacheManager extends CacheManager {
...
+                const refreshRemoval = tokenKeys.refreshToken.indexOf(key);
+                if (refreshRemoval > -1) {
+                    this.logger.info("BrowserCacheManager: removeTokenKey - refreshToken removed from map");
+                    tokenKeys.idToken.splice(refreshRemoval, 1);
+                } else {
+                    this.logger.info("BrowserCacheManager: removeTokenKey - refreshToken does not exist in map. Either it was previously removed or it was never added.");
+                }
+                break;
+            default:
+                this.logger.error("BrowserCacheManager:addTokenKey - CredentialType provided invalid. Doing nothing.");
```
peterzenz: Copy/paste error on `addTokenKey` in the logger call.  Also, like the above this is likely a programmer error and should throw here.

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -364,156 +380,110 @@ export abstract class CacheManager implements ICacheManager {
...
-        requestedClaimsHash?: string
-    ): CredentialCache {
-        const allCacheKeys = this.getKeys();
-        const matchingCredentials: CredentialCache = {
-            idTokens: {},
-            accessTokens: {},
-            refreshTokens: {},
-        };
+    isCredentialKey(key: string): boolean {
+        if (key.split(Separators.CACHE_KEY_SEPARATOR).length < 6) {
```
peterzenz, line 387: We discussed this one earlier, our browser cache doesn't need to be coherent with the shared MSAL caches as they can't be shared.  Do we want to change this to be something else?  Maybe '~' or '^' instead of '-'?

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -364,156 +380,110 @@ export abstract class CacheManager implements ICacheManager {
...
 
-            /*
-             * homeAccountId can undefined, and we want to filter out cached items that have a homeAccountId of ""
-             * because we don't want a client_credential request to return a cached token that has a homeAccountId
-             */
-            if ((typeof homeAccountId === "string") && !this.matchHomeAccountId(entity, homeAccountId)) {
-                return;
-            }
+        /*
+         * homeAccountId can undefined, and we want to filter out cached items that have a homeAccountId of ""
```
peterzenz: nit:  can be undefined

peterzenz (approved):
Would be great to make sure that we have telemetry that covers weird/unexpected edge cases:
- Specific data we've serialized into the cache does not deserialize
- We have a token we can't generate a proper cache key for, an expectation for specific data being present from a server artifact is absent.

lib/msal-common/src/cache/CacheManager.ts
tnorling: I definitely think we should discuss changes to the cache schema, but I think that discussion should happen outside the context of this PR and should not be limited to just the separators

lib/msal-browser/src/cache/BrowserCacheManager.ts
tnorling: We had agreed that we would not be adding telemetry for the migration code. Did that stance change?

lib/msal-browser/src/cache/BrowserCacheManager.ts
tnorling: Thanks, addressed!

lib/msal-browser/src/cache/BrowserCacheManager.ts
tnorling: Addressed!

sameerag (approved):

lib/msal-browser/src/cache/BrowserCacheManager.ts
tnorling: CacheManagers are not set up to log telemetry yet and adding it would involve significant changes that are beyond the scope of this PR so I'll open a separate work item to track that part. Let me know what you think regarding prioritization

PR #5817, opened 2023-03-23
Run 1p build on 3P PRs

tnorling:
Adds pipeline to build/test 1P library against 3P changes


.pipelines/1p-build.yml
```diff
@@ -0,0 +1,48 @@
+jobs:
+  - job: Build_1P
+    displayName: 'Build with 1P library'
+    pool:
+      vmImage: 'ubuntu-latest'
```
peterzenz, line 10: Why ubuntu here?  The azure piplines just below is on Windows.  Does it matter?

peterzenz (approved):

.pipelines/1p-build.yml
tnorling: Consistency with all our other build pipelines. Don't think it matters except for msal-node-extensions which we need to build for all platforms

konstantin-msft (approved):

PR #5818, opened 2023-03-23
Migrate from tsdx to use rollup and jest directly for msal-node

hectormmg:
This PR:
- Removes `tsdx` from `msal-node` dependencies
- Replaces `msal-node` tsdx build and test scripts with their rollup and jest equivalents
- Adds `rollup.config.js` to `msal-node`
- Removes Node v14 from the MSAL Node test matrix
- Uses jest requireActual to properly load symlinked msal-common


lib/msal-node/rollup.config.js
```diff
@@ -0,0 +1,71 @@
...
+        plugins: [
+            typescript( {
+                typescript: require("typescript"),
+                tsconfig: "tsconfig.build.json"
+            }),
+            nodeResolve()
+        ]
+    },
+    {
+        // for cjs build
```
hectormmg:
```suggestion
        // for esm build
```

lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -163,7 +163,7 @@ export abstract class ClientApplication {
             if (e instanceof AuthError) {
                 e.setCorrelationId(validRequest.correlationId);
             }
-            serverTelemetryManager.cacheFailedRequest(e);
+            serverTelemetryManager.cacheFailedRequest(e as AuthError);
```
konstantin-msft: `serverTelemetryManager.cacheFailedRequest()` accepts any types of error. You don't need to cast it explicitly https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/msal-browser-v3/lib/msal-common/src/telemetry/server/ServerTelemetryManager.ts#L82.

konstantin-msft (approved):

lib/msal-node/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@azure/msal-node",
-  "version": "1.15.0",
+  "version": "1.16.0",
```
tnorling: Why was this bumped?

lib/msal-node/test/client/PublicClientApplication.spec.ts
```diff
@@ -524,7 +524,7 @@ describe('PublicClientApplication', () => {
         const logger = new Logger({
             loggerCallback: (level, message, containsPii) => {
                 expect(message).toContain("Message");
-                expect(message).toContain(LogLevel.Info);
+                expect(message).toContain("Info");
```
tnorling: Why was this changed?

hectormmg: I don't know why it was passing before, but under current config LogLevel.Info evaluates to the number 2 corresponding to its enum index, not to a string. It's not present in the log string. Unless I'm missing something, this is a fix.

tnorling: Ah interesting. Thanks!

tnorling (approved):

lib/msal-node/src/client/ClientApplication.ts
hectormmg: Thanks, updated these to stop type casting unsafely.

lib/msal-node/package.json
hectormmg: Undid the change, no idea how it got bumped.

sameerag (approved):

lib/msal-node/test/client/PublicClientApplication.spec.ts
```diff
@@ -524,7 +526,7 @@ describe('PublicClientApplication', () => {
         const logger = new Logger({
             loggerCallback: (level, message, containsPii) => {
                 expect(message).toContain("Message");
-                expect(message).toContain(LogLevel.Info);
+                expect(message).toContain(LogLevel.Info.toString());
```
Robbie-Microsoft, line 529: Why is this needed here, but not on the next line?

hectormmg: Because the level is a number. These two lines should never have passed at the same time.

PR #5824, opened 2023-03-30
initial add of office addin sample

shoatman:
- See the included readme for information on where the sample comes from and how to run it
- Not sure if or when we'll be able to use this sample for end to end testing give that it will need Office to execute, but want the sample to be available for manual testing for now


samples/msal-react-samples/office-addin-sample/readme.md
derisen, line 1: Can we ensure the filename is in capitals for consistency with the rest of the samples? The path portion in a url is case sensitive, and this leads to broken links if we ever link this file from somewhere in the same manner as others.

samples/msal-react-samples/office-addin-sample/readme.md
```diff
@@ -0,0 +1,74 @@
...
+## How to run the sample
+
+### Pre-requisites
+
+- Ensure [all pre-requisites](../../../lib/msal-react/README.md#prerequisites) have been completed to run `@azure/msal-react`.
+- Install node.js if needed (<https://nodejs.org/en/>).
+
+### Configure the application
+
+- Open `./src/authConfig.js` in an editor.
```
derisen, line 37: Seems like we haven't added auth logic to the sample yet -should this PR be a draft instead?

shoatman: The sample isn't ready to do anything yet, but i'd like to add what the yeoman output gives us... so that we can clearly understand the delta between the 2 things.

PR #5832, opened 2023-04-01
MSAL Angular v3 docs

jo-arroyo:


jo-arroyo:
Closing in favor of #5896

PR #5855, opened 2023-04-04
Unit tests for NativeBrokerPlugin (msal-node-extensions)

tnorling:
- Adds unit tests for the NativeBrokerPlugin implementation in msal-node-extensions
- Fixes incorrect expiresOn property


extensions/msal-node-extensions/test/broker/NativeBrokerPlugin.spec.ts
```diff
@@ -0,0 +1,1320 @@
...
+            const nativeBrokerPlugin = new NativeBrokerPlugin();
+            const request: NativeRequest = {
+                clientId: TEST_CLIENT_ID,
+                scopes: testAuthenticationResult.scopes,
+                correlationId: testCorrelationId,
+                authority: testAuthenticationResult.authority,
+                redirectUri: TEST_REDIRECTURI,
+                prompt: PromptValue.NONE
+            }
+            const response = await nativeBrokerPlugin.acquireTokenInteractive(request);
```
lalimasharda, line 494: Just want to make sure here that when prompt = none, acquireTokenInteractive request falls back to SignInSilentlyAsync request? Because the test description says "Calls SignInSilentlyAsync".

lalimasharda (approved):
LGTM

extensions/msal-node-extensions/src/broker/NativeBrokerPlugin.ts
```diff
@@ -356,7 +356,7 @@ export class NativeBrokerPlugin implements INativeBrokerPlugin {
             idTokenClaims: idTokenClaims,
             accessToken: authResult.accessToken,
             fromCache: fromCache,
-            expiresOn: new Date(authResult.expiresOn * 1000),
+            expiresOn: new Date(authResult.expiresOn),
```
sameerag, line 359: Don't we always send ms? I noticed we changed this recently but do we need to document this?

tnorling: No, we return a Date object, I'm not aware that this has changed.

extensions/msal-node-extensions/test/broker/NativeBrokerPlugin.spec.ts
tnorling: Yes. If prompt=none then we call either SignInSilentlyAsync or AcquireTokenSilentlyAsync depending on whether or not an account was included in the request.

peterzenz (approved):

PR #5862, opened 2023-04-06
Check app init concurrency

konstantin-msft:


lib/msal-browser/test/app/PublicClientApplication.spec.ts
```diff
@@ -136,6 +136,75 @@ describe("PublicClientApplication.ts Class Unit Tests", () => {
     });
 
     describe("intialize tests", () => {
+        jest.setTimeout(9999999);
```
tnorling: This is a pretty high timeout, anything smaller we can use?

lib/msal-browser/test/app/PublicClientApplication.spec.ts
```diff
@@ -136,6 +136,75 @@ describe("PublicClientApplication.ts Class Unit Tests", () => {
...
+        it("serializes concurrent calls", async () => {
+            const config = {
+                auth: {
+                    clientId: TEST_CONFIG.MSAL_CLIENT_ID
+                },
+                system: {
+                    allowNativeBroker: true
+                }
+            };
+            const samples = 100;
```
tnorling: Why do we need to run this test 100 times?

lib/msal-browser/test/app/PublicClientApplication.spec.ts
```diff
@@ -136,6 +136,75 @@ describe("PublicClientApplication.ts Class Unit Tests", () => {
...
+                const handshakeSpy: sinon.SinonSpy = sinon.spy(NativeMessageHandler.prototype, "sendHandshakeRequest");
+                const initEventSpy = sinon.stub(EventHandler.prototype, "emitEvent");
+                let mcPort: MessagePort;
+
+                const eventHandler = function (event: MessageEvent) {
+                    event.stopImmediatePropagation();
+                    const request = event.data;
+                    const req  = {
+                        channel: "53ee284d-920a-4b59-9d30-a60315b26836",
+                        extensionId: "test-ext-id",
```
tnorling: What about cases where the first handshake (to the specific extensionId) fails and the second handshake (without extensionId) is executed?

lib/msal-browser/test/app/PublicClientApplication.spec.ts
konstantin-msft: It's just a one-time test to verify that `app.init` calls are serialized by the event loop. I have no plans to merge it.

lib/msal-browser/test/app/PublicClientApplication.spec.ts
tnorling: Similarly, what's the behavior when one or more handshakes fail?

lib/msal-browser/test/app/PublicClientApplication.spec.ts
konstantin-msft: To make sure the serialization pattern is consistent.

lib/msal-browser/test/app/PublicClientApplication.spec.ts
```diff
@@ -136,6 +136,75 @@ describe("PublicClientApplication.ts Class Unit Tests", () => {
...
+                        }
+                    };
+
+                    mcPort = postMessageSpy.args[0][2][0];
+                    mcPort.postMessage(req);
+                };
+                window.addEventListener("message", eventHandler, true);
+
+                const promises = [];
+                for (let x = 1; x <= concurrency; x++) {
```
sameerag: nit: index to `0`?

fepimentel (approved):

PR #5865, opened 2023-04-10
Add CIAM authority support

sameerag:
**Add CIAM support in MSAL JS**

Azure Active Directory (Azure AD) is adding support to customer identity access management (CIAM) solution that lets you create secure, customized sign-in experiences for your customer-facing apps and services. With these built-in CIAM features, Azure AD can serve as the identity provider and access management service for your customer scenarios.

> The CIAM authority format is still in the process of standardization. It is determined to be split from AAD and currently supported in the below formats:

* `https://tenantName.ciamlogin.com/tenant` 
* `https://tenantName.ciamlogin.com`

> All MSALs have currently agreed to add two rules for CIAM Authority:

1. Add a new `CIAM` authority type
2. If the i/p CIAM  authority does not have a path, append `"tenant.onmicrosoft.com"` to the PATH as a temporary measure to support the STS endpoints. 

The second part of this support eventually will go away.

> Please also note that this is an emerging space and the endpoints may change eventually before productization. More details about CIAM can be found here: [CIAM previews](https://github.com/microsoft/entra-previews)

**_Completed :_**
- Code changes: Added `AuthorityType: Ciam` and the path tranformation logic stated above
- Unit tests
- Docs updates
- Samples tested

**_Coming soon :_**
- e2e tests will be in a separate PR


lib/msal-common/src/authority/AuthorityFactory.ts
```diff
@@ -36,10 +36,12 @@ export class AuthorityFactory {
         correlationId?: string
     ): Promise<Authority> {
         performanceClient?.addQueueMeasurement(PerformanceEvents.AuthorityFactoryCreateDiscoveredInstance, correlationId);
+        
+        const authorityUriFinal = Authority.transformCIAMAuthority(authorityUri);
```
tnorling, line 40: Can we do this in the authority constructor/instance methods instead?

lib/msal-common/src/authority/Authority.ts
```diff
@@ -768,5 +783,29 @@ export class Authority {
 
         return metadata;
     }
-}
 
+    /**
+     * Transform CIAM_AUTHORIY as per the below rules:
+     * If no path segments found and it is a CIAM authority, then transform it
+     * hostname ends with .ciamlogin.com (a constant of-course) and any Path => do not transform
+     * hostname ends with .ciamlogin.com (a constant of-course) and noPath  => transform
```
tnorling: nit to make this a little less verbose:

```suggestion
     * If no path segments found and it is a CIAM authority (hostname ends with .ciamlogin.com), then transform it
```

lib/msal-common/src/authority/AuthorityFactory.ts
sameerag: Guess we can. I chose this as most of the `authority` instance methods are getters and setters and having this code here made it more readable IMHO than in the other class. Let me wait for a few more reviews and make a call.

lib/msal-common/src/authority/Authority.ts
sameerag: Sorry, was meaning to fix this, pasted verbatim for a discussion as a place holder :). Committed the suggestion. Doc changes and samples are in progress for end users.

lib/msal-common/src/authority/AuthorityFactory.ts
tnorling: The argument for doing it in `Authority` is that we may not always go through the `AuthorityFactory` in which case this would be missed. I'll leave it up to you but it feels safer to do important things like this in the `Authority` class itself.

tnorling:
Changefile?

lib/msal-common/src/authority/AuthorityFactory.ts
sameerag: I expect this to go away. The transition is temporary until STS supports the standard format without needing to append anything. So leaving as is for now.

lib/msal-common/src/authority/Authority.ts
```diff
@@ -79,6 +79,12 @@ export class Authority {
 
     // See above for AuthorityType
     public get authorityType(): AuthorityType {
+
+        // CIAM auth pattern is being standardized as: tenant.ciamlogin.com
+        if(this.canonicalAuthorityUrlComponents.HostNameAndPort.endsWith(Constants.CIAM_AUTH_URL)) {
```
derisen:
```suggestion
        // CIAM auth url pattern is being standardized as: <tenant>.ciamlogin.com
        if (this.canonicalAuthorityUrlComponents.HostNameAndPort.endsWith(Constants.CIAM_AUTH_URL)) {
```

lib/msal-common/src/authority/Authority.ts
```diff
@@ -510,6 +521,13 @@ export class Authority {
      * Parse cloudDiscoveryMetadata config or check knownAuthorities
      */
     private getCloudDiscoveryMetadataFromConfig(): CloudDiscoveryMetadata | null {
+
+        // CIAM does not support cloud discovery metadata
+        if(this.authorityType === AuthorityType.Ciam) {
```
derisen:
```suggestion
        if (this.authorityType === AuthorityType.Ciam) {
```

lib/msal-common/src/authority/Authority.ts
```diff
@@ -768,5 +786,27 @@ export class Authority {
...
+     * 
+     * @param authority 
+     */
+    static transformCIAMAuthority(authority: string): string {
+        let ciamAuthority = authority;
+        const authorityUrl = new UrlString(authority);
+        const authorityUrlComponents = authorityUrl.getUrlComponents();
+
+        // check if transformation is needed
+        if(authorityUrlComponents.PathSegments.length === 0 && (authorityUrlComponents.HostNameAndPort.endsWith(Constants.CIAM_AUTH_URL))){
```
derisen:
```suggestion
        if (authorityUrlComponents.PathSegments.length === 0 && (authorityUrlComponents.HostNameAndPort.endsWith(Constants.CIAM_AUTH_URL))){
```

lib/msal-common/src/utils/Constants.ts
```diff
@@ -17,6 +17,9 @@ export const Constants = {
     DSTS: "dstsv2",
     // Default AAD Instance Discovery Endpoint
     AAD_INSTANCE_DISCOVERY_ENDPT: "https://login.microsoftonline.com/common/discovery/instance?api-version=1.1&authorization_endpoint=",
+    // CIAM URL
+    CIAM_AUTH_URL: ".ciamlogin.com",
+    CIAM_ENDPOINT: ".onmicrosoft.com",
```
derisen: > CIAM_ENDPOINT: ".onmicrosoft.com",

super nit: maybe something like `AAD_TENANT_DOMAIN_SUFFIX` would be more accurate

lib/msal-common/test/authority/AuthorityFactory.spec.ts
```diff
@@ -133,4 +133,40 @@ describe("AuthorityFactory.ts Class Unit Tests", () => {
...
+    it("createDiscoveredInstance does not transform when there is a PATH", async () => {
+        const resolveEndpointsStub = jest.spyOn(Authority.prototype, "resolveEndpointsAsync").mockResolvedValue();
+        const authorityInstance = await AuthorityFactory.createDiscoveredInstance("https://test.ciamlogin.com/tenant/", networkInterface, mockStorage, authorityOptions, logger);
+        expect(authorityInstance.authorityType).toBe(AuthorityType.Ciam);
+        expect(authorityInstance.canonicalAuthority).toBe("https://test.ciamlogin.com/tenant/");
+        expect(authorityInstance instanceof Authority);
+        expect(resolveEndpointsStub).toHaveBeenCalledTimes(1);
+    });
+
+    it("createDiscoveredInstance does not transform when there is a PATH & does not add a trailing `slash`", async () => {
```
derisen: is the test here aptly named? (seems like trailing slash is added)

lib/msal-common/test/authority/AuthorityFactory.spec.ts
```diff
@@ -133,4 +133,40 @@ describe("AuthorityFactory.ts Class Unit Tests", () => {
...
+        const resolveEndpointsStub = jest.spyOn(Authority.prototype, "resolveEndpointsAsync").mockResolvedValue();
+        const authorityInstance = await AuthorityFactory.createDiscoveredInstance("https://test.ciamlogin.com/tenant", networkInterface, mockStorage, authorityOptions, logger);
+        expect(authorityInstance.authorityType).toBe(AuthorityType.Ciam);
+        expect(authorityInstance.canonicalAuthority).toBe("https://test.ciamlogin.com/tenant/");
+        expect(authorityInstance instanceof Authority);
+        expect(resolveEndpointsStub).toHaveBeenCalledTimes(1);
+    })
+
+    it("createDiscoveredInstance transforms CIAM authority", async () => {
```
derisen: same comment here: seems like it does not transform authority here

derisen:
[.NET impl.](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4054) does not seem to allow setting CIAM authority at request level, though this doesn't seem to create an issue in here. Will test a bit more and confirm

derisen:
> [.NET impl.](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4054) does not seem to allow setting CIAM authority at request level, though this doesn't seem to create an issue in here. Will test a bit more and confirm

Had another go at this. There isn't an issue from the service side, but we end up with the same situation as in B2C user-flows: multiple set of entries in the cache, making it complicated to maintain the right account for UI and token requests (without a good use case it seems to me, unlike B2C). 

Should we align the SDKs here? cc @trwalke for more context

lib/msal-common/src/utils/Constants.ts
sameerag: Sure. Can change this.

lib/msal-common/test/authority/AuthorityFactory.spec.ts
sameerag: No, this is a miss, I was moving around tests and did not rename. Will change.

sameerag:
> > [.NET impl.](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4054) does not seem to allow setting CIAM authority at request level, though this doesn't seem to create an issue in here. Will test a bit more and confirm
> 
> Had another go at this. There isn't an issue from the service side, but we end up with the same situation as in B2C user-flows: multiple set of entries in the cache, making it complicated to maintain the right account for UI and token requests (without a good use case it seems to me, unlike B2C).
> 
> Should we align the SDKs here? cc @trwalke for more context

Sure. The reason why I have transformed it before and not later is that it is easy to clear this code once this is no longer needed. Can you please elaborate on multiple entries here? You mean all authorities of the format `*.ciamlogin.com/*` with the same `tenant` has to considered the same authority?

derisen:
> > > [.NET impl.](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4054) does not seem to allow setting CIAM authority at request level, though this doesn't seem to create an issue in here. Will test a bit more and confirm
> > 
> > 
> > Had another go at this. There isn't an issue from the service side, but we end up with the same situation as in B2C user-flows: multiple set of entries in the cache, making it complicated to maintain the right account for UI and token requests (without a good use case it seems to me, unlike B2C).
> > Should we align the SDKs here? cc @trwalke for more context
> 
> Sure. The reason why I have transformed it before and not later is that it is easy to clear this code once this is no longer needed. Can you please elaborate on multiple entries here? You mean all authorities of the format `*.ciamlogin.com/*` with the same `tenant` has to considered the same authority?

No I mean, if I sign-in with `authority1`, and then later acquire a token for `authority2`, I would end up with a set of entries (account and tokens) in the cache for each. This is normal behavior of our SDK, regardless of AAD, B2C or CIAM. But I'm not sure why [.NET doesn't allow overriding authority in the case of CIAM](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4054/files#r1161000570) -hopefully @trwalke can step in to elaborate.

sameerag:
> Changefile?

Added

sameerag:
> > > > [.NET impl.](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4054) does not seem to allow setting CIAM authority at request level, though this doesn't seem to create an issue in here. Will test a bit more and confirm
> > > 
> > > 
> > > Had another go at this. There isn't an issue from the service side, but we end up with the same situation as in B2C user-flows: multiple set of entries in the cache, making it complicated to maintain the right account for UI and token requests (without a good use case it seems to me, unlike B2C).
> > > Should we align the SDKs here? cc @trwalke for more context
> > 
> > 
> > Sure. The reason why I have transformed it before and not later is that it is easy to clear this code once this is no longer needed. Can you please elaborate on multiple entries here? You mean all authorities of the format `*.ciamlogin.com/*` with the same `tenant` has to considered the same authority?
> 
> No I mean, if I sign-in with `authority1`, and then later acquire a token for `authority2`, I would end up with a set of entries (account and tokens) in the cache for each. This is normal behavior of our SDK, regardless of AAD, B2C or CIAM. But I'm not sure why [.NET doesn't allow overriding authority in the case of CIAM](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4054/files#r1161000570) -hopefully @trwalke can step in to elaborate.

Discussed with @jmprieur and I believe we have it right. 

@jmprieur can you please review this too?

lib/msal-common/src/utils/Constants.ts
sameerag: Done

lib/msal-common/test/authority/AuthorityFactory.spec.ts
sameerag: Done

lib/msal-common/test/authority/AuthorityFactory.spec.ts
sameerag: Done

tnorling (approved):

samples/msal-browser-samples/VanillaJSTestApp2.0/app/ciam/index.html
```diff
@@ -0,0 +1,63 @@
...
+      </div>
+    </nav>
+    <br>
+    <h5 class="card-header text-center">Vanilla JavaScript SPA with MSAL.JS</h5>
+    <br>
+    <div class="row" style="margin:auto" >
+    <div id="card-div" class="col-md-3" style="display:none">
+    <div class="card text-center">
+      <div class="card-body">
+        <h5 class="card-title" id="WelcomeMessage">Welcome! You have successfully signed in with CIAM Auhtority</h5>
```
derisen:
```suggestion
<h5 class="card-title" id="WelcomeMessage">Please sign-in</h5>
```

derisen: forgot that this will be replaced with username once they sign-in

derisen (approved):
LGTM!

PR #5869, opened 2023-04-10
passport-azure-ad update node-jose dependency

tnorling:
Updates node-jose dependency to 2.2.0+ to address vulnerability with versions < 2.2.0


peterzenz (approved):
Thanks!

jo-arroyo (approved):

PR #5872, opened 2023-04-11
Updates to use prettier for code formatting

shoatman:
- added prettier to root of project (fixed version intentionally to avoid folks updating to new rules until we want to universally)
- added prettier to all projects.... looks like msal-react has a dependency which uses a older version of prettier, by explicitly declaring dependency ensure we get the correct version
- ran prettier using commands created below
- Made changes post apply fixed around (moving the comments as needed):
  - @ts-ignore
  - eslint-ignore-next-line / eslint-ignore-line
 - added format:check and format:fix scripts/commands to package.json of all updated projects
 - only applied to src / test folders (where present) and ignored all files ignored in .gitignore
 - updated the eslintrc file to remove things that conflict with the default rules of prettier... intentionally using the default rules
   - Some of the things I cut were no longer recommended to be there


.eslintrc.json
```diff
@@ -20,7 +20,8 @@
         "plugin:react/recommended",
         "plugin:react-hooks/recommended",
         "plugin:security/recommended",
-        "plugin:import/typescript"
+        "plugin:import/typescript",
+        "prettier"
```
tnorling, line 24: Does this mean that the prettier rules will be checked alongside eslint rules in CI?

lib/msal-angular/src/IMsalService.ts
```diff
@@ -4,26 +4,28 @@
...
-} from "@azure/msal-browser";
-import { Observable } from "rxjs";
+  EndSessionRequest,
+  AuthorizationUrlRequest,
+  AuthenticationResult,
+  PopupRequest,
+  RedirectRequest,
+  SilentRequest,
+  Logger,
+} from '@azure/msal-browser';
```
tnorling, line 14: Why the change from double to single quotes? That's going to be a ton of changes...

shoatman: Yes.  I didn't override any of the recommended settings from Prettier, mostly to avoid ever having to discuss the pros/cons of quote types in the future ;-)

.eslintrc.json
shoatman: I'll double check to see what the behavior is exactly.  We made need to add an explicit format check/step to the CI process.

lib/msal-angular/src/IMsalService.ts
derisen: Perhaps we can rely on the [TypeScript repo's style guide](https://github.com/microsoft/TypeScript/wiki/Coding-guidelines)?

tnorling:
> fixed version intentionally to avoid folks updating to new rules until we want to universally

My preference would be to set the rules we want explicitly and use ^ version rather than fixing the version and using defaults. What do other folks think?

lib/msal-angular/src/IMsalService.ts
shoatman: Here's the prettier teams opinion on options: https://prettier.io/docs/en/option-philosophy.html and Dan Abramov's issue referenced in that article: https://github.com/prettier/prettier/issues/40

shoatman:
> > fixed version intentionally to avoid folks updating to new rules until we want to universally
> 
> My preference would be to set the rules we want explicitly and use ^ version rather than fixing the version and using defaults. What do other folks think?

See the link on options and related issue above.

.eslintrc.json
shoatman: This is just to avoid the conflict.  Effectively ES Lint formatting errors are turned off.  Looks like we need to run prettier as a separate CI check if we want that.  Prettier does provide a package that will run prettier and then execute ES lint, but I'm be inclined to keep these separate.

shoatman:
@peterzenz , @tnorling , @sameerag - Let me know if you want to discuss this in a meeting rather than in the PR.  I'd really like to get this in today in order to unblock a much larger PR where I don't want to have to deal with eslint formatting errors.

.eslintrc.json
peterzenz: The prettier docs suggest a separation here to have linting rules not look at formatting, but to still check for code errors and just have prettier handle formatting/presentation.  https://prettier.io/docs/en/comparison.html

lib/msal-angular/src/IMsalService.ts
peterzenz: I mentioned in teams that this is likely worth a discussion.  I see where you're coming from here, but let's discuss the tradeoffs and any concerns if folks have any.

.eslintrc.json
tnorling: I think having style be enforced will be important to avoid the code churn/irrelevant changes we're facing today (bonus points if the workflow can just auto commit a fix instead of failing). I don't have a strong preference about whether it's done as a separate step or as part of ES Lint.

lib/msal-angular/src/IMsalService.ts
tnorling: Sounds good, my concern is just that keeping dependencies up to date is a constant struggle and pinning dependencies makes that harder. I'm ok to try this and re-evaluate later if it becomes burdensome.

peterzenz (approved):

Robbie-Microsoft:
> @peterzenz , @tnorling , @sameerag - Let me know if you want to discuss this in a meeting rather than in the PR. I'd really like to get this in today in order to unblock a much larger PR where I don't want to have to deal with eslint formatting errors.

I'd like to listen in to a meeting regarding this, if one materializes.

.eslintrc.json
shoatman: I'll open a second PR to make the pipeline changes.  Will definitely auto-fix; however, with tests that's likely to break something (@ts-ignore) statements need to be on the line above the offending bit of code.

tnorling (approved):

shoatman:
@tnorling @sameerag - Possible to get one more approval?

Thanks

shane

shoatman:
Thanks @peterzenz @tnorling !  Very much appreciated!

konstantin-msft (approved):

PR #5875, opened 2023-04-11
fix: fix lexical declaration in case clause, type of acquireToken, ad…

Dmytro-Lukashenko:
Hello!
I have searched the [issues](https://github.com/AzureAD/microsoft-authentication-library-for-js/issues) of this repository and believe that this is not a duplicate.

Version 
@azure/msal-browser
2.30.0

Environment
vue
3.2.37

1. plugins/msalPlugin.ts
String 63: There is a lexical declaration in the "case" clause. It can be a problem because the lexical declaration is visible in the entire switch block but it only gets initialized when it is assigned, which will only happen if the case where it is defined is reached.
I suggest adding a block in this "case":

```
case EventType.ACQUIRE_TOKEN_FAILURE: {
                    const currentAccounts = msalInstance.getAllAccounts();
                    if (!accountArraysAreEqual(currentAccounts, state.accounts)) {
                        state.accounts = currentAccounts;
                    }
                    break;
                }
```
2. composition-api/useMsalAuthentication.ts
String 6: `acquireToken: Function`
The `Function` type accepts any function-like value
I suggest explicitly define the function shape
`acquireToken: (requestOverride?: PopupRequest|RedirectRequest|SilentRequest) => Promise<void>;`

3. composition-api/useMsalAuthentication.ts
String 68:  ` if (!result && !error) {`
`result` and `error` are Refs, and to access the values we should use `result.value, error.value`
I suggest: `if (!result.value && !error.value) {`

Thanks! Happy coding!

### DIFF
--- a/samples/msal-browser-samples/vue3-sample-app/src/composition-api/useMsalAuthentication.ts
+++ b/samples/msal-browser-samples/vue3-sample-app/src/composition-api/useMsalAuthentication.ts
@@ -3,7 +3,7 @@ import { Ref, ref, watch } from "vue";
 import { useMsal } from "./useMsal";
 
 export type MsalAuthenticationResult = {
-    acquireToken: Function;
+    acquireToken: (requestOverride?: PopupRequest|RedirectRequest|SilentRequest) => Promise<void>;
     result: Ref<AuthenticationResult|null>;
     error: Ref<AuthError|null>;
     inProgress: Ref<boolean>;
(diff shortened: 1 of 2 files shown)


Dmytro-Lukashenko:
> 

Dmytro-Lukashenko:
It was my mistake to commit to the dev branch. I'll close this pull request, create a new branch from dev and then create a pull request from the new branch.

PR #5883, opened 2023-04-12
Adding nested app auth support

shoatman:
- Feature is opt-in via optional configuration setting, defaults to false
- Rollup config needed to be modified to support dynamic imports for UMD output (I'm really using them now)
- jest.config was updated to run a BridgeSetup script prior to running the unit tests in order to provide a mock implementation of the bridge (test/utils/bridgesetup)
- Just addressing the basic scenarios for Bearer Tokens at the moment
- Added a new error type specific to nested app auth to indicate when something is unsupported in that context
- Added a bridge proxy class for communication with the bridge


shoatman:
@peterzenz - I don't think it's possible or adviseable to provide everything in one PR.  This provides the implementation of the core and some tests for the same.  I'll add a sample project based on the Office Addin Template and try to figure out what if anything that can be done for end-to-end testing.  

@iambmelt - Please take a look at the NestedAppAuthController implementation as this will illustrate what is and what is not supported when operating in the presence of the bridge when the user has opted into it.

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
```diff
@@ -0,0 +1,198 @@
...
+                request.correlationId !== undefined
+                    ? request.correlationId
+                    : "",
+            prompt: request.prompt !== undefined ? request.prompt : "",
+            claims: request.claims !== undefined ? request.claims : "",
+            authenticationScheme:
+                request.authenticationScheme !== undefined
+                    ? request.authenticationScheme
+                    : "",
+            clientCapabilities: undefined,
```
iambmelt: For this, I think we pass the value provided by the caller [of MSAL.js], right? Doesn't the embedded app declare what its capabilities are?

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
```diff
@@ -0,0 +1,198 @@
...
+
+        return account;
+    }
+
+    /**
+     *
+     * @param error BridgeError
+     * @returns AuthError, ClientAuthError, ClientConfigurationError, ServerError, InteractionRequiredError
+     */
+    public fromBridgeError(
```
iambmelt, line 182: TODO - Need to review these error transformations with somebody from the MSAL.js team to see if they can live this

lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -0,0 +1,396 @@
...
+        throw NestedAppAuthError.createUnsupportedError();
+    }
+    acquireTokenByRefreshToken(
+        commonRequest: CommonSilentFlowRequest, // eslint-disable-line @typescript-eslint/no-unused-vars
+        silentRequest: SilentRequest // eslint-disable-line @typescript-eslint/no-unused-vars
+    ): Promise<AuthenticationResult> {
+        throw NestedAppAuthError.createUnsupportedError();
+    }
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    addEventCallback(callback: Function): string | null {
```
iambmelt: @tnorling IIRC these are necessary to support Angular/React -- what do these frameworks except us to do here?

lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -0,0 +1,396 @@
...
+    getActiveAccount(): AccountInfo | null {
+        throw NestedAppAuthError.createUnsupportedError();
+    }
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    initializeWrapperLibrary(sku: WrapperSKU, version: string): void {
+        // Do nothing for this for now
+        return;
+    }
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    setNavigationClient(navigationClient: INavigationClient): void {
```
iambmelt, line 429: Open question on what this should do...

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
shoatman: I think that's what we will have to do... since we don't have the declarative model yet.

iambmelt:
@shoatman I didn't quite finish looking through this yesterday, I'll pick this back up today with spare cycles

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
shoatman: Done.

lib/msal-browser/src/controllers/NestedAppAuthController.ts
tnorling: There's code in each of the acquireToken functions which emits "events" (such as Login started, succeeded, failed, etc). Essentially it's invoking any callbacks which have been registered using this API. You can look at the standard implementation to make more sense of it.

lib/msal-browser/src/controllers/NestedAppAuthController.ts
tnorling: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/navigation.md

lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -0,0 +1,435 @@
...
+        logoutRequest?: EndSessionRequest | undefined // eslint-disable-line @typescript-eslint/no-unused-vars
+    ): Promise<void> {
+        throw NestedAppAuthError.createUnsupportedError();
+    }
+    logoutPopup(
+        logoutRequest?: EndSessionPopupRequest | undefined // eslint-disable-line @typescript-eslint/no-unused-vars
+    ): Promise<void> {
+        throw NestedAppAuthError.createUnsupportedError();
+    }
+    ssoSilent(
```
iambmelt, line 370: I understand why this isn't needed for NAA, but I am wondering how we should handle this since developers in browsers will have to call this. How do we prevent them from having to conditionally fork/catch this?

lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -0,0 +1,435 @@
...
+    }
+    getLogger(): Logger {
+        throw NestedAppAuthError.createUnsupportedError();
+    }
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    setLogger(logger: Logger): void {
+        throw NestedAppAuthError.createUnsupportedError();
+    }
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    setActiveAccount(account: AccountInfo | null): void {
```
iambmelt, line 405: Something I am thinking about - if this function is called by the Hub in some cases, and by the app developer when running as a standalone page, could it make sense to have this function do nothing and log a warning, rather than `throw` in the nested case? Thinking out loud, a developer would have to fork their code (if/else) to avoid calling this if they detect they are embedded in a rich client.

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -259,6 +263,7 @@ export function buildConfiguration(
             tenant: Constants.EMPTY_STRING,
         },
         skipAuthorityMetadataCache: false,
+        supportsNestedAppAuth: false,
```
codexeon, line 266: This parameter works as an opt-in if brk-multihub is the only possible broker. If we want to allow the possibility of other brokers, an option would be supportedNestedAppAuthBrokers or trustedNestedAppAuthBrokers that would take an array of values and use that to opt-in: ["brk-<client_id_here>", "brk-multihub"]

lib/msal-browser/src/controllers/ControllerFactory.ts
```diff
@@ -27,20 +27,21 @@ export class ControllerFactory {
...
-        const metaOS = new TeamsAppOperatingContext(this.config);
+        const teamsApp = new TeamsAppOperatingContext(this.config);
 
-        const operatingContexts = [standard.initialize(), metaOS.initialize()];
+        const operatingContexts = [
+            standard.initialize(),
+            teamsApp.initialize(),
+        ];
 
         return Promise.all(operatingContexts).then(async () => {
```
codexeon, line 37: Prefer await here instead of then.

samples/msal-browser-samples/OfficeAddin/src/taskpane/taskpane.js
```diff
@@ -0,0 +1,89 @@
...
+
+const loginRequest = {
+  scopes: ["User.Read"]
+};
+
+let pca = undefined;
+
+PublicClientApplication.createPublicClientApplication(msalConfig).then((result) => {
+  pca = result;
+  ssoGetToken();
```
codexeon, line 50: For the test app could you add a button that tries to acquire the token. It would be helpful if it could call both acquireTokenSilent and acquireTokenPopup with scopes specified by user.

samples/msal-browser-samples/OfficeAddin/src/taskpane/taskpane.js
```diff
@@ -0,0 +1,89 @@
...
+  //scope: "User.Read",
+  //redirectUri: "https://localhost:3000",
+  const activeAccount = pca.getActiveAccount();
+
+  const tokenRequest = {
+    scopes: ["User.Read"],
+    account: activeAccount
+  };
+
+  pca.acquireTokenSilent(tokenRequest).then(async (result) => {
```
codexeon, line 64: change to await

samples/msal-browser-samples/OfficeAddin/src/taskpane/taskpane.js
```diff
@@ -0,0 +1,89 @@
...
+/* global document, Office, Word */
+
+Office.onReady((info) => {
+  if (info.host === Office.HostType.Word) {
+    document.getElementById("sideload-msg").style.display = "none";
+    document.getElementById("app-body").style.display = "flex";
+  }
+});
+
+export async function run() {
```
codexeon, line 17: Probably not used in this sample and could be removed

samples/msal-browser-samples/OfficeAddin/src/taskpane/taskpane.js
```diff
@@ -0,0 +1,89 @@
...
+PublicClientApplication.createPublicClientApplication(msalConfig).then((result) => {
+  pca = result;
+  ssoGetToken();
+});
+
+export async function ssoGetToken(){
+  //clientId: "a076930c-cfc9-4ebd-9607-7963bccbf666", //My msdn dev tenant
+  //scope: "User.Read",
+  //redirectUri: "https://localhost:3000",
+  const activeAccount = pca.getActiveAccount();
```
codexeon, line 57: The sample should ideally work in embedded context when nested app auth isn't available.

samples/msal-browser-samples/OfficeAddin/src/commands/commands.js
```diff
@@ -0,0 +1,44 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
+ * See LICENSE in the project root for license information.
+ */
+
+/* global global, Office, self, window */
+
+Office.onReady(() => {
```
codexeon, line 8: This file probably not necessary and could be removed, we can just have the taskpane.

samples/msal-browser-samples/OfficeAddin/manifest.xml
```diff
@@ -0,0 +1,85 @@
...
+    </Hosts>
+    <Resources>
+      <bt:Images>
+        <bt:Image id="Icon.16x16" DefaultValue="https://localhost:3000/assets/icon-16.png"/>
+        <bt:Image id="Icon.32x32" DefaultValue="https://localhost:3000/assets/icon-32.png"/>
+        <bt:Image id="Icon.80x80" DefaultValue="https://localhost:3000/assets/icon-80.png"/>
+      </bt:Images>
+      <bt:Urls>
+        <bt:Url id="GetStarted.LearnMoreUrl" DefaultValue="https://go.microsoft.com/fwlink/?LinkId=276812"/>
+        <bt:Url id="Commands.Url" DefaultValue="https://localhost:3000/commands.html"/>
```
codexeon, line 71: Probably could remove all references to commands.html and just support taskpane test app.

lib/msal-browser/test/utils/BridgeSetup.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    if (typeof window !== "undefined") {
+        /*
+         * JSDOM get's used except for tests that we're marked with node test environment
+         * that's done to simulate server side usage of the library
+         */
+        window.nestedAppAuthBridge = new MockBridge();
+    }
+});
+afterAll(() => {});
```
codexeon, line 25: undefine nestedAppAuthBridge after test.

lib/msal-browser/jest.config.js
```diff
@@ -13,5 +13,6 @@ module.exports = {
         url: "https://localhost:8081/index.html"
     },
     collectCoverageFrom: ["src/**/*.ts"],
-    coverageReporters: [["lcov", {"projectRoot": "../../"}]]
+    coverageReporters: [["lcov", {"projectRoot": "../../"}]],
+    setupFilesAfterEnv: ["./test/utils/BridgeSetup.ts"]
```
codexeon, line 17: should BridgeSetup just be used for NAA tests, so that it doesn't impact other test cases?

lib/msal-browser/src/naa/AccountInfo.ts
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+export type AccountInfo = {
```
codexeon, line 6: There is already an AccountInfo in msal-common, can we use that one or extend it if new properties are needed and provide a different name.

lib/msal-browser/src/naa/TokenRequest.ts
```diff
@@ -0,0 +1,26 @@
...
+    userObjectId?: string; // Account identifier used by OneAuth
+    clientId: string;
+    authority?: string;
+    scope: string;
+    correlationId: string;
+    prompt?: string; // Prompt used to identify interactive request
+    nonce?: string;
+    claims?: string;
+    state?: string;
+    reqCnf?: string; // Having OneAuth own the keypair is better for hardware token binding support
```
codexeon, line 16: Can we remove reqCnf and keyId since the expectation is OneAuth creates the key

lib/msal-browser/src/naa/TokenRequest.ts
```diff
@@ -0,0 +1,26 @@
...
+ * Licensed under the MIT License.
+ */
+
+export type TokenRequest = {
+    userObjectId?: string; // Account identifier used by OneAuth
+    clientId: string;
+    authority?: string;
+    scope: string;
+    correlationId: string;
+    prompt?: string; // Prompt used to identify interactive request
```
codexeon, line 12: the comment is no longer true, we have a different method for interactive requests.

lib/msal-browser/src/naa/AccountInfo.ts
codexeon: I see this is representing the response returned from native, probably prefixing with NestedAppAuth so that it isn't confused with the AccountInfo used everywhere else.

lib/msal-browser/src/naa/BridgeProxy.ts
```diff
@@ -0,0 +1,232 @@
...
+            throw new Error("window.crypto is undefined");
+        }
+
+        try {
+            BridgeProxy.crypto = window.crypto;
+
+            window.nestedAppAuthBridge.addEventListener(
+                "message",
+                (response: string) => {
+                    window.console.log(response);
```
codexeon, line 64: console.log needs to be removed, you can use tracePii if you want.

lib/msal-browser/src/naa/BridgeProxy.ts
```diff
@@ -0,0 +1,232 @@
...
+                    const responseEnvelope: BridgeResponseEnvelope =
+                        JSON.parse(response);
+                    const request = BridgeProxy.bridgeRequests.find(
+                        (element) =>
+                            element.requestId === responseEnvelope.requestId
+                    );
+                    if (request !== undefined) {
```
codexeon, line 71: if (request) is sufficient

lib/msal-browser/src/naa/BridgeProxy.ts
```diff
@@ -0,0 +1,232 @@
...
+}
+
+/**
+ * BridgeProxy
+ * Provides a proxy for accessing a bridge to a host app and/or
+ * platform broker
+ */
+export class BridgeProxy implements IBridgeProxy {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    static bridgeRequests: any[] = [];
```
codexeon, line 35: you may want to use a Map for performance, and for better helpers around lookup/remove/add.

lib/msal-browser/src/naa/BridgeProxy.ts
```diff
@@ -0,0 +1,232 @@
...
+                        if (responseEnvelope.success) {
+                            request.resolve(responseEnvelope.body);
+                        } else {
+                            request.reject(responseEnvelope.body);
+                        }
+                    }
+                }
+            );
+
+            const promise = new Promise<InitializeBridgeResponse>(
```
codexeon, line 85: instead of duplicating the code, can you reuse sendRequest to get init context.

lib/msal-browser/src/naa/BridgeProxy.ts
```diff
@@ -0,0 +1,232 @@
...
+            | TokenRequest
+            | AccountByHomeIdRequest
+            | AccountByLocalIdRequest
+            | AccountByUsernameRequest
+            | undefined
+    ): Promise<TResponse> {
+        const message: BridgeRequestEnvelope = {
+            messageType: "NestedAppAuthRequest",
+            method: method,
+            requestId: BridgeProxy.getRandomId(),
```
codexeon, line 183: I think a counter is sufficient here unless there is a need for a uuid or security string that can't be guessed. This is just used to link a request/response together.

lib/msal-browser/src/naa/BridgeProxy.ts
```diff
@@ -0,0 +1,232 @@
...
+            | AccountByLocalIdRequest
+            | AccountByUsernameRequest
+    ): Promise<AccountInfo> {
+        let method: BridgeMethods = "GetAccountByHomeId";
+
+        if ((request as AccountByHomeIdRequest).homeAccountId !== undefined) {
+            method = "GetAccountByHomeId";
+        }
+
+        if ((request as AccountByLocalIdRequest).localAccountId !== undefined) {
```
codexeon, line 151: Probably use elseif for these cases since only one type of request can be made.

lib/msal-browser/src/naa/BridgeProxy.ts
```diff
@@ -0,0 +1,232 @@
...
+                    BridgeProxy.bridgeRequests.push(request);
+                    window.nestedAppAuthBridge.postMessage(
+                        JSON.stringify(message)
+                    );
+                }
+            );
+
+            return promise;
+        } catch (error) {
+            window.console.log(error);
```
codexeon, line 107: remove the console.log

lib/msal-browser/src/naa/BridgeProxy.ts
codexeon: I guess there could be two MSAL instances on the same page, in that case it would need to be unique. Feel free to leave this with cryptographically random, but we really only care about uniqueness and performance.

lib/msal-browser/src/controllers/ControllerFactory.ts
```diff
@@ -27,20 +27,21 @@ export class ControllerFactory {
...
-                 * pull down metaos module
-                 * create associated controller
-                 */
-                // return await StandardController.createController(standard);
-                const controller = await import("./StandardController");
-                return await controller.StandardController.createController(
-                    standard
+            if (
+                teamsApp.isAvailable() &&
+                teamsApp.getConfig().auth.supportsNestedAppAuth
```
codexeon, line 40: is it necessary to initialize teamsApp if config.auth.supportsNestedAppAuth is false?

lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -0,0 +1,435 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    addEventCallback(callback: Function): string | null {
+        throw NestedAppAuthError.createUnsupportedError();
+    }
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    removeEventCallback(callbackId: string): void {
+        throw NestedAppAuthError.createUnsupportedError();
+    }
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    addPerformanceCallback(callback: PerformanceCallbackFunction): string {
```
codexeon, line 270: Should the performance callback be shared with standard controller? I didn't think it would have dependencies on nested app auth.

lib/msal-browser/src/naa/TokenRequest.ts
```diff
@@ -0,0 +1,26 @@
...
+    reqCnf?: string; // Having OneAuth own the keypair is better for hardware token binding support
+    keyId?: string; // Having OneAuth own the keypair is better for hardware token binding support
+    authenticationScheme?: string;
+    shrClaims?: string;
+    shrNonce?: string;
+    clientCapabilities?: string[]; // CP1 for CAE support
+    resourceRequestMethod?: string;
+    resourceRequestUri?: string;
+    extendedExpiryToken?: boolean;
+    extraParameters?: Map<string, string>;
```
codexeon, line 25: should this be StringDict instead of Map? This just gets serialized to JSON.

lib/msal-browser/src/naa/TokenRequest.ts
```diff
@@ -0,0 +1,26 @@
...
+    prompt?: string; // Prompt used to identify interactive request
+    nonce?: string;
+    claims?: string;
+    state?: string;
+    reqCnf?: string; // Having OneAuth own the keypair is better for hardware token binding support
+    keyId?: string; // Having OneAuth own the keypair is better for hardware token binding support
+    authenticationScheme?: string;
+    shrClaims?: string;
+    shrNonce?: string;
+    clientCapabilities?: string[]; // CP1 for CAE support
```
codexeon, line 21: I believe clientCapabilities eventually ends up as claims parameter to OneAuth. Is it possible to merge claims and clientCapabilities at the MSAL.js level?

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
```diff
@@ -0,0 +1,213 @@
...
+        const tokenRequest: TokenRequest = {
+            userObjectId: request.account?.homeAccountId,
+            clientId: this.clientId,
+            authority: request.authority,
+            scope: request.scopes.join(" "),
+            correlationId:
+                request.correlationId !== undefined
+                    ? request.correlationId
+                    : "",
+            prompt: request.prompt !== undefined ? request.prompt : "",
```
codexeon, line 68: prompt: request.prompt ?? ""

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
```diff
@@ -0,0 +1,213 @@
...
+        crypto: ICrypto,
+        logger: Logger
+    ) {
+        this.clientId = clientId;
+        this.clientCapabilities = clientCapabilities;
+        this.crypto = crypto;
+        this.logger = logger;
+    }
+
+    public toNaaSilentTokenRequest(request: SilentRequest): TokenRequest {
```
codexeon, line 46: do you need both toNaaSilentTokenRequest and toNaaTokenRequest, or can you build the request object based on optional fields included in the request? It may be a way to reduce the code.

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
```diff
@@ -0,0 +1,213 @@
...
+            uniqueId: response.account.homeAccountId,
+            tenantId: response.account.tenantId,
+            scopes: response.scope.split(" "),
+            account: this.fromNaaAccountInfo(response.account),
+            /*
+             * idToken: response.id_token,
+             * idTokenClaims: authToken.claims,
+             */
+            idToken: "",
+            idTokenClaims: {},
```
codexeon, line 146: likely you will need to parse the idTokenClaims from id token (once it is available). You could still add a call to do this now and unit test.

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
```diff
@@ -0,0 +1,213 @@
...
+            accessToken: response.access_token,
+            fromCache: true,
```
codexeon, line 148: fromCache always being true is not accurate. We may need to pass an additional field. It looks like NativeInteractionClient gets this from mats.

lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -0,0 +1,435 @@
...
+import { IController } from "./IController";
+import { TeamsAppOperatingContext } from "../operatingcontext/TeamsAppOperatingContext";
+import { IBridgeProxy } from "../naa/IBridgeProxy";
+import { BrowserPerformanceClient } from "../telemetry/BrowserPerformanceClient";
+import { version, name } from "../packageMetadata";
+import { CryptoOps } from "../crypto/CryptoOps";
+import { NestedAppAuthAdapter } from "../naa/mapping/NestedAppAuthAdapter";
+import { NestedAppAuthError } from "../error/NestedAppAuthError";
+
+export class NestedAppAuthController implements IController {
```
codexeon, line 42: Lets talk outside of this review about what to do when getAccount API is not available. Also for the clients that don't support nested app auth.

lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -0,0 +1,435 @@
...
+            response
+        );
+    }
+
+    private async acquireTokenSilentInternal(
+        request: SilentRequest
+    ): Promise<AuthenticationResult> {
+        const naaRequest =
+            this.nestedAppAuthAdapter.toNaaSilentTokenRequest(request);
+        const response = await this.bridgeProxy.getTokenSilent(naaRequest);
```
codexeon: is there a caching layer planned, a common pattern I see is developers calling getToken API every web request that needs a token, regardless if current token is still valid, and it will have a large overhead without a local cache in MSAL.js.

lib/msal-browser/src/controllers/ControllerFactory.ts
```diff
@@ -27,20 +27,21 @@ export class ControllerFactory {
 
     async createController(): Promise<IController> {
         const standard = new StandardOperatingContext(this.config);
-        const metaOS = new TeamsAppOperatingContext(this.config);
+        const teamsApp = new TeamsAppOperatingContext(this.config);
```
axhaferllari, line 30: Does this refer explicitly to Teams App?

lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -0,0 +1,501 @@
...
+                this.nestedAppAuthAdapter.fromNaaAccountInfo(currentAccount),
+            ];
+        } else {
+            return [];
+        }
+    }
+    handleRedirectPromise(
+        hash?: string | undefined // eslint-disable-line @typescript-eslint/no-unused-vars
+    ): Promise<AuthenticationResult | null> {
+        throw new Error("Method not implemented.");
```
iambmelt, line 341: Should this `throw NestedAppAuthError.createUnsupportedError();` ?

lib/msal-browser/src/controllers/NestedAppAuthController.ts
shoatman: logged a warning and ignored

lib/msal-browser/src/controllers/NestedAppAuthController.ts
shoatman: Converted to acquire token silent.

lib/msal-browser/src/controllers/NestedAppAuthController.ts
shoatman: logged a warning and did nothing.

lib/msal-browser/src/config/Configuration.ts
shoatman: @iambmelt - Thoughts on this one.  For now I plan to leave as is.

samples/msal-browser-samples/OfficeAddin/src/taskpane/taskpane.js
shoatman: Will do.  @iambmelt - Anything else you'd like to see in the test app?

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
shoatman: I think this is an interesting question.... not sure the code saving is worth it in this case.

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
shoatman: The code to parse exists already with tests, etc.... I'll just be calling existing utility methods... not sure if we have this yet though.

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
shoatman: @iambmelt - Can you track this one for Dan and I.  Currently I'm planning to leave this as true....

lib/msal-browser/src/controllers/NestedAppAuthController.ts
shoatman: @iambmelt - Can you track this one for Dan and I?

lib/msal-browser/src/controllers/NestedAppAuthController.ts
shoatman: @iambmelt - Suggest we add this to a backlog.

lib/msal-browser/src/controllers/ControllerFactory.ts
shoatman: For the moment this is just for Teams Apps.  It may be generalized in the future.

lib/msal-browser/src/controllers/NestedAppAuthController.ts
shoatman: Fixed.

samples/msal-browser-samples/OfficeAddin/src/taskpane/taskpane.js
iambmelt: > 

Off hand, I think it'd be good to illustrate:
- Handling if bridge is unavailable (e.g., fallback to auth popup)
- Handling of expired tokens
- Handling of CA/Intune challenges (e.g., MFA, maybe even claimsChallenge?)

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
iambmelt: ADO item is [here](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2617671)

lib/msal-browser/src/controllers/NestedAppAuthController.ts
iambmelt: Added [here](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2617680)

lib/msal-browser/src/controllers/NestedAppAuthController.ts
iambmelt: Added [here](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2617681)

lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -0,0 +1,501 @@
...
+        request: SilentRequest
+    ): Promise<AuthenticationResult> {
+        this.eventHandler.emitEvent(
+            EventType.ACQUIRE_TOKEN_START,
+            InteractionType.Silent,
+            request
+        );
+
+        try {
+            const naaRequest =
```
codexeon, line 179: In the StandardController getRequestCorrelationId is called when no correlation id is provided. I think we should do the same here, since it is still useful for telemetry to have a correlation id originate from MSAL.js.

lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -0,0 +1,501 @@
...
+                CommonAuthorizationUrlRequest,
+                | "requestedClaimsHash"
+                | "responseMode"
+                | "codeChallenge"
+                | "codeChallengeMethod"
+                | "nativeBroker"
+            >
+        >
+    ): Promise<AuthenticationResult> {
+        throw NestedAppAuthError.createUnsupportedError();
```
codexeon, line 383: Need to provide an implementation for ssoSilent, as it is currently the recommended API to call for MSAL v3 as there is no account object required.

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
```diff
@@ -0,0 +1,213 @@
...
+
+        return tokenRequest;
+    }
+
+    public fromNaaTokenResponse(
+        request: TokenRequest,
+        response: TokenResponse
+    ): AuthenticationResult {
+        const expiresOn = new Date(
+            TimeUtils.nowSeconds() + (response.expires_in || 0) * 1000
```
codexeon, line 125: Date constructor should be milliseconds.  (TimeUtils.nowSeconds() + (response.expires_in || 0)) * 1000

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
```diff
@@ -0,0 +1,213 @@
...
+        );
+
+        /*
+         * Currently we're not getting id_token back....
+         *let authToken: AuthToken;
+         *if (response.id_token !== undefined) {
+         *    authToken = new AuthToken(response.id_token, this.crypto);
+         *}
+         */
+        const authenticationResult: AuthenticationResult = {
```
codexeon, line 135: Should you populate fromNativeBroker: true

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
```diff
@@ -0,0 +1,213 @@
...
+             * idToken: response.id_token,
+             * idTokenClaims: authToken.claims,
+             */
+            idToken: "",
+            idTokenClaims: {},
+            accessToken: response.access_token,
+            fromCache: true,
+            expiresOn: expiresOn,
+            tokenType:
+                request.authenticationScheme !== undefined
```
codexeon, line 151: This should be tokenType: request.authenticationScheme ? request.authenticationScheme : "Bearer"
If authenticationScheme is undefined we should default to Bearer.

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
```diff
@@ -0,0 +1,213 @@
...
+
+        /*
+         * Currently we're not getting id_token back....
+         *let authToken: AuthToken;
+         *if (response.id_token !== undefined) {
+         *    authToken = new AuthToken(response.id_token, this.crypto);
+         *}
+         */
+        const authenticationResult: AuthenticationResult = {
+            authority: response.account.environment,
```
codexeon, line 136: environment is not the authority, it is host name and port of the preferred cache location. Authority should be in the format "https://login.microsoftonline.com/common/". Can you infer authority from the id token issuer, or does this need to be passed from native?

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
```diff
@@ -0,0 +1,213 @@
...
+    ): AuthenticationResult {
+        const expiresOn = new Date(
+            TimeUtils.nowSeconds() + (response.expires_in || 0) * 1000
+        );
+
+        /*
+         * Currently we're not getting id_token back....
+         *let authToken: AuthToken;
+         *if (response.id_token !== undefined) {
+         *    authToken = new AuthToken(response.id_token, this.crypto);
```
codexeon, line 132: should uncomment the id token code now that id token is included.

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
```diff
@@ -0,0 +1,213 @@
...
+         *}
+         */
+        const authenticationResult: AuthenticationResult = {
+            authority: response.account.environment,
+            uniqueId: response.account.homeAccountId,
+            tenantId: response.account.tenantId,
+            scopes: response.scope.split(" "),
+            account: this.fromNaaAccountInfo(response.account),
```
codexeon, line 140: instead of creating the account object based on native response, it will be more consistent to base it on the id token. AccountEntity.createAccount to create the account entity, and then accountEntity.getAccountInfo() to create the AccountInfo object.

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
```diff
@@ -0,0 +1,213 @@
...
+        /*
+         * Currently we're not getting id_token back....
+         *let authToken: AuthToken;
+         *if (response.id_token !== undefined) {
+         *    authToken = new AuthToken(response.id_token, this.crypto);
+         *}
+         */
+        const authenticationResult: AuthenticationResult = {
+            authority: response.account.environment,
+            uniqueId: response.account.homeAccountId,
```
codexeon, line 137: for uniqueId/tenantId can you try to read it from the id token if available for consistency.

lib/msal-browser/src/naa/TokenRequest.ts
codexeon: addClientCapabilitiesToClaims is how they handle merging client capabilities and claims for other msal-browser flows. We should do the same before sending the claims parameter to the host to make it easier for native hosts to handle.

lib/msal-browser/src/operatingcontext/TeamsAppOperatingContext.ts
```diff
@@ -33,16 +39,41 @@ export class TeamsAppOperatingContext extends BaseOperatingContext {
...
          */
-        return false;
+        try {
+            if (typeof window !== "undefined") {
+                const bridgeProxy: IBridgeProxy = await BridgeProxy.create();
+                /*
+                 * Because we want to get SSO we need to grab the active account as part of initialization
+                 * this.activeAccount = await bridgeProxy.getActiveAccount();
+                 */
+                this.activeAccount = await bridgeProxy.getActiveAccount();
```
codexeon, line 67: This call should be removed because initial implementation of getActiveAccount is not the full account object (it doesn't include id token or claims). Additionally, this call will fail to initialize TeamsAppOperatingContext if host does not implement GetActiveAccount.

lib/msal-browser/src/naa/BridgeProxy.ts
codexeon, line 85: should you implement timeout to make sure connection with host can be established, or is the responsibility of timeout in the bridge itself? When in the browser scenario if there is a postMessage dependency, you can expect that there may be some messages that aren't handled and usually a timeout is used with postMessage.

lib/msal-browser/src/naa/BridgeProxy.ts
```diff
@@ -0,0 +1,232 @@
...
+        }
+
+        try {
+            BridgeProxy.crypto = window.crypto;
+
+            window.nestedAppAuthBridge.addEventListener(
+                "message",
+                (response: string) => {
+                    window.console.log(response);
+                    const responseEnvelope: BridgeResponseEnvelope =
```
amehagh, line 65: 💡 I'd recommend that we add a `try - catch` around the `JSON.parse` since we're parsing a string and in the worst case, it can be an invalid JSON

lib/msal-browser/src/naa/BridgeProxy.ts
```diff
@@ -0,0 +1,232 @@
...
+                        JSON.parse(response);
+                    const request = BridgeProxy.bridgeRequests.find(
+                        (element) =>
+                            element.requestId === responseEnvelope.requestId
+                    );
+                    if (request !== undefined) {
+                        BridgeProxy.bridgeRequests.splice(
```
amehagh, line 72: If we follow @codexeon's comment to use the `Map`, we'll gain a lot by avoiding `splice`s

lib/msal-browser/src/naa/BridgeProxy.ts
amehagh, line 65: This is a force cast. Do we instead need to massage/map the response to _create_ a new responseEnvelope so that we can ensure that consumers get the response body they'd expect?

lib/msal-browser/src/naa/BridgeProxy.ts
```diff
@@ -0,0 +1,232 @@
...
+            | AccountByHomeIdRequest
+            | AccountByLocalIdRequest
+            | AccountByUsernameRequest
+            | undefined
+    ): Promise<TResponse> {
+        const message: BridgeRequestEnvelope = {
+            messageType: "NestedAppAuthRequest",
+            method: method,
+            requestId: BridgeProxy.getRandomId(),
+            apiKey: "test_key",
```
amehagh, line 184: Curious: Is this static? And why is it called "test" key

lib/msal-browser/src/naa/BridgeRequestEnvelope.ts
```diff
@@ -0,0 +1,47 @@
...
+    requestId: string;
+    apiKey?: string;
+    body?:
+        | TokenRequest
+        | AccountByHomeIdRequest
+        | AccountByLocalIdRequest
+        | AccountByUsernameRequest;
+};
+
+export function isBridgeRequestEnvelope(
```
amehagh, line 38: Nice 👍

lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -0,0 +1,501 @@
...
+                    response
+                );
+            this.eventHandler.emitEvent(
+                EventType.ACQUIRE_TOKEN_SUCCESS,
+                InteractionType.Silent,
+                result
+            );
+            return result;
+        } catch (e) {
+            this.eventHandler.emitEvent(
```
codexeon, line 195: This error is a BridgeError, it needs to be converted to MSAL.js error before passed back to developer. this.nestedAppAuthAdapter.fromBridgeError should be used somewhere in this path.

lib/msal-browser/src/naa/BridgeError.ts
```diff
@@ -0,0 +1,21 @@
...
+    code: string; // auth_flow_last_error such as invalid_grant
+    subError: string; // server_suberror_code such as consent_required
+    description: string;
+    properties: object; // additional telemetry info
+};
+
+export function isBridgeError(error: unknown): error is BridgeError {
+    return (
+        (error as BridgeError).status !== undefined &&
+        (error as BridgeError).code !== undefined
```
codexeon, line 19: Not all BridgeErrors may have a code. Is status enough to identify that it is a bridge error, or do you need another identifier? You also know it is a bridge error when it is rejected in the event listener, you could construct a strong-typed object for isBridgeError to detect.

lib/msal-browser/src/controllers/ControllerFactory.ts
codexeon, line 37: convert .then to await Promise.all(operatingContexts)

lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -0,0 +1,501 @@
...
+    // EventHandler
+    protected readonly eventHandler: EventHandler;
+
+    // NestedAppAuthAdapter
+    protected readonly nestedAppAuthAdapter: NestedAppAuthAdapter;
+
+    constructor(operatingContext: TeamsAppOperatingContext) {
+        this.operatingContext = operatingContext;
+        const proxy = this.operatingContext.getBridgeProxy();
+        if (proxy !== undefined) {
```
codexeon, line 70: if (proxy) is more common syntax for checking if an object exists.

lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -0,0 +1,501 @@
...
+        const controller = new NestedAppAuthController(operatingContext);
+        return Promise.resolve(controller);
+    }
+
+    initialize(): Promise<void> {
+        // do nothing not required by this controller
+        return Promise.resolve();
+    }
+
+    private async acquireTokenInteractive(
```
codexeon, line 132: for naa, silent and interactive are basically the same (the only difference is a string passed. Could you refactor for them to share majority of the code?

lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -0,0 +1,501 @@
...
+                    response
+                );
+            this.eventHandler.emitEvent(
+                EventType.ACQUIRE_TOKEN_SUCCESS,
+                InteractionType.Popup,
+                result
+            );
+            return result;
+        } catch (e) {
+            this.eventHandler.emitEvent(
```
codexeon, line 159: interactive has the same problem as silent with not converting bridge error to MSAL.js error.

lib/msal-browser/src/controllers/NestedAppAuthController.ts
codexeon: also ssoSilent could share majority of the implementation.

lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -0,0 +1,501 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    getAccountByHomeId(homeAccountId: string): AccountInfo | null {
+        const currentAccount = this.operatingContext.getActiveAccount();
+        if (currentAccount !== undefined) {
+            if (currentAccount.homeAccountId === homeAccountId) {
+                return this.nestedAppAuthAdapter.fromNaaAccountInfo(
+                    currentAccount
+                );
+            } else {
+                return null;
```
codexeon, line 292: instead of nested if/else and multiple paths returning null, just return null at the end if no account is found.

lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -0,0 +1,501 @@
...
+    handleRedirectPromise(
+        hash?: string | undefined // eslint-disable-line @typescript-eslint/no-unused-vars
+    ): Promise<AuthenticationResult | null> {
+        throw new Error("Method not implemented.");
+    }
+    loginPopup(
+        request?: PopupRequest | undefined // eslint-disable-line @typescript-eslint/no-unused-vars
+    ): Promise<AuthenticationResult> {
+        if (request !== undefined) {
+            return this.acquireTokenInteractive(request);
```
codexeon, line 347: loginPopup should set default scopes as a lot of the time it is called with no parameters. The same as StandardController: ...(request || DEFAULT_REQUEST)

lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -0,0 +1,501 @@
...
+        this.logger = logger;
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    setActiveAccount(account: AccountInfo | null): void {
+        /*
+         * StandardController uses this to allow the developer to set the active account
+         * in the nested app auth scenario the active account is controlled by the app hosting the nested app
+         */
+        return;
```
codexeon, line 410: since this.operatingContext.getActiveAccount() may not be populated, this could be called as a result of ssoSilent. We should store the account in-memory (it comes from the operating context of the host via token request).

lib/msal-browser/src/error/NestedAppAuthError.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import { AuthError } from "@azure/msal-common";
+
+/**
+ * NestedAppAuthErrorMessage class containing string constants used by error codes and messages.
+ */
+export const NestedAppAuthErrorMessage = {
+    unsupportedMethod: {
+        code: "unsupported_method",
+        desc: "The PKCE code challenge and verifier could not be generated.",
```
codexeon, line 14: the description does not seem to be related to nested app auth.

lib/msal-browser/src/naa/TokenResponse.ts
```diff
@@ -0,0 +1,61 @@
...
+};
+
+/*
+ * Sample response below
+ * {
+ * "body":{
+ *   "access_token":"<token>",
+ *      "account":{"environment":"login.microsoftonline.com",
+ *        "homeAccountId":"2995ae49-d9dd-409d-8d62-ba969ce58a81.51178b70-16cc-41b5-bef1-ae1808139065",
+ *        "idTokenClaims":{
```
codexeon, line 32: idTokenClaims will not be included in the response, they can be parsed from the id token

samples/msal-browser-samples/OfficeAddin/tsconfig.json
```diff
@@ -0,0 +1,26 @@
+{
+  "compilerOptions": {
```
codexeon, line 2: are there any typescript files being compiled? It looks like a pure JavaScript sample.

samples/msal-browser-samples/OfficeAddin/stats.json
codexeon, line 1: do you need to include this file? It contains information about your local build.

shoatman:
Closing as this was superseded a while a go.

PR #5891, opened 2023-04-14
Improve Iframe Errors

tnorling:
Improves error detection and logging when monitoring hidden iframes, providing a better understanding of why the request failed.

Previously any non-successful response was thrown as `monitor_window_timeout`
With these changes we'll also throw `hash_empty_error` or `hash_does_not_contain_known_properties` error if the application is manipulating the hash before we can read it.


lib/msal-browser/test/interaction_handler/SilentHandler.spec.ts
```diff
@@ -389,5 +397,81 @@ describe("SilentHandler.ts Unit Tests", () => {
...
+                defaultTokenRequest,
+                browserRequestLogger,
+                {
+                    navigateFrameWait: DEFAULT_IFRAME_TIMEOUT_MS,
+                    pollIntervalMilliseconds: DEFAULT_POLL_INTERVAL_MS,
+                },
+                performanceClient
+            );
+            // @ts-ignore
+            silentHandler.monitorIframeForHash(iframe, 1000).catch((e) => {
```
peterzenz, line 461: Would be great to use constants for the 1000 here and the 500 on line 474 just below.

peterzenz (approved):

hectormmg (approved):

sameerag (approved):
Lgtm. Agree with Peter on Constants.

PR #5893, opened 2023-04-14
msal-node-extension drop tsdx

tnorling:
TSDX is no longer maintained and there are security alerts open for some transitive dependencies. This PR drops TSDX in favor of using the underlying tools directly


peterzenz (approved):

extensions/msal-node-extensions/src/Dpapi.ts
```diff
@@ -7,6 +7,5 @@ export interface DpapiBindings{
     protectData(dataToEncrypt: Uint8Array, optionalEntropy: Uint8Array, scope: string): Uint8Array
     unprotectData(encryptData: Uint8Array, optionalEntropy: Uint8Array, scope: string): Uint8Array
 }
-/* eslint-disable-next-line @typescript-eslint/no-var-requires, no-var, import/no-commonjs */
-export var Dpapi: DpapiBindings = require("../build/Release/dpapi.node");
-export default Dpapi;
+/* eslint-disable-next-line @typescript-eslint/no-var-requires, import/no-commonjs */
+export const Dpapi: DpapiBindings = require("../build/Release/dpapi.node");
```
derisen, line 11: Curious why this was declared as `var` before. Was it to accomodate older node versions? (or to ensure its in the global scope since it's meant to be a "singleton"?)

derisen (approved):

hectormmg (approved):

PR #5895, opened 2023-04-14
msal-react drop tsdx

tnorling:
TSDX is no longer maintained and there are security alerts open for some transitive dependencies. This PR drops TSDX in favor of using the underlying tools directly


peterzenz (approved):

hectormmg (approved):
LGTM!

PR #5896, opened 2023-04-14
Msal Angular v3 initialization changes and doc updates

jo-arroyo:
This PR:
- adds initialization to `handleRedirectObservable` and updates unit tests
- removes initialization from `MsalRedirectComponent` and updates unit tests
- updates the msal-angular README
- adds a v2-v3 upgrade guide
- updates guidance surrounding initialization and handling redirects


lib/msal-angular/README.md
```diff
@@ -44,8 +45,9 @@ At a minimum, `@azure/msal-angular` will follow the [support schedule of the mai
 
 | MSAL Angular version | MSAL support status     | Supported Angular versions |
 |----------------------|-------------------------|----------------------------|
-| MSAL Angular v2      | Active development      | 9, 10, 11, 12, 13, 14      |
-| MSAL Angular v1      | Active development      | 6, 7, 8, 9                 |
+| MSAL Angular v3      | Active development      | 15                         |
```
peterzenz, line 48: So MSAL Angular v3 is only covering Angular 15?

lib/msal-angular/README.md
```diff
@@ -148,4 +153,4 @@ Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT
 
 ## We Value and Adhere to the Microsoft Open Source Code of Conduct
 
-This project has adopted the [Microsoft Open Source Code of Conduct](https://opensource.microsoft.com/codeofconduct/). For more information see the [Code of Conduct FAQ](https://opensource.microsoft.com/codeofconduct/faq/) or contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any additional questions or comments.
+This project has adopted the [Microsoft Open Source Code of Conduct](https://opensource.microsoft.com/codeofconduct/). For more information see the [Code of Conduct FAQ](https://opensource.microsoft.com/codeofconduct/faq/) or contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any additional questions or comments.
```
peterzenz, line 156: What was the diff here?

lib/msal-angular/docs/v2-docs/redirects.md
```diff
@@ -180,6 +180,8 @@ export class AppComponent implements OnInit, OnDestroy {
 
 ## 2. Subscribing to `handleRedirectObservable` manually
 
+**Note:** As of MSAL v3.x, the `MsalRedirectComponent` is now **MANDATORY** and manual subscriptions to `handleRedirectObservable` can no longer be guaranteed to work.
```
peterzenz: Can we go with a stronger "are not supported" vs "no longer guaranteed to work"?

lib/msal-angular/docs/v2-v3-upgrade-guide.md
```diff
@@ -0,0 +1,51 @@
...
+
+### `MsalRedirectComponent` is now mandatory
+
+MSAL v3.x now requires initializing the application object. Initialization has been built into the `MsalRedirectComponent`, bootstrapping the component is now **MANDATORY**.
+
+See the [guide to redirects](./v2-docs/redirects.md) on how to add `MsalRedirectComponent` to your application.
+
+### The `allowNativeBroker` flag
+
+The `allowNativeBroker` flag is now turned on by default in the configurations. To turn it off, configure your app as follows:
```
peterzenz: Please remove the section on how to disable it.  If you feel really strongly about leaving it in, please update the comment to say ''// Disables native brokering support" vs "// Disables WAM Broker"

peterzenz (changes requested):

lib/msal-angular/README.md
```diff
@@ -44,8 +45,9 @@ At a minimum, `@azure/msal-angular` will follow the [support schedule of the mai
 
 | MSAL Angular version | MSAL support status     | Supported Angular versions |
 |----------------------|-------------------------|----------------------------|
-| MSAL Angular v2      | Active development      | 9, 10, 11, 12, 13, 14      |
-| MSAL Angular v1      | Active development      | 6, 7, 8, 9                 |
+| MSAL Angular v3      | Active development      | 15                         |
+| MSAL Angular v2      | In maintenance          | 9, 10, 11, 12, 13, 14      |
```
tnorling, line 49: Do we want to adopt "Long term support" instead of "In maintenance"? Or make it clear we mean security fixes only? 

And probably mention v0 as being out of support?

lib/msal-angular/docs/v2-docs/redirects.md
tnorling: I agree, I think we need to take a firmer stance one way or the other. I am curious though why `handleRedirectObservable` is no longer expected to work?

lib/msal-angular/docs/v2-v3-upgrade-guide.md
tnorling: I'd suggest the messaging around turning it off be focused on unsupported scenarios e.g.

```suggestion
The `allowNativeBroker` flag is now turned on by default in the configurations. If you're using a B2C authority you can turn it off as follows:
```

lib/msal-angular/README.md
jo-arroyo: Hmmm I just used the language from our release cadence doc. Technically, from the time v3 is released, it will be in LTS.

lib/msal-angular/README.md
jo-arroyo: Not sure why this came up with a diff. Nothing was changed here.

lib/msal-angular/README.md
jo-arroyo: Yes. Due to breaking changes in Angular 13 and 14, MSAL Angular v3 being in Angular 15 is not backwards compatible. MSAL Angular v3 may cover Angular 16 once we test and determine compatibility.

lib/msal-angular/docs/v2-docs/redirects.md
jo-arroyo: As of this release, bootstrapping `MsalRedirectComponent` is the only way to properly initialize the application. Using `handleRedirectObservable` to manually handle redirects was an alternative to using `MsalRedirectComponent`. At point, using it directly as a redirect strategy (which is what this doc covers) will no longer work as the app will not be initialized properly.

tnorling: Is there a way someone could initialize & use handleRedirectObservable (initializationObservable perhaps?) I seem to remember quite a few people stating they could not or did not want to bootstrap the RedirectComponent

tnorling: Or maybe add initialize right before handleRedirectPromise on this [line](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/da7ec294115f8046d513e7efc18866533901241b/lib/msal-angular/src/msal.service.ts#L56)?

e.g. 

```js
handleRedirectObservable(hash?: string): Observable<AuthenticationResult> {
    return from(this.instance.initialize().then(() => this.instance.handleRedirectPromise(hash || this.redirectHash)));
  }
```

boschni: The `MsalRedirectComponent` does not seem compatible with the new `bootstrapApplication` API from Angular as it only allows bootstrapping one component. Maybe it can be exposed as standalone component? Or should `handleRedirectObservable` still be supported?

lib/msal-angular/docs/v2-v3-upgrade-guide.md
```diff
@@ -0,0 +1,51 @@
...
+        system: {
+            allowNativeBroker: false, // Disables native brokering support
+        }
+    });
+}
+```
+
+## Angular 15 and rxjs@7
+
+MSAL Angular now expects that your application is built with `@angular/core@15`, `@angular/common@15`, `rxjs@7`. As with MSAL Angular v2, `rxjs-compat` is not required.
```
sameerag: Should we also note that this isn't backward compatible?

jo-arroyo: Good point. Noted and updated.

lib/msal-angular/docs/v2-docs/redirects.md
jo-arroyo: Thanks for the suggestions. I've made the change to `handleRedirectObservable` that allows it to still be supported, and updated the guidance surrounding initialization.

lib/msal-angular/docs/v2-docs/configuration.md
```diff
@@ -29,6 +29,8 @@ Please see our [MsalInterceptor](https://github.com/AzureAD/microsoft-authentica
 
 We recommend importing `MsalRedirectComponent` and bootstrapping with the `AppComponent` if you intend to use redirects. Please see the [redirect documentation](https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev/lib/msal-angular/docs/v2-docs/redirects.md) for more details. 
 
+**Note:** As of MSAL v3.x, initialization of the application object is now required. See the [v2-v3 upgrade guide](../v2-v3-upgrade-guide.md) for more details.
```
tnorling, line 32: This doc is in v2-docs folder. Do we need to call out v3 things here?

peterzenz (approved):

tnorling (approved):

sameerag (approved):

lib/msal-angular/docs/v2-docs/configuration.md
jo-arroyo: Since there aren't many changes between v2 and v3, I will move the v2 docs into the main docs folder, so they can apply to both v2 and v3. v1 docs will remain in their own folder.

jo-arroyo: Actually, since there are external links to the v2 folder, I won't move the docs. Instead, I will call out in the v3 migration doc that most of the v2 docs apply. If there are many more changes, I will create specific v3 docs per topic.

PR #5911, opened 2023-04-19
Remove no account error in removeAccount

tnorling:
- Removes no account error from `removeAccount` as this should not be a failure case
- Re-enables tests that were skipped

### DIFF
--- a/lib/msal-common/src/cache/CacheManager.ts
+++ b/lib/msal-common/src/cache/CacheManager.ts
@@ -709,7 +709,7 @@ export abstract class CacheManager implements ICacheManager {
     async removeAccount(accountKey: string): Promise<void> {
         const account = this.getAccount(accountKey);
         if (!account) {
-            throw ClientAuthError.createNoAccountFoundError();
+            return;
         }
         await this.removeAccountContext(account);
         this.removeItem(accountKey);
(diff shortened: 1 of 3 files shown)


konstantin-msft (approved):

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -709,7 +709,7 @@ export abstract class CacheManager implements ICacheManager {
     async removeAccount(accountKey: string): Promise<void> {
         const account = this.getAccount(accountKey);
         if (!account) {
-            throw ClientAuthError.createNoAccountFoundError();
+            return;
```
sameerag, line 712: Does the end user ever call this? Swallowing the error is okay, should have been there from the beginning as it should be the "we will do best to remove but do not fail in case the account isn't".

Just curious if there is any end user impact.

sameerag (approved):

peterzenz (approved):

lib/msal-common/src/cache/CacheManager.ts
tnorling: Not directly AFAIK, always through a logout API. Even if it is though, this shouldn't impact anything. The end result we're looking for here is that the account is not cached when the promise resolves, the state of the cache before the removal happens is irrelevant.

PR #5912, opened 2023-04-19
Remove null from acquireTokenSilent response type

tnorling:
`acquireTokenSilent` does not return anything other than `AuthenticationResult`. This PR updates the return type to reflect that.


konstantin-msft (approved):

sameerag (approved):
Long time due :)

peterzenz (approved):

PR #5915, opened 2023-04-20
Add CIAM support for v3

sameerag:
Ported PR #5865 to mainline!

**Add CIAM support in MSAL JS**

Azure Active Directory (Azure AD) is adding support to customer identity access management (CIAM) solution that lets you create secure, customized sign-in experiences for your customer-facing apps and services. With these built-in CIAM features, Azure AD can serve as the identity provider and access management service for your customer scenarios.

> The CIAM authority format is still in the process of standardization. It is determined to be split from AAD and currently supported in the below formats:

* `https://tenantName.ciamlogin.com/tenant` 
* `https://tenantName.ciamlogin.com`

> All MSALs have currently agreed to add two rules for CIAM Authority:

1. Add a new `CIAM` authority type
2. If the i/p CIAM  authority does not have a path, append `"tenant.onmicrosoft.com"` to the PATH as a temporary measure to support the STS endpoints. 

The second part of this support eventually will go away.

> Please also note that this is an emerging space and the endpoints may change eventually before productization. More details about CIAM can be found here: [CIAM previews](https://github.com/microsoft/entra-previews)

**_Completed :_**
- Code changes: Added `AuthorityType: Ciam` and the path tranformation logic stated above
- Unit tests
- Docs updates
- Samples tested

### DIFF
--- a/change/@azure-msal-common-a41521ad-cf75-4f37-bc79-d009688b83cb.json
+++ b/change/@azure-msal-common-a41521ad-cf75-4f37-bc79-d009688b83cb.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add CIAM support for v3(#5915)",
+  "packageName": "@azure/msal-common",
+  "email": "sameera.gajjarapu@microsoft.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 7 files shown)


sameerag:
cc @derisen This PR should have the fix for the bug you identified. I added a change and corresponding test case. Will soon create another PR for `v2` which should be released next month, FYI.

Will share a private build on v2 soon for you to test.

tnorling (approved):

derisen (approved):

PR #5917, opened 2023-04-20
Support missing slash in CIAM authority support

sameerag:
This PR fixes a bug in CIAM support, a specific use case where the `authority` without `PATH` without a trailing slash is failing:

```
clientId: 'Enter_the_Application_Id_Here', 
authority: '[https://Enter_the_Tenant_Name_Here.ciamlogin.com']
```

This PR fixes the bug and also adds a test case to cover this use case.

cc @derisen


derisen (approved):
LGTM, thanks!

peterzenz (approved):

PR #5918, opened 2023-04-21
Bundle msal-common into msal-node

konstantin-msft:


lib/msal-node/tsconfig.json
```diff
@@ -19,21 +19,25 @@
     "noImplicitThis": true,
     "alwaysStrict": true,
     "noUnusedLocals": true,
-    "noUnusedParameters": true,
+    "noUnusedParameters": false,
```
konstantin-msft, line 22: Had to switch to false due to the high number of unused params in `msal-common`.

lib/msal-node/tsconfig.json
```diff
@@ -19,21 +19,25 @@
...
     "resolveJsonModule": true,
     "moduleResolution": "node",
     "jsx": "react",
     "esModuleInterop": true,
     "suppressImplicitAnyIndexErrors": true,
-    "experimentalDecorators": true
+    "experimentalDecorators": true,
+    "baseUrl": ".",
+    "paths": {
+        "@azure/msal-common": ["src/msal-common"]
```
konstantin-msft, line 33: Create an alias for bundled `msal-common` to avoid changing references across the module.

lib/msal-node/package.json
```diff
@@ -57,14 +56,14 @@
     "jest": "^29.5.0",
     "prettier": "2.8.7",
     "rollup": "^3.20.1",
+    "rollup-plugin-typescript-paths": "^1.4.0",
```
konstantin-msft, line 59: Rollup plugin to properly resolve `tsconfig.compileroptions.paths`.

lib/msal-node/tsconfig.json
```diff
@@ -19,21 +19,25 @@
...
+    }
   },
   "compileOnSave": false,
   "buildOnSave": false,
   "references": [
     {
-      "path": "../msal-common"
+      "path": "src/msal-common"
```
tnorling: We probably should leave this unchanged. This is what makes "go to definition" work between libs.

peterzenz (approved):

lib/msal-node/tsconfig.json
sameerag: Can we please file an ADO to tackle this later?

sameerag (approved):
Lgtm. Would like to check if `noUsedParameters:true` can be fixed eventually.

lib/msal-node/tsconfig.json
konstantin-msft: done

konstantin-msft:
@peterzenz @sameerag @tnorling I removed the original path alias for `msal-common` and replaced it with relative paths across the `msal-node` module. I didn't manage to make [jest moduleNameMapper](https://jestjs.io/docs/configuration#modulenamemapper-objectstring-string--arraystring) work for the e2e tests. Jest module name mapper doesn't work well for tests/utilities located outside of the `msal-node` that refer to the module directly, such as [this one](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/samples/e2eTestUtils/NodeCacheTestUtils.ts#L7).  Please let me know if you have any objections.

tnorling:
@konstantin-msft Did we consider [`bundleDepedencies`](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#bundledependencies) or a similar approach of simply bundling msal-common into its dependents? I have some concerns about the ease of use/maintainability of this approach of copying the source...

In any case I think it would be good to write up our options and discuss pros/cons before we land on something

konstantin-msft:
> @konstantin-msft Did we consider [`bundleDepedencies`](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#bundledependencies) or a similar approach of simply bundling msal-common into its dependents? I have some concerns about the ease of use/maintainability of this approach of copying the source...
> 
> In any case I think it would be good to write up our options and discuss pros/cons before we land on something

Bundling dependencies and specifying `@azure/msal-common` as a path alias was my initial approach.  However, it breaks e2e tests that import `msal-node` classes directly. For [example](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/samples/e2eTestUtils/NodeCacheTestUtils.ts#L7).

The options I can think of:
1. Bundling `msal-common` dependency and creating a path alias for it. Pros: less changes to `msal-node`, better maintainability. Cons: e2e tests won't work, requires changes to all jest config files.
2. Using relative references to `msal-common`. Pros: e2e tests work, no changes to jest config files required. Cons: worse maintainability because of `msal-common` being imported with relative paths.

tnorling:
> However, it breaks e2e tests that import `msal-node` classes directly. For [example](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/samples/e2eTestUtils/NodeCacheTestUtils.ts#L7).

Interesting...this should not have happened. We should be importing things from the package, not individual files/classes in samples and e2e tests. I'm also not sure I understand why our msal-common approach breaks node imports...anyway if this is the only blocker I would suggest we fix the e2e tests (we should fix this regardless of the approach we choose here IMO)

sameerag:
@konstantin-msft Can we start adding PR description to add better context?

PR #5931, opened 2023-04-25
Fix relative `msal-node` paths for `msal-node-samples`

konstantin-msft:
- Export `Serializer` and `Deserializer` from `msal-node`.
- Fix relative `msal-node` paths for `msal-node-samples`.
- Add a step to install `msal-node-samples` packages.
- Move `NodeCacheTestUtils` to `msal-node-samples` folder.


lib/msal-node/src/index.ts
```diff
@@ -47,6 +47,8 @@ export {
     SerializedAppMetadataEntity,
     SerializedRefreshTokenEntity,
 } from "./cache/serializer/SerializerTypes";
+export { Serializer } from "./cache/serializer/Serializer";
```
tnorling: @Sameerag Are these things reasonable to export on the public surface or do we need to reserve the right to change API signatures? If so, perhaps we should export them under 'internals', similar to msal-browser, or find a way not to rely on them in E2E. Thoughts?

tnorling (approved):
One question for the team, otherwise looks good. Thanks!

lib/msal-node/src/index.ts
sameerag: This is meant to serialize all the artifacts we store (creds like tokens, appmetadata etc). I prefer `internal` as this will change in case we add a new type of credential ever. Practically we haven't changed in couple years but would hate to do a major version release in case we need to add a new type.

peterzenz (approved):

lib/msal-node/src/index.ts
konstantin-msft: Let me export them under internals then.

sameerag (approved):

lib/msal-node/src/index.ts
konstantin-msft: done

PR #5932, opened 2023-04-25
Bundle msal-common into msal-browser

hectormmg:


lib/msal-browser/rollup.config.js
```diff
@@ -40,21 +40,14 @@ export default [
...
         preserveModules: false,
         output: [
             {
                 dir: "lib",
                 format: "cjs",
                 banner: fileHeader,
                 sourcemap: true,
-            },
-            {
-                dir: "lib",
```
hectormmg: Removing temporarily while I figure out how to avoid the error:

```
[!] RollupError: Invalid value "umd" for option "output.format" - UMD and IIFE output formats are not supported for code-splitting builds.
https://rollupjs.org/configuration-options/#output-format
```

PR #5933, opened 2023-04-26
Angular environment variables for angular15 sample

jo-arroyo:
Changes the Angular 15 sample app to use Angular environment variables.


tnorling (approved):

samples/msal-angular-v3-samples/angular15-sample-app/src/environments/environment.ts
```diff
@@ -3,7 +3,17 @@
 // The list of file replacements can be found in `angular.json`.
 
 export const environment = {
-  production: false
+  production: false,
+  msalConfig: {
+      auth: {
+          clientId: '',
```
derisen: nit: might use placeholders like "ENTER_CLIENT_ID" instead of empty strings here

derisen (approved):

samples/msal-angular-v3-samples/angular15-sample-app/src/environments/environment.ts
jo-arroyo: That's fair. Addressed and updated. Thanks!

PR #5940, opened 2023-04-27
changes to package.json, rollup config and tsconfig

shoatman:
- Added dev dependency on msal-common via file path
- Added msal-common to bundleDependencies (bundledDependencies - also would work I believe)
- Modified rollup config to create separate UMD and ES6 output
- NOTE: I haven't changed the package json to reflect the 2 new output locations.... we should probably look at package exports for this... i use the legacy approach in msal-browser currently.


tnorling:
Closing as we went in another direction with this

PR #5953, opened 2023-05-01
Bundle local version of msal-common into msal-browser

konstantin-msft:
- Declare `msal-common` as bundled dependency.
- Add `prepare` step to make lerna bootstrap local `msal-common` package.
- Update `prepack` step to filter out `msal-common` devDeps before publishing.

**_NOTE:_** `msal-common` is bundled under `node_modules` in tar file.


tnorling (approved):

hectormmg (approved):
Thanks, looks great!

peterzenz (approved):

PR #5967, opened 2023-05-02
Fix prepack hook

konstantin-msft:
- Fix prepack hook by building `msal-common` package before removing its non-production dependencies.

### DIFF
--- a/extensions/msal-node-extensions/package.json
+++ b/extensions/msal-node-extensions/package.json
@@ -34,7 +34,7 @@
     "lint": "cd ../../ && npm run lint:node:extensions",
     "lint:fix": "npm run lint -- -- --fix",
     "prepare": "npm install ../../lib/msal-common",
-    "prepack": "cd ../../lib/msal-common && npm install --production && cd ../../extensions/msal-node-extensions && npm run build:all"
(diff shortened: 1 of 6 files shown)


hectormmg (approved):

peterzenz (approved):

extensions/msal-node-extensions/package.json
```diff
@@ -34,7 +34,7 @@
     "lint": "cd ../../ && npm run lint:node:extensions",
     "lint:fix": "npm run lint -- -- --fix",
     "prepare": "npm install ../../lib/msal-common",
-    "prepack": "cd ../../lib/msal-common && npm install --production && cd ../../extensions/msal-node-extensions && npm run build:all"
+    "prepack": "npm run build:all && cd ../../lib/msal-common && npm install --production && cd ../../extensions/msal-node-extensions"
```
tnorling, line 37: Don't think I understand what's happening here...

konstantin-msft: We are removing `msal-common` non-production dependencies before packing.

tnorling: Why do we need to install common's dependencies AFTER building it? And why do we then change back to the extensions directory? All the steps after the build feel unnecessary, what am I missing?

PR #6002, opened 2023-05-09
added new ClientConfigurationError for invalid authority in request - v2

lalimasharda:
Throw authority mismatch error in the acquire token silent and native cases if the authority in request does not match environment specified in the account obj or MSAL config.

### DIFF
--- a/change/@azure-msal-common-0e8eab18-58d4-40bd-869e-aa2e9705cef2.json
+++ b/change/@azure-msal-common-0e8eab18-58d4-40bd-869e-aa2e9705cef2.json
@@ -1,5 +1,5 @@
 {
-  "type": "minor",
+  "type": "patch",
   "comment": "Added new ClientConfigurationError type for authority mismatch in login request #6002",
   "packageName": "@azure/msal-common",
   "email": "lalimasharda@microsoft.com",
--- a/change/@azure-msal-common-0e8eab18-58d4-40bd-869e-aa2e9705cef2.json
+++ b/change/@azure-msal-common-0e8eab18-58d4-40bd-869e-aa2e9705cef2.json
@@ -1,7 +1,7 @@
 {
   "type": "patch",
-  "comment": " added new ClientAuthError type for environment mismatch in account #6002",
+  "comment": "Added new ClientConfigurationError type for authority mismatch in login request #6002",
   "packageName": "@azure/msal-common",
   "email": "lalimasharda@microsoft.com",
-  "dependentChangeType": "patch"
+  "dependentChangeType": "minor"
 }
--- a/change/@azure-msal-common-0e8eab18-58d4-40bd-869e-aa2e9705cef2.json
+++ b/change/@azure-msal-common-0e8eab18-58d4-40bd-869e-aa2e9705cef2.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": " added new ClientAuthError type for environment mismatch in account #6002",
+  "packageName": "@azure/msal-common",
+  "email": "lalimasharda@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@azure-msal-common-0e8eab18-58d4-40bd-869e-aa2e9705cef2.json
+++ b/change/@azure-msal-common-0e8eab18-58d4-40bd-869e-aa2e9705cef2.json
@@ -1,7 +1,7 @@
 {
-  "type": "patch",
+  "type": "minor",
   "comment": "Added new ClientConfigurationError type for authority mismatch in login request #6002",
   "packageName": "@azure/msal-common",
   "email": "lalimasharda@microsoft.com",
-  "dependentChangeType": "minor"
+  "dependentChangeType": "patch"
 }
--- a/lib/msal-browser/src/interaction_client/BaseInteractionClient.ts
+++ b/lib/msal-browser/src/interaction_client/BaseInteractionClient.ts
@@ -138,9 +138,8 @@ export abstract class BaseInteractionClient {
      */
     async validateRequestAuthority(authority: string, account: AccountInfo): Promise<void> {
         const discoveredAuthority = await this.getDiscoveredAuthority(authority);
-        const validAuthority = discoveredAuthority.isAlias(account.environment) || discoveredAuthority.isAlias(this.config.auth.authority);
         
-        if(!validAuthority) {
+        if(!discoveredAuthority.isAlias(account.environment)) {
             throw ClientConfigurationError.createAuthorityMismatchError();
         }
     }
--- a/lib/msal-browser/src/interaction_client/BaseInteractionClient.ts
+++ b/lib/msal-browser/src/interaction_client/BaseInteractionClient.ts
@@ -77,10 +77,23 @@ export abstract class BaseInteractionClient {
      * Initializer function for all request APIs
      * @param request
      */
-    protected async initializeBaseRequest(request: Partial<BaseAuthRequest>): Promise<BaseAuthRequest> {
+    protected async initializeBaseRequest(request: Partial<BaseAuthRequest>, account?: AccountInfo): Promise<BaseAuthRequest> {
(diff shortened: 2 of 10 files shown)


tnorling (commented):
Does any logic need to be added to the library to throw this new error?

tnorling:
Also, I wonder if this should be a `ClientConfigurationError`?

lalimasharda:
> Also, I wonder if this should be a `ClientConfigurationError`?

@tnorling I thought that the ClientConfigurationError type is reserved for MSAL.js config issues. In our use case, ClientAuthError is probably more suitable because MSAL.js is actually configured correctly but someone is trying to deliberately change the account environment while MSAL is running. Open to discussing on this more though.

tnorling:
> I thought that the ClientConfigurationError type is reserved for MSAL.js config issues. 

Anything passed in by a developer is a "Configuration", it's not just limited to the PCA constructor.

lalimasharda:
> Does any logic need to be added to the library to throw this new error?

@tnorling I added this method: createEnvironmentMismatchError() to throw that error. Am I missing something else?

tnorling:
> I added this method: createEnvironmentMismatchError() to throw that error. Am I missing something else?

Where does this new method get called?

sameerag:
Are we considering the use case that the user can provide different authorities for `login` (configuration) and `request`?

lalimasharda:
> Are we considering the use case that the user can provide different authorities for `login` (configuration) and `request`?

@sameerag This fix is for authority being different in request (and not valid against the one provided in account or msal config). 

I think we discussed the case if the config authority is different than expected and came to the conclusion that that case would be a client issue.

lib/msal-browser/src/interaction_client/BaseInteractionClient.ts
```diff
@@ -128,6 +132,19 @@ export abstract class BaseInteractionClient {
         return UrlString.getAbsoluteUrl(redirectUri, BrowserUtils.getCurrentUri());
     }
 
+    /*
+     * If authority provided in the request does not match environment/authority specified 
+     * in the account or MSAL config, we throw an error.
+     */
+    async validateRequestAuthority(authority: string, account: AccountInfo): Promise<void> {
+        const discoveredAuthority = await this.getDiscoveredAuthority(authority);
+        const validAuthority = discoveredAuthority.isAlias(account.environment) || discoveredAuthority.isAlias(this.config.auth.authority);
```
tnorling: The check should be `request.authority || config.authority === account.authority)` instead of `request.authority === account.authority || config.authority`

lib/msal-common/src/error/ClientConfigurationError.ts
```diff
@@ -100,6 +100,10 @@ export const ClientConfigurationErrorMessage = {
     invalidAuthenticationHeader: {
         code: "invalid_authentication_header",
         desc: "Invalid authentication header provided"
+    },
+    authorityMismatch: {
+        code: "authority_mismatch",
+        desc: "Authority mismatch error. Authority provided in login request does not match with environment/authority provided in account or MSAL.js configuration options. Please make an interactive request to login."
```
tnorling:
```suggestion
        desc: "Authority mismatch error. Authority provided in login request or PublicClientApplication config does not match the environment of the provided account. Please use a matching account or make an interactive request to login to this authority."
```

lib/msal-browser/src/interaction_client/BaseInteractionClient.ts
```diff
@@ -128,6 +132,18 @@ export abstract class BaseInteractionClient {
         return UrlString.getAbsoluteUrl(redirectUri, BrowserUtils.getCurrentUri());
     }
 
+    /*
+     * If authority provided in the request does not match environment/authority specified 
+     * in the account or MSAL config, we throw an error.
+     */
+    async validateRequestAuthority(authority: string, account: AccountInfo): Promise<void> {
```
tnorling, line 139: Can you add test coverage for this method?

lib/msal-browser/src/interaction_client/BaseInteractionClient.ts
```diff
@@ -128,6 +132,18 @@ export abstract class BaseInteractionClient {
...
+        const discoveredAuthority = await this.getDiscoveredAuthority(authority);
```
sameerag, line 140: This is an additional round trip. We need to leverage it to reuse it for the following AT call. However we do not need this information for native calls right? I want to review the end to end, make sure we do not tank our perf.

tnorling: The network call is only made if it hasn't already been done previously, so no it's not an extra round-trip. And, no we don't need to do it for native calls, but we may need to do it for any cache lookups

sameerag: Yeah that makes more sense. Thanks, resolving.

sameerag (approved):

lib/msal-browser/test/interaction_client/BaseInteractionClient.spec.ts
```diff
@@ -117,5 +117,77 @@ describe("BaseInteractionClient", () => {
...
+            window.sessionStorage.clear();
+        });
+
+        it("Throw error when authority in request or MSAL config does not match with environment set for account", async () => {
+            let loginRequest = {
+                authority: "https://login.windows-ppe.net/common",
+                account: pca.getActiveAccount()
+            };
+
+            if(loginRequest.account) {
```
tnorling: If this turns out to be null, this test will still pass. Rather than using getActiveAccount and setActiveAccount can we just directly pass in an account object?

lib/msal-browser/test/interaction_client/BaseInteractionClient.spec.ts
```diff
@@ -117,5 +117,77 @@ describe("BaseInteractionClient", () => {
...
+                await testClient.validateRequestAuthority(loginRequest.authority, loginRequest.account)
+                    .catch(error => {
```
tnorling: If this method doesn't throw this test will still pass. Either throw in the .then or use the done() method instead of async/await

tnorling (changes requested):
Some changes needed in tests

lib/msal-browser/test/interaction_client/BaseInteractionClient.spec.ts
```diff
@@ -117,5 +117,77 @@ describe("BaseInteractionClient", () => {
...
+
+        it("Does not throw error when authority in request or MSAL config matches with environment set for account", async () => {
+            let loginRequest = {
+                authority: "https://login.microsoftonline.com/common",
+                account: pca.getActiveAccount()
+            };
+
+            if(loginRequest.account) {
+                await testClient.validateRequestAuthority(loginRequest.authority, loginRequest.account)
+                    .then( error => {
```
tnorling: You should use the done() pattern here.

lib/msal-browser/test/interaction_client/BaseInteractionClient.spec.ts
```diff
@@ -117,5 +117,65 @@ describe("BaseInteractionClient", () => {
...
+
+        it("Throw error when authority in request or MSAL config does not match with environment set for account", async () => {
+            let loginRequest = {
+                authority: "https://login.windows-ppe.net/common",
+                account: testAccountInfo1
+            };
+
+            await testClient.validateRequestAuthority(loginRequest.authority, loginRequest.account)
+                .then(() => {
+                    throw "String unexpected";
```
tnorling:
```suggestion
                    throw "This is unexpected. This call should have failed.";
```

change/@azure-msal-common-0e8eab18-58d4-40bd-869e-aa2e9705cef2.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
tnorling:
```suggestion
  "type": "patch",
```

tnorling (approved):

PR #6004, opened 2023-05-10
Update rxjs dev dependency for MSAL Angular

jo-arroyo:
Updates RxJS dev dependency for MSAL Angular to allow e2e tests to run properly.

### DIFF
--- a/lib/msal-angular/package.json
+++ b/lib/msal-angular/package.json
@@ -65,7 +65,7 @@
     "karma-jasmine-html-reporter": "^2.0.0",
     "ng-packagr": "^15.1.2",
     "prettier": "2.8.7",
-    "rxjs": "^7.0.0",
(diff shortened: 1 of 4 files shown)


lib/msal-angular/package.json
```diff
@@ -75,4 +75,4 @@
     "@azure/msal-browser": "^3.0.0-alpha.1",
     "rxjs": "^7.0.0"
```
tnorling, line 76: Was this an issue specific to our sample or do we need to update the peer dependency as well?

jo-arroyo: This was specifically for our CI E2E tests and shouldn't affect our customers. Samples with different RxJS versions work fine with the peer dependency of ^7.0.0.

tnorling (approved):

peterzenz (approved):

PR #6007, opened 2023-05-10
Remove MSAL Angular v1, v2, and AngularJS samples from dev

jo-arroyo:


jo-arroyo:
Wrong base, closing in favor of #6008.

PR #6011, opened 2023-05-11
Support tenant switching for tenant domain names

konstantin-msft:
- Support tenant switching for tenant domain names.
- Move `replaceTenant()` into `replacePath()`.

### DIFF
--- a/lib/msal-common/src/authority/Authority.ts
+++ b/lib/msal-common/src/authority/Authority.ts
@@ -59,7 +59,13 @@ export class Authority {
     // Correlation Id
     protected correlationId: string | undefined;
     // Reserved tenant domain names that will not be replaced with tenant id
-    private static reservedTenantDomains: Set<string> = new Set([Constants.DEFAULT_COMMON_TENANT, "{tenant}", "{tenantid}"]);
+    private static reservedTenantDomains: Set<string> = new Set([
+        Constants.DEFAULT_COMMON_TENANT,
+        "{tenant}",
+        "{tenantid}",
+        "consumers",
+        "organizations"
+    ]);
 
     constructor(
         authority: string,
@@ -235,11 +241,11 @@ export class Authority {
     }
 
     /**
-     * Returns a flag indicating that authority {@link IUri} type is AAD
+     * Returns a flag indicating that tenant name can be replaced in authority {@link IUri}
      * @param authorityUri {@link IUri}
      * @private
      */
-    private isAADAuthority(authorityUri: IUri): boolean {
+    private canReplaceTenant(authorityUri: IUri): boolean {
         return authorityUri.PathSegments.length === 1
             && !Authority.reservedTenantDomains.has(authorityUri.PathSegments[0])
             && this.getAuthorityType(authorityUri) === AuthorityType.Default
@@ -267,7 +273,7 @@ export class Authority {
 
         currentAuthorityParts.forEach((currentPart, index) => {
             let cachedPart = cachedAuthorityParts[index];
-            if (index === 0 && this.isAADAuthority(cachedAuthorityUrlComponents))
+            if (index === 0 && this.canReplaceTenant(cachedAuthorityUrlComponents))
             {
                 const tenantId = (new UrlString(this.metadata.authorization_endpoint)).getUrlComponents().PathSegments[0];
                 /**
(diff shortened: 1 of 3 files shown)


lib/msal-common/src/authority/Authority.ts
```diff
@@ -220,13 +228,24 @@ export class Authority {
...
     }
 
+    /**
+     * Returns a flag indicating that authority {@link IUri} type is AAD
+     * @param authorityUri {@link IUri}
+     * @private
+     */
+    private isAADAuthority(authorityUri: IUri): boolean {
+        return authorityUri.PathSegments.length === 1
+            && !Authority.reservedTenantDomains.has(authorityUri.PathSegments[0])
```
tnorling, line 250: An authority with these reservedDomains could still be (is probably) an AAD authority. If you still need to do this check, can we do it at the call-site? Alternatively, we could rename the method to something more accurate. Something like `canReplaceTenant` perhaps?

lib/msal-common/src/authority/Authority.ts
```diff
@@ -55,6 +58,8 @@ export class Authority {
     protected performanceClient: IPerformanceClient | undefined;
     // Correlation Id
     protected correlationId: string | undefined;
+    // Reserved tenant domain names that will not be replaced with tenant id
+    private static reservedTenantDomains: Set<string> = new Set([Constants.DEFAULT_COMMON_TENANT, "{tenant}", "{tenantid}"]);
```
tnorling: Super nitpicky - does this have to be a Set or can it be a basic array? Also, let's include the `consumers` and `organizations` reserved words here as well

lib/msal-common/src/authority/Authority.ts
konstantin-msft: That's a good point, thank you. Renamed to `canReplaceTenant`.

lib/msal-common/src/authority/Authority.ts
konstantin-msft: I use Set for faster lookups. Added additional items, thank you!

lib/msal-common/src/authority/Authority.ts
```diff
@@ -55,6 +58,14 @@ export class Authority {
     protected performanceClient: IPerformanceClient | undefined;
     // Correlation Id
     protected correlationId: string | undefined;
+    // Reserved tenant domain names that will not be replaced with tenant id
+    private static reservedTenantDomains: Set<string> = new Set([
+        Constants.DEFAULT_COMMON_TENANT,
+        "{tenant}",
+        "{tenantid}",
+        "consumers",
```
tnorling: You can use this enum for authority constants FYI: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/eeb136bcb815f32f931fda39d6f9e99d6be29a17/lib/msal-common/src/utils/Constants.ts#LL104C19-L104C19

konstantin-msft: Good point, done.

tnorling (approved):

lib/msal-common/src/authority/Authority.ts
```diff
@@ -55,6 +58,14 @@ export class Authority {
     protected performanceClient: IPerformanceClient | undefined;
     // Correlation Id
     protected correlationId: string | undefined;
+    // Reserved tenant domain names that will not be replaced with tenant id
+    private static reservedTenantDomains: Set<string> = (new Set([
```
sameerag, line 62: nit: The first two are replacement strings and not technically part of the `authority`. The next three are STS understood `authority strings`. Can we separate them conceptually?

lib/msal-common/src/authority/Authority.ts
```diff
@@ -220,13 +234,24 @@ export class Authority {
...
 
+    /**
+     * Returns a flag indicating that tenant name can be replaced in authority {@link IUri}
+     * @param authorityUri {@link IUri}
+     * @private
+     */
+    private canReplaceTenant(authorityUri: IUri): boolean {
+        return authorityUri.PathSegments.length === 1
+            && !Authority.reservedTenantDomains.has(authorityUri.PathSegments[0])
+            && this.getAuthorityType(authorityUri) === AuthorityType.Default
```
sameerag, line 251: As an aside: @tnorling do you think it is time we identity B2C as its own `authorityType`?

lib/msal-common/src/authority/Authority.ts
konstantin-msft: The only purpose of this Set is to filter out names that should not be replaced with tenant id. I am not sure it would be better to split it into pars.

sameerag (approved):
Approved with a few comments

Issue #6015, opened 2023-05-13
Requesting clarity on forced Request scopes


MattWatt:
### Core Library

MSAL.js v2 (@azure/msal-browser)

### Core Library Version

2.37.0

### Wrapper Library

Not Applicable

### Wrapper Library Version

None

### Public or Confidential Client?

Public

### Description

I'm looking for clarity around why the MSAL.js v2.x library is forcing request scopes, like "profile", on `loginPopup()` or `loginRedirect()`?

I'm trying to use the library on a public SPA application to request an id_token from an Azure B2C instance.  I don't manage or control that instance, but they've configured it so permissions are only allowed for "openid" and "offline_access" scopes, but not "profile".  No matter what specific request scopes I send to `loginPopup()`, the request is always made with the scopes: "openid", "profile", "offline_access".

BTW, it took me a lot of digging to discover that the "invalid_grant" error with code AADB2C90085 was a result of scope permissions.  That was not clear in the error description.

I'm seeing conflicting info from B2C examples for MSAL.js v2.x that just show the use of the "openid", as well as some of the FAQ information that seems less than clear around whether or why these scopes are forced.

I can however pinpoint the exact code block in the `msal-common` library that indeed shows that it is appended to every Auth Request.:
https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/eeb136bcb815f32f931fda39d6f9e99d6be29a17/lib/msal-common/src/request/RequestParameterBuilder.ts#L86C12-L89

I'm now in a pickle since the library is forcing scopes I can't control and integrating with a partner that does not want to allow permissions to the "profile" scope.

I appreciate any help or clarity you can provide.

### MSAL Configuration

```javascript
{
	"auth": {
		"clientId": "7b695e04-fe0f-41d8-b16c-03207d6c3f31",
		"authority": "https://{omitted_tenant_name}.b2clogin.com/{omitted_tenant_name}.onmicrosoft.com/b2c_1a_signup_signin_drg",
		"knownAuthorities": [
			"{omitted_tenant_name}.b2clogin.com"
		],
		"cloudDiscoveryMetadata": "",
		"authorityMetadata": "",
		"redirectUri": "http://localhost:11400/oidc",
		"postLogoutRedirectUri": "http://localhost:11400/signout/oidc",
		"navigateToLoginRequestUrl": false,
		"clientCapabilities": [],
		"protocolMode": "AAD",
		"azureCloudOptions": {
[...]

PR #6021, opened 2023-05-16
fixing issue with default OCID scopes

tamdatameshgroup:
this allow adding new optional config value addOidcScopes into request to let parameter to not  auto add OIDC default scopes
this is for fixing this opened issue
https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/6015


tamdatameshgroup:

tnorling:
Thanks for the PR. Unfortunately, we cannot accept this right now. These scopes give us important artifacts we rely on, removing this dependency will require more work on our end - something we're considering but have not committed to yet.

PR #6045, opened 2023-05-19
removed unneeded method definitions from controller interface

shoatman:
- Removed method definitions from the IController interface that did not belong there (Looks like doing so did not break any build.... will check the end 2 end tests to confirm)

### DIFF
--- a/lib/msal-browser/src/controllers/IController.ts
+++ b/lib/msal-browser/src/controllers/IController.ts
@@ -8,8 +8,6 @@ import {
     AccountInfo,
     Logger,
     PerformanceCallbackFunction,
-    IPerformanceClient,
-    ICrypto,
     CommonSilentFlowRequest,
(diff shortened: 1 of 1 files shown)


tnorling (commented):
You can probably remove `acquireTokenNative` and `acquireTokenByRefreshToken` from this as well. Those are implementation details.

Also changelog?

konstantin-msft (changes requested):
These methods are used by the 1p library, please do not remove.

shoatman:
> These methods are used by the 1p library, please do not remove.

Which methods?

konstantin-msft:
> > These methods are used by the 1p library, please do not remove.
> 
> Which methods?

All of them

tnorling:
Closing as a lot of this has been addressed and the remaining ones will be addressed

PR #6068, opened 2023-05-31
Switch from enums to object literals to reduce the bundle size

konstantin-msft:
- Switch from enums to object literals to reduce the bundle size.

### DIFF
--- a/change/@azure-msal-node-extensions-947e9597-9fab-4896-9e83-227c0b90a3bf.json
+++ b/change/@azure-msal-node-extensions-947e9597-9fab-4896-9e83-227c0b90a3bf.json
@@ -1,7 +1,7 @@
 {
-  "type": "none",
+  "type": "prerelease",
   "comment": "Switch from enums to object literals to reduce the bundle size #6068",
   "packageName": "@azure/msal-node-extensions",
   "email": "kshabelko@microsoft.com",
-  "dependentChangeType": "none"
+  "dependentChangeType": "prerelease"
 }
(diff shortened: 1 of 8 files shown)


change/@azure-msal-node-extensions-947e9597-9fab-4896-9e83-227c0b90a3bf.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
```
tnorling:
```suggestion
  "type": "prerelease",
```

lib/msal-react/src/MsalProvider.tsx
```diff
@@ -31,10 +31,11 @@ type MsalState = {
     accounts: AccountInfo[];
 };
 
-enum MsalProviderActionType {
-    UNBLOCK_INPROGRESS = "UNBLOCK_INPROGRESS",
-    EVENT = "EVENT",
-}
+const MsalProviderActionType = {
```
tnorling, line 34: This can probably just be `const enum` as it's not exported.

tnorling (approved):

lib/msal-react/src/MsalProvider.tsx
konstantin-msft: I would prefer to be on the safe side and use object literal instead of const enum in case it needs to be exported.

change/@azure-msal-node-extensions-947e9597-9fab-4896-9e83-227c0b90a3bf.json
konstantin-msft: fixed, thank you

lib/msal-react/src/MsalProvider.tsx
tnorling: This will never be exported as it is an implementation detail used for state management. I'll leave the final call up to you but please remove the export on the line below if you choose to keep it as an object literal.

konstantin-msft: Good point, done. Thank you.

lalimasharda (approved):

PR #6077, opened 2023-05-31
Remove expression syntax in job conditional

tnorling:
Apparently Github Actions has [a bug](https://github.com/actions/runner/issues/1173) where the usage of expression syntax in an if conditional can result in false positives. The expression syntax is [implicit in if conditionals](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif) anyway so this PR removes it to address false positives in our own pipelines.

### DIFF
--- a/.github/workflows/ci-template.yml
+++ b/.github/workflows/ci-template.yml
@@ -24,7 +24,7 @@ concurrency:
 
 jobs:
   build-test:
-    if: ${{ inputs.should-run }}}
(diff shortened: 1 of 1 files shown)


lalimasharda (approved):

konstantin-msft (approved):

tnorling:
Will be addressed in #6084

PR #6078, opened 2023-05-31
Add First Party Wrapper SKU

hectormmg:
This PR adds the `WrapperSKU.FirstParty` entry corresponding to the `msal-javascript-1p` SDK.


lib/msal-browser/src/utils/BrowserConstants.ts
```diff
@@ -186,6 +186,7 @@ export const KEY_FORMAT_JWK = "jwk";
 export const WrapperSKU = {
     React: "@azure/msal-react",
     Angular: "@azure/msal-angular",
+    FirstParty: "@azure/msal-javascript-1p"
```
tnorling, line 189: How will this work when using both 1P and React or Angular?

hectormmg: I'm sorry if we've discussed this before, I kind of lost some of the context I had. 

Should we then use the ESTS SKU to distinguish? 

If I understand correctly now, the combination of ESTS telemetry SKU (browser/node/1p) and this wrapper SKU in the telemetry gives us the complete data point of what's being used right?

tnorling: I would try to avoid changing the eSTS SKU as they build some logic around the value AFAIK. My suggestion is to replicate the behavior of the React/Angular SKUs for 1P (rather than making it a distinct option on the existing logic) so that if both are used they are both present on the server telemetry fields.

hectormmg:
Closed since we're leveraging the ESTS SKU instead.

Issue #6012, opened 2023-05-11
Add Next.js sample that utilises the new App Router (and potentially make TypeScript first too)


jenewland1999:
### Core Library

MSAL.js v2 (@azure/msal-browser)

### Wrapper Library

MSAL React (@azure/msal-react)

### Public or Confidential Client?

Public

### Description

Next.js recently promoted their new app router from beta to stable and fundamentally changes the way the framework works by using React Server Components (RSC) to be a server-rendered-first framework. 

The existing sample docs for Next.js use the traditional Pages Router and while these are still helpful, it would be great if there were docs for the new App Router model too.

Also, TypeScript-first samples would be great as it's a lot easier to remove TypeScript syntax than it is to add it (as you have to figure out what the correct types are for various things).

https://nextjs.org/docs

### Source

External (Customer)

hectormmg:
cc: @EmLauber can we track this sample request internally please? Thanks!

EmLauber:
Tracking the request and hopefully we can get it addressed via the community submitted sample linked.

sakgoyal:
any updates?

kletse:
Any updates on this? Would love a working sample of this as I'm just starting on a new project with Next.js and MSAL...

spacecat:
Next.js 14 was announced last thurday.
Vercel also made a fantastic job of creating an entirely new 16 chapter tutorial using the new App Router.
Nextjs.org/learn
They have a dedicated chapter for authentication. 
They use nextauth.js.
In their main docs they mention Clerk, Auth0, etc but no MSAL React.
I think that in order for you to complete with these 3rd party auth services you need to create a sample based on nextjs.org/learn, add ALL real world features, or at least the most common ones, and on top, create a tutorial with lots of images and easy to follow text.

I think if you did that the community would be very happy and at the same time MSAL would become even more popular and Azure AD B2C a more interesting alternative. 

What is the status btw on the latest Next.js msal samples?

tobi1220:
Any news on this?

ehsan-shv:
I guess there are several issues with Next.js App Router and msal.js. It's better to publish an official example with Next.js v13 and v14 ASAP.

spacecat:
[...]

PR #6083, opened 2023-06-01
NextJS App Router RSC sample

h3rmanj:
Addresses #6012

This PR adds a sample that integrates `@azure/msal-node` with React Server Components in the NextJS App Router.  
Submitting a Draft PR as I've gotten most core parts to work, but need to finalize the sample.

Saw you did some stuff with client components already @tnorling, but this sample aims to integrate `@azure/msal-node`, so the full capabilities of the NextJS server with RSC can be utilized. I'll gladly accept feedback and help to make this sample meet the standards of this repository 😊

### Sample goals

- [x] React Server Component support
- [x] Forced Auth
- [x] Optional Auth
- [x] Gradual Consent
- [x] Works with Route Handlers
- [x] Works with Server Actions
- [x] Logout
- [x] JSDoc on `AuthProvider`
- [ ] ~~Multiple accounts~~
  I don't have any stake in this feature, but will gladly help if required.
- [ ] ~~Error handling~~
  I've mostly done the happy path, but this depends on auth scenario, so didn't pursue it for this sample.
- [ ] ~~Work with Pages Router as well?~~

### Remarks

I tried to make a generic React Server Component integration with the `AuthProvider`, but that might have made this sample more complicated than needed.

I also needed to find a suitable cookie/session management library. Most I could find was either too specific for the old pages router, or so small and specific I would need to manually combine and cook something. I ended up using `createCookieSessionStorage` from Remix, which might sound a little weird, but since it works directly against core Web APIs it works out.


h3rmanj:
Marking as ready, I'd love some feedback, if this is something you'd like to include in this repo.

For some time now, the React tooling/frameworks has moved in a direction to SSR. The official msal-react library was made and designed with SPAs in mind, and using it in one of those frameworks would negate many of the reasons to use such a framework in the first place. As a developer in a company with Azure at it's core I've basically felt a little left out, with no official library or samples for those kinds of frameworks.

Now with RSC on the way, I feel this might be a chance to get ahead of the curve, at least with a sample. Allthough the end goal should probably be an official library that supports Server Components.

samples/msal-node-samples/nextjs-rsc-ts/.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
```
tnorling, line 1: Please remove as we already have repo level settings

samples/msal-node-samples/nextjs-rsc-ts/README.md
```diff
@@ -0,0 +1,68 @@
+# @azure/msal-node for Next.js App Router example
```
tnorling:
```suggestion
# @azure/msal-node for Next.js with Server Components
```

samples/msal-node-samples/nextjs-rsc-ts/README.md
```diff
@@ -0,0 +1,68 @@
...
+Install and run the development server:
+
+```bash
+npm install
+npm start
+```
+
+Open [http://localhost:3000](http://localhost:3000) with your browser to see the result.
+
+## Remarks
```
tnorling, line 122: Empty section?

samples/msal-node-samples/nextjs-rsc-ts/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "who-next",
```
tnorling:
```suggestion
  "name": "nextjs-server-component-sample",
```

samples/msal-node-samples/nextjs-rsc-ts/src/utils/AuthProvider.ts
```diff
@@ -0,0 +1,174 @@
...
+    const instance = await this.getInstance();
+
+    const requestWithRedirectUri = {
+      ...request,
+      redirectUri: this.redirectUri,
+    };
+
+    const state = this.cryptoProvider.base64Encode(
+      JSON.stringify({
+        request: requestWithRedirectUri,
```
tnorling: Quite a few of the parameters included on this object are not needed on the code exchange later on and do not need to be passed through the auth server. I'd suggest only passing through the bits you'll actually need or store them somewhere temporarily for later reference.

samples/msal-node-samples/nextjs-rsc-ts/src/services/auth.ts
```diff
@@ -0,0 +1,48 @@
...
+  const cookie = cookies().get("__session");
+
+  const session = await getSession(`__session=${cookie?.value}`);
+
+  return new SessionPartitionManager(session);
+}
+
+export const authProvider = new AuthProvider(
+  msalConfig,
+  "http://localhost:3000/auth/callback",
```
tnorling: I'd suggest breaking out any configuration values/constants like this into a separate file

samples/msal-node-samples/nextjs-rsc-ts/src/components/mui.tsx
```diff
@@ -0,0 +1,12 @@
+"use client";
+
+export {
```
tnorling: What's the purpose of this file versus simply importing what you need directly from mui?

samples/msal-node-samples/nextjs-rsc-ts/src/components/WelcomeName.tsx
```diff
@@ -0,0 +1,16 @@
+import { authProvider } from "~/services/auth";
+import { Typography } from "./mui";
+
+export default async function WelcomeName() {
+  const account = await authProvider.getAccount();
```
tnorling, line 5: Will this work both on server and client? Or should this component be marked as a server component?

samples/msal-node-samples/nextjs-rsc-ts/src/components/SignOutButton.tsx
```diff
@@ -0,0 +1,43 @@
...
+        keepMounted
+        transformOrigin={{
+          vertical: "top",
+          horizontal: "right",
+        }}
+        open={Boolean(anchorEl)}
+        onClose={() => setAnchorEl(null)}
+      >
+        <form action={logout}>
+          <MenuItem component={"button"} type="submit" key="logoutPopup">
```
tnorling: logoutPopup?

samples/msal-node-samples/nextjs-rsc-ts/src/app/profile/page.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const token = await instance.acquireTokenSilent({
+    account,
+    scopes: ["User.Read"],
+  });
+
+  if (!token) {
+    return null;
+  }
+
+  const response = await fetch("https://graph.microsoft.com/v1.0/me", {
```
tnorling: Endpoint is another good thing to put in a config file

tnorling (commented):
First of all, thank you for taking the time to put this together and opening a pull request. It's been on my to-do list for ages but can't seem to find the time so your contribution is very much appreciated! I've given it a quick glance and left a few minor comments but have not had a chance yet to spin it up and try it out. 

A couple general comments/questions:

- There's a lot going on here... I agree some of these things could be abstracted away by an official library but prioritization is the name of the game... In the meantime it would be great if you could move all configurations to a central file and update the README to list out each file with a short description of what it's responsible for as well as a short high-level overview of how this works (some of the pieces are there but I think it assumes the reader is already familiar with both MSAL and Next.js RSC)
- Is the idea here that every time the application needs to get data from an API it needs to go back to the server, make a token request, make the API request and then render the client content? It might be good to demonstrate the hybrid flow so that once the client has rendered it can continue to do authenticated things on the client rather than going back to the server. That can be a future enhancement, however.

I'd also like @derisen and @bgavrilMS to review this as well.

derisen:
@h3rmanj this seems exciting, thank you for your time! I'll get to review it as soon as I can.

> * Is the idea here that every time the application needs to get data from an API it needs to go back to the server, make a token request, make the API request and then render the client content? It might be good to demonstrate the hybrid flow so that once the client has rendered it can continue to do authenticated things on the client rather than going back to the server. That can be a future enhancement, however.

This sounds like the [backend for frontend proxy architecture](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-browser-based-apps-13#name-backend-for-frontend-bff-pr), am I right? We recently wrote [a sample implementing this pattern](https://github.com/Azure-Samples/ms-identity-javascript-nodejs-tutorial/tree/main/5-AdvancedScenarios/1-call-graph-bff#scenario). I actually would love this sample to appear along the other BFF sample in that repo, but I'll leave it your and @tnorling's preference.

Regarding the hybrid spa code flow, it might be optional but I wouldn't say its a necessary improvement on top of this. The BFF pattern has the advantage of keeping the frontend side clear of auth logic and browser storage free of auth artifacts against XSS (though this will be less of an issue in future with token binding), but hybrid spa flow would be more performant for sure (since the roundtrip for every API call means additional latency).

samples/msal-node-samples/nextjs-rsc-ts/.vscode/settings.json
jenewland1999: I believe the lines 5 and 6 would be required here as Next.js now ships with its own TypeScript plugin.

Source: https://nextjs.org/docs/app/building-your-application/configuring/typescript

samples/msal-node-samples/nextjs-rsc-ts/src/components/WelcomeName.tsx
jenewland1999: By default, all components in Next.js that are used in the App Router are considered server components. Only components marked with the `"use client";` directive are rendered on the server then hydrated on the client.

Source: https://nextjs.org/docs/getting-started/react-essentials#the-use-client-directive

The use of the `"use server";` directive further up I believe is part of the server actions RFC that is currently experimental in Next.js/React and imho should probably be avoided using in docs until it is stable. Though it could be split into a separate example so when they are stable the example is already there.

samples/msal-node-samples/nextjs-rsc-ts/src/components/mui.tsx
jenewland1999: This is a temporary thing until third party libraries like MUI adopt the React `"use client";` directive for components that utilise React hooks or context. Next.js suggest this as a work around in the meantime.

Though, it's worth adding a comment with a link to the docs to explain that in the file so it's clear.

Source: https://nextjs.org/docs/getting-started/react-essentials#third-party-packages

samples/msal-node-samples/nextjs-rsc-ts/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "eslint-config-next": "13.4.1",
+    "next": "^13.4.4",
+    "react": "18.2.0",
+    "react-dom": "18.2.0",
+    "redis": "^4.6.6",
+    "typescript": "^5.1.3"
+  },
+  "devDependencies": {
+    "eslint-config-prettier": "^8.8.0",
+    "eslint-plugin-prettier": "^4.2.1",
```
jenewland1999: It's considered a bad practice to use Linters for formatting code. I've linked some sources below that explain why, but the tl;dr is you should use linters for linting and formatters for formatting. 😃 

Sources:
- https://eslint.org/blog/2020/05/changes-to-rules-policies/#what-are-the-changes
- https://typescript-eslint.io/linting/troubleshooting/formatting/
- https://prettier.io/docs/en/comparison.html
- https://prettier.io/docs/en/integrating-with-linters.html
- https://www.youtube.com/watch?v=Cd-gBxzcsdA
- https://www.youtube.com/watch?v=sSJBeWPIipQ

samples/msal-node-samples/nextjs-rsc-ts/src/actions/auth.ts
```diff
@@ -0,0 +1,31 @@
+"use server";
```
jenewland1999, line 1: While it's great to see how this is done with Server Actions, given they're in alpha not sure if it's wise to showcase them in this sample? Maybe have a separate sample that showcases their use with a warning to be explicit about their early-stage development.

Source: https://nextjs.org/docs/app/building-your-application/data-fetching/server-actions

jenewland1999:
Thank you so much for putting together this PR @h3rmanj. I left a couple comments, but feel free to disregard them, just a couple of tips 😃.

@tnorling I replied to a couple of your comments where you had some questions about some of the Next.js/React specific stuff that's relatively new. Hope they're helpful.

samples/msal-node-samples/nextjs-rsc-ts/src/actions/auth.ts
h3rmanj: I agree it might be a little early for server actions. But I do feel they are an integral part to the RSC story, so I'd still like them to be a part of this sample.

But I have added some sample buttons that utilizes Route Handlers as well 🙂

samples/msal-node-samples/nextjs-rsc-ts/src/utils/AuthProvider.ts
h3rmanj: I'm not sure what's required from the original request when I call aquireTokenByCode. It seemed I needed most of it, but maybe scopes are enough?

samples/msal-node-samples/nextjs-rsc-ts/.vscode/settings.json
h3rmanj: Removing some of the properties, but as of now the TypeScript extension ships with version 5.0.4 of TypeScript, while 5.1.0 contains some enhancements specifically for Server Components, so I had to include it.

samples/msal-node-samples/nextjs-rsc-ts/src/components/mui.tsx
h3rmanj: I've added a comment, but as @jenewland1999 points out it's to allow using these components from server components without Next.js assuming they are server components too

samples/msal-node-samples/nextjs-rsc-ts/src/components/WelcomeName.tsx
h3rmanj: It will only work in server components. As stated all components in App Router are server by default.

I will however poison the authProvider modules with `import "server-only"` to avoid them accidentally being imported in client components.

https://nextjs.org/docs/getting-started/react-essentials#keeping-server-only-code-out-of-client-components-poisoning

h3rmanj:
> There's a lot going on here... I agree some of these things could be abstracted away by an official library but prioritization is the name of the game... In the meantime it would be great if you could move all configurations to a central file and update the README to list out each file with a short description of what it's responsible for as well as a short high-level overview of how this works (some of the pieces are there but I think it assumes the reader is already familiar with both MSAL and Next.js RSC)

Updated the README a bit more, it's more of a first draft of things, so feedback on the structure is greatly appreciated. I think some familiarity with MSAL, RSC and Next.js would be required in general if you want to use this sample, but I could elaborate on specific things if necessary.

--- 
> Is the idea here that every time the application needs to get data from an API it needs to go back to the server, make a token request, make the API request and then render the client content? It might be good to demonstrate the hybrid flow so that once the client has rendered it can continue to do authenticated things on the client rather than going back to the server. That can be a future enhancement, however.

Yes, as pointed out, this would be the BFF pattern, which is essentially what all these new React-frameworks (Remix, Next.js etc) are, BFFs with the added ability to render React.

I'm not sure how I would go about implementing a hybrid flow that can do things both on the server and client though. Would I serialize the cache on the server, then initialize an instance on the client with that cache? Or do the server just send the user to azure, and the client handle the redirect? Either way I think that would be best kept in it's own sample.

Regarding the performance part, it's essentially a tradeoff depending on the users of your application. Is it better that the acquire token + fetch api (maybe even multiple apis) waterfall happens on the client, or on the server? This greatly depends on the users condition, and the BFF would make the performance consistent between differing user conditions. Since there already are client samples with msal-react, this sample would cater to the other end.

--- 
> We recently wrote [a sample implementing this pattern](https://github.com/Azure-Samples/ms-identity-javascript-nodejs-tutorial/tree/main/5-AdvancedScenarios/1-call-graph-bff#scenario). I actually would love this sample to appear along the other BFF sample in that repo, but I'll leave it your and @tnorling's preference.

If that is a better (or additional) location for this sample is up to you, I've personally missed something like this along the official library repo, so I just started here.

---

I feel like I also have to mention that while App Router is stable in Next.js, React Server Components aren't really documented by React themselves yet, and Server Actions are still experimental. I've included an alternative to server actions in the sample, but I still feel like they should be included, as they play an important part to RSCs story.

Also, if anyone have any suggestion of a library that handles cookie sessions/signing, I'd greatly appreciate it. The only ones I could find was too tied up to the old pages router, or very specific for express middleware etc. While the one I use now works straight with the cookie header, which is kind of what I'm after, I think "importing remix in next" might raise some eyebrows.

samples/msal-node-samples/nextjs-rsc-ts/src/utils/AuthProvider.ts
tnorling: These 4 would be a good start, we can adjust if needed based on feedback:

```
correlationId
scopes
claims
azureCloudOptions
```

samples/msal-node-samples/nextjs-rsc-ts/README.md
```diff
@@ -0,0 +1,134 @@
...
+if (!account) {
+  redirect(await authProvider.getAuthCodeUrl(codeRequest, getCurrentUrl()));
+}
+```
+
+This will redirect the user to azure at the earliest possible time, using HTTP headers on the initial request to the page.
+
+This can be useful if your whole app relies on being authenticated.
+
+### Gradual consent
```
derisen:
```suggestion
### Incremental consent
```

samples/msal-node-samples/nextjs-rsc-ts/.gitignore
```diff
@@ -0,0 +1,35 @@
...
+
+# local env files
+.env*.local
+
+# vercel
+.vercel
+
+# typescript
+*.tsbuildinfo
+next-env.d.ts
```
derisen, line 35:
```suggestion
next-env.d.ts

# vscode settings
.vscode
```

samples/msal-node-samples/nextjs-rsc-ts/README.md
```diff
@@ -0,0 +1,134 @@
...
+- Redirect URI (**WEB**): http://localhost:3000/auth/callback
+- API permissions `User.Read` and `Calendars.Read`
+- A client secret
+
+Set up a `.env.local` file with the following values
+
+```env
+AZURE_AD_CLIENT_ID=YOUR_CLIENT_ID
+AZURE_AD_CLIENT_SECRET=YOUR_CLIENT_SECRET
+AZURE_AD_TENANT_ID=YOUR_TENANT_ID
```
derisen:
```suggestion
AZURE_AD_AUTHORITY=YOUR_AZURE_AD_AUTHORITY
```

derisen: per serverConfig.ts

samples/msal-node-samples/nextjs-rsc-ts/src/serverConfig.ts
```diff
@@ -0,0 +1,32 @@
...
+};
+
+export const msalConfig: Configuration = {
+  auth: {
+    clientId: process.env.AZURE_AD_CLIENT_ID!,
+    clientSecret: process.env.AZURE_AD_CLIENT_SECRET!,
+    authority:
+      process.env.AZURE_AD_AUTHORITY ??
+      "https://login.microsoftonline.com/common",
+  },
```
derisen, line 19: can we add logging as well? https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/logging.md

samples/msal-node-samples/nextjs-rsc-ts/src/utils/AuthProvider.ts
```diff
@@ -0,0 +1,174 @@
...
+
+    const code = url.searchParams.get("code");
+
+    if (!code) {
+      throw new Error("No code found.");
+    }
+
+    const authResult = await instance.acquireTokenByCode({
+      ...state.request,
+      code,
```
derisen: you can pass the entire query response into a JS object, so that acquireTokenByCode also validates the returned parameters ([see](https://azuread.github.io/microsoft-authentication-library-for-js/ref/classes/_azure_msal_node.confidentialclientapplication.html#acquiretokenbycode)). Something like:

```js
const urlParams = new URLSearchParams(url);
const params = Object.fromEntries(urlParams)

const authResult = await instance.acquireTokenByCode({
      ...state.request,
      params
});
```

Though I'd much setting the response mode to `form_post` in the authCodeUrl request, and then receiving the auth code response as a POST request to your server (recommended for confidential client apps in general).

samples/msal-node-samples/nextjs-rsc-ts/src/utils/AuthProvider.ts
```diff
@@ -0,0 +1,174 @@
...
+    });
+  }
+
+  /**
+   * Handles token acquisition based on the url that the user was sent to from Azure.
+   * @param url The return url from Azure
+   * @returns An object containing the logged in account, and where the user should be redirected to.
+   */
+  async handleAuthCodeCallback(url: URL) {
+    const instance = await this.getInstance();
```
derisen: A new CCA instance will come with an empty metadata cache. See this for more: [performance-considerations-for-confidential-client-applications](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/docs/performance.md#performance-considerations-for-confidential-client-applications). 

There's an ongoing discussion on what's the best way to cache metadata for web app samples, so leaving this here as a reminder, but something [like this](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/a29f54835d039b109929bad313d488c29d7a1485/samples/msal-node-samples/auth-code-distributed-cache/src/AuthProvider.ts#L135) might be here.

samples/msal-node-samples/nextjs-rsc-ts/.gitignore
h3rmanj: Do you want the `.vscode/settings.json` to be removed, or locked from further changes?

I see the shipped TypeScript version is now up to date, but nextjs still ships their own plugin. The included settings only suggest to the user to use the workspace version when opening the folder.

samples/msal-node-samples/nextjs-rsc-ts/src/utils/AuthProvider.ts
h3rmanj: Switched to using `ResponseMode.FORM_POST`, and passing all form data to the second argument.

samples/msal-node-samples/nextjs-rsc-ts/.gitignore
derisen: I meant to remove, but I see where you're coming from. Do they (nextjs) have a guidance on this?

samples/msal-node-samples/nextjs-rsc-ts/src/utils/SessionPartitionManager.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export class SessionPartitionManager implements IPartitionManager {
+  session: Session;
+
+  constructor(session: Session) {
+    this.session = session;
+  }
+
+  async getKey() {
+    const homeAccountId = this.session.get("homeAccountId");
```
derisen:
```suggestion
    const homeAccountId = this.session.get("homeAccountId") || "";
```

samples/msal-node-samples/nextjs-rsc-ts/src/utils/AuthProvider.ts
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    return data as Omit<AuthorizationCodePayload, "state"> &
+      Required<Pick<AuthorizationCodePayload, "state">>;
+  }
+
+  private async getMetadata(clientId: string, authority: string) {
+    const tenantId = authority!.split("/").pop()!;
+
+    try {
```
derisen, line 226:
```suggestion
            /**
             * If the current msal configuration does not have cloudDiscoveryMetadata or authorityMetadata, we will 
             * make a request to the relevant endpoints to retrieve the metadata. This allows MSAL to avoid making 
             * metadata discovery calls, thereby improving performance of token acquisition process. For more, see:
             * https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/docs/performance.md
             */
    try {
```

derisen: or something similar to the effect

samples/msal-node-samples/nextjs-rsc-ts/src/serverConfig.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export const loginRequest = {
+  scopes: ["User.Read"],
+};
+
+export const calendarRequest = {
+  scopes: ["Calendars.Read"],
+};
+
+export const authCallbackUri = "http://localhost:3000/auth/callback";
```
derisen: nit: might as well read the callback uri from environment variables too (would come handy when deploying the app)

samples/msal-node-samples/nextjs-rsc-ts/src/utils/AuthProvider.ts
```diff
@@ -0,0 +1,285 @@
...
+
+    const data = Object.fromEntries(stringEntries);
+
+    if (!("state" in data)) {
+      throw new Error("No state found in payload.");
+    }
+
+    if (!("code" in data)) {
+      throw new Error("No code found in payload.");
+    }
```
derisen, line 211: >     if (!("state" in data)) {
>       throw new Error("No state found in payload.");
>     }
> 
>     if (!("code" in data)) {
>       throw new Error("No code found in payload.");
>     }

I'm thinking if there's value in these early checks or would it be better to let msal throw it's own errors. I'm leaning on the latter.

samples/msal-node-samples/nextjs-rsc-ts/src/components/CalendarEvent.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export type GraphCalendarEvent = {
+  subject: string;
+  start: {
+    dateTime: string;
+    timezone: string;
+  };
+};
+
+export const CalendarEvent = ({ event }: { event: GraphCalendarEvent }) => {
```
derisen, line 20:
```suggestion
export const CalendarEvent = ({ event }: { event: GraphCalendarEvent }) => {
  if (!event) {
    return (<div>Could not found any events</div>);
  }
```

samples/msal-node-samples/nextjs-rsc-ts/src/app/auth/eventConsent/route.ts
```diff
@@ -0,0 +1,10 @@
+import { redirect } from "next/navigation";
+import { NextRequest } from "next/server";
+import { loginRequest } from "~/serverConfig";
+import { authProvider } from "~/services/auth";
+
+export async function GET(request: NextRequest) {
+  const returnTo = new URL("/event", request.url).toString();
+
+  redirect(await authProvider.getAuthCodeUrl(loginRequest, returnTo));
```
derisen: > redirect(await authProvider.getAuthCodeUrl(loginRequest, returnTo));

Did you mean to use `calendarRequest` here instead?

samples/msal-node-samples/nextjs-rsc-ts/src/utils/AuthProvider.ts
h3rmanj: Added note where the metadata injection happens, and jsdoc to the metadata methods

samples/msal-node-samples/nextjs-rsc-ts/src/app/auth/eventConsent/route.ts
h3rmanj: Yes, fixed

samples/msal-node-samples/nextjs-rsc-ts/src/utils/AuthProvider.ts
h3rmanj: State is needed before it is passed to MSAL, as it stores the request made, which is needed for incremental consent. I'm not sure it's the right place to store it though.

As for code, the types of msal doesn't allow me to pass a null/undefined `code` value, so based on the types alone I wouldn't think msal could throw its own errors even. Since I cast it to the required type below, I want to validate that the required properties are present.

samples/msal-node-samples/nextjs-rsc-ts/.gitignore
h3rmanj: They have some docs for it here  
https://nextjs.org/docs/app/building-your-application/configuring/typescript#typescript-plugin

I also added a path to the installed prettier version, when I opened the directory it crashed, probably because one of the parent folders also has prettier installed and configured.

This would come down to whether a sample should work as it's own standalone folder, or require the context of the parent directories, which is your call.

samples/msal-node-samples/nextjs-rsc-ts/src/utils/AuthProvider.ts
derisen: Got it, thanks for elaborating this. About storing the initial request via the state param -If possible, I would prefer using the session for this, since the state param travels over the address bar i.e. goes into the browser history. Not that there's any sensitive info in the request object here, but more of a general precaution.

samples/msal-node-samples/nextjs-rsc-ts/.gitignore
derisen: Ok, I tried and ran into the same problem. I'd prefer the least amount of friction for devs, so let's keep it then : )

samples/msal-node-samples/nextjs-rsc-ts/src/utils/AuthProvider.ts
h3rmanj: I need to think about this, because of the way I made it right now. The `AuthProvider` doesn't really do any session-specific logic, or nextjs specific logic, other that calling the partition manager. Currently I also use session with cookie storage, and should probably switch to something else due to cookie size limitations.

I'm thinking of the following options for now;

- Use [`createSessionStorage`](https://remix.run/docs/en/1.17.1/utils/sessions#createsessionstorage) with redis, but again that's from remix which I should probably move away from in this sample.
- Use the next version of [`iron-session`](https://github.com/vvo/iron-session/issues/586), which is framework-agnostic and would support app router, however this is in alpha.
- Use `AuthProvider.cacheClient` to store the request, with a key generated from `AuthProvider.cryptoProvider.createNewGuid()`, which i then store in the state, or session cookie storage.

derisen: I see. In that case I would say using the state for storing request object is fine, as far as samples go. I'd feel easier though if we could add some comments to encourage using sessions whenever possible, or if they have to use the state param, make sure to not include any sensitive information in it.

renchris: @h3rmanj where may I find the alpha release that supports app router? What is your solution to work with app router + iron-session in the short-medium development term?

h3rmanj: @renchris I haven't tried integrating with it yet. The alpha is released under the `alpha` tag on npm, but no docs are available yet. The issue I linked has some discussion with people managing to integrate it.

derisen (approved):
LGTM -thanks @h3rmanj

samples/msal-node-samples/nextjs-rsc-ts/src/utils/AuthProvider.ts
h3rmanj: Added a remark in the README.

tnorling:
@h3rmanj Finally had a chance to spin this up and try it out. It mostly works but I'm seeing an issue with the logout server action. It doesn't seem to rerender the page until I manually refresh and API calls are failing in the console (expected but the calls shouldn't be made at all after a logout)

h3rmanj:
> @h3rmanj Finally had a chance to spin this up and try it out. It mostly works but I'm seeing an issue with the logout server action. It doesn't seem to rerender the page until I manually refresh and API calls are failing in the console (expected but the calls shouldn't be made at all after a logout)

I've upgraded next from `13.4.4` to `13.4.9`, which seems to have fixed the issue with the page not revalidating properly after server action logout. Not sure which API calls you mean, but I don't see any now at least. Did the errors pop up in the browser console or the server console?

arolariu:
Trying to revive this PR discussion...

I've ran the sample on my machine - LGTM.
I've also upped the `msal-node` package to 2.2.0 and the `next` package to 13.5.4 - all works fine.
I did not need to do anything else - no breaking changes detected. Cool!

Really great job, @h3rmanj. Thank you for your dedication on this aspect.
I have some newbie questions that maybe you could answer...:
- Do we still need `createCookieSessionStorage` from Remix? Is there any workaround, maybe something built into Next.JS - what about: https://nextjs.org/docs/app/api-reference/functions/cookies ?
- Do we still need the `server-only` package? Could you expand a bit on this topic, how it works, why is it needed..?
- I see that we also have some other dependencies on Remix Run, such as the `Session` type on the SessionPartitionManager.ts file.. Could you also expand on this, are there any NextJS alternatives?

I would also be interested in seeing how this example can be expanded with other IdPs such as GitHub, Google; do you have any tips on how complex it would be, to integrate other accounts besides MSAs, in your example? 

My knowledge on the front-end domain is rather limited, as I am not a front-end developer, and my questions are out of sheer curiosity.

bgavrilMS:
This looks good to me from the confidential client perspective. I am not sure Redis back-end storage is mandatory here, as that's a high bar of entry, we should maybe describe how to use this without a persistent back-end storage. 

I did not review the front-end code.

bgavrilMS:
Where is the session management done? We should advise (in a comment) that sessions should be set to expire every now and then (e.g. 60 min), because this will force the user to re-login (even if they don't see it, AAD will mostly redirect silently). The re-login aspect is important, as it ensures that conditional access policies are re-evaluated periodically.

bgavrilMS (approved):

h3rmanj:
Been some time since I touched this! The ecosystem for React Server Components and Next.js is moving fast, but I'll try to answer with what I encountered when setting this up in June.

h3rmanj:
@arolariu 

> Do we still need createCookieSessionStorage from Remix? Is there any workaround, maybe something built into Next.JS - what about: https://nextjs.org/docs/app/api-reference/functions/cookies ?

I am using built-in `headers` and `cookies`. However, those just return the raw values of a cookie. If I were to use them alone, the application would be open to some huge security issues. Since the token cache lives on the server side, I need to trust some input to access the users partition of the cache. The solution is to sign the cookies the application set.

I tried to explore iron-session, but the current version only supports the Pages Router of Next.js, while this sample utilizes the App Router. There is a [v8 in progress](https://github.com/vvo/iron-session/issues/586) which is supposed to support App Router, but it's undocumented and not released yet.

Since this is a sample, I don't really want it to own the logic of signing cookies manually, while adhering to rules of how many bytes a cookie is allowed to be etc. `createCookieSessionStorage` from remix is the only solution I'm familiar with, and works in isolation, which ironically is exactly what I need to work directly with the built-in `headers` and `cookie` apis from next.

As I've written before I'm open to suggestions here though, as it is kinda weird.

> Do we still need the server-only package? Could you expand a bit on this topic, how it works, why is it needed..?

Next.js App Router utilizes React Server Components. This allows you to write React components that either run on the server, client or both. This allows you to do some really cool stuff, but you have to be careful not to leak server-side secrets into the client bundle. A good example of this is the `src/serverConfig.ts` file, I don't want that file with all the secrets to explicitly or accidentaly end up in the client. By importing the `server-only` package, next will refuse to build the application if that were to happen.

[Docs about this](https://nextjs.org/docs/app/building-your-application/rendering/composition-patterns#keeping-server-only-code-out-of-the-client-environment)

> I would also be interested in seeing how this example can be expanded with other IdPs such as GitHub, Google; do you have any tips on how complex it would be, to integrate other accounts besides MSAs, in your example?

Are you thinking about Azure B2C? I haven't set it up myself so I'm not sure if you have to do anything special on the application side. But I'd guess you'll be able to configure this sample the same way you would configure any other `@azure/msal-node` application to use B2C.

h3rmanj:
@bgavrilMS

> I am not sure Redis back-end storage is mandatory here, as that's a high bar of entry, we should maybe describe how to use this without a persistent back-end storage.

I think I started with a FileCache here. But since it's a BFF framework, I'd say a distributed cache is a requirement to go to production, might as well have it in the sample.

> Where is the session management done? We should advise (in a comment) that sessions should be set to expire every now and then (e.g. 60 min), because this will force the user to re-login (even if they don't see it, AAD will mostly redirect silently). The re-login aspect is important, as it ensures that conditional access policies are re-evaluated periodically.

My thinking here was leaning on what `@azure/msal-node` does internally in the cache, and let the session expire with the refresh token. Are conditional access policies not re-evaluated if the server gets new id and access tokens with the refresh tokens?

The token caches and auth logic in this application is server-side, so the user will always have to do a full redirect to re-login, not something that can be silently done.

h3rmanj:
Updated to next.js 14, server actions are no longer experimental, so I removed the route handler alternatives. Also changed the calendar page to show consent button using nextjs error component.

h3rmanj:
I'm closing this for now, as there seems to be a lack of interest.

I will revisit this, as this is close to the technical direction react is heading, and how we'll use it at our company. Hopefully either `next-auth` or `iron-session` will be out of alpha/beta stage by then, and the sample be more attractive to you.

renchris:
@h3rmanj iron-session released v8 to the main stable branch this past Friday https://github.com/vvo/iron-session/releases/tag/v8.0.0

h3rmanj:
Oh, I hadn't seen that! I don't have bandwidth to look at it this year, but I welcome anyone who wants to continue working on this iteration of the sample. I'll probably do another pass at this in the start of the next year.

dmitrc:
Thank you for the great effort, @h3rmanj!
Hope this PR can regain traction, since this sample is pretty essential to cover the direction where React ecosystem is heading.

PR #6084, opened 2023-06-01
Prebuild extension binaries

### LINKED ISSUE #3332: @azure/msal-node-extensions doesn't have pre-compiled binaries
## Library
- [x] `@azure/msal-node-extensions@1.0.0-alpha.6`

## Description

`@azure/msal-node-extensions@1.0.0-alpha.6` builds its binaries when the package gets installed by a developer. As the readme says, this means that developers will need to have `node-gyp`, which in turn has considerable requirements on each platform. For example, on Windows, `node-gyp` requires Visual Studio with C++, which means `@azure/msal-node-extensions@1.0.0-alpha.6` is requiring Windows users to install a separate dependency that weighs several gigabytes on its own.

Another issue of not having pre-compiled binaries is that if a developer installs `@azure/msal-node-extensions@1.0.0-alpha.6` on Node 14 and then they switch to a different version of Node (let's say, Node 15), once they try to use persistence they get an error similar to the following:

```
Error: The module '/home/user/workspace/azure-sdk-for-js/common/temp/node_modules/.pnpm/@azure/msal-node-extensions@1.0.0-alpha.6/node_modules/@azure/msal-node-extensions/build/Release/dpapi.node'
was compiled against a different Node.js version using
NODE_MODULE_VERSION 88. This version of Node.js requires
NODE_MODULE_VERSION 83. Please try re-compiling or re-installing
the module (for instance, using `npm rebuild` or `npm install`).
```

Many users won't be able to consider this dependency while it requires this setup. Ideally the binaries would come pre-compiled, specially for developers using this library on production environments.

## Source

- [x] Internal (Microsoft)
- [ ] Customer request

tnorling:
- Updates build process for node-extensions to prebuild and include Windows binaries in the published package
- Updates test pipeline to run tests against each Windows, Mac and Linux

This saves consumers of the package from having to install pre-requisite build tooling and spend time compiling the native components.


konstantin-msft (approved):

extensions/msal-node-extensions/src/persistence/PersistenceCreator.ts
```diff
@@ -51,11 +51,8 @@ export class PersistenceCreator {
             throw PersistenceError.createNotSupportedError(
                 "The current environment is not supported by msal-node-extensions yet.");
         }
-
-        // Initially suppress the error thrown during persistence verification to allow us to fallback to plain text
-        const isPersistenceVerified = await peristence.verifyPersistence().catch(() => false);
-
-        if (!isPersistenceVerified) {
+        
+        await peristence.verifyPersistence().catch(async (e) => {
```
shylasummers, line 55: What is persistence verification?

tnorling: `msal-node` stores all cached artifacts "in-memory" by default, meaning, they are stored in an object which is destroyed when the `PublicClientApplication` object is destroyed (does not persist). In order to offer persistent cache storage we built `msal-node-extensions` which is a plugin that provides methods for `msal-node` to store its cache "on disk". In other words, an encrypted file which can be accessed by many instances of `PublicClientApplication` across processes.

The `verifyPersistence` method is just confirming that the content we wrote to the file was actually written and can be read back.

hectormmg (approved):

hectormmg:
LGTM, would be great to confirm the windows pool is actually used on a pipeline test run

.pipelines/OneBranch.Release.yml
```diff
@@ -161,6 +162,7 @@ extends:
             dependsOn: 'publish_msal_common'
             publishOverride: ${{ parameters.overrideMsalNodeExtensions }}
             testMode: ${{ parameters.testMode }}
+            poolType: 'windows'
```
sameerag, line 165: one branch is only for windows?

sameerag (approved):

.pipelines/OneBranch.Release.yml
tnorling: Extensions must be published using a windows machine since we are now prebuilding the windows binaries. All other libs will be published using an ubuntu machine

PR #6102, opened 2023-06-07
Extend proactive token refresh to client_credentials

### LINKED ISSUE #6085: Extend proactive token refresh to client_credentials and OBO
### Core Library

MSAL Node (@azure/msal-node)

### Wrapper Library

Not Applicable

Confidential

### Description

Proactive token refresh is an ESTS resiliency feature. MSAL libraries will receive a refresh_in parameter alongside expires_in in the token response. 

MSAL is expected to refresh the token when the refresh_in expires. This is already implemented for `AcquireTokenSilent` but is not implemented for `client_credentials`  and for `obo` flows, which look into the cache on their own. (MSAL JS doesn't use RT in OBO flows yet)

There is a spec but it is not great except for the acceptance tests - https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=/RefreshAtExpirationPercentage/overview.md&version=GBdev&anchor=key-case-sensitivity

Have a look at how "refresh_in" property is used today.

### Source

Internal (Microsoft)

Robbie-Microsoft:
MSAL is expected to refresh the token when the refresh_in value expires. This is already implemented for AcquireTokenSilent but is not implemented for client_credentials. This PR implements refresh_in for client_credentials.

I've implemented the [long term solution](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=/RefreshAtExpirationPercentage/overview.md&version=GBdev&_a=preview&anchor=long-term-solution) from the [refresh spec](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=/RefreshAtExpirationPercentage/overview.md).


lib/msal-node/src/client/ClientCredentialClient.ts
```diff
@@ -89,6 +89,21 @@ export class ClientCredentialClient extends BaseClient {
             this.serverTelemetryManager?.setCacheOutcome(
                 CacheOutcome.CACHED_ACCESS_TOKEN_EXPIRED
             );
+
+            return null;
+        // must refresh due to the refresh_in value
+        } else if (
```
bgavrilMS: I don't think this is the correct implementation, because it doesn't help with availability. If ESTS goes down during the token refresh phase, it kills the app. Please read the spec, particularly the acceptance tests.

## The scenario that `refresh_in` is trying to protect against is the following: 

1. app gets a token from ESTS that expires in 60 min and it uses the token over and over (from the cache)
2. cached token expires and request goes to ESTS but **ESTS is down** 
3. app is no longer able to function, because it doesn't have a token to call some downstream api

## With the introduction of refresh_in, the way it's supposed to work is: 

1. app gets a token from ESTS that expires in 60 min and with refresh_in=30min, at 12:00
2. between 12:00 and 12:30, the app can get the token out of the cache, just like before
3. at 12:40 app makes a request for the token. MSAL gives the app the token from the cache, but at the same time it sends a request to ESTS for a new token. Now the token is fresh.

With this, a fresh token is always in the cache, as long as the app uses it frequently. If ESTS goes down at step 3 above, MSAL should just ignore that and return the token from the cache.

Definition of "ESTS is down" is that it responds with 429 or 5xx errors (maybe other HTTP errors), but not with other 4xx.

## Implementation details

There are 2 ways to implement this: 

1. When you detect that refresh_on has elapsed, return the token from cache and queue a background request to ESTS. If this request fails, just log the failure, as there's nothing that can be done. This is preferred way of doing it, but I am not sure if this is possible in NodeJS, as you can't queue background threads?
OR
2. When you detect that refresh_on has elapsed but it's not yet expired, make a call to ESTS. If that call fails because ESTS is unavailable, return the cached token (it's not expired).

lib/msal-node/test/client/ClientCredentialClient.spec.ts
```diff
@@ -982,6 +982,88 @@ describe("ClientCredentialClient unit tests", () => {
         expect(authResult.state).toHaveLength(0);
     });
 
+    it("re-acquires a token via a network request, because the cached token's refreshOn value is expired", async () => {
```
bgavrilMS: Test doesn't take into account ESTS being down, as per acceptance tests in the spec.

bgavrilMS (changes requested):
Not implemented as availability feature.

lib/msal-node/src/client/ClientCredentialClient.ts
```diff
@@ -73,14 +73,14 @@ export class ClientCredentialClient extends BaseClient {
...
         if (!cachedAccessToken) {
             this.serverTelemetryManager?.setCacheOutcome(
                 CacheOutcome.NO_CACHED_ACCESS_TOKEN
             );
             return null;
-        }
-
-        if (
+        // must refresh due to the expires_in value
+        } else if (
```
bgavrilMS: You don't need any of these `else if` because you already `return null`. Having fewer branches makes the code easier to read. Have a look at https://learn.microsoft.com/en-us/visualstudio/code-quality/code-metrics-cyclomatic-complexity?view=vs-2022

lib/msal-node/test/client/ClientCredentialClient.spec.ts
```diff
@@ -982,6 +982,88 @@ describe("ClientCredentialClient unit tests", () => {
...
+        sinon
+            .stub(Authority.prototype, <any>"getEndpointMetadataFromNetwork")
+            .resolves(DEFAULT_OPENID_CONFIG_RESPONSE.body);
+        sinon
+            .stub(
```
bgavrilMS, line 700: In terms of tests, you can extend the coverage if instead of mocking some method `executePostToTokenEndpoint` (which btw will break if on rename), to injecting custom HTTP responses or mocking at the HTTP Client level.

Otherwise, please point to tests that show that `executePostToTokenEndpoint` behave like your mock states when getting 500 errors from STS.

lib/msal-node/src/client/ClientCredentialClient.ts
```diff
@@ -193,6 +213,23 @@ export class ClientCredentialClient extends BaseClient {
                 headers,
                 thumbprint
             );
+
+            // check if a new token is being requested because the cached token needs to be refreshed, not because it's expired
+            if (accessTokenCacheKey) {
+                // check if ESTS is down (status code 429 or 5xx)
+                if ((response.status === 429) || (response.status >= 500)) {
```
bgavrilMS, line 220: Yes, this is good. I'd put it in a private method. Can we extend this condition to cover Http exceptions such as the ones you get there's no internet on the box ? I believe those exceptions do not result in http responses at all.

lib/msal-node/src/client/ClientCredentialClient.ts
```diff
@@ -193,6 +213,23 @@ export class ClientCredentialClient extends BaseClient {
...
+                    this.logger.warning(
+                        "ClientCredentialClient:executeTokenRequest - AAD is currently unavailable and is unable to refresh the token."
+                    );
+                // ESTS is not down, the cached token is bad and should be removed from the cache
+                } else {
+                    this.logger.error(
+                        "ClientCredentialClient:executeTokenRequest - AAD is currently available but is unable to refresh the token. Returning the unexpired token and removing it from the cache."
+                    );
+                    this.cacheManager.removeAccessToken(accessTokenCacheKey);
```
bgavrilMS: I can't think of a scenario where you'd want to remove an access token? Generally MSAL relies on access tokens being replaced.

lib/msal-node/src/client/ClientCredentialClient.ts
```diff
@@ -89,7 +89,26 @@ export class ClientCredentialClient extends BaseClient {
             this.serverTelemetryManager?.setCacheOutcome(
                 CacheOutcome.CACHED_ACCESS_TOKEN_EXPIRED
             );
+
             return null;
+        // the token is not expired, but it must be refreshed due to the refresh_in value
+        } else if (
```
bgavrilMS: The approach is good, but the structure of the code is not. These private methods should do one thing and one thing alone, .e.g. `getCachedAuthenticationResult` should touch the cache, but now it also makes HTTP requests.

A possible implementation:

- `acquireToken` is the orchestrator of the logic
- it gets a cached access token just like before and checks the CacheOutcome
- if the CacheOutcome is REFRESH_CACHED_ACCESS_TOKEN then it keep hold of the "oldAuthenticationResult" and it calls `executeTokenRequest`
- if `executeTokenRequest` succeeds, return its result
- if `executeTokenRequest` fails for any reason other than 4xx errors, it returns `oldAuthenticationResult`
- if `executeTokenRequest` fails for 4xx errors, let the exception bubble through

lib/msal-node/src/client/ClientCredentialClient.ts
bgavrilMS: Having thought about this some more, I think the safest is to check on any exception and error code except 4xx

This is because the AppTokenProvider logic should also be taken into account.

lib/msal-node/src/client/ClientCredentialClient.ts
Robbie-Microsoft: I see this bullet in the long term solution in the 'refresh access token' spec: ```If MsalServiceException is thrown and the AAD is available, log the error and remove the entry from the cache```.

In ```executeTokenRequest```, the token only gets replaced if ```responseHandler.validateTokenResponse(serverTokenResponse)``` doesn't throw an error. Immediately after that code ```responseHandler.handleServerTokenResponse``` is executed which replaces the token. According to the wording of that bullet point above, I think that the token should be removed if ```responseHandler.validateTokenResponse(serverTokenResponse)``` (aka bad token) throws an error.

lib/msal-node/src/client/ClientCredentialClient.ts
```diff
@@ -58,7 +59,20 @@ export class ClientCredentialClient extends BaseClient {
 
         const cachedAuthenticationResult =
             await this.getCachedAuthenticationResult(request);
+
         if (cachedAuthenticationResult) {
+            // if the token is not expired but must be refreshed; get a new one in the background
+            if (this.serverTelemetryManager?.getCacheOutcome()) {
```
bgavrilMS: Did you mean to check a specific outcome?

bgavrilMS: nit: you can return a tuple (token, needs_refersh) from `getCachedAuthenticationResult` to not rely on telemetry.

lib/msal-node/src/client/ClientCredentialClient.ts
bgavrilMS: Let's to skip this part as it can have unintended consequences. Removing tokens from cache is not simple (particularly for user flows, because you need to remove RT and IDT and Accounts as well). 

If AAD is available and it issues a 4xx error (e.g. cert expired), the error bubbles up to the app. That's a good enough signal. 

error does not mean cached token is bad.

bgavrilMS: feel free to change the spec on this one

lib/msal-node/src/client/ClientCredentialClient.ts
```diff
@@ -214,23 +222,41 @@ export class ClientCredentialClient extends BaseClient {
                 thumbprint
             );
 
-            // check if a new token is being requested because the cached token needs to be refreshed, not because it's expired
-            if (accessTokenCacheKey) {
-                // check if ESTS is down (status code 429 or 5xx)
-                if ((response.status === 429) || (response.status >= 500)) {
+            serverTokenResponse = response.body;
```
bgavrilMS, line 257: null ref if code goes into block at line 174

lib/msal-node/src/client/ClientCredentialClient.ts
```diff
@@ -214,23 +222,41 @@ export class ClientCredentialClient extends BaseClient {
...
+
+            // if a new token is being requested due to the refresh_in value
+            if (accessTokenRefresh) {
```
bgavrilMS: As stated before, the logic should not be here. It should be in `acquireToken`. This method does too much and at the same time it doesn't cover errors from block at L174

lib/msal-node/src/client/ClientCredentialClient.ts
```diff
@@ -58,7 +59,20 @@ export class ClientCredentialClient extends BaseClient {
...
         if (cachedAuthenticationResult) {
+            // if the token is not expired but must be refreshed; get a new one in the background
+            if (this.serverTelemetryManager?.getCacheOutcome()) {
+                this.logger.info(
+                    "ClientCredentialClient:getCachedAuthenticationResult - Cached access token's refreshOn property has been exceeded'. It's not expired, but must be refreshed."
+                );
+
+                // return the cached token, don't wait for the result of this request
+                const accessTokenRefresh = true;
+                this.executeTokenRequest(request, this.authority, accessTokenRefresh);
```
bgavrilMS: How about something like 

```ts
if (cacheOutcome == REFRESH) {
     bool serverError = false;
      try
      {
              var newResponse =  this.executeTokenRequest(request, this.authority);
              estsOutage = newResponse.Status >= 500 && newResponse.Status < 600;
      }
      catch (Exception e) // other exceptions like HTTP stack e.g. SocketTimeout
      {
            Log(e);
            estsOutage = true;
      }
      
      // later
     if (estsOutage) { return cachedAuthenticationResult; } // still valid
     else { return newResponse;} // upper layer should transform this into correct exception
}

```

lib/msal-node/src/client/ClientCredentialClient.ts
```diff
@@ -214,23 +222,41 @@ export class ClientCredentialClient extends BaseClient {
...
                     );
+
+                    // we already know there is a valid cached access token
+                    const cachedAccessToken = this.readAccessTokenFromCache();
+                    // get its cache key
+                    const accessTokenCacheKey = (cachedAccessToken as AccessTokenEntity).generateCredentialKey();
+                    // and remove it from the cache
                     this.cacheManager.removeAccessToken(accessTokenCacheKey);
+
+                    const errString = `${serverTokenResponse.error_codes} - [${serverTokenResponse.timestamp}]: ${serverTokenResponse.error_description} - Correlation ID: ${serverTokenResponse.correlation_id} - Trace ID: ${serverTokenResponse.trace_id}`;
```
bgavrilMS: This is not the right place to transform server token response into errors. It makes the code brittle as we have multiple places that do that.

lib/msal-node/src/client/ClientCredentialClient.ts
Robbie-Microsoft: I did, oops. Didn't adjust the unit tests yet so I didn't catch this error. Also, you recommended checking the cache outcome?

bgavrilMS: Yes, but I'm saying that I don't trust `this.serverTelemetryManager?.getCacheOutcome()` - it might be doing some unexpected stuff like sampling or aggregations in there. Easier to have the private function return a tuple (token, needs_refresh ?)

lib/msal-node/src/client/ClientCredentialClient.ts
```diff
@@ -58,7 +58,20 @@ export class ClientCredentialClient extends BaseClient {
...
         if (cachedAuthenticationResult) {
+            // if the token is not expired but must be refreshed; get a new one in the background
+            if (this.serverTelemetryManager?.getCacheOutcome() === CacheOutcome.REFRESH_CACHED_ACCESS_TOKEN) {
+                this.logger.info(
+                    "ClientCredentialClient:getCachedAuthenticationResult - Cached access token's refreshOn property has been exceeded'. It's not expired, but must be refreshed."
+                );
+
+                // return the cached token, don't wait for the result of this request
+                const accessTokenRefresh = true;
+                this.executeTokenRequest(request, this.authority, accessTokenRefresh);
```
bgavrilMS: @Robbie-Microsoft @derisen - what happens with background request here? Does NodeJS keep the thread alive until the request finishes and the promise returns?

bgavrilMS (changes requested):
Code looks good now, please add the acceptance tests.

lib/msal-node/src/client/ClientCredentialClient.ts
Robbie-Microsoft: Instead of a tuple, how about a private class variable: lastCacheOutcome ?

lib/msal-node/src/client/ClientCredentialClient.ts
Robbie-Microsoft: yep

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -133,16 +133,37 @@ export class ResponseHandler {
...
     ): void {
         // Check for error
         if (
             serverResponse.error ||
             serverResponse.error_description ||
             serverResponse.suberror
         ) {
+
+            // check if 500 error
+            if (accessTokenRefresh && serverResponse.status && String(serverResponse.status)[0] === "5") {
```
tnorling: Magic string - please put the possible serverResponse statuses in an enum

lib/msal-node/src/client/ClientCredentialClient.ts
```diff
@@ -58,7 +59,22 @@ export class ClientCredentialClient extends BaseClient {
...
+            }
+
+            this.lastCacheOutcome = "";
```
tnorling: Please type lastCacheOutcome to the CacheOutcome enum and add a default/empty option if it doesn't already exist

change/@azure-msal-common-b0f24abd-2c63-4840-990b-29ee443b640a.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Extend proactive token refresh to client_credentials",
```
tnorling:
```suggestion
  "comment": "Extend proactive token refresh to client_credentials #6102",
```

change/@azure-msal-common-b0f24abd-2c63-4840-990b-29ee443b640a.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
tnorling:
```suggestion
  "type": "prerelease",
```

change/@azure-msal-node-9b01df06-64f9-4f34-8a6b-0bb05ab12f14.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
tnorling:
```suggestion
  "type": "prerelease",
```

change/@azure-msal-node-9b01df06-64f9-4f34-8a6b-0bb05ab12f14.json
```diff
@@ -0,0 +1,7 @@
...
+  "comment": "Extend proactive token refresh to client_credentials",
```
tnorling:
```suggestion
  "comment": "Extend proactive token refresh to client_credentials #6102",
```

lib/msal-common/src/response/ResponseHandler.ts
derisen: +1 

Also, why check the first character only?

lib/msal-node/src/client/ClientCredentialClient.ts
derisen: > this.lastCacheOutcome = "";

there is an empty string constant somewhere if I'm not mistaken

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -133,16 +133,37 @@ export class ResponseHandler {
     /**
      * Function which validates server authorization token response.
      * @param serverResponse
+     * @param accessTokenRefresh
      */
     validateTokenResponse(
-        serverResponse: ServerAuthorizationTokenResponse
+        serverResponse: ServerAuthorizationTokenResponse,
+        accessTokenRefresh?: boolean
```
derisen: nit: refreshAccessToken (verb followed by noun) seems more grammatical to me (even better perhaps: `refreshCachedAccessToken`)

lib/msal-node/src/client/ClientCredentialClient.ts
derisen: Nodejs uses a single thread (the event loop) to execute JS code (+k worker threads, but used mainly for things like GC and disk access etc. unless explicitly created). So it will invoke the `executeTokenRequest` but will not wait for it to complete and continue with statements below

lib/msal-common/src/response/ResponseHandler.ts
bgavrilMS: nit: another way to name bool is to use the verb `is` or the plural `are`. In this case `isTokenFromRefreshIn`

lib/msal-common/src/response/ResponseHandler.ts
bgavrilMS: @Robbie-Microsoft wants to check that the error code is between 500 and 599, which indicates there's something wrong with the STS. 

Not sure about an enum, since there could be new HTTP errors in the 500 range that are defined in the future. 

http status code should be a number, so int comparison is best here.

lib/msal-node/src/client/ClientCredentialClient.ts
bgavrilMS: @derisen - I get that it will just go to line 73, but what happens to the `executeTokenRequest` ? It's a long running async operation, will the current thread be killed before it's complete?

Scenario is: 

1. a request is made to the app and nodejs assigns a thread to it
1. app makes a token request
1. MSAL figures out that a refresh is needed and executes a token fetch, but it does **not** await it (!)
1. application is given a token
1. application uses the token to do stuff and finishes the requesst

Point 5 might or might not happen before the token acquisition (at step 3) is complete. Does NodeJS guarantee to not kill the thread and wait for step 3 to complete? (ASP.NET Core does this btw).

From looking at SO, I see that most ppl point to use https://nodejs.org/api/worker_threads.html#worker-threads to schedule step 3. 

CC @pmaytak

lib/msal-node/test/client/ClientCredentialClient.spec.ts
```diff
@@ -982,6 +982,159 @@ describe("ClientCredentialClient unit tests", () => {
...
+                ClientCredentialClient.prototype,
+                <any>"readAccessTokenFromCache"
+            )
+            .returns(expectedAtEntity);
+
+        const authResult = (await client.acquireToken(
+            clientCredentialRequest
+        )) as AuthenticationResult;
+        
+        const expectedScopes = [TEST_CONFIG.DEFAULT_GRAPH_SCOPE[0]];
```
bgavrilMS, line 864: How do you know that the background request was kicked off and that the log warning was issued?

lib/msal-node/test/client/ClientCredentialClient.spec.ts
```diff
@@ -982,6 +982,159 @@ describe("ClientCredentialClient unit tests", () => {
...
+            )
+        ).toBe(true);
+        expect(
+            returnVal.includes(
+                `${AADServerParamKeys.GRANT_TYPE}=${GrantType.CLIENT_CREDENTIALS_GRANT}`
+            )
+        ).toBe(true);
+        expect(
+            returnVal.includes(
+                `${AADServerParamKeys.CLIENT_SECRET}=${TEST_CONFIG.MSAL_CLIENT_SECRET}`
```
bgavrilMS: I don't get the point of most of these asserts. They make the test hard to read and are not relevant to the scenario. Can you use a private method to hide them?

lib/msal-node/test/client/ClientCredentialClient.spec.ts
```diff
@@ -982,6 +982,159 @@ describe("ClientCredentialClient unit tests", () => {
...
+        
+        const expectedScopes = [TEST_CONFIG.DEFAULT_GRAPH_SCOPE[0]];
+        expect(authResult.scopes).toEqual(expectedScopes);
+        expect(authResult.accessToken).toEqual("an_access_token");
+        expect(authResult.account).toBeNull();
+        expect(authResult.fromCache).toBe(true);
+        expect(authResult.uniqueId).toHaveLength(0);
+        expect(authResult.state).toHaveLength(0);
+
+        expect(createTokenRequestBodySpy.calledWith(clientCredentialRequest)).toBe(true);
```
bgavrilMS: how do you know that a background token request was made and that the access token from the cache has been updated?

lib/msal-node/src/client/ClientCredentialClient.ts
derisen: > Point 5 might or might not happen before the token acquisition (at step 3) is complete. Does NodeJS guarantee to not kill the thread and wait for step 3 to complete? (ASP.NET Core does this btw).

That's the thing though, there's only one thread (event loop) and it's alive as long as the app runs. In other words, no new thread is created for the network call here -it will be queued in the event loop.

> From looking at SO, I see that most ppl point to use https://nodejs.org/api/worker_threads.html#worker-threads to schedule step 3.

Worker threads are recommended for CPU intensive tasks. As mentioned in the docs, they don't provide much advantage for IO operations.

So I believe it's guaranteed, but @tnorling might have more insight on this.

tnorling: The node process will not close until the event loop is empty (unless the process is explicitly killed)

tnorling: Probably good to have test coverage for this scenario though

lib/msal-common/src/response/ResponseHandler.ts
sameerag: I like `isTokenFromRefreshIn`, more relevant name

lib/msal-common/src/response/ResponseHandler.ts
sameerag: Can we just create a constant then, which says, `ServerErrorRangeHint` or something?

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -133,16 +133,37 @@ export class ResponseHandler {
...
+            // check if 500 error
+            if (accessTokenRefresh && serverResponse.status && String(serverResponse.status)[0] === "5") {
+                this.logger.warning(
+                    "executeTokenRequest:validateTokenResponse - AAD is currently unavailable and the access token is unable to be refreshed."
+                );
+                
+                // don't throw an exception, but alert the user via a log that the token was unable to be refreshed
+                return;
+            // check if 400 error
+            } else if (accessTokenRefresh && serverResponse.status && String(serverResponse.status)[0] === "4") {
```
sameerag: Same as above

lib/msal-node/src/client/ClientCredentialClient.ts
sameerag: There is: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/6f5a21499f915fd5a49515c4d0927bf4be81fd07/lib/msal-common/src/utils/Constants.ts#L46

lib/msal-node/src/client/ClientCredentialClient.ts
sameerag: So, we are updating the `accessToken` in the background and is the below use case valid in this case: 

1. access_token fetched
2. App uses the access_token 
3. background refresh initiated
4. App calls atSilent, gets the old token, which can be expired between 3 and 4 - fails get interaction needed error
5. background fetch successful
6. App retries after step 4 and now gets a valid token??

lib/msal-node/test/client/ClientCredentialClient.spec.ts
sameerag: We never had `http` level mocking. I remember we scheduled it, never got around it.

lib/msal-common/src/response/ResponseHandler.ts
tnorling: Agreed, int comparison would be better than converting to string in this case

lib/msal-node/test/client/ClientCredentialClient.spec.ts
Robbie-Microsoft: On line 996 we declare that we're spying on the function createTokenRequestBody which will be called when a network request is made. On line 1041 we're listening to the spy and seeing if it was called.

lib/msal-node/test/client/ClientCredentialClient.spec.ts
Robbie-Microsoft: They ensure that the createTokenRequestBodySpy function was called and a valid response was received. Basically, they ensure that the network request to refresh the token is successful. I could put them in a private function, and we'd need another PR to clean up the several other tests in this file that use these assertions.

lib/msal-node/test/client/ClientCredentialClient.spec.ts
Robbie-Microsoft: See other comment regarding how we know the background token request was made.

lib/msal-common/src/utils/Constants.ts
```diff
@@ -64,6 +64,13 @@ export const Constants = {
     INVALID_INSTANCE: "invalid_instance",
 };
 
+export enum ERROR_CODE_BOUNDS {
```
tnorling: Please define this as an object literal instead (look at HeaderNames or PersistentCacheKeys a few lines below). We stopped using the native enums due to size concerns

Robbie-Microsoft: interesting!

lib/msal-common/src/response/ServerAuthorizationTokenResponse.ts
```diff
@@ -26,6 +26,7 @@ import { AuthenticationScheme } from "../utils/Constants";
  * - correlation_id: A unique identifier for the request that can help in diagnostics across components.
  */
 export type ServerAuthorizationTokenResponse = {
+    status?: number;
```
pmaytak, line 30: Add a comment above for this.

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -133,16 +134,47 @@ export class ResponseHandler {
...
             serverResponse.error ||
             serverResponse.error_description ||
             serverResponse.suberror
         ) {
+
+            // check if 500 error
+            if (
+                refreshAccessToken &&
+                serverResponse.status &&
+                (serverResponse.status >= ErrorCodeBounds.stsErrorLowerLimit) &&
```
bgavrilMS, line 154: nit: code was more readable before when numbers were harcoded.

lib/msal-node/test/client/ClientCredentialClient.spec.ts
bgavrilMS: Consider moving them to a private method to avoid repetition, at least in this test class.

lib/msal-node/src/client/ClientCredentialClient.ts
bgavrilMS: @sameerag - I think MSAL will **either** kick off a background fetch (refresh logic) **or** treat the cached token expired. Both cannot happen at the same time, since it's the call to `atSilent` / `atClientCreds` that initiates the entire logic.

bgavrilMS: Thanks for weighing in on this @tnorling @sameerag! 

@Robbie-Microsoft - please see https://nodejs.org/en/docs/guides/event-loop-timers-and-nexttick and consider doing a manual test first to confirm what @tnorling said. You can automate a test by adding a 1-2s timeout in the the refresh logic

lib/msal-common/src/request/RequestValidator.ts
```diff
@@ -93,7 +93,7 @@ export class RequestValidator {
         }
 
         // Remove any query parameters already included in SSO params
-        queryParams.forEach((value, key) => {
+        queryParams.forEach((_value, key) => {
```
bgavrilMS, line 96: is it because it's not used?

lib/msal-common/src/response/AuthenticationResult.ts
```diff
@@ -32,10 +33,11 @@ export type AuthenticationResult = {
     accessToken: string;
     fromCache: boolean;
     expiresOn: Date | null;
+    extExpiresOn?: Date;
+    refreshOn: Date | null;
```
bgavrilMS: What is the difference between

`refreshOn: Date | null;` 
and
`refreshOn?: Date;`

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -151,16 +152,47 @@ export class ResponseHandler {
...
+                (serverResponse.status <= ErrorCodeBounds.stsErrorUpperLimit)
+            ) {
+                this.logger.warning(
```
bgavrilMS, line 175: I'd go with error

bgavrilMS, line 183: Make sure to include as much information as possible about the error in the log message.

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -151,16 +152,47 @@ export class ResponseHandler {
...
+                // don't throw an exception, but alert the user via a log that the token was unable to be refreshed
+                return;
+            // check if 400 error
+            } else if (
+                refreshAccessToken &&
+                serverResponse.status &&
+                (serverResponse.status >= ErrorCodeBounds.clientErrorLowerLimit) &&
+                (serverResponse.status <= ErrorCodeBounds.clientErrorUpperLimit)
+            ) {
+                this.logger.warning(
```
bgavrilMS, line 196: Same as above. If a bug is raised and you see this log message, it is not sufficient to understand what's going on.

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -582,6 +615,9 @@ export class ResponseHandler {
             extExpiresOn = new Date(
                 Number(cacheRecord.accessToken.extendedExpiresOn) * 1000
             );
+            refreshOn = new Date(
+                Number(cacheRecord.accessToken.refreshOn) * 1000
```
bgavrilMS: What happens if "refreshOn" is null? This is quite possible: 

- ESTS doesn't emit refresh_in at all today
- ESTS will only emit refresh_in in certain cases anyway

bgavrilMS: (this is a blocking comment)

lib/msal-common/src/utils/Constants.ts
```diff
@@ -64,6 +64,14 @@ export const Constants = {
     INVALID_INSTANCE: "invalid_instance",
 };
 
+export const ErrorCodeBounds = {
+    clientErrorLowerLimit: 400,
```
bgavrilMS: 1. Naming of constants seems to be inconsistent. I see that constants are all UPPER_CASE 
2. Consider renaming the class to "HttpStatus" and the constants to `CLIENT_ERROR_LOWER_LIMIT` etc.

lib/msal-common/src/utils/Constants.ts
```diff
@@ -64,6 +64,14 @@ export const Constants = {
...
+    clientErrorUpperLimit: 499,
+    stsErrorLowerLimit: 500,
```
bgavrilMS: same, 500 is an http status, not an sts concept.

lib/msal-common/test/response/ResponseHandler.spec.ts
```diff
@@ -110,6 +110,7 @@ const testServerTokenResponse = {
         scope: TEST_CONFIG.DEFAULT_SCOPES.join(" "),
         expires_in: TEST_TOKEN_LIFETIMES.DEFAULT_EXPIRES_IN,
         ext_expires_in: TEST_TOKEN_LIFETIMES.DEFAULT_EXPIRES_IN,
+        refresh_in: TEST_TOKEN_LIFETIMES.DEFAULT_REFESH_IN,
```
bgavrilMS: I think it's better to have another response, because you want tests that cover both cases: 

- refresh_in is present in the response
- refresh_in is not present

Consider having a function `GetTestServerResponse(refresh_in? = 123)`

lib/msal-node/src/client/ClientCredentialClient.ts
```diff
@@ -59,7 +60,23 @@ export class ClientCredentialClient extends BaseClient {
...
+                    "ClientCredentialClient:getCachedAuthenticationResult - Cached access token's refreshOn property has been exceeded'. It's not expired, but must be refreshed."
+                );
+
+                // return the cached token, don't wait for the result of this request
+                const refreshAccessToken = true;
+                this.executeTokenRequest(request, this.authority, refreshAccessToken);
+            }
+
+            // reset the last cache outcome
+            this.lastCacheOutcome = CacheOutcome.NO_CACHE_HIT;
```
bgavrilMS, line 84: Isn't `lastCacheOutcome` used elsewhere? In case of "refresh_in" you are resetting it to "NO_CACHE_HIT" ?

bgavrilMS (changes requested):
RefreshOn is optional, not mandatory.

lib/msal-common/src/request/RequestValidator.ts
Robbie-Microsoft: Exactly. And I wasn't able to build msal-common to run the unit tests because this was throwing an eslint-unused-var error.

lib/msal-node/src/client/ClientCredentialClient.ts
Robbie-Microsoft: It's not used elsewhere. We set it in the function that attempts to get the token from the cache. We're just resetting the value here for the next time acquire token is called.

bgavrilMS (approved):

pmaytak (approved):

lib/msal-common/src/utils/Constants.ts
```diff
@@ -64,6 +64,17 @@ export const Constants = {
     INVALID_INSTANCE: "invalid_instance",
 };
 
+export const HttpStatus = {
```
sameerag, line 67: Unrelated to this PR, don't we have similar constants in `msal-node`? can we leverage these from common?

lib/msal-node/src/client/ClientCredentialClient.ts
```diff
@@ -59,7 +60,23 @@ export class ClientCredentialClient extends BaseClient {
...
             await this.getCachedAuthenticationResult(request);
+
         if (cachedAuthenticationResult) {
+            // if the token is not expired but must be refreshed; get a new one in the background
+            if (this.lastCacheOutcome === CacheOutcome.REFRESH_CACHED_ACCESS_TOKEN) {
+                this.logger.info(
+                    "ClientCredentialClient:getCachedAuthenticationResult - Cached access token's refreshOn property has been exceeded'. It's not expired, but must be refreshed."
+                );
+
+                // return the cached token, don't wait for the result of this request
```
sameerag, line 74: Can we edit this: 
// Refresh the access token in the background and return the current value

lib/msal-node/src/client/ClientCredentialClient.ts
```diff
@@ -35,6 +35,7 @@ import {
 export class ClientCredentialClient extends BaseClient {
     private scopeSet: ScopeSet;
     private readonly appTokenProvider?: IAppTokenProvider;
+    private lastCacheOutcome: CacheOutcome;
```
sameerag, line 38: No default value?

sameerag (approved):

lib/msal-common/src/utils/Constants.ts
Robbie-Microsoft: They were in msal-node. I moved them to msal-common.

lib/msal-node/src/client/ClientCredentialClient.ts
Robbie-Microsoft: no, because there hasn't been a cache search yet

PR #6104, opened 2023-06-07
ClientCredentialRequest Documentation Update

### LINKED ISSUE #6069: acquireTokenByClientCredential fails with multiple scopes
### Core Library

MSAL Node (@azure/msal-node)

### Core Library Version

1.17.2

### Wrapper Library

Not Applicable

### Wrapper Library Version

None

Confidential

### Description

acquireTokenByClientCredential indicates that multiple 'scopes' can be supplied, but using more than one scope gives an error.

```typescript
const token = await this.#cca.acquireTokenByClientCredential({
      scopes: [
        'https://graph.microsoft.com/.default',
        'https://xxx.sharepoint.com/.default'
      ]
    })
```

  ServerError: invalid_scope: 70011 - [2023-05-31 17:07:41Z]: AADSTS70011: The provided request must include a 'scope' input parameter. The provided value for the input parameter 'scope' is not valid. The scope https://graph.microsoft.com/.default https://xxx.sharepoint.com/.default is not valid.

### Error Message

  ServerError: invalid_scope: 70011 - [2023-05-31 17:07:41Z]: AADSTS70011: The provided request must include a 'scope' input parameter. The provided value for the input parameter 'scope' is not valid. The scope https://graph.microsoft.com/.default https://xxx.sharepoint.com/.default is not valid.

### Msal Logs

_No response_

### MSAL Configuration

```javascript
{
      auth: {
        clientId: this.#options.clientId,
        clientCertificate: {
          thumbprint: this.#options.clientCertFingerprint,
          privateKey: this.#options.clientCertKey
        },
        authority: `https://login.microsoftonline.com/${this.#options.tenantId}`
      }
    }
```

### Relevant Code Snippets

```javascript
const token = await this.#cca.acquireTokenByClientCredential({
      scopes: [
        'https://graph.microsoft.com/.default',
        'https://xxx.sharepoint.com/.default'
      ]
    })
```

### Reproduction Steps

see description

### Expected Behavior

multiple scopes supported

### Identity Provider

Azure AD / MSA

None (Server)

### Regression

_No response_

### Source

External (Customer)

Robbie-Microsoft:
This PR contains an update for the ClientCredentialRequest documentation that indicates that only one scope should be passed in.


change/@azure-msal-node-f9a3368f-1984-4a2c-97cd-85b238e3e141.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
derisen: shouldn't this be "none" since its a doc change?

lib/msal-node/src/request/ClientCredentialRequest.ts
```diff
@@ -7,7 +7,7 @@ import { CommonClientCredentialRequest } from "@azure/msal-common";
 
 /**
  * CommonClientCredentialRequest
- * - scopes                  - Array of scopes the application is requesting access to.
+ * - scopes                  - An array containing one (string) scope the application is requesting access to.
```
derisen:
```suggestion
 * - scopes                  - Array of scopes the application is requesting access to. Typically contains only the .default scope for a single resource. See: https://learn.microsoft.com/azure/active-directory/develop/scopes-oidc#the-default-scope
```

derisen: This rubs me in the wrong way, since it seems a contingent restriction to me -but I'll defer to @bgavrilMS

bgavrilMS: What do you propose we do @derisen ? AAD can only issue 1 token per resource. Maybe we should add an overload that takes only 1 string?

bgavrilMS (approved):
LGTM with suggestions from @derisen

bgavrilMS (approved):
LGTM with suggestions from @derisen

lib/msal-node/src/request/ClientCredentialRequest.ts
derisen: This is the case for delegated permissions too though, and we don't call it out specifically -eSTS just ignores the scopes for the 2..n resources and issues a token for the first one only. I feel like like it would be better for STS to align the behavior with application permissions with delegated permissions i.e. ignore the scopes for other resources instead of responding with an error (unless I'm missing something).

What I feel contingent is the single scope/string part, since multiple scopes != multiple resources. It's a single scope because we are using .default with client credentials, but in theory another IdP might allow ["resource1.scope1" and "resource1.scope2"]

change/@azure-msal-node-f9a3368f-1984-4a2c-97cd-85b238e3e141.json
Robbie-Microsoft:
```suggestion
  "type": "none",
```

change/@azure-msal-node-f9a3368f-1984-4a2c-97cd-85b238e3e141.json
```diff
@@ -0,0 +1,7 @@
...
+  "comment": "Updated ClientCredentialRequest documentation",
+  "packageName": "@azure/msal-node",
+  "email": "rginsburg@microsoft.com",
+  "dependentChangeType": "patch"
```
Robbie-Microsoft:
```suggestion
  "dependentChangeType": "none"
```

derisen (approved):

PR #6129, opened 2023-06-13
ignore native broker initialize check if application is not top-frame

lalimasharda:


lib/msal-browser/src/controllers/StandardController.ts
```diff
@@ -1299,11 +1299,13 @@ export class StandardController implements IController {
         // Block redirectUri opened in a popup from calling MSAL APIs
         BrowserUtils.blockAcquireTokenInPopups();
 
-        // Block token acquisition before initialize has been called if native brokering is enabled
-        BrowserUtils.blockNativeBrokerCalledBeforeInitialized(
+        // Block token acquisition before initialize has been called if native brokering is enabled in top-frame
+        if (setInteractionInProgress) {
+            BrowserUtils.blockNativeBrokerCalledBeforeInitialized(
             this.config.system.allowNativeBroker,
```
peterzenz: Nit:  Indent this line and the next

peterzenz (approved):
Note that small changes like this should have a test covering this feature explicitly.  As you look at the test automation plan, please keep this in mind.

lib/msal-browser/src/controllers/StandardController.ts
lalimasharda: Fixed!

lib/msal-browser/src/controllers/StandardController.ts
```diff
@@ -1299,11 +1299,13 @@ export class StandardController implements IController {
         // Block redirectUri opened in a popup from calling MSAL APIs
         BrowserUtils.blockAcquireTokenInPopups();
 
-        // Block token acquisition before initialize has been called if native brokering is enabled
-        BrowserUtils.blockNativeBrokerCalledBeforeInitialized(
-            this.config.system.allowNativeBroker,
-            this.initialized
-        );
+        // Block token acquisition before initialize has been called if native brokering is enabled in top-frame
+        if (setInteractionInProgress) {
```
sameerag: The check has always been `interactionInProgress` to identify an embedded frame I presume?

sameerag (approved):

lib/msal-browser/src/controllers/StandardController.ts
lalimasharda: Yes, I just need to change the naming.

sameerag (approved):
Makes it much clear. Thanks.

tnorling (approved):

PR #6139, opened 2023-06-19
Fix performance bug with regional authority detection

jo-arroyo:
This PR addresses a performance bug where the Azure region is still being auto-detected even when the Azure region is configured, for telemetry purposes. Refactored function will use Azure region where configured for performance advantages. This fix passes all existing unit tests for regional discovery and user provided region overrides.

Note: Although not all telemetry outcomes will be used after refactoring, `RegionDiscoveryOutcomes` constant remains the same to reflect spec-defined telemetry data points.


peterzenz (approved):

tnorling (approved):

change/@azure-msal-common-3cdc9d22-67a6-41c7-8d7c-6b42b43c8106.json
```diff
@@ -0,0 +1,7 @@
+{
```
sameerag, line 1: Do we need this in v2 also?

lib/msal-common/src/authority/Authority.ts
```diff
@@ -579,47 +580,38 @@ export class Authority {
...
-            if (autodetectedRegionName) {
-                this.regionDiscoveryMetadata.region_outcome =
-                    this.authorityOptions.azureRegionConfiguration
-                        ?.azureRegion === autodetectedRegionName
-                        ? RegionDiscoveryOutcomes.CONFIGURED_MATCHES_DETECTED
-                        : RegionDiscoveryOutcomes.CONFIGURED_NOT_DETECTED;
-            } else {
-                this.regionDiscoveryMetadata.region_outcome =
-                    RegionDiscoveryOutcomes.CONFIGURED_NO_AUTO_DETECTION;
+        if (userConfiguredAzureRegion) {
```
sameerag, line 585:
```suggestion
        if (userConfiguredAzureRegion && userConfiguredAzureRegion !== Constants.AZURE_REGION_AUTO_DISCOVER_FLAG) {
```

lib/msal-common/src/authority/Authority.ts
```diff
@@ -579,47 +580,38 @@ export class Authority {
...
+            if (userConfiguredAzureRegion !== Constants.AZURE_REGION_AUTO_DISCOVER_FLAG) {
+                this.regionDiscoveryMetadata.region_outcome = RegionDiscoveryOutcomes.CONFIGURED_NO_AUTO_DETECTION;
+                this.regionDiscoveryMetadata.region_used = userConfiguredAzureRegion;
```
sameerag, line 588: Do we cover for user passing an `invalid region`? How does the e2e flow work there? Does the service throw?

sameerag (approved):
Left few comments, approved.

change/@azure-msal-common-3cdc9d22-67a6-41c7-8d7c-6b42b43c8106.json
jo-arroyo: I will open a separate PR for this change in v2.

lib/msal-common/src/authority/Authority.ts
jo-arroyo: According to the spec it seems we just pass on the user input region, and it is up to the service to throw if the region is invalid.

PR #6163, opened 2023-06-26
Make claims-based caching configurable

hectormmg:
This PR:
- Sets ability to cache based on requested claims as a configurable option under `CacheOptions.claimsBasedCachinEnabled`
- Changes default behavior so in v3 claims-based caching is off by default and any request containing claims will go to the network
- Provides a mitigation to the issue where requesting different claims results in a cache miss and a cached token accumulation with no practical way to clear unused tokens
- Backport for LTS packages (msal-node v1/msal-browser v2) in progress


lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -499,8 +503,12 @@ export abstract class ClientApplication {
 
         authRequest.authenticationScheme = AuthenticationScheme.BEARER;
 
-        // Set requested claims hash if claims were requested
-        if (authRequest.claims && !StringUtils.isEmpty(authRequest.claims)) {
+        // // Set requested claims hash if claims-based caching is enabled and claims were requested
+        if (
```
bgavrilMS, line 507: Consider adding tests for both client_credentials and OBO, as both of these methods look into the cache on their own (not via acquire token silent).

lib/msal-common/test/client/SilentFlowClient.spec.ts
```diff
@@ -268,6 +268,120 @@ describe("SilentFlowClient unit tests", () => {
...
+                )
+                .resolves(DEFAULT_OPENID_CONFIG_RESPONSE.body);
+            sinon
+                .stub(AuthToken, "extractTokenClaims")
+                .returns(ID_TOKEN_CLAIMS);
+            sinon
+                .stub(CacheManager.prototype, "readAccountFromCache")
+                .returns(testAccountEntity);
+            sinon
+                .stub(CacheManager.prototype, "getIdToken")
```
bgavrilMS, line 340: Instead of mocking so much of MSAL's internal state, consider mocking at the HTTP level, similar to `getEndpointMetadataFromNetwork`. So a test would be: 

1. make token request with claims -> goes to STS
2. make second toke request with claims -> goes to STS when `claimsBasedCachingEnabled=false`

bgavrilMS (approved):

lib/msal-node/src/client/ClientApplication.ts
hectormmg: Yep, I opened this up for review before adding the node tests, sorry about that just trying to get ahead on the reviews browser-side.

lib/msal-browser/src/interaction_client/BaseInteractionClient.ts
```diff
@@ -178,8 +178,12 @@ export abstract class BaseInteractionClient {
             );
         }
 
-        // Set requested claims hash if claims were requested
-        if (request.claims && !StringUtils.isEmpty(request.claims)) {
+        // Set requested claims hash if claims-based caching is enabled and claims were requested
+        if (
+            this.config.cache.claimsBasedCachingEnabled &&
+            request.claims &&
+            !StringUtils.isEmpty(request.claims)
```
tnorling:
```suggestion
            request.claims
```

lib/msal-browser/test/cache/TokenCache.spec.ts
```diff
@@ -56,6 +56,7 @@ describe("TokenCache tests", () => {
             storeAuthStateInCookie: false,
             secureCookies: false,
             cacheMigrationEnabled: false,
+            claimsBasedCachingEnabled: false,
```
tnorling, line 59: Any new tests to validate the bug you're fixing here?

lib/msal-common/src/client/SilentFlowClient.ts
```diff
@@ -75,6 +76,15 @@ export class SilentFlowClient extends BaseClient {
                 "SilentFlowClient:acquireCachedToken - Skipping cache because forceRefresh is true."
             );
             throw ClientAuthError.createRefreshRequiredError();
+        } else if (
+            !this.config.cacheOptions.claimsBasedCachingEnabled &&
+            !StringUtils.isEmptyObj(request.claims)
```
tnorling, line 81:
```suggestion
            request.claims
```

tnorling, line 81: Is request.claims an object or a stringified object?

lib/msal-common/src/config/ClientConfiguration.ts
```diff
@@ -53,6 +54,7 @@ export type CommonClientConfiguration = {
     authOptions: Required<AuthOptions>;
     systemOptions: Required<SystemOptions>;
     loggerOptions: Required<LoggerOptions>;
+    cacheOptions: CacheOptions;
```
tnorling: `Required<CacheOptions>`?

lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -499,8 +503,12 @@ export abstract class ClientApplication {
...
+            this.config.cache.claimsBasedCachingEnabled &&
+            authRequest.claims &&
+            !StringUtils.isEmpty(authRequest.claims)
```
tnorling:
```suggestion
            authRequest.claims
```

lib/msal-common/src/client/SilentFlowClient.ts
hectormmg: Stringified object, this will skip the cache if the claims value is a valid stringified object that isn't "{}" (empty) and claims based caching isn't enabled.

lib/msal-node/src/client/ClientApplication.ts
hectormmg: Punting Client Credentials and OBO tests and bug fixes to a different PR over the next couple of weeks so we can get these changes merged for PCA before our next release.

lib/msal-node/src/client/ClientApplication.ts
hectormmg: Changing to check for empty stringified object instead, which I think was the original intention that I misinterpreted back when I first added claims-based caching

lib/msal-browser/test/cache/TokenCache.spec.ts
hectormmg: There's a few on `SilentFlowClient.spec.ts`, sorry for the massive diff, the automatic format on save has been doing that to all my PRs lately.

lib/msal-common/src/config/ClientConfiguration.ts
hectormmg: Updated. What's our criteria for which types we pull with Required and which we don't?

lib/msal-browser/src/interaction_client/BaseInteractionClient.ts
hectormmg: Fixed by switching to isEmptyObj and adding an explanatory comment.

lib/msal-browser/test/cache/TokenCache.spec.ts
hectormmg: They do cover the following cases:
1. Should go to network if claimsBasedCaching is off and claims are present
2. Should go to cache (and match) if claimsBasedCaching is on and claims are present
3. Should go to network after going to cache if claimsBasedCaching is on and claims are present but different from cached claims
4. Should go to cache if claimsBasedCaching is off but no claims are present in request, returning a matching previously cached token even if it has requested claims

lib/msal-common/src/config/ClientConfiguration.ts
tnorling: Customer facing = not `Required`, during initialization we assign defaults to everything so any internal references should be `Required`. This helps us by not needing to constantly check whether things are defined or not

lib/msal-common/src/config/ClientConfiguration.ts
```diff
@@ -108,6 +110,15 @@ export type LoggerOptions = {
     correlationId?: string;
 };
 
+/**
+ *  Use this to configure credential cache preferences in the ClientConfiguration object
+ *
+ * - claimsBasedCachingEnabled   - Sets whether tokens should be cached based on the claims hash. Default is true.
```
hectormmg:
```suggestion
 * - claimsBasedCachingEnabled   - Sets whether tokens should be cached based on the claims hash. Default is false.
```

lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -499,8 +503,13 @@ export abstract class ClientApplication {
 
         authRequest.authenticationScheme = AuthenticationScheme.BEARER;
 
-        // Set requested claims hash if claims were requested
-        if (authRequest.claims && !StringUtils.isEmpty(authRequest.claims)) {
+        // // Set requested claims hash if claims-based caching is enabled and claims were requested
```
hectormmg:
```suggestion
        // Set requested claims hash if claims-based caching is enabled and claims were requested
```

tnorling (approved):
Please note the change to the default value in the migration guide. Otherwise looks good

PR #6164, opened 2023-06-26
Flatten e2e test utils

tnorling:
npm workspaces doesn't seem to handle nested packages very well. To get our repo ready for workspaces this PR moves the e2e test utils to be a sibling of the samples rather than a parent


samples/msal-node-samples/jest.config.js
jo-arroyo, line 1: Out of curiosity, why is this being deleted?

tnorling: Each tested sample has its own config. This was here in case we wanted to run all tests/samples at the same time which we haven't been doing. Once we move to workspaces however, this will still be an option if needed by using the workspace commands

peterzenz (approved):

jo-arroyo (approved):

PR #6165, opened 2023-06-27
Remove sample dependency on msal-common

tnorling:
Some of our samples are taking direct dependency on msal-common. This PR refactors these samples to use exports from msal-browser/msal-node instead & adds additional exports to those libraries where needed.

### DIFF
--- a/change/@azure-msal-browser-1fe30015-5db8-41ae-988f-d27fa55e15e3.json
+++ b/change/@azure-msal-browser-1fe30015-5db8-41ae-988f-d27fa55e15e3.json
@@ -1,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "Export IdTokenClaims & PromptValue types",
+  "packageName": "@azure/msal-browser",
+  "email": "thomas.norling@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/samples/msal-node-samples/ElectronSystemBrowserTestApp/src/CustomLoopbackClient.ts
+++ b/samples/msal-node-samples/ElectronSystemBrowserTestApp/src/CustomLoopbackClient.ts
@@ -185,7 +185,7 @@ export class CustomLoopbackClient implements ILoopbackClient {
      * @param query
      */
     static queryStringToObject(query: string): ServerAuthorizationCodeResponse {
-        const obj: ServerAuthorizationCodeResponse = {};
+        const obj: {[key:string]:string} = {};
         const params = query.split("&");
         const decode = (s: string) => decodeURIComponent(s.replace(/\+/g, " "));
         params.forEach((pair) => {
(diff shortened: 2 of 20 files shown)


samples/msal-node-samples/ElectronSystemBrowserTestApp/src/CustomLoopbackClient.ts
```diff
@@ -59,12 +58,12 @@ export class CustomLoopbackClient implements ILoopbackClient {
                     res.end(errorTemplate || "Error occurred loading redirectUrl");
                     reject(new Error('Loopback server callback was invoked without a url. This is unexpected.'));
                     return;
-                } else if (url === CommonConstants.FORWARD_SLASH) {
+                } else if (url === "/") {
```
peterzenz, line 61: Are you removing the const as well?

konstantin-msft (approved):

change/@azure-msal-browser-1fe30015-5db8-41ae-988f-d27fa55e15e3.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "prerelease",
+  "comment": "Export IdTokenClaims & PromptValue types",
+  "packageName": "@azure/msal-browser",
+  "email": "thomas.norling@microsoft.com",
+  "dependentChangeType": "patch"
```
konstantin-msft, line 6: AFAIK `dependentChangeType` should be `prerelease` as well until we GA.

samples/msal-node-samples/ElectronSystemBrowserTestApp/src/CustomLoopbackClient.ts
```diff
@@ -140,4 +139,63 @@ export class CustomLoopbackClient implements ILoopbackClient {
...
+            throw "Unable to deserialize query string";
+        }
+        return deserializedQueryString;
+    }
+
+    /**
+     * Parses query string from given string. Returns empty string if no query symbol is found.
+     * @param queryString
+     */
+    static parseQueryString(queryString: string): string {
```
peterzenz, line 171: Regex match of https://domain.com/path?.. or https://domain.com?... etc.  might be safer.

change/@azure-msal-browser-1fe30015-5db8-41ae-988f-d27fa55e15e3.json
tnorling: We've never bothered with this field, AFAIK it only impacts where the dependency version bump log appears in the changelog but does not affect the bump itself.

jo-arroyo (approved):

peterzenz (approved):

samples/msal-node-samples/ElectronSystemBrowserTestApp/src/CustomLoopbackClient.ts
tnorling: No, the const is used internally in msal-common/browser/node. I just don't want to demonstrate consuming things directly from common in our samples (and this isn't something we need to or should export from any of our libs)

PR #6174, opened 2023-06-29
Remove isEmpty helper function

tnorling:
This PR removes the isEmpty helper in favor of direct validation.

StringUtils.isEmpty doesn't provide any value over directly evaluating the truthiness or falsiness of a string and usage of it has the downsides of both being more verbose and that the typechecker doesn't understand that the variable has already been validated as truthy (so you may _still_ need to directly check truthiness).


hectormmg:
Is this a legacy utility function we have from IE11 support?

tnorling:
> Is this a legacy utility function we have from IE11 support?

No, to be honest, this one has always been a bit of a head scratcher to me...think it dates back to v1, perhaps even earlier

lib/msal-common/src/logger/Logger.ts
```diff
@@ -126,14 +125,7 @@ export class Logger {
         const timestamp = new Date().toUTCString();
 
         // Add correlationId to logs if set, correlationId provided on log messages take precedence
-        let logHeader: string;
-        if (!StringUtils.isEmpty(options.correlationId)) {
-            logHeader = `[${timestamp}] : [${options.correlationId}]`;
-        } else if (!StringUtils.isEmpty(this.correlationId)) {
-            logHeader = `[${timestamp}] : [${this.correlationId}]`;
```
peterzenz, line 133: When would this.correlationId not be the same as options.correlationId?

peterzenz (approved):
What is driving this change?  Why are we making it?  Was there a bug/issue we're trying to resolve?  It doesn't look like we're eliminating any external dependencies, and (to me) this doesn't improve readability or clarity in the code.

lib/msal-common/src/logger/Logger.ts
tnorling: Before the controllers / interaction clients are instantiated we use the global logger instance and pass in a correlationId (options), the global instance doesn't have the `this.correlationId` defined

tnorling:
> What is driving this change? Why are we making it? Was there a bug/issue we're trying to resolve? It doesn't look like we're eliminating any external dependencies, and (to me) this doesn't improve readability or clarity in the code.

The util function is doing 3 separate, redundant validations and hiding the validation behind a util function confuses the typechecker, which causes people to write things like `if (string_name && !isEmpty(string_name))` which is even more redundant.

The correct way to do this, if we wanted to keep a named `isEmpty` check would be to write this as a type guard but I personally find this to be unnecessarily verbose when a simple `!string_name` will suffice.

konstantin-msft (approved):
FYI: `StringUtils.isEmpty()` is widely used across the 1p

PR #6186, opened 2023-07-03
Export additional internal classes for the 1p

konstantin-msft:
- Export additional internal classes for the 1p.
- Extend common constants from browser constants.

### DIFF
--- a/lib/msal-browser/src/internals.ts
+++ b/lib/msal-browser/src/internals.ts
@@ -8,11 +8,10 @@
  * Breaking changes to these APIs will be shipped under a minor version, instead of a major version.
  */
 
-// Cache Manager
+// Cache
 export { BrowserCacheManager } from "./cache/BrowserCacheManager";
-
-// Cache entities
 export { CacheRecord } from "./cache/entities/CacheRecord";
+export { ICachePlugin, ISerializableTokenCache, TokenCacheContext, CacheManager } from "@azure/msal-common";
 
 // Clients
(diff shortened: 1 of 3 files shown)


lib/msal-browser/src/internals.ts
```diff
@@ -8,11 +8,10 @@
  * Breaking changes to these APIs will be shipped under a minor version, instead of a major version.
  */
 
-// Cache Manager
+// Cache
 export { BrowserCacheManager } from "./cache/BrowserCacheManager";
-
-// Cache entities
 export { CacheRecord } from "./cache/entities/CacheRecord";
+export { ICachePlugin, ISerializableTokenCache, TokenCacheContext, CacheManager } from "@azure/msal-common";
```
tnorling, line 14: These are interfaces used only by msal-node, why do they need to be exposed to 1P?

tnorling (commented):
There's a lot of things being exported here which are implementation details of common, can you please explain why these are needed in 1P?

lib/msal-browser/src/internals.ts
konstantin-msft: They are used by 1p `BrokerResponseHandler`

tnorling: They shouldn't be - I'd prefer to do the right thing here and remove the unneeded imports from the 1P lib rather than exporting additional things from msal-browser

PR #6217, opened 2023-07-06
Multiple msal instances support for SSO

sameerag:
When apps run multiple msal instances, attempt to SSO is not made for certain instances as there is not pre-fetched RT in the cache. Though this can be circumvented by setting appropriate cache policy or calling `ssoSilent` instead of `acquireTokenSilent`,  certain apps cannot afford the perf hit needed to determine the right API.

This fix will have MSAL JS attempt `sso` when RT is not found in the cache.

### DIFF
--- a/lib/msal-browser/FAQ.md
+++ b/lib/msal-browser/FAQ.md
@@ -23,6 +23,7 @@
 
 1. [How to get single sign-on in my application with MSAL.js?](#how-to-get-single-sign-on-in-my-application-with-msaljs)
 1. [How can my application recognize a user after sign-in? How do I correlate users between applications?](#how-can-my-application-recognize-a-user-after-sign-in-how-do-i-correlate-users-between-applications)
+1. [How can I SSO when I have multiple MSAL instances running?](#how-can-i-sso-when-i-have-multiple-msal-instances-running)
 
 **[Accounts](#Accounts)**
 
@@ -211,6 +212,23 @@ loginPopup().then((response) => {
     const uniqueID = response.account.homeAccountId;
 })
 ```
+
+## How can I SSO when I have multiple MSAL instances running?
+
+For use cases where applications run multiple msal instances in a single window and expect SSO, please enable `attemptSSO: true` in authOptions in the Configuration. 
+
+```javascript
+const config = {
+    auth: {
+        clientId: "your-client-id",
+        authority: "https://yourApp.b2clogin.com/yourApp.onmicrosoft.com/your_policy",
+        attemptSSO: true;
+    }
+}
+const pca = new PublicClientApplication(config);
+```
+
+Please note that this will be default to `true` in v3 and applications that do not want SSO attempted can set the cache policies accordingly.
 
 # Accounts
 
--- a/lib/msal-browser/FAQ.md
+++ b/lib/msal-browser/FAQ.md
@@ -23,7 +23,7 @@
 
 1. [How to get single sign-on in my application with MSAL.js?](#how-to-get-single-sign-on-in-my-application-with-msaljs)
 1. [How can my application recognize a user after sign-in? How do I correlate users between applications?](#how-can-my-application-recognize-a-user-after-sign-in-how-do-i-correlate-users-between-applications)
-1. [How can I SSO when I have multiple MSAL instances running?](#how-can-i-sso-when-i-have-multiple-msal-instances-running)
+1. [Is SSO enabled when I have multiple MSAL instances running?](#is-sso-enabled-when-i-have-multiple-msal-instances-running)
 
 **[Accounts](#Accounts)**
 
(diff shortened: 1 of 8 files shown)


lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -248,7 +249,10 @@ export class PublicClientApplication extends ClientApplication implements IPubli
                             requestWithCLP.cacheLookupPolicy === CacheLookupPolicy.AccessTokenAndRefreshToken ||
                             requestWithCLP.cacheLookupPolicy === CacheLookupPolicy.RefreshToken)
                         && (requestWithCLP.cacheLookupPolicy !== CacheLookupPolicy.Skip)
+                        && !attemptSsoRTNotFound
                     ) {
+                        // eslint-disable-next-line no-console
+                        console.log("acquireTokenSilentAsync: retrieving refresh token failed.");
```
```suggestion
```

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -240,6 +240,7 @@ export class PublicClientApplication extends ClientApplication implements IPubli
                 return this.acquireTokenByRefreshToken(silentRequest, requestWithCLP).catch((refreshTokenError: AuthError) => {
                     const isServerError = refreshTokenError instanceof ServerError;
                     const isInteractionRequiredError = refreshTokenError instanceof InteractionRequiredAuthError;
+                    const attemptSsoRTNotFound = (refreshTokenError.errorCode === BrowserConstants.NO_TOKENS_FOUND_ERROR) && this.config.auth.attemptSSO;
```
tnorling: What do we think about just attempting ssoSilent when `No_tokens_found` is thrown and let the cache lookup policy determine whether or not ssoSilent should be skipped? Not a huge fan of introducing another config option...

lib/msal-browser/src/app/PublicClientApplication.ts
sameerag: Yeah, removing this, was trying to test something and forgot to remove it. Thanks.

lib/msal-browser/src/app/PublicClientApplication.ts
sameerag: I am fine with it. I was conflicted too, just trying to make sure we do not break anyone who assumes this. Do you mean something like this:

```javascript

if(noTokensFoundError) {
    acquireTokenByIframe(...);
 }
 ```
 
 or 
 
 ```javascript

if(noTokensFoundError || CachePolicy.Skip) {
    ssoSilent(...);
 }
 ```
 
 Option 1, we can accommodate with the above change by removing the config option altogether. It will attempt SSO for anyone who has default cache look up policy.

Comment (sameerag) [reply lib/msal-browser/src/app/PublicClientApplication.ts]: Updated to not use a `config` param. This behavior will be default (try SSO if RT is missing in the cache) unless specified otherwise in cache policy or `forceRefresh`

File: lib/msal-browser/FAQ.md [hunk 3]
```diff
@@ -212,6 +213,11 @@ loginPopup().then((response) => {
 })
 ```
 
+## Is SSO enabled when I have multiple MSAL instances running?
+
+Yes. We try to SSO by default unless the cache policies say otherwise.
```
tnorling: This does not have enough context and is very open to interpretation imo - I would either expand on this or remove it, our acquireToken docs should cover this topic in more detail.

lib/msal-browser/src/utils/BrowserConstants.ts
```diff
@@ -19,6 +19,10 @@ export const BrowserConstants = {
      * Invalid grant error code
      */
     INVALID_GRANT_ERROR: "invalid_grant",
+    /**
+     * No tokens found error code
+     */
+    NO_TOKENS_FOUND_ERROR: "no_tokens_found",
```
tnorling: This error code is already defined elsewhere (in whichever class defines the error) - can we use that instead?

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -240,6 +240,7 @@ export class PublicClientApplication extends ClientApplication implements IPubli
                 return this.acquireTokenByRefreshToken(silentRequest, requestWithCLP).catch((refreshTokenError: AuthError) => {
                     const isServerError = refreshTokenError instanceof ServerError;
                     const isInteractionRequiredError = refreshTokenError instanceof InteractionRequiredAuthError;
+                    const attemptSsoRTNotFound = (refreshTokenError.errorCode === BrowserConstants.NO_TOKENS_FOUND_ERROR);
```
konstantin-msft: Should we name it just `rtNotFound` instead? Application logic that handles this scenario (refreshTokenError) is defined outside of this function.

lib/msal-browser/src/utils/BrowserConstants.ts
sameerag: It is defined as a `code` in an error class which is not exported. Same with `invalid_grant` above if you check this code. I thought this is the best way to do this.

lib/msal-browser/src/app/PublicClientApplication.ts
sameerag: Makes sense. Done.

lib/msal-browser/FAQ.md
sameerag: Removed. The FAQ initially was added as a guidance for `config` param and now I agree that it does not have enough context to qualify for an FAQ.

lib/msal-browser/src/utils/BrowserConstants.ts
sameerag: Done

tnorling (approved):

konstantin-msft (approved):

PR #6218, opened 2023-07-06
ClientCredentials: Fixed bug where user-supplied cache is loaded into memory only after network request

Robbie-Microsoft:
While working on the PR [Extend proactive token refresh to client_credentials](https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/6102), I discovered a bug while manually testing the fix.

A user-supplied cache (provided as ```cache: { cachePlugin }``` in the configuration) is not loaded into memory until after a network request, inside of handleServerTokenResponse, inside of the ResponseHandler.

We want to load the user-defined cache (if applicable) into memory BEFORE checking the in-memory-cache for existing tokens.

Without this fix, when a token is acquired in the client_credentials flow, ```fromCache``` will always be set to false, and a request will be sent over the network for a token.

Question to others and follow-up: Does this bug exist in other flows?

Edit: looks like this PR will solve the problem described in msal-node's performance doc [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/docs/performance.md). Looking at the code at the bottom of that page, the line directly above the acquireToken call: ```await cca.getTokenCache().getAllAccounts(); // required for cache read```


lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -216,6 +219,25 @@ export abstract class CacheManager implements ICacheManager {
         credential: ValidCredentialType
     ): string;
 
+    /**
+     * Read the user-supplied cache into memory, if applicable
+     * @param serializableCache 
+     * @param persistencePlugin 
+     */
+    async readUserDefinedCacheIntoMemory(
```
bgavrilMS: CacheManager doesn't deal with calling beforeCache / afterCache hooks. 

See https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/533afa12305e9a4bc685e591fc4c10b3afcf1f5d/lib/msal-node/src/cache/TokenCache.ts#L123 for an example of where these cache operations go 

Or https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/533afa12305e9a4bc685e591fc4c10b3afcf1f5d/lib/msal-common/src/response/ResponseHandler.ts#L256 for how the cache hooks are used when a token response comes in.

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -216,6 +219,25 @@ export abstract class CacheManager implements ICacheManager {
...
+        serializableCache: ISerializableTokenCache | null,
+        persistencePlugin: ICachePlugin | null
+    ): Promise<void> {
+        if (serializableCache && persistencePlugin) {
+            const cacheContext = new TokenCacheContext(
+                serializableCache,
+                true,
+            );
+
+            await persistencePlugin.beforeCacheAccess(cacheContext);
```
bgavrilMS: The pattern is: 
```
try { 
     beforeCacheAccess
} 
finally { 
     afterCacheAccess 
}
```



`AfterCacheAccess` needs to be implemented as well, to allow implemeters to make changes.

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -216,6 +219,25 @@ export abstract class CacheManager implements ICacheManager {
...
+     * @param persistencePlugin 
+     */
+    async readUserDefinedCacheIntoMemory(
+        serializableCache: ISerializableTokenCache | null,
+        persistencePlugin: ICachePlugin | null
+    ): Promise<void> {
+        if (serializableCache && persistencePlugin) {
+            const cacheContext = new TokenCacheContext(
+                serializableCache,
+                true,
```
bgavrilMS: Flag should be set to false, since the cache hasn't changed. It only changes when: 

- MSAL gets new tokens from AAD and merges them in
- Some `Remove` API is used.

lib/msal-node/src/client/ClientCredentialClient.ts
```diff
@@ -71,6 +71,9 @@ export class ClientCredentialClient extends BaseClient {
     private async getCachedAuthenticationResult(
         request: CommonClientCredentialRequest
     ): Promise<AuthenticationResult | null> {
+        // read the user-supplied cache into memory, if applicable
+        await this.cacheManager.readUserDefinedCacheIntoMemory(this.config.serializableCache, this.config.persistencePlugin);
```
bgavrilMS: I think it's ok to invoke `beforeCacheAccess` / `afterCacheAccess` here.

Robbie-Microsoft: That's how I had it originally before I moved it to the cache manager. I'll move it back.

pmaytak (approved):

lib/msal-node/src/client/ClientCredentialClient.ts
pmaytak: I think short-term it's ok to put it here as it's consistent with other code. But it seems more appropriate to put this somewhere in more internal cache classes, that's why CacheManager seemed better. IMO ClientCredentialClient and ResponseHandler shouldn't need to know the proper steps to call / cache; these event handlers are internal implementation that should be abstracted away somwhere.

sameerag: CacheManager is intended as a generic abstraction for any cache lookups/storage/removal and the exact implementation is delegated to the platform layer (node in this case). Serialization and snapshots of cache are a part of node and not the common layer. 

I would prefer adding this code (if it is ever moved) to the `msal-node` cache layer and not in common.

sameerag (approved):
Added context for `CacheManager` class in comments, please keep that in mind for future changes.

lib/msal-node/src/client/ClientCredentialClient.ts
```diff
@@ -71,8 +72,19 @@ export class ClientCredentialClient extends BaseClient {
...
+        // read the user-supplied cache into memory, if applicable
+        let cacheContext;
+        if (this.config.serializableCache && this.config.persistencePlugin) {
+            cacheContext = new TokenCacheContext(this.config.serializableCache, false);
+            await this.config.persistencePlugin.beforeCacheAccess(cacheContext);
+        }
+        
         const cachedAccessToken = this.readAccessTokenFromCache();
 
+        if (this.config.serializableCache && this.config.persistencePlugin && cacheContext) {
```
bgavrilMS, line 84: Please use the `try`  / `finally `pattern that you see in other places that call `beforeCacheAccess` and `after`

derisen:
@Robbie-Microsoft @bgavrilMS @sameerag this issue pertains to other flows as well (especially public client auth code flow in relation to msal-node-extensions). It was being tracked here: #5502 

For CCA, there's a non-trivial cost to hydrating the in memory cache with the persistent cache i.e. if the persistence layer is hosted on cloud or similar, each check will involve a network call (couple 100ms worst case -consider latency differences between a redis cache vs a sql db). It might be fine for client credentials but it could be a bottleneck for a resource server using OBO.

Essentially the question is how often the in-memo cache should be hydrated before a token request, given that the in-memo cache is the single source of truth for msal. "Always" is probably the best wrt cache coherence in a horizontal scaling scenario, but worth discussing what other options might be viable for devs optimizing for perf. My 2 cents: perhaps it makes sense to have a cache policy configuration parameter for the cache field in msal-node configuration, with options like "check in-memo cache first, then persistence if needed", "check persistence first, then in-memo cache" and etc..

pmaytak:
> For CCA, there's a non-trivial cost to hydrating the in memory cache with the persistent cache i.e. if the persistence layer is hosted on cloud or similar, each check will involve a network call (couple 100ms worst case -consider latency differences between a redis cache vs a sql db). It might be fine for client credentials but it could be a bottleneck for a resource server using OBO.
> 
> Essentially the question is how often the in-memo cache should be checked before a token request, given that the in-memo cache is the single source of truth for msal. "Always" is probably the best wrt cache coherence in a horizontal scaling scenario, but worth discussing what other options might be viable for devs optimizing for perf. My 2 cents: perhaps it makes sense to have a cache policy configuration parameter for the cache field in msal-node configuration, with options like "check in-memo cache first, then persistence if needed", "check persistence first, then in-memo cache" and etc..

I would expect it to be something like - always check in-memory cache first, if not empty, return data, but if it's empty, check persistence cache. The issue now is if new CCA is created per request, the in-memory cache will always be empty (unless that workaround is used and getAccounts is called first),

derisen:
> I would expect it to be something like - always check in-memory cache first, if not empty, return data, but if it's empty, check persistence cache. The issue now is if new CCA is created per request, the in-memory cache will always be empty (unless that workaround is used and getAccounts is called first),

@pmaytak agreed, that would be the general approach (for PCA and CCA), but as you said, CCA will typically be a fresh instance and so, empty in-memory cache..

PR #6219, opened 2023-07-06
Adding OIDCOptions parameter to config

shylasummers:
Adds OIDCOptions parameter to config, which allows the customer to set the default OIDC scopes and the hash response type (query or fragment) when using OIDC protocol mode.


sameerag (commented):
nit: Please add description for context in the future

lib/msal-browser/src/controllers/StandardController.ts
```diff
@@ -324,14 +327,39 @@ export class StandardController implements IController {
...
+                throw ClientConfigurationError.createCannotSetOIDCOptionsError();
+        }
+
+        let found_hash = hash;
+        
+        if(this.config.auth.protocolMode === ProtocolMode.OIDC && 
+            this.config.auth.OIDCOptions?.serverResponseType === ServerResponseType.QUERY) {
+                /**
+                 * Check from ?code to make sure we don't get a random query string
+                 * until # since some IDPs add stuff that doesn't concern MSAL after the #
```
peterzenz: Wording.  "add stuff that doesn't concern MSAL..."  
Suggest something more like:  extract hash between '?code=' and '#' if trailing '# is present.

lib/msal-browser/src/interaction_client/PopupClient.ts
```diff
@@ -540,16 +585,34 @@ export class PopupClient extends StandardInteractionClient {
...
 
-                let href: string = Constants.EMPTY_STRING;
-                let hash: string = Constants.EMPTY_STRING;
+                const href = popupWindow.location.href;
+                let hash;
+                if(this.config.auth.protocolMode === ProtocolMode.OIDC && 
+                    this.config.auth.OIDCOptions?.serverResponseType === ServerResponseType.QUERY) {
+                    /**
+                     * Check from ?code to make sure we don't get a random query string
+                     * until # since some IDPs add stuff that doesn't concern MSAL after the #
```
peterzenz: Ditto for this one.

peterzenz (changes requested):

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -74,6 +75,10 @@ export type BrowserAuthOptions = {
      * Enum that represents the protocol that msal follows. Used for configuring proper endpoints.
      */
     protocolMode?: ProtocolMode;
+    /**
+     * Enum that configures options for the OIDC protocol mode.
+     */
+    OIDCOptions?: OIDCOptions | null;
```
tnorling: This probably doesn't need `null` as an option since `undefined` is already acceptable as well

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -240,6 +245,7 @@ export function buildConfiguration(
         navigateToLoginRequestUrl: true,
         clientCapabilities: [],
         protocolMode: ProtocolMode.AAD,
+        OIDCOptions: null,
```
tnorling: Please give this a default value of type `OIDCOptions`

lib/msal-browser/src/controllers/StandardController.ts
```diff
@@ -324,14 +327,39 @@ export class StandardController implements IController {
             this.initialized
         );
 
+        // Throw an error if user has set OIDCOptions without being in OIDC protocol mode
+        if(this.config.auth.protocolMode !== ProtocolMode.OIDC && 
+            this.config.auth.OIDCOptions) {
+                throw ClientConfigurationError.createCannotSetOIDCOptionsError();
```
tnorling: When you remove the `null` type this check could move into the `buildConfiguration` function

lib/msal-browser/src/controllers/StandardController.ts
```diff
@@ -324,14 +327,39 @@ export class StandardController implements IController {
...
+        if(this.config.auth.protocolMode === ProtocolMode.OIDC && 
+            this.config.auth.OIDCOptions?.serverResponseType === ServerResponseType.QUERY) {
+                /**
+                 * Check from ?code to make sure we don't get a random query string
+                 * until # since some IDPs add stuff that doesn't concern MSAL after the # 
+                 */
+                const url = window.location.href;
+                if(url.indexOf("?code") > -1) {
+                    if(url.indexOf("#") > -1) {
+                        found_hash = url.substring(url.indexOf("?code") + 1, url.indexOf("#"));
```
tnorling: We have utility functions in `UrlString` (msal-common) that can handle breaking the down url into its separate parts. Please take advantage of those APIs.

tnorling: nit: can we call this something other than `hash` given it's coming from the query string? `serverResponse` or `serverResponseString` perhaps?

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
```diff
@@ -437,12 +439,23 @@ export abstract class StandardInteractionClient extends BaseInteractionClient {
             PerformanceEvents.InitializeBaseRequest,
             this.correlationId
         );
+
+        let configResponseMode = null;
+        if(!this.config.auth.OIDCOptions || 
+            (this.config.auth.OIDCOptions.serverResponseType &&
+            this.config.auth.OIDCOptions.serverResponseType === ServerResponseType.HASH)) {
+            configResponseMode = ResponseMode.FRAGMENT;
```
tnorling: Setting defaults should happen in the configuration builder

lib/msal-browser/src/interaction_client/PopupClient.ts
```diff
@@ -413,6 +415,49 @@ export class PopupClient extends StandardInteractionClient {
...
+                    this.browserStorage.removeIdToken(validRequest.account?.homeAccountId);
+                    this.browserStorage.removeRefreshToken(validRequest.account?.homeAccountId);
+                    
+                    this.eventHandler.emitEvent(
+                        EventType.LOGOUT_SUCCESS,
+                        InteractionType.Popup,
+                        validRequest
+                    );
+
+                    const popupWindow = this.openPopup(window.location.href, {
```
tnorling: We don't need to open the popup at all if there's no logout endpoint, no?

lib/msal-browser/src/interaction_client/PopupClient.ts
```diff
@@ -413,6 +415,49 @@ export class PopupClient extends StandardInteractionClient {
             );
             this.logger.verbose("Auth code client created");
 
+            try {
+                authClient.authority.endSessionEndpoint;
+            } catch {
+                if (validRequest.account?.homeAccountId && validRequest.postLogoutRedirectUri && authClient.authority.protocolMode === ProtocolMode.OIDC){
+                    this.browserStorage.removeAccount(validRequest.account?.homeAccountId);
```
tnorling, line 422: The `removeAccount` function will also remove associated tokens so the next 3 lines shouldn't be needed

lib/msal-browser/src/interaction_client/RedirectClient.ts
```diff
@@ -507,6 +507,35 @@ export class RedirectClient extends StandardInteractionClient {
...
+                        this.eventHandler.emitEvent(
+                            EventType.LOGOUT_SUCCESS,
+                            InteractionType.Redirect,
+                            validLogoutRequest
+                        );
+        
+                        if (!this.browserStorage.getInteractionInProgress()) {
+                            this.browserStorage.setInteractionInProgress(true);
+                        }
+                        await this.navigationClient.navigateExternal(
```
tnorling: We don't need to navigate anywhere if there's no logout endpoint. Let's simply log a warning

lib/msal-common/src/authority/OIDCOptions.ts
```diff
@@ -0,0 +1,18 @@
...
+ * Options for the OIDC protocol mode.
+ */
+export type OIDCOptions = {
+    serverResponseType?: ServerResponseType;
+    defaultScopes?: Array<string>;
+};
+
+export const ServerResponseType = {
+    QUERY: "query",
+    HASH: "hash",
```
tnorling: Can we reuse the `ResponseMode` enum here instead of creating a new one?

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -200,8 +203,13 @@ export class AuthorizationCodeClient extends BaseClient {
         // Deserialize hash fragment response parameters.
         const hashUrlString = new UrlString(hashFragment);
         // Deserialize hash fragment response parameters.
-        const serverParams: ServerAuthorizationCodeResponse =
-            UrlString.getDeserializedHash(hashUrlString.getHash());
+        let serverParams : ServerAuthorizationCodeResponse;
+        if(this.config.authOptions.authority.options.OIDCOptions?.serverResponseType === ServerResponseType.QUERY) {
+            serverParams = UrlString.getDeserializedHash(hashFragment);
```
tnorling: `getDeserializedQueryString`?

lib/msal-browser/src/config/Configuration.ts
shylasummers: The reason I have it like this is because, as I understand, we don't want to set OIDCOptions to anything when using AAD protocol mode (in fact, we should be throwing an error if this happens). This is also why I have the null type, because it throws an error when I try to set it as just undefined.

lib/msal-browser/src/controllers/StandardController.ts
shylasummers: I don't think we have one that does exactly this; am I wrong about that, or should I make a utility function that will do this?

lib/msal-common/src/authority/OIDCOptions.ts
shylasummers: I made a new one because ResponseMode also has form_post, which we don't want for OIDCOptions.

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
shylasummers: Can you clarify what you mean by this? The response mode is configured in the OIDCOptions parameter (which is in the configuration builder).

lib/msal-browser/src/interaction_client/RedirectClient.ts
shylasummers: I have this because I want the window to refresh or redirect so the UI doesn't still say the user is logged in. If there's a better way to do that, I'm happy to implement it.

lib/msal-browser/src/controllers/StandardController.ts
hectormmg: @shylasummers try to leverage whatever you find there, if you can use functions in there to achieve parts of this then you can re-use that code, even if you do have to add a new function to wrap the steps and add the extra logic.

lib/msal-common/src/client/AuthorizationCodeClient.ts
hectormmg: I'd vote for `parseQueryString` instead. If the method doesn't already exist, which I'm fairly certain it does.

lib/msal-common/src/authority/OIDCOptions.ts
hectormmg: You can create a subtype called OidcResponseMode to re-use the ResponseMode type like so:

```typescript
export const OidcResponseMode = Exclude<ResponseMode, ResponseMode.FORM_POST>
```

Removes as much repetition as possible. For readability I'd put that type in Constants right after ResponseMode and import it wherever it's used.

lib/msal-browser/src/config/Configuration.ts
tnorling: You can do both - throw an error here if AAD protocol mode is provided with OIDCOptions. And if it's not provided (regardless) give it a default value that matches our default behavior today (ResponseMode Fragment and existing default scopes)

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -316,5 +323,12 @@ export function buildConfiguration(
         system: { ...DEFAULT_BROWSER_SYSTEM_OPTIONS, ...providedSystemOptions },
         telemetry: { ...DEFAULT_TELEMETRY_OPTIONS, ...userInputTelemetry },
     };
+
+    // Throw an error if user has set OIDCOptions without being in OIDC protocol mode
+    if(overlayedConfig.auth.protocolMode !== ProtocolMode.OIDC &&
```
tnorling: If you move this up a little bit and use the user provided values you can build the behavior I mentioned above

lib/msal-browser/src/interaction_client/RedirectClient.ts
tnorling: Refreshing the UI is the responsibility of the app and doesn't necessarily require a total refresh of the page. I'd rather leave it up to developers to decide how they want to handle that.

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
tnorling: If you ensure that `this.config.auth.OIDCOptions` is always defined (either with a default or with the user provided object) then you can reduce all of this (442-451) into just `const configResponseMode = this.config.auth.OIDCOptions.serverResponseType`

lib/msal-browser/src/config/Configuration.ts
shylasummers: I get it now, I can change that

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -310,11 +321,28 @@ export function buildConfiguration(
         },
     };
 
+    // Throw an error if user has set OIDCOptions without being in OIDC protocol mode
+    if(userInputAuth &&
+        (!userInputAuth.protocolMode || userInputAuth.protocolMode !== ProtocolMode.OIDC) && 
+        userInputAuth.OIDCOptions) {
```
tnorling: nit for readability
```suggestion
    if(userInputAuth?.protocolMode !== ProtocolMode.OIDC && 
        userInputAuth.OIDCOptions) {
```

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -310,11 +321,28 @@ export function buildConfiguration(
...
+            throw ClientConfigurationError.createCannotSetOIDCOptionsError();
+    }
+
+    // Throw an error if user has set allowNativeBroker to true without being in AAD protocol mode
+    if(userInputAuth &&
+        userInputAuth.protocolMode &&
+        userInputAuth.protocolMode !== ProtocolMode.AAD && 
+        providedSystemOptions &&
+        providedSystemOptions.allowNativeBroker) {
```
tnorling:
```suggestion
    if(userInputAuth?.protocolMode &&
        userInputAuth.protocolMode !== ProtocolMode.AAD &&
        providedSystemOptions?.allowNativeBroker) {
```

lib/msal-browser/src/controllers/StandardController.ts
```diff
@@ -72,6 +74,7 @@ import { BaseOperatingContext } from "../operatingcontext/BaseOperatingContext";
 import { version, name } from "../packageMetadata";
 import { IController } from "./IController";
 import { AuthenticationResult } from "../response/AuthenticationResult";
+import { UrlString } from "..";
```
tnorling: Please add this to the msal-common import above instead

lib/msal-browser/src/interaction_client/PopupClient.ts
```diff
@@ -540,16 +578,23 @@ export class PopupClient extends StandardInteractionClient {
                     return;
                 }
 
-                let href: string = Constants.EMPTY_STRING;
-                let hash: string = Constants.EMPTY_STRING;
+                const href = popupWindow.location.href;
+                let hash;
```
tnorling:
```suggestion
                let serverResponseString;
```

lib/msal-browser/src/interaction_client/RedirectClient.ts
```diff
@@ -507,6 +507,28 @@ export class RedirectClient extends StandardInteractionClient {
...
+                        
+                        this.eventHandler.emitEvent(
+                            EventType.LOGOUT_SUCCESS,
+                            InteractionType.Redirect,
+                            validLogoutRequest
+                        );
+        
+                        if (!this.browserStorage.getInteractionInProgress()) {
+                            this.browserStorage.setInteractionInProgress(true);
+                        }
```
tnorling: Not needed since we're not redirecting
```suggestion
```

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
```diff
@@ -437,12 +439,21 @@ export abstract class StandardInteractionClient extends BaseInteractionClient {
...
             this.correlationId
         );
+
+        let configResponseMode = null;
+        if(this.config.auth.OIDCOptions.serverResponseType === ServerResponseType.FRAGMENT) {
+            configResponseMode = ResponseMode.FRAGMENT;
+        }
+        else {
+            configResponseMode = ResponseMode.QUERY;
+        }
```
tnorling:
```suggestion
        const configResponseMode: ResponseMode = this.config.auth.OIDCOptions.serverResponseType;
```

tnorling (commented):
Last few nits, looking good!

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
shylasummers: I'm running into a small problem with this one. Since the user can choose to set only one part of OIDCOptions (e.g. just the default scopes), OIDCOptions.serverResponseType is allowed to be either undefined or of type ServerResponseType. undefined does not map to ResponseMode. So even though, at this point, OIDCOptions.serverResponseType is defined because of the changes you suggested to Configuration.ts, Typescript throws an error because it has the potential to be undefined.

tnorling: Ah, this is because the `BrowserConfiguration` type makes every member of `auth` required but it's not recursive. You can either mess around with the types in the Configuration file to make sure nested objects are also wrapped in a `Required` utility type or, an easier way to address this, might be to make `OIDCOptions` its own category (sibling of auth, cache, system, etc)

shylasummers: I don't want to do the second option, because I think it makes more sense for the user to set OIDCOptions in a different part of the config than ProtocolMode. I'm having trouble finding how to do the first option, both when just playing around with the code and when looking it up online, and I'm wondering if it's more effort than is necessary to save 5 lines of code.

tnorling: You can define a type like this:

```js
type InternalAuthOptions = Required<BrowserAuthOptions> & {
    OIDCOptions: Required<OIDCOptions>;
};
```

And replace existing references to `Required<BrowserAuthOptions>` with the new `InternalAuthOptions`.

Let me know if that doesn't work or if you want to work through this together

lib/msal-browser/src/interaction_client/PopupClient.ts
```diff
@@ -540,16 +578,23 @@ export class PopupClient extends StandardInteractionClient {
                     return;
                 }
 
-                let href: string = Constants.EMPTY_STRING;
-                let hash: string = Constants.EMPTY_STRING;
+                const href = popupWindow.location.href;
+                let serverResponseString;
```
hectormmg: A bit of a nit: These ~7 lines are all about setting the serverResponseString value. Maybe we can factor them out into a new method to clean this up a bit.

lib/msal-browser/src/interaction_client/StandardInteractionClient.ts
```diff
@@ -437,12 +439,15 @@ export abstract class StandardInteractionClient extends BaseInteractionClient {
...
+
+        const configResponseMode: ResponseMode = this.config.auth.OIDCOptions.serverResponseType;
+
         const validatedRequest: AuthorizationUrlRequest = {
             ...(await this.initializeBaseRequest(request)),
             redirectUri: redirectUri,
             state: state,
             nonce: request.nonce || this.browserCrypto.createNewGuid(),
-            responseMode: ResponseMode.FRAGMENT,
+            responseMode: configResponseMode,
```
hectormmg: Since it's only used here, I think I'd rather we don't reassign the value to a new variable and just access the config object here instead. Mainly because `this.config.auth.OIDCOptions.serverResponseType` gives me a much better idea of what the value is supposed to represent than `configResponseMode`.

lib/msal-browser/test/app/PublicClientApplication.spec.ts
```diff
@@ -825,6 +829,38 @@ describe("PublicClientApplication.ts Class Unit Tests", () => {
             expect(tokenResponse1).toEqual(tokenResponse2);
             expect(tokenResponse4).toEqual(tokenResponse1);
         });
+        it("Looks for hash in query param if OIDCOptions.serverResponseType is set to query", async () => {
```
hectormmg:
```suggestion
        it("Looks for server code response in query param if OIDCOptions.serverResponseType is set to query", async () => {
```

lib/msal-browser/test/app/PublicClientApplication.spec.ts
```diff
@@ -825,6 +829,38 @@ describe("PublicClientApplication.ts Class Unit Tests", () => {
...
+            /**
+             * The testing environment does not accept query params, so instead we see that it ignores a hash fragment.
+             * That means handleRedirectPromise is looking for a hash in a query param instead of a hash fragment.
+             */
+            pca = new PublicClientApplication({
```
hectormmg, line 834: I would suggest adding a `describe` block that contains this test to isolate and better identify tests for PCA under OIDC protocol mode. Then you can build the PCA in a `beforeEach` block and we can add more tests here with the same config later.

lib/msal-browser/test/interaction_client/PopupClient.spec.ts
```diff
@@ -1584,6 +1586,58 @@ describe("PopupClient", () => {
             });
         });
 
+        it("returns hash in query form when serverResponseType in OIDCOptions is query", (done) => {
```
hectormmg:
```suggestion
        it("returns server code response in query form when serverResponseType in OIDCOptions is query", (done) => {
```

In this context, `hash` used to refer to the server auth code response because we only supported "hash" mode, so the thing we were parsing was always a "hash". Now that we support `query` in OIDC as well, calling it a "hash" is no longer accurate.

lib/msal-browser/test/interaction_client/PopupClient.spec.ts
```diff
@@ -1584,6 +1586,58 @@ describe("PopupClient", () => {
...
+                pca.performanceClient,
+                //@ts-ignore
+                pca.nativeInternalStorage,
+                undefined,
+                TEST_CONFIG.CORRELATION_ID
+            );
+
+            const popup = {
+                location: {
+                    href: "http://localhost/?code=hello",
```
hectormmg: Nit: this value can be moved to a constant (i.e. TEST_QUERY_CODE_RESPONSE)` or something like that.

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -200,8 +203,13 @@ export class AuthorizationCodeClient extends BaseClient {
         // Deserialize hash fragment response parameters.
         const hashUrlString = new UrlString(hashFragment);
         // Deserialize hash fragment response parameters.
-        const serverParams: ServerAuthorizationCodeResponse =
-            UrlString.getDeserializedHash(hashUrlString.getHash());
+        let serverParams : ServerAuthorizationCodeResponse;
```
hectormmg: This can also be factored out into a method called getDeserializedCodeResponse or something like that.

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -336,7 +344,7 @@ export class AuthorizationCodeClient extends BaseClient {
         }
 
         // Add scope array, parameter builder will add default scopes and dedupe
-        parameterBuilder.addScopes(request.scopes);
+        parameterBuilder.addScopes(request.scopes, undefined, this.config.authOptions.authority.options.OIDCOptions?.defaultScopes);
```
hectormmg: As opposed to my previous comment, here the config value is used in multiple places and there's validation done on whether OIDCOptions is set. This means it's more practical to set `const oidcDefaultScopes = this.config.authOptions.authority.options.OIDCOptions?.defaultScopes` somewhere above this and just reference the value here and in line 507.

hectormmg: This way it's more readable and the logical check for OIDCOptions is only executed once.

lib/msal-common/src/url/UrlString.ts
```diff
@@ -237,6 +237,28 @@ export class UrlString {
...
     }
 
+    /**
+     * Parses query server response string from given string. 
+     * Extract hash between '?code=' and '#' if trailing '# is present.
+     * Returns empty string if no query symbol is found.
+     * @param queryString
+     */
+    static parseQueryServerResponse(queryString: string): string {
+        const queryIndex1 = queryString.indexOf("?code");
```
hectormmg, line 247: Let's try to refactor this method to be more readable here. I think it can be rewrittent o be more self-documenting and even then it might benefit from comments so anyone who reads it knows why and what it's doing.

hectormmg: On further thought: The manual processing of the string may be more readable in general, but we should consider using RegEx. If we do decide to go with RegEx here, we just need to make sure the expression you come up with works for all browsers, since we've had cases where we missed an error with a regular expression in safari.

lib/msal-common/src/utils/Constants.ts
```diff
@@ -213,12 +213,19 @@ export const CodeChallengeMethodValues = {
...
     FRAGMENT: "fragment",
-    FORM_POST: "form_post",
+} as const;
+export type ServerResponseType = typeof ServerResponseType[keyof typeof ServerResponseType];
+
+/**
+ * allowed values for response_mode
+ */
+export const ResponseMode = {
+    ...ServerResponseType, FORM_POST: "form_post"
```
hectormmg, line 228: Let's make sure we have the business logic to back "form_post" up as an option. If we only have successful code paths for `query` and `hash` then we should reconsider allowing the config option.

hectormmg (commented):
Some comments/suggestions/considerations, but it's looking great so far.

lib/msal-common/src/utils/Constants.ts
shylasummers: Could you clarify what you mean by this?

lib/msal-common/src/url/UrlString.ts
shylasummers: Could you explain why I might choose RegEx over how it's currently written if it's more readable right now?

lib/msal-common/src/utils/Constants.ts
hectormmg: We don't have any code that actually processes a response in "form_post" format do we? If that's the case we probably don't need to include it as an option here.

lib/msal-common/src/url/UrlString.ts
hectormmg: Here's a few thoughts, but I'll preface it by saying: I don't think these differences are super important, your current implementation works well (you can always add a bunch of edge cases to unit tests to test this function) and it would be perfectly fine to leave as is. 

1. I don't think the string size or pattern we're searching for are large enough for performance differences between `indexOf` and RegEx to be a concern here
2. My first thought is "we're trying to match and extract a substring between '?code' or '/?code' and '#'and disregard anything else, which is more of a pattern match (what RegEx does) than a simple check to see if a substring is present in the string (what `indexOf` technically does)
3. A RegEx is a bit tougher to read at a glance (comments that explain what it should do mitigate this), but in this case we have about 10 lines of logic with multiple conditions and index increment operations that can be hard to read and easy to hide a bug in. We'd be trading that for 1 or 2 lines with a RegEx, and there's plenty of tools to debug a RegEx.

Those are just my $0.02, now, assuming you decide against using a RegEx, I also think you can consider a 3rd alternative: leverage the `parseQueryString` method above and check if there's a "code" attribute:


```typescript
const queryString = window.location.url // or wherever we get it from
const serverResponse = UrlString.parseQueryString(queryString);
const authCode = serverResponse.code;

if (authCode) {
// keep going
} else {
// there was no code in the query string, so not an auth code response
}
```

lib/msal-common/src/utils/Constants.ts
shylasummers: Oh, this is the old ResponseMode type. One of the earlier comments said to avoid repetition in the types, and since I had trouble removing form_post from ResponseMode to make ServerResponseType, I did it the other way around. form_post is there to make it the same as ResponseMode was before.

lib/msal-common/src/url/UrlString.ts
shylasummers: parseQueryString returns a string instead of a ServerAuthorizationCodeResponse, so I don't know if we can get a code attribute from that. 
I think I'll keep the current logic, but let me know if there's a way to make it clearer (maybe moving something around or adding some more comments).

lib/msal-common/src/utils/Constants.ts
hectormmg: Ah, that's great then.

lib/msal-browser/src/index.ts
```diff
@@ -20,6 +20,7 @@ export { PublicClientApplication } from "./app/PublicClientApplication";
 export {
     Configuration,
     BrowserAuthOptions,
+    InternalAuthOptions,
```
tnorling: This does not need to be exported
```suggestion
```

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -204,7 +215,7 @@ export type Configuration = {
 };
 
 export type BrowserConfiguration = {
-    auth: Required<BrowserAuthOptions>;
+    auth: Required<InternalAuthOptions>;
```
tnorling: Your type definition above already makes everything required
```suggestion
    auth: InternalAuthOptions;
```

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -229,7 +240,7 @@ export function buildConfiguration(
     isBrowserEnvironment: boolean
 ): BrowserConfiguration {
     // Default auth options for browser
-    const DEFAULT_AUTH_OPTIONS: Required<BrowserAuthOptions> = {
+    const DEFAULT_AUTH_OPTIONS: Required<InternalAuthOptions> = {
```
tnorling:
```suggestion
    const DEFAULT_AUTH_OPTIONS: InternalAuthOptions = {
```

hectormmg (approved):
LGTM, thanks for addressing feedback!

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -310,11 +325,25 @@ export function buildConfiguration(
...
+    // Throw an error if user has set allowNativeBroker to true without being in AAD protocol mode
+    if(userInputAuth?.protocolMode &&
+        userInputAuth.protocolMode !== ProtocolMode.AAD &&
+        providedSystemOptions?.allowNativeBroker) {
+            throw ClientConfigurationError.createCannotAllowNativeBrokerError();
+    }
+
     const overlayedConfig: BrowserConfiguration = {
-        auth: { ...DEFAULT_AUTH_OPTIONS, ...userInputAuth },
+        auth: { ...DEFAULT_AUTH_OPTIONS, ...userInputAuth as InternalAuthOptions },
```
tnorling: We should not be typecasting the user input, if this is needed to pass typechecking something is not quite right above
```suggestion
        auth: { ...DEFAULT_AUTH_OPTIONS, ...userInputAuth },
```

lib/msal-browser/src/interaction_client/PopupClient.ts
```diff
@@ -540,16 +578,16 @@ export class PopupClient extends StandardInteractionClient {
                     return;
                 }
 
-                let href: string = Constants.EMPTY_STRING;
-                let hash: string = Constants.EMPTY_STRING;
+                const href = popupWindow.location.href;
```
tnorling: The popupWindow.location.href call needs to stay inside the try because this will not succeed when the popup is not same origin

lib/msal-browser/src/interaction_client/PopupClient.ts
```diff
@@ -540,16 +578,16 @@ export class PopupClient extends StandardInteractionClient {
...
+                const serverResponseString = this.extractServerResponseStringFromPopup(popupWindow, href);
```
tnorling: This also needs to be inside the try block

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -336,7 +336,7 @@ export class AuthorizationCodeClient extends BaseClient {
         }
 
         // Add scope array, parameter builder will add default scopes and dedupe
-        parameterBuilder.addScopes(request.scopes);
+        parameterBuilder.addScopes(request.scopes, undefined, this.oidcDefaultScopes);
```
```suggestion
        parameterBuilder.addScopes(request.scopes, true, this.oidcDefaultScopes);
```

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -496,7 +496,7 @@ export class AuthorizationCodeClient extends BaseClient {
             ...(request.scopes || []),
             ...(request.extraScopesToConsent || []),
         ];
-        parameterBuilder.addScopes(requestScopes);
+        parameterBuilder.addScopes(requestScopes, undefined, this.oidcDefaultScopes);
```
```suggestion
        parameterBuilder.addScopes(requestScopes, true, this.oidcDefaultScopes);
```

lib/msal-browser/src/config/Configuration.ts
shylasummers: I had it like this, because otherwise, Typescript would throw an error that the types are not assignable to each other. I've been trying to change types above to fix the error, but I haven't been able to successfully do it.

tnorling: The problem in this case is that you start with the default (which has everything defined) then you're overwriting the defaults with whatever the user provided. If, let's say, the user doesn't provide default scopes your final object will have default scopes undefined because the whole default object gets overwritten with the object the user provided.

You could address this by doing something like this:

```js
const { OIDCOptions as userInputOIDCOptions, ...restUserInput } = userInputAuth;

...
auth: {
 ...DEFAULT_AUTH_OPTIONS, 
 ...restUserInput, 
 OIDCOptions: { ...DEFAULT_AUTH_OPTIONS.OIDCOptions, ...userInputOIDCOptions }
}
```

You don't see this pattern in the existing logic because we haven't really done nested objects like this before, instead typically opting to create a new category if config options make sense to be grouped together like this. That said I'm ok with either, I'll leave it up to you what you prefer.

peterzenz (approved):

lib/msal-browser/src/config/Configuration.ts
shylasummers: The stuff about the nested objects makes sense. I'd prefer to have OIDCOptions in auth (to me, it makes the most sense to set ProtocolMode and OIDCOptions in the same object), so I implemented something like this. Thank you!

lib/msal-browser/src/interaction_client/PopupClient.ts
```diff
@@ -833,4 +871,22 @@ export class PopupClient extends StandardInteractionClient {
...
+    /** 
+     * Extracts the server response from the popup window
+     */
+    extractServerResponseStringFromPopup(
+        popupWindow: Window,
+        href: string
+    ): string {
+        let serverResponseString;
+        if(this.config.auth.protocolMode === ProtocolMode.OIDC && 
+            this.config.auth.OIDCOptions?.serverResponseType === ServerResponseType.QUERY) {
```
tnorling: Now that the validation is done in the config builder and you've given this a default value we can just directly check responseType

```suggestion
        if(this.config.auth.OIDCOptions.serverResponseType === ServerResponseType.QUERY) {
```

lib/msal-browser/src/controllers/StandardController.ts
```diff
@@ -324,14 +327,22 @@ export class StandardController implements IController {
             this.initialized
         );
 
+        let foundServerResponse = hash;
+        
+        if(this.config.auth.protocolMode === ProtocolMode.OIDC && 
+            this.config.auth.OIDCOptions?.serverResponseType === ServerResponseType.QUERY) {
```
tnorling: Same here
```suggestion
        if(this.config.auth.OIDCOptions.serverResponseType === ServerResponseType.QUERY) {
```

tnorling (approved):
Looks great! Thanks for addressing all my feedback :)

PR #6226, opened 2023-07-10
Catch errors thrown by "decodeURIComponent"

konstantin-msft:
- Catch errors thrown by "decodeURIComponent"
- Negate state condition check for better readability.

### DIFF
--- a/lib/msal-common/src/response/ResponseHandler.ts
+++ b/lib/msal-common/src/response/ResponseHandler.ts
@@ -82,15 +82,33 @@ export class ResponseHandler {
         cryptoObj: ICrypto
     ): void {
         if (!serverResponseHash.state || !cachedState) {
-            throw !serverResponseHash.state
-                ? ClientAuthError.createStateNotFoundError("Server State")
-                : ClientAuthError.createStateNotFoundError("Cached State");
+            throw serverResponseHash.state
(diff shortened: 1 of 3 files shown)


tnorling (approved):
Any tests you can add for the new error cases?

konstantin-msft:
> Any tests you can add for the new error cases?

done

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -82,15 +82,33 @@ export class ResponseHandler {
         cryptoObj: ICrypto
     ): void {
         if (!serverResponseHash.state || !cachedState) {
-            throw !serverResponseHash.state
-                ? ClientAuthError.createStateNotFoundError("Server State")
-                : ClientAuthError.createStateNotFoundError("Cached State");
+            throw serverResponseHash.state
```
lalimasharda, line 85: Is this change for better readability?

lalimasharda (approved):

lib/msal-common/src/response/ResponseHandler.ts
konstantin-msft: Yep

peterzenz (approved):

PR #6231, opened 2023-07-11
Prioritize hardcoded metadata over network-sourced metadata

hectormmg:
- Fixes a bug where hardcoded metadata for Cloud Instance Discovery was incorrectly built and therefore never worked
- Promotes hardcoded Cloud Instance Discovery metadata over cached and network metadata sources
- Fixes a bug where azure region update method mutates the existing metadata object instead of returning a new object causing MSAL to prepend azure region cumulatively
- Promotes hardcoded Endpoint metadata over cached and network metadata sources


konstantin-msft (approved):

lib/msal-common/src/authority/Authority.ts
```diff
@@ -653,23 +737,41 @@ export class Authority {
...
+            metadataEntity.updateCloudDiscoveryMetadata(
+                hardcodedMetadata,
+                false
+            );
+            return AuthorityMetadataSource.HARDCODED_VALUES;
+        }
+
         this.logger.verbose(
-            "Did not find cloud discovery metadata in the config... Attempting to get cloud discovery metadata from the cache."
+            "Did not find cloud discovery metadata in hardcoded values... Attempting to get cloud discovery metadata from the cache."
```
peterzenz: Not entirely accurate, we may have hardcoded metadata, but the user has configured the option to skip it.  The logs won't represent that state as this is written.

peterzenz (approved):

tnorling (approved):

lib/msal-common/src/authority/Authority.ts
hectormmg: You're absolutely right, updated with an extra check and log message to be accurate. Thanks for catching this!

PR #6233, opened 2023-07-12
Throw when initialize has not been called

hectormmg:
This PR:

- Replaces `BrowserUtils.blockNativeBrokerCalledBeforeInitialize` with `BrowserUtils.blockCallsBeforeInitialize` since now `initialize` call is required regardless of whether `allowNativeBroker` is enabled
- Calls `BrowserUtils.blockCallsBeforeInitialize` in all public APIs related to token acquisition and accounts
- Replaces `BrowserAuthError.createNativeBrokerCalledBeforeInitialize()` with `BrowserAuthError.createUninitializedPublicClientApplication()` and updates error message
- Updates browser spec files to match new behavior


lib/msal-browser/src/controllers/StandardController.ts
```diff
@@ -1437,6 +1438,8 @@ export class StandardController implements IController {
     public getNativeAccountId(
         request: RedirectRequest | PopupRequest | SsoSilentRequest
     ): string {
+        // Block token acquisition before initialize has been called
+        BrowserUtils.blockCallsBeforeInitialize(this.initialized);
```
tnorling: `getNativeAccountId` shouldn't be public so this isn't needed here

tnorling (approved):
Looks good besides one minor comment, thanks for this!

lib/msal-browser/src/controllers/StandardController.ts
hectormmg: Are we planning on changing this method signature? Or are we waiting for v4 to not break the surface API?

tnorling: We should change it in v3, it's not public surface - I'm fairly certain this got changed during the move to the StandardController. If it's needed in 1P let's make it protected

sameerag (commented):
Can we not initialize ourselves by default instead of adding this check at a zillion places?

sameerag (changes requested):
Can we sync up once before I approve this on why we chose not to initialize ourselves or if there is a better way to do this?

hectormmg:
> Can we sync up once before I approve this on why we chose not to initialize ourselves or if there is a better way to do this?

There's not much we can do since initialize is async. We need these checks because someone could instantiate a PCA and start calling APIs without initializing.

sameerag (approved):
As discussed, approving with suggestions to improve on this for the next major version. Also:
nit: Can we rename the function name to `blockAPICallsBeforeInitialize()`?

shoatman (approved):
lgtm

tnorling (approved):

PR #6237, opened 2023-07-14
Angular standalone sample

jo-arroyo:
This PR adds a sample for using MSAL Angular v3 with Angular standalone components and updates documentation.


schmkr:
question: with v3 of msal-angular, shouldn't the public client be explicitly initialised (with `await client.initialize();`). I don't see that in the example added it, but without the explicit initialization we see in the console these errors:

```
BrowserAuthError: native_broker_called_before_initialize: You must call and await the initialize function before attempting to call any other MSAL API when native brokering is enabled
```

jo-arroyo:
> question: with v3 of msal-angular, shouldn't the public client be explicitly initialised (with `await client.initialize();`). 

@schmkr great question. `initialize()` is called by `handleRedirectObservable`, so as long as the `handleRedirectObservable` is correctly subscribed to (as in this sample) or bootstrapped with the `MsalRedirectComponent` (other msal-angular samples), the public client should be initialized and the error you quoted should not appear. Please see the source code for `handleRedirectObservable` [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/8c382b3b677bf9388154ab6c2863d6951905e91a/lib/msal-angular/src/msal.service.ts#L58).

If you are still seeing that error, please try subscribing to `handleRedirectObservable` as shown in [this sample app](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/cae2279e005b626e17f3f2a29e76e1c10199f821/samples/msal-angular-v3-samples/angular-standalone-sample/src/app/app.component.ts#L34) and [open an issue](https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/new/choose) if you are still seeing that error.

lib/msal-angular/FAQ.md
```diff
@@ -43,21 +44,25 @@ Please see [here](https://github.com/AzureAD/microsoft-authentication-library-fo
 
 ### What versions of Angular are supported?
 
-MSAL Angular v3 is in alpha and currently supports Angular 15 and 16.
+MSAL Angular v3 is in beta and currently supports Angular 15 and 16.
```
peterzenz: Another update point for the upcoming release.  @hectormmg

samples/msal-angular-v3-samples/angular-standalone-sample/src/main.ts
```diff
@@ -0,0 +1,119 @@
...
+import { Route, provideRouter, withDisabledInitialNavigation, withEnabledBlockingInitialNavigation } from '@angular/router';
+import { MsalInterceptor, MSAL_INSTANCE, MsalInterceptorConfiguration, MsalGuardConfiguration, MSAL_GUARD_CONFIG, MSAL_INTERCEPTOR_CONFIG, MsalService, MsalGuard, MsalBroadcastService } from '@azure/msal-angular';
+import { IPublicClientApplication, PublicClientApplication, InteractionType, BrowserCacheLocation, LogLevel, BrowserUtils } from '@azure/msal-browser';
+import { AppComponent } from './app/app.component';
+import { HomeComponent } from './app/home/home.component';
+import { ProfileComponent } from './app/profile/profile.component';
+import { FailedComponent } from './app/failed/failed.component';
+import { environment } from './environments/environment';
+
+const isIE = window.navigator.userAgent.indexOf("MSIE ") > -1 || window.navigator.userAgent.indexOf("Trident/") > -1;
```
peterzenz: Any isIE references should be gone from our code in our MSAL.js v3 release.  We're dropping IE support in v3.

peterzenz (approved):

samples/msal-angular-v3-samples/angular-standalone-sample/src/main.ts
jo-arroyo: Makes sense. I've removed all references to it. Thanks!

lib/msal-angular/FAQ.md
tnorling: Since GA is imminent
```suggestion
MSAL Angular v3 currently supports Angular 15 and 16.
```

tnorling (approved):

PR #6248, opened 2023-07-18
Add storeInCache request parameter to control which tokens are persisted to the cache

tnorling:
Adds new request parameter to control which tokens should be stored in the cache. This will give us more flexibility and enable our 1P library to use our public APIs rather than depend on and/or re-implement internal implementation details.

Note: msal-node is out of scope for this change for now. If there's a need to extend this feature to node in the future the param can be removed from the "Omit" list of the request types and tests added.


hectormmg (approved):
LGTM

lalimasharda (approved):
LGTM!

sameerag (approved):

peterzenz:
How does this work with WAM/Native Brokers?  If the app experience is going to be very different between a pure browser flow and a JS/Native flow because of this, then I'd suggest we back this change out.

tnorling:
> How does this work with WAM/Native Brokers? If the app experience is going to be very different between a pure browser flow and a JS/Native flow because of this, then I'd suggest we back this change out.

This is referring to the local MSAL cache (browser storage) and applies regardless of where the token came from. I'm not the biggest fan of the parameter name myself but if you have suggestions for making this clearer or an alternative solution let me know.

PR #6256, opened 2023-07-21
Append v2 to endpoint when using a Microsoft authority under OIDC protocol mode

shylasummers:
Currently, MSAL appends v2 to the default open id configuration endpoint only when using AAD protocol mode or when authority type is ADFS. This PR changes that check to append v2 to the endpoint when using a known Microsoft authority or when authority type is ADFS.


lib/msal-browser/src/config/Configuration.ts
```diff
@@ -340,6 +341,14 @@ export function buildConfiguration(
             throw ClientConfigurationError.createCannotAllowNativeBrokerError();
     }
 
+    // Throw an error if using a known Microsoft authority with OIDC compliance mode
+    if(userInputAuth?.authority && userInputAuth?.protocolMode === ProtocolMode.OIDC) {
+        const knownMSAuthorities = Object.keys(rawMetdataJSON.endpointMetadata);
```
sameerag: nit: `rawMetadataJSON`?

shylasummers: I noticed that too, the original variable is actually called that and I wasn't sure why, I can change it if there isn't a reason for it.

shylasummers: For some reason this change caused many failing tests. Because the typo appears at many points in the code, I assume it's purposeful.

hectormmg: You can import the named `EndpointMetadata` variable from `AuthorityMetadata` instead. It's called `rawMetadataJson` because it is a "raw" copy-paste from the network response.

These are the last couple lines in AuthorityMetadata:
```typescript
export const EndpointMetadata = rawMetdataJSON.endpointMetadata;
export const InstanceDiscoveryMetadata =
    rawMetdataJSON.instanceDiscoveryMetadata;
```

lib/msal-common/src/index.ts
```diff
@@ -47,6 +47,7 @@ export {
 } from "./authority/AuthorityOptions";
 export { AuthorityFactory } from "./authority/AuthorityFactory";
 export { AuthorityType } from "./authority/AuthorityType";
+export { rawMetdataJSON } from "./authority/AuthorityMetadata";
```
hectormmg:
```suggestion
export { EndpointMetadata } from "./authority/AuthorityMetadata";
```

hectormmg (changes requested):

lib/msal-common/src/index.ts
```diff
@@ -47,6 +47,7 @@ export {
 } from "./authority/AuthorityOptions";
 export { AuthorityFactory } from "./authority/AuthorityFactory";
 export { AuthorityType } from "./authority/AuthorityType";
+export { rawMetdataJSON, EndpointMetadata } from "./authority/AuthorityMetadata";
```
hectormmg: Is `rawMetadataJSON` still used now that you've exported `EndpointMetadata`? If not then we can remove it from exports.

hectormmg (approved):
LGTM

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -340,6 +341,14 @@ export function buildConfiguration(
             throw ClientConfigurationError.createCannotAllowNativeBrokerError();
     }
 
+    // Throw an error if using a known Microsoft authority with OIDC compliance mode
+    if(userInputAuth?.authority && userInputAuth?.protocolMode === ProtocolMode.OIDC) {
```
lalimasharda: This check only happens when MSAL is being instantiated with the user input config right? Just out of curiosity, is this error only thrown for config issues or do we also want to add it in the case where authority might be provided in the login request as well?

shylasummers: Yep, only thrown in the config right now. My thought process is that, since the error would mean MSAL never gets initialized, you can't authenticate with the known Microsoft authority in OIDC compliance mode. Is there a case where somebody can bypass this check in the login request? I wouldn't know about it if there is.

tnorling: Yup, you can pass a different authority on a per request basis - I'm curious though do we really need this check? It's not clear to me what we're trying to protect against

lalimasharda: I don't think anyone will be able to bypass this check as it is. But I do think that it is possible to pass in a different authority in the login request (once MSAL is initialized properly) and in that case authority provided in login request is given [precedence](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-common/docs/authority.md#:~:text=(2)%20takes%20precedence%20over%20(1)%20i.e.%20you%20can%20override%20the%20initial%20authority%20URL%20set%20in%20the%20configuration%20object%20later%20on%20during%20each%20token%20request.%20See%20msal%2Dbrowser%20configuration%20and%20msal%2Dnode%20configuration%20for%20more.) over the authority provided in the config. Although, I am not entirely sure if this is a possible use case for new IDPs. 
To reframe my question, do you foresee a scenario where MSAL might be initialized with a non-Microsoft authority, but someone is trying to login with a Microsoft authority or vice versa?

*PS: I just updated the wording of my last question*

shylasummers: @tnorling The check is because the AAD protocol mode is optimized for Microsoft authority, so there will probably be less issues if people can't use OIDC protocol mode with Microsoft authorities. Also because we don't want people to change the default scopes sent to Microsoft authorities, without this check they theoretically would.

@lalimasharda I don't foresee anybody doing this in practice to get around this check (because there wouldn't really be any additional benefit for them, if anything it would cause them more problems) but it might be possible in theory, I'm unsure on that.

hectormmg: I do have a slight hesitation of throwing on `buildConfiguration`. Maybe it's better to throw during the request validation, since the final authority will have been set by then, per what @lalimasharda is pointing out. Just the fact that `buildConfiguration` doesn't even have access to the logger makes me feel no validation should happen at this point, outside of setting defaults.

tnorling: AAD may be optimized for Microsoft authorities (and that is certainly our recommendation) but I'm not sure I understand why we would want to block the "less optimal" path if that's what someone wants to do. This feels kind of arbitrary to me. 

We're also not capable of blocking OIDC mode for B2C authorities in this same way (we don't have a known list of all the possible authorities) so this will cause an inconsistent experience between eSTS and B2C which I think is undesirable.

tnorling: r.e. Switching to a different authority on a per-request basis. This is absolutely a valid, supported scenario which we need to make sure is covered.

hectormmg: The B2C/DSTS/ADFS parity argument does make me doubt, but there's still a good reason to throw, which is we don't want supported response modes and default scopes (and whatever else we end up making configurable in OIDC Mode) overwritten for AAD mode.

From a COGS point of view, this could cause a lot of preventable failed requests to reach ESTS with incorrect default scopes. It could also let devs downgrade to use query response mode when we've previously made the decision to only use hash response mode for security.

@peterzenz I remember discussing this with you, was there another reason to block other than the COGS argument?

tnorling: The service supports any combination of scopes, not including one of these shouldn't result in failed requests. We mandate them because we think it will give our customers the best experience. Important to remember this is "OIDC Mode" and Microsoft authorities are OIDC compliant, so any of these options should be compatible. In fact, for use cases calling MSA directly you _must_ use OIDC protocol mode (this was the original use case for adding the ProtocolMode config)

tnorling: By all means, we should set sensible defaults to optimize for things like perf, security and reliability but I'm not sure I agree with actively blocking a less optimal path if these options are supported by the spec

hectormmg: @tnorling  Ok, with that in mind I'd agree there's then no reason to throw with AAD authorities. I was under the impression that default scopes were necessary. Would we need to change MSAL's expectations of the response though?

tnorling: I assume you're asking if we properly handle the absence of a refresh token in the response, for example? Ensuring we handle this gracefully should be in-scope for this work and it was my understanding this was both manually tested and unit tests were added prior to merging the previous PRs. Is that accurate?

peterzenz: There were asks for things like folks using OIDC mode and a Microsoft authority so they could disable RTs, something we require today.  There's also a long-term consistency approach with defaults and JS/Native broker flows.  If we end up in a place where we get radically different behavior when an extension is either present or not, I expect we'll get a lot of customer confusion there.  So if native bridge is enabled, OIDC mode is selected, and a Microsoft Authority is passed in, is the result consistent with just enabling native bridge and using a Microsoft authority?  It *might* be.  From a support perspective, that *maybe* answer complicates things for us and makes things harder for our customers.  If we're using OIDC mode and changing the authority on a per-request basis, I'd suggest we throw there as well as the initial config so we're consistent.

tnorling: As part of Shyla's previous PR, OIDC mode and native brokering can't be used together. If OIDC mode is on, we will turn off native brokering and if the user had explicitly enabled it we will log a warning saying these configs are incompatible with each other. I think that addresses your concerns, no?

peterzenz: That works, thanks.

hectormmg: @tnorling right, I meant scenarios like no offline_access (my assumption is MSAL is already set up to throw interaction required since it's the same behavior as a missing or expired RT), AAD authority with query response mode (which I believe may now be covered by Shyla's new code), no OIDC scopes, etc. Shyla and I are discussing how to test for and cover these scenarios, should be fine to not block AAD authorities in OIDC mode as long as features such as native brokering are disabled.

konstantin-msft (approved):

hectormmg (approved):

PR #6264, opened 2023-07-26
Fix formatting and run formatting check in CI

tnorling:
Formatting changes cause code churn and make PRs harder to review. This adds a CI check to ensure formatting is correct on each PR and applies formatting to all the files which were missed in the original move to Prettier.

Future enhancement could auto commit a formatting fix, if needed


peterzenz (approved):
Appreciate you not mixing functional changes with the broad formatting ones.

konstantin-msft (approved):

PR #6270, opened 2023-07-27
Update "PerformanceClient" for better usability/extendibility

konstantin-msft:
- Make measurement name param a string.
- Make measurement payload type a string.
- Merge telemetry "Counters" and "StaticFields" into "PerformanceEvent". 
- Simplify "InProgressPerformanceEvent" method names.
- Make "int fields" a parameter of "PerformanceClient" constructor.
- Make "performance client" a parameter of "PublicClientApplication" constructor.
- Stub "PublicClientApplication.performanceClient" by default to eliminate telemetry instrumentation overhead.

### DIFF
--- a/lib/msal-browser/src/config/Configuration.ts
+++ b/lib/msal-browser/src/config/Configuration.ts
@@ -12,12 +12,15 @@ import {
     ProtocolMode,
     OIDCOptions,
     ServerResponseType,
+    Logger,
     LogLevel,
     StubbedNetworkModule,
     AzureCloudInstance,
     AzureCloudOptions,
     ApplicationTelemetry,
-    ClientConfigurationError
+    ClientConfigurationError,
+    IPerformanceClient,
+    StubPerformanceClient
 } from "@azure/msal-common";
 import { BrowserUtils } from "../utils/BrowserUtils";
 import {
@@ -26,6 +29,7 @@ import {
 } from "../utils/BrowserConstants";
 import { INavigationClient } from "../navigation/INavigationClient";
 import { NavigationClient } from "../navigation/NavigationClient";
+import { name, version } from "../packageMetadata";
 
 // Default timeout for popup windows and iframes in milliseconds
 export const DEFAULT_POPUP_TIMEOUT_MS = 60000;
@@ -190,6 +194,8 @@ export type BrowserTelemetryOptions = {
      * - appVersion: Version of the application using MSAL
      */
     application?: ApplicationTelemetry;
+
+    client?: IPerformanceClient;
 };
 
 /**
@@ -251,8 +257,8 @@ export function buildConfiguration(
         navigateToLoginRequestUrl: true,
         clientCapabilities: [],
         protocolMode: ProtocolMode.AAD,
-        OIDCOptions: { 
-            serverResponseType: ServerResponseType.FRAGMENT, 
+        OIDCOptions: {
+            serverResponseType: ServerResponseType.FRAGMENT,
             defaultScopes: [Constants.OPENID_SCOPE, Constants.PROFILE_SCOPE, Constants.OFFLINE_ACCESS_SCOPE]
         },
         azureCloudOptions: {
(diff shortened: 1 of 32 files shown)


lib/msal-browser/src/config/Configuration.ts
```diff
@@ -190,6 +194,8 @@ export type BrowserTelemetryOptions = {
      * - appVersion: Version of the application using MSAL
      */
     application?: ApplicationTelemetry;
+
+    client?: IPerformanceClient;
```
tnorling, line 198: can we call this something like "telemetryClient" instead? We use "client" for a fairly wide range of things

tnorling: Oh nevermind - didn't see this was a child of the "telemetry" object. This is fine.

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
```diff
@@ -440,18 +439,20 @@ export abstract class PerformanceClient implements IPerformanceClient {
 
     /**
      * Increment counters to be emitted when the measurements are flushed
-     * @param counters {Counters}
+     * @param fields {string[]}
      * @param correlationId {string} correlation identifier
      */
-    increment(counters: Counters, correlationId: string): void {
+    incrementFields(fields: { [key: string]: {} | undefined; }, correlationId: string): void {
```
tnorling: Should this be number? 

```suggestion
    incrementFields(fields: { [key: string]: number | undefined; }, correlationId: string): void {
```

lib/msal-node/src/client/UsernamePasswordClient.ts
```diff
@@ -53,7 +53,7 @@ export class UsernamePasswordClient extends BaseClient {
         );
 
         const httpVerToken = response.headers?.[HeaderNames.X_MS_HTTP_VERSION];
-        atsMeasurement?.addStaticFields({
+        atsMeasurement?.add({
```
tnorling: Why does msal-node have a measurement at all? This can probably be removed?

tnorling (approved):
Love these changes! Thanks!

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -323,10 +329,21 @@ export function buildConfiguration(
             appName: Constants.EMPTY_STRING,
             appVersion: Constants.EMPTY_STRING,
         },
+        client: new StubPerformanceClient(
+            DEFAULT_AUTH_OPTIONS.clientId,
+            DEFAULT_AUTH_OPTIONS.authority,
+            new Logger(DEFAULT_LOGGER_OPTIONS, name, version),
+            name,
```
tnorling, line 336: Could the PerformanceClient get the name and version itself instead of passing it in the constructor?

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
konstantin-msft: It can be a string, array or anything else.

tnorling: How do you increment a string? Or am I misunderstanding what this does?

lib/msal-browser/src/config/Configuration.ts
konstantin-msft: Unfortunately, we can't do that because 1p and 3p have their own `packageMetadata.ts` files.

lib/msal-node/src/client/UsernamePasswordClient.ts
konstantin-msft: Good point, thank you. Removed it.

lib/msal-browser/src/config/Configuration.ts
tnorling: Could we wrap the client in 1P and set the correct metadata values? Trying to see how we can simplify the constructor, given this will be public

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
konstantin-msft: Sorry I thought you were referring to `add` method. Fixed it, thanks!

lib/msal-browser/src/config/Configuration.ts
sameerag: Will this also help with setting up the correct SKU if we generalize this?

lib/msal-common/src/telemetry/performance/IPerformanceClient.ts
```diff
@@ -4,34 +4,33 @@
...
+    end: (
         event?: Partial<PerformanceEvent>
     ) => PerformanceEvent | null;
-    discardMeasurement: () => void;
-    addStaticFields: (staticFields: StaticFields) => void;
-    increment: (counters: Counters) => void;
+    discard: () => void;
+    add: (fields: { [key: string]: {} | undefined; }) => void;
+    increment: (fields: { [key: string]: number | undefined; }) => void;
     event: PerformanceEvent;
```
sameerag, line 21: Why undefined? Can we not determine the type in advance? Or is it to keep it generic?

konstantin-msft: To keep it compatible with the existing data points that can be undefined. Let me check if i can fix that without making breaking changes.

sameerag (approved):
Had a few comments. Great improvement.

lib/msal-common/src/telemetry/performance/IPerformanceClient.ts
konstantin-msft: We have to keep `undefined` for now due to the handful of data points that can be null/undefined.

lib/msal-browser/src/config/Configuration.ts
konstantin-msft: I have simplified `BrowserPerformanceClient` constructor as per your suggestion.

PR #6271, opened 2023-07-27
Hydrate cache API

tnorling:
In the 1P library we require a mechanism to hydrate the browser cache from an `AuthenticationResult` object created out of band (parent frame). We have an existing `loadExternalTokens` API, however, this API makes some assumptions that MSAL was not used to acquire the tokens and is intended for use inside test suites so does not seem appropriate to repurpose.

This PR adds a public API to `PublicClientApplication` that takes in an `AuthenticationResult` object and a request object and generates the corresponding cache entities for the account, id and access tokens and stores them in the configured cache location.


lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -682,7 +660,7 @@ export class NativeInteractionClient extends BaseInteractionClient {
             );
 
         const nativeCacheRecord = new CacheRecord(
-            accountEntity,
+            undefined,
```
konstantin-msft, line 662: May i ask you why accountEntity is undefined?

lib/msal-common/src/cache/entities/AccountEntity.ts
```diff
@@ -137,18 +118,25 @@ export class AccountEntity {
...
+        clientInfo?: string,
         cloudGraphHostName?: string,
         msGraphHost?: string,
         environment?: string,
         nativeAccountId?: string
     ): AccountEntity {
         const account: AccountEntity = new AccountEntity();
 
-        account.authorityType = CacheAccountType.MSSTS_ACCOUNT_TYPE;
+        if (authority.authorityType === AuthorityType.Adfs) {
```
konstantin-msft, line 126: Should we use a map here instead of `if-else` statement?

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -375,19 +378,22 @@ export class NativeInteractionClient extends BaseInteractionClient {
...
         // generate identifiers
         const idTokenObj = this.createIdTokenObj(response);
         const homeAccountIdentifier = this.createHomeAccountIdentifier(
             response,
             idTokenObj
         );
-        const accountEntity = this.createAccountEntity(
-            response,
+        const accountEntity = AccountEntity.createAccount(
             homeAccountIdentifier,
```
konstantin-msft: It's out of scope of this PR, but something we can address later on. It would be great to be able to pass an object instead of plain params to make sure we can omit undefined ones and not accidentally mix string parameters.

lib/msal-browser/src/app/IPublicClientApplication.ts
```diff
@@ -57,6 +57,7 @@ export interface IPublicClientApplication {
     initializeWrapperLibrary(sku: WrapperSKU, version: string): void;
     setNavigationClient(navigationClient: INavigationClient): void;
     getConfiguration(): BrowserConfiguration;
+    hydrateCache(result: AuthenticationResult, request: SilentRequest) : Promise<void>;
```
sameerag: hydrateCredentialCache?

konstantin-msft (approved):

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
tnorling: In the native flow we cache tokens in-memory (the native broker contains its own persistent cache). The account object, however, needs to persist in the browser. Storing the account in both in-memory and browser storage creates two sources of truth so instead, we store the account in browser storage above and then this call stores only the id and access tokens in in-memory storage

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
tnorling: That's a great suggestion, let me see if I can squeeze that in here

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
konstantin-msft: Got it, Thanks for the detailed explanation, appreciate it!

lib/msal-common/src/cache/entities/AccountEntity.ts
tnorling: Since we're comparing against both authorityType and protocolMode a map seems trickier to implement here :/

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
tnorling: Made this change! Thanks for the suggestion!

lib/msal-browser/src/app/IPublicClientApplication.ts
tnorling: I'd prefer to keep it short but if there's inclination towards a different verb such as "saveTokens", "loadCache", etc. I'm open to suggestions/feedback

sameerag: This hydrates account and tokens, so I am good with the name. I do not think we ever needed `app meta data` and that is the only extra base type in cache for us.

sameerag (approved):

PR #6284, opened 2023-08-02
Expose response handler and update attributes' access modifiers

konstantin-msft:
- Make ResponseHandler attributes protected for extendability.
- Add optional "clientId" param to generateCacheRecord.
- Expose "CacheRecord" and "ResponseHandler" as part of msal-browser "internals".

### DIFF
--- a/lib/msal-browser/src/internals.ts
+++ b/lib/msal-browser/src/internals.ts
@@ -8,8 +8,9 @@
  * Breaking changes to these APIs will be shipped under a minor version, instead of a major version.
  */
 
-// Cache Manager
+// Cache
 export { BrowserCacheManager } from "./cache/BrowserCacheManager";
+export { CacheRecord } from "@azure/msal-common";
 
 // Clients
 export { StandardInteractionClient } from "./interaction_client/StandardInteractionClient";
@@ -24,6 +25,7 @@ export { NativeInteractionClient } from "./interaction_client/NativeInteractionC
 export { RedirectHandler } from "./interaction_handler/RedirectHandler";
 export { EventHandler } from "./event/EventHandler";
 export { NativeMessageHandler } from "./broker/nativeBroker/NativeMessageHandler";
+export { ResponseHandler } from "@azure/msal-common";
 
 // Utilities
 export { BrowserStateObject } from "./utils/BrowserProtocolUtils";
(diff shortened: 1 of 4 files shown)


lib/msal-browser/src/internals.ts
```diff
@@ -24,6 +25,7 @@ export { NativeInteractionClient } from "./interaction_client/NativeInteractionC
 export { RedirectHandler } from "./interaction_handler/RedirectHandler";
 export { EventHandler } from "./event/EventHandler";
 export { NativeMessageHandler } from "./broker/nativeBroker/NativeMessageHandler";
+export { ResponseHandler } from "@azure/msal-common";
```
sameerag, line 28: Is this to avoid `common` exports in the future?

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -42,14 +42,14 @@ import { PerformanceEvents } from "../telemetry/performance/PerformanceEvent";
...
-    private clientId: string;
-    private cacheStorage: CacheManager;
-    private cryptoObj: ICrypto;
-    private logger: Logger;
-    private homeAccountIdentifier: string;
-    private serializableCache: ISerializableTokenCache | null;
-    private persistencePlugin: ICachePlugin | null;
-    private performanceClient?: IPerformanceClient;
+    private readonly clientId: string;
+    protected readonly cacheStorage: CacheManager;
```
sameerag, line 46: I think we can avoid this with the cache changes made by Thomas.

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -42,14 +42,14 @@ import { PerformanceEvents } from "../telemetry/performance/PerformanceEvent";
...
+    protected readonly cryptoObj: ICrypto;
```
sameerag, line 47: Same here, if logger and crypto are available. Let us check on this.

lib/msal-browser/src/internals.ts
tnorling: I don't really want to add new stuff to internals, let's instead focus on what we need to do to remove internals all-up

lib/msal-common/src/response/ResponseHandler.ts
tnorling: Yeah none of this should be needed. I understand this may be a stopgap but all of this will be superseded by the 1P changes and we'll then have to revert all of this again. Would prefer to just hold out until the 1P changes are done

PR #6286, opened 2023-08-03
Pick up broker extra query params

konstantin-msft:
- Add "child_client_id" and "child_redirect_uri" query params.
- Manage "client_id" based on extra query params for web flow.
- Manage child query params based on extra query params for native flow.

### DIFF
--- a/lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
+++ b/lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
@@ -27,6 +27,7 @@ import {
     CommonSilentFlowRequest,
     AccountInfo,
     CacheRecord,
+    AADServerParamKeys,
 } from "@azure/msal-common";
 import { BaseInteractionClient } from "./BaseInteractionClient";
 import { BrowserConfiguration } from "../config/Configuration";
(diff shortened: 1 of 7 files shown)


lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -55,6 +56,11 @@ import { BrowserAuthError } from "../error/BrowserAuthError";
 import { SilentCacheClient } from "./SilentCacheClient";
 import { AuthenticationResult } from "../response/AuthenticationResult";
 
+const BrokerServerParamKeys = {
```
sameerag, line 59: To retain the boundary? Usually constants are defined in `Constants.ts`

konstantin-msft: These doesn't look like a constant that needs to be exported all across the module to me.

lib/msal-common/src/utils/Constants.ts
```diff
@@ -164,6 +164,8 @@ export const AADServerParamKeys = {
     RETURN_SPA_CODE: "return_spa_code",
     NATIVE_BROKER: "nativebroker",
     LOGOUT_HINT: "logout_hint",
+    CHILD_CLIENT_ID: "child_client_id",
+    CHILD_REDIRECT_URI: "child_redirect_uri",
```
tnorling: Doesn't look like these are used anywhere - did I miss something?

tnorling (approved):

lib/msal-common/src/utils/Constants.ts
konstantin-msft: That's a good catch, thank you!

jo-arroyo (approved):

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
sameerag: Fair.

sameerag (approved):

PR #6287, opened 2023-08-03
Remove beta tags from publish commands

hectormmg:
Remove beta tags from publish commands in publish_template.yml to prepare for stabel v3 release

### DIFF
--- a/.pipelines/custom-templates/publish-template.yml
+++ b/.pipelines/custom-templates/publish-template.yml
@@ -81,7 +81,7 @@ jobs:
             inputs:
                 command: "custom"
                 workingDir: "${{ parameters.path }}/${{ parameters.libName }}"
-                customCommand: "publish --tag beta" # Remove --tag beta when ready to GA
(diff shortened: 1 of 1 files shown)


konstantin-msft (approved):

.pipelines/custom-templates/publish-template.yml
```diff
@@ -98,7 +98,7 @@ jobs:
             inputs:
                 command: "custom"
                 workingDir: "${{ parameters.path }}/${{ parameters.libName }}"
-                customCommand: "run deploy:beta" # Remove :beta when ready to GA"
```
tnorling, line 101: Should we remove this script from the package.json as well?

tnorling (approved):

.pipelines/custom-templates/publish-template.yml
hectormmg: We can, I assume we'll use it for future beta releases though?

PR #6311, opened 2023-08-09
Replaces `multiple matched tokens error` with a log message, removing matching tokens from cache

konstantin-msft:
Replaces `multiple matched tokens error` with a log message, removing matching tokens from cache.

### DIFF
--- a/lib/msal-common/src/cache/CacheManager.ts
+++ b/lib/msal-common/src/cache/CacheManager.ts
@@ -882,7 +882,13 @@ export abstract class CacheManager implements ICacheManager {
             this.commonLogger.info("CacheManager:getIdToken - No token found");
             return null;
         } else if (numIdTokens > 1) {
-            throw ClientAuthError.createMultipleMatchingTokensInCacheError();
+            this.commonLogger.info(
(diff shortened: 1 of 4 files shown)


lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -1033,7 +1039,13 @@ export abstract class CacheManager implements ICacheManager {
             );
             return null;
         } else if (numAccessTokens > 1) {
-            throw ClientAuthError.createMultipleMatchingTokensInCacheError();
```
tnorling, line 1036: Should we also remove this function? Is it used anywhere else?

tnorling (approved):

lib/msal-common/src/cache/CacheManager.ts
konstantin-msft: It's used in `msal-node` by `ClientCredentialClient` and `OnBehalfOfClient`.

tnorling: Got it thanks. @Robbie-Microsoft It's probably worth implementing a similar fix in those flows but we'll let ya'll make that call

jo-arroyo (approved):

PR #6318, opened 2023-08-09
Fix msal-node export file extensions

hectormmg:


lib/msal-node/package.json
```diff
@@ -21,24 +21,25 @@
...
     "./package.json": "./package.json"
   },
   "files": [
     "dist"
   ],
   "scripts": {
-    "build": "rollup -c --strictDeprecations --bundleConfigAsCjs",
+    "build": "npm run clean && rollup -c --strictDeprecations --bundleConfigAsCjs",
     "build:watch": "rollup -c --watch --strictDeprecations --bundleConfigAsCjs",
+    "clean": "shx rm -rf dist lib",
```
tnorling, line 42: Looks like shx needs to be added to devDependencies

tnorling (approved):

lib/msal-node/package.json
hectormmg: You're absolutely right

konstantin-msft (approved):

PR #6342, opened 2023-08-14
Fix authority endpoint version for B2C authorities

### LINKED ISSUE #6340: The scope provided in the request is not supported
### Core Library

MSAL.js (@azure/msal-browser)

### Core Library Version

3.0.1

### Wrapper Library

### Wrapper Library Version

3.0.1

Public

### Description

After updating to Angular 15 and MSAL 3.0.1, we are getting an error when trying to authenticate.

Since nothing in this regard is documented in the Upgrade Guide, I would not expect any change to be required in this regard or any deprecations.

### Error Message

```
invalid_request: AADB2C90012: The scope '<our-scopes> openid profile offline_access' provided in request is not supported.
```

### Msal Logs

_No response_

### MSAL Configuration

```javascript
new PublicClientApplication({
    auth: {
      clientId: environment.AzureClientId,
      authority: b2cPolicies.authorities.signUpSignIn.authority,
      redirectUri: "/",
      postLogoutRedirectUri: "/",
      knownAuthorities: [`${environment.AzureB2CName}.b2clogin.com`],
    },
    cache: {
      storeAuthStateInCookie: isIE, // set to true for IE 11
    },
    system: {
      loggerOptions: {
        loggerCallback,
        logLevel: LogLevel.Warning,
        piiLoggingEnabled: false,
      },
    },
  });
```

### Relevant Code Snippets

```javascript
-
```

### Reproduction Steps

1. Start the login flow
2. The page blanks out and this error appears in the console

### Expected Behavior

Nothing should error out, because there are no deprecations in the [Upgrade Guide](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-angular/docs/v2-v3-upgrade-guide.md)

### Identity Provider

Azure B2C Basic Policy

Chrome, Firefox, Edge, Safari, Internet Explorer, Other

### Regression

@azure/msal-browser 2.x

### Source

External (Customer)

hectormmg:
This PR:

- Fixes a bug where B2C authority URLs were not having /v2 appended because they were not in the list of known Microsoft authorities

### DIFF
--- a/lib/msal-common/src/authority/Authority.ts
+++ b/lib/msal-common/src/authority/Authority.ts
@@ -356,7 +356,7 @@ export class Authority {
         const canonicalAuthorityHost = this.hostnameAndPort;
         if (
             this.authorityType === AuthorityType.Adfs ||
-            (this.protocolMode === ProtocolMode.OIDC &&
+            (this.protocolMode !== ProtocolMode.AAD &&
                 !this.isAliasOfKnownMicrosoftAuthority(canonicalAuthorityHost))
         ) {
             return `${this.canonicalAuthority}.well-known/openid-configuration`;
--- a/lib/msal-common/src/authority/Authority.ts
+++ b/lib/msal-common/src/authority/Authority.ts
@@ -356,7 +356,8 @@ export class Authority {
         const canonicalAuthorityHost = this.hostnameAndPort;
         if (
             this.authorityType === AuthorityType.Adfs ||
-            !this.isAliasOfKnownMicrosoftAuthority(canonicalAuthorityHost)
+            (this.protocolMode === ProtocolMode.OIDC &&
+                !this.isAliasOfKnownMicrosoftAuthority(canonicalAuthorityHost))
         ) {
             return `${this.canonicalAuthority}.well-known/openid-configuration`;
         }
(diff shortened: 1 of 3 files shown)


lib/msal-common/src/authority/Authority.ts
```diff
@@ -356,7 +356,8 @@ export class Authority {
         const canonicalAuthorityHost = this.hostnameAndPort;
         if (
             this.authorityType === AuthorityType.Adfs ||
-            !this.isAliasOfKnownMicrosoftAuthority(canonicalAuthorityHost)
+            (this.protocolMode !== ProtocolMode.AAD &&
```
hectormmg, line 359: Note: This is an immediate solution, but it may be a good idea to add a way to either detect B2C authorities or indicate the authority is B2C in some way so B2C can be used in OIDC mode.

konstantin-msft (approved):

tnorling (approved):

tnorling:
Tests?

hectormmg:
@tnorling on it, will update the PR a bit later.

lib/msal-common/src/authority/Authority.ts
sameerag: We are not detecting CIAM so detecting B2C should be okay. The decision not to was made when we wanted to be IDP agnostic which is no longer true in many cases.

sameerag (approved):

PR #6356, opened 2023-08-17
Clear claims cache for apps that switched their cache preferences [LTS]

sameerag:
Some apps fetch tokens with high frequency and changing claims (new claim per request). Prior to #6187, these tokens over flowed the local cache. However, there is still the issue of stale tokens to be cleared when the `claimsBasedCachingEnabled`  feature is set to false, which is not addressed.

This PR adds a new utility function `clearTokensAndKeysWithClaims()` which will be triggered when the application is `initialized` to clear any old tokens cached with claims.

The initial choice is to clear the cache at token acquisition. However apps facing this issue will not be able to fetch tokens since the `tokenKeys` cache entry is already large and freezes the app. Hence the choice of adding this when the app is initialized and the extra requirement of any application on the msal v2 to specifically call `msal.initialize()` as a pre-requisite.

Applications are mandated to call `initialize()` for `msal v3` and we are defaulting to `claimsBasedCachingEnabled: false` in v3. Hoping both these settings will mitigate this issue from day one in the new version.

So here is the guidance:

`msal-browser v2`: App needs to explicitly set `claimsBasedCachingEnabled: false` and call `initialize()` to use msal to mitigate this issue.
`msal-browser v3`: MSAL JS should handle this in the back ground.

P.S: Currently this approach is in test and once approved will push the changes to v3.

### DIFF
--- a/lib/msal-browser/src/app/ClientApplication.ts
+++ b/lib/msal-browser/src/app/ClientApplication.ts
@@ -188,7 +188,9 @@ export abstract class ClientApplication {
 
         if(!this.config.cache.claimsBasedCachingEnabled) {
             this.logger.verbose("Claims-based caching is disabled. Clearing the previous cache with claims");
+            const claimsTokensRemovalMeasurement = this.performanceClient.startMeasurement(PerformanceEvents.ClearTokensAndKeysWithClaims);
             await this.browserStorage.clearTokensAndKeysWithClaims();
+            claimsTokensRemovalMeasurement.endMeasurement(PerformanceEvents.ClearTokensAndKeysWithClaims);
         }
 
         this.initialized = true;
--- a/lib/msal-browser/src/app/ClientApplication.ts
+++ b/lib/msal-browser/src/app/ClientApplication.ts
@@ -185,6 +185,12 @@ export abstract class ClientApplication {
                 this.logger.verbose(e);
             }
         }
+
+        if(!this.config.cache.claimsBasedCachingEnabled) {
+            this.logger.verbose("Claims-based caching is disabled. Clearing the previous cache with claims");
+            await this.browserStorage.clearTokensWithClaimsInCache();
+        }
+
         this.initialized = true;
         this.eventHandler.emitEvent(EventType.INITIALIZE_END);
 
--- a/lib/msal-browser/src/app/ClientApplication.ts
+++ b/lib/msal-browser/src/app/ClientApplication.ts
@@ -188,7 +188,7 @@ export abstract class ClientApplication {
 
         if(!this.config.cache.claimsBasedCachingEnabled) {
             this.logger.verbose("Claims-based caching is disabled. Clearing the previous cache with claims");
-            await this.browserStorage.clearTokensWithClaimsInCache();
+            await this.browserStorage.clearTokensAndKeysWithClaims();
         }
 
         this.initialized = true;
--- a/lib/msal-browser/src/app/ClientApplication.ts
+++ b/lib/msal-browser/src/app/ClientApplication.ts
@@ -190,7 +190,7 @@ export abstract class ClientApplication {
             this.logger.verbose("Claims-based caching is disabled. Clearing the previous cache with claims");
             const claimsTokensRemovalMeasurement = this.performanceClient.startMeasurement(PerformanceEvents.ClearTokensAndKeysWithClaims);
             await this.browserStorage.clearTokensAndKeysWithClaims();
-            claimsTokensRemovalMeasurement.endMeasurement(PerformanceEvents.ClearTokensAndKeysWithClaims);
+            claimsTokensRemovalMeasurement.endMeasurement({success: true});
         }
 
         this.initialized = true;
--- a/lib/msal-browser/src/cache/BrowserCacheManager.ts
+++ b/lib/msal-browser/src/cache/BrowserCacheManager.ts
@@ -954,6 +954,26 @@ export class BrowserCacheManager extends CacheManager {
     }
 
     /**
+     * Clears all access tokes that have claims prior to saving the current one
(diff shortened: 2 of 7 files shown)


peterzenz:
> This code is not required for `msal v3` 

How does this work if I'm migrating from v2 to v3 and I'm not using this specific v2 build to clear out large caches?

sameerag:
> > This code is not required for `msal v3`
> 
> How does this work if I'm migrating from v2 to v3 and I'm not using this specific v2 build to clear out large caches?

Sorry I meant to say `initialize()` is mandated for v3, hence the app inherits this migration code there. If an app is using hot swap as in not logging out, yes, this issue will still remain. Correcting the description.

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -185,6 +185,12 @@ export abstract class ClientApplication {
                 this.logger.verbose(e);
             }
         }
+
+        if(!this.config.cache.claimsBasedCachingEnabled) {
+            this.logger.verbose("Claims-based caching is disabled. Clearing the previous cache with claims");
+            await this.browserStorage.clearTokensAndKeysWithClaims();
```
tnorling, line 192: Can we add some telemetry submeasurements around this?

lib/msal-browser/test/cache/BrowserCacheManager.spec.ts
```diff
@@ -944,6 +944,92 @@ describe("BrowserCacheManager tests", () => {
                         expect(browserLocalStorage.getAccessTokenCredential(testAccessTokenWithAuthScheme.generateCredentialKey())).toBeInstanceOf(AccessTokenEntity);
                     }
                 )
+
+                it.only("clearTokensWithClaimsInCache clears all access tokens with claims in tokenKeys", () => {
```
tnorling:
```suggestion
                it("clearTokensWithClaimsInCache clears all access tokens with claims in tokenKeys", () => {
```

tnorling (approved):

lib/msal-browser/test/cache/BrowserCacheManager.spec.ts
hectormmg: I'll check to see if there's a linter rule for this or something, I check in onlies a lot.

sameerag: Thanks @hectormmg. Yes, after many many test runs, forgot to take this out.

JasonGore:
In an old repro case I had from a few months ago and I'm not seeing any reduction in local key quantities or value sizes with this change. I have a repro condition that I'll go through with Sameera, but I think I'd hold off merging this for now until we understand why it's not clearing stale values for me locally

lib/msal-browser/src/app/ClientApplication.ts
JasonGore: I almost think a message of type `Warning` is merited if any keys are actually cleared. This will help highlight any issues in telemetry, particularly if this is occurring more than we expect. (For our use case I only ever expect to see this once per user and would like to verify that if possible.)

sameerag: Sure. Can change this to warning and will add telemetry as needed.

sameerag:
> In an old repro case I had from a few months ago and I'm not seeing any reduction in local key quantities or value sizes with this change. I have a repro condition that I'll go through with Sameera, but I think I'd hold off merging this for now until we understand why it's not clearing stale values for me locally

Sure, as discussed the keys with what I tested seem to clear, so may be there is some other criteria that I am missing. Will test with the actual repro and lets see where we land.

lib/msal-browser/src/app/ClientApplication.ts
JasonGore: Just to clarify I meant that it would be `Warning` only if tokens and keys were cleared (if easily possible) rather than a `Warning` as it is right now (which would occur on every app load.) Thanks!

lib/msal-browser/src/cache/BrowserCacheManager.ts
```diff
@@ -953,6 +953,26 @@ export class BrowserCacheManager extends CacheManager {
...
+    async clearTokensAndKeysWithClaims(): Promise<void> {
+
+        const tokenKeys = this.getTokenKeys();
+            
+        const removedAccessTokens: Array<Promise<void>> = [];
+        tokenKeys.accessToken.forEach((key: string) => {
+            // if the access token has claims in its key, remove the token key and the token
+            const credential = this.getAccessTokenCredential(key);
+            if(credential?.requestedClaimsHash && key.includes(credential.requestedClaimsHash)) {
+                removedAccessTokens.push(this.removeAccessToken(key));
```
JasonGore, line 971: I do see `Info` of `BrowserCacheManager: remoteTokenKey - accessToken removed from map` but am wondering if we can add a `Warning` here as well. I think `Warning` is warranted as these should only occur once under very specific config changes, and will make it easier for us to track in telemetry

JasonGore:
Looks good! The casing fix seemed to have cleared out all remaining keys for my repro. I'm just hoping we can get the `Warning` fix in so I can make sure this is behaving as we expect (and not more than we expect)

lib/msal-browser/src/cache/BrowserCacheManager.ts
sameerag: Added warning if `tokens` are deleted.

lib/msal-browser/src/app/ClientApplication.ts
sameerag: Addressed both: Added telemetry and a warning.

konstantin-msft (approved):

PR #6357, opened 2023-08-17
Do not sign tokens if `reqCnf` is set

sameerag:
* This PR signs the `POP` tokens only if the `reqCnf` is not passed in as a `request` parameter. This is to enable any clients that choose to sign their tokens. However, please consider this an advanced feature only.
* This PR also addresses the native flow bug where `cnf` is to be sent a string instead of a hash!
* Removes `reqCnfHash` in the `ReqCnfData` since we do not use it. It is only internal, so this should not be a breaking change.


lib/msal-common/src/request/BaseAuthRequest.ts
```diff
@@ -43,4 +44,5 @@ export type BaseAuthRequest = {
     maxAge?: number;
     tokenQueryParameters?: StringDict;
     storeInCache?: StoreInCache;
+    reqCnf?: ReqCnfData;
```
sameerag, line 52: Looks like `kid` is needed for the native broker case, else we could just make this a `string`.

lib/msal-browser/docs/access-token-proof-of-possession.md
```diff
@@ -151,6 +151,10 @@ The Proof-of-Possession authentication scheme relies on an asymmetric cryptograp
 
 In the event of refreshing a bound access token, MSAL will delete the cryptographic keypair that was generated when requesting the expired bound access token, generate a new cryptographic keypair for the new access token, and store the new keypair in the keystore.
 
+## Advanced feature: User generated cryptographic keypair
+
+MSAL does not sign the POP tokens if the application chooses to generate its own random cryptographic keypair and returns the unsigned `access_token` generated by the Service to the application. This is fairly advanced feature and we recommend not to use it unless the developer understands POP and finds a use case on their end to manage the `reqCnf` generation themselves. `reqCnf` can be provided as an optional parameter on the `request` type.
```
sameerag: First cut. Soliciting feedback to make sure:
* Developers are not confused when they read this
* Do not have unnecessary POP failures if someone chooses to set this

lib/msal-browser/docs/access-token-proof-of-possession.md
```diff
@@ -151,6 +151,10 @@ The Proof-of-Possession authentication scheme relies on an asymmetric cryptograp
 
 In the event of refreshing a bound access token, MSAL will delete the cryptographic keypair that was generated when requesting the expired bound access token, generate a new cryptographic keypair for the new access token, and store the new keypair in the keystore.
 
+## Advanced feature: User generated cryptographic keypair
```
tnorling: nit: don't like the term "User" here - what about this instead?
```suggestion
## Advanced feature: Application managed cryptographic keypair
```

lib/msal-browser/docs/access-token-proof-of-possession.md
tnorling: I would also add a clear usage example below and make note of the fact that you should not provide the other PoP params.

```suggestion
> :warning: This is an advanced feature and we do not recommend its use unless you are intimately familiar with the Proof of Possession protocol and have a requirement to manage your own keypairs. For most use cases we recommend implementing PoP as described in the rest of this document.

If you would like to generate your own cryptographic keypair and be responsible for signing the resulting tokens you can provide your own `reqCnf` value on the request. If this value is provided MSAL will ensure the token issuer embeds it in the token but will return the token _unsigned_. It is then the application's responsibility to sign the token before attempting to send it to the intended resource. 
```

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -863,12 +864,12 @@ export class NativeInteractionClient extends BaseInteractionClient {
             };
 
             const popTokenGenerator = new PopTokenGenerator(this.browserCrypto);
-            const reqCnfData = await popTokenGenerator.generateCnf(
-                shrParameters
-            );
+            const reqCnfData = request.reqCnf
+                ? request.reqCnf
+                : await popTokenGenerator.generateCnf(shrParameters);
```
```suggestion
            const reqCnfData = request.reqCnf ||
                await popTokenGenerator.generateCnf(shrParameters);
```

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -400,7 +400,9 @@ export class AuthorizationCodeClient extends BaseClient {
                 PerformanceEvents.PopTokenGenerateCnf,
                 request.correlationId
             );
-            const reqCnfData = await popTokenGenerator.generateCnf(request);
+            const reqCnfData = request.reqCnf
+                ? request.reqCnf
+                : await popTokenGenerator.generateCnf(request);
```
```suggestion
            const reqCnfData = request.reqCnf ||
                await popTokenGenerator.generateCnf(request);
```

lib/msal-common/src/request/BaseAuthRequest.ts
tnorling: Not sure why it would be needed in addition to reqCnf - let's confirm that

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -392,7 +392,9 @@ export class RefreshTokenClient extends BaseClient {
                 PerformanceEvents.PopTokenGenerateCnf,
                 request.correlationId
             );
-            const reqCnfData = await popTokenGenerator.generateCnf(request);
+            const reqCnfData = request.reqCnf
+                ? request.reqCnf
+                : await popTokenGenerator.generateCnf(request);
```
tnorling:
```suggestion
            const reqCnfData = request.reqCnf ||
                await popTokenGenerator.generateCnf(request);
```

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -503,8 +503,11 @@ export class ResponseHandler {
         let familyId: string = Constants.EMPTY_STRING;
 
         if (cacheRecord.accessToken) {
+            // if the request object has `reqCnf` property, the token will be returned unsigned
             if (
-                cacheRecord.accessToken.tokenType === AuthenticationScheme.POP
+                cacheRecord.accessToken.tokenType ===
+                    AuthenticationScheme.POP &&
+                !request.reqCnf
```
tnorling, line 598: Is the request here the original request or the "validatedRequest"? If it's the latter this will always be defined, right?

sameerag: Good q. Let me verify that.

lib/msal-common/src/request/BaseAuthRequest.ts
sameerag: I will follow up on this in a separate PR.

lib/msal-common/src/response/ResponseHandler.ts
sameerag: `request` and `request.code` are the only data checked  prior to this. Since `reqCnf` is never defined unless explicitly set by the user, the check should still hold.

lib/msal-browser/docs/access-token-proof-of-possession.md
sameerag: Updated text

lib/msal-common/src/response/ResponseHandler.ts
sameerag: `request` is ensured to be defined by the time we reach here. However `request.reqCnf` should still be undefined unless set explicitly. Does that clarify?

lib/msal-browser/docs/access-token-proof-of-possession.md
sameerag: Added more detailed text and warning.

lib/msal-browser/docs/access-token-proof-of-possession.md
```diff
@@ -151,6 +151,14 @@ The Proof-of-Possession authentication scheme relies on an asymmetric cryptograp
 
 In the event of refreshing a bound access token, MSAL will delete the cryptographic keypair that was generated when requesting the expired bound access token, generate a new cryptographic keypair for the new access token, and store the new keypair in the keystore.
 
+## Advanced feature: Application managed cryptographic keypair
+
+> :warning: We do not recommend to use this feature unless you are familiar with the [Proof of Possession protocol](https://oauth.net/2/dpop/) and have a use case to generate your own cryptographic keypair. For most cases, we recommend the PoP usage as described in the rest of the document.
```
tnorling:
```suggestion
> :warning: We do not recommend using this feature unless you are familiar with the [Proof of Possession protocol](https://oauth.net/2/dpop/) and have a specific requirement to generate your own cryptographic keypair. For most cases, we recommend the PoP usage as described in the rest of this document.
```

lib/msal-common/src/response/ResponseHandler.ts
tnorling: I'm not sure I believe you :) Can we add a test for the msal-browser APIs to verify? The first thing we do in the top level APIs is create a validated request object which should create a reqCnf if not already provided. The original request object is never passed down into msal-common AFAIK

lib/msal-core/dist/msal.js
```diff
@@ -0,0 +1,7306 @@
+/*! msal v1.4.18 2023-04-18 */
+'use strict';
```
tnorling: What's going on here?

lib/msal-common/src/response/ResponseHandler.ts
sameerag: Ha ha, Sure, lets add tests.

lib/msal-core/dist/msal.js
sameerag: Shoot! I have these ghost files that keep getting added in dev from msal-core. Let me clean it forcefully, this is a local repo issue.

lib/msal-common/src/response/ResponseHandler.ts
sameerag: Added tests. Please check.

lib/msal-browser/test/interaction_client/StandardInteractionClient.spec.ts
```diff
@@ -140,6 +141,56 @@ describe("StandardInteractionClient", () => {
...
+            challenge: TEST_CONFIG.TEST_CHALLENGE,
+            verifier: TEST_CONFIG.TEST_VERIFIER,
+        });
+
+        const authCodeRequest =
+            await testClient.initializeAuthorizationCodeRequest(request);
+        expect(authCodeRequest.reqCnf).toBeUndefined;
+    });
+
+    it("initializeAuthorizationCodeRequest validates the request and adds reqCnf param when user defined", async () => {
```
tnorling, line 172: Can you please add a test that validates the end to end flow? i.e. "Includes provided reqCnf on /token request and Returns token unsigned"

konstantin-msft (approved):

lib/msal-browser/test/interaction_client/StandardInteractionClient.spec.ts
sameerag: Did one better and added a test in the e2e pipeline.

lib/msal-browser/src/index.ts
```diff
@@ -94,6 +94,9 @@ export {
     SignedHttpRequestOptions,
 } from "./crypto/SignedHttpRequest";
 
+export { base64Encode as Base64Encode } from "./encode/Base64Encode";
```
sameerag, line 90: We can probably avoid this if we generate or stringify the JSON later. This will however help with any encoding needs that may arise later in 1P. Let me know what you folks think about it and I can do changes accordingly.
cc @konstantin-msft  and @tnorling

sameerag: I do not mind either way but we cannot remove these later as it will be public surface.

tnorling: Would prefer not to add these types of things to the public surface if we can avoid it. Assume you want to use this in 1P? For what purpose?

samples/msal-browser-samples/VanillaJSTestApp2.0/app/pop/test/browser.spec.ts
```diff
@@ -111,4 +111,46 @@ describe("Browser tests", function () {
...
+        expect(tokenStore.refreshTokens).toHaveLength(1);
+        const cachedAccount = await BrowserCache.getAccountFromCache(
+            tokenStore.idTokens[0]
+        );
+        const defaultCachedToken =
+            await BrowserCache.accessTokenForScopesExists(
+                tokenStore.accessTokens,
+                ["openid", "profile", "user.read"]
+            );
+        expect(cachedAccount).toBeDefined();
```
tnorling, line 150: Can we also verify the cached token is unsigned and contains the cnf claim?

lib/msal-browser/src/index.ts
sameerag: `reqCnfData` generation. We can also pass the `kid` as is and encode it while constructing the server request. I need to update the 1P PR and this one to reflect it. Do we not foresee any encoding/decoding needs in 1P? Then we can go that route.

tnorling: I'd prefer that approach. We can discuss future needs when they come up, I don't forsee any right now

sameerag:
Abandoning this for a later PR.

PR #6359, opened 2023-08-17
Shoatman/naa latest

shoatman:


lib/msal-browser/src/naa/BridgeError.ts
```diff
@@ -0,0 +1,21 @@
...
+    code: string; // auth_flow_last_error such as invalid_grant
+    subError: string; // server_suberror_code such as consent_required
+    description: string;
+    properties: object; // additional telemetry info
+};
+
+export function isBridgeError(error: unknown): error is BridgeError {
+    return (
+        (error as BridgeError).status !== undefined &&
+        (error as BridgeError).code !== undefined
```
codexeon: code will not always be populated, status will always have a value

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
```diff
@@ -0,0 +1,206 @@
...
+                request.correlationId !== undefined
+                    ? request.correlationId
+                    : "",
+            prompt: request.prompt !== undefined ? request.prompt : "",
+            claims: request.claims !== undefined ? request.claims : "",
+            authenticationScheme:
+                request.authenticationScheme !== undefined
+                    ? request.authenticationScheme
+                    : "",
+            clientCapabilities: this.clientCapabilities,
```
codexeon, line 77: clientCapabilities should be merged with claims before sending across the bridge.

lib/msal-browser/src/operatingcontext/TeamsAppOperatingContext.ts
```diff
@@ -33,16 +39,41 @@ export class TeamsAppOperatingContext extends BaseOperatingContext {
...
          */
-        return false;
+        try {
+            if (typeof window !== "undefined") {
+                const bridgeProxy: IBridgeProxy = await BridgeProxy.create();
+                /*
+                 * Because we want to get SSO we need to grab the active account as part of initialization
+                 * this.activeAccount = await bridgeProxy.getActiveAccount();
+                 */
+                this.activeAccount = await bridgeProxy.getActiveAccount();
```
codexeon: try/catch this or remove the call, getActiveAccount may not be supported but we should still allow ssoSilent.

lib/msal-browser/src/naa/BridgeProxy.ts
```diff
@@ -0,0 +1,232 @@
...
+            throw new Error("window.crypto is undefined");
+        }
+
+        try {
+            BridgeProxy.crypto = window.crypto;
+
+            window.nestedAppAuthBridge.addEventListener(
+                "message",
+                (response: string) => {
+                    window.console.log(response);
```
codexeon: remove console.log

lib/msal-browser/src/controllers/ControllerFactory.ts
```diff
@@ -25,31 +25,33 @@ export class ControllerFactory {
...
+            teamsApp.isAvailable() &&
+            teamsApp.getConfig().auth.supportsNestedAppAuth
+        ) {
+            const controller = await import("./NestedAppAuthController");
+            return await controller.NestedAppAuthController.createController(
+                teamsApp
+            );
+        } else if (standard.isAvailable()) {
+            const controller = await import("./StandardController");
+            return await controller.StandardController.createController(
```
codexeon, line 58: hitting a race condition inside this function. StandardController.createController is not properly awaiting initialize call.

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
```diff
@@ -0,0 +1,206 @@
...
+
+    public fromNaaTokenResponse(
+        request: TokenRequest,
+        response: TokenResponse
+    ): AuthenticationResult {
+        const expiresOn = new Date(
+            TimeUtils.nowSeconds() + (response.expires_in || 0) * 1000
+        );
+
+        let authToken: AuthToken | undefined;
```
codexeon: should be called idToken

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
```diff
@@ -0,0 +1,206 @@
...
+            requestId: "",
+            extExpiresOn: expiresOn,
+            state: response.state,
+        };
+
+        return authenticationResult;
+    }
+
+    public fromNaaAccountInfo(fromAccount: NaaAccountInfo): MsalAccountInfo {
+        const account: MsalAccountInfo = {
```
codexeon, line 190: this is not a full MsalAccountInfo. It should primarily be built from idToken. Can you use AccountEntity.createAccount to create the account object?

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
```diff
@@ -0,0 +1,206 @@
...
+                Object.entries(request.extraQueryParameters)
+            );
+        }
+
+        const tokenRequest: TokenRequest = {
+            userObjectId: request.account?.homeAccountId,
+            clientId: this.clientId,
+            authority: request.authority,
+            scope: request.scopes.join(" "),
+            correlationId:
```
codexeon, line 101: for standard controller if correlationId is not passed, one is generated. We should follow the same pattern for nested app auth controller.

lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -0,0 +1,554 @@
...
+        ssoSilentMeasurement?.add({
+            nestedAppAuthRequest: true,
+        });
+
+        try {
+            const naaRequest =
+                this.nestedAppAuthAdapter.toNaaSilentTokenRequest(request);
+            const response = await this.bridgeProxy.getTokenSilent(naaRequest);
+
+            const result: AuthenticationResult =
```
codexeon, line 210: **state** parameter is being lost when calling ssoSilent. This is a difference in behavior between standard controller and nested app auth controller. It at least needs to be returned to the caller.

lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -0,0 +1,554 @@
...
+            );
+            ssoSilentMeasurement?.add({
+                accessTokenSize: result.accessToken.length,
+                idTokenSize: result.idToken.length,
+            });
+            ssoSilentMeasurement?.end({
+                success: true,
+                requestId: result.requestId,
+            });
+            return result;
```
codexeon, line 230: what is the expected behavior of pca.getAllAccounts() after a successful token acquisition? For standard controller account is populated, but for nested app auth account is just returned in response and not available globally.

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
```diff
@@ -0,0 +1,206 @@
...
+            extraParams = new Map<string, string>(
+                Object.entries(request.extraQueryParameters)
+            );
+        }
+        /**
+         * Need to get information about the client to populate request correctly
+         * For example: client id, client capabilities
+         */
+        const tokenRequest: TokenRequest = {
+            userObjectId: request.account?.homeAccountId,
```
codexeon, line 63: it would be good to be able to pass some account id in ssoSilent to help disambiguate accounts if the host is signed-in with multiple accounts. ssoSilent already accepts loginHint, although this is not as reliable as accountId. userObjectId sounds like it could be a reliable identifier, if allowed to pass as a string without providing full account object.

lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -0,0 +1,554 @@
...
+     * Removes callback with provided id from callback array
+     * @param callbackId
+     */
+    removeEventCallback(callbackId: string): void {
+        this.eventHandler.removeEventCallback(callbackId);
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    addPerformanceCallback(callback: PerformanceCallbackFunction): string {
+        throw NestedAppAuthError.createUnsupportedError();
```
codexeon, line 314: it would be better to just no-op than to throw an exception, since the app calling this API might suddenly fail when running in nested context. Could this actually be shared in the base class, it doesn't seem specific to nested apps?

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -20,25 +20,34 @@ import {
...
 
     public static async createPublicClientApplication(
         configuration: Configuration
     ): Promise<IPublicClientApplication> {
         const factory = new ControllerFactory(configuration);
         const controller = await factory.createController();
-        const pca = new PublicClientApplication(configuration, controller);
-
+        let pca;
+        if (controller !== null) {
```
sameerag: nit: 
```suggestion
        pca = controller ? new PublicClientApplication(configuration, controller) : new PublicClientApplication(configuraration);
```

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -65,21 +74,30 @@ export class PublicClientApplication implements IPublicClientApplication {
...
      */
     public constructor(configuration: Configuration, controller?: IController) {
+        this.configuration = configuration;
         if (controller) {
             this.controller = controller;
         } else {
-            const standardOperatingContext = new StandardOperatingContext(
-                configuration
+            const operatingContext = new UnknownOperatingContext(configuration);
+            this.controller = new UnknownOperatingContextController(
```
konstantin-msft: MVC model and dynamic imports allow us splitting up modules gracefully. I would rather prefer all this functionality to be incorporated into a separate `msal-naa` module to keep status quo as well as aim for the smaller `msal-browser` bundle size.

konstantin-msft (dismissed):

lib/msal-browser/src/naa/BridgeRequestEnvelope.ts
```diff
@@ -0,0 +1,47 @@
...
+    | "GetTokenPopup";
+
+export type BridgeRequestEnvelope = {
+    messageType: "NestedAppAuthRequest";
+    method: BridgeMethods;
+    sendTime?: number; // Assume this is epoch
+    clientLibrary?: string;
+    clientLibraryVersion?: string;
+    requestId: string;
+    apiKey?: string;
```
codexeon: can remove this parameter

lib/msal-browser/src/naa/BridgeProxy.ts
```diff
@@ -0,0 +1,232 @@
...
+            | AccountByHomeIdRequest
+            | AccountByLocalIdRequest
+            | AccountByUsernameRequest
+            | undefined
+    ): Promise<TResponse> {
+        const message: BridgeRequestEnvelope = {
+            messageType: "NestedAppAuthRequest",
+            method: method,
+            requestId: BridgeProxy.getRandomId(),
+            apiKey: "test_key",
```
codexeon: remove this parameter

lib/msal-browser/src/app/PublicClientApplication.ts
sameerag: +1, Dynamic modules only help with download control not the bundle size from what I gathered. Can we redesign to entirely install two different modules based on the OS context? It will be cleaner and also maintaining `msal-naa` as a layer on `msal-browser` is more manageable. Only downside I can think if is that it will be a new SDK for apps which may go against the initial promise. We can still mask it I assume by loading it internally.

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
shoatman: This is what it was called in the code.... I think because it was being used for all token types.... now it's just called TokenClaims (there is no token of type AuthToken... pure abstraction)

lib/msal-browser/src/app/PublicClientApplication.ts
shoatman: That's what webpack is doing for consumers of the library.

lib/msal-browser/src/naa/BridgeError.ts
shoatman: Yes, I know, i can see that when running the tests.... it's a bit unfortunate that there is no other field which we can use to identity the type other than status... I consider this funky TypeScript anyway... i'll remove code.

lib/msal-browser/src/controllers/ControllerFactory.ts
shoatman: someone fixed this....

lib/msal-browser/src/controllers/NestedAppAuthController.ts
shoatman: Not sure I followed this.  I didn't see a state parameter on the SilentRequest or it's parents.

lib/msal-browser/src/controllers/NestedAppAuthController.ts
shoatman: It should be returned.  I fixed this.

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
shoatman: Going to leave this as a feature request for the team taking over

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
shoatman: I add the id_token and id_token claims.... the rest of the fields are optional....

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
shoatman: I'm going to create a work item for the TAOS team to take this on once they start.

shoatman: Work item created under Alicia's Feature: https://office.visualstudio.com/MetaOS/_workitems/edit/8426403

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -68,21 +77,30 @@ export class PublicClientApplication implements IPublicClientApplication {
      * @param IController Optional parameter to explictly set the controller. (Will be removed when we remove public constructor)
      */
     public constructor(configuration: Configuration, controller?: IController) {
+        this.configuration = configuration;
         if (controller) {
             this.controller = controller;
         } else {
-            const standardOperatingContext = new StandardOperatingContext(
-                configuration
+            const operatingContext = new UnknownOperatingContext(configuration);
```
tnorling: Would this not be a breaking change? How do existing consumers of the library calling `new PublicClientApplication(configuration)` get the standard controller?

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -68,21 +77,30 @@ export class PublicClientApplication implements IPublicClientApplication {
...
         }
     }
 
     /**
      * Initializer function to perform async startup tasks such as connecting to WAM extension
      */
     async initialize(): Promise<void> {
-        return this.controller.initialize();
+        if (this.controller instanceof UnknownOperatingContextController) {
+            const factory = new ControllerFactory(this.configuration);
```
tnorling: Ah so the StandardController gets built in initialize instead? How does that work for APIs that don't require initialize (getAllAccounts for instance) that may be invoked before initialize?

lib/msal-browser/src/controllers/UnknownOperatingContextController.ts
```diff
@@ -0,0 +1,530 @@
...
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    getAccount(accountFilter: AccountFilter): AccountInfo | null {
+        BrowserUtils.blockAPICallsBeforeInitialize(this.initialized);
+        BrowserUtils.blockNonBrowserEnvironment(this.isBrowserEnvironment);
+        return null;
+    }
+
+    getAccountByHomeId(homeAccountId: string): AccountInfo | null {
```
tnorling, line 131: Not a fan of the code duplication this introduces. Any way to avoid reimplementing all of these?

lib/msal-browser/src/controllers/UnknownOperatingContextController.ts
```diff
@@ -0,0 +1,530 @@
...
+            return account;
+        } else {
+            this.logger.verbose(
+                "getAccountByUsername: No matching account found, returning null"
+            );
+            return null;
+        }
+    }
+    getAllAccounts(): AccountInfo[] {
+        BrowserUtils.blockAPICallsBeforeInitialize(this.initialized);
```
tnorling: Why is this blocked?

lib/msal-browser/src/controllers/UnknownOperatingContextController.ts
```diff
@@ -0,0 +1,530 @@
...
+    }
+    acquireTokenByCode(
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        request: AuthorizationCodeRequest
+    ): Promise<AuthenticationResult> {
+        BrowserUtils.blockAPICallsBeforeInitialize(this.initialized);
+        BrowserUtils.blockNonBrowserEnvironment(this.isBrowserEnvironment);
+        return {} as Promise<AuthenticationResult>;
+    }
+    acquireTokenNative(
```
tnorling, line 177: Shouldn't need this, not a public API

lib/msal-browser/src/controllers/UnknownOperatingContextController.ts
```diff
@@ -0,0 +1,530 @@
...
+                      | "codeChallengeMethod"
+                      | "requestedClaimsHash"
+                      | "nativeBroker"
+                  >
+              >
+    ): string {
+        BrowserUtils.blockAPICallsBeforeInitialize(this.initialized);
+        BrowserUtils.blockNonBrowserEnvironment(this.isBrowserEnvironment);
+        return "";
+    }
```
tnorling: All of these shouldn't be needed

lib/msal-browser/src/controllers/UnknownOperatingContextController.ts
```diff
@@ -0,0 +1,530 @@
...
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        apiId: ApiId,
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        accountId?: string | undefined
+    ): Promise<AuthenticationResult> {
+        BrowserUtils.blockAPICallsBeforeInitialize(this.initialized);
+        BrowserUtils.blockNonBrowserEnvironment(this.isBrowserEnvironment);
+        return {} as Promise<AuthenticationResult>;
+    }
+    acquireTokenByRefreshToken(
```
tnorling, line 201: Dont need this

lib/msal-browser/src/controllers/UnknownOperatingContextController.ts
```diff
@@ -0,0 +1,530 @@
...
+    preflightBrowserEnvironmentCheck(
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        interactionType: InteractionType,
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        isAppEmbedded?: boolean | undefined
+    ): void {
+        BrowserUtils.blockAPICallsBeforeInitialize(this.initialized);
+        BrowserUtils.blockNonBrowserEnvironment(this.isBrowserEnvironment);
+    }
+    canUseNative(
```
tnorling: Not needed

lib/msal-browser/src/controllers/UnknownOperatingContextController.ts
```diff
@@ -0,0 +1,530 @@
...
+        return {} as PopupClient;
+    }
+    createSilentIframeClient(
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        correlationId?: string | undefined
+    ): SilentIframeClient {
+        BrowserUtils.blockAPICallsBeforeInitialize(this.initialized);
+        BrowserUtils.blockNonBrowserEnvironment(this.isBrowserEnvironment);
+        return {} as SilentIframeClient;
+    }
```
tnorling: Not needed

lib/msal-browser/src/error/NestedAppAuthError.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export class NestedAppAuthError extends AuthError {
+    constructor(errorCode: string, errorMessage?: string) {
+        super(errorCode, errorMessage);
+
+        Object.setPrototypeOf(this, NestedAppAuthError.prototype);
+        this.name = "NestedAppAuthError";
+    }
+
+    public static createUnsupportedError(): NestedAppAuthError {
```
tnorling, line 26: Suggest refactoring this file to follow the new convention set by the rest of our errors (better for minification/treeshaking)

lib/msal-browser/src/index.ts
```diff
@@ -13,6 +13,11 @@
  * Breaking changes to these APIs will be shipped under a minor version, instead of a major version.
  */
 
+/*
+ * import * as internals from "./internals";
+ * export { internals };
+ */
+
```
tnorling: Please remove
```suggestion
```

lib/msal-browser/src/naa/AccountInfo.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export type AccountInfo = {
+    homeAccountId: string;
+    environment: string;
+    tenantId: string;
+    username: string;
+    localAccountId: string;
+    name?: string;
+    idToken?: string; // idTokenClaims can be parsed from idToken in MSAL.js
+    platformBrokerId?: string; // Used by WAM previous called nativeAccountId
```
tnorling, line 14: Curious why you can't just use the existing nativeAccountId to save you from redefining this type?

If absolutely necessary I would suggest redefining to inherit the existing type so you get new additions out of the box

```
export type AccountInfo = Omit<CommonAccountInfo, "nativeAccountId"> & {
platformBrokerId?: string
}
```

lib/msal-browser/src/naa/AccountRequests.ts
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+export type AccountByHomeIdRequest = {
```
tnorling, line 6: Why objects with a single property instead of just using the property directly?

lib/msal-browser/src/naa/BridgeCapabilities.ts
```diff
@@ -0,0 +1,8 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+export interface BridgeCapabilities {
+    queryAccount: boolean;
```
tnorling, line 7: Same question here

samples/msal-browser-samples/OfficeAddin/package.json
```diff
@@ -0,0 +1,66 @@
...
+        "prettier": "office-addin-lint prettier",
+        "start": "office-addin-debugging start manifest.xml",
+        "start:desktop": "office-addin-debugging start manifest.xml desktop",
+        "start:web": "office-addin-debugging start manifest.xml web",
+        "stop": "office-addin-debugging stop manifest.xml",
+        "validate": "office-addin-manifest validate manifest.xml",
+        "watch": "webpack --mode development --watch"
+    },
+    "dependencies": {
+        "@azure/msal-browser": "file:../../../lib/msal-browser",
```
tnorling: This can reference the msal-browser version so that it can be consumed both locally via workspaces or via published npm. I'll leave it up to you if you think that would be useful

lib/msal-react/test/components/AuthenticatedTemplate.spec.tsx
```diff
@@ -26,8 +26,9 @@ describe("AuthenticatedTemplate tests", () => {
         },
     };
 
-    beforeEach(() => {
+    beforeEach(async () => {
         pca = new PublicClientApplication(msalConfig);
+        await pca.initialize();
```
tnorling, line 31: It's not clear to me why this is needed in each of these test files. msal-react calls this under the hood and these tests pass today

lib/msal-browser/test/setupCrypto.ts
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import crypto from "crypto";
```
tnorling: Please remove this file, it should be addressed by the shared config

lib/msal-browser/test/telemetry/BrowserPerformanceClient.spec.ts
```diff
@@ -1,4 +1,14 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import {
+    ApplicationTelemetry,
+    Logger,
+    PerformanceEvents,
+} from "@azure/msal-common";
+import { name, version } from "../../src/packageMetadata";
```
tnorling, line 11: Reason for additional imports?

lib/msal-browser/test/interaction_client/SilentCacheClient.spec.ts
```diff
@@ -78,13 +78,14 @@ describe("SilentCacheClient", () => {
...
 
-    beforeEach(() => {
+    beforeEach(async () => {
         pca = new PublicClientApplication({
             auth: {
                 clientId: TEST_CONFIG.MSAL_CLIENT_ID,
             },
         });
 
+        await pca.initialize();
```
tnorling, line 88: I'm curious what changed that necessitates adding this everywhere in this PR specifically...

lib/msal-browser/src/naa/TokenRequest.ts
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+export type TokenRequest = {
```
tnorling, line 6: Can you extend one of the existing request types? e.g.

```
BaseAuthRequest & {
newProperty: string 
}
```

lib/msal-browser/src/naa/TokenResponse.ts
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { AccountInfo } from "./AccountInfo";
+
+export type TokenResponse = {
```
tnorling, line 8: Reason for not using the existing AuthenticationResult type?

lib/msal-browser/src/naa/TokenResponse.ts
```diff
@@ -0,0 +1,61 @@
...
+    id_token: string;
+    properties: TokenResponseProperties | null;
+    scope: string;
+    state: string;
+    shr?: string; // token binding enabled at native layer it is the access token, not the signing keys
+    extendedLifetimeToken?: boolean;
+};
+
+export type TokenResponseProperties = {
+    MATS?: string;
```
tnorling, line 22: We have a MATS type defined if you want to re-use that

lib/msal-browser/src/naa/BridgeStatusCode.ts
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+export enum BridgeStatusCode {
```
tnorling, line 6: Please define as an object literal instead

lib/msal-browser/src/app/PublicClientApplication.ts
shoatman: They don't ... why would someone want a specific implementation?  That's the point of the controller interface.

lib/msal-browser/src/app/PublicClientApplication.ts
shoatman: We had this conversation in the Teams channel... I don't buy the it's sync in standard controller so it shouldn't require from the public API point of view a call to initialize.

lib/msal-browser/src/controllers/UnknownOperatingContextController.ts
shoatman: I think this was added recently.  And I wasn't familiar with how to implement it.

lib/msal-browser/src/controllers/UnknownOperatingContextController.ts
shoatman: Yep.  Left over from the stuff that leaked in previously.  I'll remove.

lib/msal-browser/src/controllers/UnknownOperatingContextController.ts
shoatman: That's true now :-)

lib/msal-browser/test/telemetry/BrowserPerformanceClient.spec.ts
shoatman: Which ones?

lib/msal-browser/test/interaction_client/SilentCacheClient.spec.ts
shoatman: Where do you think msal-browser calls initialize?

lib/msal-browser/src/naa/TokenRequest.ts
shoatman: This is not an interface owned by us... it's defined by the bridge.

lib/msal-browser/src/naa/TokenResponse.ts
shoatman: Owned by bridge.

lib/msal-browser/src/naa/TokenResponse.ts
shoatman: Owned by bridge.

lib/msal-browser/src/controllers/UnknownOperatingContextController.ts
shoatman: What are you thoughts on placing the implementation only in UnknownOperatingContextController?  This controller is always downloaded.  Another option would be to have a collection of functions that perform these operations.

lib/msal-browser/src/app/PublicClientApplication.ts
shoatman: The default controller is UnknownOperatingContextController... and it has the implementation for those methods.

lib/msal-browser/src/index.ts
shoatman: done.

samples/msal-browser-samples/OfficeAddin/package.json
shoatman: Yes.  I see that the npm workspace is up and running now.  I'll update this.

lib/msal-browser/test/setupCrypto.ts
shoatman: done.

lib/msal-react/test/components/AuthenticatedTemplate.spec.tsx
shoatman: You won't have the standardcontroller until initialize is called.... in this way if you don't need it... we don't have to download it....

samples/msal-browser-samples/OfficeAddin/package.json
shoatman: done.

lib/msal-browser/test/interaction_client/SilentCacheClient.spec.ts
shoatman: Today... standard controller is automatically there.... in this PR... it's only downloaded when needed.  That was the point of introducing the controller.  Get the controller that you need for the context you find yourself running in.  For example: you could have a WAM Controller completely separate from the non-wam code instead of all the if/else statements in the standard controller implementation.  (Meaning a much smaller download... since today you get both implementations).  I don't personally buy into the we may need to fallback inline argument... if WAM is that unreliable.... seems like there are bigger problems.

lib/msal-browser/src/controllers/UnknownOperatingContextController.ts
shoatman: Fixed.

lib/msal-browser/src/controllers/UnknownOperatingContextController.ts
shoatman: Fixed.

lib/msal-browser/src/naa/AccountInfo.ts
shoatman: This is the bridge model.  If the bridge model and the "native" model converge at some point that could happen.

tnorling: Isn't the "bridge model" a new thing we're defining and building? Why are we not designing the thing to use the interfaces we already have?

lib/msal-browser/src/app/PublicClientApplication.ts
sameerag: name change: `customOperatingContext`?

lib/msal-browser/src/naa/AccountInfo.ts
shoatman: Because in this case the bridge interface was design by the authors of the bridge.  Not us.

lib/msal-browser/src/controllers/ControllerFactory.ts
```diff
@@ -25,31 +25,42 @@ export class ControllerFactory {
         this.logger = new Logger(loggerOptions, name, version);
     }
 
-    async createController(): Promise<IController> {
+    async createV3Controller(): Promise<IController> {
```
tnorling, line 28: nit: Prefer that this remain `createController` and the other one becomes `createNextController` for now

change/@azure-msal-react-e44ef43a-73f0-4eb4-99e6-8cceeb26058f.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
tnorling: No changes to published package 
```suggestion
  "type": "none",
```

change/@azure-msal-browser-50170fbc-d837-4531-83f6-28b62449a07b.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Added new PublicClientNext and additional Operating Contexts and associated Controllers; Initially to address NestedAppAuth.",
+  "packageName": "@azure/msal-browser",
+  "email": "email not defined",
```
tnorling, line 5: Suggest updating email

change/@azure-msal-browser-50170fbc-d837-4531-83f6-28b62449a07b.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Added new PublicClientNext and additional Operating Contexts and associated Controllers; Initially to address NestedAppAuth.",
```
tnorling:
```suggestion
  "comment": "Added new PublicClientNext and additional Operating Contexts and associated Controllers; Initially to address NestedAppAuth. #6359",
```

change/@azure-msal-common-497a7ea5-49f3-4b9b-ac83-55d1410db08c.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Added some new error codes/messages for NestedAppAuth.",
```
tnorling:
```suggestion
  "comment": "Added some new error codes/messages for NestedAppAuth. #6359",
```

lib/msal-react/test/components/AuthenticatedTemplate.spec.tsx
tnorling: Are all these test changes still needed now that you have a separate export?

lib/msal-browser/test/app/PublicClientApplication.spec.ts
```diff
@@ -207,7 +222,12 @@ describe("PublicClientApplication.ts Class Unit Tests", () => {
             sinon.restore();
         });
 
-        it("handles concurrent calls", async () => {
+        /**
+         * TODO: Speak to someone on MSAL.js team about how this test is supposed to work
```
tnorling, line 226: Several TODOs in this file - have these been resolved and can we remove?

lib/msal-browser/test/app/PublicClientApplication.spec.ts
```diff
@@ -207,7 +222,12 @@ describe("PublicClientApplication.ts Class Unit Tests", () => {
...
+         * Currently the test fails as a result of the fact that the spys do not record anything
+         * Most likely because this test is operating against PCA instead of StandardController
+         */
+        xit("handles concurrent calls", async () => {
```
tnorling, line 230: xit?

tnorling: There's several in this file

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
```diff
@@ -0,0 +1,252 @@
...
+                case BridgeStatusCode.ACCOUNT_UNAVAILABLE:
+                    return new ClientAuthError(
+                        ClientAuthErrorCodes.noAccountFound
+                    );
+                case BridgeStatusCode.DISABLED:
+                    return new ClientAuthError(
+                        ClientAuthErrorCodes.nestedAppAuthBridgeDisabled
+                    );
+                case BridgeStatusCode.NESTED_APP_AUTH_UNAVAILABLE:
+                    return new ClientAuthError(error.code, error.description);
```
tnorling, line 236: Should either define a code for this or let this hit default case

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
```diff
@@ -0,0 +1,252 @@
...
+     *     nativeAccountId?: string;
+     *     authorityType?: string;
+     * };
+     */
+    public fromNaaAccountInfo(fromAccount: NaaAccountInfo): MsalAccountInfo {
+        let tokenClaims: TokenClaims | undefined;
+        if (fromAccount.idToken !== undefined) {
+            tokenClaims = AuthToken.extractTokenClaims(
+                fromAccount.idToken,
+                this.crypto.base64Decode
```
tnorling, line 184: Suggest directly passing in the base64decode function exported by encode/Base64Decode.ts instead. NAA likely does not need ICrypto/CryptoOps at all

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
```diff
@@ -0,0 +1,252 @@
...
+            requestId: "",
+            extExpiresOn: expiresOn,
+            state: response.state,
+        };
+
+        return authenticationResult;
+    }
+
+    /*
+     *  export type AccountInfo = {
```
tnorling, line 158: This comment still needed?

lib/msal-browser/src/naa/BridgeProxy.ts
```diff
@@ -0,0 +1,230 @@
...
+
+/**
+ * BridgeProxy
+ * Provides a proxy for accessing a bridge to a host app and/or
+ * platform broker
+ */
+export class BridgeProxy implements IBridgeProxy {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    static bridgeRequests: any[] = [];
+    static crypto: Crypto;
```
tnorling, line 36: Reason for static member rather than just accessing window.crypto directly when needed?

lib/msal-browser/src/controllers/UnknownOperatingContextController.ts
tnorling: Now that you've moved to PublicClientNext where you expect initialize to be called before all of these, are these implementations still needed? Can we return null / empty array everywhere?

lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -0,0 +1,525 @@
...
+
+        this.nestedAppAuthAdapter = new NestedAppAuthAdapter(
+            this.config.auth.clientId,
+            this.config.auth.clientCapabilities,
+            this.browserCrypto,
+            this.logger
+        );
+    }
+    getBrowserStorage(): BrowserCacheManager {
+        throw NestedAppAuthError.createUnsupportedError();
```
tnorling, line 99: Do you need to define this at all anymore?

tnorling: I see a few unsupported APIs that can likely be removed here

lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -0,0 +1,525 @@
...
+        // Set the configuration.
+        this.config = operatingContext.getConfig();
+
+        // Initialize logger
+        this.logger = this.operatingContext.getLogger();
+        // Initialize performance client
+        this.performanceClient = this.config.telemetry.client;
+
+        // Initialize the crypto class.
+        this.browserCrypto = operatingContext.isBrowserEnvironment()
```
tnorling, line 85: Do you need to sign tokens inside NAA? If not you can remove this and use the encoding/decoding/guid generation functions directly

tnorling (commented):
Still generally concerned about the number of test changes so want to confirm that these are no longer necessary with the change in approach

lib/msal-browser/test/telemetry/BrowserPerformanceClient.spec.ts
tnorling: All of them, looks like adding imports is the only change in this file, want to understand why

lib/msal-browser/src/controllers/UnknownOperatingContextController.ts
shoatman: This address the SSR rendering context.  If the window object is available.... initialize will just keep this controller.

lib/msal-browser/src/controllers/NestedAppAuthController.ts
shoatman: Seems like an optimization that TAOS can pick up.

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
shoatman: This is for quick and easy reference.   Do these get left behind after bundling/minification?  I don't remember.

tnorling: This type of comment tends to get outdated pretty quick so I prefer to avoid these but I'll leave it up to you and TAOS. Not sure what minifier TAOS is using but I would expect it to strip comments so that part is not a concern. Our minifier only runs on the UMD output which I assume you aren't using for this use case.

tnorling (approved):
Approving as my remaining suggestions are non-blocking but please do create work items for anything you plan to address as a follow up or hand off

lib/msal-browser/src/naa/AccountInfo.ts
sameerag: native and platform is going to be confusing. I will mark it as a follow up with Taos to unify this.

lib/msal-browser/src/naa/AccountRequests.ts
sameerag: Agree. Could bloat the code for no reason.

sameerag (approved):

PR #6365, opened 2023-08-18
Fix Readme links

tnorling:
Relative links don't work properly when clicked from npm - update the migration doc links to absolute urls


konstantin-msft (approved):
Thanks for addressing this!

jo-arroyo (approved):

PR #6374, opened 2023-08-18
New API to clear cache

lalimasharda:
New SilentCacheClient API to clear cache silently using the clearCacheOnLogout() method.


lib/msal-browser/src/controllers/StandardController.ts
```diff
@@ -1124,6 +1124,16 @@ export class StandardController implements IController {
         }
     }
 
+    /**
+     * Creates a cache interaction client to clear broswer cache.
+     * @param logoutRequest
+     */
+    async clearCache(logoutRequest?: EndSessionRequest): Promise<void> {
+        const correlationId = this.getRequestCorrelationId(logoutRequest);
```
sameerag, line 1133: if `logoutRequest` is optional, all of these are undefined. What is the default behavior?

lib/msal-browser/src/interaction_client/SilentCacheClient.ts
```diff
@@ -78,13 +79,13 @@ export class SilentCacheClient extends StandardInteractionClient {
...
-     * Currently Unsupported
+     * API to silenty clear the browser cache.
+     * @param logoutRequest
      */
-    logout(): Promise<void> {
-        // Synchronous so we must reject
-        return Promise.reject(
-            BrowserAuthError.createSilentLogoutUnsupportedError()
-        );
+    logout(logoutRequest?: EndSessionRequest): Promise<void> {
```
sameerag: Same here, I do not see validation of the optional param or the fallback code when `logout` is called without the `request`.

lib/msal-browser/src/controllers/StandardController.ts
lalimasharda: This API exposes clearCacheOnLogout() which clears only account related items from cache if account is provided in the logout request. And if account or the logout request is not provided, it will clear out the entire cache by default.

tnorling: getRequestCorrelationId will either return the correlationId passed in or create a new one if it was undefined

lib/msal-browser/src/interaction_client/SilentCacheClient.ts
tnorling: What fallback are you expecting here? The only relevant piece for clearing cache is the account. If account was passed we will clear artifacts for just that account - if it was not we will clear artifacts for all accounts. This logic exists in clearCacheOnLogout

lib/msal-browser/src/controllers/StandardController.ts
```diff
@@ -1124,6 +1124,16 @@ export class StandardController implements IController {
         }
     }
 
+    /**
+     * Creates a cache interaction client to clear broswer cache.
+     * @param logoutRequest
+     */
+    async clearCache(logoutRequest?: EndSessionRequest): Promise<void> {
```
tnorling: I wonder if we should accept a smaller request type here as many of the params aren't relevant. i.e. ClearCacheRequest with optional members correlationId and account

lib/msal-browser/src/interaction_client/SilentCacheClient.ts
lalimasharda: I don't think we need fallback code here since we are checking for account in the `clearCacheOnLogout` method. It is not a required parameter, and the default case is being handled down the line.

lib/msal-browser/src/controllers/StandardController.ts
lalimasharda: Okay, that makes sense. I can create another request type.

lib/msal-browser/src/interaction_client/SilentCacheClient.ts
sameerag: Ah okay. So `EndRequest -> Account` and we check for account. Makes sense. Next q I had is already covered before, can we thin down the `request` in that case.

lib/msal-browser/src/controllers/StandardController.ts
```diff
@@ -1124,6 +1124,16 @@ export class StandardController implements IController {
...
+        const correlationId = this.getRequestCorrelationId(logoutRequest);
+        const cacheClient = this.createSilentCacheClient(correlationId);
```
sameerag, line 1134: Do we need an instance of `silentCache` or can we call `logout` independently? It is not really dependent on the client is it? If it is, would like to make it independent if we can.

sameerag (commented):
Can we add some description please?

lib/msal-browser/src/app/IPublicClientApplication.ts
```diff
@@ -67,6 +67,7 @@ export interface IPublicClientApplication {
             | RedirectRequest
             | PopupRequest
     ): Promise<void>;
+    clearCache(logoutRequest?: EndSessionRequest): Promise<void>;
```
tnorling:
```suggestion
    clearCache(logoutRequest?: ClearCacheRequest): Promise<void>;
```

lib/msal-browser/src/controllers/StandardController.ts
tnorling: I think the interaction clients could stand to be de-classed but I'd like to take that as its own PR. For now this is consistent with our other methods so I'm ok with it

lib/msal-browser/src/request/ClearCacheRequest.ts
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * ClearCacheRequest
+ * - account                - Account object that will be logged out of. All tokens tied to this account will be cleared.
+ * - correlationId          - Unique GUID set per request to trace a request end-to-end for telemetry purposes.
+ * - logoutHint             - A string that specifies the account that is being logged out in order to skip the server account picker on logout.
+ */
+export type ClearCacheRequest = {
+    correlationId?: string;
+    account?: AccountInfo | null;
+    loginHint?: string;
```
tnorling: loginHint? Is this needed?

tnorling (approved):

sameerag (approved):
Approved. I think we can get rid of the `silentCache` class later, for now this works.

PR #6380, opened 2023-08-22
Change relative import to extensioned variations for Node16 and NodeNext type support

### LINKED ISSUE #6377: Types are broken for moduleResolution: Node16 and NodeNext
### Core Library

MSAL Node (@azure/msal-node)

### Core Library Version

2.0.2

### Wrapper Library

Not Applicable

### Wrapper Library Version

None

Public, Confidential

### Description

reminiscent of #6151 

I'm using latest typescript, my server is fully esm.
```json
{
  "type": "module"
}
```
```json
{
  "compilerOptions": {
    "module": "NodeNext",
    "moduleResolution": "NodeNext"
  }
}
```
and all the types from msal are broken. This is because typescript doesn't map `.mjs` to `.d.ts` naturally. So you get this, if you're looking inside:
image

Looking at `index.d.ts`, for `esm` the imports don't point to any _typed_ file,
image

If directed explicitly to the `d.ts`, this is resolved

image

### **!NOTE! This is not enough, for this to work _all_  imports in `d.ts` files have to explicitly point to `d.ts` files**

I'm not sure how to solve this, but from my personal experience I found that meme extensions are very painful to work with

### Error Message

_No response_

### Msal Logs

_No response_

### MSAL Configuration

```javascript
{}
```

### Relevant Code Snippets

```javascript
import { CryptoProvider } from "@azure/msal-node"
// or anything, really
```

### Reproduction Steps

1. create modern esm node project
2. install `@azure/msal-node`
3. ??????
4. profit

### Expected Behavior

Typed correctly

### Identity Provider

Azure AD / MSA

None (Server)

### Regression

@azure/msal-node@~2

### Source

External (Customer)

datner:
- chore(msal-node): add $schema props for schema validation
- test(msal-node): add support for .js extension import
- build(msal-node): remove redundant pass
- feat(msal-node): change all relative imports to .js

I've made meaningful commits, so I recommend you review them instead of all the files at once.

The other solutions are:
1. to manually transform d.ts files after the build, which is a great way to introduce both complexity and brittleness to the project
2. use self-imports, allowing many entry-points `import { Configuration } from "@azure/msal-node/config/Configuration"` letting typescript and node to each resolve it in the way they understand. But this is a massive design-shift
3. using package.json `"imports"` entry. Just changing one problem for another, not a real solution and doesn't scale
4. emitting 2 `d.ts` versions, `d.mts` and `d.cts`. Thats possible but just.. redundant?

There are a few oddities, like the meme extensions importing only within their meme domain, but `d.ts` referencing non-existent `.js` files. Which is fine because it will just direct _Typescript_ to the correct `d.ts` and there isn't an interface difference between cjs and esm. Also `jest` had to be adjusted since it never got the memo that 3 LTS versions ago nodejs started supporting and advocating for `.js` extensions. If we ditch one of the meme extensions then this would be visually less scary, though it should be just as safe as it is now.

closes #6377

personal note:
Please don't bike-shed this pr for months, yes it would probably break node versions before 12 but come on. [even 16 is literally days away from deprecation](https://github.com/nodejs/release#release-schedule), it would be absolutely unreasonable.. This is over all a small change with a visually large foot print

### DIFF
--- a/lib/msal-node/jest.config.cjs
+++ b/lib/msal-node/jest.config.cjs
@@ -5,19 +5,18 @@
 
 module.exports = {
     verbose: true,
-    moduleFileExtensions: [
-        "ts",
-        "tsx",
-        "js",
-        "json",
-        "jsx",
-        "node"
-    ],
-    testMatch: [
-        "<rootDir>/test/**/*.spec.ts"
-    ],
+    moduleFileExtensions: ["ts", "tsx", "js", "json", "jsx", "node"],
+    testMatch: ["<rootDir>/test/**/*.spec.ts"],
     transform: {
         "^.+\\.(ts|tsx)$": "ts-jest",
     },
-    coverageReporters: [["lcov", {"projectRoot": "../../"}]]
+    /**
+     * "Not a ts-jest issue" but a jest one. Consider vitest?
+     *  this morphs "./some/relative/path.js" to "./some/relative/path"
+     *  https://github.com/kulshekhar/ts-jest/issues/1057
+     */
+    moduleNameMapper: {
+        "^(\\.\\.?\\/.+)\\.js$": "$1",
+    },
+    coverageReporters: [["lcov", { projectRoot: "../../" }]],
 };
--- a/lib/msal-node/package.json
+++ b/lib/msal-node/package.json
@@ -1,4 +1,5 @@
 {
+  "$schema": "https://json.schemastore.org/package.json",
   "name": "@azure/msal-node",
   "version": "2.0.2",
   "author": {
--- a/lib/msal-node/rollup.config.js
+++ b/lib/msal-node/rollup.config.js
@@ -5,68 +5,49 @@
 
 import { nodeResolve } from "@rollup/plugin-node-resolve";
 import typescript from "@rollup/plugin-typescript";
+import { defineConfig } from "rollup";
 import pkg from "./package.json";
 
-const libraryHeader = `/*! ${pkg.name} v${pkg.version} ${new Date().toISOString().split("T")[0]} */`;
+const libraryHeader = `/*! ${pkg.name} v${pkg.version} ${
+    new Date().toISOString().split("T")[0]
+} */`;
 const useStrictHeader = "'use strict';";
 const fileHeader = `${libraryHeader}\n${useStrictHeader}`;
 
-export default [
-    {
-        // for cjs build
-        input: "src/index.ts",
-        output: {
+export default defineConfig({
+    input: "src/index.ts",
+    output: [
+        {
             dir: "dist",
             format: "cjs",
             entryFileNames: "[name].cjs",
             preserveModules: true,
             preserveModulesRoot: "src",
             banner: fileHeader,
-            sourcemap: true
+            sourcemap: true,
         },
-        treeshake: {
-            moduleSideEffects: false,
-            propertyReadSideEffects: false
-        },
-        external: [
-            ...Object.keys(pkg.dependencies || {}),
-            ...Object.keys(pkg.peerDependencies || {})
-        ],
-        plugins: [
-            typescript( {
-                typescript: require("typescript"),
-                tsconfig: "tsconfig.build.json"
-            }),
-            nodeResolve()
-        ]
-    },
-    {
-        // for esm build
-        input: "src/index.ts",
-        output: {
+        {
             dir: "dist",
             format: "es",
             entryFileNames: "[name].mjs",
             preserveModules: true,
             preserveModulesRoot: "src",
             banner: fileHeader,
-            sourcemap: true
+            sourcemap: true,
         },
-        treeshake: {
-            moduleSideEffects: false,
-            propertyReadSideEffects: false
-        },
-        external: [
-            ...Object.keys(pkg.dependencies || {}),
-            ...Object.keys(pkg.peerDependencies || {})
-        ],
-        plugins: [
-            typescript( {
-                typescript: require("typescript"),
-                tsconfig: "tsconfig.build.json"
-            }),
-            nodeResolve()
-        ]
-    }
-
-];
+    ],
+    treeshake: {
+        moduleSideEffects: false,
+        propertyReadSideEffects: false,
+    },
+    external: [
+        ...Object.keys(pkg.dependencies || {}),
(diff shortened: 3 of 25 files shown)


datner:

lib/msal-node/jest.config.cjs
```diff
@@ -19,5 +19,12 @@ module.exports = {
     transform: {
         "^.+\\.(ts|tsx)$": "ts-jest",
     },
+    /**
+     * "Not a ts-jest issue" but a jest one. Consider vitest?
+     *  https://github.com/kulshekhar/ts-jest/issues/1057
+     */
+    moduleNameMapper: {
+      "^(\\.\\.?\\/.+)\\.js$": "$1"
```
datner: this just morphs `"./some/relative/path.js"` to `"./some/relative/path"`

lib/msal-node/package.json
```diff
@@ -1,4 +1,5 @@
 {
+  "$schema": "https://json.schemastore.org/package.json",
```
datner, line 2: autocompletion and instant feedback is cool

lib/msal-node/rollup.config.js
```diff
@@ -5,17 +5,18 @@
 
 import { nodeResolve } from "@rollup/plugin-node-resolve";
 import typescript from "@rollup/plugin-typescript";
+import { defineConfig } from "rollup"
```
datner: btw prettier does not cover these files

lib/msal-node/rollup.config.js
```diff
@@ -5,17 +5,18 @@
...
 import pkg from "./package.json";
 
 const libraryHeader = `/*! ${pkg.name} v${pkg.version} ${new Date().toISOString().split("T")[0]} */`;
 const useStrictHeader = "'use strict';";
 const fileHeader = `${libraryHeader}\n${useStrictHeader}`;
 
-export default [
+export default defineConfig([
```
datner: this does not change the config, only adds types

lib/msal-node/rollup.config.js
```diff
@@ -62,11 +44,9 @@ export default [
         ],
         plugins: [
             typescript( {
-                typescript: require("typescript"),
```
datner, line 65: this is not required, its the default behavior

lib/msal-node/rollup.config.js
```diff
@@ -5,17 +5,18 @@
...
 const useStrictHeader = "'use strict';";
 const fileHeader = `${libraryHeader}\n${useStrictHeader}`;
 
-export default [
+export default defineConfig([
     {
         // for cjs build
         input: "src/index.ts",
-        output: {
+        output: [{
```
datner: we can can do multiple output instead of multiple bundling because everything else is identical

lib/msal-node/jest.config.cjs
tnorling: Could you please add this explanation to the comment above?

tnorling (commented):
Thanks for the PR! Could you please run `npm run beachball:change` from the root of the repo to generate changefiles

lib/msal-node/src/index.ts
```diff
@@ -118,4 +118,4 @@ export {
     AppTokenProviderResult,
 } from "@azure/msal-common";
```
tnorling, line 119: Do these changes need to be made for msal-common exports as well?

datner: `msal-common` is still `cjs`, once you change it to `"type": "module"` then the types will break and you'll have to move to the `.js` extension. At least thats as far as I understand

datner: This does also mean that if you remove `"type": "module"` from the distributed package.json, this would also solve the issue and with the smallest footprint. But I would consider this a hack more than a solution

tnorling: All of our libraries export both cjs and esm

lib/msal-node/jest.config.cjs
datner: Ofc

lib/msal-node/src/index.ts
datner: But they share a `package.json`, take a peek at the one both msal-common and msal-nodes ship.
This is the difference between them, and what broke typescript

datner: I'll be more specific. `msal-common` is still _written_ as a commonjs module for the simple fact that its type is not set to `"module"`-- So it defaults to `"commonjs"`. It and the rest are bundled for both `esm` and `cjs`, but the module type stays the same. Node decides which version to take wether you `import` or `require`, as outlined in your `"exports"` but can chew though both either way. the typescript type checker on the other hand, from what I understand, does not give 2 licks about how much effort we go though to bundle things universally, it knows either cjs _or_ esm. Depending on what the type of the module is-- which is decided by the package.json `"type"` field.

I'm sure you already know everything I said here, and I'm also sure I made mistakes while oversimplifying the issue, but thats the long version of why msal-common doesn't _need_ to do this yet, though it probably should

datner:
No clue how coverage changed, there was literally 0 logical changes (tests or otherwise) 
This PR introduces a 20% improvement to build speed from roughly 1.3s to about 1s 😎 

I rebased from dev, formatted, and fixed conflicts

lib/msal-node/rollup.config.js
```diff
@@ -5,68 +5,49 @@
...
-        output: {
+export default defineConfig({
+    input: "src/index.ts",
+    output: [
+        {
             dir: "dist",
             format: "cjs",
             entryFileNames: "[name].cjs",
             preserveModules: true,
             preserveModulesRoot: "src",
```
datner, line 25: seeing as the only valid entry-point for the module is `@azure/msal-node`, I don't understand why these are used

lib/msal-node/rollup.config.js
sameerag: We can change that if it helps.

datner: @sameerag  I just ran the prettier command manually, just raising a flag 😄

tnorling:
Some of our CI tasks don't have permission to run on external PRs so I've pulled this into an internal branch and have opened #6399

PR #6383, opened 2023-08-23
Add wrapper functions for telemetry & logging

tnorling:
- Introduces `invoke` and `invokeAsync` helper functions to wrap APIs with telemetry measurements and trace logging.
- Adds perf measurements for ssoSilent/iframed flows
- Moves `httpVerToken` and `refreshTokenSize` to more central place right after POST
- Remove `httpVerAuthority` is it represents the same thing as `httpVerToken` but with a different name

### Pros

- Makes it easier and less verbose to add logging & take perf measurements
- In theory should be less expensive from a package size perspective

### Cons

- Makes code a bit harder to read
- If a function is called from multiple places each callsite must wrap the function (in contrast to being baked in today)


lib/msal-common/src/utils/FunctionWrappers.ts
```diff
@@ -0,0 +1,105 @@
...
+export const invoke = <T extends Array<any>, U>(
+    callback: (...args: T) => U,
+    eventName: string,
+    logger: Logger,
+    telemetryClient?: IPerformanceClient,
+    correlationId?: string
+) => {
+    return (...args: T): U => {
+        logger.trace(`Executing function ${eventName}`);
+        const inProgressEvent = telemetryClient?.startMeasurement(
```
konstantin-msft, line 30: The main issue with a wrapper (i was trying to implement a similar solution before) is that it allows you to easily measure function execution time, but doesn't allow adding additional payload inside the function (callback) because `inProgressEvent` is not being passed to it.

tnorling: You can call `performanceClient.addFields()` anywhere you have access to a performanceClient instance. Additional fields do not require an inProgressEvent as long as the correlationId matches

sameerag (approved):
Let us track the user experience with a separate PR, approving as is. This is great improvement over redundant code.

konstantin-msft (approved):

lib/msal-browser/src/interaction_client/SilentIframeClient.ts
```diff
@@ -74,11 +75,6 @@ export class SilentIframeClient extends StandardInteractionClient {
             PerformanceEvents.SilentIframeClientAcquireToken,
             request.correlationId
         );
-        this.logger.verbose("acquireTokenByIframe called");
```
peterzenz, line 77: Did you intend to remove this logger line as well?

peterzenz (approved):
Definitely a nice readability and maintenance improvement.  Thanks.

lib/msal-browser/src/interaction_client/SilentIframeClient.ts
tnorling: Yes, the invoke functions will handle logging start/end messages as well

PR #6396, opened 2023-08-25
Add multi-tenant support [LTS]

hectormmg:
This PR:
- Adds `multiTenantAccountsEnabled` flag to auth config
- Adds `TenantProfile` type and `tenantProfiles: Map<String, TenantProfile>` optional attribute to `AccountInfo`
- Enables cross-tenant cache storage and retrieval when `multiTenantAccountsEnabled` is `true`
- Adds `getAllAccountsMultiTenant` API to allow applications to get an AccountInfo object of each home account in the cache with a map of tenant profiles for the matching cross-tenant accounts (alternate branch `multi-tenant-support-all-accounts` returns all accounts, home or guest, with their cross-tenant profiles)
- Add a multi-tenant sample demonstrating how to get accounts by tenant profile and how to get tokens for cross-tenant accounts
- Adds logic to logout to clear ALL tenant profile accounts from the cache matching the account passed in when `multiTenantAccountsEnabled` is `true`
- PENDING TESTS/DOCS

### DIFF
--- a/change/@azure-msal-browser-5ccdb56d-acb7-4277-bb93-3d847632b65f.json
+++ b/change/@azure-msal-browser-5ccdb56d-acb7-4277-bb93-3d847632b65f.json
@@ -2,6 +2,6 @@
   "type": "minor",
   "comment": "Add multi-tenant support [LTS] #6396",
   "packageName": "@azure/msal-browser",
-  "email": "hectormg@hey.com",
+  "email": "hemoral@microsoft.com",
   "dependentChangeType": "patch"
 }
--- a/change/@azure-msal-browser-5ccdb56d-acb7-4277-bb93-3d847632b65f.json
+++ b/change/@azure-msal-browser-5ccdb56d-acb7-4277-bb93-3d847632b65f.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add multi-tenant support [LTS] #6396",
+  "packageName": "@azure/msal-browser",
+  "email": "hectormg@hey.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@azure-msal-common-64829f7e-0b68-4227-b348-ba607182da56.json
+++ b/change/@azure-msal-common-64829f7e-0b68-4227-b348-ba607182da56.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add multi-tenant support [LTS] #6396",
+  "packageName": "@azure/msal-common",
+  "email": "hectormg@hey.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@azure-msal-common-64829f7e-0b68-4227-b348-ba607182da56.json
+++ b/change/@azure-msal-common-64829f7e-0b68-4227-b348-ba607182da56.json
@@ -2,6 +2,6 @@
   "type": "minor",
   "comment": "Add multi-tenant support [LTS] #6396",
   "packageName": "@azure/msal-common",
-  "email": "hectormg@hey.com",
+  "email": "hemoral@microsoft.com",
   "dependentChangeType": "patch"
 }
--- a/lib/msal-browser/docs/multi-tenant-accounts.md
+++ b/lib/msal-browser/docs/multi-tenant-accounts.md
@@ -1,0 +1,120 @@
+# Multi-tenant Support in MSAL Browser
+
+> This is an advanced document regarding the acquisition of tokensacross tenants in `msal-browser`. For basic account information, please review the [Accounts](./accounts.md) document.
+
+## Multi-tenant Accounts
+
+As of v2.39.0, MSAL Browser supports the acquisition of access tokens for accounts across multiple tenants. In order to achieve this, MSAL caches the account information and auth artifacts for a user for each tenant it has authenticated with. 
+
+This means that with the exception of refresh tokens, which are shared across tenants, the MSAL cache may store multiple accounts, access tokens, and ID tokens for a single user. Each of these account records, in a multi-tenant context, will be referred to as a **"tenant profile"** of the main account.
+
+### Tenant Profiles
+
+A tenant profile is the record of a particular account in a particular tenant. When multi-tenant support is enabled, a distinct `AccountInfo` object will be cached containing the entire tenant profile for every tenant that the user authenticates with.
+
+All tenant profiles of an account include a reference to the original or "home" account. In MSAL, this identifier is the `homeAccountId` in the `AccountInfo` object.
+
+Although full tenant profiles are actually stored as `AccountInfo` objects, MSAL exports a type called `TenantProfile` which encapsulates the information required to retrieve a particular `AccountInfo` object representing a specific tenant profile. This type is shown and explained below:
+
+```javascript
+export type TenantProfile = {
+    objectId: string;
+    tenantId: string;
+    isHomeTenant: boolean;
+};
+```
+
+| Property | Definition |
+| -------- | ---------- |
+| `objectId`| Sourced from the `oid` claim in an ID token, the `objectId` is the value that uniquely identifies a user object within a specific tenant, also referred to as `localAccountId`.|
+| `tenantId` | Source from the `tid` claim in an ID token, the `tenantId` is the value that uniquely identifies a tenant. The combination of `objectId` and `tenantId` will fully and uniquely identify a user account's specfic tenant profile.|
+| `isHomeTenant` | A flag that specifies whether the tenant profile is the home tenant profile, determined by comparing the string `"[objectId].[tenantId]"` to the account record's `homeAccountId`.|
+
+## Usage
+
+### Configuration
+
+To make use of this advanced feature, you must enable multi-tenant accounts through the `Configuration` object.
+
+```javascript
+const msalConfig = {
+	auth: {
+		clientId: "your_client_id",
+		multiTenantAccountsEnabled: true
+	}
+}
+
+const myMSALObj = new PubliClientApplication(msalConfig);
+```
+
+### Authenticating with Multiple Tenants
+
+In order to authenticate with multiple tenants, you can use either `login` or `acquireToken` APIs normally, only setting the authority to the particular tenant you are requesting tokens for.
+
+```javascript
+const msalConfig = {
+    auth: {
+        clientId: "ENTER_CLIENT_ID",
+        authority: "https://login.microsoftonline.com/HOME_TENANT", // This is the authority that MSAL will default to for requests that don't specify their own authority.
+        multiTenantAccountsEnabled: true
+    }
+}
+
+const myMSALObj = new PublicClientApplication(msalConfig);
+
+// handleRedirectPromise has to be called to resume redirect requests
+myMSALObj.handleRedirectPromise().then(handleResponse).catch(err => {
+    console.error(err);
+});
+
+// Authority isn't specified because it is already the default set in the MSAL Config object
+const homeTenantRequest = {
+    scopes: ["HOME_TENANT_SCOPE"]
+};
+
+// Guest tenant authority overrides the MSAL config authority, meaning the user can authenticate with the tenant they are a guest in and acquire tokens for it
+const guestTenantRequest = {
+    scopes: ["GUEST_TENANT_SCOPE"],
+    authority: "https://login.microsoftonline.com/GUEST_TENANT"
+};
+
+const homeTenatAuthResponse = await myMSALObj.loginPopup(homeTenantRequest);
+const guestTenantAuthResponse = await myMSALObj.loginPopup(homeTenantRequest);
+```
+
+Once both login requests are successfully completed and the user has authenticated with each tenant, there will be one `AccountInfo` object, one ID token and one access token for each tenant. There will also be a single refresh token that will be used for both in the cache.
+
+
+### Acquiring Tokens Silently for Multiple Tenants
+
+In order to acquire cached tokens for a particular tenant, the `AccountInfo` object for that particular tenant must be passed into `acquireTokenSilent`. In order to facilitate this, when multi-tenant support is enabled, all returned account objects will include a `Map<String, TenantProfile>` object containing references to all the alternate tenant profiles in the cache. Also, a `getAccountByFilter()` API has been added that allows applications to pass in a `TenantProfile` object that will be used to find the `AccountInfo` object for that specific tenant profile.
+
+The sample code below shows how one can:
+
+- Filter the result of `getAllAccounts()` using the new optional `accountFilter` parameter to "flatten" the cached accounts into home accounts with a map of their tenant profiles (otherwise getAllAccounts will return the `AccountInfo` object for each tenant profile)
+- Extract the desired `TenantProfile` object from the home `AccountInfo` object
+- Use the `TenantProfile` object to get the `AccountInfo` object for that tenant profile
+- Use the guest tenant account object to acquire cached tokens that belong to it
+
+```javascript
+// When a filter is passed into getAllAccounts, it returns all cached accounts that match the filter. Use the special isHomeTenant filter to get the home accounts only.
+const homeAccount = myMSALObj.getAllAccounts({ isHomeTenant: true })[0]; // Assuming all cached accounts belong to the same user, there should only be one home account in the cache
+const tenantId = "GUEST_TENANT_ID"; // This will be the tenant you want to retrieve a cached token for
+const guestTenantProfile = homeAccount.tenantProfiles.get(tenantId);
+const guestTenantAccount = myMSALObj.getAccountByFilter({ tenantProfile: tenantProfile});
+
+
+const guestTenantAuthResponse = await myMSALObj.acquireTokenSilent({ ...guestTenantRequest, account: guestTenantAccount }).catch(async (error) => {
+        if (error instanceof msal.InteractionRequiredAuthError) {
+            // fallback to interaction when silent call fails
+            myMSALObj.acquireTokenRedirect(request);
+        } else {
+            console.error(error);
+        }
+});
+```
+
+
+## Multi-tenant Logout
+
+Calling the `logout` API with an account object passed in will result in all tenant profiles corresponding to that account being logged out and all of their account information and auth artifacts being removed from the cache.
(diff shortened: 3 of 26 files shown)


change/@azure-msal-browser-5ccdb56d-acb7-4277-bb93-3d847632b65f.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add multi-tenant support [LTS] #6396",
+  "packageName": "@azure/msal-browser",
+  "email": "hectormg@hey.com",
```
tnorling: @microsoft.com?

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -845,6 +875,30 @@ export abstract class ClientApplication {
         }
     }
 
+    /**
+     * Returns the signed in account matching the tenant profile passed in or null if no account is found.
+     * @param tenantProfile
+     * @returns The account object stored in MSAL 
+     */
+    getAccountByTenantProfile(tenantProfile: TenantProfile): AccountInfo | null {
```
tnorling: Rather than adding another getAccount API can we add a `getAccountByFilter` API so we don't need to keep adding additional APIs in the future? Let's also mark the existing getAccountBy APIs deprecated and point folks to the new one

lib/msal-browser/src/cache/BrowserCacheManager.ts
```diff
@@ -739,6 +765,15 @@ export class BrowserCacheManager extends CacheManager {
         return null;
     }
 
+    /**
+     * Returns active account with tenant profiles if found 
+     * @returns 
+     */
+    getActiveAccountMultiTenant(): AccountInfo | null {
+        const activeAccount = this.getActiveAccount();
+        return activeAccount ? this.buildMultiTenantAccount(activeAccount) : null;
```
tnorling, line 769: Could this additional logic live in the existing getActiveAccount function?

lib/msal-browser/src/cache/BrowserCacheManager.ts
```diff
@@ -343,6 +343,32 @@ export class BrowserCacheManager extends CacheManager {
...
+        super.removeAccount(accountCacheKey, true);
+
+        // Remove cross-tenant account keys from the map
+        const accountId = account.generateAccountId();
+        const accountKeys = this.getAccountKeys();
+        accountKeys.forEach((key) => {
+            if (key.indexOf(accountId) === 0) {
+                this.removeAccountKeyFromMap(key);
+            }
+        });
```
tnorling: Same question here, could this logic live in removeAccount?

lib/msal-common/src/authority/Authority.ts
```diff
@@ -856,4 +856,13 @@ export class Authority {
...
         return ciamAuthority;
     }
+
+    /**
+     * Extract tenantId from authority
+     */
+    static getTenantFromAuthorityString(authority: string): string | undefined {
+        const authorityUrl = new UrlString(authority);
+        const authorityUrlComponents = authorityUrl.getUrlComponents();
+        return authorityUrlComponents.PathSegments[0];
```
tnorling: This doesn't hold for non-AAD authorities (e.g. B2C) - is this only supported for AAD? Would it perhaps be safer to take the entire path?

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -624,6 +675,21 @@ export abstract class CacheManager implements ICacheManager {
             }
         });
 
+        // Remove other account tenant profiles for this account from cache
+        if (multiTenant) {
+            const allAccountKeys = this.getAccountKeys();
+            allAccountKeys.forEach((key) => {
+                // All account keys accross tenants will have the same account ID at the beginning (homeAccountId-environment-*)
+                if (key.indexOf(accountId) === 0) {
```
tnorling, line 730: We can use .startsWith now :)

samples/msal-browser-samples/VanillaJSTestApp2.0/app/multi-tenant/auth.js
```diff
@@ -0,0 +1,115 @@
+// Browser check variables
+// If you support IE, our recommendation is that you sign-in using Redirect APIs
+// If you as a developer are testing using Edge InPrivate mode, please add "isEdge" to the if check
+const ua = window.navigator.userAgent;
+const msie = ua.indexOf("MSIE ");
+const msie11 = ua.indexOf("Trident/");
+const msedge = ua.indexOf("Edge/");
+const isIE = msie > 0 || msie11 > 0;
+const isEdge = msedge > 0;
```
tnorling: Not needed anymore ;)

lib/msal-common/src/cache/CacheManager.ts
tnorling: nvm - missed that this was for LTS

lib/msal-browser/src/cache/BrowserCacheManager.ts
hectormmg: @tnorling It can, I wanted to avoid adding an optional parameter to the method signature, but I guess since it doesn't constitute a breaking change it's fine and probably better.

change/@azure-msal-browser-5ccdb56d-acb7-4277-bb93-3d847632b65f.json
hectormmg:
```suggestion
  "email": "hemoral@microsoft.com",
```

change/@azure-msal-common-64829f7e-0b68-4227-b348-ba607182da56.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add multi-tenant support [LTS] #6396",
+  "packageName": "@azure/msal-common",
+  "email": "hectormg@hey.com",
```
hectormmg:
```suggestion
  "email": "hemoral@microsoft.com",
```

samples/msal-browser-samples/VanillaJSTestApp2.0/app/multi-tenant/auth.js
hectormmg: Removed, thanks!

lib/msal-browser/src/cache/BrowserCacheManager.ts
hectormmg: Moved it back into removeAccount using a new optional parameter.

lib/msal-common/src/authority/Authority.ts
hectormmg: Fixed by using PathSegments.join("/") so tenant IDs for B2C are covered as well. Will add tests for this specifically as well.

lib/msal-browser/src/cache/BrowserCacheManager.ts
hectormmg: Fixed, thanks!

lib/msal-browser/src/app/ClientApplication.ts
hectormmg: Done. I did add a new account API called `getAccountsFilteredBy` that returns all the accounts that match the filter, not just the first. It enables a getAllAccounts with filter "isHomeTenant" to only return home or guest accounts depending on the app needs.

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -784,7 +796,8 @@ export abstract class ClientApplication {
             return null;
         }
 
-        const account = this.browserStorage.getAccountInfoFilteredBy({username});
+        const account: AccountInfo | null = this.browserStorage.getAccountInfoFilteredBy({username}, this.config.auth.multiTenantAccountsEnabled);
```
tnorling, line 791: Can we also add `@deprecated - Use getAccountByFilter instead` to each of the doc comments over the old getAccountBy APIs?

lib/msal-browser/src/app/ClientApplication.ts
```diff
@@ -845,6 +860,30 @@ export abstract class ClientApplication {
         }
     }
 
+    /**
+     * Returns the first account found in the cache that matches the account filter passed in.
+     * @param accountFilter 
+     * @returns The first account found in the cache matching the provided filter or null if no account could be found.
+     */
+    getAccountByFilter(accountFilter: AccountFilter): AccountInfo | null {
```
tnorling, line 862: Do we need both getAccount and getAccounts or can we get by with just one?

lib/msal-browser/src/cache/BrowserCacheManager.ts
tnorling: Doesn't look like those changes were pushed?

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -204,23 +214,103 @@ export abstract class CacheManager implements ICacheManager {
...
             return allAccounts;
         }
     }
+    
+    /**
+     * Returns all accounts with their respective tenant profiles in the cache
+     */
+    private getAllAccountsMultiTenant(): AccountInfo[] {
+        const allAccountKeys = this.getAccountKeys();
+        const allAccounts: AccountInfo[] = this.getAllAccounts();
```
tnorling: This will cause an infinite loop getAllAccounts -> getAllAccountsMultiTenant -> getAllAccounts -> ...

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -204,23 +214,103 @@ export abstract class CacheManager implements ICacheManager {
...
+        if (multiTenantAccountsEnabled) {
+            return fitleredAccounts.map((accountInfo) => { return this.buildMultiTenantAccount(accountInfo); });
+        } 
+        return fitleredAccounts;
+    }
+
+    /** 
+     * Gets first AcccountInfo object that matches the provided filters
+     */
+    getAccountInfoFilteredBy(accountFilter: AccountFilter, multiTenantAccountsEnabled?: boolean): AccountInfo | null {
```
tnorling, line 305: We've now got `getAllAccountsFilteredBy`, `getAccountInfoFilteredBy` and `getAccountsFilteredBy` - what's the difference between all of these and are they all needed? Any way to converge?

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -204,23 +214,103 @@ export abstract class CacheManager implements ICacheManager {
...
+     * @param accountFilter 
+     * @param multiTenantAccountsEnabled 
+     * @returns all accounts in the cache that match the filter
+     */
+    getAllAccountsFilteredBy(accountFilter: AccountFilter, multiTenantAccountsEnabled?: boolean): AccountInfo[] {
+        const fitleredAccounts: AccountInfo[] = this.getAccountsFilteredBy(accountFilter).map((accountEntity) => { return accountEntity.getAccountInfo(); });
+        if (multiTenantAccountsEnabled) {
+            return fitleredAccounts.map((accountInfo) => { return this.buildMultiTenantAccount(accountInfo); });
+        } 
+        return fitleredAccounts;
```
tnorling: Typo

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -204,23 +214,103 @@ export abstract class CacheManager implements ICacheManager {
...
+            });
         }
+
+        // Map tenant profile accounts to their tenantProfiles for baseAccount
+        tenantProfileAccounts.forEach((tenantProfileAccount: AccountInfo) => {
+            tenantProfiles.set(
+                tenantProfileAccount.tenantId, { 
+                    objectId: tenantProfileAccount.localAccountId,
+                    tenantId: tenantProfileAccount.tenantId,
+                    isHomeTenant: tenantProfileAccount.tenantId === tenantProfileAccount.homeAccountId.split(".")[1]
```
tnorling: Need to make sure any whitespace on either end is stripped off before comparison

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -182,16 +184,32 @@ export class ResponseHandler {
...
+                        if(crossTenantAccount){
+                            this.logger.info("Account used to refresh tokens not in persistence, but another tenant profile for requested account is in persistence, refreshed tokens will be stored in the cache");
+                        } else {
+                            this.logger.warning("Neither the account used to refresh tokens nor a cross-tenant profile of it were found in persistence, refreshed tokens will not be stored in the cache");
+                            return ResponseHandler.generateAuthenticationResult(this.cryptoObj, authority, cacheRecord, false, request, idTokenObj, requestStateObj, undefined, serverRequestId);
+                        }
+                    } else {
+                        this.logger.warning("Account used to refresh tokens not in persistence, refreshed tokens will not be stored in the cache");
+                        return ResponseHandler.generateAuthenticationResult(this.cryptoObj, authority, cacheRecord, false, request, idTokenObj, requestStateObj, undefined, serverRequestId);
```
tnorling: Looks like the same signature as above, can we simplify this?

samples/msal-browser-samples/VanillaJSTestApp2.0/app/multi-tenant/authConfig.js
```diff
@@ -0,0 +1,68 @@
+const homeTenant = "ENTER_HOME_TENANT";
+const guestTenant = "ENTER_GUEST_TENANT"
+const baseAuthority = "https://login.microsoftonline.com"
+
+// Config object to be passed to Msal on creation
+const msalConfig = {
+    auth: {
+        clientId: "ENTER_CLIENT_ID",
+        authority: `${baseAuthority}/${homeTenant}`,
+        multiTenantAccountsEnabled: true
```
tnorling, line 10: We should write some docs for this feature. Once those docs are written I would challenge whether we need a separate sample for this at all

lib/msal-browser/src/app/ClientApplication.ts
hectormmg: Done

lib/msal-common/src/cache/CacheManager.ts
hectormmg: I have no idea why this didn't break my sample but thanks for catching it! I probably missed it while refactoring. It was probably going to show up when writing the tests ofc.

hectormmg: Ok the loop wasn't infinite because the getAllAccounts call here doesn't have the multinenantAccountsEnabled parameter set to true. It did loop once though, fixing by changing to this.getAllCachedAccounts instead, which was originally getAllAccounts.

lib/msal-browser/src/app/ClientApplication.ts
hectormmg: Here's how I see it:

- getAccountByFilter - Replaces current getAccountBy APIs so they can be deprecated
- getAccountsByFilter - Allows devs to leverage MSAL's existing account filtering and (I think more importantly) enables  powerful actions in a multi-tenant scenario like getting truncated home account objects that contain their cross-tenant profiles (instead of getting an exhaustive array of every account info object in the cache each with every cross-tenant profile represented)

hectormmg: Removed `getAccountsByFilter` since it was incorporated into `getAllAccounts` with an optional filter

lib/msal-common/src/response/ResponseHandler.ts
hectormmg: Best I could do was refactor the logic out into a boolean method, still pretty tricky to flatten into a logical structure with less returns. Please let me know if you can think of a cleaner way.

lib/msal-browser/docs/multi-tenant-accounts.md
```diff
@@ -0,0 +1,120 @@
+# Multi-tenant Support in MSAL Browser
+
+> This is an advanced document regarding the acquisition of tokensacross tenants in `msal-browser`. For basic account information, please review the [Accounts](./accounts.md) document.
```
sameerag, line 3:
```suggestion
> This is an advanced feature that enables to acquire tokens across tenants in `msal-browser`. For basic account information, please review the [Accounts](./accounts.md) document.
```

lib/msal-browser/docs/multi-tenant-accounts.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Multi-tenant Accounts
+
+As of v2.39.0, MSAL Browser supports the acquisition of access tokens for accounts across multiple tenants. In order to achieve this, MSAL caches the account information and auth artifacts for a user for each tenant it has authenticated with. 
+
+This means that with the exception of refresh tokens, which are shared across tenants, the MSAL cache may store multiple accounts, access tokens, and ID tokens for a single user. Each of these account records, in a multi-tenant context, will be referred to as a **"tenant profile"** of the main account.
```
sameerag, line 9:
```suggestion
With the exception of refresh tokens, which are shared across tenants, the MSAL cache will now store accounts, access tokens, and ID tokens `per tenant` for a single user. Each of these account records, in a multi-tenant context, will be referred to as a **"tenant profile"** of the main account.
```

lib/msal-browser/docs/multi-tenant-accounts.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Tenant Profiles
+
+A tenant profile is the record of a particular account in a particular tenant. When multi-tenant support is enabled, a distinct `AccountInfo` object will be cached containing the entire tenant profile for every tenant that the user authenticates with.
```
sameerag, line 13:
```suggestion
A tenant profile is the record of an account in a particular tenant. When multi-tenant support is enabled, a distinct `AccountInfo` object will be cached containing the entire tenant profile for every tenant that the user authenticates with.
```

sameerag, line 13: What does "containing entire tenant profile" mean? Can we expand that - is it that `tenant_info` or all `tenants`?

chris-oliver-zartis:
Why was this closed? This actually looks like it would be a useful feature.

sameerag:
We have this feature integrated with 3.x, albeit with some changes in design. @chris-oliver-zartis.

chris-oliver-zartis:
Thanks @sameerag, do you perhaps have documentation covering how this is implemented?

PR #6397, opened 2023-08-25
Modified refreshOn in MSAL Common Silent Flow to return cached token before refreshing in the background

Robbie-Microsoft:
refreshOn functionality already exists in Silent Flow. That is, if a token needs to be refreshed but is not expired, an error was thrown and the token was refreshed. This PR contains changes that will now return the cached token if it needs to be refreshed and is not expired, while also refreshing the token in the background.


lib/msal-node/test/client/ClientCredentialClient.spec.ts
```diff
@@ -748,9 +748,10 @@ describe("ClientCredentialClient unit tests", () => {
             clientCredentialRequest
         )) as AuthenticationResult;
 
-        // Check the cache to ensure the refreshed token exists (the network request was successful).
-        // Typically, the network request may not have completed by the time the below code runs.
-        // However, the network requests are mocked in these tests, so the refreshed token should be in the cache at this point.
+        // Wait two seconds for the acquireToken and its mocked network requests to complete,
+        // then check the cache to ensure the refreshed token exists (the network request was successful)
+        await new Promise((r) => setTimeout(r, 2000));
```
bgavrilMS: 2s for a unit test is quite a lot and generally waiting in tests is a code smell. 

A better pattern to check that an internal async operation has finished is: 


1. check if some field X has been modified / condition is satisfied (I guess in your case the condition at line 755)
2. if it hasn't, wait for a short amount of time, e.g. 100ms
3. goto 1

Of course, if the whole thing takes too long, e.g. 2000ms, then bail and fail the test.

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -308,7 +308,12 @@ export class RefreshTokenClient extends BaseClient {
             acquireTokenMeasurement?.end({
                 success: true,
             });
-            return result;
+
+            if (!response) {
```
bgavrilMS: How can this happen? What is the error condition that the user / app dev is facing and how can they fix it? Errors should be actionable.

bgavrilMS (changes requested):
No sleep in tests; Errors must be actionable.

lib/msal-common/src/client/SilentFlowClient.ts
```diff
@@ -39,8 +41,34 @@ export class SilentFlowClient extends BaseClient {
         request: CommonSilentFlowRequest
     ): Promise<AuthenticationResult> {
         try {
-            return await this.acquireCachedToken(request);
+            const cachedToken = await this.acquireCachedToken(request);
```
pmaytak: Does JS/TS have a concept of tuples? Like in C# you can return multiple values: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/value-tuples#use-cases-of-tuples

Would be helpful, so that lastCacheOutcome wouldn't have to be set/reset.

lib/msal-common/src/client/SilentFlowClient.ts
```diff
@@ -70,6 +98,7 @@ export class SilentFlowClient extends BaseClient {
 
         if (request.forceRefresh) {
             // Must refresh due to present force_refresh flag.
+            this.lastCacheOutcome = CacheOutcome.FORCE_REFRESH;
```
pmaytak: Line 110 should also set this for consistency.

Not related to this, but 110 should also set telemetry to ForceRefresh? 

In .NET (and I think per spec) Claims and Force refresh are counted as the same. 
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/90ae6b9c6d33655a66cbc6413e8aadee487f052a/src/client/Microsoft.Identity.Client/Cache/CacheRefreshReason.cs#L18

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -70,10 +70,13 @@ export class RefreshTokenClient extends BaseClient {
...
-            request,
-            this.authority
-        );
+
+        let response;
+        try {
+            response = await this.executeTokenRequest(request, this.authority);
+        } catch (error) {
+            throw error;
+        }
```
pmaytak: Remove? I assume this was for debugging?

pmaytak (approved):

lib/msal-common/src/client/RefreshTokenClient.ts
Robbie-Microsoft: Reverted these changes. I was messing around in this file while testing the unit tests.

Robbie-Microsoft: It will never happen; I made a mistake leaving that code in this PR. There will always either be a response, or a thrown exception.

lib/msal-common/src/client/SilentFlowClient.ts
Robbie-Microsoft: Where can I find the spec for confirmation? However, good point. Maybe a more specific error should be created for the claims check instead of throwing a createRefreshRequiredError? Thoughts?

lib/msal-common/src/client/SilentFlowClient.ts
Robbie-Microsoft: There is concept of [typescript tuples](https://www.w3schools.com/typescript/typescript_tuples.php). I can see how this would make sense, however, all the unit tests using acquireCachedToken would need to updated to account for the tuple return type. I think that would be fine. Thoughts?

bgavrilMS: Yes, do it https://levelup.gitconnected.com/the-boy-scouting-rule-a-simple-yet-effective-programming-principle-5d504c7dd993#:~:text=The%20Boy%20Scouting%20Rule%20is,environment%20cleaner%20than%20they%20found.

lib/msal-common/src/client/SilentFlowClient.ts
bgavrilMS: .NET implementation and Peter's advice are the best spec :)

Better error messages for the app developers using the SDK is always good. Error messages are not part of tghe public api, so we're not breaking them. But error codes are.

lib/msal-common/src/client/SilentFlowClient.ts
pmaytak: The issue with using lastCacheOutcome like currently is if one forgets to set/reset it, it can create bugs, which can be difficult to track down. Since Bogdan gave you the blessing on refactoring, I would just put it in a separate PR, you can use this branch as the base and then merge right after this one. (Smaller and more focused PRs are easier to review and track.)

Hmm, I knew about the rule; didn't know what it was called...

bgavrilMS (approved):

lib/msal-common/src/client/SilentFlowClient.ts
sameerag: Just make sure the `acquireToken` call is handled properly as I think we use that API in `msal-browser`. Also may be `cachedToken` is not named correctly, the return type is `AuthenticationResult` and not just a `token`.

lib/msal-common/src/client/SilentFlowClient.ts
```diff
@@ -40,8 +42,34 @@ export class SilentFlowClient extends BaseClient {
...
+                this.logger.info(
+                    "SilentFlowClient:acquireCachedToken - Cached access token's refreshOn property has been exceeded'. It's not expired, but must be refreshed."
+                );
+
+                // refresh the access token in the background
+                const refreshTokenClient = new RefreshTokenClient(
+                    this.config,
+                    this.performanceClient
+                );
+                refreshTokenClient.acquireTokenByRefreshToken(request);
```
sameerag: Please mark this appropriately, we have added a linter rule to mark floating promises.

lib/msal-common/src/client/SilentFlowClient.ts
```diff
@@ -40,8 +42,34 @@ export class SilentFlowClient extends BaseClient {
...
+            }
+
+            // reset the last cache outcome
+            this.lastCacheOutcome = CacheOutcome.NO_CACHE_HIT;
```
konstantin-msft: Can place this repeated line into `finally`.

konstantin-msft (approved):

sameerag (approved):

lib/msal-common/src/client/SilentFlowClient.ts
Robbie-Microsoft: yea, I saw that... Since this is supposed to run in the background while the cached token is returned, can't I just add an ignore statement right above it? We're not taking any action upon the success or failure of this code.

pmaytak (approved):

PR #6406, opened 2023-08-29
Create shareable eslint config

tnorling:
Creates a private package which contains the eslint config and dependencies. 

This allows each lib to run linting within its own directory instead of running from the root - which will become more important when we move the CI pipelines to the 1P repo


konstantin-msft (approved):

eslint-config-msal/package.json
```diff
@@ -0,0 +1,19 @@
+{
+    "name": "eslint-config-msal",
+    "version": "0.0.0",
```
peterzenz, line 3: 0.0.0?

peterzenz (approved):

eslint-config-msal/package.json
tnorling: We can give this any arbitrary number we want as it won't be published and won't ever be incremented. It's only defined because it's a required field.

PR #6412, opened 2023-08-30
Shared jest config

tnorling:
Creates a shared jest configuration with the necessary crypto polyfills


shared-configs/jest-config/jest.config.cjs
```diff
@@ -0,0 +1,17 @@
...
+
+module.exports = {
+    preset: "ts-jest",
+    testEnvironment: "jsdom",
+    testEnvironmentOptions: {
+        url: "https://localhost:8081/index.html"
+    },
+    collectCoverageFrom: ["src/**/*.ts"],
+    coverageReporters: [["lcov", { "projectRoot": path.join(__dirname, "../../") }]],
+    setupFilesAfterEnv: [path.join(__dirname, "setupCrypto.cjs")],
```
sameerag, line 16: Curious: Was this optional before on `common` and `node-extensions`?

sameerag (approved):

shared-configs/jest-config/jest.config.cjs
tnorling: Crypto APIs aren't used in those libs today so it hasn't been needed

peterzenz (approved):

PR #6413, opened 2023-08-30
Remove redundant controller internal methods and classes from internals

konstantin-msft:
- Remove redundant controller internal methods.
- Remove redundant classes from internals.ts.
- Add `MemoryStorage` and `BrowserStorage` to internals.ts. 
- Update tests.

**Note**
Will push a follow-up 1p PR right after to reconcile changes.


lib/msal-browser/src/controllers/StandardController.ts
```diff
@@ -1713,17 +1713,10 @@ export class StandardController implements IController {
...
-    public getNativeInternalStorage(): BrowserCacheManager {
-        return this.nativeInternalStorage;
-    }
-
-    /**
-     * Returns the instance of interface for crypto functions
-     */
-    public getBrowserCrypto(): ICrypto {
-        return this.browserCrypto;
+    public setInteractionInProgress(progress: boolean): void {
```
konstantin-msft: Public access for 1p.

peterzenz (approved):

lib/msal-browser/src/controllers/IController.ts
```diff
@@ -123,54 +118,17 @@ export interface IController {
...
     getEventHandler(): EventHandler;
 
-    /** @internal */
-    getNavigationClient(): INavigationClient;
-
-    /** @internal */
-    getRedirectResponse(): Map<string, Promise<AuthenticationResult | null>>;
-
     /** @internal */
     preflightBrowserEnvironmentCheck(
```
tnorling, line 127: What about the rest of these internal functions? Future PR?

lib/msal-browser/src/controllers/StandardController.ts
tnorling: 1P shouldn't need to set this, if it does we should see what we need to do to make it not needed

lib/msal-browser/src/internals.ts
```diff
@@ -8,36 +8,18 @@
...
-export { CacheRecord } from "@azure/msal-common";
+// Storage
+export { MemoryStorage } from "./cache/MemoryStorage";
+export { BrowserStorage } from "./cache/BrowserStorage";
 
 // Clients
-export { StandardInteractionClient } from "./interaction_client/StandardInteractionClient";
-export { RedirectClient } from "./interaction_client/RedirectClient";
-export { PopupClient } from "./interaction_client/PopupClient";
 export { SilentIframeClient } from "./interaction_client/SilentIframeClient";
```
tnorling, line 12: Are the rest of these exports still being used in 1P?

lib/msal-browser/src/controllers/IController.ts
sameerag: I think this is only for the current changes already made in 1P?

lib/msal-browser/src/controllers/StandardController.ts
sameerag: I would like that in a separate PR though. @konstantin-msft  please correct me but my assumption is that this is reflecting current 1P status?

sameerag (approved):

lib/msal-browser/src/controllers/StandardController.ts
tnorling: No, but this is _adding_ new functionality to 3P. We shouldn't add things we don't need and since this is public we won't be able to remove it later

tnorling (changes requested):
Please remove new API

lib/msal-browser/src/internals.ts
```diff
@@ -8,36 +8,18 @@
  * Breaking changes to these APIs will be shipped under a minor version, instead of a major version.
  */
 
-// Cache Manager
-export { BrowserCacheManager } from "./cache/BrowserCacheManager";
-
-// Cache entities
-export { CacheRecord } from "@azure/msal-common";
+// Storage
+export { MemoryStorage } from "./cache/MemoryStorage";
```
tnorling: Why are we adding new things to internals? Goal is to remove internals entirely

lib/msal-browser/src/controllers/IController.ts
konstantin-msft: The rest are still used in 1P.

lib/msal-browser/src/internals.ts
konstantin-msft: Yes, other exports are still used being used in 1P.

lib/msal-browser/src/internals.ts
konstantin-msft: We need that for temporary broker storage in 1p.

tnorling: Then let's export this in the main exports

konstantin-msft:
> Please remove new API

New API will be used in 1P. Instead of exposing `BrowserStorage` that is solely used for managing the interaction status we can create an API to so directly.

tnorling:
> > Please remove new API
> 
> New API will be used in 1P. Instead of exposing `BrowserStorage` that is solely used for managing the interaction status we can create an API to so directly.

My question is - why do we need to manage interaction status in 1P at all? Those types of details should be managed solely by 3P and if 1P is taking a requirement on these details something is wrong with the design

konstantin-msft:
> 

Removed the API.

lib/msal-browser/src/controllers/StandardController.ts
konstantin-msft: Removed the API after brief discussion.

lib/msal-browser/src/controllers/IController.ts
tnorling: I just want to confirm removing the rest of these is still on our radar, just isn't ready yet?

lib/msal-browser/src/controllers/StandardController.ts
```diff
@@ -524,7 +524,7 @@ export class StandardController implements IController {
                         );
                         return redirectClient.acquireToken(request);
                     }
-                    this.browserStorage.setInteractionInProgress(false);
+                    this.getBrowserStorage().setInteractionInProgress(false);
```
tnorling, line 527: Reason for this change?

tnorling (approved):

PR #6414, opened 2023-08-31
Refactor BrowserAuthError for reduced size

tnorling:
- Removes static `create<some>Error` helpers in favor of a single `createBrowserAuthError` which accepts an error code (method not belonging to a class allows its name to be minified)
- Exports error codes as constants (allows bundlers to minify references to the const)
- `BrowserAuthErrorMessage` uses new constants to avoid duplication, marked deprecated and no longer used internally (allows bundlers to treeshake this out if the app isn't importing it)

This allows for better minification and reduces minified size by ~10kb. Other error classes to follow.


lib/msal-browser/src/error/BrowserAuthErrorCodes.ts
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+export const pkceNotGenerated = "pkce_not_created";
```
sameerag: Will all these minify into the array they are referred in above? I thought minification cannot work when the `constants` scope cannot be defined. Since the usage is spread across the library, was curious how this helps.

tnorling: It's a bit complex, I'll try to break it down (I also plan to do a more in depth demonstration for the team at a later date)

Each of these will be minified down to something like `a="pkce_not_created"` and any internal reference to this constant will instead reference `a`.

It also means that any unused error codes can be removed because they are all individual constants rather than being lumped together in some shared object.

We also ` export * as BrowserAuthErrorCodes` so that consumers of the library can do something like `BrowserAuthErrorCodes.pkceNotGenerated`. This gets included in our output bundle separately as something along the lines of `BrowserAuthErrorCodes={pkceNotGenerated: a, ...etc}`

The final bundler (application) can then map references to `BrowserAuthErrorCodes.pkceNotGenerated` in the app back to `a` and remove the `BrowserAuthErrorCodes` object from the final bundle.

lib/msal-browser/src/error/BrowserAuthErrorCodes.ts
```diff
@@ -0,0 +1,57 @@
...
+export const cryptoDoesNotExist = "crypto_nonexistent";
+export const emptyNavigateUriError = "empty_navigate_uri";
+export const hashEmptyError = "hash_empty_error";
+export const hashDoesNotContainStateError = "no_state_in_hash";
```
peterzenz: suggest matching const names to the text:
`export const noStateInHash = "no_state_in_hash"`
There are quite a few of these in the list.  We likely can't alter things like "no_state_in_hash" as folks might worry if there's a change in error characteristics from one release to another.

peterzenz (approved):

lib/msal-browser/src/error/BrowserAuthErrorCodes.ts
tnorling: Great suggestion thanks! Done.

sameerag (approved):

lib/msal-browser/src/error/BrowserAuthErrorCodes.ts
sameerag: Thanks. Good improvement over errors and probably adds to better docs.

konstantin-msft (approved):

lib/msal-browser/src/error/BrowserAuthError.ts
```diff
@@ -4,672 +4,337 @@
...
 };
 
 /**
  * Browser library error class thrown by the MSAL.js library for SPAs
  */
 export class BrowserAuthError extends AuthError {
-    constructor(errorCode: string, errorMessage?: string) {
-        super(errorCode, errorMessage);
+    constructor(errorCode: string) {
+        super(errorCode, BrowserAuthErrorMessages[errorCode]);
```
JasonGore, line 331: @tnorling I have some test utilities in our app to make sure we're receiving and handling MSAL errors in a consistent manner. Should the second arg to `super` be `BrowserAuthErrorMessages[errorCode].desc` here instead? I noticed this when `BrowserAuthError` started emitting differently than the other errors after being converted to string. 

image

JasonGore: Sorry, overlooked the `s` at the end of `BrowserAuthErrorMessages`

PR #6416, opened 2023-08-31
Declass encoder/decoder for reduced size

tnorling:
- Moves Base64Encoder and Base64Decoder into functions rather than class methods
- Updates their implementations to use built-in browser APIs (now available since we don't support IE anymore!)


change/@azure-msal-browser-a12bca4f-39ee-49e6-bd21-5e6af072d408.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Make base64 encoder and decoder methods instead of classes",
```
sameerag:
```suggestion
  "comment": "Make base64 encoder and decoder methods instead of classes #6416",
```

lib/msal-browser/src/cache/BrowserCacheManager.ts
```diff
@@ -1696,10 +1690,10 @@ export class BrowserCacheManager extends CacheManager {
...
         }
 
-        const parsedRequest = this.validateAndParseJson(
-            browserCrypto.base64Decode(encodedTokenRequest)
-        ) as CommonAuthorizationCodeRequest;
-        if (!parsedRequest) {
+        let parsedRequest: CommonAuthorizationCodeRequest;
+        try {
+            parsedRequest = JSON.parse(base64Decode(encodedTokenRequest));
+        } catch (e) {
```
peterzenz, line 1696: Include the caught exception inside the rethrow?

peterzenz (approved):

lalimasharda (approved):

lib/msal-browser/src/cache/BrowserCacheManager.ts
tnorling: Added a couple log messages instead.

PR #6457, opened 2023-09-11
Do not add v2.0 to authority endpoint if already exists

### LINKED ISSUE #6456: Regression in metadata discovery after upgrading msal-common v13.1.0 to v14.0.2
### Core Library

MSAL.js (@azure/msal-browser)

### Core Library Version

3.0.2

### Wrapper Library

### Wrapper Library Version

3.0.2

Public

### Description

After upgrading the @azure/msal-common library the metadata discovery goes into an infinite loop.

### Error Message

GET https://login.microsoftonline.com/<redacted tenant id>/v2.0/v2.0/.well-known/openid-configuration net::ERR_FAILED 404 (Not Found)

### Msal Logs

_No response_

### MSAL Configuration

```javascript
{
    auth: {
        authority: 'https://login.microsoftonline.com/<redacted tenant id>/v2.0',
        protocolMode: ProtocolMode.OIDC,
    },
}
```

### Relevant Code Snippets

```javascript
function MsalInstanceFactory(configuration): IPublicClientApplication {
    return new PublicClientApplication(configuration);
}
```

### Reproduction Steps

1. Initialize the PublicClientApplication with the provided authority
2. Request a login
3. Metadata discovery fails because the discovery endpoint has duplicate `v2.0` in URL.

### Expected Behavior

The discovery URL is correct (as was the behaviour with msal-common v13.1.0).

### Identity Provider

Azure AD / MSA

Chrome, Firefox

### Regression

@azure/msal-common v13.1.0

### Source

External (Customer)

konstantin-msft:
- Do not add v2.0 to authority endpoint if already exists.

### DIFF
--- a/change/@azure-msal-common-ddf6cb76-26e2-40de-b764-8b2c753ad77a.json
+++ b/change/@azure-msal-common-ddf6cb76-26e2-40de-b764-8b2c753ad77a.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Do not add v2.0 to authority endpoint if already exists #6457",
+  "packageName": "@azure/msal-common",
+  "email": "kshabelko@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/lib/msal-common/src/authority/Authority.ts
+++ b/lib/msal-common/src/authority/Authority.ts
@@ -356,6 +356,7 @@ export class Authority {
     protected get defaultOpenIdConfigurationEndpoint(): string {
         const canonicalAuthorityHost = this.hostnameAndPort;
         if (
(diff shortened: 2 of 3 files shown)


tnorling (approved):

hectormmg (approved):

sameerag (approved):
Usage docs mention to always pass only `domainname/tenantId` format. Curious about the usage here, it is good to have but recommend the proper usage as much as possible.

hectormmg:
> Usage docs mention to always pass only `domainname/tenantId` format. Curious about the usage here, it is good to have but recommend the proper usage as much as possible.

I agree that the developer shouldn't be setting the API version, but I also prefer we gracefully avoid duplication rather than throw if the7 set v2.0/

PR #6465, opened 2023-09-12
Created new tuple return type for SilentFlow acquireCachedToken

Robbie-Microsoft:


lib/msal-common/src/client/SilentFlowClient.ts
```diff
@@ -42,13 +40,12 @@ export class SilentFlowClient extends BaseClient {
         request: CommonSilentFlowRequest
     ): Promise<AuthenticationResult> {
         try {
-            const cachedToken = await this.acquireCachedToken(request);
+            const response = await this.acquireCachedToken(request);
+            const authResponse = response[0];
+            const cacheOutcome = response[1];
```
Robbie-Microsoft: const [authResponse, cacheOutcome] = await this.acquireCachedToken(request);

lib/msal-common/src/client/SilentFlowClient.ts
```diff
@@ -43,13 +41,12 @@ export class SilentFlowClient extends BaseClient {
         request: CommonSilentFlowRequest
     ): Promise<AuthenticationResult> {
         try {
-            const cachedToken = await this.acquireCachedToken(request);
+            const [authResponse, cacheOutcome] = await this.acquireCachedToken(
+                request
+            );
```
pmaytak, line 46:
```suggestion
            const [authResponse, cacheOutcome] = await this.acquireCachedToken(request);
```

pmaytak (approved):

lib/msal-common/src/client/SilentFlowClient.ts
Robbie-Microsoft: I agree with you, but we have some crazy linting rules now that auto-fix the code to look like this.

PR #6469, opened 2023-09-13
Improve robustness of LoopbackClient

xirzec:
This change addresses a few rough edges caused by Node's `http.server` APIs and how they behave in unexpected ways as well as some cleanup of the `ILoopbackClient` contract:

1. `server.close()` only prevents new connections from being accepted but does not clean up any open ones.
2. A race condition where we call `server.address()` potentially before the service has begun listening.
3. General cleanup like listening for the `listening` and `error` events on `server` instead of polling in a setInterval and avoiding mixing `async/await` with Promise constructors and chaining.
5. A small enhancement where we now clean up the server as soon as we have the final auth response or have failed in case the caller forgets to call `closeServer`

Unfortunately, I wasn't able to make this change without introducing some small breaks to the `ILoopbackClient`. This is because I needed to split apart waiting for the server to start listening and waiting for the auth code to come back, otherwise starting the redirect dance in the middle could run afoul of race conditions.

Any feedback, stylistic or otherwise would be greatly appreciated!

### DIFF
--- a/lib/msal-node/src/network/ILoopbackClient.ts
+++ b/lib/msal-node/src/network/ILoopbackClient.ts
@@ -10,10 +10,11 @@ import { ServerAuthorizationCodeResponse } from "@azure/msal-common";
  * @public
  */
 export interface ILoopbackClient {
-    listenForAuthCode(
-        successTemplate?: string,
-        errorTemplate?: string
-    ): Promise<ServerAuthorizationCodeResponse>;
+    startServer(options: {
+        successTemplate?: string;
+        errorTemplate?: string;
+    }): Promise<void>;
+    waitForAuthCode(): Promise<ServerAuthorizationCodeResponse>;
     getRedirectUri(): string;
     closeServer(): void;
 }
--- a/lib/msal-node/src/client/PublicClientApplication.ts
+++ b/lib/msal-node/src/client/PublicClientApplication.ts
@@ -167,12 +167,11 @@ export class PublicClientApplication
         const loopbackClient: ILoopbackClient =
             customLoopbackClient || new LoopbackClient();
 
-        let authCodeListener: Promise<ServerAuthorizationCodeResponse>;
         try {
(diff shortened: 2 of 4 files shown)


lib/msal-node/src/network/ILoopbackClient.ts
```diff
@@ -10,10 +10,11 @@ import { ServerAuthorizationCodeResponse } from "@azure/msal-common";
...
 export interface ILoopbackClient {
-    listenForAuthCode(
-        successTemplate?: string,
-        errorTemplate?: string
-    ): Promise<ServerAuthorizationCodeResponse>;
+    startServer(options: {
+        successTemplate?: string;
+        errorTemplate?: string;
+    }): Promise<void>;
+    waitForAuthCode(): Promise<ServerAuthorizationCodeResponse>;
```
tnorling, line 17: This is a breaking change we cannot accept.

xirzec:
@tnorling per your request, I have created two standalone repros that showcase the major failings of the current design: https://github.com/xirzec/msal-repro

Let me know if this is sufficient for your investigation or if you would like me to walk you through why `msal-node` is doing the wrong thing in each case.

tnorling:
> @tnorling per your request, I have created two standalone repros that showcase the major failings of the current design: https://github.com/xirzec/msal-repro
> 
> Let me know if this is sufficient for your investigation or if you would like me to walk you through why `msal-node` is doing the wrong thing in each case.

I appreciate it, thanks! I'll take a look tomorrow or Friday.

xirzec:
Closing in favor of https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/6481

xirzec:
@tnorling @peterzenz any suggestions how to tell the bot to not reopen? 😅

tnorling:
It's a bug with the "close and comment" action. Commenting then closing should keep it closed. We have it on our backlog to find an alternative but haven't gotten around to it yet.

PR #6480, opened 2023-09-14
Add missing queue measurement instrumentation

konstantin-msft:
- Add missing queue measurement instrumentation.
- Default queue time to 0 for manually completed queue events.
- Update unit tests.

### DIFF
--- a/lib/msal-browser/src/cache/BrowserCacheManager.ts
+++ b/lib/msal-browser/src/cache/BrowserCacheManager.ts
@@ -77,8 +77,6 @@ export class BrowserCacheManager extends CacheManager {
     protected temporaryCacheStorage: IWindowStorage<string>;
     // Logger instance
     protected logger: Logger;
-    // Performance client
-    protected readonly performanceClient: IPerformanceClient;
 
     // Cookie life calculation (hours * minutes * seconds * ms)
     protected readonly COOKIE_LIFE_MULTIPLIER = 24 * 60 * 60 * 1000;
@@ -87,8 +85,7 @@ export class BrowserCacheManager extends CacheManager {
         clientId: string,
         cacheConfig: Required<CacheOptions>,
         cryptoImpl: ICrypto,
-        logger: Logger,
-        performanceClient: IPerformanceClient
+        logger: Logger
     ) {
         super(clientId, cryptoImpl, logger);
         this.cacheConfig = cacheConfig;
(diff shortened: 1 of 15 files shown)


peterzenz (approved):

sameerag (commented):
Is there any way we can identify when `addQueueMeasurement` is not called when it is supposed to?

lib/msal-browser/src/cache/BrowserCacheManager.ts
```diff
@@ -96,6 +101,7 @@ export class BrowserCacheManager extends CacheManager {
             this.cacheConfig.temporaryCacheLocation,
             this.cacheConfig.cacheLocation
         );
+        this.performanceClient = performanceClient;
```
tnorling: Can we pass this into the function rather than adding it to the instance? It will make de-classing this easier later.

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
```diff
@@ -227,7 +227,8 @@ export abstract class PerformanceClient implements IPerformanceClient {
 
         const queueMeasurement: QueueMeasurement = {
             eventName,
-            queueTime,
+            // Always default queue time to 0 for manually completed (improperly instrumented)
+            queueTime: manuallyCompleted ? 0 : queueTime,
```
tnorling, line 231: Just a thought, not sure how much I like this idea myself, but what if we make it -1 so we can easily identify bugs versus a legitimate 0?

tnorling: We would then also have to throw out -1s when summing up all the queue times

sameerag: I would prefer -1 than 0.

sameerag: What is the measurement unit: milli seconds?

konstantin-msft: We already have `manuallyCompleted` attribute in place which then get accumulated and tracked. There is no need to set it to `-1`.

konstantin-msft:
> Is there any way we can identify when `addQueueMeasurement` is not called when it is supposed to?

We can do that by checking the number of manually completed queue events, which is, unfortunately, non-proactive.

lib/msal-browser/src/cache/BrowserCacheManager.ts
konstantin-msft: done

sameerag (approved):

tnorling (approved):

PR #6484, opened 2023-09-15
Update measurements to use invoke

tnorling:
[msal-browser] Updates remaining measurements to use the invoke wrappers. This will also result in a few more measurements we're not collecting today.

Will follow up with msal-common changes in a separate PR

### DIFF
--- a/change/@azure-msal-browser-4b59ec24-18be-486e-8bb9-33d5262c4a0f.json
+++ b/change/@azure-msal-browser-4b59ec24-18be-486e-8bb9-33d5262c4a0f.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Update telemetry measurements to use invoke",
+  "comment": "Update telemetry measurements to use invoke #6484",
   "packageName": "@azure/msal-browser",
   "email": "thomas.norling@microsoft.com",
   "dependentChangeType": "patch"
--- a/change/@azure-msal-browser-4b59ec24-18be-486e-8bb9-33d5262c4a0f.json
+++ b/change/@azure-msal-browser-4b59ec24-18be-486e-8bb9-33d5262c4a0f.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Update telemetry measurements to use invoke",
+  "packageName": "@azure/msal-browser",
+  "email": "thomas.norling@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/lib/msal-browser/src/controllers/StandardController.ts
+++ b/lib/msal-browser/src/controllers/StandardController.ts
@@ -1042,7 +1042,13 @@ export class StandardController implements IController {
             case CacheLookupPolicy.Default:
             case CacheLookupPolicy.AccessToken:
             case CacheLookupPolicy.AccessTokenAndRefreshToken:
-                return silentCacheClient.acquireToken(commonRequest);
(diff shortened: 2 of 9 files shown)


lib/msal-browser/src/controllers/StandardController.ts
```diff
@@ -2010,72 +2020,75 @@ export class StandardController implements IController {
...
+                        this.logger,
+                        this.performanceClient,
                         silentRequest.correlationId
+                    )(silentRequest, requestWithCLP).catch(
+                        (refreshTokenError: AuthError) => {
+                            const isServerError =
+                                refreshTokenError instanceof ServerError;
+                            const isInteractionRequiredError =
+                                refreshTokenError instanceof
+                                InteractionRequiredAuthError;
```
peterzenz, line 2058: nit:  Odd line break here, maybe remove the newline or indent

lib/msal-browser/src/controllers/StandardController.ts
```diff
@@ -2010,72 +2020,75 @@ export class StandardController implements IController {
...
+                            const isInvalidGrantError =
+                                refreshTokenError.errorCode ===
+                                BrowserConstants.INVALID_GRANT_ERROR;
+
+                            if (
+                                (!isServerError ||
+                                    !isInvalidGrantError ||
+                                    isInteractionRequiredError ||
```
peterzenz, line 2066: Can you list the or clauses in the order of most common to least in this expression?  I'm guessing interaction required would be before invalid grant but I could be mistaken.

peterzenz (approved):

change/@azure-msal-browser-4b59ec24-18be-486e-8bb9-33d5262c4a0f.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Update telemetry measurements to use invoke",
```
sameerag:
```suggestion
  "comment": "Update telemetry measurements to use invoke #6484",
```

lib/msal-browser/src/controllers/StandardController.ts
tnorling: This is the formatter's doing, agree it's a bit odd. We'd have to change the default rules to change the behavior here

lib/msal-browser/src/controllers/StandardController.ts
tnorling: Do we have data on this? I'm not actually changing anything here, it's just showing in the diff due to new indentation

tnorling: Invalid grant will be every 24 hours for sure, not sure if interaction_required should be expected to be more or less often than that

sameerag: interaction_required can show for background changes like MFA or CA changes apart from the token expiry.  If we take off the policy or password changes which cannot be predicted, invalid_grant still precedes interaction_required as the later is invoked when the session expires and the former is when the RT expires (predictably 24 hrs).

lib/msal-browser/src/controllers/StandardController.ts
sameerag: May be increasing line length limits will do it, can be a separate PR.

sameerag (approved):

PR #6499, opened 2023-09-20
Add getAccount and enhance account filtering

hectormmg:
This PR:
- Adds and `accountFilter: AccountFilter` param to `getAllAccounts()` in order to enable the narrowing-down of cached accounts returned
- Adds `getAccountByFilter` API to replace now marked-for-deprecation `getAccountBy` APIs
- Extends `AccountFilter` class to optionally include all the properties in `AccountInfo` except `idToken` and `idTokenClaims` so accounts are filterable by all of their searchable properties
- Adds new `AccountFilter` properties to the set of matchers in `CacheManager.getAccountsFilteredBy()` as well as their respective private matching methods
- Adds filtering by `loginHint`


lib/msal-browser/docs/accounts.md
```diff
@@ -6,10 +6,32 @@
...
+-   `getAccountByFilter()`: returns the first cached account that matches the filter passed in. The order in which accounts are read from the cache is arbitrary and there is no guarantee that the first account in the filtered list will be the same for any two calls of `getAccountByFilter`.
+
+### Account Filter Object
+
+The following table presents the properties that can be used in the `AccountFilter` object that can be passed into the APIs listed above.
+
+> Note: A single account filter attribute is usually not guaranteed to uniquely identify a cached account object. Adding a combination of attributes that don't repeat together, such as `homeAccountId` + `localAccountId`, can help refine the search.
+
+| Property Name        | Description                                                                                                                                                                                  | Example                                                                   |
```
tnorling: Let's reference the typedocs here instead of writing it all out - less to maintain later :)

lib/msal-browser/docs/accounts.md
```diff
@@ -35,25 +57,79 @@ function handleResponse(resp) {
...
-       // Handle error
+        // Handle error
         return await myMSALObj.acquireTokenPopup(request);
     });
 }
 ```
 
+### Filtering by login hint
+
+As of `@azure/msal-browser@3.11.0`, all login hint values can be used to search for and filter accounts. In order to filter by login hint, MSAL will compare the `loginHint` value in the `AccountFilter` object against the following account attributes (in order of precedence) to search for matches:
```
tnorling: 3.11?

lib/msal-browser/docs/accounts.md
```diff
@@ -35,25 +57,79 @@ function handleResponse(resp) {
...
+    return await myMSALObj.acquireTokenPopup(request);
+});
+```
+
+#### Using `username``
+
+```javascript
+const accountUsername = userProfile.username;
+const accountFilter = {
+    loginHint: accountUsername;
```
tnorling, line 99: Based on the AccountFilter type above, could you also provide this as `username`?

lib/msal-browser/docs/accounts.md
```diff
@@ -6,10 +6,32 @@
 
 The `@azure/msal-browser` library provides the following APIs to access cached accounts:
 
-* `getAllAccounts()`: returns all the accounts currently in the cache. An application must choose an account to acquire tokens silently.
-* `getAccountByHomeId()`: receives a `homeAccountId` string and returns the matching account from the cache.
-* `getAccountByLocalId()`: receives a `localAccountId` string and returns the matching account from the cache.
-* `getAccountByUsername()`: receives a `username` string and returns the matching account from the cache.
+-   `getAllAccounts()`: returns all the accounts currently in the cache. A filter can be passed in to narrow down the returned accounts. An application must choose an account to acquire tokens silently.
```
sameerag:
```suggestion
-   `getAllAccounts()`: returns all the accounts currently in the cache. Supports a filter param to return a specific set of accounts. An application must choose an account to acquire tokens silently.
```

lib/msal-browser/docs/accounts.md
```diff
@@ -6,10 +6,32 @@
...
+| `localAccountId`     | Local, tenant-specific account identifer for this account object, usually used in legacy cases                                                                                               | 00000000-0000-0000-0000-000000000000                                      |
+| `name`               | Full name for the account, including given name and family name.                                                                                                                             | John Doe                                                                  |
+| `nativeAccountId`    | The user's native account ID                                                                                                                                                                 | 00000000-0000-0000-0000-000000000000                                      |
+
+> Note: `realm` is the name that `tenantId` receives in the cache.
```
sameerag: nit: 
```suggestion
> Note: `realm` is the name that `tenantId` received in the cache.
```

lib/msal-browser/docs/accounts.md
hectormmg: The next release, this will be a minor change so the 3.11 is going to be the next version.

lib/msal-browser/docs/accounts.md
```diff
@@ -6,10 +6,32 @@
 
 The `@azure/msal-browser` library provides the following APIs to access cached accounts:
 
-* `getAllAccounts()`: returns all the accounts currently in the cache. An application must choose an account to acquire tokens silently.
-* `getAccountByHomeId()`: receives a `homeAccountId` string and returns the matching account from the cache.
-* `getAccountByLocalId()`: receives a `localAccountId` string and returns the matching account from the cache.
-* `getAccountByUsername()`: receives a `username` string and returns the matching account from the cache.
+-   `getAllAccounts()`: returns all the accounts currently in the cache. A filter can be passed in to narrow down the returned accounts. An application must choose an account to acquire tokens silently.
+-   `getAccountByFilter()`: returns the first cached account that matches the filter passed in. The order in which accounts are read from the cache is arbitrary and there is no guarantee that the first account in the filtered list will be the same for any two calls of `getAccountByFilter`.
```
sameerag:
```suggestion
-   `getAccountByFilter()`: returns the first cached account that matches the filter passed in. The order in which accounts are read from the cache is arbitrary and there is no guarantee that the first account in the filtered list will be the same for any two calls of `getAccountByFilter`. Please check below for guidance to make a unique match possible.
```

lib/msal-browser/docs/accounts.md
hectormmg: Yes you could, that was already enabled though. Considering it in login hint is mostly useful if you pick a login_hint out of whichever is available between the three and then stop caring what the source of the login_hint was (in case you lose context). So 10k lines above const loginHint was assigned to something, but you know for sure you can pass it into getAccount and it'll get compared to all the options available.

lib/msal-browser/docs/accounts.md
hectormmg: I knew this was coming 😆

lib/msal-browser/docs/accounts.md
```diff
@@ -6,10 +6,32 @@
...
+| `localAccountId`     | Local, tenant-specific account identifer for this account object, usually used in legacy cases                                                                                               | 00000000-0000-0000-0000-000000000000                                      |
+| `name`               | Full name for the account, including given name and family name.                                                                                                                             | John Doe                                                                  |
+| `nativeAccountId`    | The user's native account ID                                                                                                                                                                 | 00000000-0000-0000-0000-000000000000                                      |
+
+> Note: `realm` is the name that `tenantId` receives in the cache.
+
+The following `getAccountBy` APIs are marked for deprecation and will be removed in a future version of MSAL. Please migrate to `getAccountByFilter()`:
```
sameerag: I assume you added `@deprecated` tag for typedocs too.

lib/msal-browser/docs/accounts.md
tnorling: We're on 3.1 right now, next release will be 3.2

lib/msal-browser/docs/accounts.md
tnorling: That's fair, if you know you have a username however, providing it as username will be slightly more performant. We should call out both so folks don't think they must provide it as loginHint

lib/msal-browser/docs/accounts.md
hectormmg: Sorry my bad, misread :S

lib/msal-browser/docs/accounts.md
hectormmg: Yep you can verify that in the diff for PublicClientApplication.

lib/msal-browser/docs/accounts.md
```diff
@@ -6,10 +6,22 @@
...
-* `getAccountByHomeId()`: receives a `homeAccountId` string and returns the matching account from the cache.
-* `getAccountByLocalId()`: receives a `localAccountId` string and returns the matching account from the cache.
-* `getAccountByUsername()`: receives a `username` string and returns the matching account from the cache.
+-   `getAllAccounts()`: returns all the accounts currently in the cache. Supports an optional filter to return a specific set of accounts. An application must choose an account to acquire tokens silently.
+-   `getAccount()`: returns the first cached account that matches the filter passed in. The order in which accounts are read from the cache is arbitrary and there is no guarantee that the first account in the filtered list will be the same for any two calls of `getAccount`. As explained below, increasing the number of filter attributes will provide more exact matches.
+
+### Account Filter Object
+
+The [AccountFilter](https://azuread.github.io/microsoft-authentication-library-for-js/ref/types/_azure_msal_common.AccountFilter.html) type documentation lists the properties that can be used and combined to filter accounts.
```
hectormmg, line 14: This link's destination doesn't exist yet because the type wasn't public before. If the link doesn't work I'll replace it once type docs are published.

lib/msal-browser/docs/accounts.md
hectormmg: I added some context/disambiguation.

lib/msal-browser/docs/accounts.md
hectormmg: Changed the wording

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -188,12 +189,22 @@ export class PublicClientApplication implements IPublicClientApplication {
...
+        return this.controller.getAccount(accountFilter);
+    }
+
     /**
      * Returns the signed in account matching homeAccountId.
      * (the account object is created at the time of successful login)
      * or null when no matching account is found
      * @param homeAccountId
      * @returns The account object stored in MSAL
+     * @deprecated - Use getAccountByFilter instead
```
```suggestion
     * @deprecated - Use getAccount instead
```

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -205,6 +216,7 @@ export class PublicClientApplication implements IPublicClientApplication {
      * or null when no matching account is found
      * @param localAccountId
      * @returns The account object stored in MSAL
+     * @deprecated - Use getAccountByFilter instead
```
```suggestion
     * @deprecated - Use getAccount instead
```

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -217,19 +229,19 @@ export class PublicClientApplication implements IPublicClientApplication {
      * This API is provided for convenience but getAccountById should be used for best reliability
      * @param userName
      * @returns The account object stored in MSAL
+     * @deprecated - Use getAccountByFilter instead
```
```suggestion
     * @deprecated - Use getAccount instead
```

lib/msal-browser/src/controllers/StandardController.ts
```diff
@@ -1196,18 +1197,50 @@ export class StandardController implements IController {
...
     }
 
+    /**
+     * Returns the first account found in the cache that matches the account filter passed in.
+     * @param accountFilter
+     * @returns The first account found in the cache matching the provided filter or null if no account could be found.
+     */
+    getAccount(accountFilter: AccountFilter): AccountInfo | null {
+        this.logger.trace("getAccount called");
+        if (!accountFilter || Object.keys(accountFilter).length === 0) {
```
tnorling: nit: accountFilter is required so we don't need to check for truthiness
```suggestion
        if (Object.keys(accountFilter).length === 0) {
```

tnorling: I wonder if we should also validate the values? Something like `{username: ""}` would bypass this check

lib/msal-browser/src/controllers/StandardController.ts
```diff
@@ -1196,18 +1197,50 @@ export class StandardController implements IController {
...
+
+        if (account) {
+            this.logger.verbose(
+                "getAccount: Account matching provided filter found, returning"
+            );
+            this.logger.verbosePii(
+                `getAccount: Returning signed-in account matching accountFilter: ${JSON.stringify(
+                    accountFilter
+                )}`
+            );
```
tnorling: Would prefer not to log this for the simple reason that JSON operations are expensive and this will execute regardless of whether or not it will end up being logged. If someone needs to debug they can add their own log statement at the callsite
```suggestion
```

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -1324,6 +1357,19 @@ export abstract class CacheManager implements ICacheManager {
...
 
+    /**
+     * helper to match names
+     * @param entity
+     * @param name
+     * @returns true if the downcased name properties are present and match in the filter and the entity
+     */
+    private matchName(entity: AccountEntity, name: string): boolean {
+        return !!(
+            typeof entity.name === "string" &&
```
tnorling: Can name be something other than string?

lib/msal-browser/docs/accounts.md
```diff
@@ -35,25 +47,97 @@ function handleResponse(resp) {
...
+    return await myMSALObj.acquireTokenPopup(request);
+});
+```
+
+Passing `username` as `username`
+
+```javascript
+const accountUsername = userProfile.username;
+const accountFilter = {
+    username: accountUsername; // More performant filtering
```
tnorling: nit: The explanation above is sufficient, I don't want to inadvertently encourage folks to use username instead of the claim
```suggestion
    username: accountUsername;
```

tnorling (approved):
A couple nits but otherwise looks great! Thanks for doing this!

lib/msal-browser/docs/accounts.md
```diff
@@ -6,10 +6,22 @@
...
+-   `getAllAccounts()`: returns all the accounts currently in the cache. Supports an optional filter to return a specific set of accounts. An application must choose an account to acquire tokens silently.
+-   `getAccount()`: returns the first cached account that matches the filter passed in. The order in which accounts are read from the cache is arbitrary and there is no guarantee that the first account in the filtered list will be the same for any two calls of `getAccount`. As explained below, increasing the number of filter attributes will provide more exact matches.
+
+### Account Filter Object
+
+The [AccountFilter](https://azuread.github.io/microsoft-authentication-library-for-js/ref/types/_azure_msal_common.AccountFilter.html) type documentation lists the properties that can be used and combined to filter accounts.
+
+> Note: A single account filter attribute is usually not guaranteed to uniquely identify a cached account object. Adding a combination of attributes that don't repeat together, such as `homeAccountId` + `localAccountId`, can help refine the search.
+
+> Note: `realm` is the name given to `tenantId` in the cache.
```
sameerag:
```suggestion
> Note: `realm` is `tenantId` in the cache.
```

lib/msal-browser/docs/accounts.md
```diff
@@ -35,25 +47,97 @@ function handleResponse(resp) {
...
+});
+```
+
 ## Active Account APIs
 
 The `@azure/msal-browser` library also provides 2 convenient APIs to help you keep track of which account is currently "active" and should be used for token requests.
 
-* `getActiveAccount()`: Returns the currently active account
-* `setActiveAccount()`: Receives an account object and sets it as the active account
+-   `getActiveAccount()`: Returns the currently active account
```
sameerag:
```suggestion
-   `getActiveAccount()`: Returns the current active account
```

lib/msal-browser/src/controllers/StandardController.ts
sameerag: You mean a non-empty, non-null value always?

sameerag (approved):
Minor comments, approved. Much clear now, Thanks.

peterzenz (approved):

lib/msal-browser/src/controllers/StandardController.ts
hectormmg: @tnorling  @sameerag I think empty username can technically be a valid search filter though. I think filter debugging can be left to the devs, but I'm still open to adding validation if my first assumption isn't true.

tnorling: Empty filters evaluate as true. So a filter of `{username: ""}` would match ALL accounts in the cache.

lib/msal-common/src/cache/CacheManager.ts
hectormmg: Refined the check to assume entity.name can only be passed in as string

lib/msal-browser/test/utils/StringConstants.ts
```diff
@@ -86,10 +86,12 @@ export const TEST_TOKENS = {
...
     REFRESH_TOKEN: "thisIsARefreshT0ken",
+    // [SuppressMessage("Microsoft.Security", "CS002:SecretInNextLine", Justification="Fake credential used for testing purposes")]
```
sameerag, line 89: Should we add this to all test tokens? We can take it as a new item in the backlog.

hectormmg: We could, but cred scan only flags new credentials it finds in the diff

Nodios:
Hi,

Just a quick question regarding this change. Is there a reason why the interface `IPublicClientApplication` was not updated with the method definition changes? I don't know if I'm missing something

See example below:

Interface https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/d055ad8cb91961dbc0639861c618f11900485853/lib/msal-browser/src/app/IPublicClientApplication.ts#L49

Implementation https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/d055ad8cb91961dbc0639861c618f11900485853/lib/msal-browser/src/app/PublicClientApplication.ts#L245

PR #6504, opened 2023-09-21
Address missing params in refreshtoken flow

sameerag:
* Address missing params in `acquireTokenByRefreshToken` for specific flows
* Correct request thumbprint `clientId` in `AuthorizationCodeClient` and `RefreshTokenClient`


lib/msal-browser/src/controllers/IController.ts
```diff
@@ -102,9 +103,12 @@ export interface IController {
 
     setNavigationClient(navigationClient: INavigationClient): void;
 
+    getRequestCorrelationId(request?: Partial<BaseAuthRequest>): string;
```
sameerag: We can rename it to something like `generateCorrelationId` if preferred. Suggestions welcome. This change is to enable PWB or any app to call this API to fetch the request correlationId or generate a GUID.

sameerag: We can also just restrict this to GUID generation API.

lib/msal-common/src/request/CommonRefreshTokenRequest.ts
```diff
@@ -22,4 +23,6 @@ export type CommonRefreshTokenRequest = BaseAuthRequest & {
     refreshToken: string;
     ccsCredential?: CcsCredential;
     forceCache?: boolean;
+    tokenBodyParameters?: StringDict;
```
sameerag, line 26: I am also toying around the idea to add this in `BaseAuthRequest` as this is supposed to be a generic param to be added to the `body` of a request sent to `/token` end point. 

These changes fix the bugs, however would like to consolidate the param addition at *many* places if `tokenBodyParameters` can be repurposed.

lib/msal-common/src/client/RefreshTokenClient.ts
```diff
@@ -293,7 +296,14 @@ export class RefreshTokenClient extends BaseClient {
         const correlationId = request.correlationId;
         const parameterBuilder = new RequestParameterBuilder();
 
-        parameterBuilder.addClientId(this.config.authOptions.clientId);
+        parameterBuilder.addClientId(
+            request.tokenBodyParameters?.[AADServerParamKeys.CLIENT_ID] ||
+                this.config.authOptions.clientId
+        );
+
+        if (request.redirectUri) {
```
sameerag, line 304: We need to add `redirectUri` in case of brokering as per the spec.

lib/msal-browser/src/controllers/IController.ts
sameerag: Let me do a different PR for correlation-id related fixes, to track this for RT use case.

konstantin-msft (approved):

lalimasharda (approved):

hectormmg (approved):

peterzenz (approved):

PR #6512, opened 2023-09-25
Instrument Crypto APIs

tnorling:
- Adds instrumentation for Crypto APIs, specifically around PKCE generation
- Also refactors `PkceGenerator` and `BrowserCrypto` classes into pure functions for better minification and simpler usage


peterzenz (approved):

lib/msal-browser/src/crypto/CryptoOps.ts
```diff
@@ -9,14 +9,20 @@ import {
...
-import { BrowserCrypto } from "./BrowserCrypto";
+import {
+    createNewGuid,
+    exportJwk,
+    generateKeyPair,
+    importJwk,
+    sha256Digest,
+    sign,
+    validateCryptoAvailable,
+} from "./BrowserCrypto";
```
sameerag: Can we do namespaces or modules and invoke any function in the namespace/module instead of importing each function? I think as we do more of this modeling for all utils, the imports will get too long.

konstantin-msft: Does adding namespaces negatively impacts tree shaking?

sameerag (approved):
One comment on function names imports, approved.

lib/msal-browser/src/crypto/CryptoOps.ts
tnorling: It does not. It's a good suggestion, I'll make the change here. My general opinion is if we're importing just one or two things I think it's fine to do so, anything more should be namespaced.

tnorling: Done, thanks for the suggestion!

konstantin-msft (approved):

PR #6514, opened 2023-09-26
MSAL Node's Managed Identity Api Implementation

Robbie-Microsoft:
This PR covers the API for the new (MSAL-Node specific) Managed Identity Client. Only the App Service source has been implemented for now. The Azure Arc, CloudShell, IMDS and Service Fabric sources will be in another PR.

The following is how a developer will create a new managed identity client and acquire a token

```
import * as msal from "@azure/msal-node";

// optional
const config: msal.ManagedIdentityConfiguration = {
    // optional
    managedIdentityIdParams: {
        userAssignedClientId: "managed_identity_id"
    },
    system: { // optional
        loggerOptions: msal.LoggerOptions; // optional
        networkClient: msal.INetworkModule; // optional
        proxyUrl: string; // optional
        customAgentOptions: http.AgentOptions | https.AgentOption; // optional
    },
};

// required
const request: msal.ManagedIdentityRequest = {
    forceRefresh: true, // optional, defaults to false
    resource: "URI_of_the_managed_identity_assigned_to_the_azure_resource" // required
};

..................

try {
    const managedIdentityClient = new msal.ManagedIdentityClient(config);
    const response = await managedIdentityClient.acquireToken(request);
    console.log(response);
} catch (error) {
    console.log(error);
}
```


bgavrilMS:
CC @pmaytak and @neha-bhargava on API proposal

lib/msal-node/src/client/ManagedIdentityClient.ts
```diff
@@ -0,0 +1,157 @@
...
+    ManagedIdentityConfiguration,
+    ManagedIdentityNodeConfiguration,
+    buildManagedIdentityConfiguration,
+} from "../config/Configuration";
+import { version, name } from "../packageMetadata.js";
+
+/**
+ * Class to initialize a managed identity and identify the service
+ */
+export class ManagedIdentityClient {
```
bgavrilMS, line 30: Will this work though? For example are you going to be able to reuse the token caching logic that is already tied to ConfidentialClientApplication?

lib/msal-node/src/config/Configuration.ts
```diff
@@ -107,6 +111,14 @@ export type Configuration = {
     telemetry?: NodeTelemetryOptions;
 };
 
+export type ManagedIdentityConfiguration = {
+    forceRefresh?: boolean;
+    id?: string;
+    resourceType: ManagedIdentityIdType;
```
bgavrilMS: Not part of the public API. MSAL deals with this internally.

lib/msal-node/src/client/ManagedIdentityApplication.ts
```diff
@@ -0,0 +1,137 @@
...
+                    this.managedIdentityId
+                ) ||
+                AzureArc.tryCreate(
+                    this.config,
+                    this.logger,
+                    this.cacheManager,
+                    this.networkManager,
+                    this.managedIdentityId
+                ) ||
+                Imds.tryCreate(
```
neha-bhargava: For IMDS, we directly call the constructor rather than try create the source. If none of the above sources are available IMDS will be used by default.

lib/msal-node/src/config/Configuration.ts
```diff
@@ -107,6 +111,11 @@ export type Configuration = {
     telemetry?: NodeTelemetryOptions;
 };
 
+export type ManagedIdentityConfiguration = Omit<
+    Configuration,
+    "auth" | "broker" | "cache" | "telemetry"
```
neha-bhargava: Does MSAL Node support any telemetry other than server side telemetry?

lib/msal-node/src/config/Configuration.ts
```diff
@@ -199,3 +208,25 @@ export function buildAppConfiguration({
...
+    NodeConfiguration,
+    "auth" | "broker" | "cache" | "telemetry"
+>;
+
+export function buildManagedIdentityConfiguration({
+    system: system,
+}: ManagedIdentityConfiguration): ManagedIdentityNodeConfiguration {
+    const systemOptions: Required<NodeSystemOptions> = {
+        ...DEFAULT_SYSTEM_OPTIONS,
+        loggerOptions: system?.loggerOptions || DEFAULT_LOGGER_OPTIONS,
```
neha-bhargava, line 233: Not sure but is this where the AppTokenCache should be enabled and default authority should be set?

lib/msal-node/src/utils/Constants.ts
```diff
@@ -3,6 +3,37 @@
  * Licensed under the MIT License.
  */
 
+// MSI Constants. Docs for MSI are available here https://docs.microsoft.com/azure/app-service/overview-managed-identity
+export const APP_SERVICE_MSI_API_VERSION: string = "2019-08-01";
+export const SECRET_HEADER_NAME: string = "X-IDENTITY-HEADER";
```
pmaytak: The constants that are specific to each source should probably go into those specific files, so it's clearer to which source they apply. (see [MSAL.NET](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/9029e1c78aa62a918ce1db1aea54a2cfe3c15293/src/client/Microsoft.Identity.Client/ManagedIdentity/AppServiceManagedIdentitySource.cs#L19))

lib/msal-node/src/config/ManagedIdentityId.ts
```diff
@@ -0,0 +1,90 @@
...
+     * Creates an instance of ManagedIdentityId for a user assigned managed identity from an object id
+     * @param objectId Object id of the user assigned managed identity assigned to the azure resource
+     * @returns Instance of ManagedIdentityId
+     */
+    public static createUserAssignedObjectId(
+        objectId: string
+    ): ManagedIdentityId {
+        // TODO: throw exception if objectId is null
+        return new ManagedIdentityId(
+            ManagedIdentityIdType.USER_ASSIGNED_RESOURCE_ID,
```
pmaytak:
```suggestion
            ManagedIdentityIdType.USER_ASSIGNED_OBJECT_ID,
```

lib/msal-node/src/config/ManagedIdentityId.ts
```diff
@@ -0,0 +1,90 @@
...
+     * Creates an instance of ManagedIdentityId for a user assigned managed identity from a client id
+     * @param clientId Client id of the user assigned managed identity assigned to the azure resource
+     * @returns Instance of ManagedIdentityId
+     */
+    public static createUserAssignedClientId(
+        clientId: string
+    ): ManagedIdentityId {
+        // TODO: throw exception if clientId is null
+        return new ManagedIdentityId(
+            ManagedIdentityIdType.USER_ASSIGNED_RESOURCE_ID,
```
pmaytak:
```suggestion
            ManagedIdentityIdType.USER_ASSIGNED_CLIENT_ID,
```

lib/msal-node/src/client/ManagedIdentityApplication.ts
```diff
@@ -0,0 +1,137 @@
...
+    buildManagedIdentityConfiguration,
+} from "../config/Configuration";
+import { version, name } from "../packageMetadata.js";
+import { ManagedIdentityRequest } from "../request/ManagedIdentityRequest";
+import { ManagedIdentityId } from "../config/ManagedIdentityId";
+
+/**
+ * Class to initialize a managed identity and identify the service
+ */
+export class ManagedIdentityApplication {
```
pmaytak, line 42: So do I understand the design:
- ManagedIdentityApplication is equivalent to ConfidentialClientApplication.
- Then each MI source class will be equivalent to ClientCredentialClient/OnBehalfOfClient/etc - classes that actually check the cache and get the tokens?

Right now, I do agree that inheriting from ClientApplication or BaseClient is not really useful.

pmaytak (approved):
Yea, I think looks good so far, I think once at least 1 source is implemented, the full picture would be a bit clearer.

lib/msal-node/src/client/ManagedIdentityApplication.ts
bgavrilMS: No, the "Sources" are an implementation detail. There is only 1 "AcquireToken" method. The user only needs to know: 

- their managed identity id (for user assigned) 
- the resource they want a token for (e.g. KeyVault)

That's it. No clientId, tenantid, authority. No cache either. 

Have a look at https://learn.microsoft.com/en-us/entra/msal/dotnet/advanced/managed-identity#examples for the .NET sample.

lib/msal-node/src/client/ManagedIdentityApplication.ts
Robbie-Microsoft: I'm adding extra error checking in here. Maybe all the environment variables are malformed?

lib/msal-node/src/config/ManagedIdentityId.ts
Robbie-Microsoft: good catch

lib/msal-node/src/config/Configuration.ts
Robbie-Microsoft: yes, it supports client telemetry. I will need to implement that eventually

lib/msal-node/src/config/Configuration.ts
Robbie-Microsoft: No, they should be initialized in ManagedIdentityApplication.ts

lib/msal-common/src/config/ClientConfiguration.ts
```diff
@@ -7,7 +7,10 @@ import { INetworkModule } from "../network/INetworkModule";
 import { DEFAULT_CRYPTO_IMPLEMENTATION, ICrypto } from "../crypto/ICrypto";
 import { AuthError } from "../error/AuthError";
 import { ILoggerCallback, Logger, LogLevel } from "../logger/Logger";
-import { Constants } from "../utils/Constants";
+import {
+    Constants,
+    DEFAULT_TOKEN_RENEWAL_OFFSET_SEC,
```
bgavrilMS, line 12: Why do we need to add this now? It should just default to 5 min

lib/msal-common/src/network/INetworkModule.ts
```diff
@@ -39,7 +39,8 @@ export interface INetworkModule {
      */
     sendPostRequestAsync<T>(
         url: string,
-        options?: NetworkRequestOptions
+        options?: NetworkRequestOptions,
+        cancellationToken?: number
```
bgavrilMS: Do we really need to add timeout to this interface? If we really do, can we add it as part of `options` ?

Btw, if this is a timeout, just call it timeout, not "token". CancellationToken has more functionality, timeout is just one of them.

lib/msal-common/src/config/ClientConfiguration.ts
```diff
@@ -18,9 +21,6 @@ import { ISerializableTokenCache } from "../cache/interface/ISerializableTokenCa
 import { ClientCredentials } from "../account/ClientCredentials";
 import { ProtocolMode } from "../authority/ProtocolMode";
 
-// Token renewal offset default in seconds
-const DEFAULT_TOKEN_RENEWAL_OFFSET_SEC = 300;
```
bgavrilMS, line 25: Is this part of the public API? Isn't this a breaking change?

lib/msal-common/src/network/NetworkManager.ts
```diff
@@ -43,23 +43,62 @@ export class NetworkManager {
...
-     * Wraps sendPostRequestAsync with necessary preflight and postflight logic
+     * Wraps the networkClient's sendGetRequestAsync with necessary preflight and postflight logic
+     * Managed Identity sends GET requests to communicate to the following managed identity sources:
+     * App Service, Azure Arc, Imds, Service Fabric
+     *
+     * @param thumbprint
+     * @param tokenEndpoint
+     * @param options
+     */
+    async sendGetRequest<T extends ServerAuthorizationTokenResponse>(
```
bgavrilMS: We discussed that `NetworkManager` is actually `TokenEndpointNetworkManager`. But in MSI case, you'll be hitting different endpoints, which will not return `ServerAuthorizationTokenResponse`. So I don't think `NetworkManager` can be used here. 

Consider creating an `MsiNetworkManager` or using the `networkClient` directly.

lib/msal-node/src/client/ManagedIdentityApplication.ts
```diff
@@ -0,0 +1,275 @@
...
+    private cryptoProvider: CryptoProvider;
+
+    private identitySource:
+        | ServiceFabric
+        | AppService
+        | CloudShell
+        | AzureArc
+        | Imds;
+
+    private fakeAuthority: Authority;
```
bgavrilMS, line 51: Pls add a comment here on why this is fake etc.

lib/msal-node/src/config/Configuration.ts
```diff
@@ -83,6 +84,8 @@ export type NodeSystemOptions = {
     networkClient?: INetworkModule;
     proxyUrl?: string;
     customAgentOptions?: http.AgentOptions | https.AgentOptions;
+    tokenRenewalOffsetSeconds?: number;
```
bgavrilMS: This is not a system option. Why do we need this to be configurable anyway?

lib/msal-node/src/config/Configuration.ts
```diff
@@ -83,6 +84,8 @@ export type NodeSystemOptions = {
...
+    cancellationToken?: number;
```
bgavrilMS: We should not add this new option just for MSI. Cancellation / timeout always existed in MSAL .NET but we are not bringing it in NodeJS as part of MSI. If ppl need it, we can bring it separately. 

Cancellation of async operations in NodeJS is different than cancellation in .NET

lib/msal-node/src/config/Configuration.ts
bgavrilMS: Actually I don't think you need to do that @Robbie-Microsoft . The way public client and SPA **client** telemetry is collected is different than confidential client. @neha-bhargava will do a demo on this soon I've heard :)

lib/msal-node/src/config/ManagedIdentityId.ts
```diff
@@ -0,0 +1,98 @@
...
+            case ManagedIdentityIdType.SYSTEM_ASSIGNED:
+                this._isUserAssignedId = false;
+                break;
+            case ManagedIdentityIdType.USER_ASSIGNED_CLIENT_ID:
+            case ManagedIdentityIdType.USER_ASSIGNED_RESOURCE_ID:
+            case ManagedIdentityIdType.USER_ASSIGNED_OBJECT_ID:
+                this._isUserAssignedId = true;
+                break;
+            default:
+                // TODO: throw error
```
bgavrilMS: There are a few simple TODOs here that can be easily solved.

lib/msal-node/src/config/ManagedIdentityId.ts
```diff
@@ -0,0 +1,98 @@
...
+                break;
+        }
+    }
+
+    /**
+     * Creates an instance of ManagedIdentityId for a system assigned managed identity
+     */
+    public static createSystemAssigned(): ManagedIdentityId {
```
bgavrilMS: How about `public static get systemAssignedIdentity()`

lib/msal-node/src/config/ManagedIdentityId.ts
```diff
@@ -0,0 +1,98 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import {
+    DEFAULT_MANAGED_IDENTITY_ID,
```
bgavrilMS, line 11: Is this constant part of the public API? Why?

lib/msal-node/src/config/ManagedIdentityId.ts
```diff
@@ -0,0 +1,98 @@
...
+    ManagedIdentityIdType,
+} from "../utils/Constants";
+
+export class ManagedIdentityId {
+    private _id: string;
```
bgavrilMS, line 17: inconsistent naming. Pls don't follow the .NET naming conventions. There is a linter in MSAL JS, pls ensure it runs and passes.

lib/msal-node/src/config/ManagedIdentityId.ts
```diff
@@ -0,0 +1,98 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import {
+    DEFAULT_MANAGED_IDENTITY_ID,
+    ManagedIdentityIdType,
+} from "../utils/Constants";
+
+export class ManagedIdentityId {
```
bgavrilMS, line 16: Since this is JS and ppl like JSON, I think you should have an object model that is more declarative and less static-y. 
```json
{ 
    "systemAssigned": true
}
```
```json
{ 
    "systemAssigned": false,
    "userAssignedClientId": 12345
}
```

```json
{ 
    "systemAssigned": false,
    "userAssignedObjectId": 12345
}
```

bgavrilMS: @rayluo for a perspective, as I think this is what we did in MSAL PY as well.

lib/msal-node/src/client/ManagedIdentityApplication.ts
```diff
@@ -0,0 +1,275 @@
...
+
+    constructor(
```
bgavrilMS: For consistency with `ConfidentialClientApplication`, we should have only 1 config param here, not 2. And that config param should be mandatory.

lib/msal-node/src/client/ManagedIdentityApplication.ts
```diff
@@ -0,0 +1,275 @@
...
+
+        this.networkClient = this.config.system.networkClient;
+        this.networkManager = new NetworkManager(
+            this.networkClient,
+            this.cacheManager
+        );
+
+        this.cryptoProvider = new CryptoProvider();
+
+        this.identitySource = this.selectManagedIdentitySource();
```
bgavrilMS: We shouldn't try to select the MI source when creating the app, but when firing the request.
Once a request has figured out what MI source we're on (e.g. VM), all other requests need to remeber this (i.e. static cache this)

lib/msal-node/src/client/ManagedIdentityApplication.ts
```diff
@@ -0,0 +1,275 @@
...
+        this.config = buildManagedIdentityConfiguration(configuration || {});
+        this.managedIdentityId = managedIdentityId;
+
+        this.logger = new Logger(
+            this.config.system.loggerOptions,
+            name,
+            version
+        );
+
+        this.cacheManager = new DefaultStorageClass(
```
bgavrilMS: What is the difference between `NodeStorage`, which ClientApplicaiton uses and this one?

lib/msal-node/src/client/ManagedIdentityApplication.ts
```diff
@@ -0,0 +1,275 @@
...
+            this.managedIdentityId.id,
+            DEFAULT_CRYPTO_IMPLEMENTATION,
+            this.logger
+        );
+
+        this.networkClient = this.config.system.networkClient;
+        this.networkManager = new NetworkManager(
```
bgavrilMS: Forget about network manager. It is a wrapper over networkClient that adds throttling. The throttling logic is specific to talking to ESTS and does not apply here.

lib/msal-node/src/request/ManagedIdentityRequest.ts
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { BaseAuthRequest } from "@azure/msal-common/request/BaseAuthRequest";
+
+/**
+ * ManagedIdentityRequest
```
bgavrilMS, line 10: Please add better comments, based on MSAL.NET ones which were already reviewed etc.

lib/msal-node/src/request/ManagedIdentityRequest.ts
```diff
@@ -0,0 +1,16 @@
...
+ * - forceRefresh - forces managed identity requests to skip the cache and make network calls if true
+ * - resourceUri  - the URI of the managed identity
```
bgavrilMS: See https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/IManagedIdentityApplication.cs 

Please add a task to the board to make sure we support both formats. Can be addressed separately. For now just focus on "resourceUri"

lib/msal-node/src/client/ManagedIdentityApplication.ts
```diff
@@ -0,0 +1,275 @@
...
+     * @param managedIdentityRequest
+     * @param managedIdentityId
+     * @returns the access token
+     */
+    public async acquireToken(
+        managedIdentityRequest: ManagedIdentityRequest,
+        managedIdentityId: ManagedIdentityId
+    ): Promise<AuthenticationResult | null> {
+        return (
+            (await this.getCachedToken(
```
bgavrilMS: Think how you can refactor the code from ClientCredentialClient so as to reuse the fetching from cache logic. We should not rewrite this because it is  common between client_credentials flow and MSI. For example, you already have a bug in your code because you don't skip the cache when forceRefresh=true. 

I see 2 ways: 

- create a ClietnCredentialsRequest and set an AppTokenProvider. The AppTokenProvider makes MSI calls etc. Caching is ensured by the ClietnCredentialsRequest.
- OR share code from ClietnCredentialsRequest 

I'd go for the first one, but @neha-bhargava can also suggest.

lib/msal-common/src/network/INetworkModule.ts
Robbie-Microsoft: It seems to me that this was a bug where the timeout was missing from the post request definition in InNetworkModule. In httpClient, we allow the timeout (cancellationToken) to be optionally provided. @sameerag, can you chime in here? Was this originally supposed to be timeout functionality, or a cancellationToken?

lib/msal-common/src/config/ClientConfiguration.ts
Robbie-Microsoft: It is part of the public api, and it's not a breaking change. I updated the imports in msal-common.

It's used in all clients (public api, msal-common), but Managed Identity (msal-node) needs this constant.

lib/msal-common/src/network/NetworkManager.ts
Robbie-Microsoft: Based on your other comment about MSI not needing throttling, I think it makes more sense to just call the networkClient directly.

lib/msal-node/src/config/Configuration.ts
Robbie-Microsoft: It's used to compare the access token to, to see if the access token is expired. This is something that already exists in CommonClientConfiguration, but MSI does not extend CommonClientConfiguration. We could just keep it at it's default value of 300 ms and not allow the developer to optionally configure it, if you'd like.

lib/msal-node/src/config/Configuration.ts
Robbie-Microsoft: I mentioned this in a response to the previous comment, but these already exist in the CommonClientConfiguration which all other msal-node clients extend. MSI doesn't extend CommonClientConfiguration, so I added them to the MSI configuration. Thoughts? @sameerag what are your thoughts?

lib/msal-node/src/config/ManagedIdentityId.ts
Robbie-Microsoft: I don't understand your question.

lib/msal-node/src/config/ManagedIdentityId.ts
Robbie-Microsoft: I don't understand what you're saying. I run the linter before every push to GitHub.

lib/msal-node/src/client/ManagedIdentityApplication.ts
Robbie-Microsoft: Making this change would mean editing the MSI config and making managedIdentityId a required part of the config, provided by the developer. I can do that if you'd like.

lib/msal-node/src/client/ManagedIdentityApplication.ts
Robbie-Microsoft: I disagree. If we select the MSI source when creating the app, then it will already be known when firing the requests. Isn't this how .dotnet does it?

lib/msal-node/src/client/ManagedIdentityApplication.ts
Robbie-Microsoft: Good question, I'll have to look into this more. It appears NodeStorage is used in TokenCache, and TokenCache's methods are exposed to the user. For example: serialize, deserialize, getAllAccounts, etc. I don't think we want to expose these to the developer of an MSI application? If not, CacheManager should suffice for storing tokens in a cache. @sameerag am I understanding this correctly?

lib/msal-node/src/request/ManagedIdentityRequest.ts
Robbie-Microsoft: I don't understand? forceRefresh is optional here.

lib/msal-node/src/client/ManagedIdentityApplication.ts
Robbie-Microsoft: I don't understand this comment. Refactoring the code from ClientCredentialClient is exactly how I wrote the caching logic here. I do skip the cache when forceRefresh is true... on line 144.

lib/msal-node/src/config/Configuration.ts
Robbie-Microsoft: I believe it is a system option. Look in lib/msal-node/src/client/ClientCredentialClient.ts on line 127

lib/msal-common/src/network/INetworkModule.ts
bgavrilMS: Let's not get bogged down in this @Robbie-Microsoft - just open a bug and we'll treat it separately. Let's not make changes to the existing interface if they are not needed.

lib/msal-node/src/client/ManagedIdentityApplication.ts
bgavrilMS: I think that's a better developer experience? Managed Identity ID is mandatory.

This is very similar to `client_id`,. which is mandatory for CCA

lib/msal-node/src/client/ManagedIdentityApplication.ts
bgavrilMS: No, .NET selects the source when making the request itself. It then caches the sources in a **static** variable (so for the duration of the process). 

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs#L156

The important bit here is to cache it statically, so that no matter how many MSIAppcalition objects you created, the discovery process should only occur once. 
For consistency purposes, I'd keep this at request level.

lib/msal-node/src/client/ManagedIdentityApplication.ts
```diff
@@ -0,0 +1,270 @@
...
+    /**
+     * Acquire an access token from the cache or the managed identity
+     * @param managedIdentityRequest
+     * @param managedIdentityId
+     * @returns the access token
+     */
+    public async acquireToken(
+        managedIdentityRequest: ManagedIdentityRequest,
+        managedIdentityId: ManagedIdentityId
+    ): Promise<AuthenticationResult | null> {
```
bgavrilMS: Will it ever return null? I don't think so. It should always return `AuthenticationResult` or throw exception

lib/msal-node/src/client/ManagedIdentityApplication.ts
bgavrilMS: ClientCredentialClient does more things like calculates the outcome and does pro-active token refresh. We need that functionality as well. See `getCachedAuthenticationResult` - can we reuse that?

lib/msal-node/src/config/Configuration.ts
bgavrilMS: Can we not expose this for MSI please? I have no idea why it's exposed for ConfidentialClient, there are no scenarios that I know of where we'd want app developers to change this. Maybe in Public Client and SPA, but not in Confidential Client and MSI.

lib/msal-node/src/config/Configuration.ts
bgavrilMS: Let's start without. We can always add later?

lib/msal-node/src/request/ManagedIdentityRequest.ts
bgavrilMS: I am not talking about force refresh, I am talking about the format of the resource URI string (btw, please rename the param as `resource` since it's not a URI ). 

See https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/IManagedIdentityApplication.cs#L22

lib/msal-node/src/config/ManagedIdentityId.ts
bgavrilMS: If I look at existing files like `ClientCredentialClient.ts` I see that private variables are defined as 

`private scopeSet: ScopeSet;`

But in your case you use 

`private _id: string`

So the naming is inconsistent. Maybe the linter doesn't work or doesn't pick up on this? In any case, pls use consistent naming of vars.

lib/msal-common/src/network/INetworkModule.ts
Robbie-Microsoft: Well regardless, this change is needed for Managed Identity if we're exposing the timeout to the developer.

But per your other comment, we can start without exposing this value and always add it later.

lib/msal-node/src/config/Configuration.ts
Robbie-Microsoft: How should we check if the access token is expired? Use the default value?

lib/msal-node/src/client/ManagedIdentityApplication.ts
Robbie-Microsoft: It will return null if forceRefresh has been set to true, if there is no cached access token, or if the cached access token is expired. This is how the ClientCredentials flow functions as well. I think it makes more sense to return null.

lib/msal-node/src/client/ManagedIdentityApplication.ts
Robbie-Microsoft: That is essentially what I've already done, but I left those parts out. I will add them.

lib/msal-node/src/config/Configuration.ts
bgavrilMS: Yes. 5 min.

lib/msal-node/src/client/ManagedIdentityApplication.ts
Robbie-Microsoft: On second pass, it looks like NodeStorage extends CacheManager, which this uses. 

NodeStorage "implements Storage for node, reading cache from user specified storage location or an  extension library". Managed Identity doesn't do either of those thing, so CacheManager should suffice.

lib/msal-node/src/client/ManagedIdentityApplication.ts
neha-bhargava: I think we should reuse the code for getting token from cache and saving the token to cache from client credential. And add conditions to skip the parts not relevant to managed identity.

lib/msal-node/src/config/ManagedIdentityId.ts
Robbie-Microsoft: resolved per our conversation

lib/msal-node/src/config/ManagedIdentityId.ts
Robbie-Microsoft: resolved per our conversation

lib/msal-node/src/config/ManagedIdentityId.ts
Robbie-Microsoft: I've started implementing this. I think this is more error prone as the developer could input all three of userAssignedClientId, userAssignedResourceId and userAssignedObjectId. I can add if statements to mitigate this, but it will be much more code than before.

bgavrilMS: Ok, how about this then. Let's support the JSON approach for the most common use case. I believe 99% of app developers rely on system assigned identity or on user assigned identity + client_id. 

```json
{ 
    "systemAssigned": true
}
```

```json
{ 
    "systemAssigned": false, // optional, but if set, it can't be true
    "userAssignedIdenityClientId": 12345
}
```

The 1% needing user assigned identity by resourceId can use the static method.

lib/msal-node/src/client/ManagedIdentityApplication.ts
Robbie-Microsoft: resolved per our discussion

lib/msal-node/src/client/ManagedIdentityApplication.ts
Robbie-Microsoft: resolved, per our conversation

lib/msal-node/src/client/ManagedIdentityApplication.ts
Robbie-Microsoft: Another Update. I made a mistake when first using DefaultStorageClass. None of the cache methods are implemented in DefaultStorageClass. I should have been using NodeStorage initially.

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -1432,11 +1432,16 @@ export abstract class CacheManager implements ICacheManager {
                     this.staticAuthorityOptions.canonicalAuthority
                 ) ||
                 this.staticAuthorityOptions.knownAuthorities;
-            if (
-                staticAliases &&
-                staticAliases.includes(environment) &&
-                staticAliases.includes(entity.environment)
-            ) {
+
+            const validEnvironment = staticAliases?.some((staticAlias) => {
```
bgavrilMS, line 1436: What was the bug? Is it related to the use "indexOf" instead of "includes" ?

pmaytak: Why doesn't `includes` work here?
Is this related to managed identity? If not, I think it should be included in a separate PR into dev.

lib/msal-common/src/utils/Constants.ts
```diff
@@ -442,3 +442,6 @@ export const JsonTypes = {
 export type JsonTypes = (typeof JsonTypes)[keyof typeof JsonTypes];
 
 export const ONE_DAY_IN_MS = 86400000;
+
+// Token renewal offset default in seconds
+export const DEFAULT_TOKEN_RENEWAL_OFFSET_SEC = 300;
```
pmaytak, line 447: If this is not related to managed identity, separate into a separate PR?

lib/msal-common/src/response/AuthenticationResult.ts
```diff
@@ -44,4 +44,6 @@ export type AuthenticationResult = {
     msGraphHost?: string;
     code?: string;
     fromNativeBroker?: boolean;
+    managedIdentityClientId?: string;
+    managedIdentityResource?: string;
```
pmaytak: Do we need to return these? The user already knows what they are since they pass them into a request.

lib/msal-node/src/config/Configuration.ts
```diff
@@ -107,6 +108,17 @@ export type Configuration = {
...
+export type ManagedIdentityIdParams = {
+    userAssignedClientId?: string;
+    userAssignedResourceId?: string;
+    userAssignedObjectId?: string;
+};
+
+export type ManagedIdentityConfiguration = {
+    managedIdentityIdParams?: ManagedIdentityIdParams;
+    system?: NodeSystemOptions;
+};
```
pmaytak, line 120: What about instead of `ManagedIdentityIdParams`, `ManagedIdentityConfiguration` accepts `ManagedIdentityIdType` (required) and `UserAssignedId` (optional, in case of system assigned Id) to make API clearer. Currently with `ManagedIdentityIdParams`, what if multiple IDs are set, what if none are set?

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/AppConfig/ManagedIdentityId.cs

lib/msal-node/src/config/Configuration.ts
```diff
@@ -199,3 +211,31 @@ export function buildAppConfiguration({
...
+
+export type ManagedIdentityNodeConfiguration = {
+    managedIdentityId: ManagedIdentityId;
+    system: Required<NodeSystemOptions>;
+};
+
+export function buildManagedIdentityConfiguration({
+    managedIdentityIdParams,
+    system,
+}: ManagedIdentityConfiguration): ManagedIdentityNodeConfiguration {
```
pmaytak, line 226:
```typescript
({
    managedIdentityIdParams,
    system,
}: ManagedIdentityConfiguration)
```
What does this syntax mean? This method accepts an instance of `ManagedIdentityConfiguration`, and it's syntactic sugar, so you don't have to call config.paramBlah?

lib/msal-node/src/config/Configuration.ts
```diff
@@ -107,6 +108,17 @@ export type Configuration = {
     telemetry?: NodeTelemetryOptions;
 };
 
+export type ManagedIdentityIdParams = {
+    userAssignedClientId?: string;
+    userAssignedResourceId?: string;
+    userAssignedObjectId?: string;
+};
+
+export type ManagedIdentityConfiguration = {
```
pmaytak, line 117: What about instead of `ManagedIdentityIdParams`, `ManagedIdentityConfiguration` accepts `ManagedIdentityIdType` (required) and `UserAssignedId` (optional, in case of system assigned Id) to make API clearer. Currently with `ManagedIdentityIdParams`, what if multiple IDs are set, what if none are set?

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/AppConfig/ManagedIdentityId.cs

lib/msal-node/src/config/Configuration.ts
```diff
@@ -199,3 +211,31 @@ export function buildAppConfiguration({
         telemetry: { ...DEFAULT_TELEMETRY_OPTIONS, ...telemetry },
     };
 }
+
+export type ManagedIdentityNodeConfiguration = {
```
pmaytak, line 218: Can `ManagedIdentityNodeConfiguration` and `ManagedIdentityConfiguration` be combined? Let the user provide `ManagedIdentityId`? These classes look very similar.

lib/msal-node/src/config/ManagedIdentityId.ts
pmaytak: So @bgavrilMS you're suggesting not to follow how .NET did it? +1 to Robbie's comment that I would narrow the API down. We could allow the user to set (ManagedIdentityIdType, UserProvidedId?) (although in .NET IdType is internal).

lib/msal-node/src/config/ManagedIdentityRequestParameters.ts
```diff
@@ -0,0 +1,54 @@
...
+
+        const queryParametersString = parameterBuilder.createQueryString();
+
+        return UrlString.appendQueryString(
+            this._baseEndpoint,
+            queryParametersString
+        );
+    }
+
+    public computeParametersBodyString(): string {
```
pmaytak, line 39: What is this meant to do?

lib/msal-node/src/request/ManagedIdentityRequest.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import { CommonClientCredentialRequest } from "@azure/msal-common";
+import { ManagedIdentityRequestParams } from "./ManagedIdentityRequestParams";
+
+/**
+ * ManagedIdentityRequest
+ * - forceRefresh - forces managed identity requests to skip the cache and make network calls if true
+ * - resource  - resource requested to access the protected API. It should be of the form "{ResourceIdUri}" or {ResourceIdUri/.default}. For instance https://management.azure.net or, for Microsoft Graph, https://graph.microsoft.com/.default
+ */
+export type ManagedIdentityRequest = ManagedIdentityRequestParams &
```
pmaytak, line 14: I assume this syntax means `ManagedIdentityRequest` inherits all properties from both of these classes?

lib/msal-node/src/request/ManagedIdentityRequestParams.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+/**
+ * ManagedIdentityRequest
+ * - forceRefresh - forces managed identity requests to skip the cache and make network calls if true
+ * - resource  - resource requested to access the protected API. It should be of the form "{ResourceIdUri}" or {ResourceIdUri/.default}. For instance https://management.azure.net or, for Microsoft Graph, https://graph.microsoft.com/.default
+ */
+export type ManagedIdentityRequestParams = {
```
pmaytak, line 11: - There's already similarly named `ManagedIdentityRequestParameters` class. Can this class be eliminated and these properties directly be in `ManagedIdentityRequest`, and then for consistency `ManagedIdentityApplication.acquireToken` will accept `ManagedIdentityRequest` instance?

lib/msal-node/src/request/ManagedIdentityRequestParams.ts
```diff
@@ -0,0 +1,14 @@
...
+    forceRefresh?: boolean;
+    resource: string;
```
pmaytak, line 13: Does JS care about order of required and optional parameters or is there some coding convention? In .NET required parameters must come before any optional ones.

lib/msal-common/src/response/ServerAuthorizationTokenResponse.ts
```diff
@@ -51,4 +51,15 @@ export type ServerAuthorizationTokenResponse = {
     trace_id?: string;
     correlation_id?: string;
     claims?: string;
+
+    /**
+     * Unique to Managed Identity
+     */
+    // success
+    client_id?: string;
```
pmaytak: Since these properties are kind of separate from the above ones, can they be extracted into something like `ManagedIdentityResponse` and then called like `networkClient.sendPostRequestAsync<ManagedIdentityResponse>?

lib/msal-node/src/utils/Constants.ts
```diff
@@ -3,6 +3,44 @@
...
+    USER_ASSIGNED_RESOURCE_ID: "user-assigned-resource-id",
+    USER_ASSIGNED_OBJECT_ID: "user-assigned-object-id",
+} as const;
+export type ManagedIdentityIdType =
+    (typeof ManagedIdentityIdType)[keyof typeof ManagedIdentityIdType];
+
+/**
+ * Managed Identity Authentication Sources
+ */
+export const ManagedIdentitySource = {
```
pmaytak: Is this used anywhere?

lib/msal-node/src/client/ManagedIdentityApplication.ts
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    /**
+     * Acquire an access token from the cache or the managed identity
+     * @param managedIdentityRequest
+     * @returns the access token
+     */
+    public async acquireToken(
+        managedIdentityRequestParams: ManagedIdentityRequestParams
+    ): Promise<ManagedIdentityResult> {
```
pmaytak: Shouldn't this return `AuthenticationResult` like other APIs?

lib/msal-node/src/client/ManagedIdentityClient.ts
```diff
@@ -0,0 +1,126 @@
...
+             *  )
+             */
+
+            if (!source) {
+                throw createManagedIdentityError(
+                    ManagedIdentityErrorCodes.unableToCreateSource
+                );
+            }
+            return source;
+        } catch (error) {
```
pmaytak: nit: try/catch not needed if we're just throwing?

lib/msal-node/src/client/ManagedIdentityClient.ts
```diff
@@ -0,0 +1,126 @@
...
+                this.selectManagedIdentitySource(
+                    logger,
+                    cacheManager,
+                    networkClient,
+                    cryptoProvider
+                );
+        }
+    }
+
+    public async sendMSITokenRequest(
```
pmaytak: nit: `sendManagedIdentityTokenRequest` for consistency.

lib/msal-node/src/client/ManagedIdentitySources/BaseManagedIdentitySource.ts
```diff
@@ -0,0 +1,141 @@
...
+        this.networkClient = networkClient;
+        this.cryptoProvider = cryptoProvider;
+    }
+
+    abstract createRequest(
+        request: string,
+        managedIdentityId: ManagedIdentityId
+    ): ManagedIdentityRequestParameters;
+
+    public async authenticateWithMSI(
```
pmaytak: nit: `acquireTokenWithManagedIdentity` or `authenticateWithManagedIdentity`

lib/msal-node/src/client/ManagedIdentitySources/BaseManagedIdentitySource.ts
```diff
@@ -0,0 +1,141 @@
...
+import { CryptoProvider } from "../../crypto/CryptoProvider";
+import { ManagedIdentityRequest } from "../../request/ManagedIdentityRequest";
+import { HttpMethod } from "../../utils/Constants";
+import { ManagedIdentityResult } from "../../response/ManagedIdentityResult";
+import { ManagedIdentityUtils } from "../../utils/ManagedIdentityUtils";
+
+export abstract class BaseManagedIdentitySource {
+    protected logger: Logger;
+    private cacheManager: CacheManager;
+    private networkClient: INetworkModule;
```
pmaytak, line 33: I couldn't find it but what is the actual implementation of INetworkMobile that gets passed in here?

lib/msal-common/src/cache/entities/AccessTokenEntity.ts
```diff
@@ -76,7 +78,9 @@ export class AccessTokenEntity extends CredentialEntity {
         userAssertionHash?: string,
         keyId?: string,
         requestedClaims?: string,
-        requestedClaimsHash?: string
+        requestedClaimsHash?: string,
+        managedIdentityClientId?: string,
+        managedIdentityResource?: string
```
neha-bhargava: I think you can avoid adding these 2 parameters and just pass managedIdentityClientId as clientId and managedIdentityResource as scopes. Should work fine to generate cache entity.

lib/msal-common/src/response/AuthenticationResult.ts
neha-bhargava: We do not need the managedIdentityClientId but pass the managedIdentityResource as scopes on line 29.

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -635,6 +640,12 @@ export class ResponseHandler {
                 cacheRecord.account?.msGraphHost || Constants.EMPTY_STRING,
             code: serverTokenResponse?.spa_code,
             fromNativeBroker: false,
+            managedIdentityClientId:
```
neha-bhargava: Does not seem right to have managedIdentityClientId and managedIdentityResource added in general. These parameters map to clientId and scopes respectively. Any method that has scopes in the flow should ideally contain the resource from managed identity request and client id should be managedIdentityClientId.

lib/msal-common/src/response/ServerAuthorizationTokenResponse.ts
Robbie-Microsoft: I could, and that's how I originally did it, but then I'd have to change more code in msal-common, which is what I'm trying to avoid. I'd have two different server responses. In ResponseHandler on line 399, I'd have to change the type from ServerAuthorizationTokenResponse to ServerAuthorizationTokenResponse | ManagedIdentityTokenResponse, and then the rest of the code in that function would throw errors and I'd have to add a bunch of conditional checks to see if properties on the server response exist. Thoughts?

lib/msal-common/src/response/AuthenticationResult.ts
bgavrilMS: Also this is a public type, any changes to public API should be highlighted and reviewed. 

In this case, the change is not needed. AuthResult does not need to change for MSI.

lib/msal-common/src/response/ServerAuthorizationTokenResponse.ts
bgavrilMS: Agree with Peter, this is violating the Single Responsability Principle. This object is only for ESTS responses. We should have a different response for MSI - MSIResponse and if needed, an adapter (e.g. method on ServerAuthorizationResponse) that converts between MSIResponse to ServerAuthorizationResponse.

lib/msal-node/src/client/ManagedIdentitySources/BaseManagedIdentitySource.ts
```diff
@@ -0,0 +1,150 @@
...
+            managedIdentityRequest
+        );
+
+        return ManagedIdentityUtils.convertAuthResultToManagedIdentityResult(
+            authResult
+        );
+    }
+
+    // used in unit tests
+    public isAppService(): boolean {
```
bgavrilMS: This is a code smell. Base class should not know about derived class. 

Why don't you just copy this logic to the tests?

lib/msal-node/test/client/ManagedIdentityApplication.spec.ts
```diff
@@ -0,0 +1,326 @@
...
+                expect(
+                    managedIdentityApplication.getConfig.managedIdentityId
+                        .getIdType
+                ).toEqual(ManagedIdentityIdType.SYSTEM_ASSIGNED);
+                expect(
+                    managedIdentityApplication.getConfig.managedIdentityId
+                        .getIsUserAssignedId
+                ).toEqual(false);
+
+                // wait two seconds for the token to be saved to the cache, before checking the cache for the token
```
bgavrilMS: no dead code pls

lib/msal-node/src/config/Configuration.ts
bgavrilMS: Are you talking about the public API here @pmaytak ? I think the public API in NodeJS should be JSON driven and similar to the config option (also JSON) from [ID.Web](https://github.com/AzureAD/microsoft-identity-web/issues/1775)

image

lib/msal-node/test/client/ManagedIdentityApplication.spec.ts
```diff
@@ -0,0 +1,326 @@
...
+        },
+    };
+
+    const systemAssignedConfig: ManagedIdentityConfiguration = {
+        system: {
+            networkClient:
+                ManagedIdentityTestUtils.getManagedIdentityNetworkClient(
+                    DEFAULT_MANAGED_IDENTITY_ID
+                ),
+            // managedIdentityIdParams will be omitted for system-assigned
```
bgavrilMS: @localden  @pmaytak  - this (and line 43) is the public API currently. Please have a look.

lib/msal-node/test/client/ManagedIdentityApplication.spec.ts
```diff
@@ -0,0 +1,326 @@
...
+    afterAll(() => {
+        // restore old environment
+        process.env = OLD_ENVS;
+    });
+
+    const managedIdentityRequestParams: ManagedIdentityRequestParams = {
+        resource: MANAGED_IDENTITY_RESOURCE,
+    };
+
+    const nonSystemAssignedConfig: ManagedIdentityConfiguration = {
```
bgavrilMS: Please call this "userAssignedMSIConfig" or smth similar.

lib/msal-node/test/client/ManagedIdentityApplication.spec.ts
```diff
@@ -0,0 +1,326 @@
...
+            expect(
+                managedIdentityApplication instanceof ManagedIdentityApplication
+            ).toBe(true);
+        });
+    });
+
+    describe("Acquires a token successfully via an App Service Managed Identity", () => {
+        beforeAll(() => {
+            // make a copy
+            process.env = { ...OLD_ENVS };
```
bgavrilMS, line 123: Isn't this done already for all tests?

lib/msal-node/test/client/ManagedIdentityApplication.spec.ts
```diff
@@ -0,0 +1,326 @@
...
+            process.env["IDENTITY_ENDPOINT"] = "fake_IDENTITY_ENDPOINT";
+            process.env["IDENTITY_HEADER"] = "fake_IDENTITY_HEADER";
+        });
+
+        afterAll(() => {
+            // restore old environment
+            process.env = OLD_ENVS;
+        });
+
+        test("acquires a Non-System-Assigned token", async () => {
```
bgavrilMS: Pls avoid calling it "non-system-assigned" and use the term "user assigned" when you mean user assigned + client ID and "user assigned by object ID" etc.

lib/msal-node/test/client/ManagedIdentityApplication.spec.ts
```diff
@@ -0,0 +1,326 @@
...
+            const managedIdentityApplication = new ManagedIdentityApplication(
+                nonSystemAssignedConfig
+            );
+            expect(managedIdentityApplication.isAppService()).toBe(true);
```
bgavrilMS: Not a public API we want to support.

lib/msal-node/test/client/ManagedIdentityApplication.spec.ts
```diff
@@ -0,0 +1,326 @@
...
+
+            const networkManagedIdentityResult =
```
bgavrilMS: What is the type here? `AuthenticationResult` ?

lib/msal-node/test/client/ManagedIdentityApplication.spec.ts
```diff
@@ -0,0 +1,326 @@
...
+                await managedIdentityApplication.acquireToken(
+                    managedIdentityRequestParams
+                );
+
+            expect(networkManagedIdentityResult.accessToken).toEqual(
+                DEFAULT_MANAGED_IDENTITY_AUTHENTICATION_RESULT.accessToken
+            );
+            expect(
+                managedIdentityApplication.getConfig.managedIdentityId.getId
```
bgavrilMS: Is this null for System assigned?

lib/msal-node/test/client/ManagedIdentityApplication.spec.ts
```diff
@@ -0,0 +1,326 @@
...
+            ).toEqual(MANAGED_IDENTITY_CLIENT_ID);
+            expect(
+                managedIdentityApplication.getConfig.managedIdentityId.getIdType
```
bgavrilMS: Can we hide "getIdType" from the public API please?

lib/msal-node/src/config/ManagedIdentityId.ts
```diff
@@ -0,0 +1,141 @@
...
+    /*
+     *     return new ManagedIdentityId({ userAssignedObjectId: objectId });
+     * }
+     */
+
+    public get getId(): string {
+        return this.id;
+    }
+
+    public get getIdType(): ManagedIdentityIdType {
```
bgavrilMS: void using "getX" in getters. Use just "X". E.g. `public get Id`

lib/msal-node/src/config/ManagedIdentityId.ts
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    /**
+     * Creates an instance of ManagedIdentityId for a system assigned managed identity
+     */
+    public static createSystemAssigned(): ManagedIdentityId {
+        return new ManagedIdentityId();
+    }
+
+    // /**
```
bgavrilMS: Pls remove dead code from PRs

lib/msal-node/src/config/ManagedIdentityId.ts
bgavrilMS: MSAL .NET is not JSON driven, but MSAL Node is and Id.Web are. So proposal is more like: https://github.com/AzureAD/microsoft-identity-web/issues/1775#issuecomment-1769864869

Agree on hiding "IdType" - this is implementation detail. 

@Robbie-Microsoft - can you please organize a meeting and include all reviewers and @localden and discuss the public API you have ? In particular we need to see how to specify SAMI and UAMI.

lib/msal-node/src/config/ManagedIdentityId.ts
bgavrilMS: I don't think this should be public API

lib/msal-node/src/config/ManagedIdentityId.ts
```diff
@@ -0,0 +1,141 @@
...
+
+    public get getId(): string {
+        return this.id;
+    }
+
+    public get getIdType(): ManagedIdentityIdType {
+        return this.idType;
+    }
+
+    public get getIsUserAssignedId(): boolean {
```
bgavrilMS: I don't think this should be public API

lib/msal-node/test/client/ManagedIdentityApplication.spec.ts
```diff
@@ -0,0 +1,326 @@
...
+    };
+
+    const nonSystemAssignedConfig: ManagedIdentityConfiguration = {
+        system: {
+            networkClient:
+                ManagedIdentityTestUtils.getManagedIdentityNetworkClient(
+                    MANAGED_IDENTITY_CLIENT_ID
+                ),
+        },
+        managedIdentityIdParams: {
```
bgavrilMS, line 85: In hindsight, I think this would be better named as "managedIdentityConfig"

lib/msal-node/test/client/ManagedIdentityApplication.spec.ts
```diff
@@ -0,0 +1,326 @@
...
+                        .getIdType
+                ).toEqual(ManagedIdentityIdType.SYSTEM_ASSIGNED);
+                expect(
+                    managedIdentityApplication.getConfig.managedIdentityId
+                        .getIsUserAssignedId
+                ).toEqual(false);
+            });
+        });
+    });
+});
```
bgavrilMS, line 745: We need tests that validate the access token cache key for all configurations - system assigned, user assigned + client etc.

lib/msal-node/src/config/ManagedIdentityId.ts
```diff
@@ -0,0 +1,141 @@
...
+export class ManagedIdentityId {
+    private id: string;
+    private idType: ManagedIdentityIdType;
+    private isUserAssignedId: boolean;
+
+    constructor(managedIdentityIdParams?: ManagedIdentityIdParams) {
+        const userAssignedClientId =
+            managedIdentityIdParams?.userAssignedClientId;
+        const userAssignedResourceId =
+            managedIdentityIdParams?.userAssignedResourceId;
```
bgavrilMS, line 37: Are these 3 options all implemented for IMDS and MSI? If they are, are you missing tests?

lib/msal-node/test/client/ManagedIdentityApplication.spec.ts
```diff
@@ -0,0 +1,326 @@
...
+                expect(
+                    managedIdentityApplication.getConfig.managedIdentityId
+                        .getIdType
+                ).toEqual(ManagedIdentityIdType.SYSTEM_ASSIGNED);
+                expect(
+                    managedIdentityApplication.getConfig.managedIdentityId
+                        .getIsUserAssignedId
+                ).toEqual(false);
+            });
+        });
```
bgavrilMS, line 743: Please add more complex tests, for example: 

1. Request token for resource R1 for SAMI
2. Request token for resource R2 for SAMI
3. Request token for resource R2 for UAMI

Assert how they are in the cache (i.e. 3 separate tokens)

4. Make requests again and assert everything comes from the cache.

These tests do not need to be written for each individual source, as they mostly test the state of the cache.

@neha-bhargava - can you please articulate some more complex unit tests based on your work in MSAL Java  and MSAL .NET?

lib/msal-node/test/test_kit/StringConstants.ts
```diff
@@ -348,6 +350,22 @@ export const AUTHENTICATION_RESULT_DEFAULT_SCOPES = {
...
+    "URI_of_the_managed_identity_assigned_to_the_azure_resource";
+
+export const MANAGED_IDENTITY_CLIENT_ID =
+    "unique_identifier_generated_by_Azure_AD_for_the_azure_resource";
+
+export const DEFAULT_MANAGED_IDENTITY_AUTHENTICATION_RESULT: ManagedIdentityResult =
+    {
+        accessToken: TEST_TOKENS.ACCESS_TOKEN,
+        expiresOn: TEST_TOKEN_LIFETIMES.DEFAULT_EXPIRES_IN,
+        resource: MANAGED_IDENTITY_RESOURCE,
```
bgavrilMS: Is "resource" new? I thought AuthResult only has scopes.

lib/msal-node/test/test_kit/StringConstants.ts
```diff
@@ -348,6 +350,22 @@ export const AUTHENTICATION_RESULT_DEFAULT_SCOPES = {
...
+export const MANAGED_IDENTITY_RESOURCE =
+    "URI_of_the_managed_identity_assigned_to_the_azure_resource";
+
+export const MANAGED_IDENTITY_CLIENT_ID =
+    "unique_identifier_generated_by_Azure_AD_for_the_azure_resource";
+
+export const DEFAULT_MANAGED_IDENTITY_AUTHENTICATION_RESULT: ManagedIdentityResult =
+    {
+        accessToken: TEST_TOKENS.ACCESS_TOKEN,
+        expiresOn: TEST_TOKEN_LIFETIMES.DEFAULT_EXPIRES_IN,
```
bgavrilMS: This must be a bug. `expiresOn` is a date but expires_in is a number (of seconds)

bgavrilMS (changes requested):
- public API needs a review and it needs to  include PM, as there seems to be confusion. Let's get a sing-off from @localden 
- some existing public API like AuthenticationResult must not be modified
- tests do not check cache keys
- more complex tests needed
- skipped tests must be fixed

lib/msal-node/src/config/Configuration.ts
pmaytak: I didn't understand what you meant by "JSON-driven", makes sense it's about config now. Yea, I think it's fine to do a check and throw if multiple mutually exclusive IDs are set. The other option of specifying Type and IdValue is probably less intuitive.

lib/msal-node/src/client/ManagedIdentityApplication.ts
Robbie-Microsoft: No, a managed identity result has been created which will be returned to the developer.

lib/msal-node/src/config/Configuration.ts
Robbie-Microsoft: It's called [Destructuring Assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment). Scroll down to the example with function parameters, like what you're looking at here.

lib/msal-node/src/config/Configuration.ts
Robbie-Microsoft: I've implemented this

lib/msal-node/src/config/ManagedIdentityId.ts
Robbie-Microsoft: I've implemented this

lib/msal-node/src/request/ManagedIdentityRequest.ts
Robbie-Microsoft: yep

lib/msal-node/src/request/ManagedIdentityRequestParams.ts
Robbie-Microsoft: It doesn't matter for a type definition, which this is. In this case, I put them in alphabetical order.

lib/msal-node/src/utils/Constants.ts
Robbie-Microsoft: no, I think I can delete this

lib/msal-node/src/client/ManagedIdentityClient.ts
Robbie-Microsoft: consistency with what?

lib/msal-node/src/client/ManagedIdentitySources/BaseManagedIdentitySource.ts
Robbie-Microsoft: It's an interface with two methods that aren't implemented: sendGetRequestAsync and sendPostRequestAsync. It's in msal-common: lib/msal-common/src/network/INetworkModule.ts

lib/msal-node/src/config/Configuration.ts
Robbie-Microsoft: I based this off existing implementation for Configuration and NodeConfiguration (for existing clients). In this case, Configuration is passed in by the user, and NodeConfiguration is what's sent to the client. Optional parameters not provided by the developers' Configuration are set to default values here.

lib/msal-node/src/config/ManagedIdentityRequestParameters.ts
Robbie-Microsoft: It does a pre-transformation to a network request's body; it turns it into a string. In the case of managed identity, I think forceRefresh will be the only body parameter that could ever be sent.

lib/msal-node/src/request/ManagedIdentityRequestParams.ts
Robbie-Microsoft: I see your confusion. However, ManagedIdentityRequest only exists so that caching functionality from ClientCredentials can be re-used by ManagedIdentity. This functionality requires an authority, correlationId, etc. The  developer can't pass in this information, they can only pass in forceRefresh and resource, which are in ManagedIdentityRequestParameters.

lib/msal-common/src/cache/CacheManager.ts
Robbie-Microsoft: These are the values during Managed Identity Flow:
```
staticAlias = [ 'https://login.microsoftonline.com/common/' ]
environment:  login.microsoftonline.com
entity.environment:  https://login.microsoftonline.com/common/
```

The existing logic did not recognize that there is indeed a match here. The authority is fake, and so is the environment, which is why there is an issue here.

lib/msal-node/test/client/ManagedIdentityApplication.spec.ts
Robbie-Microsoft: no, the one at the top of the file defines OLD_ENVS

lib/msal-node/test/client/ManagedIdentityApplication.spec.ts
Robbie-Microsoft: ManagedIdentityResult. I've updated the code to include the return type

lib/msal-node/test/client/ManagedIdentityApplication.spec.ts
Robbie-Microsoft: no, it defaults to "system_assigned_managed_identity", the same way .net does

lib/msal-node/src/config/ManagedIdentityId.ts
Robbie-Microsoft: It needs to be declared public because the type is checked in AppService.ts and IMDS.ts. However, I can delete getIsUserAssigned because it's only used in the unit tests.

lib/msal-node/src/config/ManagedIdentityId.ts
Robbie-Microsoft: I believe the implementation for each of them is one and the same. I need to write unit tests for each one though.

lib/msal-node/test/test_kit/StringConstants.ts
Robbie-Microsoft: not returning resource anymore since the developer already knows it

lib/msal-node/test/client/ManagedIdentityApplication.spec.ts
Robbie-Microsoft: see other comment

lib/msal-node/test/client/ManagedIdentityApplication.spec.ts
Robbie-Microsoft: I don't think this makes sense because the type is already ManagedIdentityConfiguration

lib/msal-node/test/test_kit/StringConstants.ts
Robbie-Microsoft: I believe expiresOn should be a number. It's returned from the managed identity as a number. See [here](https://learn.microsoft.com/en-us/azure/app-service/overview-managed-identity?tabs=portal%2Chttp). We also return it as a number in lib/msal-node/src/utils/ManagedIdentityUtils.ts

bgavrilMS: Then rename the variable as `expiresIn`?

`expiresOn` - name suggests this is a date
`expiresIn` - name suggestions this is a duration in seconds

When it stores tokens in the cache, MSAL uses dates.

lib/msal-node/test/client/ManagedIdentityApplication.spec.ts
bgavrilMS: Let's review the public API in a meeting.

lib/msal-node/src/config/ManagedIdentityId.ts
bgavrilMS: I don't understand this. We fully control the public API?

bgavrilMS: Could we have `ManagedIdentityId` be only a thin wrapper class for the config JSON and we keep all the extra details such as the IDType oject non-exportable?




```typescript

// This is the public API
export class ManagedIdentityId
{
      public UserAssignedClientId;
      public UserAssignedObjectId;
      public UserAssignedResourceId;
}

// Class that we use in MSAL interally only
class ManagedIdentityIdInternal {

     public readonly type: ManagedIdType 
     constructor(ManagedIdentityId) { 
          type = ... // figure out the type and throw exceptions if incorrect type specified
     }
```     

}

```

Comment (Robbie-Microsoft) [reply lib/msal-node/src/config/ManagedIdentityId.ts]: Even though this class is "exported", it's not available to the developer. In order for an msal-node class or type to be exposed to the developer, it must be exported in lib/msal-node/src/index.ts. In the change-files for this PR, you'll see that only ManagedIdentityApplication and ManagedIdentityRequestParams are exposed (exported).

File: lib/msal-node/test/client/ManagedIdentityApplication.spec.ts (hunk 66 above)
Comment (neha-bhargava) [reply lib/msal-node/test/client/ManagedIdentityApplication.spec.ts:743]: Shared the tests in MSAL .Net for the scenarios: [microsoft-authentication-library-for-dotnet/tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs at main · AzureAD/microsoft-authentication-library-for-dotnet (github.com)](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs).

You can also look at some tests in MSAL Java: https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/712/files#diff-c1e78d16db8b6cb86ef4e37c4f14c7950221b0fd619b2ab501499c588b5c0833.

One note page that contains some raw responses from the server can help with mocking: https://microsoft.sharepoint.com/teams/ADAL/_layouts/OneNote.aspx?id=%2Fteams%2FADAL%2FSiteAssets%2FDevEx%20Notebook&wd=target%28ID4S%2FMSALs%2FTechnical%2FMSI.one%7CA2A57687-9A17-4C9E-86D1-60AA32AB5005%2FMSI%20Errors%20%26%20Exceptions%7C41FE3612-4FAC-475D-9667-4A114AEBEEDA%2F%29

File: lib/msal-node/test/client/ManagedIdentityApplication.spec.ts (hunk 64 above)
Comment (neha-bhargava) [reply lib/msal-node/test/client/ManagedIdentityApplication.spec.ts:745]: Updated the api review doc to add cache keys for all the managed identity scenarios. https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=/ManagedIdentity/ManagedIdentity.md&version=GBdev&_a=preview&anchor=acceptance-tests-for-cache-keys.

Comment (Robbie-Microsoft) [reply lib/msal-node/test/client/ManagedIdentityApplication.spec.ts:211]: I just read Neha's addition to the API review doc regarding the cache keys. I don't think I can write unit tests to verify them... the nodeStorage variable is private and can't be accessed from the unit tests. In Client Credential, we can pass in a custom node storage, and verify cache items that way. We've disabled that for Managed Identity, so I don't think I will be able to have access to the cache to check the cache key.

File: lib/msal-node/test/client/ManagedIdentityApplication.spec.ts (hunk 66 above)
Comment (Robbie-Microsoft) [reply lib/msal-node/test/client/ManagedIdentityApplication.spec.ts:209]: @bgavrilMS, I'm unable to implement that test you suggested. It requires two different managed identity applications, with two different mocked network responses. However, this is impossible, because the Managed Identity Application's ManagedIdentityClient.identitySource is static. Once created, we don't create it again (your feedback, and this means that it persists across Managed Identity Application instances). So whatever the mocked network response is upon creation... it will be unable to be changed for a different resource (client id).

Comment (Robbie-Microsoft) [reply lib/msal-node/test/client/ManagedIdentityApplication.spec.ts:209]: If I comment out the following if statement, then the test I wrote works (based on what you listed above)
```
// if (!ManagedIdentityClient.identitySource) {
        ManagedIdentityClient.identitySource = this.selectManagedIdentitySource(
            this.logger,
            this.nodeStorage,
            this.networkClient,
            this.cryptoProvider
        );
        // }
```

bgavrilMS (approved):
Approved with comments about testing the cache keys, which you are working on

PR #6529, opened 2023-10-01
Build account objects from cached ID Token

### LINKED ISSUE #6228: The account info will be overwritten by another SPA deployed under the same domain
### Core Library

MSAL.js (@azure/msal-browser)

### Core Library Version

2.33.0

### Wrapper Library

Not Applicable

### Wrapper Library Version

None

Public

### Description

Hello MSAL team, 

I have two applications deployed under the same domain:
- app1: https://some-domain.com/main has the `client-id-1`
- app2: https://some-domain.com/app2 has the `client-id-2`

When I sign in to the `app1`, the MSAL library stores account info in the browser's local storage. After that, I go to the `app2`. During the `app2` bootstrap, I manually filter out the `app1`'s account info by the following code:
```typescript
    // calls MSAL public application instance to get all accounts
    const accounts = this.application.getAllAccounts(); 
    // get rid of accounts that have the app1's client id
    const cachedAccounts = accounts.filter(
      (account) => (account.idTokenClaims as IdTokenClaims)?.aud === this.clientId
    );
    // If no app2's account info being found, it opens a popup to sign in
    if (cachedAccounts.length < 1) {
      const loginPopupRequest: PopupRequest = {
        scopes,
      };

      await this.application.loginPopup(loginPopupRequest);
    }
```
The problem is after I sign in with the same user in `app2`, the account info in the local storage will be overwritten by the `app2`'s account info. It seems the MSAL library cannot properly process the same user's account info for different apps under the same domain.
 

### Error Message

_No response_

### Msal Logs

_No response_

### MSAL Configuration

```javascript
{
      auth: {
        authority: 'https://login.microsoftonline.com/organizations',
        clientId: 'client-id',
        postLogoutRedirectUri: 'post-logout-url',
        redirectUri: 'redirect-url',
      },
      cache: {
      },
      system: {
        loggerOptions,
        pollIntervalMilliseconds: 20,
      },
    }
```

### Relevant Code Snippets

```javascript
N/A
```

### Reproduction Steps

1. Create two apps and deploy them under the same domain
2. Sign in to the first app
3. Go to the second app
4. Sign in to the second app with the same user
5. The first app user's account info in the local storage has been overwritten by the second app

### Expected Behavior

Account info for different apps should be separated.

### Identity Provider

Azure AD / MSA

Chrome, Firefox, Edge, Safari

### Regression

_No response_

### Source

Internal (Microsoft)

hectormmg:
This PR:

- Removes ID token claims from cached account entities
- Refactors `getAllAccounts` to only use `getAccountsFilteredBy` in order to remove duplicated logic
- Adds `staticAuthorityOptions` to CacheManager so environment matching can work in cases where `resolveEndpointsAsync` hasn't resolved and there is no authority metadata in memory
- Refactors `Authority.getCloudDiscoveryMetadataFromHardCodedSources` into an exported function and moves it into `AuthorityMetadata` so it can be reused in environment matching
- Fixes #6228


change/@azure-msal-node-695c255d-9b72-4c81-b93a-886e651da57a.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Build account objects from cached ID Token #6455",
```
sameerag:
```suggestion
  "comment": "Build account objects from cached ID Token #6529",
```

lib/msal-common/src/authority/Authority.ts
```diff
@@ -972,24 +1059,6 @@ export class Authority {
         return match;
     }
 
-    /**
-     * Get cloud discovery metadata for common authorities
-     */
-    private getCloudDiscoveryMetadataFromHardcodedValues(): CloudDiscoveryMetadata | null {
```
hectormmg, line 978: Moved to AuthorityMetadata.ts as an exported function and imported at the top

lib/msal-common/src/authority/Authority.ts
```diff
@@ -1046,25 +1115,6 @@ export class Authority {
         };
     }
 
-    /**
-     * Searches instance discovery network response for the entry that contains the host in the aliases list
-     * @param response
-     * @param authority
-     */
```
hectormmg, line 1053: Also moved to AuthoritMetadata.ts

lib/msal-common/src/authority/Authority.ts
```diff
@@ -1184,9 +1234,7 @@ export class Authority {
      * @param authority
      */
     static transformCIAMAuthority(authority: string): string {
-        let ciamAuthority = authority.endsWith(Constants.FORWARD_SLASH)
```
hectormmg, line 1187: Adding forward slash moved to `formatAuthorityUri` so it can be reused and called before passing authority into this method

lib/msal-common/src/authority/AuthorityMetadata.ts
```diff
@@ -951,3 +958,84 @@ for (const key in InstanceDiscoveryMetadata) {
...
+ * Returns aliases for from the raw cloud discovery metadata given in configuration or null if no configuration was provided
+ * @param rawCloudDiscoveryMetadata
+ * @returns
+ */
+export function getAliasesFromConfigMetadata(
+    rawCloudDiscoveryMetadata?: string
+): string[] | null {
+    if (rawCloudDiscoveryMetadata) {
+        try {
+            const metadata = JSON.parse(rawCloudDiscoveryMetadata);
```
tnorling: How many times will this function be called? We'll have to parse the metadata at least one more time when we build the authority object, right?

lib/msal-common/src/authority/AuthorityMetadata.ts
```diff
@@ -951,3 +958,84 @@ for (const key in InstanceDiscoveryMetadata) {
...
+            if (metadata) {
```
tnorling: Do we also need to check if the metadata contains the authority here like you did below?

lib/msal-browser/test/interaction_client/BaseInteractionClient.spec.ts
```diff
@@ -117,11 +154,34 @@ describe("BaseInteractionClient", () => {
...
+                    openIdConfigResponse,
+                    true
+                );
+                return authorityMetadata;
+            });
         });
 
         afterEach(() => {
             window.sessionStorage.clear();
+            jest.clearAllMocks();
```
tnorling: Clear only resets counters and such, restore will reset the function to its original state
```suggestion
            jest.restoreAllMocks();
```

hectormmg: Didn't know that, thanks!

lib/msal-common/src/authority/AuthorityMetadata.ts
hectormmg: Since config is the highest precedence and we now try static sources before in-memory cache this will actually be called every time matchEnvironement is called, so quite a lot. I can move the parsing to when the static metadata gets built and that should make the parsing only happen once per page-load.

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -302,6 +296,8 @@ export abstract class CacheManager implements ICacheManager {
         }
 
         if (!!cacheRecord.account) {
+            // Remove ID token claims before saving account entity
+            cacheRecord.account.idTokenClaims = undefined;
```
tnorling: Why are the claims part of the object in the first place if we're going to remove it?

On a different note, changing the value of object properties here changes it everywhere else as well even outside the context of this function. If we need to do this it's best to spread the account into a new object prior to reassigning the claims

lib/msal-common/src/authority/AuthorityMetadata.ts
hectormmg: TL;DR: It's a bit different but there was indeed a missing step.

There was in fact a step missing (iterating the metadata array to find the object with matching alias). It's a bit different though, the hardcoded metadata object is a KV map with canonical authority as key and the value being the network instance discovery response (tenant_discovery_endpoint + metadata array). So for hardcoded first you check that the canonical authority is one of the keys and then you iterate the value's metadata array and check that the host is one of the aliases to find a match.

For config, our docs demonstrate passing the network response object directly, so the first step can be skipped, but we still have to iterate the object's metadata array to find the one with matching aliases (this was the missing step).

lib/msal-browser/src/cache/BrowserCacheManager.ts
```diff
@@ -431,8 +434,10 @@ export class BrowserCacheManager extends CacheManager {
      */
     setAccount(account: AccountEntity): void {
         this.logger.trace("BrowserCacheManager.setAccount called");
+        // Remove ID token claims before saving account entity
+        const baseAccount = { ...account, idTokenClaims: undefined };
```
hectormmg: Make sure account is cached without ID token claims. This explicit removal is necessary because setAccount can be called outside of the normal browser response handling flow, so this is the only way to make sure the claims aren't cached.

lib/msal-browser/src/cache/BrowserCacheManager.ts
```diff
@@ -1098,94 +1100,6 @@ export class BrowserCacheManager extends CacheManager {
...
-            return true;
-        });
-    }
-
-    /**
-     * Checks the cache for accounts matching loginHint or SID
-     * @param loginHint
-     * @param sid
-     */
-    getAccountInfoByHints(
```
hectormmg, line 1159: This is already supported by `getAccountInfoFilteredBy` in the parent CacheManager

hectormmg: Note: Only the sid match was missing and is implemented in this PR

lib/msal-browser/test/cache/BrowserCacheManager.spec.ts
```diff
@@ -3417,641 +3412,5 @@ describe("BrowserCacheManager tests", () => {
                 refreshToken: ["refreshToken2"],
             });
         });
-
-        describe("getAccountInfoByFilter", () => {
```
hectormmg, line 3421: This and getAccountInfoByHints APIs were removed and their logic is already covered in the tests for getAccountX APIs

lib/msal-browser/src/cache/BrowserCacheManager.ts
hectormmg: Update: Not necessary, went back and made sure setAccount calls are preceded by account entity building that doesn't add ID token claims

lib/msal-common/src/cache/CacheManager.ts
hectormmg: The AccountEntity.createAccount method was adding idTokenClaims to the entity. Went back and made sure that accounts passed into setAccount were always build by methods that didn't include ID token claims and made sure that accounts were returned in auth results with claims added.

lib/msal-common/src/account/AccountInfo.ts
```diff
@@ -41,3 +41,10 @@ export type ActiveAccountFilters = {
     homeAccountId: string;
     localAccountId: string;
 };
+
+export function updateAccountInfo(
+    accountInfo: AccountInfo,
+    updatedAttributes: Partial<AccountInfo>
+): AccountInfo {
+    return { ...accountInfo, ...updatedAttributes };
```
tnorling: This feels basic enough to not need a helper method

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -599,14 +600,24 @@ export class ResponseHandler {
...
+        if (cacheRecord.account && idTokenClaims) {
+            const fullAccountEntity: AccountEntity = Object.assign(
+                new AccountEntity(),
+                {
+                    ...cacheRecord.account,
+                    idTokenClaims: idTokenClaims,
+                }
+            );
+            accountInfo = fullAccountEntity.getAccountInfo();
+        }
```
tnorling: Does this work instead?
```suggestion
            accountInfo = { ...cacheRecord.account.getAccountInfo(), idTokenClaims };
```

tnorling: Believe I saw this in one other place as well

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -599,14 +600,24 @@ export class ResponseHandler {
                 serverTokenResponse?.spa_accountid;
         }
 
+        let accountInfo: AccountInfo | null = null;
+        if (cacheRecord.account && idTokenClaims) {
```
tnorling: What about if cacheRecord.account is truthy but idTokenClaims is not? Looks like this will return null for accountInfo

tnorling (approved):
Last few minor suggestions but otherwise looks good. Approving.

lib/msal-common/src/response/ResponseHandler.ts
tnorling: Ideally the AccountEntity doesn't have idTokenClaims defined on it at all if they won't be cached.

lib/msal-common/src/account/AccountInfo.ts
hectormmg: Yeah originally I thought it'd be more complex, removing.

lib/msal-common/src/response/ResponseHandler.ts
hectormmg: Removing the claims check and letting the spread operator deal with ignoring if undefined.

jo-arroyo (approved):
LGTM!

peterzenz (approved):

PR #6530, opened 2023-10-02
Create devcontainer.json

NahNick:


tnorling:
Description? What's this for, why should we accept this?

PR #6533, opened 2023-10-03
Instrument create and remove hidden iframe & refactor silent handler into exported functions

tnorling:
- Adds instrumentation for loadFrame, loadFrameSync and removeHiddenIframe
- Refactors SilentHandler into static functions
- Refactors monitorIframeForHash to do less work inside setInterval


lib/msal-browser/src/interaction_client/SilentAuthCodeClient.ts
```diff
@@ -113,18 +113,19 @@ export class SilentAuthCodeClient extends StandardInteractionClient {
             this.logger.verbose("Auth code client created");
 
             // Create silent handler
-            const silentHandler = new SilentHandler(
+            const interactionHandler = new InteractionHandler(
```
sameerag, line 116: Update the comment.
nit: Also isn't the name misleading a bit? `silent` should mean `no interaction` right? I know you did not introduce this now but just wondering if changing it is possible.

lib/msal-browser/src/interaction_handler/SilentHandler.ts
```diff
@@ -4,266 +4,210 @@
...
+ * @param timeout
+ */
+export async function monitorIframeForHash(
+    iframe: HTMLIFrameElement,
+    timeout: number,
+    pollIntervalMilliseconds: number,
+    performanceClient: IPerformanceClient,
+    logger: Logger,
+    correlationId: string
+): Promise<string> {
```
sameerag, line 71: If we end up passing `logger`, `correlationId` and `performanceClient` to all functions (I am hoping more of these are slated as you dig through each flow), does it make sense to create a single data structure when we initiate a flow as `RequestMetaData` or something similar to flow through instead of passing them separately?

lib/msal-browser/src/interaction_client/SilentAuthCodeClient.ts
tnorling: What I'm slowly working towards is not having any "Handlers" at all anymore, so this will come in time.

lib/msal-browser/src/interaction_handler/SilentHandler.ts
tnorling: I've been thinking about potentially having the invoke/invokeAsync functions auto-inject these into each function call. Less to pass in manually and keeps the ordering consistent, however, they would still need to be defined in the function definitions. Unfortunately grouping all of these in an object prevents them from being minified, putting us back to where we are today in terms of size.

Still mulling that over but if you have ideas/suggestions would love to discuss

sameerag: How much is the minification cost if it is one instance that flows across a given flow? We do not intend to create new instances of these unlike classes everytime, just at the beginning until we close a given request?

tnorling: It's not a cost per instance it's a cost per reference

tnorling: If we reference a locally scoped `logger` let's say, each reference can be minified to `a`. But if we nest that inside an object `requestMetadata`, to take your example, each reference to `requestMetadata.logger` can only be minified to `a.logger`

hectormmg (approved):
LGTM

jo-arroyo (approved):

Issue #6492, opened 2023-09-18
DOMException: Failed to execute 'atob' on 'Window': The string to be decoded is not correctly encoded.


Michael-006:
### Core Library

MSAL.js (@azure/msal-browser)

### Core Library Version

?

### Wrapper Library

MSAL Angular (@azure/msal-angular)

### Wrapper Library Version

3.0.4

### Public or Confidential Client?

Public

### Description

When having special characters in the displayName like _my tester special charactersčćęèéêëėįìíîï_ msal throws an error "Failed to execute 'atob' on 'Window': The string to be decoded is not correctly encoded." 

After this error I'm getting:

ClientAuthError: token_parsing_error: Token cannot be parsed. Please review stack trace to determine root cause. Failed with error: InvalidCharacterError: Failed to execute 'atob' on 'Window': The string to be decoded is not correctly encoded.

### Error Message

DOMException: Failed to execute 'atob' on 'Window': The string to be decoded is not correctly encoded.

### MSAL Configuration

```javascript
not relevant
```

### Relevant Code Snippets

```javascript
try {
    const oe = S(Q);
    return JSON.parse(oe)
} catch (oe) {
    throw d.e.createTokenParsingError(oe)
}
```

### Reproduction Steps

Set display name to my tester special charactersčćęèéêëėįìíîï

### Expected Behavior

no error, should decode correctly

### Identity Provider

Azure B2C Custom Policy

### Browsers Affected (Select all that apply)

Chrome

### Source

External (Customer)

Mexx77:
we see the same issue with the special character ł in the lastname of the user

jansepke:
possible duplicate of #6476

fredliex:
atob can't process non-latin, ex japanese or chinese characters.
maybe change base64 encode/decode like follow
encode : btoa(unescape(encodeURIComponent(str)))
decode: decodeURIComponent(escape(atob(decodedStr)))

fredliex:
in commit https://github.com/AzureAD/microsoft-authentication-library-for-js/commit/2ef47fe40e817cfd4428ed75dd8dbf11a3f13632  , some code are missing.



https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/b69ed83ede29bd39b04eb71ae7c15f0a983e0442/lib/msal-browser/src/encode/Base64Decode.ts#L24

should be as follows
```csharp
[...]

PR #6535, opened 2023-10-04
Bug fix: id token Base64 decoding

lalimasharda:
Addresses base64 decoding errors reported by customers and #6492

### DIFF
--- a/lib/msal-browser/src/encode/Base64Decode.ts
+++ b/lib/msal-browser/src/encode/Base64Decode.ts
@@ -21,6 +21,19 @@ export function base64Decode(input: string): string {
  * @param base64String
  */
 function base64DecToArr(base64String: string): Uint8Array {
-    const binString = atob(base64String);
+    let encodedString = base64String.replace(/-/g, "+").replace(/_/g, "/");
+    switch (encodedString.length % 4) {
+        case 0:
+            break;
+        case 2:
+            encodedString += "==";
+            break;
+        case 3:
+            encodedString += "=";
+            break;
+        default:
+            throw new Error("Invalid base64 string");
+    }
+    const binString = atob(encodedString);
     return Uint8Array.from(binString, (m) => m.codePointAt(0) || 0);
 }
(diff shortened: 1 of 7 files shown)


lib/msal-browser/src/encode/Base64Decode.ts
```diff
@@ -21,6 +21,19 @@ export function base64Decode(input: string): string {
...
+        case 0:
+            break;
+        case 2:
+            encodedString += "==";
+            break;
+        case 3:
+            encodedString += "=";
+            break;
+        default:
+            throw new Error("Invalid base64 string");
```
tnorling: Can we make this a first class error? e.g. a BrowserAuthError?

lib/msal-browser/test/encode/Base64Decode.spec.ts
```diff
@@ -95,6 +95,25 @@ describe("Base64Decode.ts Unit Tests", () => {
             );
         });
 
+        it("Decode idtokenclaims with special characters", () => {
```
tnorling, line 99: There's a bunch of [localization test cases](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/d1bbf52b5c3ab14e456af076a8b4b1e7c5c024c1/lib/msal-node/test/crypto/CryptoProvider.spec.ts#L49) in msal-node. I'm surprised we don't have these in msal-browser as well. Would be good to get these added here too

tnorling (approved):
Looks great, thanks for addressing this quickly!

lib/msal-browser/test/encode/Base64Decode.spec.ts
sameerag, line 99: Do we already have corresponding encode test?

sameerag (approved):

lib/msal-browser/src/encode/Base64Decode.ts
lalimasharda: Done

lib/msal-browser/test/encode/Base64Decode.spec.ts
lalimasharda: Added one more case in browser localization tests

lalimasharda: Added a test for it in the encoding spec.

PR #6536, opened 2023-10-04
Add support for Multi-tenant accounts and cross-tenant token caching

hectormmg:
This PR:
- Adds `tenantProfiles` map to `AccountInfo` and `tenantProfiles` array (serialization) to `AccountEntity`
- Adds `realm` filtering to `getIdToken` and `getAccessToken` so cached tokens are only matched if they were issued by the tenant in the request's authority
- Updates cached account setting logic to cache `AccountEntity` objects with the available data of the home tenant profile in client_info (localAccountId = uid, realm = utid). 
- Adds logic to update the `AccountInfo` objects that `getAccount` APIs return to reflect the tenant-specific data (tenant profile) in the ID token claims from the ID token that matches the account filter passed in
- Adds logic to `getAllAccounts` to expand and return all tenant profiles that match the filter into full `AccountInfo` objects to maintain backward compatibility
- Sets access token and ID token realm values to the best available value from ID token claims with precedence tid > tfp > acr to cover AAD and B2C cases


lib/msal-browser/src/controllers/StandardController.ts
```diff
@@ -1584,10 +1584,10 @@ export class StandardController implements IController {
     ): string {
         const account =
             request.account ||
-            this.browserStorage.getAccountInfoByHints(
```
hectormmg: Will most likely remove this API once I finish adding tests

lib/msal-common/src/account/AccountInfo.ts
```diff
@@ -35,9 +36,11 @@ export type AccountInfo = {
     };
     nativeAccountId?: string;
     authorityType?: string;
+    tenants?: string[];
 };
 
 export type ActiveAccountFilters = {
     homeAccountId: string;
     localAccountId: string;
+    tenantId: string;
```
tnorling: How does upgrade / rollback work for this addition?

lib/msal-common/src/authority/Authority.ts
```diff
@@ -1314,4 +1314,20 @@ export class Authority {
 
         return ciamAuthority;
     }
+
+    /**
+     * Extract tenantId from authority
+     */
+    static getTenantFromAuthorityString(authority: string): string | undefined {
```
tnorling: Instead of static function can we pull this out of the class? Functions > static functions for minification/treeshaking. You can export it at the bottom of this file or create a new file for "AuthorityUtils" (or pick another name if you're tired of "Utils" ;) )

lib/msal-common/src/account/AccountInfo.ts
hectormmg: Good observation, will probably have to add the extra logic so this gets ignored by lower versions and it modernizes old entries.

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -241,31 +241,117 @@ export abstract class CacheManager implements ICacheManager {
...
+            // Tenant Profile filtering
+            if (tenantProfileFilter) {
+                if (
+                    !!tenantProfileFilter.localAccountId &&
+                    !this.matchLocalAccountId(
+                        idTokenClaims,
+                        tenantProfileFilter.localAccountId
+                    )
+                ) {
+                    return accountInfo;
```
tnorling: We're returning if the filter _does not_ match? Should this return only if all filters match or no filters are defined?

tnorling: If this is by design - can you add some comments explaining what's happening here?

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -387,13 +467,6 @@ export abstract class CacheManager implements ICacheManager {
                 return;
             }
 
-            if (
-                !!accountFilter.localAccountId &&
```
tnorling, line 450: Just confirming these are being removed because they've been moved somehwere else?

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -852,10 +911,16 @@ export abstract class CacheManager implements ICacheManager {
...
+            const idTokenClaims = extractTokenClaims(
                 cachedIdToken.secret,
                 this.cryptoImpl.base64Decode
             );
+
+            cachedAccount.idTokenClaims = idTokenClaims;
+            cachedAccount.localAccountId =
+                idTokenClaims.oid || cachedAccount.localAccountId;
+            cachedAccount.name = idTokenClaims.name || cachedAccount.name;
+            cachedAccount.realm = idTokenClaims.tid || cachedAccount.realm;
```
tnorling: Why are we overwriting the account properties here? In any case this logic feels weird to me - write contents from idToken or write property from self. 

Makes more sense to only overwrite if contents from idToken are truthy and != to existing contents IMO.

lib/msal-common/src/cache/entities/AccountEntity.ts
```diff
@@ -91,18 +92,29 @@ export class AccountEntity {
...
+            tenants: this.tenants,
         };
     }
 
+    /**
+     * Update attributes that are sourced from ID token claims and vary by tenant profile
+     */
+    updateTenantProfile(idTokenClaims: TokenClaims): void {
+        this.idTokenClaims = idTokenClaims;
+        this.realm = idTokenClaims.tid || Constants.EMPTY_STRING;
```
tnorling:
```suggestion
        this.realm = idTokenClaims.tid || "";
```

lib/msal-common/src/cache/entities/AccountEntity.ts
```diff
@@ -91,18 +92,29 @@ export class AccountEntity {
...
+        this.name = idTokenClaims.name || Constants.EMPTY_STRING;
```
tnorling:
```suggestion
        this.name = idTokenClaims.name || "";
```

lib/msal-common/src/cache/entities/AccountEntity.ts
```diff
@@ -91,18 +92,29 @@ export class AccountEntity {
...
      * Generates account key from interface
      * @param accountInterface
      */
     static generateAccountCacheKey(accountInterface: AccountInfo): string {
+        const homeTenantId = accountInterface.homeAccountId.split(".")[1];
         const accountKey = [
             accountInterface.homeAccountId,
             accountInterface.environment || Constants.EMPTY_STRING,
-            accountInterface.tenantId || Constants.EMPTY_STRING,
+            homeTenantId || accountInterface.tenantId || Constants.EMPTY_STRING,
```
tnorling:
```suggestion
            homeTenantId || accountInterface.tenantId || "",
```

lib/msal-common/src/cache/entities/AccountEntity.ts
```diff
@@ -134,6 +147,19 @@ export class AccountEntity {
...
+        try {
+            if (accountDetails.clientInfo && cryptoObj) {
+                clientInfo = buildClientInfo(
+                    accountDetails.clientInfo,
+                    cryptoObj
+                );
+            }
+        } catch (e) {
+            throw e;
+        }
```
tnorling: Point of wrapping in a try/catch if all you're doing is rethrowing?

lib/msal-common/src/cache/entities/AccountEntity.ts
```diff
@@ -151,12 +177,15 @@ export class AccountEntity {
         account.environment = env;
         // non AAD scenarios can have empty realm
         account.realm =
-            accountDetails.idTokenClaims.tid || Constants.EMPTY_STRING;
+            clientInfo?.utid ||
+            accountDetails.idTokenClaims.tid ||
+            Constants.EMPTY_STRING;
```
tnorling:
```suggestion
            "";
```

lib/msal-common/src/cache/entities/AccountEntity.ts
```diff
@@ -227,31 +256,30 @@ export class AccountEntity {
...
+                    }
+                } catch (e) {}
+            }
+            logger.verbose("No client info in response");
         }
 
         // default to "sub" claim
-        logger.verbose("No client info in response");
-        return accountId;
+        return idTokenClaims?.sub ? idTokenClaims.sub : Constants.EMPTY_STRING;
```
tnorling:
```suggestion
        return idTokenClaims.sub || "";
```

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -518,6 +515,48 @@ export class ResponseHandler {
...
     }
 
+    private setCachedAccount(
+        authority: Authority,
+        serverTokenResponse: ServerAuthorizationTokenResponse,
+        idTokenClaims: TokenClaims,
+        authCodePayload?: AuthorizationCodePayload
+    ): AccountEntity | undefined {
+        this.logger.verbose("setCachedAccount called");
+        const responseTenantId = idTokenClaims.tid || Constants.EMPTY_STRING;
```
tnorling:
```suggestion
        const responseTenantId = idTokenClaims.tid || "";
```

lib/msal-common/src/cache/CacheManager.ts
hectormmg: Changed to use a boolean function to make it clearer that it's checking if the ID token claims don't match the filter.

lib/msal-common/src/cache/CacheManager.ts
hectormmg: Yes, these are now part of the ID token claims fitlering that's done when the tenant profile is build (once the base account info has been matched to the filter)

lib/msal-common/src/cache/CacheManager.ts
hectormmg: Added a method that updates the tenant specific data from the claims

lib/msal-browser/test/interaction_client/NativeInteractionClient.spec.ts
```diff
@@ -764,6 +771,7 @@ describe("NativeInteractionClient Tests", () => {
                 name: ID_TOKEN_CLAIMS.name,
                 idTokenClaims: ID_TOKEN_CLAIMS,
                 nativeAccountId: mockWamResponse.account.id,
+                tenants: [ID_TOKEN_CLAIMS.tid],
```
peterzenz: So much repetition around creating the test account.  Maybe a method to create these for each test?

lib/msal-browser/test/utils/StringConstants.ts
```diff
@@ -78,6 +78,9 @@ export const TEST_TOKENS = {
...
+    // [SuppressMessage("Microsoft.Security", "CS002:SecretInNextLine", Justification="Fake credential used for testing purposes")]
```
peterzenz: Is this a newly flagged thing?  Otherwise I'd expect this to be needed for each of the fake tokens in this file.

lib/msal-common/src/account/AccountInfo.ts
peterzenz: This is a requirement for merging this PR.  Also note that deployments mean that a single client could bounce back and forth repeatedly between old/new versions of MSAL during a rollout.

lib/msal-common/test/test_kit/StringConstants.ts
```diff
@@ -19,6 +19,8 @@ export const TEST_TOKENS = {
...
     IDTOKEN_V2:
         "eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiIsImtpZCI6IjFMVE16YWtpaGlSbGFfOHoyQkVKVlhlV01xbyJ9.eyJ2ZXIiOiIyLjAiLCJpc3MiOiJodHRwczovL2xvZ2luLm1pY3Jvc29mdG9ubGluZS5jb20vOTE4ODA0MGQtNmM2Ny00YzViLWIxMTItMzZhMzA0YjY2ZGFkL3YyLjAiLCJzdWIiOiJBQUFBQUFBQUFBQUFBQUFBQUFBQUFJa3pxRlZyU2FTYUZIeTc4MmJidGFRIiwiYXVkIjoiNmNiMDQwMTgtYTNmNS00NmE3LWI5OTUtOTQwYzc4ZjVhZWYzIiwiZXhwIjoxNTM2MzYxNDExLCJpYXQiOjE1MzYyNzQ3MTEsIm5iZiI6MTUzNjI3NDcxMSwibmFtZSI6IkFiZSBMaW5jb2xuIiwicHJlZmVycmVkX3VzZXJuYW1lIjoiQWJlTGlAbWljcm9zb2Z0LmNvbSIsIm9pZCI6IjAwMDAwMDAwLTAwMDAtMDAwMC02NmYzLTMzMzJlY2E3ZWE4MSIsInRpZCI6IjMzMzgwNDBkLTZjNjctNGM1Yi1iMTEyLTM2YTMwNGI2NmRhZCIsIm5vbmNlIjoiMTIzNTIzIiwiYWlvIjoiRGYyVVZYTDFpeCFsTUNXTVNPSkJjRmF0emNHZnZGR2hqS3Y4cTVnMHg3MzJkUjVNQjVCaXN2R1FPN1lXQnlqZDhpUURMcSFlR2JJRGFreXA1bW5PcmNkcUhlWVNubHRlcFFtUnA2QUlaOGpZIn0=.1AFWW-Ck5nROwSlltm7GzZvDwUkqvhSQpm55TQsmVo9Y59cLhRXpvB8n-55HCr9Z6G_31_UbeUkoz612I2j_Sm9FFShSDDjoaLQr54CreGIJvjtmS3EkK9a7SJBbcpL1MpUtlfygow39tFjY7EVNW9plWUvRrTgVk7lYLprvfzw-CIqw3gHC-T7IK_m_xkr08INERBtaecwhTeN4chPC4W3jdmw_lIxzC48YoQ0dB1L9-ImX98Egypfrlbm0IBL5spFzL6JDZIRRJOu8vecJvj1mq-IUhGt0MacxX8jdxYLP-KUu2d9MbNKpCKJuZ7p8gwTL5B7NlUdh_dmSviPWrw",
+    IDTOKEN_V2_ALT:
```
peterzenz, line 24: No fake cred indicator here?

peterzenz (approved):
Would be great to know there's some rollout/rollback testing done with this to ensure nothing breaks in this scenario.

lib/msal-common/src/account/AccountInfo.ts
hectormmg: I made it optional and added update logic for the cached account entities

lib/msal-browser/src/cache/BrowserCacheManager.ts
```diff
@@ -518,6 +520,15 @@ export class BrowserCacheManager extends CacheManager {
         this.removeAccountKeyFromMap(key);
     }
 
+    /**
+     * Remove account entity from the platform cache if it's outdated
+     * @param accountKey
+     */
+    removeOutdatedAccount(accountKey: string): void {
```
tnorling, line 528: Can you use the existing removeAccount API instead? Should do the same things

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -280,52 +278,127 @@ export abstract class CacheManager implements ICacheManager {
...
+                    idToken: idToken.secret,
+                    idTokenClaims: idTokenClaims,
+                    localAccountId:
+                        idTokenClaims.oid || accountInfo.localAccountId,
+                    name: idTokenClaims.name || accountInfo.name,
+                    tenantId: idTokenClaims.tid || accountInfo.tenantId,
+                };
             }
         }
         return null;
```
tnorling: I'm a little surprised we return null here even if accountInfo is truthy. Do you happen to know why?

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -821,6 +866,32 @@ export abstract class CacheManager implements ICacheManager {
...
+    protected updateOutdatedCachedAccount(
+        accountKey: string,
+        accountEntity: AccountEntity
+    ): AccountEntity {
+        if (accountEntity.isSingleTenant()) {
+            const updatedAccount = Object.assign(new AccountEntity(), {
+                ...accountEntity,
+                tenants: [accountEntity.realm],
+            });
+            this.removeOutdatedAccount(accountKey);
```
tnorling: Do we have to remove then add because the keys won't necessarily be the same? How much work would it be to compare old key to new key and then only remove if they don't match?

lib/msal-node/src/cache/NodeStorage.ts
```diff
@@ -456,6 +459,14 @@ export class NodeStorage extends CacheManager {
         return result;
     }
 
+    /**
+     * Remove account entity from the platform cache if it's outdated
+     * @param accountKey
+     */
+    removeOutdatedAccount(accountKey: string): void {
```
tnorling, line 475: removeAccount should do

lib/msal-browser/src/cache/BrowserCacheManager.ts
hectormmg: Unfortunately that method is async and also removes all tokens related to the account context, I'll try to find a way to factor out the part that removes the account object only.

lib/msal-common/src/cache/CacheManager.ts
hectormmg: The main issue is the tenants array not being populated. If you are come from a previous version and have an account with home tenant ID token, then come to multi-tenant and authenticate with a new tenant, the tenants array would include the new tenant but not the home tenant that was already logged in.

I guess a workaround might be to add the home tenant when new tenants are registered if it's not already in the array?

lib/msal-browser/test/utils/StringConstants.ts
hectormmg: CredScan only flags on the diff, so stuff that was already there doesn't need suppression. I can suppress the others as well though.

lib/msal-common/test/test_kit/StringConstants.ts
hectormmg: Weirdly it's not getting flagged

lib/msal-browser/src/cache/BrowserCacheManager.ts
tnorling: Ah thats right ok. Nvm this can stay.

lib/msal-common/src/cache/CacheManager.ts
tnorling: Understood but writing to a key will overwrite the value's contents with the new value. If the keys are the same you can just do a write and skip the remove.

lib/msal-common/src/cache/CacheManager.ts
hectormmg: The accountInfo object, at the point it's checked for truthiness, only contains the base (cached) account entity data. If there is an ID token that matches the filter, then it gets updated with the ID token data and becomes an actual accoutn info object that represents the account in either the tenant from the filter or the home tenant.

If no ID tokens match the filter, then the tenant profile doesn't actually exist, so we return null. Would you like me to change the method name to make that more obvious?

tnorling: Got it. At least a comment to that effect would be helpful

lib/msal-common/src/cache/CacheManager.ts
hectormmg: Sorry, I forgot to say that doesn't mean the key doesn't matter. The realm in the key is now the home tenant since we only store the home account info now and source tenant-specific data from ID tokens.

 If in a previous version someone logs in with a guest tenant account, their account key will end with -guestTenantId and we now look for -homeTenantId so it may result in cache misses. This would catch those accounts and convert them.

hectormmg: I'll add the key comparison and only remove if they don't match

lib/msal-common/src/cache/CacheManager.ts
hectormmg: Added some comments

lib/msal-common/src/cache/entities/AccountEntity.ts
hectormmg: I thought I might have to deal with the error somehow but not the case, removing try/catch

tnorling (approved):

lib/msal-browser/src/cache/BrowserCacheManager.ts
sameerag: Can we rename it to removeAccountOnly() as a sync function and call it inside `removeAccount`? Can be done later, I think that is viable and removes code duplicity

lib/msal-browser/test/utils/StringConstants.ts
sameerag: + will be nice to add this wherever we are using fake credentials.

lib/msal-common/src/account/TokenClaims.ts
```diff
@@ -68,3 +76,23 @@ export type TokenClaims = {
     tenant_region_scope?: string;
     tenant_region_sub_scope?: string;
 };
+
+/**
+ * Gets tenantId from available ID token claims to set as credential realm with the following precedence:
+ * 1. tid - if the token is acquired from an Azure AD tenant tid will be present
```
sameerag, line 82: Any case where `tid` can be blank for AAD? Is that flagged as a bug?

lib/msal-browser/src/cache/BrowserCacheManager.ts
sameerag: Name is fine. Just reusing is what I want to focus on.

lib/msal-browser/src/cache/TokenCache.ts
```diff
@@ -240,40 +241,43 @@ export class TokenCache implements ITokenCache {
...
-        } else if (authority.authorityType !== undefined && clientInfo) {
-            homeAccountId = AccountEntity.generateHomeAccountId(
-                clientInfo,
-                authority.authorityType,
-                this.logger,
-                this.cryptoObj,
-                idTokenClaims
-            );
-        }
+        if (this.isBrowserEnvironment) {
```
hectormmg, line 244: Moved the check up here before all the work is done instead of checking after building the account and then throwing

lib/msal-browser/test/cache/TestStorageManager.ts
```diff
@@ -187,3 +210,64 @@ export class TestStorageManager extends CacheManager {
         return currentCacheKey;
     }
 }
+
+export function buildAccountFromIdTokenClaims(
```
hectormmg: Hardcoding test accounts and ID tokens increases:
1. LOC (by a lot) -> bloats tests and makes them unreadable
2. Bugs caused by the artificial entities being incomplete, meaning they could not match in some place where they should and it would be missed by tests
3. Bugs caused by a small change in one entity (i.e. Account) that isn't changed in a related entity (i.e. IdToken) that causes them to not match when they should

These methods should help us make sure the entities that use the same value are made programatically from the same source

lib/msal-common/src/account/TokenClaims.ts
hectormmg: @sameerag  Since it's listed in the default payload claims, it shouldn't ever come back blank/undefined for AAD:

https://learn.microsoft.com/en-us/entra/identity-platform/id-token-claims-reference#payload-claims

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -633,3 +646,62 @@ export class ResponseHandler {
         };
     }
 }
+
+export function buildAccountToCache(
```
hectormmg, line 650: This logic is duplicated in `ResposneHandler`, `NativeInteraction` and `TokenCache.loadAccount` so I factored it out into a function that is exported here.

lib/msal-common/test/cache/MockCache.ts
```diff
@@ -51,84 +64,38 @@ export class MockCache {
...
-            username: "John Doe",
-            localAccountId: "object1234",
-            realm: "microsoft",
-            environment: "login.microsoftonline.com",
-            homeAccountId: "uid.utid",
-            authorityType: "MSSTS",
-            clientInfo: "eyJ1aWQiOiJ1aWQiLCAidXRpZCI6InV0aWQifQ==",
-        };
-        const account = CacheManager.toObject(new AccountEntity(), accountData);
+        const account = buildAccountFromIdTokenClaims(ID_TOKEN_CLAIMS, [
```
hectormmg, line 60: Added this helper function (and one for ID tokens) since there's a lot of repetitive account creation across the test suite. I've updated all the account definitions that were touched on this PR but there are more instances that can benefit from this incrementally.

lib/msal-node/src/cache/TokenCache.ts
```diff
@@ -124,7 +124,7 @@ export class TokenCache implements ISerializableTokenCache, ITokenCache {
         let cacheContext;
         try {
             if (this.persistence) {
-                cacheContext = new TokenCacheContext(this, false);
+                cacheContext = new TokenCacheContext(this, true);
```
hectormmg, line 127: This is necessary because `getAllAccounts` is now a read/write operation. Setting the cache context here works differently than expected.

I expected the cacheContext.hasChanged property to be updated automatically upon CRUD on the cache, but it doesn't. Setting the cacheContext like this before and after the operation seems to function as a declaration of whether the cache should be updated after the operation (i.e. passing in false means the cache shouldn't update, passing in true means it should).

samples/e2eTestUtils/src/Deserializer.ts
```diff
@@ -1,191 +0,0 @@
-/*
```
hectormmg, line 1: This class as well as Serializer in the test utils are unecessary, they can be imported from msal-node directly.

lib/msal-common/docs/multi-tenant-accounts.md
```diff
@@ -0,0 +1,184 @@
+# Multi-tenant Support in MSAL JS SDKs
+
+> This is an advanced document regarding the handling of accounts to acquire tokens across tenants in the `msal-browser` and `msal-node` SDKs. For basic account information, please review the [Accounts](./accounts.md) document.
```
peterzenz: State what MTAs are as the subject/topic of the doc rather than saying it's "advanced".  If there is recommended reading on what MTAs are on the public docs, you should link them here too.

lib/msal-node/docs/accounts.md
```diff
@@ -1,14 +1,14 @@
 # Accounts in MSAL Node
 
-> This is the platform-specific Accounts documentation for `msal-node`. For the general documentation of the `AccountInfo` object structure, please visit the `msal-common` [Accounts document](../../msal-common/docs/Accounts.md).
+> This is the platform-specific Accounts documentation for `msal-node`. For the general documentation of the `AccountInfo` object structure, please visit the `msal-common` [Accounts document](../../msal-common/docs/Accounts.md).For documentation relating to multi-tenant accounts, please visit the [Multi-tenant Accounts document](../../msal-common/docs/multi-tenant-accounts.md).
```
peterzenz: nit: space missing ". For documentation..."

lib/msal-browser/src/cache/TokenCache.ts
```diff
@@ -240,40 +241,43 @@ export class TokenCache implements ITokenCache {
...
 
-        const accountEntity = AccountEntity.createAccount(
-            {
+            const cachedAccount = buildAccountToCache(
+                this.storage,
+                authority,
                 homeAccountId,
-                idTokenClaims: idTokenClaims,
+                idTokenClaims,
+                this.cryptoObj,
```
tnorling: Is this only needed for the base64Decode function? Can we pass that function in directly instead?

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -580,14 +586,11 @@ export class NativeInteractionClient extends BaseInteractionClient {
...
-            ? Object.assign(new AccountEntity(), {
-                  ...accountEntity,
-                  idTokenClaims: idTokenClaims,
-              })
-            : accountEntity;
-
-        const accountInfo = fullAccountEntity.getAccountInfo();
+        const accountInfo: AccountInfo | null = updateAccountTenantProfileData(
+            accountEntity.getAccountInfo(),
+            undefined,
```
tnorling: Anywhere you pass in an explicit undefined it would be good to include a comment stating what that's supposed to represent

lib/msal-browser/test/cache/TestStorageManager.ts
```diff
@@ -33,6 +38,25 @@ export class TestStorageManager extends CacheManager {
         return null;
     }
 
+    removeAccountKeyFromMap(key: string): void {
+        const currentAccounts = this.getAccountKeys();
+        const removalIndex = currentAccounts.indexOf(key);
+        if (removalIndex > -1) {
+            currentAccounts.splice(removalIndex, 1);
+            this.store[ACCOUNT_KEYS] = currentAccounts;
+        }
```
tnorling: A bit more efficient:
```suggestion
            this.store[ACCOUNT_KEYS] = currentAccounts.filter(entry => entry !== key);
```

tnorling: Ah, missed the fact that this is for tests - not super important then. Up to you

hectormmg: Honestly our tests can also use a perf boost. This is minor but gradual improvements are the way to go, thanks!

samples/msal-browser-samples/VanillaJSTestApp2.0/app/multi-tenant/authConfig.js
```diff
@@ -0,0 +1,67 @@
+const homeTenant = "72f988bf-86f1-41af-91ab-2d7cd011db47";
```
peterzenz: Do we want the hardcoded tenants and clientIds in here?

peterzenz (approved):

lib/msal-common/docs/multi-tenant-accounts.md
hectormmg: I haven't found any public-facing docs that talk about multi-tenant accounts specifically, if I find some I'll add them here.

samples/msal-browser-samples/VanillaJSTestApp2.0/app/multi-tenant/authConfig.js
hectormmg: I'm actually removing this sample, after talking to @tnorling it's better to add this scenario to the customizable-e2e-sample and add some tests while I'm at it.

lib/msal-common/docs/multi-tenant-accounts.md
```diff
@@ -0,0 +1,184 @@
...
+
+`AccountInfo` objects returned from `acquireToken` and `getAccount` APIs contain a `Map<string, TenantProfile>` object called `tenantProfiles` where the key is the tenant ID and the value is the `TenantProfile` for that account in that tenant.
+
+MSAL uses these `TenantProfile` objects to match and build the tenant-specific `AccountInfo` objects required through all of MSAL's flows. They can also be used by client applications for different purposes such as facilitating account selection logic and displaying account data for the user across tenants.
+
+> Warning: While MSAL can return a tenant-specific `AccountInfo` object for each tenant profile, tenant profiles are not actually different accounts, they are only representations of the same account in the different tenants the account has authentiated with. **If a user uses different accounts to authenticate to each tenant, these will not be linked as tenant profiles of eachother and will be treated as completely different accounts that cannot be used to access eachother's tokens**.
```
tnorling: A couple typos:

```suggestion
> Warning: While MSAL can return a tenant-specific `AccountInfo` object for each tenant profile, tenant profiles are not actually different accounts, they are only representations of the same account in the different tenants the account has authenticated with. **If a user uses different accounts to authenticate to each tenant, these will not be linked as tenant profiles of each other and will be treated as completely different accounts that cannot be used to access each other's tokens**.
```

lib/msal-common/docs/multi-tenant-accounts.md
```diff
@@ -0,0 +1,184 @@
...
+    auth: {
+        clientId: "ENTER_CLIENT_ID",
+        authority: "https://login.microsoftonline.com/HOME_TENANT", // This is the authority that MSAL will default to for requests that don't specify their own authority.
+    },
+};
+
+const myMSALObj = new PublicClientApplication(msalConfig);
+
+// handleRedirectPromise has to be called to resume redirect requests
+myMSALObj
```
tnorling: initialize?

lib/msal-common/docs/multi-tenant-accounts.md
```diff
@@ -0,0 +1,184 @@
...
+// Get the guest tenant account from the AuthenticationResult
+const guestTenantAccount = guestTenantAuthResponse.account;
+```
+
+Assuming the user logs into both tenants with the same account, once both login requests are successfully completed and the user has authenticated with each tenant, the cache will contain:
+
+-   2 ID tokens (one per tenant)
+-   2 Access tokens (one per tenant)
+-   1 Refresh token (one per account, shared across tenants)
+-   1 AccountEntity containing both tenant profiles. The `homeTenantAccount` and `guestTenantAccount` objects shown above are both created from the same cached account entity augmented by the claims in their respective ID tokens.
```
tnorling: Cache details are technically implementation details and we don't actually guarantee this (especially in the case of RT when we start talking about any sort of brokering). Perhaps better to rework this to simply state that ID and Access tokens are tenant specific but RTs are not and skip cache discussion.

lib/msal-common/docs/multi-tenant-accounts.md
```diff
@@ -0,0 +1,184 @@
...
+.
+// Acquire a home tenant access token silently
+const homeTenantSilentResponse = await myMSALObj.acquireTokenSilent({
+    account: homeTenantAccount,
+    ...homeTenantRequest,
+});
+
+
+// Original interactive call to guest tenant
+const guestTenantAuthResponse = await myMSALObj.loginPopup(homeTenantRequest);
```
tnorling: Do you actually need an interactive call for each separate tenant or can you go straight to ATS since the RTs are shared?

tnorling: Assuming this was needed due to not having the accountinfo object yet for the tenant you're requesting. I'd suggest instead that if you don't have the tenanted account yet folks can invoke ATS with the home account or account hint, tenanted authority & forceRefresh or CachePolicy to silently get the guest tenant tokens

lib/msal-common/docs/multi-tenant-accounts.md
```diff
@@ -0,0 +1,184 @@
...
+// Get home tenant token
+const homeTenantAuthResponse = await myMSALObj.acquireTokenSilent({
+    account: homeTenantAccount,
+    ...homeTenantRequest,
+});
+```
+
+### getAllAccounts with Multiple Tenants
+
+When any account in the MSAL cache contains multiple tenant profiles, `getAllAccounts()` is expected to return a full `AccountInfo` object for each tenant profile that satisfies the filter provided. If no filter is provided, every tenant profile for every account will be returned as a full `AccountInfo` object by default. This means that even if there is only one account object in the cache, multiple `AccountInfo` objects may be returned from `getAllAccounts`.
```
tnorling: Again, avoid discussion around cache details

lib/msal-common/src/account/AccountInfo.ts
```diff
@@ -35,9 +36,100 @@ export type AccountInfo = {
     };
     nativeAccountId?: string;
     authorityType?: string;
+    tenantProfiles?: Map<string, TenantProfile>;
+};
+
+/**
+ * Account details that vary across tenants for the same user
+ *
+ * - tenantId               - Full tenant or organizational id that this account belongs to
```
tnorling: Doc comments for params should go above each individual field. You can leave the general description of the type here.

lib/msal-common/src/account/AccountInfo.ts
```diff
@@ -35,9 +36,100 @@ export type AccountInfo = {
...
+export function buildTenantProfileFromIdTokenClaims(
+    homeAccountId: string,
+    idTokenClaims: TokenClaims
+): TenantProfile {
+    const { oid, sub, tid, name, tfp, acr } = idTokenClaims;
+
+    const tenantId = tid || tfp || acr || "";
+
+    return {
+        tenantId: tenantId.toLowerCase(),
```
tnorling: Why is this lowercased here? Should probably leave this and just lowercase for comparison

lib/msal-common/src/account/AccountInfo.ts
```diff
@@ -35,9 +36,100 @@ export type AccountInfo = {
...
+    );
+}
+
+export function buildTenantProfileFromIdTokenClaims(
+    homeAccountId: string,
+    idTokenClaims: TokenClaims
+): TenantProfile {
+    const { oid, sub, tid, name, tfp, acr } = idTokenClaims;
+
+    const tenantId = tid || tfp || acr || "";
```
tnorling, line 90: Please add a comment explaining the meaning of these claims / reasoning behind this precedence

lib/msal-common/src/account/TokenClaims.ts
```diff
@@ -68,3 +76,23 @@ export type TokenClaims = {
...
+ * @param idTokenClaims
+ * @returns
+ */
+export function getTenantIdFromIdTokenClaims(
+    idTokenClaims?: TokenClaims
+): string | null {
+    if (idTokenClaims) {
+        const tenantId =
+            idTokenClaims.tid || idTokenClaims.tfp || idTokenClaims.acr;
+        return tenantId?.toLowerCase() || null;
```
tnorling: Lowercase should be used for comparison only

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -285,52 +297,226 @@ export abstract class CacheManager implements ICacheManager {
...
+        });
+    }
+
+    private getAccountInfoForTenantProfiles(
         accountEntity: AccountEntity,
         accountFilter?: AccountFilter
+    ): AccountInfo[] {
+        const tenantProfileFilter: TenantProfileFilter | undefined =
+            accountFilter
+                ? {
```
tnorling: reason you can't simply use accountFilter directly here?

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -826,6 +1009,88 @@ export abstract class CacheManager implements ICacheManager {
...
+                        name: account.name,
+                        isHomeTenant: tenantIdMatchesHomeTenant(
+                            account.realm,
+                            account.homeAccountId
+                        ),
+                    };
+                }
+            );
+
+            const updatedAccount = Object.assign(new AccountEntity(), {
```
tnorling: I thought we had a helper for this `CacheManager.toObject`. Not super critical since I plan to refactor AccountEntity into a type after this is merged

lib/msal-common/src/cache/entities/AccountEntity.ts
```diff
@@ -53,15 +60,15 @@ export class AccountEntity {
...
     nativeAccountId?: string;
+    tenantProfiles?: Array<TenantProfile>;
 
     /**
      * Generate Account Id key component as per the schema: <home_account_id>-<environment>
      */
     generateAccountId(): string {
         const accountId: Array<string> = [this.homeAccountId, this.environment];
-        return accountId.join(Separators.CACHE_KEY_SEPARATOR).toLowerCase();
+        return accountId.join("-").toLowerCase();
```
tnorling: Let's continue to use the constant here since it does have a value - we'll refactor the constants later for the size gains

lib/msal-common/src/cache/entities/AccountEntity.ts
```diff
@@ -88,24 +95,37 @@ export class AccountEntity {
...
         const accountKey = [
             accountInterface.homeAccountId,
-            accountInterface.environment || Constants.EMPTY_STRING,
-            accountInterface.tenantId || Constants.EMPTY_STRING,
+            accountInterface.environment || "",
+            homeTenantId || accountInterface.tenantId || "",
         ];
 
-        return accountKey.join(Separators.CACHE_KEY_SEPARATOR).toLowerCase();
+        return accountKey.join("-").toLowerCase();
```
tnorling: Same comment here around constant usage

lib/msal-common/src/cache/entities/AccountEntity.ts
```diff
@@ -224,31 +276,30 @@ export class AccountEntity {
...
+                    const clientInfo = buildClientInfo(
+                        serverClientInfo,
+                        cryptoObj.base64Decode
+                    );
+                    if (clientInfo.uid && clientInfo.utid) {
+                        return `${clientInfo.uid}.${clientInfo.utid}`;
+                    }
+                } catch (e) {}
+            }
+            logger.verbose("No client info in response");
```
tnorling: This should probable be a warning

lib/msal-node/src/cache/NodeStorage.ts
```diff
@@ -218,13 +218,22 @@ export class NodeStorage extends CacheManager {
...
         return null;
     }
 
+    /**
+     * Reads account from cache, builds it into an account entity and returns it.
+     * @param accountKey
+     * @returns
+     */
+    getCachedAccountEntity(accountKey: string): AccountEntity | null {
+        return Object.assign(new AccountEntity(), this.getItem(accountKey));
```
tnorling: Will this throw if getItem returns undefined/null?

lib/msal-common/docs/multi-tenant-accounts.md
hectormmg: Reworked the sample code to handle this better using a custom function that checks if the tenant profile is already in the cache and if not it uses acquireTokenSilent with the `CacheLookupPolicy.RefreshToken` with the guest tenant authority.

FWIW, `ssoSilent` can also be used but I didn't want to add complexity since the doc is already long and verbose and that's msal-browser specific. I'll add a section to the login docs or a special doc for it soon.

lib/msal-common/docs/multi-tenant-accounts.md
hectormmg: Removed implementation details re: caching logic.

lib/msal-common/src/cache/CacheManager.ts
hectormmg: I think it was useful in a previous iteration of this function, but it ended up being an unnecessary refinement. Still keeping the TenantProfileFilter type in the downstream function the account filter gets passed into so there's a type limit on which properties are used for matching in that filtering step.

lib/msal-common/src/cache/CacheManager.ts
hectormmg: Right, these were used interchangeably, and I think they're functionally equivalent. Not sure why we have CacheManager.toObject, but since this is in fact inside CacheManager I'll use that for now for consistency.

lib/msal-node/src/cache/NodeStorage.ts
hectormmg: It won't throw (Object.assign handles it gracefully) but this is actually a bug since it'll return an empty AccountEntity when it should be returning null. This is a great catch, thanks!

lib/msal-common/src/cache/entities/AccountEntity.ts
hectormmg: Updated

lib/msal-common/src/account/AccountInfo.ts
hectormmg: Done

lib/msal-browser/src/cache/TokenCache.ts
```diff
@@ -268,7 +268,7 @@ export class TokenCache implements ITokenCache {
                 authority,
                 homeAccountId,
                 idTokenClaims,
-                this.cryptoObj,
+                this.cryptoObj.base64Decode,
```
tnorling: pass in the base64Decode function from ./encode/base64Decode.ts instead (for better minification)

lib/msal-common/docs/multi-tenant-accounts.md
```diff
@@ -172,10 +230,11 @@ if (guestTenantProfile) {
             }
         });
 } else {
-    // Authenticate with guest tenant first
-    const guestTenantAuthResponse = await myMSALObj.loginPopup(
-        homeTenantRequest
-    );
+    // If the tenant profile isn't found in the account, that means the user hasn't authenticated with that tenant. This is the custom getToken function from the first example.
+    const guestTenantAuthResponse = await myMSALObj.getToken({
```
tnorling: getToken?

lib/msal-common/src/account/AccountInfo.ts
```diff
@@ -83,10 +81,16 @@ export function buildTenantProfileFromIdTokenClaims(
 ): TenantProfile {
     const { oid, sub, tid, name, tfp, acr } = idTokenClaims;
 
+    /**
+     * Since there is no way to determine if the authority is AAD or B2C, we exhaust all the possible claims that can serve as tenant ID with the following precedence:
+     * tid - TenantID claim that identifies the tenant that issued the token in AAD. Expected in all AAD ID tokens, not present in B2C ID Tokens.
+     * tfp - Trust Framework Policy claim that identifies the policy that was used to authenticate the user. Functions as tenant for B2C scenarios.
+     * acr - Authentication Context Class Reference claim used only with older B2C policies. Fallback in case tfp is not present, but likely won't be present anyway.
```
tnorling, line 88: I think acr is not B2C specific. Can be returned by other authorities but not sure under what conditions

tnorling (approved):

lib/msal-common/docs/multi-tenant-accounts.md
hectormmg: Updated, thanks!

PR #6550, opened 2023-10-06
Implement preconnect to speed up /token calls, refactor static class methods to exported methods for better minification

tnorling:
Adds a [preconnect](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preconnect) link element to the document header in flows where we're reasonably sure we will hit the /token endpoint (acquireTokenPopup, ssoSilent, acquireTokenSilent after RT expiration). This tells the browser to start resolving DNS and establishing the SSL connection so that when the /token call is made it doesn't need to wait for these steps to complete (can take up to 300ms). This is a short lived connection and the browser will kill it if not used promptly so I've included a cleanup function on a timer (10s)

Also de-classed BrowserUtils for size reduction while I was adding a new function there anyway.

### DIFF
--- a/lib/msal-browser/src/utils/BrowserUtils.ts
+++ b/lib/msal-browser/src/utils/BrowserUtils.ts
@@ -3,7 +3,7 @@
  * Licensed under the MIT License.
  */
 
-import { Constants, UrlString } from "@azure/msal-common";
+import { UrlString } from "@azure/msal-common";
 import {
     createBrowserAuthError,
     BrowserAuthErrorCodes,
@@ -15,12 +15,12 @@ import { InteractionType, BrowserConstants } from "./BrowserConstants";
  */
 export function clearHash(contentWindow: Window): void {
     // Office.js sets history.replaceState to null
-    contentWindow.location.hash = Constants.EMPTY_STRING;
+    contentWindow.location.hash = "";
     if (typeof contentWindow.history.replaceState === "function") {
         // Full removes "#" from url
         contentWindow.history.replaceState(
             null,
-            Constants.EMPTY_STRING,
+            "",
             `${contentWindow.location.origin}${contentWindow.location.pathname}${contentWindow.location.search}`
         );
     }
@@ -32,8 +32,7 @@ export function clearHash(contentWindow: Window): void {
 export function replaceHash(url: string): void {
     const urlParts = url.split("#");
     urlParts.shift(); // Remove part before the hash
-    window.location.hash =
-        urlParts.length > 0 ? urlParts.join("#") : Constants.EMPTY_STRING;
+    window.location.hash = urlParts.length > 0 ? urlParts.join("#") : "";
 }
 
 /**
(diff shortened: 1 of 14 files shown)


lib/msal-browser/src/utils/BrowserUtils.ts
```diff
@@ -11,146 +11,155 @@ import {
...
+        );
     }
 }
+
+/**
+ * Adds a preconnect link element to the header which begins DNS resolution and SSL connection in anticipation of the /token request
+ * @param loginDomain Authority domain, including https protocol e.g. https://login.microsoftonline.com
+ * @returns
+ */
+export function preconnect(authority: string): void {
```
tnorling, line 151: This is the new function in this file. The rest are simply changed from static class methods to exported methods with unchanged implementations

lib/msal-browser/src/utils/BrowserUtils.ts
```diff
@@ -11,146 +11,155 @@ import {
...
+    const link = document.createElement("link");
+    link.rel = "preconnect";
+    link.href = new URL(authority).origin;
+    link.crossOrigin = "anonymous";
```
sameerag, line 155: because l.mso origin is different from the SPA url?

sameerag (approved):
Looks good. Did we test how much gain from this?

lib/msal-browser/src/utils/BrowserUtils.ts
tnorling: That's right

tnorling:
> Looks good. Did we test how much gain from this?

In limited testing I observed the connection steps can take up to 300ms, though tend to be somewhere between 100-200ms. It's a really small sample size so I'm taking it with a big grain of salt but we'll see what our telemetry tells us after we push this change

lib/msal-browser/src/utils/BrowserUtils.ts
peterzenz: Add this to the PR description as well.  "Implement preconnect to speed up /token calls, refactor static class methods to exported methods for better minification."  Or something like that.

lib/msal-browser/src/utils/BrowserUtils.ts
```diff
@@ -3,154 +3,162 @@
...
+    document.head.appendChild(link);
+
+    // The browser will close connection if not used within a few seconds, remove element from the header after 10s
```
peterzenz, line 158: Does the 10s timeout align with the browser implementation?  I'm not sure if the browser were to close after 3 seconds, would this implementation open then close several times before the preconnect is removed?

peterzenz (approved):

lib/msal-browser/src/utils/BrowserUtils.ts
tnorling: If I'm parsing this question correctly - The connection to the STS is opened when this preconnect element is added to the page which we only do once per request. If the browser chooses to close the connection after 3 seconds it will stay closed until another preconnect element is added (another ssoSilent request is made) so this 10s timeout is just cleaning up dead elements.

Also of note - if a connection already exists to a given domain when a preconnect element is added it is a no-op.

PR #6565, opened 2023-10-13
Rollup bug for un-minified UMD output with regard to constants

shoatman:
* Minified works likely because of Terser doing a better job of understanding what's going on
* For reasons only known to Rollup.... it seems to get confused when it starts resolve module dependencies.... when the same constant is referenced from 2 distinct entry points.

instead of normal output:

const noAccountFound = "no_account_found";

You see:

const index.noAccountFound = "no_account_found" 

This PR is a temporary fix?? Until we have time to understand rollup better and propose a PR to them.  OR until it gets fixed in a future release of rollup.


hectormmg (approved):
Worked on my machine, approving

shoatman:
Providing a better fix.

PR #6567, opened 2023-10-16
addressing rollup issue based when using 2 outputs for one input

shoatman:
I saw random discussion about a problem with trying to output both cjs and umd at the same time.  I split and the issues with the UMD output went away.  (no longer prefixing inlined dyanamic imports with "index."

I confirmed the UMD output no longer is prefixed.... let's see how the various builds do now.


tnorling (approved):
Changefiles?

shoatman:
Yep.  I'll run the beachball command.

hectormmg (approved):

PR #6580, opened 2023-10-19
Refactor token cache entities into types

tnorling:
Refactors `CredentialEntity`, `IdTokenEntity`, `AccessTokenEntity`, and `RefreshTokenEntity` to be Types rather than a Class and moves static class methods into functions exported to the `CacheHelpers` namespace.

Making these types and separating the class methods from the type definition allows us to read from the cache and directly use the value without needing to first copy each key/value pair into an instance of the class (the `toObject` helper function). Doing it this way also results in a small bundle size improvement.

Reviews can be focused on the `msal-common/src/cache` folder. The rest of the changes are repetitively changing references to the affected functions


lib/msal-common/src/cache/entities/AccessTokenEntity.ts
```diff
@@ -4,164 +4,28 @@
...
- *      familyId: Family ID identifier, usually only used for refresh tokens
- *      realm: Full tenant or organizational identifier that the account belongs to
- *      target: Permissions that are included in the token, or for refresh tokens, the resource identifier.
- *      cachedAt: Absolute device time when entry was created in the cache.
- *      expiresOn: Token expiry time, calculated based on current UTC time in seconds. Represented as a string.
- *      extendedExpiresOn: Additional extended expiry time until when token is valid in case of server-side outage. Represented as string in UTC seconds.
- *      keyId: used for POP and SSH tokenTypes
- *      tokenType: Type of the token issued. Usually "Bearer"
- * }
+ * Access token cache type
```
sameerag, line 10: Can we please retain the schema in the comments? Until we actually change the cache, we may need to keep the structure

tnorling: The schema is still documented, I only moved the comments to be above each field rather than all together in this big block. It allows Typedocs/Intellisense to accurately describe each field and is easier for maintenance

hectormmg (approved):
Looks good. Is there a reason why we're not doing the same for AccountEntity?

tnorling:
> Looks good. Is there a reason why we're not doing the same for AccountEntity?

I will do the other entities in separate PR(s). Was trying to keep this as small as possible, but it ended up pretty big anyway :/

lib/msal-common/src/cache/entities/AccessTokenEntity.ts
sameerag: I meant literally these lines, helps gauge structure of the entiry
 ```
 * Key:Value Schema:
 *
 * Key Example: uid.utid-login.microsoftonline.com-accesstoken-clientId-contoso.com-user.read
 ```

lib/msal-common/src/cache/entities/CredentialEntity.ts
```diff
@@ -3,218 +3,34 @@
...
     familyId?: string;
+    /** Full tenant or organizational identifier that the account belongs to */
     realm?: string;
+    /** Permissions that are included in the token, or for refresh tokens, the resource identifier. */
     target?: string;
+    /** Matches the SHA 256 hash of the obo_assertion for the OBO flow */
     userAssertionHash?: string;
+    /** Matches the authentication scheme for which the token was issued (i.e. Bearer or pop) */
     tokenType?: AuthenticationScheme;
+    /** KeyId for PoP tokens stored in the kid claim */
```
sameerag: nit: SSH also?

sameerag (approved):
Some small comments.

lib/msal-common/src/cache/entities/AccessTokenEntity.ts
tnorling: Gotcha, I'll add this back to the key generation function. Does that work for you?

lib/msal-common/src/cache/entities/CredentialEntity.ts
tnorling: done

lib/msal-common/src/cache/entities/AccessTokenEntity.ts
sameerag: Yes.

PR #6583, opened 2023-10-20
Update contributing.md

trwalke:
Update to build steps

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -94,7 +94,7 @@ npm run test:coverage
 
 #### Issues with Jest?
 ```bash
-//if you find that there are errors with the jest apis, you can install "jest" extension 
+//if you find that there are errors with the jest apis, and you are on VS Code, you can install "jest" extension 
 //from the market and run the folowing command
 npm i --save-dev @types/jest
 ``````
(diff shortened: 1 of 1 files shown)


contributing.md
```diff
@@ -56,6 +56,8 @@ cd lib/<package-name>
 npm run build:all
 // To run build only for a single package (not its dependencies)
 npm run build
+//If you run into an issue with shx, run the following
+npm install shx
```
tnorling, line 60: If there's issues related to dependencies not being installed let's make sure they are included in the dependency list rather than requiring one-off installs

sameerag: +1. Do we know why this is needed or rather when we run into issues without this install?

Robbie-Microsoft: I think maybe this should have been a global install? This is the first time Travis is using npm on his machine so it makes that it wouldn't be installed yet.

Robbie-Microsoft: Nevermind, shx didn't show up in my global install list. @trwalke what version of NodeJS and npm are you using?

tnorling: Should be installed as a devDependency in each lib. If it's not there and it's causing issues, please add it.

trwalke: @Robbie-Microsoft v18.18.2, the latest LTS version. Lets sync up on this so I can add it to the dependencies

PR #6586, opened 2023-10-20
Don't use temporary cache for silent & popup flows

tnorling:
Silent and Popup flows don't need to use temporary cache because they don't lose context the way the Redirect flow does. Applying this across the board causes unnecessary stringification, storage, lookup and parsing in those flows, especially since the code isn't shared across those flows anyway.

This PR refactors the popup and silent flows to pass the request object through the stack instead of storing and looking up in sessionStorage. The redirect flow will continue to utilize sessionStorage for temp cache.

### DIFF
--- a/lib/msal-browser/src/interaction_client/PopupClient.ts
+++ b/lib/msal-browser/src/interaction_client/PopupClient.ts
@@ -43,14 +43,13 @@ import { EventHandler } from "../event/EventHandler";
 import { BrowserCacheManager } from "../cache/BrowserCacheManager";
 import { BrowserConfiguration } from "../config/Configuration";
 import {
-    InteractionHandler,
-    InteractionParams,
+    InteractionHandler
 } from "../interaction_handler/InteractionHandler";
 import { PopupWindowAttributes } from "../request/PopupWindowAttributes";
 import { EventError } from "../event/EventMessage";
 import { AuthenticationResult } from "../response/AuthenticationResult";
 
-export type PopupParams = InteractionParams & {
+export type PopupParams = {
     popup?: Window | null;
     popupName: string;
     popupWindowAttributes: PopupWindowAttributes;
@@ -212,13 +211,6 @@ export class PopupClient extends StandardInteractionClient {
             InteractionType.Popup
         );
         BrowserUtils.preconnect(validRequest.authority);
-        this.browserStorage.updateCacheEntries(
-            validRequest.state,
-            validRequest.nonce,
-            validRequest.authority,
-            validRequest.loginHint || Constants.EMPTY_STRING,
-            validRequest.account || null
-        );
 
         try {
             // Create auth code request and generate PKCE params
@@ -295,11 +287,6 @@ export class PopupClient extends StandardInteractionClient {
             // Deserialize hash fragment response parameters.
             const serverParams: ServerAuthorizationCodeResponse =
                 UrlString.getDeserializedHash(hash);
-            const state = this.validateAndExtractStateFromHash(
-                serverParams,
-                InteractionType.Popup,
-                validRequest.correlationId
-            );
             // Remove throttle if it exists
             ThrottlingUtils.removeThrottle(
                 this.browserStorage,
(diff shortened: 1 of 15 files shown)


lib/msal-browser/src/interaction_handler/InteractionHandler.ts
```diff
@@ -122,65 +104,49 @@ export class InteractionHandler {
...
-                authority,
-                networkModule
+                request.correlationId
             );
         }
 
         // Nonce validation not needed when redirect not involved (e.g. hybrid spa, renewing token via rt)
         if (validateNonce) {
-            authCodeResponse.nonce = cachedNonce || undefined;
+            // TODO: Assigning "response nonce" to "request nonce" is confusing. Refactor the function doing validation to accept request nonce directly
```
peterzenz, line 134: Agreed.  Please file work items to track ToDo's.

peterzenz (approved):

lib/msal-browser/src/interaction_client/PopupClient.ts
```diff
@@ -340,25 +327,19 @@ export class PopupClient extends StandardInteractionClient {
                 );
                 const { userRequestState } = ProtocolUtils.parseRequestState(
                     this.browserCrypto,
-                    state
+                    validRequest.state
```
sameerag, line 328: Isn't state validation expected?

sameerag: We do not need to cache but we need to check state between request and response, may be you are doing it downstream, just want to verify this.

tnorling: I assume you're asking about the removal of `validateAndExtractStateFromHash` above, which was perhaps not accurately named. That function is not doing the comparison of request to response (that part is done later). Instead this is extracting the state from the hash, parsing it and checking if the interactionType is equal to the interactionType of the API we're using (this check is to prevent handleRedirectPromise from handling a silent or popup response inside the popup or iframe but is not necessary for popup/iframe response handling).

tnorling: I'm also not actually sure state request==response validation does need to happen in popup or silent calls since those are closed loops (someone previously called this out in comments in the code as well). I've left that step in for now but might be worth revisiting later.

sameerag: https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-30#section-10.12 - I think we always did it for CSRF protection as per the spec when we get the `auth-code` via STS redirect.

sameerag (commented):
Another note: Make sure the memoization is enforced for similar requests, via thumbprint and not through cache. We do that today I assume? In case you haven't verified, please confirm the same and we are good.

lib/msal-browser/src/interaction_handler/InteractionHandler.ts
```diff
@@ -190,62 +156,29 @@ export class InteractionHandler {
...
             this.performanceClient,
-            this.authCodeRequest.correlationId
+            request.correlationId
         )(this.authCodeRequest, authCodeResponse)) as AuthenticationResult;
-        this.browserStorage.cleanRequestByState(state);
         return tokenResponse;
     }
 
     /**
-     * Updates authority based on cloudInstanceHostname
```
sameerag, line 200: Curious: Why were we making an extra network call for instance, are we assuming authority changes? Hopefully the multi-tenant feature should cover this.

sameerag: Ah! never mind. I see we need this when we make the request which you moved below. I am interested to check if we can avoid this when applicable, as tenant-id switch should not change anything with cloud discovery.

sameerag (approved):
Approved, some minor comments reg existing behavior.

lib/msal-browser/src/interaction_client/PopupClient.ts
tnorling: Right but that's only applicable to the _redirect_ flow (specifically handleRedirectPromise) because it's not a closed loop. The vulnerability is that you click on a link on site A which takes you to site B with auth code or tokens in the query string or hash which and site B logs you in despite the fact that site B never requested a login. That vulnerability isn't exploitable with popup and silent flows because we (MSAL.js) own the end to end of opening the popup or iframe and closing it when we're done.

tnorling:
> Another note: Make sure the memoization is enforced for similar requests, via thumbprint and not through cache. We do that today I assume? In case you haven't verified, please confirm the same and we are good.

Yes, all request memoization is done in-memory.

lib/msal-browser/src/interaction_handler/InteractionHandler.ts
tnorling: This is different from tenant switch. This is switching the _domain_ of the authority, I believe for outage scenarios

lib/msal-browser/src/interaction_handler/InteractionHandler.ts
tnorling: Done. Assigned to myself.

PR #6590, opened 2023-10-20
Backported fix for multiple matching tokens error

### LINKED ISSUE #6425: Multiple tokens in cache but same claims
### Core Library

MSAL Node (@azure/msal-node)

### Core Library Version

1.18.0

### Wrapper Library

Not Applicable

### Wrapper Library Version

None

Confidential

### Description

While using a CachePlugin that implements a distributed cache using Mongo (following the example [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/samples/msal-node-samples/auth-code-distributed-cache/src/middleware.ts)), that segments the cache based on homeaccount id. There is an OIDC token claim made as part of login.

After adding in claim request where we fetch 2 access tokens (graph and scoped access token) in the same event cycle, we're seeing an error 

This seems only affect the claims being passed back from the graph token. 

### Error Message

"multiple_matching_tokens": the cache contains multiple tokens satisfying the requirements. 

### Msal Logs

JSON cache (looks like the scope target is out of order):
```json
"9b109d69-e09e-46b5-83de-844311fb828b.XXXX-login.windows.net-accesstoken-XXXX-XXXX-user.read profile openid email--": {
      "home_account_id": "9b109d69-e09e-46b5-83de-844311fb828b.XXXX",
      "environment": "login.windows.net",
      "credential_type": "AccessToken",
      "client_id": "XXXX",
      "secret": "XXXX",
      "realm": "XXXX",
      "target": "User.Read profile openid email",
      "cached_at": "1693700315",
      "expires_on": "1693705213",
      "extended_expires_on": "1693710112",
      "token_type": "Bearer"
    },
    "9b109d69-e09e-46b5-83de-844311fb828b.XXXX-login.windows.net-accesstoken-XXXX-XXXX-profile openid email user.read--": {
      "home_account_id": "9b109d69-e09e-46b5-83de-844311fb828b.XXXX",
      "environment": "login.windows.net",
      "credential_type": "AccessToken",
      "client_id": "XXXX",
      "secret": "XXXX",
      "realm": "XXXX",
      "target": "profile openid email User.Read",
      "cached_at": "1693662534",
      "expires_on": "1693666342",
      "extended_expires_on": "1693670151",
      "token_type": "Bearer"
    }
```

### MSAL Configuration

```javascript
const msalConfig = {
      auth: {
        clientId: authConfig.clientId,
        authority: `${authConfig.instance}/${authConfig.tenantId}`,
        clientSecret: authConfig.clientSecret,
      },
      system: {
        tokenRenewalOffsetSeconds: authConfig.tokenRenewalOffsetSeconds,
        loggerOptions: {
          loggerCallback(loglevel, message) {
            console.log(message);
          },
          logLevel: isNaN(parseInt(process.env.MSAL_LOG_LEVEL, 10)) ? LogLevel.Error : parseInt(process.env.MSAL_LOG_LEVEL, 10),
          piiLoggingEnabled: false,
        },
      },
    } as Configuration;
```

### Relevant Code Snippets

```javascript
async getTokenSilent(
    silentRequest: SilentFlowRequest,
    sessionId: string
  ): Promise<AuthenticationResult | null> {
    const msalInstance = this.getMsalInstance(sessionId);
    let tokenResponse = null;

    await msalInstance.getTokenCache().getAllAccounts(); // required for triggering beforeCacheACcess
    tokenResponse = await msalInstance.acquireTokenSilent(silentRequest);

    return tokenResponse;
  }

const [tokenResponse, graphToken] = await Promise.all([
        options.authProvider.getTokenSilent(
          {
            account: req.session.account,
            scopes,
          },
          req.session.id
        ),
        options.authProvider.getTokenSilent(
          {
            account: req.session.account,
            scopes: ['User.Read'],
          },
          req.session.id
        ),
      ]);
```

### Reproduction Steps

Login that seems to fail the most is a guest of the tenant. We're not changing the order of the scopes when requesting so the return shouldn't be out of order.

### Expected Behavior

There shouldn't be 2 claims in the cache.

### Identity Provider

Azure AD / MSA

None (Server)

### Regression

_No response_

### Source

External (Customer)

Robbie-Microsoft:
Backporting [this bugfix](https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/6311) for "multiple matching tokens" error.

We have a customer who won't upgrade to MSALJS v2.x

### DIFF
--- a/lib/msal-common/src/cache/CacheManager.ts
+++ b/lib/msal-common/src/cache/CacheManager.ts
@@ -964,7 +964,7 @@ export abstract class CacheManager implements ICacheManager {
             this.commonLogger.info("CacheManager:getRefreshToken - No refresh token found.");
             return null;
         }
-        // TODO: address the else case after remove functions address environment aliases
+        // address the else case after remove functions address environment aliases
 
         this.commonLogger.info("CacheManager:getRefreshToken - returning refresh token");
         return refreshTokens[0] as RefreshTokenEntity;
--- a/lib/msal-common/src/cache/CacheManager.ts
+++ b/lib/msal-common/src/cache/CacheManager.ts
@@ -731,7 +731,13 @@ export abstract class CacheManager implements ICacheManager {
             this.commonLogger.info("CacheManager:getIdToken - No token found");
             return null;
         } else if (numIdTokens > 1) {
-            throw ClientAuthError.createMultipleMatchingTokensInCacheError();
+            this.commonLogger.info(
+                "CacheManager:getIdToken - Multiple id tokens found, clearing them"
+            );
+            idTokens.forEach((idToken) => {
+                this.removeIdToken(idToken.generateCredentialKey());
+            });
+            return null;
         }
 
         this.commonLogger.info("CacheManager:getIdToken - Returning id token");
@@ -845,7 +851,13 @@ export abstract class CacheManager implements ICacheManager {
             this.commonLogger.info("CacheManager:getAccessToken - No token found");
             return null;
         } else if (numAccessTokens > 1) {
-            throw ClientAuthError.createMultipleMatchingTokensInCacheError();
+            this.commonLogger.info(
+                "CacheManager:getAccessToken - Multiple access tokens found, clearing them"
+            );
+            accessTokens.forEach((accessToken) => {
+                this.removeAccessToken(accessToken.generateCredentialKey());
+            });
+            return null;
         }
 
         this.commonLogger.info("CacheManager:getAccessToken - Returning access token");
@@ -952,7 +964,7 @@ export abstract class CacheManager implements ICacheManager {
             this.commonLogger.info("CacheManager:getRefreshToken - No refresh token found.");
             return null;
         }
-        // address the else case after remove functions address environment aliases
+        // TODO: address the else case after remove functions address environment aliases
 
         this.commonLogger.info("CacheManager:getRefreshToken - returning refresh token");
         return refreshTokens[0] as RefreshTokenEntity;
--- a/change/@azure-msal-common-a6532552-313d-484f-b0f9-4a399a3ffff0.json
+++ b/change/@azure-msal-common-a6532552-313d-484f-b0f9-4a399a3ffff0.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Backported fix for multiple matching tokens error",
+  "packageName": "@azure/msal-common",
+  "email": "rgins16@gmail.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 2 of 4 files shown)


lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -845,7 +851,13 @@ export abstract class CacheManager implements ICacheManager {
             this.commonLogger.info("CacheManager:getAccessToken - No token found");
             return null;
         } else if (numAccessTokens > 1) {
-            throw ClientAuthError.createMultipleMatchingTokensInCacheError();
+            this.commonLogger.info(
+                "CacheManager:getAccessToken - Multiple access tokens found, clearing them"
```
bgavrilMS, line 855: This does not seem to cover refresh token, which I believe can also get into this state?

lib/msal-common/src/cache/CacheManager.ts
```diff
@@ -952,7 +964,7 @@ export abstract class CacheManager implements ICacheManager {
             this.commonLogger.info("CacheManager:getRefreshToken - No refresh token found.");
             return null;
         }
-        // address the else case after remove functions address environment aliases
+        // TODO: address the else case after remove functions address environment aliases
```
bgavrilMS: no TODOs please,

bgavrilMS (approved):

bgavrilMS:
Approved with comments.

lib/msal-common/src/cache/CacheManager.ts
konstantin-msft: AFAIK this issue affects id and access tokens only. Also, only first matched RT is returned by [getRefreshToken()](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-common/src/cache/CacheManager.ts#L1270)

konstantin-msft (approved):

lib/msal-common/src/cache/CacheManager.ts
pmaytak: Previously the exception was thrown only for access and ID tokens, not refresh. The first token in the collection was always returned for all three tokens. With this fix, there could still be 2 duplicate RTs in the cache but the first one is returned. The questions are why wasn't the exception thrown for multiple RTs and should this fix also apply to RTs then too?

pmaytak (approved):

lib/msal-common/src/cache/CacheManager.ts
konstantin-msft: We can't remove duplicated RTs as it essentially logs out user completely. I will instrument `getRefreshToken` function to check how often it comes across multiple matches to make informed decision whether it is worth making a change. Long-term goal is to implement cache schema versioning. Meanwhile, the logic stays the same.

PR #6599, opened 2023-10-23
Fix iframe fallback when RT is not found in cache

tnorling:
It seems this fallback behavior was missed in the transition from v2 to v3. When the RT cannot be found in the cache we should try to fallback to the iframe flow. Also includes minor refactor to make the code more readable and remove some unnecessary work.

### DIFF
--- a/lib/msal-browser/src/interaction_client/SilentRefreshClient.ts
+++ b/lib/msal-browser/src/interaction_client/SilentRefreshClient.ts
@@ -53,7 +53,6 @@ export class SilentRefreshClient extends StandardInteractionClient {
             silentRequest.authority,
             silentRequest.azureCloudOptions
         );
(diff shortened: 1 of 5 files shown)


lib/msal-browser/src/interaction_client/SilentRefreshClient.ts
```diff
@@ -63,24 +62,11 @@ export class SilentRefreshClient extends StandardInteractionClient {
             this.logger,
             this.performanceClient,
             request.correlationId
-        )(silentRequest)
-            .then((result) => result as AuthenticationResult)
-            .then((result: AuthenticationResult) => {
-                this.performanceClient.addFields(
```
tnorling, line 69: These things are added at the top level API

peterzenz (approved):

lalimasharda (approved):

PR #6601, opened 2023-10-24
Nested App Auth fixes

codexeon:
Some minor fixes from initial NAA fix including:

- expiresOn AuthenticationResult contained the wrong value.
- expires_in should be used with timestamp before request was made to server, to account for time deltas between servers.
- authenticationScheme was returning empty
- AccountInfo was not returning idTokenClaims, and should use id token for properties as a fallback.
- Added authority parameter for server to host to return authority used in the request.
- Avoid making call to getActiveAccount if host does not support it.
- Merge client capabilities and claims before sending request to host.
- uniqueId should be localAccountId to match implementation of existing msal-browser.
- Removes toNaaSilentTokenRequest which had the same behavior as existing toNaaTokenRequest.

### DIFF
--- a/lib/msal-browser/src/naa/AccountInfo.ts
+++ b/lib/msal-browser/src/naa/AccountInfo.ts
@@ -4,7 +4,7 @@
  */
 
 export type AccountInfo = {
-    homeAccountId: string;
+    homeAccountId?: string;
     environment: string;
     tenantId: string;
     username: string;
@@ -13,5 +13,4 @@ export type AccountInfo = {
     idToken?: string; // idTokenClaims can be parsed from idToken in MSAL.js
     platformBrokerId?: string; // Used by WAM previous called nativeAccountId
     idTokenClaims?: object;
-    client_info?: string;
 };
--- a/lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
+++ b/lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
@@ -15,16 +15,18 @@ import {
     ClientConfigurationError,
     InteractionRequiredAuthError,
     ServerError,
-    TimeUtils,
     ICrypto,
     Logger,
     AuthToken,
     TokenClaims,
     ClientAuthErrorCodes,
+    AuthenticationScheme,
+    RequestParameterBuilder,
+    StringUtils,
+    Constants,
 } from "@azure/msal-common";
 import { isBridgeError } from "../BridgeError";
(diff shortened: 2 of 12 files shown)


lib/msal-browser/src/naa/AccountInfo.ts
```diff
@@ -4,7 +4,7 @@
  */
 
 export type AccountInfo = {
-    homeAccountId: string;
+    homeAccountId?: string;
```
shoatman, line 7: Why is this optional?

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
```diff
@@ -119,34 +87,31 @@ export class NestedAppAuthAdapter {
...
+            this.crypto.base64Decode
         );
-
-        const account = this.fromNaaAccountInfo(response.account);
 
         const authenticationResult: AuthenticationResult = {
-            authority: response.account.environment,
-            uniqueId: response.account.homeAccountId,
+            authority: response.authority || response.account.environment,
+            uniqueId: response.account.localAccountId,
```
shoatman, line 103: Is this going to differ between standard and NAA?  If so, I think that's a problem.

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
```diff
@@ -176,24 +141,36 @@ export class NestedAppAuthAdapter {
...
+        const localAccountId =
+            fromAccount.localAccountId ||
+            tokenClaims?.oid ||
+            tokenClaims?.sub ||
+            Constants.EMPTY_STRING;
+
+        const tenantId =
+            fromAccount.tenantId || tokenClaims?.tid || Constants.EMPTY_STRING;
+
+        const homeAccountId =
```
shoatman, line 157: Going to need to understand the reasoning here.  Is the assumption that the active account in the Hub.... is always operating in their home tenant?  They are not working in another tenant?

shoatman (commented):
I have some questions before I approve... mostly about account identifiers.... OneAuth is confused on this topic... and I don't think we can do something different for NAA then we do outside of a hub..... if we don't have the information, because it's not available from OneAuth.... or someone said you can't have it.... then that's another discussion.  I wouldn't block in that case.... but would suggest that Alicia follow up with Angela....

shoatman:
We discussed offline.

shoatman (approved):

PR #6634, opened 2023-10-30
October 2023 Patch Release

jo-arroyo:


hectormmg (approved):
Let's just make sure there's no escaped double quotes on the changelogs to lower the risk of GitHub releases failing because of unparsable JSON.

tnorling (approved):

Issue #6510, opened 2023-09-25
Enabling regional STS using the azureRegion variable is not working.


niallb96:
### Core Library

MSAL Node (@azure/msal-node)

### Core Library Version

2.1.0

### Wrapper Library

Not Applicable

### Wrapper Library Version

None

### Public or Confidential Client?

Confidential

### Description

Enabling regional STS using the `azureRegion` variable is not working.   I've set the `azureRegion ` to `eastus `and deployed this changed for testing  but the logs still show a login url of: `https://login.microsoftonline.com` rather than `eastus.login.microsoftonline.com` indicating it's not working.  All the other logs look the same as when this variable isn't configured and there are no error logs.  I've also tried configuring regional STS using the environment variable with no success.

### Msal Logs

Info - acquireTokenByClientCredential called
Verbose - initializeRequestScopes called
Verbose - buildOauthClientConfiguration called
Verbose - building oauth client configuration with the authority: https://login.microsoftonline.com/***
Verbose - createAuthority called
Verbose - Attempting to get cloud discovery metadata in the config
Verbose - Did not find cloud discovery metadata in the config... Attempting to get cloud discovery metadata from the cache.
Verbose - The metadata entity is expired.
Verbose - Did not find cloud discovery metadata in the cache... Attempting to get cloud discovery metadata from the network.
Verbose - Attempting to find a match between the developer's authority and the CloudInstanceDiscoveryMetadata returned from the network request.
Verbose - cloud discovery metadata was successfully returned from getCloudDiscoveryMetadataFromNetwork()
Verbose - Client credential client created

### MSAL Configuration

```javascript
auth: {
        clientId,
        clientCertificate: {
          thumbprint: certificate.thumbprint.toLowerCase().replace(/:/g, ''),
          privateKey: certificate.privateKeyWithNewLines,
          x5c: certificate.certificateWithNewLines,
        },
      }
```

### Relevant Code Snippets

```javascript
public async getFirstPartyToken (stsHostname: string, tenantId: string, resourceAppId: string, skipCache: boolean): Promise<string> {
    const request: ClientCredentialRequest = {
      scopes: [`${resourceAppId}/.default`],
      authority: `${stsHostname}/${tenantId}`,
      skipCache,
      azureRegion: 'eastus2',
    };
    const response = await this.application.acquireTokenByClientCredential(request);
    return response!.accessToken;
  }
```

### Reproduction Steps

1. Set up request with the same configuration.
[...]

PR #6641, opened 2023-10-31
Adding logging to make regional scenarios more clear

trwalke:
Was able to confirm the regional configuration api works. Logs make the used endpoint misleading. 
Fix for https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/6510

### DIFF
--- a/lib/msal-common/src/authority/Authority.ts
+++ b/lib/msal-common/src/authority/Authority.ts
@@ -237,19 +237,6 @@ export class Authority {
         }
     }
 
-    /**
-     * OAuth /token endpoint host for requests
-     */
-        public get tokenEndpointHost(): string {
-            if (this.discoveryComplete()) {
-                return new UrlString(this.metadata.token_endpoint).getUrlComponents().HostNameAndPort;
-            } else {
-                throw createClientAuthError(
-                    ClientAuthErrorCodes.endpointResolutionError
-                );
-            }
-        }
-
     public get deviceCodeEndpoint(): string {
         if (this.discoveryComplete()) {
             return this.replacePath(
--- a/lib/msal-common/src/authority/Authority.ts
+++ b/lib/msal-common/src/authority/Authority.ts
@@ -237,6 +237,19 @@ export class Authority {
         }
     }
 
+    /**
+     * OAuth /token endpoint host for requests
+     */
+        public get tokenEndpointHost(): string {
+            if (this.discoveryComplete()) {
+                return new UrlString(this.metadata.token_endpoint).getUrlComponents().HostNameAndPort;
+            } else {
+                throw createClientAuthError(
+                    ClientAuthErrorCodes.endpointResolutionError
+                );
+            }
+        }
+
     public get deviceCodeEndpoint(): string {
         if (this.discoveryComplete()) {
             return this.replacePath(
--- a/lib/msal-node/src/client/ClientApplication.ts
+++ b/lib/msal-node/src/client/ClientApplication.ts
@@ -412,7 +412,7 @@ export abstract class ClientApplication {
             : this.config.auth.azureCloudOptions;
 
         // using null assertion operator as we ensure that all config values have default values in buildConfiguration()
-        const discoveredAuthority: Authority = await this.createAuthority(
+        const discoveredAuthority = await this.createAuthority(
             authority,
             azureRegionConfiguration,
             requestCorrelationId,
--- a/lib/msal-node/src/client/ClientApplication.ts
+++ b/lib/msal-node/src/client/ClientApplication.ts
@@ -412,10 +412,23 @@ export abstract class ClientApplication {
             : this.config.auth.azureCloudOptions;
 
         // using null assertion operator as we ensure that all config values have default values in buildConfiguration()
-        this.logger.verbose(
-            `building oauth client configuration with the authority: ${authority}`,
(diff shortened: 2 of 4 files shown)


lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -412,10 +412,23 @@ export abstract class ClientApplication {
             : this.config.auth.azureCloudOptions;
 
         // using null assertion operator as we ensure that all config values have default values in buildConfiguration()
-        this.logger.verbose(
-            `building oauth client configuration with the authority: ${authority}`,
-            requestCorrelationId
-        );
+        this.logger.verbose("here-----------------------------------------------------------------------------------------------------------------");
```
sameerag: Can we stick to one format for logs? I do not think we had separators in general and if you want to add, would help if they are standardized.

lib/msal-node/src/client/ClientCredentialClient.ts
```diff
@@ -229,6 +229,9 @@ export class ClientCredentialClient extends BaseClient {
                 authority.tokenEndpoint,
                 queryParametersString
             );
+
+            this.logger.info("Sending token request to host: " + authority.tokenEndpointHost);
```
pmaytak: - Can we just print out the full tokenEndpoint, instead of just the host? Maybe helpful to have full string for debugging? Unless this is enough?
- Can this log be put near the actual call, like in `executePostToTokenEndpoint`? Unless we already log stuff there.

pmaytak (approved):

lib/msal-common/src/authority/Authority.ts
```diff
@@ -237,6 +237,19 @@ export class Authority {
         }
     }
 
+    /**
+     * OAuth /token endpoint host for requests
+     */
+        public get tokenEndpointHost(): string {
```
bgavrilMS: Is it needed? Isn't `tokenEndpoint` enough?

lib/msal-node/src/client/ClientApplication.ts
bgavrilMS: I think @trwalke meant to delete this

lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -412,10 +412,23 @@ export abstract class ClientApplication {
...
+
+        if(azureRegionConfiguration?.azureRegion)
+        {
+            this.logger.verbose(
```
bgavrilMS: I'd bump to info

lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -412,10 +412,23 @@ export abstract class ClientApplication {
             : this.config.auth.azureCloudOptions;
 
         // using null assertion operator as we ensure that all config values have default values in buildConfiguration()
-        this.logger.verbose(
-            `building oauth client configuration with the authority: ${authority}`,
-            requestCorrelationId
-        );
+        this.logger.verbose("here-----------------------------------------------------------------------------------------------------------------");
+
+        if(azureRegionConfiguration?.azureRegion)
```
bgavrilMS: Shouldn't this logging block go after line 432, where you do discovery. This way you can output "msal was configured with authority X and region Y and it uses the token endpoint Z"

bgavrilMS (approved):

bgavrilMS (changes requested):
See comments.

lib/msal-node/src/client/ClientApplication.ts
trwalke: Yes, not needed

lib/msal-common/src/authority/Authority.ts
trwalke: `tokenEndpoint` includes the path which we consider pii from what I see in MSAL.net. I wanted a way to log it without the path. Is the path still considered pii?

trwalke: Is the path considered Pii?

bgavrilMS: We have discussed about PII and put some guidance here - https://learn.microsoft.com/en-us/entra/msal/dotnet/resources/handling-pii 

I don't see how the token endpoint URL can have PII. The authorization URL can have PII because it can have login_hint as query parameter, but the token endpoint cannot.

bgavrilMS (approved):

lib/msal-node/src/client/ClientApplication.ts
```diff
@@ -412,17 +412,18 @@ export abstract class ClientApplication {
...
-            requestCorrelationId
-        );
         const discoveredAuthority = await this.createAuthority(
             authority,
             azureRegionConfiguration,
             requestCorrelationId,
             userAzureCloudOptions
         );
 
+        this.logger.info(
```
Robbie-Microsoft, line 422: Why did you change this from verbose to info? Additionally, why did you move this log to after the createAuthority call?

bgavrilMS: There are some many bugs / incidents where the authority that MSAL computes is wrong, it's worth having this at info level.

trwalke: Also, putting this after the discovery will allow us to get the updated authority with the regional environment if it is available.

PR #6644, opened 2023-10-31
Fix no refresh required log message

hectormmg:
This PR:
- Updates ATS instrumentation to change logger message when cache outcome is 0 and doesn't require a refresh

### DIFF
--- a/lib/msal-common/src/client/SilentFlowClient.ts
+++ b/lib/msal-common/src/client/SilentFlowClient.ts
@@ -190,9 +190,17 @@ export class SilentFlowClient extends BaseClient {
             },
             correlationId
         );
-        this.logger.info(
-            `Token refresh is required due to cache outcome: ${cacheOutcome}`
-        );
+        switch (cacheOutcome) {
+            case CacheOutcome.NOT_APPLICABLE:
+                this.logger.info(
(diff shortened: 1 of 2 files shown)


lalimasharda (approved):

lib/msal-common/src/client/SilentFlowClient.ts
```diff
@@ -190,9 +190,17 @@ export class SilentFlowClient extends BaseClient {
...
+            case CacheOutcome.NOT_APPLICABLE:
+                this.logger.info(
+                    `Token refresh not required due to cache outcome: ${cacheOutcome}`
+                );
+                break;
+            default:
+                this.logger.info(
+                    `Token refresh is required due to cache outcome: ${cacheOutcome}`
+                );
+        }
```
tnorling: Case switch feels like overkill here. I'd suggest if/else but I think we already have cache hit messages so we likely should just log on a miss
```suggestion
        if (cacheOutcome !== CacheOutcome.NOT_APPLICABLE) {
                this.logger.info(
                    `Token refresh is required due to cache outcome: ${cacheOutcome}`
                );
        }
```

sameerag: +1. Especially since we are just logging and not adding any new functionality.

tnorling (approved):

PR #6646, opened 2023-10-31
Optimize response parsing

### LINKED ISSUE #6452: Is acquireTokenPopup() supposed to work in query mode?
### Core Library

MSAL.js (@azure/msal-browser)

### Core Library Version

3.0.2

### Wrapper Library

Not Applicable

### Wrapper Library Version

None

Public

### Description

Our software has to support multiple OpenId-Connect providers. The code below works perfectly for Keycloak and AzureAD in fragment mode. For Auth0 we have to configure query mode since Auth0 does not support fragment mode.

When using the query mode, the login in the popup window works, but the popup does not close and our application is rendered in the popup window instead. 

I had a look to the msal code and it seems as if it finally always ends in the method

```
authCodeResponse = this.authModule.handleFragmentResponse(
    locationHash,
    requestState
);

```            
independently of the configured server response type (QUERY or FRAGMENT).

Am I missing something or is the acquireTokenPopup() method not supposed to work in query mode?

Thanks for clarification
Stefan      

### MSAL Configuration

```javascript
const msalConfig = {
        auth:  {
            protocolMode: ProtocolMode.OIDC,
            OIDCOptions: {
                serverResponseType: serverResponseType(provider.serverResponseType),
                defaultScopes: ["openid"]
            },
            authorityMetadata: JSON.stringify({
                issuer: provider.issuer,
                authorization_endpoint: provider.authorizationEndpoint,
                token_endpoint: provider.tokenEndpoint,
                userinfo_endpoint: provider.userinfoEndpoint
            }),
            authority: provider.authority,
            clientId: provider.clientId,
            knownAuthorities: provider.knownAuthorities,
            redirectUri: provider.redirectUri || (documentUri.scheme + "://" + documentUri.authority),
            navigateToLoginRequestUrl: true
        },
        cache: {
            cacheLocation: "memoryStorage", // This configures where your cache will be stored
            storeAuthStateInCookie: false   // Set this to "true" if you are having issues on IE11 or Edge
        },
        system: {
            allowNativeBroker: false
        }
    }
```

### Relevant Code Snippets

```javascript
async function onLoginWithOpenIdProvider(provider: OpenIdConnectProvider) {
    const msalConfig = {
        auth:  {
            protocolMode: ProtocolMode.OIDC,
            OIDCOptions: {
                serverResponseType: serverResponseType(provider.serverResponseType),
                defaultScopes: ["openid"]
            },
            authorityMetadata: JSON.stringify({
                issuer: provider.issuer,
                authorization_endpoint: provider.authorizationEndpoint,
                token_endpoint: provider.tokenEndpoint,
                userinfo_endpoint: provider.userinfoEndpoint
            }),
            authority: provider.authority,
            clientId: provider.clientId,
            knownAuthorities: provider.knownAuthorities,
            redirectUri: provider.redirectUri || (documentUri.scheme + "://" + documentUri.authority),
            navigateToLoginRequestUrl: true
        },
        cache: {
            cacheLocation: "memoryStorage", // This configures where your cache will be stored
            storeAuthStateInCookie: false   // Set this to "true" if you are having issues on IE11 or Edge
        },
        system: {
            allowNativeBroker: false
        }
    }

    const myMSALObj = new PublicClientApplication(msalConfig);
    await myMSALObj.initialize();

    const request = {
        scopes: ["openid email profile"]
    };

    myMSALObj.acquireTokenSilent(
        request
    ).then(response => {
        emit('idTokenReceived', response.idToken);
    }).catch(() => {
        myMSALObj.acquireTokenPopup(
            request
        ).then(response => {
            emit('idTokenReceived', response.idToken);
        }).catch(error => {
            notifyRestFailure("Failed to initialize the ID token", error);
        });
    });
}

function serverResponseType(value: string): ServerResponseType {
    if (value == "fragment") {
        return ServerResponseType.FRAGMENT;
    } else if (value == "query") {
        return ServerResponseType.QUERY;
    } else {
        return ServerResponseType.FRAGMENT;
    }
}
```

### Identity Provider

Other

### Source

External (Customer)

tnorling:
This PR optimizes response fragment/query parsing by:
- ensuring it only happens once (today /authorize responses are parsed at least twice and up to 3 times in a single request)
- using APIs provided by the browser/node runtime rather than our own, slower, implementation
- ensuring we correctly extract the fragment or query depending on what was configured by OIDCOptions (there are bugs related to this today)
- renames references to "hash" to "response" to reflect the fact that it can come from either the hash or query
- removes popup timeout logic which only started counting once the popup returned to the redirectUri. We should consider implementing a timeout for the entire operation but this may be a breaking change so left that out of scope for this PR.


lib/msal-browser/src/interaction_client/PopupClient.ts
```diff
@@ -464,7 +473,9 @@ export class PopupClient extends StandardInteractionClient {
                 null
             );
 
-            await this.waitForLogoutPopup(popupWindow);
+            await this.monitorPopupForHash(popupWindow).catch(() => {
+                // Swallow any errors related to monitoring the window. Server logout is best effort
```
peterzenz, line 477: Observability datapoint goes here.  Likely just a boolean we can use to check for volume.  If it's high we'd look for more detail.

lib/msal-browser/src/interaction_client/RedirectClient.ts
```diff
@@ -379,27 +391,31 @@ export class RedirectClient extends StandardInteractionClient {
...
 
-        this.logger.verbose(
-            "Hash does not contain known properties, returning cached hash"
-        );
-        return cachedHash;
+        if (cachedHash) {
+            response = UrlUtils.getDeserializedResponse(cachedHash);
+            if (response) {
+                this.logger.verbose(
+                    "Hash does not contain known properties, returning cached hash"
```
peterzenz, line 398: Many of these verbose logging points with errors that we're eating are good places for extra observability.

peterzenz (approved):

lib/msal-browser/src/interaction_client/PopupClient.ts
tnorling: I didn't think we are tracking interactive requests at all in our telemetry today?

sameerag: We are getting request for reliability measures for sure. I think it will be good to start adding observability metrics for all flows.

lib/msal-browser/src/interaction_handler/InteractionHandler.ts
```diff
@@ -53,22 +54,18 @@ export class InteractionHandler {
      * @param locationHash
      */
     async handleCodeResponseFromHash(
```
hectormmg: Nit: Can this be renamed now that it's not hash-specific? Not a dealbreaker if it's too large a change.

lib/msal-browser/src/interaction_handler/SilentHandler.ts
```diff
@@ -110,12 +112,17 @@ export async function monitorIframeForHash(
...
-            const contentHash = contentWindow
-                ? contentWindow.location.hash
-                : "";
+            let responseString = "";
+            if (contentWindow) {
+                if (responseType === ServerResponseType.QUERY) {
+                    responseString = contentWindow.location.search;
+                } else {
+                    responseString = contentWindow.location.hash;
+                }
```
hectormmg, line 121: We may want to factor this logic out into a function since it also shows up on `PopupClient` and `RedirectClient`.

hectormmg (approved):
Couple of nits but LGTM

konstantin-msft (approved):

sameerag (approved):
Approved. Since we are removing ticks (and I agree), let us make sure we track the popup time out in V4.

lib/msal-browser/src/interaction_handler/SilentHandler.ts
tnorling: There's consolidation that needs to happen across flows for sure. Leaving that as future work as this PR is already pushing the limits of reviewability.

lib/msal-browser/src/interaction_handler/InteractionHandler.ts
tnorling: Done

PR #6647, opened 2023-11-01
Instrument functions that drop multiple matched tokens

konstantin-msft:
- Instrument functions that drop multiple matched tokens.

### DIFF
--- a/lib/msal-common/src/telemetry/performance/PerformanceEvent.ts
+++ b/lib/msal-common/src/telemetry/performance/PerformanceEvent.ts
@@ -558,6 +558,13 @@ export type PerformanceEvent = {
      * Nested App Auth Fields
      */
     nestedAppAuthRequest?: boolean;
+
+    /**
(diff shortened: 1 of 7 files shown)


lib/msal-common/src/telemetry/performance/PerformanceEvent.ts
```diff
@@ -558,6 +558,13 @@ export type PerformanceEvent = {
      * Nested App Auth Fields
      */
     nestedAppAuthRequest?: boolean;
+
+    /**
+     * Multiple matched access/id/refresh tokens in the cache
+     */
+    multiMatchedAT?: number;
```
tnorling, line 565: Should these be `IntFields`?

konstantin-msft: Good point. Added to IntFields.

tnorling (approved):

sameerag (approved):

PR #6652, opened 2023-11-02
Fix redirect clearing popup hash

tnorling:
The handleRedirectPromise logic contains a bug which can result in popup windows not closing. This PR fixes that bug.

Today handleRedirectPromise does the following things:
1. Parses the hash and looks for known properties
2. Clears the hash from the window
3. Parses state to determine if the response belongs to a Redirect
4. Returns null if not

Because we clear the hash before we've determined if the response is even a redirect response this can result in race conditions preventing popup windows from closing if the popup uses a redirectUri which invokes handleRedirectPromise (most apps using React or Angular do this out of the box)

This PR addresses this bug by parsing state for the interactionType _before_ clearing the hash.


lib/msal-browser/src/utils/BrowserProtocolUtils.ts
```diff
@@ -16,26 +16,24 @@ export type BrowserStateObject = {
...
-                ProtocolUtils.parseRequestState(browserCrypto, state);
-            return requestStateObj.libraryState.meta as BrowserStateObject;
-        } catch (e) {
-            throw createClientAuthError(ClientAuthErrorCodes.invalidState);
-        }
+    try {
+        const requestStateObj: RequestStateObject =
+            ProtocolUtils.parseRequestState(browserCrypto, state);
+        return requestStateObj.libraryState.meta as BrowserStateObject;
+    } catch (e) {
```
peterzenz, line 36: Any exception details we should extract from e to log?

peterzenz (approved):

lib/msal-browser/src/utils/BrowserProtocolUtils.ts
tnorling: I'm not sure the specific error here would give us anything more than invalid state already does. If anything we could log the value that failed to parse but it's potentially PII which very few, if any, turn on logging for so value may be limited.

konstantin-msft (approved):

hectormmg (approved):

PR #6658, opened 2023-11-06
Return requestId in OBO AuthenticationResult

SlyryD:
Return requestId in OBO AuthenticationResult so we can get the requestId for figuring out why some requests succeed and others don't.

### DIFF
--- a/lib/msal-node/src/client/OnBehalfOfClient.ts
+++ b/lib/msal-node/src/client/OnBehalfOfClient.ts
@@ -31,6 +31,7 @@ import {
     TokenClaims,
     UrlString,
 } from "@azure/msal-common";
+import { HeaderNames } from "../utils/Constants";
 import { EncodingUtils } from "../utils/EncodingUtils";
 
 /**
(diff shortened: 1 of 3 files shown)


SlyryD:
Did local validation that `requestId` on the AuthenticationResult matches x-ms-request-id in the response headers.

bgavrilMS:
Hi @SlyryD - the design of all MSALs focused on correlation ID. Every request is associated with a correlation ID and I expect the AuthResult and the server errors to include the correlation ID (programatically as a string property and inside the error message). Correlation ID + Timestamp can be used to troubleshoot issues just as well as request ID. 

Are you not able to log correlation ID from MSAL JS?

SlyryD:
@bgavrilMS what is the recommended way to parse the correlationId from an error response? Thanks!

SlyryD:
> @bgavrilMS what is the recommended way to parse the correlationId from an error response? Thanks!

The oauth2 error response is parsed and placed into an AuthError, which makes it harder to get things like `correlation_id` out of it.

bgavrilMS:
@Robbie-Microsoft @trwalke - can you pls help answer this? In MSAL .NET the correlation ID to be part of: 

- AuthResult
- all service exceptions.

SlyryD:
> @Robbie-Microsoft @trwalke - can you pls help answer this? In MSAL .NET the correlation ID to be part of:
> 
> * AuthResult
> * all service exceptions.

`correlationId` is part of the `AuthError` now. I must have been looking at an old version. Thanks!

PR #6659, opened 2023-11-07
Bump package versions

tnorling:


peterzenz (approved):

hectormmg (approved):

lalimasharda (approved):

lib/msal-angular/src/packageMetadata.ts
```diff
@@ -1,3 +1,3 @@
 /* eslint-disable header/header */
-export const name = '@azure/msal-angular';
-export const version = '3.0.7';
+export const name = "@azure/msal-angular";
```
jo-arroyo: Hector mentioned last week that we have to change this back to single quotes for the format check to pass

lib/msal-angular/src/packageMetadata.ts
```diff
@@ -1,3 +1,3 @@
...
+export const version = "3.0.8";
```
hectormmg:
```suggestion
export const name = '@azure/msal-angular';
export const version = '3.0.8';
```

For some reason es-lint complains about this for angular only

jo-arroyo (approved):

sameerag (approved):

lib/msal-angular/src/packageMetadata.ts
tnorling: Good to know - we might consider adding the format:fix command after the bump to resolve

PR #6672, opened 2023-11-08
Nested App Auth fixes

codexeon:
Some minor fixes from initial NAA fix including:

- expiresOn AuthenticationResult contained the wrong value.
- expires_in should be used with timestamp before request was made to server, to account for time deltas between servers.
- authenticationScheme was returning empty
- AccountInfo was not returning idTokenClaims, and should use id token for properties as a fallback.
- Added authority parameter for server to host to return authority used in the request.
- Avoid making call to getActiveAccount if host does not support it.
- Merge client capabilities and claims before sending request to host.
- uniqueId should be localAccountId to match implementation of existing msal-browser.
- Removes toNaaSilentTokenRequest which had the same behavior as existing toNaaTokenRequest.
- Handle state parameter locally in MSAL.js

### DIFF
--- a/lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
+++ b/lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
@@ -24,6 +24,7 @@ import {
     RequestParameterBuilder,
     StringUtils,
     Constants,
+    createClientAuthError,
 } from "@azure/msal-common";
 import { isBridgeError } from "../BridgeError";
 import { BridgeStatusCode } from "../BridgeStatusCode";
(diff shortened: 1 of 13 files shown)


lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
```diff
@@ -176,26 +150,43 @@ export class NestedAppAuthAdapter {
...
+        idTokenClaims?: TokenClaims
+    ): MsalAccountInfo {
+        const effectiveIdTokenClaims =
+            idTokenClaims || (fromAccount.idTokenClaims as TokenClaims);
+
+        const localAccountId =
+            fromAccount.localAccountId ||
+            effectiveIdTokenClaims?.oid ||
+            effectiveIdTokenClaims?.sub ||
+            Constants.EMPTY_STRING;
```
konstantin-msft: Please use `` instead of `Constants.EMPTY_STRING` for smaller bundle size.

konstantin-msft (approved):

shoatman (approved):

PR #6681, opened 2023-11-10
Implementation of Azure Arc Managed Identity into MSAL-Node's MSI Feature Branch

Robbie-Microsoft:


lib/msal-node/src/client/ManagedIdentitySources/AzureArc.ts
```diff
@@ -3,17 +3,127 @@
...
+        request.queryParameters["api-version"] = ARC_API_VERSION;
+        request.queryParameters["resource"] = resource;
+        // bodyParameters calculated in BaseManagedIdentity.acquireTokenWithManagedIdentity
+
+        return request;
     }
 }
+
+const validateEnvironmentVariables = (
+    identityEndpoint: string | undefined,
```
bgavrilMS, line 172: A string can be `null`. Isn't it a better solution than using `undefined` which seems to be more lax?

lib/msal-node/src/client/ManagedIdentitySources/BaseManagedIdentitySource.ts
```diff
@@ -132,4 +146,43 @@ export abstract class BaseManagedIdentitySource {
             managedIdentityRequest
         );
     }
+
+    private async retryWithWWWAuthenticate(
```
bgavrilMS: This is Azure ARC specific, it should not be in base class?

lib/msal-node/src/client/ManagedIdentitySources/BaseManagedIdentitySource.ts
```diff
@@ -96,6 +101,15 @@ export abstract class BaseManagedIdentitySource {
             }
         }
 
+        // Azure Arc
+        if (response.status === HttpStatus.UNAUTHORIZED) {
```
bgavrilMS: It seems that handling the response is specific to Azure ARC. Since you went down the road of inheritance, like Neha did in .NET, the solution would be to have "HandleResponse" as protected and for AzureArc to ADD (not replace) to the HandleResponse functionality.

The code in .NET is: 

```
protected HandleResponse(HttpResponse response)
{
     if (response.StatusCode == System.Net.HttpStatusCode.Unauthorized)
    { 
          newHttpResponse = HandleComplexRetryLogic();
          base.HandleResponse(newHttpResponse);
    }
    else {
      base.HandleResponse(response);
}

```

bgavrilMS (changes requested):
1. Base class should not have details about any derived classes
2. All PRs that are ready for review need unit tests. This is because making the code testable sometimes requires significant re-write.

lib/msal-node/src/client/ManagedIdentitySources/BaseManagedIdentitySource.ts
```diff
@@ -132,4 +146,43 @@ export abstract class BaseManagedIdentitySource {
...
+        }
+
+        const splitChallenge = challenge.split("=");
+        if (splitChallenge.length !== 2) {
+            throw createManagedIdentityError(
+                ManagedIdentityErrorCodes.wwwAuthenticateHeaderUnsupportedFormat
+            );
+        }
+
+        const authHeaderValue = `Basic ${splitChallenge[1]}`;
```
neha-bhargava: You will need to read the file at the path in `splitChallenge[1]` and append the contents to `Basic `. The contents of this file is a token, for testing you can use any file with a single string in it.

lib/msal-node/src/client/ManagedIdentitySources/BaseManagedIdentitySource.ts
neha-bhargava: Yes this is specific to azure arc.

lib/msal-node/src/client/ManagedIdentitySources/BaseManagedIdentitySource.ts
```diff
@@ -132,4 +146,43 @@ export abstract class BaseManagedIdentitySource {
...
+
+        this.logger.info(
+            `[Managed Identity] Adding authorization header to the request.`
+        );
+        networkRequest.headers[AUTHORIZATION_HEADER_NAME] = authHeaderValue;
+
+        try {
+            return await this.networkClient.sendGetRequestAsync<ManagedIdentityTokenResponse>(
+                networkRequest.computeUri(),
+                networkRequestOptions
```
neha-bhargava: Does the networkRequestOptions have the header added above at line 173?

lib/msal-node/src/client/ManagedIdentitySources/AzureArc.ts
Robbie-Microsoft: this is just verifying that the environment variable exists. It either exists, or it doesn't and therefore it's undefined. If the environment exists as null string, then that's bad, and this code will correctly register that as it not existing.

lib/msal-node/src/client/ManagedIdentitySources/BaseManagedIdentitySource.ts
Robbie-Microsoft: oops. good catch

lib/msal-node/src/client/ManagedIdentitySources/BaseManagedIdentitySource.ts
Robbie-Microsoft: I've implemented this feedback, but with a twist. It's a little different than how dotnet implements it, but I've begun to implement a "retry policy", and moved this logic inside of it. Please let me know your thoughts.

lib/msal-node/src/client/ManagedIdentitySources/BaseManagedIdentitySource.ts
```diff
@@ -50,6 +51,13 @@ export abstract class BaseManagedIdentitySource {
         managedIdentityId: ManagedIdentityId
     ): ManagedIdentityRequestParameters;
 
+    retryPolicy?(
```
neha-bhargava: This is very specific to azure arc. Should not be in the base class at all.

lib/msal-node/src/client/ManagedIdentitySources/AzureArc.ts
```diff
@@ -3,17 +3,181 @@
...
+
+        request.headers[METADATA_HEADER_NAME] = "true";
+        request.queryParameters["api-version"] = ARC_API_VERSION;
+        request.queryParameters["resource"] = resource;
+        // bodyParameters calculated in BaseManagedIdentity.acquireTokenWithManagedIdentity
+
+        return request;
+    }
+
+    public async retryPolicy(
```
neha-bhargava: This method should not be named retryPolicy. The concept of retry policy is different where you retry for certain error codes. This is handling the response and making the request with authHeader. In MSAL .Net and Java this is implemented by overriding the `HandleResponse` method from the base class and adding the logic of extracting the auth header from the response and making the request again.

Try to use a similar pattern.

lib/msal-node/test/client/ManagedIdentityApplication.spec.ts
```diff
@@ -194,6 +199,103 @@ describe("ManagedIdentityApplication unit tests", () => {
...
+                        },
+                    });
+
+                const sendGetRequestAsyncSpy = sinon.spy(
+                    Azure401CustomHttpClient.prototype,
+                    <any>"sendGetRequestAsync"
+                );
+
+                try {
+                    // this request will fail because of the way it's mocked
```
neha-bhargava: Add another test where you get the token successfully after the header is added to the request.

lib/msal-node/test/client/ManagedIdentityApplication.spec.ts
```diff
@@ -194,6 +199,103 @@ describe("ManagedIdentityApplication unit tests", () => {
...
+            process.env["IMDS_ENDPOINT"] = "fake_IMDS_ENDPOINT";
+        });
+
+        afterAll(() => {
+            // restore old environment
+            process.env = OLD_ENVS;
+        });
+
+        // Azure Arc Managed Identities can only be system assigned
+        describe("System Assigned", () => {
```
neha-bhargava, line 214: Add a test to verify the error message when user assigned identities are used.

lib/msal-node/test/client/ManagedIdentityApplication.spec.ts
neha-bhargava: Also add tests to verify the error response when the auth header cannot be parsed or is empty or the file path is not in the response.

lib/msal-node/test/client/ManagedIdentityApplication.spec.ts
```diff
@@ -194,6 +199,103 @@ describe("ManagedIdentityApplication unit tests", () => {
         });
     });
 
+    describe("Acquires a token successfully via an Azure Arc Managed Identity", () => {
```
neha-bhargava, line 199: Add tests in all the sources to verify general error scenarios:
1. Response is empty
2. Response has status code 200 but the response body is null
3. Error message is parsed when the scopes are wrong

Another test case is to acquire a token for scope 1 and acquire a token for scope 2 and both should be acquired from IDP and not cache.

pmaytak (commented):
+1 to Neha's feedback. Here's the .NET's Azure Arc tests for reference: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/AzureArcTests.cs

When writing tests, I try to think of them in terms of inputs + outputs of each tested unit and write all (or at least the most common ones) the combinations for success and failure.

lib/msal-node/test/client/ManagedIdentityApplication.spec.ts
pmaytak, line 1: Should these tests be separated into each separate test file? Wasn't there an issue with nested describes and beforeX events? Or is everything working now?

lib/msal-node/test/client/ManagedIdentityApplication.spec.ts
Robbie-Microsoft: I'm going to make this a separate PR since it involves all sources.

lib/msal-node/test/client/ManagedIdentityApplication.spec.ts
Robbie-Microsoft: Everything is working now. I don't currently see a reason to break each source into their own test files.

lib/msal-node/test/client/ManagedIdentityApplication.spec.ts
Robbie-Microsoft: Regarding your first comment: that's not possible because of the way the http client is mocked. I can't modify the managed identity's http client after it's created. That's why this test is set up the way it is. I check if the get request was called twice, and ensure that the second request has the auth header.

lib/msal-node/test/client/ManagedIdentityApplication.spec.ts
```diff
@@ -194,6 +199,199 @@ describe("ManagedIdentityApplication unit tests", () => {
...
+            let managedIdentityApplication: ManagedIdentityApplication;
+            beforeEach(() => {
+                managedIdentityApplication = new ManagedIdentityApplication(
+                    systemAssignedConfig
+                );
+            });
+
+            test("acquires a token", async () => {
+                expect(ManagedIdentityTestUtils.isAzureArc()).toBe(true);
+
```
trwalke, line 224: Seems like there are a lot of random empty lines. Wouldn't hurt to remove them and shorten the file.

lib/msal-node/test/client/ManagedIdentityApplication.spec.ts
trwalke: Might make it better for readability and making the large file of tests seem a little less overwhelming

lib/msal-node/src/client/ManagedIdentitySources/BaseManagedIdentitySource.ts
```diff
@@ -50,6 +50,56 @@ export abstract class BaseManagedIdentitySource {
...
+        refreshAccessToken: boolean | undefined
+    ): Promise<ServerAuthorizationTokenResponse> {
+        return await this.validateResponse(
+            response,
+            responseHandler,
+            refreshAccessToken
+        );
+    }
+
+    public validateResponse(
```
trwalke: This feels like there is too much functionality in here. The method name says validateResponse but it is also creating the response which can be misleading. I reccomend breaking the functionality into 2 functions. something like `CreateTokenResponse` which feeds its result into `validateTokenResponse`

lib/msal-node/src/client/ManagedIdentitySources/AzureArc.ts
```diff
@@ -3,17 +3,204 @@
...
+
+        request.headers[METADATA_HEADER_NAME] = "true";
+        request.queryParameters["api-version"] = ARC_API_VERSION;
+        request.queryParameters["resource"] = resource;
+        // bodyParameters calculated in BaseManagedIdentity.acquireTokenWithManagedIdentity
+
+        return request;
+    }
+
+    public async validateResponseAsync(
```
neha-bhargava: Looks good.

lib/msal-node/src/client/ManagedIdentitySources/AzureArc.ts
```diff
@@ -3,17 +3,204 @@
...
+        this.endpoint = endpoint;
+    }
+
+    public static tryCreate(
+        logger: Logger,
+        nodeStorage: NodeStorage,
+        networkClient: INetworkModule,
+        cryptoProvider: CryptoProvider
+    ): AzureArc | null {
+        const imdsEndpoint: string | undefined = process.env["IMDS_ENDPOINT"];
```
bgavrilMS, line 62: nit: why not just make imdsEndpoint null if the process.env returns undefined?

lib/msal-node/src/client/ManagedIdentitySources/AzureArc.ts
```diff
@@ -3,17 +3,204 @@
...
+
+        const [areEnvironmentVariablesValidated, endpoint]: [
```
bgavrilMS, line 64: too many endpoints :). Please use a longer name to disambiguate

bgavrilMS (approved):

lib/msal-node/test/client/ManagedIdentityApplication.spec.ts
Robbie-Microsoft: We're at the mercy of the linter.

lib/msal-node/test/client/ManagedIdentityApplication.spec.ts
Robbie-Microsoft: Every other client in msal-common and msal-node have all their tests situated in one file.

lib/msal-node/src/client/ManagedIdentitySources/AzureArc.ts
Robbie-Microsoft: As an environment variable, it's either defined or it's not. I'm not sure what you're getting at.

lib/msal-node/src/client/ManagedIdentitySources/AzureArc.ts
Robbie-Microsoft: I'm not sure what you mean... the imdsEndpoint variable represents the IMDS_ENDPOINT environment variable. What's ambiguous about that?

lib/msal-node/test/client/ManagedIdentityApplication.spec.ts
Robbie-Microsoft: I'll do this in another PR

PR #6688, opened 2023-11-11
fix: added missing isInteractionRequiredError to export

jaskirat8:
isInteractionRequiredError function was missing from the exported functions list and hence cannot be used by apps already relying on this. Added it back to the export list for msal-common module.


tnorling:
This is an internal function not intended to be used by anything other than MSAL, nor should applications be consuming msal-common. You should validate errors using `error instanceof InteractionRequiredAuthError` instead

PR #6694, opened 2023-11-13
Check HostNameAndPort is defined before calling endsWith

dylank:
Occasionally the HostNameAndPort is undefined which causes the page to crash. This change checks it is defined before trying to call `endsWith`.

Using:
```
    "@azure/msal-browser": "^2.32.2",
    "@azure/msal-react": "^1.5.2",
```


tnorling:
HostNameAndPort should never be undefined. There's something else going on here. Please upgrade to the latest (v3) and check if this has been resolved. If it's still occurring please open an issue so we can dig in deeper to understand why you're getting undefined here.

PR #6696, opened 2023-11-14
Implementation for Cloud Shell and Service Fabric in Managed Identity

trwalke:
Tasks:

- [x] Implement Cloud Shell
- [x] Implement Service Fabric
- [ ] Unit tests for Cloud Shell
- [ ] Uni tests for Service Fabric


lib/msal-node/src/client/ManagedIdentitySources/CloudShell.ts
```diff
@@ -3,17 +3,107 @@
...
+    public createRequest(): ManagedIdentityRequestParameters {
+        const request: ManagedIdentityRequestParameters =
+            new ManagedIdentityRequestParameters(
+                HttpMethod.POST,
+                this.endpoint
+            );
+
+        request.headers["ContentType"] = "application/x-www-form-urlencoded";
+        request.headers["Metadata"] = "true";
+        // bodyParameters calculated in BaseManagedIdentity.acquireTokenWithManagedIdentity
```
trwalke, line 69: Can you confirm that resource is added as a body parameter?
@Robbie-Microsoft

Robbie-Microsoft: No, it's a query parameter. Look at the IMDS code.

trwalke: Ok, I see it but in the Cloud Shell implementation in .NET it is in the body parameters
`request.BodyParameters.Add("resource", resource);`

Shouldn't it also be a body parameter here?

Robbie-Microsoft: Oops, you're right actually. CloudShell uses a post request, and the rest of the sources use get requests.

PR #6701, opened 2023-11-16
Make AADServerParamKeys individual exports

tnorling:
Moves AADServerParamKeys to separate exports under a new "Constants" folder. Removes SSOTypes which was redundant.

### DIFF
--- a/lib/msal-common/src/utils/Constants.ts
+++ b/lib/msal-common/src/utils/Constants.ts
@@ -123,62 +123,6 @@ export type AADAuthorityConstants =
     (typeof AADAuthorityConstants)[keyof typeof AADAuthorityConstants];
 
 /**
- * Keys in the hashParams sent by AAD Server
- */
-export const AADServerParamKeys = {
-    CLIENT_ID: "client_id",
-    REDIRECT_URI: "redirect_uri",
-    RESPONSE_TYPE: "response_type",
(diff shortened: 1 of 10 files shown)


lib/msal-common/src/utils/Constants.ts
```diff
@@ -202,22 +146,6 @@ export const PromptValue = {
...
-export const SSOTypes = {
-    ACCOUNT: "account",
-    SID: "sid",
-    LOGIN_HINT: "login_hint",
-    ID_TOKEN: "id_token",
-    DOMAIN_HINT: "domain_hint",
-    ORGANIZATIONS: "organizations",
-    CONSUMERS: "consumers",
-    ACCOUNT_ID: "accountIdentifier",
-    HOMEACCOUNT_ID: "homeAccountIdentifier",
```
jo-arroyo, line 217: Some of these removed don't appear in the new AADServerParamKeys. Were they not needed/used at all?

jo-arroyo (approved):

lib/msal-common/src/utils/Constants.ts
tnorling: They were not referenced anywhere, likely weren't cleaned up when some feature or another was changed or removed

konstantin-msft (approved):

PR #6704, opened 2023-11-17
Allow overriding log level and PII setting with session storage key-values

konstantin-msft:
- Allow overriding log level and PII setting with session storage key-values to troubleshoot errors in non-dev environments.


lib/msal-browser/docs/logging.md
```diff
@@ -59,5 +59,26 @@ An example usage in a sample can be accessed [here](https://github.com/AzureAD/m
...
 
+## Override log level and PII setting
+
+These are the steps to override MSAL log level and PII settings to troubleshoot errors in non-dev environments:
+
+### Navigate to local storage
+
+1. Open developer tools by pressing F12
+2. Navigate to `Application` tab
+3. Click `Storage` and expand `Local Storage`
```
tnorling: Why localStorage rather than sessionStorage?

konstantin-msft: Updated to session storage.

lib/msal-browser/src/operatingcontext/BaseOperatingContext.ts
```diff
@@ -24,20 +29,87 @@ export abstract class BaseOperatingContext {
     protected available: boolean;
     protected browserEnvironment: boolean;
 
+    protected static LOGGER_CALLBACKS: Map<number, ILoggerCallback> = new Map([
```
tnorling: Shouldn't this just be one single callback with a switch/case for the logLevels?

lib/msal-browser/src/operatingcontext/BaseOperatingContext.ts
```diff
@@ -24,20 +29,87 @@ export abstract class BaseOperatingContext {
...
+                          LogLevel[logLevelKey]
+                      )
+                    : undefined;
+            const loggerOptions = this.config.system.loggerOptions;
+
+            this.logger = new Logger(
+                {
+                    correlationId: loggerOptions.correlationId,
+                    loggerCallback:
+                        loggerCallback || loggerOptions.loggerCallback,
```
tnorling: Since we OR with the provided config values here the else condition below is likely redundant. Can we simplify?

konstantin-msft: Given that this PR introduces two flags, customer may enable `msal.browser.log.pii` only, leaving `msal.browser.log.level` unset. In that case `loggerCallback` will be initialized as `undefined`, making us to fall back to the logger callback provided in the app config.

lib/msal-browser/src/operatingcontext/BaseOperatingContext.ts
konstantin-msft: I think that the map looks cleaner. I am fine with both approaches given that there is no perf difference.

tnorling: Maybe I'm reading this incorrectly but it looks to me like this map approach results in only logging out messages that are an exact match for the log level configured rather than all messages that match the log level and above. Is that accurate?

lib/msal-browser/src/operatingcontext/BaseOperatingContext.ts
tnorling: Right understood. I'm suggesting that the else condition below can be made redundant if you refactor this a bit.

lib/msal-browser/src/operatingcontext/BaseOperatingContext.ts
konstantin-msft: My original idea was to display messages that match the particular log level. I think you concern is valid. It is better to use a callback to display messages for the provided log level and below. Fixed, thank you.

lib/msal-browser/src/operatingcontext/BaseOperatingContext.ts
konstantin-msft: Simplified.

hectormmg (approved):

lib/msal-browser/docs/logging.md
```diff
@@ -59,5 +59,24 @@ An example usage in a sample can be accessed [here](https://github.com/AzureAD/m
...
+Additional details can be found [here](https://learn.microsoft.com/en-us/microsoft-edge/devtools-guide-chromium/storage/sessionstorage).
+
+### Override log level
+
+Add `msal.browser.log.level` key to `Session Storage`, set it's value to the desired log level (`Verbose`, for example) and refresh the page.
+Check `LogLevel` enum for the available options [here](../../msal-common/src/logger/Logger.ts).
 
+### Override PII log setting
 
+Add `msal.browser.log.pii` key to `Session Storage`, set it's value to `true` or `false` and refresh the page.
```
tnorling: Minor suggestion - 0 or 1 may be easier/less error prone than using string values

lib/msal-browser/docs/logging.md
```diff
@@ -59,5 +59,24 @@ An example usage in a sample can be accessed [here](https://github.com/AzureAD/m
...
     ![browser console](./images/BrowserLogEnablement.png)
 
+## Override log level and PII setting
+
+These are the steps to override MSAL log level and PII settings to troubleshoot errors in non-dev environments:
+
+### Navigate to session storage
+
+1. Open developer tools by pressing F12
+2. Navigate to `Application` tab
```
peterzenz: These instructions vary per browser.  For instance, Firefox is Developer Tools -> Storage Tab...  The Application tab isn't needed here.  Can you update these to be called out per browser.

lib/msal-browser/src/operatingcontext/BaseOperatingContext.ts
```diff
@@ -24,15 +29,67 @@ export abstract class BaseOperatingContext {
...
+                    : undefined;
+            if (logLevel) {
+                loggerOptions.loggerCallback =
+                    BaseOperatingContext.loggerCallback;
+                loggerOptions.logLevel = logLevel;
+            }
+            if (piiLoggingEnabled) {
+                loggerOptions.piiLoggingEnabled = piiLoggingEnabled;
+            }
+        }
```
tnorling: Outer `if` is redundant
```suggestion
            const logLevel =
                logLevelKey && Object.keys(LogLevel).includes(logLevelKey)
                    ? LogLevel[logLevelKey]
                    : undefined;
            if (logLevel) {
                loggerOptions.loggerCallback =
                    BaseOperatingContext.loggerCallback;
                loggerOptions.logLevel = logLevel;
            }
            if (piiLoggingEnabled) {
                loggerOptions.piiLoggingEnabled = piiLoggingEnabled;
            }
```

tnorling (approved):
Looks good - recommend adding a couple tests as well

lib/msal-browser/docs/logging.md
peterzenz: We need more steps here.  This should be something we can refer folks to for log collection and troubleshooting.  Something along the lines of:
`Once you have enabled Verbose logging and (optionally) enabled PII logging, refresh the page and retry the sign-in operation.  The captured logs can be found *here*.  Please review them to ensure there is no sensitive data included in them before sharing them with the team requesting them.`

peterzenz (commented):
Tests?

lib/msal-browser/src/operatingcontext/BaseOperatingContext.ts
konstantin-msft: Fixed.

lib/msal-browser/docs/logging.md
konstantin-msft: I would prefer to stick to `true`/`false` to be consistent with the pii setting in the MSAL config.

lib/msal-browser/docs/logging.md
konstantin-msft: That's a very good call, thank you. Updated docs.

lib/msal-browser/docs/logging.md
konstantin-msft: Updated docs, thank you for the suggestion.

konstantin-msft:
> Tests?

Added tests

PR #6707, opened 2023-11-20
Nested App Auth is not working with Android response

codexeon:
Android API to return a response through Nested App Auth bridge is in a slightly different format than the other platforms. The payload looks like { data: string } instead of string. The other implementations are all using a consistent format, and there is only one location in code that we depend on bridge response format. Fix is to allow both response formats.


jekloudaMSFT (approved):
Thanks for picking this up

konstantin-msft (approved):

PR #6713, opened 2023-11-22
Fix cachePlugin.js not resolving when cache file doesn't exist

Rablet:
cachePlugin.js is used for caching in the msal-node-samples. 

If the cache file does not currently exist it crates the file, however in that scenario it currently never resolves the promise so calling something like acquireTokenByCode will not return on the first invocation. Calling acquireTokenByCode again will return as in that case the cache file has been created and the promise will be resolved

This PR fixes this by resolving the promise once the file has been created.


tnorling (approved):

Rablet:
This went stale so just wanted to check if you are all good with this PR?

lalimasharda (approved):

PR #6717, opened 2023-11-23
Fix HybidSample running

markmnl:
Fixes:
* yargs dependency missing, otherwise on fresh clone and `npm install`, `npm start` fails with this missing dependency
* localhost URIs should be http as per README to match application registration
* AADSTS70018 by removing extraQueryParameters.nativebroker from authCodeUrlParameters


markmnl:

tnorling (dismissed):

samples/msal-browser-samples/HybridSample/views/client-redirect.hbs
```diff
@@ -6,7 +6,7 @@
     const msalInstance = new msal.PublicClientApplication({
         auth: {
             clientId: "{{clientId}}",
-            redirectUri: "https://localhost:3000/auth/client-redirect",
```
lalimasharda, line 9: Hey @markmnl do you mind also turning off the nativebroker in system config too on line 20. Just change the value of allowNativeBroker to false. Thank you!

tnorling (approved):

PR #6725, opened 2023-11-26
Version bump automation enhancements

hectormmg:
This PR:
- Migrates msal-angular to us double quotes
- Updates `/vscode/settings.json` to set jest `runMode` to `on-demand`

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,1 +1,1 @@
-**/src/packageMetadata.ts
+**/packageMetadata.ts
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,1 +1,0 @@
-**/packageMetadata.ts
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,1 @@
+**/src/packageMetadata.ts
--- a/lib/msal-angular/package.json
+++ b/lib/msal-angular/package.json
@@ -29,7 +29,7 @@
     "e2e": "ng e2e",
     "prepack": "npm run build:all",
     "prepublishOnly": "echo Use npm run deploy && exit 1",
-    "format:check": "npx prettier --ignore-path --ignore-path .gitignore --check src",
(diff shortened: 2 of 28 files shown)


.prettierignore
```diff
@@ -0,0 +1 @@
+**/packageMetadata.ts
```
tnorling: I think rather than ignoring this I slightly prefer updating the settings for angular to be more consistent with the rest - if possible.

Otherwise, can you add a comment explaining why this should be ignored? e.g. "This is an auto-generated file"

hectormmg: Are we ok with the PR including pretty much every msal-angular file with quotes being changed?

hectormmg: Ok, format:fix did all the work, it's only about 19 files anyway and the only changes are on quotes, so should be a swift review.

lib/msal-angular/package.json
```diff
@@ -29,8 +29,8 @@
     "e2e": "ng e2e",
     "prepack": "npm run build:all",
     "prepublishOnly": "echo Use npm run deploy && exit 1",
-    "format:check": "npx prettier --ignore-path .gitignore --check src",
-    "format:fix": "npx prettier --ignore-path .gitignore --write src"
+    "format:check": "npx prettier --ignore-path --ignore-path .gitignore --check src",
```
tnorling:
```suggestion
    "format:check": "npx prettier --ignore-path .gitignore --check src",
```

lib/msal-angular/package.json
```diff
@@ -29,8 +29,8 @@
...
+    "format:fix": "npx prettier --ignore-path --ignore-path .gitignore --write src"
```
tnorling:
```suggestion
    "format:fix": "npx prettier --ignore-path .gitignore --write src"
```

tnorling (approved):

konstantin-msft (approved):

PR #6735, opened 2023-11-30
Fixed Linter Warnings

Robbie-Microsoft:


lib/msal-common/src/telemetry/performance/PerformanceClient.ts
```diff
@@ -113,8 +113,10 @@ export abstract class PerformanceClient implements IPerformanceClient {
      * @returns {IPerformanceMeasurement}
      */
     startPerformanceMeasurement(
-        measureName: string, // eslint-disable-line @typescript-eslint/no-unused-vars
-        correlationId: string // eslint-disable-line @typescript-eslint/no-unused-vars
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
tnorling, line 116: This still needed?

Robbie-Microsoft: VSCode shows an error when I remove it, but `npm run build` still completes successfully. Should I remove it?

`'_measureName' is defined but never used.eslint[@typescript-eslint/no-unused-vars](https://typescript-eslint.io/rules/no-unused-vars)`

tnorling: Build doesn't run the linter - `npm run lint` would be what you should run to verify. In any case it's not clear to me why you need to both disable the linter and prefix the variable name with _ here. It's also not clear to me why prefixing with _ is enough in the changes to `CacheManager.ts` above but not enough here.

change/@azure-msal-common-767f39e9-0011-4e70-8631-642117a1e0f7.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
```
tnorling, line 2: Any change to anything within the src folder should be at least patch

bgavrilMS (approved):

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
Robbie-Microsoft: ah, I think I was seeing problems that didn't actually exist because I haven't "run npm run" lint in quite some time. I'm going to close this PR.

PR #6737, opened 2023-12-01
Update Nested App Auth internal schema

codexeon:
Currently Nested App Auth uses a schema that contains union types such as response.body where the body can contain a different structure depending on the request type. This works in JavaScript that uses a dynamic JSON parser, but some native implementations require a strongly typed schema. There are workarounds on native, but it is easier for all platforms to support Nested App Auth if union types are removed from the schema.

Also remove AccountByHomeIdRequest, AccountByLocalIdRequest, and AccountByUsernameRequest that are not currently implemented.


lib/msal-browser/src/naa/BridgeProxy.ts
```diff
@@ -119,49 +117,35 @@ export class BridgeProxy implements IBridgeProxy {
...
-    public getTokenInteractive(request: TokenRequest): Promise<TokenResponse> {
-        return this.sendRequest<TokenResponse>("GetTokenPopup", request);
+    public getTokenInteractive(request: TokenRequest): Promise<AuthResult> {
+        return this.getToken("GetTokenPopup", request);
     }
 
     /**
      * getTokenSilent Attempts to get a token silently from the bridge
      * @param request A token request
      * @returns a promise that resolves to a token response or rejects with a BridgeError
```
baljesingh: Do we need to update the description, or will the existing one work? i.e "a promise that resolves to a auth result"

baljesingh: Need to update the description.

peterzenz (approved):

baljesingh (approved):
Looks good to me.

PR #6744, opened 2023-12-04
Fix external account loading

hectormmg:
This PR:
- Adds environment as an optional parameter to `buildAccountToCache` utility function and passes in `authority.hostNameAndPort` as environment when `loadAccount` in `TokenCache` calls it.


peterzenz (approved):
Is there a regression we should be adding a test for?

hectormmg:
@peterzenz yes, writing the tests right now. Pushing them in a bit.

hectormmg:
No new tests needed, just had to remove a mock that was hiding the bug

tnorling (approved):

PR #6748, opened 2023-12-05
Dynamically load `BrowserPerformanceMeasurement` to capture browser perf measurements if session storage flag is set

konstantin-msft:
**Changes:**
- Dynamically load `BrowserPerformanceMeasurement` to capture browser perf measurements if session storage flag is set.
- Calculate telemetry event duration using Unix epoch if browser performance API is not available.
- Update browser performance doc.

**Background:**
Browser performance measures introduce additional overhead caused by buffer polling and linear-ish time complexity when calling [getEntriesByName()](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/src/telemetry/BrowserPerformanceMeasurement.ts#L127C57-L127C57). We can directly use `window.performance.now()` instead to measure telemetry event duration.

### DIFF
--- a/lib/msal-browser/docs/performance.md
+++ b/lib/msal-browser/docs/performance.md
@@ -109,3 +109,19 @@ const callbackId: string = msalInstance.addPerformanceCallback((events: Performa
 
 const removed: boolean = msalInstance.removePerformanceCallback(callbackId);
 ```
+
+### Measuring browser performance
+
+Browser performance measurements are disabled by default due to significant performance overhead they impose.
+Applications that want to enable performance measurements reported to the browser's performance timeline should:
+
+1. Open browser developer tools
+    - Edge, Chrome and Firefox browsers: press F12
+    - Safari: go into Safari's preferences (`Safari Menu` > `Preferences`), select the `Advanced Tab` and enable `Show features for web developers`. Once that menu is enabled, you will find the developer console by clicking on `Develop` > `Show Javascript Console`
+2. Navigate to `Session Storage`:
+    - [Edge](https://learn.microsoft.com/en-us/microsoft-edge/devtools-guide-chromium/storage/sessionstorage)
+    - [Chrome](https://developer.chrome.com/docs/devtools/storage/sessionstorage)
+    - [Firefox](https://firefox-source-docs.mozilla.org/devtools-user/storage_inspector/local_storage_session_storage)
+    - Safari: navigate to `Storage` tab and expand `Session Storage`
+3. Select target domain
+4. Add `msal.browser.performance.enabled` key to `Session Stirage`, set it's value to `1`, refresh the page and check the browser's performance timeline.
--- a/lib/msal-browser/docs/performance.md
+++ b/lib/msal-browser/docs/performance.md
@@ -124,4 +124,4 @@ Applications that want to enable performance measurements reported to the browse
     - [Firefox](https://firefox-source-docs.mozilla.org/devtools-user/storage_inspector/local_storage_session_storage)
     - Safari: navigate to `Storage` tab and expand `Session Storage`
 3. Select target domain
-4. Add `msal.browser.performance.enabled` key to `Session Stirage`, set it's value to `1`, refresh the page and check the browser's performance timeline.
+4. Add `msal.browser.performance.enabled` key to `Session Storage`, set it's value to `1`, refresh the page and check the browser's performance timeline.
--- a/lib/msal-browser/src/telemetry/BrowserPerformanceClient.ts
+++ b/lib/msal-browser/src/telemetry/BrowserPerformanceClient.ts
@@ -21,26 +21,53 @@ import {
     BrowserCacheLocation,
 } from "../utils/BrowserConstants";
 
+/**
+ * Returns browser performance measurement module if session flag is enabled. Returns undefined otherwise.
+ */
+function getPerfMeasurementModule() {
+    let sessionStorage: Storage | undefined;
+    try {
+        sessionStorage = window[BrowserCacheLocation.SessionStorage];
+        const perfEnabled = sessionStorage?.getItem(BROWSER_PERF_ENABLED_KEY);
+        if (Number(perfEnabled) === 1) {
+            return import("./BrowserPerformanceMeasurement");
+        }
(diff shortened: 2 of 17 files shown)


lib/msal-browser/docs/performance.md
```diff
@@ -109,3 +109,19 @@ const callbackId: string = msalInstance.addPerformanceCallback((events: Performa
...
+1. Open browser developer tools
+    - Edge, Chrome and Firefox browsers: press F12
+    - Safari: go into Safari's preferences (`Safari Menu` > `Preferences`), select the `Advanced Tab` and enable `Show features for web developers`. Once that menu is enabled, you will find the developer console by clicking on `Develop` > `Show Javascript Console`
+2. Navigate to `Session Storage`:
+    - [Edge](https://learn.microsoft.com/en-us/microsoft-edge/devtools-guide-chromium/storage/sessionstorage)
+    - [Chrome](https://developer.chrome.com/docs/devtools/storage/sessionstorage)
+    - [Firefox](https://firefox-source-docs.mozilla.org/devtools-user/storage_inspector/local_storage_session_storage)
+    - Safari: navigate to `Storage` tab and expand `Session Storage`
+3. Select target domain
+4. Add `msal.browser.performance.enabled` key to `Session Stirage`, set it's value to `1`, refresh the page and check the browser's performance timeline.
```
tnorling: Typo
```suggestion
4. Add `msal.browser.performance.enabled` key to `Session Storage`, set it's value to `1`, refresh the page and check the browser's performance timeline.
```

lib/msal-browser/src/telemetry/BrowserPerformanceClient.ts
```diff
@@ -4,25 +4,43 @@
...
+import {
+    BROWSER_PERF_ENABLED_KEY,
+    BrowserCacheLocation,
+} from "../utils/BrowserConstants";
 
 export class BrowserPerformanceClient
     extends PerformanceClient
     implements IPerformanceClient
 {
+    private static PERF_MEASUREMENT_MODULE = (() => {
```
tnorling: Suggest making this an unexported function instead of `static` for better minification

tnorling: Or alternatively, run this in the constructor and set a flag on the instance

lib/msal-browser/src/telemetry/BrowserPerformanceClient.ts
```diff
@@ -223,4 +259,19 @@ export class BrowserPerformanceClient
             manuallyCompleted
         );
     }
+
+    /**
+     * Returns event duration in milliseconds using window performance API if available. Returns undefined otherwise.
+     * @param startTime {DOMHighResTimeStamp | undefined}
+     * @returns {number | undefined}
+     */
+    private getPerfDurationMs(
```
tnorling: This could also probably be a standalone function rather than a class method

lib/msal-common/src/telemetry/performance/IPerformanceClient.ts
```diff
@@ -35,10 +33,6 @@ export interface IPerformanceClient {
     removePerformanceCallback(callbackId: string): boolean;
     addPerformanceCallback(callback: PerformanceCallbackFunction): string;
     emitEvents(events: PerformanceEvent[], correlationId: string): void;
-    startPerformanceMeasurement(
```
tnorling, line 38: breaking change, please revert

lib/msal-common/src/telemetry/performance/IPerformanceClient.ts
```diff
@@ -14,7 +13,6 @@ export type InProgressPerformanceEvent = {
     add: (fields: { [key: string]: {} | undefined }) => void;
     increment: (fields: { [key: string]: number | undefined }) => void;
     event: PerformanceEvent;
-    measurement: IPerformanceMeasurement;
```
tnorling, line 17: Also breaking change

tnorling (changes requested):
Looks good besides the breaking changes

lib/msal-browser/src/telemetry/BrowserPerformanceClient.ts
konstantin-msft: Made it a standalone function.

lib/msal-browser/src/telemetry/BrowserPerformanceClient.ts
konstantin-msft: Made it a standalone function.

lib/msal-common/src/telemetry/performance/IPerformanceClient.ts
konstantin-msft: Reverted and marked as deprecated.

lib/msal-common/src/telemetry/performance/IPerformanceClient.ts
konstantin-msft: Reverted and marked as deprecated.

sameerag (approved):

lib/msal-browser/src/telemetry/BrowserPerformanceClient.ts
```diff
@@ -102,30 +135,50 @@ export class BrowserPerformanceClient
...
         // Capture page visibilityState and then invoke start/end measurement
         const startPageVisibility = this.getPageVisibility();
-
         const inProgressEvent = super.startMeasurement(
             measureName,
             correlationId
         );
+        const startTime: number | undefined = supportsBrowserPerformanceNow()
+            ? window.performance.now()
+            : undefined;
```
tnorling, line 144: What do we see in our telemetry if the performance API isn't available? Do we get some sort of indication or do we toss out the data?

tnorling (approved):

lib/msal-browser/src/telemetry/BrowserPerformanceClient.ts
konstantin-msft: If performance API isn't available, then telemetry event duration is calculated using Unix epoch, a subject to clock skew, instead of `browser.performance.now()`. No data is tossed out. There is no indication of performance API availability in telemetry though.

PR #6753, opened 2023-12-06
Reduce unnecessary cache lookups

tnorling:
When looking up tokens in the cache we:
1. Get and parse: idToken, accessToken, refreshToken, account and appMetadata
2. Evaluate the accessToken and if not found or expired throw
3. Catch error and lookup the refreshToken again and attempt to exchange it over the network

There's 2 ways to improve this pattern, both addressed in this PR:
1. Don't retrieve the refresh token until it's actually needed (if and when the accessToken needs to be refreshed)
2. Don't retrieve idToken, account and appMetadata until they're needed (accessToken lookup was successful)

This saves 1 cache lookup on all calls and an additional 3 cache lookups on calls that fail to return an access token from the cache

### DIFF
--- a/change/@azure-msal-common-5d3e5b0c-ebb6-4c61-b410-108529331ca9.json
+++ b/change/@azure-msal-common-5d3e5b0c-ebb6-4c61-b410-108529331ca9.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Don't lookup tokens until they are needed",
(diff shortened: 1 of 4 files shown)


hectormmg (approved):
Is there a matching 1P PR for this?

konstantin-msft (approved):

sameerag (approved):

PR #6755, opened 2023-12-07
Implemented Static Caching + Code Cleanup

Robbie-Microsoft:


lib/msal-node/src/client/ManagedIdentityApplication.ts
```diff
@@ -43,7 +43,7 @@ export class ManagedIdentityApplication {
     private config: ManagedIdentityNodeConfiguration;
 
     private logger: Logger;
-    private nodeStorage: NodeStorage;
+    static nodeStorage: NodeStorage | undefined;
```
bgavrilMS, line 46: 1. Use `private static` to ensure visibility is correct.
2. Can't you use "null" instead of "undefined". Feels like you're propagating this "undefined" state in many places where "null" is a better abstraction. In this case, nodeStorage is always there, it's just that it's not initiatlized at first.

https://stackoverflow.com/questions/5076944/what-is-the-difference-between-null-and-undefined-in-javascript

lib/msal-node/src/client/ManagedIdentityApplication.ts
```diff
@@ -69,12 +69,14 @@ export class ManagedIdentityApplication {
             knownAuthorities: [Constants.DEFAULT_AUTHORITY],
         };
 
-        this.nodeStorage = new NodeStorage(
-            this.logger,
-            this.config.managedIdentityId.id,
-            DEFAULT_CRYPTO_IMPLEMENTATION,
-            fakeStatusAuthorityOptions
-        );
+        if (!ManagedIdentityApplication.nodeStorage) {
```
bgavrilMS, line 72: Can't you use just `nodeStorage`? Since the variable is private to the class.

lib/msal-node/test/client/ManagedIdentitySources/AppService.spec.ts
```diff
@@ -36,15 +36,16 @@ describe("Acquires a token successfully via an App Service Managed Identity", ()
     });
 
     afterEach(() => {
+        // reset static variables after each test
         ManagedIdentityClient.identitySource = undefined;
+        ManagedIdentityApplication.nodeStorage = undefined;
```
bgavrilMS, line 41: Same comment for `identitySource`. Feels like `null` is a better abstraction than `undefined`.

lib/msal-node/src/client/ManagedIdentityApplication.ts
Robbie-Microsoft: static variable cannot be access that way. The classname must be used over "this."

bgavrilMS (approved):

PR #6757, opened 2023-12-09
Handle bad_token by removing bad refresh token from cache

hectormmg:
This PR:
- Catches `bad_token` sub error and handles it by immediately removing the refresh token used in the refresh request, then lets the error bubble up to be handled as it is today
- Adds test for this scenario


lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -234,6 +238,23 @@ export class ResponseHandler {
                 );
             }
 
+            if (
```
hectormmg: There's some duplicate logic here I'll most likely refactor this before merging.

lib/msal-common/src/error/InvalidGrantAuthError.ts
```diff
@@ -0,0 +1,110 @@
...
+        desc: InvalidGrantAuthErrorMessages[
+            InvalidGrantAuthErrorCodes.badToken
+        ],
+    },
+};
+
+/**
+ * Error thrown when invalid grant is used.
+ */
+export class InvalidGrantAuthError extends AuthError {
```
tnorling: Does this really need to be its own class? Why can't we continue to use InteractionRequiredAuthError/ServerError?

lib/msal-common/src/error/InvalidGrantAuthError.ts
```diff
@@ -0,0 +1,110 @@
...
+    }
+}
+
+/**
+ * Helper function used to determine if an error thrown by the server requires bad token removal
+ * @param errorCode
+ * @param errorString
+ * @param subError
+ */
+export function isInvalidGrantAuthError(
```
tnorling: I think the logic is simply `code === "invalid_grant"`

lib/msal-common/src/error/InvalidGrantAuthError.ts
hectormmg: Moved it back to InteractionRequired and added an exception for it in acquireTokenSilent's logic to determine if an error is silently resolvable

lib/msal-common/src/error/InteractionRequiredAuthError.ts
```diff
@@ -32,6 +34,8 @@ const InteractionRequiredAuthErrorMessages = {
         "The requested account is not available in the native broker. It may have been deleted or logged out. Please sign-in again using an interactive API.",
     [InteractionRequiredAuthErrorCodes.refreshTokenExpired]:
         "Refresh token has expired.",
+    [InteractionRequiredAuthErrorCodes.badToken]:
+        "Server returned invalid_grant because of an unusable refresh token was used. Refresh token must be removed from cache.",
```
tnorling: Cache clearing is performed by MSAL so this bit doesn't need to be suggested.

```suggestion
        "Identity provider returned bad_token due to an expired or invalid refresh token. Please invoke an interactive API to resolve.",
```

lib/msal-browser/src/controllers/StandardController.ts
```diff
@@ -2081,7 +2081,9 @@ export class StandardController implements IController {
                             const isSilentlyResolvable =
                                 (!(
                                     refreshTokenError instanceof
-                                    InteractionRequiredAuthError
+                                        InteractionRequiredAuthError &&
+                                    refreshTokenError.subError !==
```
tnorling: This logic is really hard to follow. It may be time to factor this into a helper function with each condition separated to make it more readable with comments and such

tnorling (approved):
A couple minor non-blocking suggestions. Looks good otherwise.

peterzenz (approved):

PR #6770, opened 2023-12-14
Add support for extra token request headers

hectormmg:
This PR:
- Adds client application configuration-level `extraTokenRequestHeaders` option so applications can add custom headers to be sent on /token calls (Auth Code + Refresh Token Flow + Node Clients)
- Adds request configuration-level `extraTokenRequestHeaders` option that will extend and overwrite (same-name headers) app config-level option


lib/msal-common/src/client/BaseClient.ts
```diff
@@ -126,6 +127,25 @@ export abstract class BaseClient {
...
         }
+
+        // Add client app configuration headers
+        const extraClientConfigTokenRequestHeaders =
+            this.config.authOptions.extraTokenRequestHeaders;
+
+        if (extraClientConfigTokenRequestHeaders) {
+            for (const headerName in extraClientConfigTokenRequestHeaders) {
+                headers[headerName] =
+                    extraClientConfigTokenRequestHeaders[headerName];
```
peterzenz, line 138: Do we want to add any checking to ensure that the key/value strings are valid?  No added " or : in there for instance.

peterzenz (approved):

rayluo:
> This PR:
> 
>     * Adds client application configuration-level `extraTokenRequestHeaders` option so applications can add custom headers to be sent on /token calls (Auth Code + Refresh Token Flow + Node Clients)
> 
>     * Adds request configuration-level `extraTokenRequestHeaders` option that will extend and overwrite (same-name headers) app config-level option

In what scenario will these new feature be useful?

hectormmg:
@rayluo we don't have a direct scenario that demanded it yet but it would allow apps to add newly required headers without us having to ship a library update.

lib/msal-common/src/client/BaseClient.ts
hectormmg: We certainly can. I didn't add validation initially because we generally take config input as opaque. We can throw ourselves if validation fails instead of letting the engine throw.

hectormmg: @peterzenz there's options for this, interested in your input:
1. Validate and throw at the moment client app/config is built
2. Validate and throw when the token/ request is built
3. Validate and just omit invalid headers when the token request is built, log a warning

lib/msal-browser/src/config/Configuration.ts
```diff
@@ -278,6 +282,7 @@ export function buildConfiguration(
         },
         skipAuthorityMetadataCache: false,
         supportsNestedAppAuth: false,
+        extraTokenRequestHeaders: {},
```
tnorling, line 285: Would this be more appropriate under `system`? I also wonder if this is the kind of functionality that overriding the network client would be better suited for?

hectormmg: I don't have a preference. I put it under auth because it's part of the request logic, I'm not particularly married to that. As far as falling back on the network client override, the only downside I can think of is changing the headers per-request wouldn't be as easy.

hectormmg: There's also more overhead if overriding the network client is required.

tnorling: What's the scenario we're trying to unblock today? Does it require request level headers?

hectormmg: No immediate scenario, mostly we're adding this so we don't have to ship a new version later if a new header is required for some exception (things like browser feature trials, for ex. Chrome's 3p cookie extension trial).

lib/msal-common/src/client/BaseClient.ts
peterzenz: Thanks for the options.  I think I'd be OK to skip this here as is and let the call fail downstream.  Thanks.

lib/msal-browser/src/config/Configuration.ts
peterzenz: This is more a just in case we need to add something in here.  No active scenario we need to unblock at the moment.

rayluo:
> we don't have a direct scenario that demanded it yet but it would allow apps to add newly required headers without us having to ship a library update.

This topic can be subjective, but personally I am not convinced that this is the direction to go. By exposing an arbitrary headers parameter, MSAL goes to the direction of a lower level towards a generic http library. This implies that app developers would need to understand a new scenario's requirement down to the details of its http header on the protocol level, and code that header themselves. I would rather have MSAL wait until that new scenario comes up, make an educated decision on "oh this is a valid scenario that we want to support, rather than some hack that we do not want to encourage", and then add an explicit new parameter together with its corresponding documentation "if you want to use scenario X, you use this new param `settingForX=...`".

Is this new parameter expected to be supported by other MSALs, too?

peterzenz:
> I would rather have MSAL wait until that new scenario comes up, make an educated decision on "oh this is a valid scenario that we want to support, rather than some hack that we do not want to encourage"

We already support a similar mechanism today to add additional query parameters to all our requests.  This really isn't much different than that.  That QS generalization saves us when partners are told to do integration testing against test slice or prod with a specific server feature enablement flag.

lib/msal-browser/src/config/Configuration.ts
sameerag: TLDR: Let's punt this unless we need/have a requirement. My POV below:

Historically we rarely had any new STS params added (except may be the format change for `login_hint`.) Do we foresee this happening for headers? I would like to get STS involved once to check how critical is this and/or other MSALs if they support this and also check if it is applicable for browsers. Adding in configuration especially, I am very mindful of exposing this to customers for params we do not support. If the STS ignores this, it is all fine but if the service throws an error, we are opening ourselves for something unnecessary. This in turn would then open this up for sanitization and the lack of use case does not render us enough material to do so.

sameerag, line 285: And if we need to support this, I will say adding it as an override in network implementation would be a better place, however that places a dependency on the client choosing a network implementation which isn't true always. Hence it may make it a little more complex implementation.

rayluo:
> > I would rather have MSAL wait until that new scenario comes up, make an educated decision on "oh this is a valid scenario that we want to support, rather than some hack that we do not want to encourage"
> 
> We already support a similar mechanism today to add additional query parameters to all our requests. This really isn't much different than that. That QS generalization saves us when partners are told to do integration testing against test slice or prod with a specific server feature enablement flag.

Fair enough.

That scenario of "using a test slice denoted by an extra query parameter" is actually a good example to show that we could wait until a scenario comes up. For what it is worth, we encountered that test-slice scenario too, and thought it's legit to use a test slice (which is conceptually different than the normal authority), so we provided native support by allowing the authority url to contain query parameter (and then an OIDC discovery took care of the rest). Not necessarily saying that it is a better choice, but at least showing that there can be merit in postponing a decision until a scenario comes up.

Another downside of the generic additionalQueryParam, additionalFormPostedData or extraHttpHeaders approach is that, the auth code flow contains two legs, but it is never semantically clear that which leg should pick up those extra parameters. Perhaps the existing implementation applies those params on both legs, but that would break if/when a future scenario requires a param to present on one leg but not the other.

hectormmg:
Closing since we opted to table this feature until it is actually required. Thanks for all the feedback folks!

PR #6780, opened 2023-12-19
Implemented Service Fabric + Cloud Shell Managed Identity

Robbie-Microsoft:
In making the Service Fabric + Cloud Shell code as efficient as possible, several code improvements that affect the other managed identity sources were implemented as well.

Cloud Shell has been manually tested
Service Fabric still needs to be manually tested


lib/msal-node/src/client/ManagedIdentitySources/AppService.ts
```diff
@@ -80,72 +85,55 @@ export class AppService extends BaseManagedIdentitySource {
...
+            );
+
+        request.headers[APP_SERVICE_SECRET_HEADER_NAME] = this.identityHeader;
+
+        request.queryParameters[API_VERSION_QUERY_PARAMETER_NAME] =
+            APP_SERVICE_MSI_API_VERSION;
+        request.queryParameters[RESOURCE_QUERY_PARAMETER_NAME] = resource;
+
+        if (
+            managedIdentityId.idType !== ManagedIdentityIdType.SYSTEM_ASSIGNED
```
bgavrilMS, line 109: nit: `=== type.UserAssigned` seems to be more in line with your intention.

lib/msal-node/src/client/ManagedIdentitySources/BaseManagedIdentitySource.ts
```diff
@@ -166,4 +182,54 @@ export abstract class BaseManagedIdentitySource {
...
+                return ManagedIdentityUserAssignedIdQueryParameterNames.MANAGED_IDENTITY_OBJECT_ID;
+            default:
+                throw createManagedIdentityError(
+                    ManagedIdentityErrorCodes.invalidManagedIdentityIdType
+                );
+        }
+    }
 }
+
+export const getValidatedEnvVariableUrlString = (
```
bgavrilMS: Shouldn't this just be a static public method in this class or some helper class?

lib/msal-node/src/client/ManagedIdentitySources/BaseManagedIdentitySource.ts
```diff
@@ -166,4 +182,54 @@ export abstract class BaseManagedIdentitySource {
...
+    envVariableStringName: string,
+    envVariable: string,
+    sourceName: string,
+    logger: Logger
+): string => {
+    try {
+        return new UrlString(envVariable).urlString;
```
bgavrilMS: What is `UrlString` ? How is it different than `Url`?

lib/msal-node/src/client/ManagedIdentitySources/ServiceFabric.ts
```diff
@@ -3,17 +3,155 @@
...
+    private identityHeader: string;
+
+    constructor(
+        logger: Logger,
+        nodeStorage: NodeStorage,
+        networkClient: INetworkModule,
+        cryptoProvider: CryptoProvider,
+        identityEndpoint: string,
+        identityHeader: string,
+        systemAssigned: boolean
```
bgavrilMS: Why don't you pass the actual type here? It's cleaner than passing a boolean - you can at least use it in logging.

lib/msal-node/src/client/ManagedIdentitySources/ServiceFabric.ts
```diff
@@ -3,17 +3,155 @@
...
+    ) {
+        super(logger, nodeStorage, networkClient, cryptoProvider);
+
+        this.identityEndpoint = identityEndpoint;
+        this.identityHeader = identityHeader;
+
+        if (systemAssigned) {
+            logger.warning(
+                `[Managed Identity] ${ManagedIdentitySourceNames.SERVICE_FABRIC} user assigned managed identity ClientId or ResourceId is not configurable at runtime.`
```
bgavrilMS: I don't understand this warning. What does it mean?

lib/msal-node/src/client/ManagedIdentitySources/ServiceFabric.ts
```diff
@@ -3,17 +3,155 @@
...
+            string | undefined
+        ] = validateEnvironmentVariables(
+            process.env[
+                ManagedIdentityEnvironmentVariableNames.IDENTITY_ENDPOINT
+            ] || undefined,
+            identityHeader,
+            process.env[
+                ManagedIdentityEnvironmentVariableNames
+                    .IDENTITY_SERVER_THUMBPRINT
+            ] || undefined,
```
bgavrilMS: Consider having a help method that reads env variables and returns just string. Null string means env variable is not there. This whole "undefined" concept makes the code difficult to read.

lib/msal-node/src/client/ManagedIdentitySources/ServiceFabric.ts
bgavrilMS: The ID type is probably useful to the other sources? Maybe put it in the base ctor?

bgavrilMS (approved):

lib/msal-node/src/client/ManagedIdentitySources/AppService.ts
Robbie-Microsoft: that type doesn't exist. ManagedIdentityTypes looks like:
```typescript
const ManagedIdentityIdType = {
    SYSTEM_ASSIGNED: "system-assigned",
    USER_ASSIGNED_CLIENT_ID: "user-assigned-client-id",
    USER_ASSIGNED_RESOURCE_ID: "user-assigned-resource-id",
    USER_ASSIGNED_OBJECT_ID: "user-assigned-object-id",
}
```
Instead of using === for all three user assigned values, it's much less code to say ```!== ManagedIdentityIdType .SYSTEM_ASSIGNED```

lib/msal-node/src/client/ManagedIdentitySources/BaseManagedIdentitySource.ts
Robbie-Microsoft: It's MSALJS's own URL utils class.

"Url object class which can perform various transformations on url strings."

We use it to validate urls.

```Url``` is NodeJS's url class

lib/msal-node/src/client/ManagedIdentitySources/ServiceFabric.ts
Robbie-Microsoft: Let's chat offline about this. I use ```undefined``` because ```console.log(process.env["non-existant-env-var"])``` shows ```undefined```, not ```null```.

Why do you think it makes the code difficult to read? I think maybe you're confused because .net doesn't have the concept of ```undefined```? ```undefined``` is regularly used in JavaScript.

Look at this [stack overflow post](https://stackoverflow.com/questions/5076944/what-is-the-difference-between-null-and-undefined-in-javascript) for ```null``` vs ```undefined``` in JavaScript.

lib/msal-node/src/client/ManagedIdentitySources/ServiceFabric.ts
Robbie-Microsoft: I don't care about the actual type, I just want to know if it's system assigned or not. If it's not, then we use the logger to display a warning.

lib/msal-node/src/client/ManagedIdentitySources/ServiceFabric.ts
Robbie-Microsoft: See our teams chat for more information and please let me know your thoughts.

lib/msal-node/src/client/ManagedIdentitySources/ServiceFabric.ts
Robbie-Microsoft: It's not, I don't care about the actual type. I just want to know if it's system assigned or not. If it's not, then we use the logger to display a warning.

PR #6782, opened 2023-12-20
Implemented CloudShell Managed Identity

Robbie-Microsoft:


lib/msal-node/test/client/ManagedIdentitySources/AzureArc.spec.ts
```diff
@@ -164,106 +164,103 @@ describe("Acquires a token successfully via an Azure Arc Managed Identity", () =
 
             jest.restoreAllMocks();
         });
+    });
 
-        describe("Errors", () => {
-            test("throws an error when a user assigned managed identity is used", async () => {
-                expect(ManagedIdentityTestUtils.isAzureArc()).toBe(true);
+    describe("Errors", () => {
```
Robbie-Microsoft, line 169: Formatting bugfix, please ignore this file

lib/msal-node/src/client/ManagedIdentitySources/AzureArc.ts
```diff
@@ -49,18 +47,26 @@ export class AzureArc extends BaseManagedIdentitySource {
         nodeStorage: NodeStorage,
         networkClient: INetworkModule,
         cryptoProvider: CryptoProvider,
-        identityEndpoint: string
+        identityEndpoint: string,
+        systemAssigned: boolean
```
bgavrilMS, line 51: nit: As stated before, it's better to pass in the full configuration about SAMI / UAMI as it might help with logging etc. There's no need for higher level class to derive a flag and inject it here.

lib/msal-node/src/client/ManagedIdentitySources/CloudShell.ts
```diff
@@ -3,17 +3,116 @@
...
+        request.headers[METADATA_HEADER_NAME] = "true";
+
+        request.bodyParameters[RESOURCE_BODY_OR_QUERY_PARAMETER_NAME] =
+            resource;
+
+        return request;
     }
 }
+
+const validateEnvironmentVariables = (
```
bgavrilMS: Why is this a lambda outside the class? Why can't it be a private class method?

lib/msal-node/src/client/ManagedIdentitySources/CloudShell.ts
```diff
@@ -3,17 +3,116 @@
...
+    msiEndpoint: string | undefined,
+    logger: Logger
+): [boolean, string | undefined] => {
```
bgavrilMS: is it needed to send a tuple here? It seems to be that a string carries this information anyway: 

- null (not undefined) string means "false" / MI is not available
- a value means "true"

lib/msal-node/test/client/ManagedIdentitySources/CloudShell.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+        // make a copy of old environment
+        OLD_ENVS = process.env;
+
+        process.env[ManagedIdentityEnvironmentVariableNames.MSI_ENDPOINT] =
+            "msi_IDENTITY_ENDPOINT";
+    });
+
+    afterAll(() => {
+        // restore old environment
+        process.env = OLD_ENVS;
```
bgavrilMS: Does this work in all cases, for example: 

1. old_env = [VAR1 = "foo", VAR2 = "bar"]
2. then you set VAR1 = "baz" and VAR3 = "xyz"
3. then restore

You should the vars at step1 and no VAR3. At least in .NET the team had to do some manual deletions.

bgavrilMS (approved):

lib/msal-node/src/client/ManagedIdentitySources/CloudShell.ts
Robbie-Microsoft: This is just an extension of the static method ```TryCreate```. Static methods can't access the class's non-static variables/method, so I moved it outside of the class. I believe I only have two options: move the logic back inside the static method without a helper function, or make the helper function static. I don't think the latter is a good option because it's just a helper method; I don't want it to be able to be accessed by other instances of the class.

lib/msal-node/test/client/ManagedIdentitySources/CloudShell.spec.ts
Robbie-Microsoft: Thank you! This is the root cause of a bug I have been experiencing during testing <3

lib/msal-node/src/client/ManagedIdentitySources/CloudShell.ts
Robbie-Microsoft: I'm leaning towards the former, since having a helper function doesn't really contribute anything here. In this case, I won't need to deal with the tuple return value, as described in your other comment. I'll be able to eliminate several lines of code.

lib/msal-node/src/client/ManagedIdentityClient.ts
```diff
@@ -89,27 +88,37 @@ export class ManagedIdentityClient {
...
+        managedIdentityId: ManagedIdentityId
     ): ServiceFabric | AppService | CloudShell | AzureArc | Imds {
         const source =
             ServiceFabric.tryCreate(
                 logger,
                 nodeStorage,
                 networkClient,
                 cryptoProvider,
-                systemAssigned
+                managedIdentityId.idType ===
```
Robbie-Microsoft, line 99: Will change this to match Cloud Shell if these changes get approved

lib/msal-node/src/client/ManagedIdentityClient.ts
```diff
@@ -89,27 +88,37 @@ export class ManagedIdentityClient {
...
+                managedIdentityId
+            ) ||
             AzureArc.tryCreate(
                 logger,
                 nodeStorage,
                 networkClient,
-                cryptoProvider
+                cryptoProvider,
+                managedIdentityId.idType ===
+                    ManagedIdentityIdType.SYSTEM_ASSIGNED
```
Robbie-Microsoft, line 121: Will change this to match Cloud Shell if these changes get approved

bgavrilMS (approved):

PR #6801, opened 2024-01-04
3p exports for benchmarking tool

jo-arroyo:
Update exports and SilentRequest to be used in benchmarking.


lib/msal-browser/src/index.ts
```diff
@@ -133,6 +135,7 @@ export {
     // Server Response
     ExternalTokenResponse,
     // Utils
+    ProtocolUtils,
```
tnorling, line 138: This is an internal detail that should not be exported, let's discuss how you can get the functionality you need in another way. FetchClient too, though to a lesser extent given it implements a public interface

jo-arroyo:
Closing as mocking changed in benchmark tests to not require 3p exports

PR #6814, opened 2024-01-10
Fix bug affecting metadata resolution for tenanted authorities

tnorling:
Our hardcoded endpoint metadata is stored in a map keyed on full authority urls including the common tenants (common, organizations, consumers). This results in authority urls with specific tenants to fail the hardcoded lookup and go to the network for metadata which is taking ~700ms at P95. 

This PR does the following:
- Updates the hardcoded metadata map to key off domain only (consequently reducing the total number of entries)
- Removes extraneous fields that are not being used
- Adds telemetry data point indicating where we got the metadata from (hardcoded, config, network, cache)
- Removes unnecessary `createInstance` function


hectormmg (approved):

lib/msal-browser/src/cache/TokenCache.ts
```diff
@@ -142,7 +143,8 @@ export class TokenCache implements ITokenCache {
                 this.config.system.networkClient,
                 this.storage,
                 authorityOptions,
-                this.logger
+                this.logger,
+                request.correlationId || BrowserCrypto.createNewGuid()
```
sameerag, line 147: Is request.correlationId an option to the API? This is used for sideloading tokens so I assume this is not always set?

tnorling: Correct on both counts

sameerag (approved):

PR #6818, opened 2024-01-12
Update ssoSilent docs with troubleshooting information

hectormmg:
This PR:
- Adds a warning about content security policies and HTTP headers that may interfere with ssoSilent in the login docs
- Adds a troubleshooting section to the ssoSilent FAQ


lib/msal-browser/FAQ.md
```diff
@@ -203,9 +204,23 @@ You can use the `homeAccountId` property on the Account in the `AuthenticationRe
...
 
+## Troubleshooting Single Sign-On
+
+The following is a list of common causes for SSO failures when using MSAL Browser:
+
+### 1. The user has blocked third-party cookies in their browser
+
+Silent SSO requires third-party cookie access so the authentication service can persist a user's session accross tabs. If third-party cookies are blocked, silent SSO will fail and interaction will be required.
+
+### 2. There is a content security policy or HTTP header blocking the iframe from loading your redirect URI page
```
tnorling, line 218: Let's add 2 more here:

1. redirectUri must be same origin as the calling page
2. Some service error occurred which prevented the STS from redirecting back to the redirectUri (this will usually be accompanied by an X-frame options deny error and can be troubleshooted by opening the url in a new tab to view the error)

Let's also add a link to this FAQ and/or a short summary under the monitor_window_timeout error in the error doc

tnorling (approved):
Suggestion to expand but otherwise looks good. Thanks!

konstantin-msft (approved):

PR #6819, opened 2024-01-13
Support state parameter on silent requests

tnorling:
Adds support for `state` parameter on silent requests


hectormmg (approved):
LGTM, should we consider adding it to Node as well?

tnorling:
> LGTM, should we consider adding it to Node as well?

Node doesn't do silent /authorize requests so N/A

konstantin-msft (approved):

PR #6838, opened 2024-01-22
Reduce calls to resolveEndpointsAsync

tnorling:
We are building an `Authority` object and parsing/validating endpoint metadata at least twice in a single request - once to validate it matches the account object passed in and again to build the final request object. This PR refactors to do both these steps at the same time using a single `Authority` object. Additionally, telemetry is added to track the number of times an API is invoked over the course of a request to identify redundancies easier.


lib/msal-browser/src/interaction_client/BaseInteractionClient.ts
```diff
@@ -266,29 +244,59 @@ export abstract class BaseInteractionClient {
     /**
      * Used to get a discovered version of the default authority.
      * @param requestAuthority
+     * @param requestCorrelationId
```
konstantin-msft: Params in jsdoc do not match the actual ones.

konstantin-msft (approved):

lib/msal-browser/test/interaction_client/BaseInteractionClient.spec.ts
```diff
@@ -217,8 +219,10 @@ describe("BaseInteractionClient", () => {
             };
 
             await testClient
-                .validateRequestAuthority(
+                // @ts-ignore
+                .getDiscoveredAuthority(
                     "https://login.microsoftonline.com/common",
+                    undefined,
```
hectormmg: Can you add a comment next to the undefineds to document what they represent?

hectormmg (approved):

PR #6841, opened 2024-01-23
Make BrowserCrypto generate UUID v7 by default

konstantin-msft:
- Make BrowserCrypto generate UUID v7 by default.

### DIFF
--- a/lib/msal-browser/src/crypto/BrowserCrypto.ts
+++ b/lib/msal-browser/src/crypto/BrowserCrypto.ts
@@ -32,13 +32,8 @@ const MODULUS_LENGTH = 2048;
 const PUBLIC_EXPONENT: Uint8Array = new Uint8Array([0x01, 0x00, 0x01]);
 // UUID hex digits
 const UUID_CHARS = "0123456789abcdef";
-// UUID max counter
-const UUID_MAX_COUNTER = 0x3ff_ffff_ffff;
-
-const uint32Buffer = new Uint32Array(10);
-let uint32Cursor = 0;
-let uuidCounter = getUuidCounter();
-let uuidTimestamp = 0;
+// Array to store UINT32 random value
+const UINT32_ARR = new Uint32Array(1);
 
 const keygenAlgorithmOptions: RsaHashedKeyGenParams = {
     name: PKCS1_V15_KEYGEN_ALG,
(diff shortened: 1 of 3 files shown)


lib/msal-browser/src/crypto/BrowserCrypto.ts
```diff
@@ -30,6 +30,15 @@ const S256_HASH_ALG = "SHA-256";
...
 const PUBLIC_EXPONENT: Uint8Array = new Uint8Array([0x01, 0x00, 0x01]);
+// UUID hex digits
+const UUID_CHARS = "0123456789abcdef";
+// UUID max counter
+const UUID_MAX_COUNTER = 0x3ff_ffff_ffff;
+
+const uint32Buffer = new Uint32Array(10);
+let uint32Cursor = 0;
+let uuidCounter = getUuidCounter();
+let uuidTimestamp = 0;
```
tnorling: Variables should not be defined in the global scope. If you need to track things across multiple calls I'd suggest passing them in as input parameters. 

However, if this is just to guarantee uniqueness we may not need it if the likelihood of collisions is small. We don't have a requirement that I'm aware of that these be 100% unique and these values will be reset each time the page is refreshed or the lib re-imported anyway so we can't really guarantee it.

sameerag: +1

konstantin-msft: UUID v7 relies on the sequence (suffix), which is reset every millisecond. I don't think I can implement UUID v7 without using classes or global variables. Removing the sequence and making UUID suffix completely random defies the purpose of v7 structure and breaks weak causality.

tnorling: I don't understand, maybe we can take this offline, but the sequence is reset on each page load/new tab, yes? If so, what's the difference between resetting the sequence on each page versus resetting the sequence on each request?

konstantin-msft: Removed global vars as we decided not to fully adhere to UUID v7 spec.

lib/msal-browser/src/crypto/BrowserCrypto.ts
```diff
@@ -80,11 +86,54 @@ export function getRandomValues(dataBuffer: Uint8Array): Uint8Array {
...
 /**
- * Creates a new random GUID
- * @returns
+ * Creates a UUID v7 from the current timestamp.
+ * Implementation relies on the system clock to guarantee increasing order of generated identifiers.
+ * @returns {number}
  */
 export function createNewGuid(): string {
-    return window.crypto.randomUUID();
+    function getRandomUint32(): number {
```
tnorling: For consistency, anonymous function or define outside.

```suggestion
    const getRandomUint32 = (): number => {
```

tnorling (approved):

lib/msal-browser/src/crypto/BrowserCrypto.ts
konstantin-msft: Fixed.

sameerag (commented):
What changed w.r.t the sequential logic? Are we implementing this in part and still expecting the storage gains?

konstantin-msft:
> What changed w.r.t the sequential logic? Are we implementing this in part and still expecting the storage gains?

We drop maintenance of the monotonic sequence (reset every ms) in UUID suffix. However, timestamp-based prefix stays in place, providing storage size/perf gains.

sameerag (approved):

PR #6846, opened 2024-01-24
MSAL JS should not throw user account switch error when home account ids of the request/response match

lalimasharda:
This PR fixes 2 issues:
- When the parent app is requesting tokens on behalf of an iframed application, it should check request/response based on the native account id as well as the homeAccountId.
- When the tokens are brokered, the parent's native account id should not be cached in the iframed application. The iframed application should only store it's own native account id.


lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -405,24 +405,33 @@ export class NativeInteractionClient extends BaseInteractionClient {
...
         );
 
+        const cachedhomeAccountId =
+            this.browserStorage.getAccountInfoFilteredBy({
+                nativeAccountId: request.accountId,
+            })?.homeAccountId;
+
+        if (
+            homeAccountIdentifier != cachedhomeAccountId &&
+            response.account.id !== request.accountId
```
sameerag, line 437: Can we add the use cases in description of this PR where this is true that is, home account ID matches but account response differs and vice versa? I thought this was added to address the difference between embedded frame vs the parent frame in the request and am curious where this bug surfaced?

lalimasharda: Yep that's the main use case. Also, parent's native account id seemed to be getting passed on to the embeddd frame, so I have included a fix for that in this PR as well.

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -604,6 +613,10 @@ export class NativeInteractionClient extends BaseInteractionClient {
             idTokenClaims
         );
 
+        if (accountInfo.nativeAccountId !== response.account.id) {
+            accountInfo.nativeAccountId = response.account.id;
```
tnorling, line 621: How do we get into this state and why is this not an error?

lalimasharda: The accountInfo is what we get from the baseAccount passed by the handleNativeResponse() method. The baseAccount comes from the buildAccountToCache method in msal-common's ResponseHandler but that method overrides the embedded app's native account id that was passed in with the cached account (in this case it is the parent's cache). [This line specifically](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/e7e24e1b5d712c48d6ddc1374eacd7e7599bc18b/lib/msal-common/src/response/ResponseHandler.ts#L690C1-L691C1). Do we want to change that behaviour instead?

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -604,6 +613,14 @@ export class NativeInteractionClient extends BaseInteractionClient {
             idTokenClaims
         );
 
+        /**
+         * In case of brokered flows, this check prevents parent's cached native account id
+         * from overriding the brokered app's account id in the account info.
```
```suggestion
         * In pairwise broker flows, this check prevents the broker's native account id
         * from being returned over the embedded app's account id.
```

tnorling (approved):

hectormmg (approved):

PR #6849, opened 2024-01-25
Fix node-extensions tests

tnorling:
Jest only has experimental support for ESM and is failing when parsing `import.meta` (ESM only) as it's assuming CJS. This PR mocks `import.meta` as suggested [here](https://github.com/kulshekhar/ts-jest/issues/3888#issuecomment-1722524078) to address for the time being. If/when jest/ts-jest address this on their end we can remove this.


hectormmg (approved):
Thanks for finding the fix, lgtm!

konstantin-msft (approved):

PR #6854, opened 2024-01-29
Add active account update event

hectormmg:
This PR:
- Adds a new cross-tab account storage event called `ACTIVE_ACCOUNT_CHANGED` to `EventHandler` API
- Adds logic to emit the `ACTIVE_ACCOUNT_CHANGED` event when the `active-account-filters` cache entry is modified in a new tab
- Adds docs for event usage
- Adds tests


lib/msal-browser/src/event/EventHandler.ts
```diff
@@ -174,4 +181,26 @@ export class EventHandler {
...
+            const previousActiveAccountFilters = e.oldValue
+                ? JSON.parse(e.oldValue)
+                : null;
+            const newActiveAccountFilters = e.newValue
+                ? JSON.parse(e.newValue)
+                : null;
+
+            this.emitEvent(
+                EventType.ACTIVE_ACCOUNT_UPDATED,
+                undefined, // interactionType
```
sameerag: Do we care about the interactionType while emiting this?

lib/msal-browser/src/event/EventHandler.ts
```diff
@@ -139,6 +140,12 @@ export class EventHandler {
      */
     private handleAccountCacheChange(e: StorageEvent): void {
         try {
+            // Handle active account filter change
+            if (e.key?.includes(PersistentCacheKeys.ACTIVE_ACCOUNT_FILTERS)) {
```
sameerag, line 144: Confused a bit. We do not seem to compare for any change, but just the fact that the filters changed. Is that the intent?

hectormmg: Yes, it's just a notice. Given that for localStorage the active account is shared across tabs, we're just notifying the other tabs/windows that the active account has changed so they can trigger UI updates on the other tabs. Any comparison required would in theory be made in `setActiveAccount`. I don't think there's a reason to add comparisons though.

lib/msal-browser/src/event/EventHandler.ts
hectormmg: I don't believe so, `setActiveAccount` can be called from basically any scenario, I suspect most calls won't have an interaction type.

peterzenz (approved):

lib/msal-browser/docs/events.md
```diff
@@ -83,7 +99,10 @@ const callbackId = msalInstance.addEventCallback((message: EventMessage) => {
 
 ## Syncing logged in state across tabs and windows
 
-If you would like to update your UI when a user logs in or out of your app in a different tab or window you can subscribe to the `ACCOUNT_ADDED` and `ACCOUNT_REMOVED` events. The payload will be the `AccountInfo` object that was added or removed.
+If you would like to update your UI when a user logs in or out of your app or changes the active account in a different tab or window you can subscribe to the `ACCOUNT_ADDED`, `ACCOUNT_REMOVED`, and `ACTIVE_ACCOUNT_UPDATED` events.
+
+-   For account additions and removals, the payload will be the `AccountInfo` object that was added or removed.
+-   For active account updates, the payload will contain the previous and new active account filters as `AccountFilter` objects.
```
tnorling: Why filters? Why not the AccountInfo object here as well?

lib/msal-browser/src/event/EventHandler.ts
tnorling: +1 N/A

lib/msal-browser/src/event/EventType.ts
```diff
@@ -8,6 +8,7 @@ export const EventType = {
     INITIALIZE_END: "msal:initializeEnd",
     ACCOUNT_ADDED: "msal:accountAdded",
     ACCOUNT_REMOVED: "msal:accountRemoved",
+    ACTIVE_ACCOUNT_UPDATED: "msal:activeAccountChanged",
```
tnorling: Updated or Changed? Let's be consistent over both left and right hand sides

lib/msal-browser/docs/events.md
hectormmg: @tnorling the storage event is activated by the filters changing, there isn't necessarily a change at the cached account level (change can be at the active tenant level). The account objects aren't recorded in the storage event, but the old/new filters are.

Returning the account objects would require reading them from the cache (calling `getAccount` for each). The calling app can still pass the filters directly to `getAccount` to get the exact accounts in question if they want to, I'd rather we didn't force the perf hit on them.

lib/msal-browser/src/event/EventType.ts
hectormmg: Any preferences? I'll make sure it's standardized.

peterzenz: I missed this.  Changed might be clearer but I'm ok either way.

hectormmg: Updated the event name to "changed." I picked ACTIVE_ACCOUNT_UPDATED and copilot completed the line. I intended to change it to "updated" but forgot about it. I think "changed" is better as well.

lib/msal-browser/docs/events.md
tnorling: What happens if we change the way we store/identify active accounts? It sounds like we're leaking implementation details here. I don't disagree r.e. perf but given this is a public interface change we should first consider what we want the public interface to be then figure out how to build it, rather than defining the interface based on today's implementation, if that makes sense.

Personally I think returning the AccountInfo object is both more intuitive and consistent with the rest of our APIs. If you disagree or if doing so results in perf concerns that perhaps warrants further discussion.

tnorling: FWIW cache lookups should be inexpensive. If they're taking double digit or more ms that's something we should look into separately

hectormmg: @tnorling here's my take:

Semantically, `ACTIVE_ACCOUNT_CHANGED` is an event that signals a contextual/app state change across browser tabs/windows. The `AccountInfo` object is orthogonal given that it suffers no changes to cause or as a result of this storage event.

The actual object that changed in the cache is the filters object, which is both 
a) the minimum required context that the subscriber really needs to react to the event (e.g. update UI in a different tab to maintain consistency across tabs) and
b) the only context we can source from the storage event itself without additional cache lookups/parsing.

We can certainly re-design how we keep track of the active account, but I think that will also open us up to backward compatibility challenges and duplication of account info in the cache.

Happy to draft a design doc and set up some for the team to discuss this week, I agree we may benefit from further discussion.

JasonGore: Based on my current knowledge of MSAL I'm not actually sure what `AccountFilter` is, how the app would use it, or how it impacts the frequency at which this is called. Superficially (ignoring architectural implications) it does seem more consistent to pass in `AccountInfo` as `ACCOUNT_ADDED` and `ACCOUNT_REMOVED` do.

In terms of triggering, what are the situations in which this new event would get called? Whenever `tenantId` and `uniqueId` change per [Accounts in MSAL Javascript](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-common/docs/Accounts.md)? (I guess based on the test I see `homeAccountId`, `localAccountId` and `tenantId`, but not sure if there is any overlap here.) I think it might be useful to enumerate these conditions in the API event docs too.

lib/msal-browser/src/event/EventHandler.ts
JasonGore: A bit related to what can trigger this event in my other comment, but does this mean there may not be meaningful account changes when this event triggers?

lib/msal-browser/docs/events.md
hectormmg: @JasonGore `AccountFilter` is an extended subset of `AccountInfo` that is used internally for cache lookups and is exposed publically so the apps can pass it into `getAccountX` APIs. The specific fitler objects published in this event would contain the home/local account IDs and tenant ID (of previous and new active accounts).

So the way you'd use it in this case:
- To have context of what account was active before the event
- To get the new active account by passing in the newActiveAccountFilter into `getAccount` (alternatively and perhaps more correctly you could just call `getActiveAccount` and update your app state.

W.r.t. triggering, in the MSAL context, this event is triggered in subscribed tabs when `setActiveAccount` is called in any tab. In actual code, MSAL will emit the event whenever the cache entry for your clientId ending in ".active-account-filters" is modified in any way.

For your use case, would the event without a payload be good enough? Or do you really need the context/account info here? As previously stated, you can just call getActiveAccount when the event is emitted.

lib/msal-browser/src/event/EventHandler.ts
hectormmg: It's possible that the account object in the cache won't change (for example, if there is a multi-tenant account and what changed was the active tenant, but both the previous active tenant profile and the new active tenant profile had already been cached).

Generally speaking if the event is triggered, the specific account/tenant profile that is active from MSAL's point of view changed. This means that the results of calling `getActiveAccount` before and after the event will be different.

tnorling (approved):

PR #6861, opened 2024-01-30
Instrument handleRedirectPromise()

konstantin-msft:
- Instrument handleRedirectPromise().

### DIFF
--- a/lib/msal-browser/src/controllers/StandardController.ts
+++ b/lib/msal-browser/src/controllers/StandardController.ts
@@ -347,7 +347,8 @@ export class StandardController implements IController {
             if (typeof response === "undefined") {
                 const request: NativeTokenRequest | null =
                     this.browserStorage.getCachedNativeRequest();
-                const useNative = request &&
+                const useNative =
+                    request &&
                     NativeMessageHandler.isNativeAvailable(
                         this.config,
                         this.logger,
@@ -355,14 +356,15 @@ export class StandardController implements IController {
                     ) &&
                     this.nativeExtensionProvider &&
                     !hash;
-                const correlationId =
-                    useNative ? request?.correlationId :
-                    this.browserStorage.getTemporaryCache(
-                        TemporaryCacheKeys.CORRELATION_ID,
-                        true
-                    ) || "";
+                const correlationId = useNative
+                    ? request?.correlationId
+                    : this.browserStorage.getTemporaryCache(
+                          TemporaryCacheKeys.CORRELATION_ID,
+                          true
+                      ) || "";
                 const rootMeasurement = this.performanceClient.startMeasurement(
-                    "acquireTokenRedirect", correlationId
+                    "acquireTokenRedirect",
+                    correlationId
                 );
                 this.eventHandler.emitEvent(
                     EventType.HANDLE_REDIRECT_START,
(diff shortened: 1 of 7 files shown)


lib/msal-browser/src/interaction_client/RedirectClient.ts
```diff
@@ -203,6 +207,12 @@ export class RedirectClient extends StandardInteractionClient {
                 this.logger.info(
                     "handleRedirectPromise called but there is no interaction in progress, returning null."
                 );
+                if (performanceClient && correlationId) {
+                    performanceClient?.addFields(
+                        { errorCode: "no_interaction_in_progress" },
```
tnorling: This will be emitted on every page load, perhaps multiple times and is 90% of the time expected. Is this desirable? Do we have a way to distinguish expected vs. unexpected?

lib/msal-browser/src/controllers/StandardController.ts
```diff
@@ -426,35 +449,50 @@ export class StandardController implements IController {
...
+                        if (eventError instanceof AuthError) {
+                            rootMeasurement.end({
+                                success: false,
+                                errorCode: eventError.errorCode,
+                                subErrorCode: eventError.subError,
+                            });
+                        } else {
+                            rootMeasurement.end({
+                                success: false,
+                                errorCode: eventError?.message,
```
tnorling: I've found previously that tracking arbitrary error messages like this often results in long, unique strings and may even contain PII (think JSON.parse error for example that may leak the string that failed parsing)

konstantin-msft: That's a good point, thanks. Removed it.

lib/msal-browser/src/interaction_client/RedirectClient.ts
konstantin-msft: Thanks for the context, appreciate it. I am not sure I can come up with a way to distinguish expected vs unexpected. I don't think we can use `hash` as part of the unexpected condition. Happy to remove this one for now.

tnorling: You could check if hash (which would be hash passed in OR window.location.hash/search) contains known properties (there's a helper function that does this)

That would give you a pretty good idea, however, downside is that you're now parsing the url hash on every page load which could be expensive. Since this is adding some level of perf tracking, maybe we try it and if it proves to be too expensive reconsider.

konstantin-msft: Checking hash for known properties could be an option, but i would rather avoid doing that. To me the perf regression outweighs the potential tracking benefits.

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
```diff
@@ -320,15 +320,26 @@ export class NativeInteractionClient extends BaseInteractionClient {
...
         this.logger.trace(
             "NativeInteractionClient - handleRedirectPromise called."
         );
         if (!this.browserStorage.isInteractionInProgress(true)) {
             this.logger.info(
                 "handleRedirectPromise called but there is no interaction in progress, returning null."
             );
+            if (performanceClient && correlationId) {
+                performanceClient?.addFields(
+                    { errorCode: "no_interaction_in_progress" },
```
tnorling: Shall we remove this one as well?

tnorling (approved):

lib/msal-browser/src/interaction_client/NativeInteractionClient.ts
konstantin-msft: Removed this as well

hectormmg (approved):

PR #6864, opened 2024-02-01
Fix electron webpack samples

tnorling:
The electron webpack samples' transforms weren't updated when we changed to .cjs extensions which breaks native module imports (node extensions). This PR addresses by adding .cjs to the list of files to transform


peterzenz (approved):

jo-arroyo (approved):
Thanks for looking into this! I also noticed that the file path for the `build:package` script in the package.json was out of date. Wonder if it's worth updating that in this PR too?

tnorling:
> Thanks for looking into this! I also noticed that the file path for the `build:package` script in the package.json was out of date. Wonder if it's worth updating that in this PR too?

Thanks for calling that out, addressed!

PR #6867, opened 2024-02-02
Add nonce to SilentRequest

jo-arroyo:
Making `nonce` available and optional on SilentRequest.ts


konstantin-msft (approved):

lib/msal-browser/src/request/SilentRequest.ts
```diff
@@ -47,4 +47,5 @@ export type SilentRequest = Omit<
     prompt?: string;
     state?: string;
     tokenBodyParameters?: StringDict;
+    nonce?: string;
```
tnorling, line 50: What's the behavior when a token is served from cache?

hectormmg: +1, should nonce be added as a lookup filter?

tnorling: Might need to be treated as a force-iframe refresh as it only gets passed to the /authorize endpoint. @jo-arroyo What was the scenario we're trying to unblock with this?

jo-arroyo: @tnorling This is being used in the benchmark tests. When we last talked about this, I was told not being able to pass this in SilentRequest was an oversight?

jo-arroyo:
Workaround found offline. Closing as no longer needed.

PR #6873, opened 2024-02-05
Update package-lock to fix react build

hectormmg:


konstantin-msft (approved):

lib/msal-react/package.json
```diff
@@ -69,5 +69,8 @@
     "ts-jest": "^29.1.0",
     "tslib": "^2.0.0",
     "typescript": "^4.9.5"
+  },
+  "optionalDependencies": {
+    "@rollup/rollup-linux-x64-gnu": "^4.9.5"
```
tnorling: Interesting...why is this just now becoming an issue? Did the VM Image change to something unsupported?

optionalDependencies will be attempted to be installed by customers, we should make this a devDependency or install it ad-hoc in the pipeline itself instead

hectormmg: There's an npm bug, I'm still debugging it though

hectormmg: This isn't final, just had to open the PR to run the checks

hectormmg: https://github.com/npm/cli/issues/4828

PR #6880, opened 2024-02-06
Managed Identity in MSALJS

### LINKED ISSUE #6401: Managed Identity Support in MSAL JS
### Core Library

MSAL Node (@azure/msal-node)

### Wrapper Library

Not Applicable

Confidential

### Description

See example support in MSAL .NET - https://learn.microsoft.com/en-us/entra/msal/dotnet/advanced/managed-identity

More in-depth technical spec : https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=/ManagedIdentity/ManagedIdentity.md&version=GBdev&anchor=key-case-sensitivity 

Full task breakdown as per MSAL Java: https://github.com/orgs/AzureAD/projects/25

Azure Identity implementation: https://github.com/Azure/azure-sdk-for-js/tree/main/sdk/identity/identity/src/credentials/managedIdentityCredential

### Source

Internal (Microsoft)

Robbie-Microsoft:
Implemented Managed Identity in MSALJS. The following sources can be used to create a Managed Identity application:
- App Service
- Azure Arc
- Cloud Shell
- Imds
- Service Fabric

Tested all sources manually, and have unit tests which make fake network requests.


lib/msal-node/src/network/HttpClient.ts
```diff
@@ -20,13 +20,16 @@ import https from "https";
 export class HttpClient implements INetworkModule {
     private proxyUrl: string;
     private customAgentOptions: http.AgentOptions | https.AgentOptions;
+    private managedIdentity: boolean;
```
bgavrilMS: HttpClient should NOT know anything about Managed Identity. This is a violation of the Single Responsability Principle. 

At most, HTTPClient should know that it has to use the "http" module vs the "https" module. 

Can't you just look at the request URL to tell if the request URL starts with "http" and not "https" ?

lib/msal-node/src/config/Configuration.ts
```diff
@@ -199,3 +214,34 @@ export function buildAppConfiguration({
...
+
+    const systemOptions: Required<NodeSystemOptions> = {
+        ...DEFAULT_SYSTEM_OPTIONS,
+        loggerOptions: system?.loggerOptions || DEFAULT_LOGGER_OPTIONS,
+        networkClient: new HttpClient(
+            system?.proxyUrl,
+            system?.customAgentOptions as
+                | http.AgentOptions
+                | https.AgentOptions,
+            true // Managed Identity
```
bgavrilMS: Please don't leak MSI details in network client.

lib/msal-node/src/client/ManagedIdentityClient.ts
```diff
@@ -0,0 +1,128 @@
...
+ * Class to initialize a managed identity and identify the service.
+ * Original source of code: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/identity/Azure.Identity/src/ManagedIdentityClient.cs
+ */
+export class ManagedIdentityClient {
+    private logger: Logger;
+    private nodeStorage: NodeStorage;
+    private networkClient: INetworkModule;
+    private cryptoProvider: CryptoProvider;
+
+    static identitySource:
```
bgavrilMS: Pls add visibility modifier (private?) all the time.

lib/msal-node/src/client/ManagedIdentityClient.ts
```diff
@@ -0,0 +1,128 @@
...
+        | ServiceFabric
+        | AppService
+        | CloudShell
+        | AzureArc
+        | Imds
+        | undefined;
```
bgavrilMS: Please stop using "undefined" and use null checks. https://stackoverflow.com/questions/5076944/what-is-the-difference-between-null-and-undefined-in-javascript

An identity source exists, but the variable can be null at first, before it is initialized.

lib/msal-node/src/client/ManagedIdentityClient.ts
```diff
@@ -0,0 +1,128 @@
...
+ * Original source of code: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/identity/Azure.Identity/src/ManagedIdentityClient.cs
+ */
+export class ManagedIdentityClient {
+    private logger: Logger;
+    private nodeStorage: NodeStorage;
+    private networkClient: INetworkModule;
+    private cryptoProvider: CryptoProvider;
+
+    static identitySource:
+        | ServiceFabric
```
bgavrilMS: Don't enumerate the classes here. Use `BaseManagedIdentitySource` instead - the base class. That's the power of polymorphism.

lib/msal-node/src/client/ManagedIdentityClient.ts
Robbie-Microsoft: I will stop using it in this case, but there are other cases in this PR where the variables can be undefined.

neha-bhargava (approved):

neha-bhargava (approved):

bgavrilMS (approved):

pmaytak (approved):
Approving in lieu of Neha's approval.

PR #6885, opened 2024-02-07
Implemented Network Request Retry-Policy in MSAL-Node & Managed Identity

Robbie-Microsoft:
I am waiting to hear feedback on this implementation before I write the unit tests.

Closing in favor of other implementation: https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/6891

### DIFF
--- a/lib/msal-node/src/client/ManagedIdentityApplication.ts
+++ b/lib/msal-node/src/client/ManagedIdentityApplication.ts
@@ -35,6 +35,10 @@ import {
     ManagedIdentityErrorCodes,
     createManagedIdentityError,
 } from "../error/ManagedIdentityError";
+import { RetryTracker } from "../utils/RetryTracker";
+
+const RETRY_TRACKER_RETRY_LIMIT = 1;
+const RETRY_TRACKER_WAIT_TIME_IN_MILLISECONDS = 1000;
 
 /**
  * Class to initialize a managed identity and identify the service
@@ -46,6 +50,7 @@ export class ManagedIdentityApplication {
     private static nodeStorage: NodeStorage | undefined;
     private networkClient: INetworkModule;
     private cryptoProvider: CryptoProvider;
+    private retryTracker: RetryTracker;
 
     // authority needs to be faked to re-use existing functionality in msal-common: caching in responseHandler, etc.
     private fakeAuthority: Authority;
@@ -82,6 +87,12 @@ export class ManagedIdentityApplication {
 
         this.cryptoProvider = new CryptoProvider();
 
+        this.retryTracker = new RetryTracker(
+            this.logger,
+            RETRY_TRACKER_RETRY_LIMIT,
+            RETRY_TRACKER_WAIT_TIME_IN_MILLISECONDS
+        );
+
         const fakeAuthorityOptions: AuthorityOptions = {
             protocolMode: ProtocolMode.AAD,
             knownAuthorities: [DEFAULT_AUTHORITY_FOR_MANAGED_IDENTITY],
@@ -94,7 +105,7 @@ export class ManagedIdentityApplication {
             ManagedIdentityApplication.nodeStorage as NodeStorage,
             fakeAuthorityOptions,
             this.logger,
-            undefined,
+            this.config.managedIdentityId.id,
             undefined,
             true
         );
@@ -110,7 +121,8 @@ export class ManagedIdentityApplication {
             this.logger,
             ManagedIdentityApplication.nodeStorage as NodeStorage,
             this.networkClient,
-            this.cryptoProvider
+            this.cryptoProvider,
+            this.retryTracker
         );
     }
 
--- a/lib/msal-node/src/client/ManagedIdentityClient.ts
+++ b/lib/msal-node/src/client/ManagedIdentityClient.ts
@@ -22,6 +22,7 @@ import {
 import { ManagedIdentityRequest } from "../request/ManagedIdentityRequest";
 import { ManagedIdentityId } from "../config/ManagedIdentityId";
 import { NodeStorage } from "../cache/NodeStorage";
+import { RetryTracker } from "../utils/RetryTracker";
 
 /*
  * Class to initialize a managed identity and identify the service.
@@ -32,6 +33,7 @@ export class ManagedIdentityClient {
     private nodeStorage: NodeStorage;
     private networkClient: INetworkModule;
     private cryptoProvider: CryptoProvider;
+    private retryTracker: RetryTracker;
 
     static identitySource:
         | ServiceFabric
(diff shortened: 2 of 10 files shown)


lib/msal-node/src/utils/RetryTracker.ts
```diff
@@ -0,0 +1,74 @@
...
+ */
+
+import { Logger } from "@azure/msal-common";
+
+export class RetryTracker {
+    private logger: Logger;
+    private retryLimit: number;
+    private retryWaitTimeMilli: number;
+
+    private _numRetried: number;
```
bgavrilMS, line 13: inconsistent naming - all private fields should be named in the same way. 
Also, they can probably be readonly?

lib/msal-node/src/utils/RetryTracker.ts
```diff
@@ -0,0 +1,74 @@
...
+        if (!retryAfterHeader) {
+            this.logger.warning(
+                "The request can not be retried because the original network response contains the Retry-After header."
+            );
+            return false;
+        }
+
+        let retriableStatusCode: boolean = false;
+        switch (statusCode) {
+            case 404: // Not Found
```
bgavrilMS, line 50: These should not be hardcoded. Use the ctor.

lib/msal-node/src/utils/RetryTracker.ts
```diff
@@ -0,0 +1,74 @@
...
+        retryAfterHeader: string | undefined
+    ): Promise<boolean> {
+        if (this._numRetried === this.retryLimit) {
+            this.logger.warning(
+                "The request can not be retried because it has already been retried the maximum number of times."
+            );
+            return false;
+        }
+
+        if (!retryAfterHeader) {
```
bgavrilMS, line 41: @neha-bhargava @Avery-Dunn  - what are we doing in MSALs when IMDS returns a RetryAfter header? IMO we should respect this header, but still retry.

lib/msal-node/src/utils/RetryTracker.ts
```diff
@@ -0,0 +1,74 @@
...
+
+import { Logger } from "@azure/msal-common";
+
+export class RetryTracker {
+    private logger: Logger;
+    private retryLimit: number;
+    private retryWaitTimeMilli: number;
+
+    private _numRetried: number;
+    public resetNumRetried(): void {
```
bgavrilMS, line 14: Is this needed?

lib/msal-node/src/client/ManagedIdentityApplication.ts
```diff
@@ -46,6 +50,7 @@ export class ManagedIdentityApplication {
     private static nodeStorage: NodeStorage | undefined;
     private networkClient: INetworkModule;
     private cryptoProvider: CryptoProvider;
+    private retryTracker: RetryTracker;
```
bgavrilMS, line 53: Does it need to be a member of this class? It looks to me that it's more of a method variable?

lib/msal-node/src/client/ManagedIdentityClient.ts
```diff
@@ -32,6 +33,7 @@ export class ManagedIdentityClient {
     private nodeStorage: NodeStorage;
     private networkClient: INetworkModule;
     private cryptoProvider: CryptoProvider;
+    private retryTracker: RetryTracker;
```
bgavrilMS, line 36: I don't think the lifetime of this object is correct. There should be 1 retry possible for each HTTP call. I think here you will have just 1 try per ManagedIdentityClient, not per request.

lib/msal-node/src/client/ManagedIdentitySources/BaseManagedIdentitySource.ts
```diff
@@ -154,6 +142,63 @@ export abstract class BaseManagedIdentitySource {
             }
         }
 
+        let serverTokenResponse: ServerAuthorizationTokenResponse =
```
bgavrilMS, line 145: It seems to me that the RetryTracker object can be created directly in this method only. There's no need to keep it as class member in all these classes.

bgavrilMS (commented):
This is a good approach, the "retry tracker" lifetime needs to be adjusted etc. 

I want to propose an alternative, and then we can contrast and compare, knowing that: 

- the retry logic is likely to change (e.g. obey the RetryAfter header)
- it's likely that we will add RetryLogic for other flows, e.g. ConfidentialClient flows, but the logic will be different
- it's likely to we will need to provide app developers a chance to disable retries completely (for them to add their own)



# HttpClientWithRetries

Idea: the retries happen at HTTP layer, so it makes sense for the HTTP layer to handle the retries. 

Http extensibility point (INetworkModule) should not change.

### Scenarios

- MSI retry N times after X amount of time
- MSI retry M times after reading RetryAfter header (not shown here)


### Design

Let's see if we can use the Decorator pattern https://www.tutorialspoint.com/design_pattern/decorator_pattern.htm

See https://gist.github.com/bgavrilMS/ea89ad64c0204193cd62a1a2fca7d50a for the code.

bgavrilMS:
@trwalke - can you also take a look at this design and at my alternative proposal, as we'll want to do smth similar in MSAL.NET as well.

lib/msal-node/src/utils/RetryTracker.ts
neha-bhargava: @bgavrilMS I don't think IMDS or any other managed identity endpoints send back a retry-after header.

lib/msal-node/src/utils/RetryTracker.ts
Robbie-Microsoft: If you're talking about the void method, I use it in BaseManagedIdentitySource on line 173 when the retry limit has been reached (if applicable). However I like your idea about making the RetryTracker local to the network request method in BaseManagedIdentitySource which might eliminate the need for this.

lib/msal-node/src/client/ManagedIdentityApplication.ts
Robbie-Microsoft: You're right, it doesn't need to be.

lib/msal-node/src/client/ManagedIdentityClient.ts
Robbie-Microsoft: Correct, but this is where the resetNumRetried() method comes into play. I reset the counter of retriedRequests to 0 after each request (retried request, if applicable).

lib/msal-node/src/client/ManagedIdentitySources/BaseManagedIdentitySource.ts
Robbie-Microsoft: I like this idea. My original idea was to create it just once, and use a setter for resetting the number of times retried to 0 after each request (retried request, if applicable).

Robbie-Microsoft:
Closing in favor of other implementation: [HttpClient-Decorator](https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/6891)

PR #6891, opened 2024-02-13
Implemented HttpClient-Decorator Retry-Policy in MSAL-Node & Managed Identity

### LINKED ISSUE #5128: MSAL retry policy
### Core Library

MSAL Node (@azure/msal-node)

### Wrapper Library

Not Applicable

Confidential

### Description

What is the guidance on retry policy for MSAL node?

### Source

Internal (Microsoft)

## Spec

MSAL throws several types of exceptions: 

- base class is [AuthError](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/72fdd1fe926d72422caff139cf5abe2e31cef9fb/lib/msal-common/src/error/AuthError.ts#L25) 
- client errors come from the SDK itself [ClientAuthError](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/72fdd1fe926d72422caff139cf5abe2e31cef9fb/lib/msal-common/src/error/ClientAuthError.ts#L212)
- client configuration errors are a subclass of client errors [ClientConfigurationError](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/72fdd1fe926d72422caff139cf5abe2e31cef9fb/lib/msal-common/src/error/ClientConfigurationError.ts#L109)
- [ServerError](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/72fdd1fe926d72422caff139cf5abe2e31cef9fb/lib/msal-common/src/error/ServerError.ts#L11) is for when AAD (the service) returns an error as part of a 400 / 401 response. See [here](https://learn.microsoft.com/en-us/azure/active-directory/develop/reference-aadsts-error-codes) for some of the AAD error codes. 
- [UiRequiredError](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/72fdd1fe926d72422caff139cf5abe2e31cef9fb/lib/msal-common/src/error/InteractionRequiredAuthError.ts#L42) for when interaction is required
- other errors which the library does not re-package. For example, if the request to AAD returns a 500 error ("server error"), MSAL will not catch this (?). Needs confirmation @Robbie-Microsoft 

## Retriable exceptions

- server responses which return with HTTP status codes 408 (unlikely to be seen here) , 429, 5xx 
Important: If a "retry-after" header is present, it must be observed (e.g. 500 error + Retry-After: 5)
- transient HTTP errors, such as the box losing connectivity temporarily (to test, just pull the internet cable out :) )
- in the future, some of the AAD exceptions, but for now we don't have this, so all other `ServerError`s should be seen as non-retry-able.

## Task list

- [x] analyze what exceptions occur if server returns 429, 500, 501 etc. 
- [x] identify other transient HTTP exceptions
- [x] ensure that the app developer can read the HttpStatus and Headers of exceptions, particularly for 429, 500 etc.
- [ ] create a simple retry policy as example / documentation 

CC @Robbie-Microsoft @jmprieur

Robbie-Microsoft:


lib/msal-node/src/client/ManagedIdentityApplication.ts
```diff
@@ -94,7 +94,7 @@ export class ManagedIdentityApplication {
             ManagedIdentityApplication.nodeStorage as NodeStorage,
             fakeAuthorityOptions,
             this.logger,
-            undefined,
+            this.config.managedIdentityId.id,
```
Robbie-Microsoft: needed for changes made to msal-common for a bugfix

lib/msal-node/test/client/ManagedIdentitySources/AzureArc.spec.ts
```diff
@@ -143,10 +143,10 @@ describe("Acquires a token successfully via an Azure Arc Managed Identity", () =
                 DEFAULT_SYSTEM_ASSIGNED_MANAGED_IDENTITY_AUTHENTICATION_RESULT.accessToken
             );
 
-            expect(sendGetRequestAsyncSpy).toBeCalledTimes(2);
-            expect(readFileSyncSpy).toBeCalledTimes(1);
+            expect(sendGetRequestAsyncSpy).toHaveBeenCalledTimes(2);
+            expect(readFileSyncSpy).toHaveBeenCalledTimes(1);
 
-            expect(sendGetRequestAsyncSpy).nthCalledWith(
+            expect(sendGetRequestAsyncSpy).toHaveBeenNthCalledWith(
```
Robbie-Microsoft, line 149: old versions of methods are deprecated

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -192,7 +192,7 @@ export class ResponseHandler {
             if (
                 refreshAccessToken &&
                 serverResponse.status &&
-                serverResponse.status >= HttpStatus.SERVER_ERROR_RANGE_START &&
+                serverResponse.status >= HttpStatus.INTERNAL_SERVER_ERROR &&
```
bgavrilMS: Inconsistent with the "SERVER_ERROR_RANGE_END". You can have 2 constants for 500. Or just hardcode these error codes, they are well known

lib/msal-node/src/client/ManagedIdentityApplication.ts
bgavrilMS: What is the bug fix?

lib/msal-node/test/test_kit/ManagedIdentityTestUtils.ts
```diff
@@ -96,7 +96,7 @@ export class ManagedIdentityNetworkClient implements INetworkModule {
     ): Promise<NetworkResponse<T>> {
         return new Promise<NetworkResponse<T>>((resolve, _reject) => {
             resolve({
-                status: 200,
+                status: HttpStatus.SUCCESS_RANGE_START,
```
bgavrilMS: I much preferred 200. Easy to read.

lib/msal-node/test/test_kit/ManagedIdentityTestUtils.ts
```diff
@@ -179,6 +179,18 @@ export class ManagedIdentityNetworkErrorClient implements INetworkModule {
         });
     }
 
+    sendGetRequestForRetryAsync<T>(): Promise<NetworkResponse<T>> {
+        return new Promise<NetworkResponse<T>>((resolve, _reject) => {
+            resolve({
+                status: HttpStatus.INTERNAL_SERVER_ERROR,
+                body: {
+                    message: MANAGED_IDENTITY_TOKEN_RETRIEVAL_ERROR,
+                    correlationId: DEFAULT_MANAGED_IDENTITY_ID,
```
bgavrilMS: Correlation ID is not managed identity ID. Pls use a new GUID.

lib/msal-node/test/client/ManagedIdentitySources/Imds.spec.ts
```diff
@@ -170,6 +173,130 @@ describe("Acquires a token successfully via an IMDS Managed Identity", () => {
...
+                expect(serverError.errorCode).toEqual(
+                    MANAGED_IDENTITY_TOKEN_RETRIEVAL_ERROR
+                );
+                expect(sendGetRequestAsyncSpy).toHaveBeenCalledTimes(2);
+
+                jest.restoreAllMocks();
+            });
+        });
+
+        describe("System Assigned", () => {
```
bgavrilMS, line 241: I don't think these tests add anything new. Leave them since you coded them up, but the code path is the same for UAMI and SAMI in relation to HTTP

lib/msal-node/test/client/ManagedIdentitySources/Imds.spec.ts
```diff
@@ -170,6 +173,130 @@ describe("Acquires a token successfully via an IMDS Managed Identity", () => {
...
+                managedIdentityApplication = new ManagedIdentityApplication(
+                    systemAssignedConfig
+                );
+            });
+
+            test("returns a 500 error response from the network request, just the first time", async () => {
+                expect(ManagedIdentityTestUtils.isIMDS()).toBe(true);
+
+                const sendGetRequestAsyncSpy: jest.SpyInstance = jest
+                    .spyOn(networkClient, <any>"sendGetRequestAsync")
```
bgavrilMS, line 253: Wouldn't it be better to use regular HTTP level mocking instead of mocking the API? With HTTP level mocking, you also test that your change does not affect the ability of the app developer to use their own `INetworkModule`

lib/msal-node/src/client/ManagedIdentitySources/BaseManagedIdentitySource.ts
```diff
@@ -57,8 +60,27 @@ export abstract class BaseManagedIdentitySource {
...
+            HttpStatus.INTERNAL_SERVER_ERROR,
+            HttpStatus.SERVICE_UNAVAILABLE,
+            HttpStatus.SERVER_ERROR_RANGE_END,
+        ];
+        const linearRetryPolicy: LinearRetryPolicy = new LinearRetryPolicy(
+            maxRetries,
+            retryDelay,
+            httpStatusCodesToRetryOn
+        );
+        this.networkClientWithRetry = new HttpClientWithRetries(
```
bgavrilMS: What is the lifetime of BaseManagedIdentitySource object in relation to token requests? 
Will each token request have create a `BaseManagedIdentitySource` or is this object reused by all requests? If it's reused, then the code is wrong, because the HttpClient retains the "current retry" as state.

Irrespective of the answer, there needs to be a test for this: 

- create a ManagedIdentityApp 
- configure the HTTP stack so that it always returns 500
- request a token for resource R1. Assert that 2 HTTP calls are made.
- request a token for resource R2. Assert that 2 HTTP calls are made.

bgavrilMS (changes requested):
Needs another test to ensure retry policy lifetime is per request (and possible code changes)

lib/msal-node/src/client/ManagedIdentityApplication.ts
Robbie-Microsoft: In Authority.ts, the correlationId's type changed from `undefined | string`, to just `string`

lib/msal-common/src/response/ResponseHandler.ts
Robbie-Microsoft: I don't think we should hardcode. No "magic numbers", right? (even though it's obvious what 500 means)

lib/msal-node/test/test_kit/ManagedIdentityTestUtils.ts
Robbie-Microsoft: I agree... but don't we want to avoid "magic numbers", no matter how obvious 200 is?

lib/msal-node/src/retry/LinearRetryPolicy.ts
```diff
@@ -0,0 +1,39 @@
...
+    async pauseForRetry(
+        status: number,
+        currentRetry: number
+    ): Promise<boolean> {
+        if (
+            this.httpStatusCodesToRetryOn.includes(status) &&
+            currentRetry < this.maxRetries
+        ) {
+            await new Promise((resolve) =>
+                setTimeout(resolve, this.retryDelay)
```
Robbie-Microsoft: future proof: `retry-after header || this.retryDelay`

lib/msal-node/test/client/ManagedIdentitySources/Imds.spec.ts
Robbie-Microsoft: I'm really confused by this comment. I AM mocking the http request. In fact, that's the only thing I'm mocking here: the return value from the http request. I do the same thing in the Azure Arc tests.

lib/msal-node/src/client/ManagedIdentitySources/BaseManagedIdentitySource.ts
Robbie-Microsoft: ~~The lifetime of BaseManagedIdentitySource object is once per initial request (used by all retries). It's not used across all requests.~~

~~I will create the tests you suggested to confirm this.~~

Robbie-Microsoft: Expose a setting for app developers to disable internal retries.

lib/msal-common/src/response/ResponseHandler.ts
Robbie-Microsoft: I do like the idea of having two different constants for 500

lib/msal-node/test/client/ManagedIdentitySources/Imds.spec.ts
Robbie-Microsoft: Using the same Managed Identity Application, I need to mock one of the return values to simulate a 500 so that the request is retried.

lib/msal-node/src/client/ManagedIdentitySources/BaseManagedIdentitySource.ts
Robbie-Microsoft: I've been looking into this.

The `BaseManagedIdentitySource` is the static `identitySource` in `ManagedIdentityClient`, with the `ManagedIdentityClient` created per `ManagedIdentityApplication`. So `BaseManagedIdentitySource` (`identitySource`) is defined upon the first network request of each application, and is used across applications (This is desired behavior per our previous discussions and current implementation).

So according to your comments, this makes the code in this PR wrong. What do you mean by that, can you elaborate? Should the `networkClientWithRetry` be defined in `ManagedIdentityApplication` before `acquireToken` is called?

lib/msal-node/src/client/ManagedIdentityApplication.ts
bgavrilMS: That's not a correlation ID, that's a managed identity id.

lib/msal-node/src/client/ManagedIdentitySources/BaseManagedIdentitySource.ts
bgavrilMS: If `identitySource` is static, then it means that all requests share it and so the `this.networkClientWithRetry` is also shared, and so the "current retry" property is also shared etc. So once a request exhausts the retries, new requests will no longer retry.

I think you just need to write a unit test and you'll figure out why it's wrong. 


1. create a ManagedIdentity setup (public API only)
2. configure the HTTP stack so that it always returns 500
3. request a token for resource R1. 

**Expected**: Assert that 2 HTTP calls are made.

4. request a token for resource R2. 

**Expected**:  2 HTTP calls are made.
**Actual**: I believe in your case, only 1 HTTP call will be made here.

bgavrilMS: This should to be done separately.

lib/msal-node/src/network/HttpClientWithRetries.ts
```diff
@@ -0,0 +1,68 @@
...
+    async sendPostRequestAsync<T>(
+        url: string,
+        options?: NetworkRequestOptions
+    ): Promise<NetworkResponse<T>> {
+        // the underlying network module (custom or HttpClient) will make the call
+        let response: NetworkResponse<T> =
+            await this.httpClientNoRetries.sendPostRequestAsync(url, options);
+
+        let currentRetry: number = 0;
+        while (
```
bgavrilMS, line 49: nit: quite a lot of duplicate code between these 2 methods. Can you use a private mehtod to eliminate dup?

lib/msal-node/src/retry/LinearRetryPolicy.ts
bgavrilMS: Is this a TODO comment ?

lib/msal-node/test/client/ManagedIdentitySources/Imds.spec.ts
bgavrilMS: It's confusing because you are mocking `sendGetRequestAsync` - but which `sendGetRequestAsync` is it? 

- the httpClientWithNoRetries?
- the httpClientWithRetries?

bgavrilMS: Ok, I looked at this and it's ok. I was thinking that the `managedIdentityNetworkErrorClient` can have a list (or a queue) of responses and your test can just check how many responses have been served. This is ok too.

lib/msal-node/src/client/ManagedIdentityApplication.ts
Robbie-Microsoft: Correct. This is for the fake Authority, which is used to create a fake ClientCredentialClient... which is needed to re-use cache functionality. Any string value can be passed in.

lib/msal-node/src/retry/LinearRetryPolicy.ts
Robbie-Microsoft: yes

lib/msal-node/src/client/ManagedIdentitySources/BaseManagedIdentitySource.ts
Robbie-Microsoft: https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/6901

lib/msal-node/src/client/ManagedIdentityApplication.ts
bgavrilMS: Every request should have a correlation ID, even if (for now) we can't inject the correlation ID into MSI. My 2c - just generate a GUID for each request. And ensure each log statement logs it.

lib/msal-node/src/network/HttpClientWithRetries.ts
```diff
@@ -22,41 +25,64 @@ export class HttpClientWithRetries implements INetworkModule {
...
 
-        let currentRetry: number = 0;
-        while (
-            await this.retryPolicy.pauseForRetry(response.status, currentRetry)
-        ) {
-            response = await this.httpClientNoRetries.sendGetRequestAsync(
-                url,
-                options
+        // retry-after header is in seconds
+        let millisToSleep = Math.round(parseFloat(retryHeader) * 1000);
```
bgavrilMS: Is it possible to move this logic inside the retry policy itself? The HttpClientWithRetries should onyl be responsible for waiting N ms and retrying. It should figure out N (how long to wait) from the retry policy itself.

lib/msal-common/src/utils/Constants.ts
```diff
@@ -65,12 +65,20 @@ export const Constants = {
 
 export const HttpStatus = {
     SUCCESS_RANGE_START: 200,
+    SUCCESS: 200,
     SUCCESS_RANGE_END: 299,
     REDIRECT: 302,
     CLIENT_ERROR_RANGE_START: 400,
+    BAD_REQUEST: 400,
```
neha-bhargava, line 72: I don't think this is needed for managed identity

lib/msal-node/src/network/HttpClientWithRetries.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    constructor(
+        httpClientNoRetries: INetworkModule,
+        retryPolicy: IHttpRetryPolicy
+    ) {
+        this.httpClientNoRetries = httpClientNoRetries;
+        this.retryPolicy = retryPolicy;
+    }
+
+    private retryAfterMillisecondsToSleep(
```
neha-bhargava: In case we use the retry header to make retries in MSAL what will happen if there are timeouts setup? There are cancellation tokens in dotnet, can user set a timeout for an MSAL request in JS?

Robbie-Microsoft: Yes, that functionality already exists in msal-node: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/src/network/HttpClient.ts#L67

Robbie-Microsoft: currently the cancellationToken functionality only exists for post requests

lib/msal-node/test/client/ManagedIdentitySources/Imds.spec.ts
```diff
@@ -170,6 +175,224 @@ describe("Acquires a token successfully via an IMDS Managed Identity", () => {
         });
     });
 
+    describe("Linear Retry Policy", () => {
```
neha-bhargava, line 179: Also add tests to assert that the retry does not happen for the scenarios when retry policy is not setup. For example, in case of 400 the request fails but a retry does not happen.

lib/msal-node/src/network/HttpClientWithRetries.ts
Robbie-Microsoft: I'm going to make another GitHub issue for this because there appears to be some inconsistencies in the msal-node code.

neha-bhargava: Should we honor the retry header in this PR? Since managed identity responses does not contain a retry header, these are more relevant to calls to ests. Also if we implement the retry header we will need to also have the logic to honor the timeout/ cancellation tokens. IMO break the PR to have the logic for retry-after header with ests requests. 
Thoughts? @bgavrilMS

lib/msal-common/src/utils/Constants.ts
Robbie-Microsoft: It's only used for testing. I don't want to use a specific error message, but just a general client error.

lib/msal-node/test/client/ManagedIdentitySources/Imds.spec.ts
Robbie-Microsoft: Right now, there's no way to disable the retry policy. I'll add this unit test to the PR for [Add an option to disable MSAL Node's internal retry policy.](https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/6901)

lib/msal-node/src/retry/IHttpRetryPolicy.ts
```diff
@@ -13,4 +15,8 @@ export interface IHttpRetryPolicy {
         currentRetry: number,
         retryAfterHeader: number
     ): Promise<boolean>;
+
+    retryAfterMillisecondsToSleep(
```
bgavrilMS: There is no need to expose this method on the interface, as only the retry policy uses it. Can you avoid exposing it on the interface?

The retry policy object has access to the "Retry-After" header and can compute the number of seconds to pause on its own.

lib/msal-node/src/retry/LinearRetryPolicy.ts
```diff
@@ -39,4 +40,29 @@ export class LinearRetryPolicy implements IHttpRetryPolicy {
 
         return false;
     }
+
+    retryAfterMillisecondsToSleep(
```
bgavrilMS: This can be called directly by `pauseForRetry`

bgavrilMS: nit: Please add explicit visibility modifiers (public, private) to help reviewers who are not familiar with TS review.

bgavrilMS (changes requested):
Keep IHttpRetryPolicy with 1 method.

lib/msal-node/src/retry/IHttpRetryPolicy.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import http from "http";
+
+export interface IHttpRetryPolicy {
+    /*
+     * if retry conditions occur, pauses and returns true
+     * otherwise return false
+     */
+    pauseForRetry(
+        status: number,
```
bgavrilMS: rename to httpStatusCode for readability.

lib/msal-node/src/retry/IHttpRetryPolicy.ts
```diff
@@ -0,0 +1,22 @@
...
+        currentRetry: number,
+        retryAfterHeader: number
```
bgavrilMS: A header is not a number. Why don't you pass in all the headers from the response? Or better yet, the entire HttpResponse? And then the retry policy can extract what it needs from the response object.

lib/msal-node/src/client/ManagedIdentityApplication.ts
```diff
@@ -82,6 +84,8 @@ export class ManagedIdentityApplication {
 
         this.cryptoProvider = new CryptoProvider();
 
+        this.correlationId = this.cryptoProvider.createNewGuid();
```
bgavrilMS: Is there an API for app developer to pass in his own correlation ID?

lib/msal-node/src/client/ManagedIdentityApplication.ts
```diff
@@ -143,7 +147,7 @@ export class ManagedIdentityApplication {
                 managedIdentityRequestParams.resource.replace("/.default", ""),
             ],
             authority: this.fakeAuthority.canonicalAuthority,
-            correlationId: this.config.managedIdentityId.id,
+            correlationId: this.correlationId,
```
bgavrilMS: How does a log look like for an MSI token? 

Correlation ID should be per request, not per object.

lib/msal-node/test/client/ManagedIdentitySources/Imds.spec.ts
```diff
@@ -170,6 +175,224 @@ describe("Acquires a token successfully via an IMDS Managed Identity", () => {
...
+        describe("User Assigned", () => {
+            let managedIdentityApplication: ManagedIdentityApplication;
+            beforeEach(() => {
+                managedIdentityApplication = new ManagedIdentityApplication(
+                    userAssignedClientIdConfig
+                );
+            });
```
bgavrilMS, line 186: Good unit tests!

lib/msal-node/test/client/ManagedIdentitySources/Imds.spec.ts
neha-bhargava: The retry policy retries for specific error codes. I am recommending to add a test for a scenario where the retry policy would not retry and fail.

lib/msal-node/src/network/HttpClientWithRetries.ts
bgavrilMS: It's fine, @Robbie-Microsoft already implemented and unit test this ( I agree it should have been a different PR). 

We discussed that all retry policies should obey Retry-After header.

lib/msal-node/src/retry/LinearRetryPolicy.ts
Robbie-Microsoft: The visibility of all properties or methods in Typescript classes is “public“ by default. If I add it here, we should have a strategy to add it to all existing code too.

lib/msal-node/src/retry/IHttpRetryPolicy.ts
Robbie-Microsoft: This comment is moot after addressing your other feedback about restructuring the methods and removing ```retryAfterMillisecondsToSleep``` from the interface.

lib/msal-node/test/client/ManagedIdentitySources/Imds.spec.ts
Robbie-Microsoft: :)

lib/msal-node/src/client/ManagedIdentityApplication.ts
Robbie-Microsoft: It appears we didn't implement that in Managed Identity. Unlike all other msal-node clients, Managed Identity creates the request (```ManagedIdentityRequest```) inside of the ```acquireToken``` call, instead of passing in the request. We currently only pass in ```ManagedIdentityRequestParams``` (forceRefresh?, resource) to the ```aquireToken``` call.

Thoughts? Was this an error on our part? Should we pass in the ```ManagedIdentityRequest``` to the ```acquireToken``` call instead of creating it inside of ```acquireToken```?

lib/msal-node/src/client/ManagedIdentityApplication.ts
Robbie-Microsoft: I think I understand. So: the fakeAuthority (needed for fakeClientCredentialClient, which is needed to re-use caching functionality) needs it's own correlationId. The fakeAuthority's correlationId should be different from each network request's correlationId (which will unique for each network request)?

lib/msal-node/src/client/ManagedIdentitySources/BaseManagedIdentitySource.ts
Robbie-Microsoft: I wrote a unit test for this scenario:

```
test("creates two different managed identity applications (that use the same HttpClient) which both return 500 error responses from their network requests permanently, to ensure that retry policy lifetime is per request", async () => {
                expect(ManagedIdentityTestUtils.isIMDS()).toBe(true);

                const networkClient2: ManagedIdentityNetworkClient =
                    new ManagedIdentityNetworkClient(
                        MANAGED_IDENTITY_RESOURCE_ID_2 // client id
                    );
                const managedIdentityApplication2: ManagedIdentityApplication =
                    new ManagedIdentityApplication({
                        system: {
                            networkClient: networkClient2,
                            // managedIdentityIdParams will be omitted for system assigned
                        },
                    });

                const sendGetRequestAsyncSpyApp1: jest.SpyInstance = jest
                    .spyOn(networkClient, <any>"sendGetRequestAsync")
                    // permanently override the networkClient's sendGetRequestAsync method to return a 500
                    .mockReturnValue(
                        managedIdentityNetworkErrorClient.sendGetRequestForRetryAsync()
                    );

                const sendGetRequestAsyncSpyApp2: jest.SpyInstance = jest
                    .spyOn(networkClient2, <any>"sendGetRequestAsync")
                    // permanently override the networkClient's sendGetRequestAsync method to return a 500
                    .mockReturnValue(
                        managedIdentityNetworkErrorClient.sendGetRequestForRetryAsync()
                    );

                try {
                    await managedIdentityApplication.acquireToken(
                        managedIdentityRequestParams
                    );
                } catch (e) {
                    expect(sendGetRequestAsyncSpyApp1).toHaveBeenCalledTimes(2);
                    expect(sendGetRequestAsyncSpyApp2).toHaveBeenCalledTimes(0);
                }

                try {
                    await managedIdentityApplication2.acquireToken(
                        managedIdentityRequestParams
                    );
                } catch (e) {
                    expect(sendGetRequestAsyncSpyApp1).toHaveBeenCalledTimes(2);
                    expect(sendGetRequestAsyncSpyApp2).toHaveBeenCalledTimes(2);
                }

                jest.restoreAllMocks();
            });
```

It fails in the 2nd catch, on the line ```expect(sendGetRequestAsyncSpyApp1).toHaveBeenCalledTimes(2);```. The spy is being called 4 times, instead of 2.

Since ```identitySource``` being static is desired behavior, then what is your proposed solution here? 

Should the `networkClientWithRetry` be defined in `ManagedIdentityApplication` inside of each `acquireToken` call? Add the retryPolicy to the `ManagedIdentityRequest` inside of `acquireToken`?

lib/msal-node/test/client/ManagedIdentitySources/Imds.spec.ts
Robbie-Microsoft: Oh, I see. Thanks for clarifying.

lib/msal-node/src/client/ManagedIdentityApplication.ts
bgavrilMS: Yeah, each request should have a different correlation ID. This helps you (MSAL developer) troubleshoot logs, particularly in multi-threaded environments where many requests occur at once. 

IMO, you should treat this as a separate issue. Open a task and associate it with MSI - to observe how "correlationID" shows up in the logs.

lib/msal-node/src/client/ManagedIdentityApplication.ts
bgavrilMS: As suggested on a different comment, for now pls log an issue for "dealing with correlation ID" and associate it with MSI. 

Implementation details on this isn't important as long as each request has it's own correlation ID (assigned by MSAL or given by app developer)

lib/msal-node/src/client/ManagedIdentitySources/BaseManagedIdentitySource.ts
bgavrilMS: It feels that you are unsure of the object lifetime for all these objects that you've created. And I think the unit tests don't help in understanding, because they are quite complex and I am not sure you **debug** through them. Debugging is key.

Here's a suggestion to help with understanding - create a new project and add the MSI classes to it, but with very simplified logic. No async, no helpers etc. Just constructor and public methods and return dumb values. Something like this: https://gist.github.com/bgavrilMS/fd70a44bca9ad194d2d453c4e7b3669d

The key aspects you want to capture is when objects are constructed (e.g. LinearPolicy = new LinearPolicy)

And then run everything with a debugger and understand the lifetime of each object, how the interact with each other etc.

bgavrilMS: Also, this is not the test I asked to write (but it's a good test as well). 

I asked for a test where you use a single `ManagedIdentityNetworkClient` object


```ts
 try {
                    await managedIdentityApplication.acquireToken(
                        managedIdentityRequestParams1 // resource 1
                    );
                } catch (e) {
                    expect(sendGetRequestAsyncSpyApp).toHaveBeenCalledTimes(2);                 
                }

                try {
                    await managedIdentityApplication.acquireToken(
                        managedIdentityRequestParams2 // resource 2
                    );
                } catch (e) {
                    expect(sendGetRequestAsyncSpyApp).toHaveBeenCalledTimes(2);                 
                }
```

bgavrilMS (approved):

lib/msal-node/src/client/ManagedIdentitySources/BaseManagedIdentitySource.ts
Robbie-Microsoft: Spoke with Bogdan offline about this and cleared up confusion on both sides. Current implementation works based on new unit test. No change is needed.

lib/msal-node/src/client/ManagedIdentityApplication.ts
Robbie-Microsoft: https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/6919

lib/msal-node/src/client/ManagedIdentityApplication.ts
Robbie-Microsoft: https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/6919

lib/msal-node/test/client/ManagedIdentitySources/Imds.spec.ts
```diff
@@ -170,6 +176,285 @@ describe("Acquires a token successfully via an IMDS Managed Identity", () => {
...
+                            HttpStatus.BAD_REQUEST as number
+                        )
+                    );
+
+                try {
+                    await managedIdentityApplication.acquireToken(
+                        managedIdentityRequestParams
+                    );
+                } catch (e) {
+                    expect(sendGetRequestAsyncSpyApp).toHaveBeenCalledTimes(2);
```
neha-bhargava: This should expect 1 call not 2 since the 400 is not included in http status codes that need to be retried.

Robbie-Microsoft: good catch! Couple of typos on my end. fixed now.

neha-bhargava: I still see this as 2 calls been expected.

lib/msal-node/test/client/ManagedIdentitySources/Imds.spec.ts
```diff
@@ -170,6 +176,285 @@ describe("Acquires a token successfully via an IMDS Managed Identity", () => {
...
+                            HttpStatus.BAD_REQUEST
+                        )
+                    );
+
+                try {
+                    await managedIdentityApplication.acquireToken(
+                        managedIdentityRequestParams
+                    );
+                } catch (e) {
+                    expect(sendGetRequestAsyncSpyApp).toHaveBeenCalledTimes(1);
```
Robbie-Microsoft, line 450: @neha-bhargava I think you are looking at an outdated version. I updated the code.

lib/msal-node/test/client/ManagedIdentitySources/Imds.spec.ts
Robbie-Microsoft: I think you're looking at an outdated version. In the "files changed" tab, it shows as 1 call.

neha-bhargava (approved):

PR #6896, opened 2024-02-14
Fix MSA e2e tests

hectormmg:
This PR:
- Replaces sign-in button ID with #kmsiTitle ID for KMSI page in MSA flows

### DIFF
--- a/samples/e2eTestUtils/src/TestUtils.ts
+++ b/samples/e2eTestUtils/src/TestUtils.ts
@@ -307,22 +307,22 @@ export async function enterCredentials(
             throw e;
         });
     } catch (e) {
-        return;
-    }
-
-    // keep me signed in MSA
-    try {
-        await page.waitForSelector("#kmsiTitle", { timeout: 1000 });
-        await screenshot.takeScreenshot(page, "keepMeSignedInPage");
-        await Promise.all([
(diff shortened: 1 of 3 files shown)


samples/e2eTestUtils/src/TestUtils.ts
```diff
@@ -307,7 +307,22 @@ export async function enterCredentials(
             throw e;
         });
     } catch (e) {
-        return;
+        // keep me signed in MSA
+        try {
+            await page.waitForSelector("#kmsiTitle", { timeout: 1000 });
```
tnorling: Could be a bit faster and less duplication if you use Promise.any above with both tags

tnorling (approved):
One suggestion but non-blocking

samples/e2eTestUtils/src/TestUtils.ts
hectormmg: Great idea, thanks! I was wondering if there was a way to use whichever one showed up.

lalimasharda (approved):

samples/e2eTestUtils/src/TestUtils.ts
hectormmg: @tnorling ended up using race since .any wasn't available for the current build target settings.

konstantin-msft (approved):

Issue #3758, opened 2021-06-15
Angular interceptor should support APIs with optional authentication


kevincathcart-cas:
### Core Library

MSAL.js v2 (@azure/msal-browser)

### Wrapper Library

MSAL Angular (@azure/msal-angular)

### Description

There should be a way in the angular interceptor to specify an API where authentication is optional
(i.e. it should send the access token if it has a valid one, or if it can silent acquire one, but omit it otherwise). This is useful for APIs that do not strictly require authentication, but provide additional functionality if authenticated. It might return only public data to an anonymous user, but include personalized data for a signed in user, or signed in users may have less restrictive rate limiting, etc.

Furthermore, the need for authentication can vary for the same URL by http method, and right now there is no way to support that without a full.

For the first scenario, this feels like it should be a relatively easy feature to add, since it is simply skipping the `this.acquireTokenInteractively` calls. Even for the second scenario, sending the auth token (when available) when not actually required is usually harmless, so both scenarios could be supported in this same manner.

The only tricky part seems to me to be the ability to specify this option. I would simply update the `ProtectedResourceScopes` type to make httpMethod optional (with semantics that not supplied means any method), and add a new boolean to trigger the new behavior. The reason I would put it on the scopes level rather than the resource level is because that is where the httpMethod is specified, and this should allow (but not require) distinguishing by method.

Thoughts?

### Source

External (Customer)

jasonnutter:
> Furthermore, the need for authentication can vary for the same URL by http method, and right now there is no way to support that without a full.

Note, this is currently possible in v2. See more here: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-angular/docs/v2-docs/msal-interceptor.md#protected-resource-map

samuelkubai:
@kevincathcart-cas thank you for filling this issue, we'll mark it as an enhancement and put it in the backlog. We'll update this issue with any updates.

kevincathcart-cas:
> > Furthermore, the need for authentication can vary for the same URL by http method, and right now there is no way to support that without a full.
> 
> Note, this is currently possible in v2. 

Right, I misread the source, and if specify only `ProtectedResourceScopes`, but don't include a specific httpMethod in any, it will fall back to not using auth. Good.

Obviously the point about optional authentication for a route still stands.

hansmaad:
[...]

PR #6928, opened 2024-02-28
optional route interception

bytelabsco:
In scenarios where a route may optionally require a token if the user is logged in, you can now mark ProtectedResourceScopes definitions as optional.  These requests will attach (after acquiring, if necessary) the required token if a user is authenticated.  If no user is currently authenticated, the request will simply go through as if it is an unprotected endpoint, and not force the user to authenticate.

#3758


bytelabsco:

PR #6930, opened 2024-02-29
Add handleRedirectPromise docs to login-user and acquire-token docs

hectormmg:
This PR:
- Adds `handleRedirectPromise` section to docs that explain redirect API usage

### DIFF
--- a/lib/msal-browser/docs/acquire-token.md
+++ b/lib/msal-browser/docs/acquire-token.md
@@ -32,16 +32,19 @@ var request = {
     scopes: ["User.Read"],
 };
 
-msalInstance.acquireTokenSilent(request).then(tokenResponse => {
-    // Do something with the tokenResponse
-}).catch(async (error) => {
-    if (error instanceof InteractionRequiredAuthError) {
-        // fallback to interaction when silent call fails
-        return msalInstance.acquireTokenPopup(request);
-    }
+msalInstance
+    .acquireTokenSilent(request)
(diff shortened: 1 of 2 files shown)


lib/msal-browser/docs/acquire-token.md
```diff
@@ -51,18 +54,44 @@ var request = {
...
+            return msalInstance.acquireTokenRedirect(request);
+        }
+
+        // handle other errors
+    });
+```
+
+Note: If you are using `msal-angular` or `msal-react`, redirects are handled differently, and you should see the [`msal-angular` redirect doc](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-angular/docs/v2-docs/redirects.md) and [`msal-react` FAQ](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-react/FAQ.md#how-do-i-handle-the-redirect-flow-in-a-react-app) for more details.
+
+The redirect APIs are asynchronous (i.e. return a promise) `void` functions which redirect the browser window after caching some basic info. If you choose to use the redirect APIs, be aware that **you MUST call `handleRedirectPromise()` to correctly handle the API**. You can use the following function to perform an action when this token exchange is completed:
```
tnorling, line 74: Let's make this clear that you need to call this automatically on every page load. It might even help to combine this code snippet with the one above to illustrate that this comes first.

PR #6933, opened 2024-03-03
Refactor bootstrap of angualr standalone sample

andrewalderson:
The documentation and example for bootstrapping redirects for Angular standalone application is not correct. It implies that there is no way in an Angular standalone application to bootstrap the `MsalRedirectComponent` as you can in an application with an `app.module.ts`. This is actually incorrect. The `MsalRedirectComponent` can be bootstrapped in the `main.ts` file by using the `Promise<ApplicationRef>` returned by the `bootstrapApplication` function.

The PR corrects that by refactoring the Angular standalone sample and updating the documentation.

### DIFF
--- a/lib/msal-angular/docs/redirects.md
+++ b/lib/msal-angular/docs/redirects.md
@@ -183,6 +183,119 @@ export class AppComponent implements OnInit, OnDestroy {
   }
 ```
 
+## Redirects with standalone components
+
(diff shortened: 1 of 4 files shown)


andrewalderson:

andrewalderson:
Just noticed that I committed my changes to the dev branch. I am not sure what happened to the branch I created. I will close this and resubmit it.

PR #6937, opened 2024-03-05
Instrument non-auth error name and stack

konstantin-msft:
- Instrument non-auth error name and stack in Controllers.
- Add/update telemetry functions to capture non-auth error name and stack.

**Note** Non-auth errors make up 0.06% of the total errors so stack trace parsing/compacting won't impact perf. This number should get smaller once we sample and address non-auth errors.


lib/msal-common/src/telemetry/performance/IPerformanceClient.ts
```diff
@@ -31,6 +32,7 @@ export interface IPerformanceClient {
         fields: { [key: string]: {} | undefined },
         correlationId: string
     ): void;
+    addError(error: Error, correlationId: string, maxStackSize: number): void;
```
tnorling: Adding functions to this interface is potentially a breaking change - is there a way to add this functionality using the existing methods?

konstantin-msft: Done. Looks a bit ugly but does the job with regards to perf interface immutability.

lib/msal-common/src/utils/FunctionWrappers.ts
```diff
@@ -53,9 +57,8 @@ export const invoke = <T extends Array<any>, U>(
             } catch (e) {
                 logger.trace("Unable to print error message.");
             }
-            inProgressEvent?.end({
-                success: false,
-            });
+
+            captureError(e, inProgressEvent);
```
tnorling: This would capture every error that occurs in the stack, even ones we end up swallowing, potentially leading to error fields being defined in successful calls. It may be better to handle this at the top of the stack only, for example: [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dfda8fa0751a1dc924d081ab359a25860b22483f/lib/msal-browser/src/controllers/StandardController.ts#L743)

lib/msal-common/src/telemetry/performance/IPerformanceClient.ts
tnorling: Thanks this looks good

lib/msal-common/src/utils/FunctionWrappers.ts
konstantin-msft: That's a very good point. Thanks. I have made the change such that we capture errors in controllers and simplified the instrumentation code.

tnorling (approved):

jo-arroyo (approved):

PR #6941, opened 2024-03-08
Nested App Auth: allow empty parameters for loginPopup

codexeon:
This change allows empty parameters for loginPopup by using a default request to match the behavior of StandardController. It also removes the dependency on crypto.randomUUID() which is not used elsewhere in the MSAL code base. Use the shared implementation to generate UUID v7 instead.

### DIFF
--- a/lib/msal-browser/src/controllers/NestedAppAuthController.ts
+++ b/lib/msal-browser/src/controllers/NestedAppAuthController.ts
@@ -26,7 +26,12 @@ import { PopupRequest } from "../request/PopupRequest";
 import { RedirectRequest } from "../request/RedirectRequest";
 import { SilentRequest } from "../request/SilentRequest";
 import { SsoSilentRequest } from "../request/SsoSilentRequest";
-import { ApiId, WrapperSKU, InteractionType, DEFAULT_REQUEST } from "../utils/BrowserConstants";
+import {
+    ApiId,
+    WrapperSKU,
+    InteractionType,
+    DEFAULT_REQUEST,
+} from "../utils/BrowserConstants";
 import { IController } from "./IController";
 import { TeamsAppOperatingContext } from "../operatingcontext/TeamsAppOperatingContext";
 import { IBridgeProxy } from "../naa/IBridgeProxy";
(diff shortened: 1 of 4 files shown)


lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -416,11 +421,7 @@ export class NestedAppAuthController implements IController {
     loginPopup(
         request?: PopupRequest | undefined // eslint-disable-line @typescript-eslint/no-unused-vars
     ): Promise<AuthenticationResult> {
-        if (request !== undefined) {
-            return this.acquireTokenInteractive(request);
-        } else {
-            throw NestedAppAuthError.createUnsupportedError();
-        }
+        return this.acquireTokenInteractive(request || DEFAULT_REQUEST);
```
sameerag, line 426: Do you track correlation-id for NAA? I think this should be:
```javascript
        const correlationId: string = request.correlationId || BrowserCrypto.createNewGuid(),
        return this.acquireTokenPopup({
            correlationId,
            ...(request || DEFAULT_REQUEST),
        });
 ```
 
Better yet, may be utilize `getRequestCorrelationId()` from `StandardController` as a common API.

sameerag (commented):
Minor comments on correlationId.

lib/msal-browser/src/controllers/NestedAppAuthController.ts
codexeon: We do use correlation id for NAA, but it is generated in a different location than the StandardController if not provided by the caller. [Here](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/cba97b5f7a87dc8e55f2cad384ff471931b67f2a/lib/msal-browser/src/controllers/NestedAppAuthController.ts#L129C2-L132C11) it is called from all the token API supported with Nested App Auth: ssoSilent, acquireTokenSilent, acquireTokenPopup, and loginPopup.

sameerag: I see. Yeah, checked that you are validating in `acquireTokenInteractive`

sameerag (approved):

baljesingh (approved):
LGTM

PR #6960, opened 2024-03-19
Add classname to silent token renewal iframe HTML Element

lalimasharda:


lib/msal-browser/src/interaction_handler/SilentHandler.ts
```diff
@@ -145,9 +139,11 @@ export class SilentHandler extends InteractionHandler {
      * Creates a new hidden iframe or gets an existing one for silent token renewal.
      * @ignore
      */
-    private createHiddenIframe(): HTMLIFrameElement {
+    private createHiddenIframe(urlNavigate: string): HTMLIFrameElement {
         const authFrame = document.createElement("iframe");
 
+        authFrame.className = "silentTokenRenewalIframe";
+        authFrame.src = urlNavigate;
```
sameerag: Is there a reason why we did not do this before? @tnorling  any browser policy issues we need to worry about?

lib/msal-browser/src/interaction_handler/SilentHandler.ts
```diff
@@ -110,16 +110,13 @@ export class SilentHandler extends InteractionHandler {
...
-            const frameHandle = this.createHiddenIframe();
+            const frameHandle = this.createHiddenIframe(urlNavigate);
 
             setTimeout(() => {
                 if (!frameHandle) {
                     reject("Unable to load iframe");
                     return;
                 }
-
-                frameHandle.src = urlNavigate;
```
tnorling: This is a breaking change and needs to wait for v4 - can you file a work item if one doesn't exist?

lib/msal-browser/src/interaction_handler/SilentHandler.ts
tnorling: There was a limitation in some older browsers, namely IE11, that affected iframe creation. To workaround we allowed applications to configure whether the src attribute should be assigned synchronously or asynchronously - we can likely remove that configuration in v4 but we'll need to keep the existing model in v3 to avoid a change in behavior here

lib/msal-browser/src/interaction_handler/SilentHandler.ts
```diff
@@ -145,9 +139,11 @@ export class SilentHandler extends InteractionHandler {
      * Creates a new hidden iframe or gets an existing one for silent token renewal.
      * @ignore
      */
-    private createHiddenIframe(): HTMLIFrameElement {
+    private createHiddenIframe(urlNavigate: string): HTMLIFrameElement {
         const authFrame = document.createElement("iframe");
 
+        authFrame.className = "silentTokenRenewalIframe";
```
tnorling: Curious why this is needed?

lib/msal-browser/src/interaction_handler/SilentHandler.ts
lalimasharda: Noted, thanks!

lib/msal-browser/src/interaction_handler/SilentHandler.ts
sameerag: Bing is blocking Dime `ssoSilent` on 2.x (pre-broker). They want a classname to get an exception.

lib/msal-browser/src/interaction_handler/SilentHandler.ts
sameerag: That is what I was wondering. @lalimasharda  let us stick to `className` for now to unblock Dime.

lib/msal-browser/src/interaction_handler/SilentHandler.ts
tnorling: I see...interesting. Should we include "msal" in the name somewhere?

sameerag: Not a bad idea, @lalimasharda  can we do that, say `msalSilentIframe`

tnorling (approved):

sameerag (approved):

PR #6962, opened 2024-03-20
Fix for concurrent iframe calls

tnorling:
When multiple acquireTokenSilent requests are made in parallel there is a chance that they will all need to fallback to the iframe flow. It is both unnecessary for more than 1 iframe call to be made and results in a perf and reliability degradation for all calls. This PR introduces a mechanism to track in progress iframe calls and cause subsequent requests to wait before retrying the cache and/or RT redemption.

Note: Telemetry dashboards will need to be updated after this change is released to avoid counting awaited iframe calls against our perf metrics more than once.

Note: This PR does not make any changes to ssoSilent - follow up work should add ssoSilent calls to the active request tracking variable and log warnings when more than 1 ssoSilent requests are made but should **not** block the calls.


lib/msal-browser/src/controllers/StandardController.ts
```diff
@@ -1966,24 +1969,180 @@ export class StandardController implements IController {
...
+                    const awaitConcurrentIframeMeasure =
+                        this.performanceClient.startMeasurement(
+                            PerformanceEvents.AwaitConcurrentIframe,
+                            silentRequest.correlationId
+                        );
+                    awaitConcurrentIframeMeasure.add({
+                        awaitIframeCorrelationId: activeCorrelationId,
+                    });
+
+                    // Await for errors first so we can distinguish errors thrown by activePromise versus errors thrown by .then below
```
sameerag, line 2043: Comment incomplete?

tnorling: I dont think so - do I need to rephrase?

sameerag (approved):
Please leave docs for `ssoSilent` use case, for users who are using this API instead of `acquireTokenSilent` - It is anti pattern but I agree that blocking them is more harmful for prod changes.

lib/msal-browser/src/controllers/StandardController.ts
```diff
@@ -1966,24 +1969,180 @@ export class StandardController implements IController {
...
+                        `Iframe request is already in progress, awaiting resolution for request with correlationId: ${activeCorrelationId}`,
+                        silentRequest.correlationId
+                    );
+                    const awaitConcurrentIframeMeasure =
+                        this.performanceClient.startMeasurement(
+                            PerformanceEvents.AwaitConcurrentIframe,
+                            silentRequest.correlationId
+                        );
+                    awaitConcurrentIframeMeasure.add({
+                        awaitIframeCorrelationId: activeCorrelationId,
```
konstantin-msft, line 2040: Please add `awaitIframeCorrelationId` optional attribute to `PerformanceEvent` type.

konstantin-msft (approved):

PR #6987, opened 2024-03-27
MSAL Node - HttpClient Proxy Improvements

Robbie-Microsoft:
Added `customAgent` to msal-node system options so custom http(s) agents can be used by the HttpClient. This allows devs to use agents created by third party libraries (useful for proxy support).

Added support for authenticated proxy.

NOTE: faq.md was automatically linted. The only change is on line 75.


lib/msal-node/src/config/Configuration.ts
```diff
@@ -145,6 +146,7 @@ const DEFAULT_SYSTEM_OPTIONS: Required<NodeSystemOptions> = {
     loggerOptions: DEFAULT_LOGGER_OPTIONS,
     networkClient: new HttpClient(),
     proxyUrl: Constants.EMPTY_STRING,
+    customAgent: new http.Agent() || new https.Agent(),
```
bgavrilMS: What does it mean? Doesn't this just resolve to `new http.Agent()` ?

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -72,6 +78,7 @@ export class HttpClient implements INetworkModule {
                 this.proxyUrl,
                 HttpMethod.POST,
                 options,
+                this.customAgent as http.Agent,
```
bgavrilMS: What if the customer configured https.Agent? Should you not throw an exception with a good error message explaining how to configure these?

lib/msal-node/src/network/HttpClient.ts
```diff
@@ -80,6 +87,7 @@ export class HttpClient implements INetworkModule {
                 url,
                 HttpMethod.POST,
                 options,
+                this.customAgent as https.Agent,
```
bgavrilMS: How will this work with MSI? Calls to MSI are done over http.

I don't expect anyone to define a proxy for calling MSI. That can be unsupported scenario.

bgavrilMS (changes requested):
I understand this is not unit testable, but please explain how you manually tested this?

lib/msal-node/src/network/HttpClient.ts
Robbie-Microsoft: I made some improvements to address this.

Take a look at this [stack overflow](https://stackoverflow.com/questions/64117101/what-is-the-role-of-https-agent-in-node) post. 

> https.Agent internally calls and constructs with http.Agent

I interpret this as meaning either one can be used. Look at my improvements and let me know your thoughts.

lib/msal-node/src/network/HttpClient.ts
Robbie-Microsoft: My improvements should address this.

lib/msal-node/src/config/Configuration.ts
Robbie-Microsoft: My improvements should address this.

Robbie-Microsoft:
> I understand this is not unit testable, but please explain how you manually tested this?

I'll be testing this shortly on my Windows laptop with Fiddler, to simulate a proxy. I'm also going to to ask the user who filed the linked issue to test this.

PR #6997, opened 2024-04-02
Upgrade angular-standalone-sample to use Angular 17

jo-arroyo:


jo-arroyo:
Closed in favor of #7017

PR #6998, opened 2024-04-02
Fix handleRedirectPromise Memoization

### LINKED ISSUE #6893: Race condition when calling `handleRedirectPromise` in Safari using @azure/msal-react
### Core Library

MSAL.js (@azure/msal-browser)

### Core Library Version

3.9.0

### Wrapper Library

### Wrapper Library Version

2.0.11

Public

### Description

I've been troubleshooting an authentication issue that is happening intermittently in Safari for multiple of our customers.

After troubleshooting the problem for some time, I came to the conclusion that I'm having a race condition between `MsalProvider`'s call to `handleRedirectPromise` and our implementation's call to `handleRedirectPromise` when trying to acquire the `state` of the `loginRedirect` call.

What I've found is that, while `MsalProvider` is running `handleRedirectPromise`, execution is then removed and my implementation starts executing. This leads to problems because the first call is not able to cache the Promise, but, has enough time to remove the fragment state, preventing the second call from working as expected.

A few things that I believe are worth mentionting.

1) The problem is random. I can reproduce it ~50% of the times
2) The problem only seems to happen in Safari. 

We tried moving to the Events API but I started seeing other inconsistencies in Safari. The workaround so far is to add a `setTimeout` of 2 seconds in our callback component to guarantee that `MsalProvider` it's done with the initialization process. The other approach I thought could solve the problem is watching the`inProgress` status of the `MsalProvider` and only start my application when `status` is `none` but this doesn't work for me because `MsalProvider` will change it's state when trying to acquire a token in normal application flow.

I'm not 100% sure if this is a bug or something we need to protect ourselves against. Maybe there is something that can be done in `@azure/msal-react` to make life easier to the user. 

I think this issue might be related to my problem as well https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/6797

### Error Message

```json
{"errorCode":"state_not_found","errorMessage":"State not found: Cached State","subError":"","name":"ClientAuthError","correlationId":"018da7fb-acbe-7e9b-b354-f5fb0f78be59"}
```

### MSAL Logs

Sending logs in a different comment due to size.

### Network Trace (Preferrably Fiddler)

- [ ] Sent
- [X] Pending

### MSAL Configuration

```javascript
{
	auth: {
		authority: "XXXX",
		clientId: "XXXX",
		knownAuthorities: ['XXXX']
	},
	cache: {
		cacheLocation: 'localStorage'
	}
}
```

### Relevant Code Snippets

```javascript
The MSAL provider is defined in the very first component of the application.


render((
	<Provider store={store}>
		<PersistGate persistor={persistor}>
				<Root history={history} />
			</MsalProvider>
		</PersistGate>
	</Provider>
), document.getElementById('app'));


We are doing `loginRedirect` calls to start the flows. Calls look like this:

```js
msalInstance.loginRedirect({
				loginHint   : email,
				authority   :'XXXX',
				scopes      : 'XXXXX',
				redirectUri : '/ad/callback',
				prompt      : 'login',
				state       : utils.encodeADState({ operation : 'login' }),
			});
```

And then we have an ADCallback component attached to the `/ad/callback` route that looks like this:

```
const ADCallback = ({ history }) => {
	const { instance } = useMsal();

	const callback = () => {
		instance.handleRedirectPromise()
			.then((accessTokenResponse) => {
				if (accessTokenResponse) {
					const state = utils.decodeADState(accessTokenResponse.state);

					switch (state.operation) {
						// Handle case
					}
				}
			})
			.catch((error) => {
				// Handler error
			});
	};

	useEffect(() => {
		callback();
	}, []);

	return (
		<div id="ad-signup">
			<Loading />
		</div>
	);
};

const mapDispatchToProps = (dispatch) => ({
	loginWithAD : bindActionCreators(thunks, dispatch).loginWithAD,
});

ADCallback.propTypes = {
	history : PropTypes.object.isRequired,
};

export default connect(null, mapDispatchToProps)(ADCallback);
```
```


### Reproduction Steps

Calling `handleRedirectPromise` in Safari using the `MsalProvider`.

### Expected Behavior

The implementation of `handleRedirectPromise` is smart enough to realize that there is an inflight operation happening and will wait for it to complete before executing the second one.

### Identity Provider

Entra ID (formerly Azure AD) / MSA

Safari

### Regression

_No response_

### Source

External (Customer)

### PR DESCRIPTION
**Pull request by tnorling**
We're doing too many things after checking for an existing promise but before setting the current promise which can create a race condition if multiple instances of handleRedirectPromise are invoked.

To address this PR sets the promise immediately after checking for one.

Fixes #6893

### CODE REVIEW

**Review by sameerag — APPROVED** [review]

File: change/@azure-msal-browser-b7c89de9-6914-47ba-9e31-0cd60c0fb29f.json [hunk 1]
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "fix handleRedirectPromise memoization",
```
jo-arroyo:
```suggestion
  "comment": "fix handleRedirectPromise memoization #6998",
```

lib/msal-browser/src/controllers/StandardController.ts
```diff
@@ -345,155 +344,11 @@ export class StandardController implements IController {
...
-                const rootMeasurement = this.performanceClient.startMeasurement(
-                    "acquireTokenRedirect",
-                    correlationId
-                );
-                this.eventHandler.emitEvent(
-                    EventType.HANDLE_REDIRECT_START,
-                    InteractionType.Redirect
-                );
+                response = this.handleRedirectPromiseInternal(hash);
+                this.redirectResponse.set(redirectResponseKey, response);
```
jo-arroyo, line 348: Could you clarify for me: Since this is not awaiting the response from `handleRedirectPromiseInternal` but is setting it first, does it need to set the response again later on?

tnorling: No the `response` is a promise object which can resolve at any time.

In a nutshell: 

First call: create promise -> save promise -> return promise
Subsequent calls: get promise -> return promise

jo-arroyo: Okay, great. Thanks for the explanation!

jo-arroyo (approved):

PR #6999, opened 2024-04-03
ClientCredential and OBO acquireToken requests with claims will now skip the cache

### LINKED ISSUE #6173: Claims don't bypass the cache for OBO and client_credentials
### Core Library

MSAL Node (@azure/msal-node)

### Core Library Version

1.17.3

### Wrapper Library

Not Applicable

### Wrapper Library Version

n/a

Confidential

### Description

1. Acquire a token from the STS using client_creds or OBO
2. Make the same request again. Expect the result to come from the cache.
3. Make a similar request, where you add some claims e.g. `claims="foo"`

Expected: request should go to the STS
Actual: token is served from the cache

Impact: this breaks the CAE scenario for web api and the upcoming CAE support for S2S

### Error Message

see description

### Msal Logs

see description

### MSAL Configuration

```javascript
see description
```

### Relevant Code Snippets

```javascript
see description
```

### Reproduction Steps

see description

### Expected Behavior

see description

### Identity Provider

Azure AD / MSA

None (Server)

### Regression

_No response_

### Source

Internal (Microsoft)

Robbie-Microsoft:
ClientCredential and OBO acquireToken requests with claims will now skip the cache.

claimsBasedCachingEnabled has been marked as deprecated in msal-node.


lib/msal-node/test/client/ClientCredentialClient.spec.ts
```diff
@@ -446,6 +446,27 @@ describe("ClientCredentialClient unit tests", () => {
                     CONFIDENTIAL_CLIENT_AUTHENTICATION_RESULT.body.access_token
                 );
                 expect(cachedAuthResult.fromCache).toBe(true);
+
+                // validate that the same request but with new claims will make a STS request
```
bgavrilMS: But will a request **without** claims now result in a cache hit or a cache miss ?

In orther words, the scenario: 

1. Get a token without claims. Expected source: IDP
2. Get a token with claims. Expected source: IDP
3. Get a token with other claims. Expected source: IDP (you tested this I think)
4. Get a token without claims. Expected source: cache

What I really hope to avoid is a situation where step 4 comes from IDP. Because in this case, it's up to the user to ALWAYS remember the claims and to inject the claims.

bgavrilMS: Have a look at https://learn.microsoft.com/en-us/entra/identity-platform/app-resilience-continuous-access-evaluation?tabs=JavaScript for current guidance on CAE

Robbie-Microsoft: I just tested it. What happens in this case is that token #4 is the same token as token #1; #4 is retrieved from the cache. They both have a non-existant claims hash, so they match each other.

This comparison logic already existed, but just wasn't being used properly.

However, as I understand the CAE docs you linked, it appears that an application would never receive the same claims challenge twice because after the first request, a new token would be retrieved from the IDP, and the token would pass CAE. Is that right? In that case, we can do what you originally suggested and just skip the cache if there are claims in the token request.

bgavrilMS: How are the access tokens cached? Is the claims_hash made part of the token cache key?

Let's have a meeting with the other MSAL JS folks to discuss about this before proceeding.

sameerag:
```
How are the access tokens cached? Is the claims_hash made part of the token cache key?
```

Please avoid this. Just cleaning up a mess with Teams as we added this for them and it is causing cache saturation issues. By default, MSAL JS does not cache the `claims` in token key.

lib/msal-node/src/error/ConfigurationError.ts
```diff
@@ -0,0 +1,30 @@
...
+import { AuthError } from "@azure/msal-common";
+import * as ConfigurationErrorCodes from "./ConfigurationErrorCodes";
+export { ConfigurationErrorCodes };
+
+/**
+ * ConfigurationErrorMessages class containing string constants used by error codes and messages.
+ */
+export const ConfigurationErrorMessages = {
+    [ConfigurationErrorCodes.claimsBasedCachingEnabled]:
+        "Claims based caching is no longer supported.",
```
bgavrilMS:
```suggestion
        "Claims based caching is not supported on MSAL NodeJS.",
```

bgavrilMS: (it never was supported)

lib/msal-node/test/client/ClientCredentialClient.spec.ts
```diff
@@ -437,19 +443,55 @@ describe("ClientCredentialClient unit tests", () => {
...
+
+                // acquire a token with a client that has client capabilities, and has claims in the request
+                clientCredentialRequest.claims = claims;
+                const authResult2 = (await client.acquireToken(
+                    clientCredentialRequest
+                )) as AuthenticationResult;
+                expect(authResult2.accessToken).toEqual(
+                    CONFIDENTIAL_CLIENT_AUTHENTICATION_RESULT.body.access_token
+                );
+                expect(authResult2.fromCache).toBe(false);
```
bgavrilMS, line 464: Can you please add 2 more requests to this test: 

- another request with no claims - expect the result to come from cache
- another request with claims - expect the result to come from IDP

in both cases client capability continues to be set.

bgavrilMS (approved):

lib/msal-node/test/client/ClientCredentialClient.spec.ts
Robbie-Microsoft: I don't understand, I have these tests already. You want me to repeat them?

bgavrilMS (approved):

lib/msal-common/test/config/ClientConfiguration.spec.ts
```diff
@@ -31,21 +36,21 @@ describe("ClientConfiguration.ts Class Unit Tests", () => {
         expect(emptyConfig.cryptoInterface.base64Decode).not.toBeNull();
         expect(() =>
             emptyConfig.cryptoInterface.base64Decode("test input")
-        ).toThrowError(
```
Robbie-Microsoft: deprecated

neha-bhargava (approved):

sameerag (approved):

sameerag (changes requested):
Please mark this as a warning than throw - to not make this breaking for 3.x

sameerag (approved):

PR #7001, opened 2024-04-03
MSAL-Node NetworkModule: Fixed inconsistencies with cancellationToken (timeout)

Robbie-Microsoft:
- renamed cancellationToken to timeout
- it can only be used in get requests (region discovery, for example)
- it's not part of the public api


bgavrilMS (approved):

lib/msal-common/src/utils/Constants.ts
```diff
@@ -64,22 +64,26 @@ export const Constants = {
 };
 
 export const HttpStatus = {
+    SUCCESS: 200,
```
neha-bhargava, line 67: This already exists below at line 69

lib/msal-common/src/utils/Constants.ts
```diff
@@ -64,22 +64,26 @@ export const Constants = {
...
     REDIRECT: 302,
+    CLIENT_ERROR: 400,
     CLIENT_ERROR_RANGE_START: 400,
     BAD_REQUEST: 400,
     UNAUTHORIZED: 401,
     NOT_FOUND: 404,
     REQUEST_TIMEOUT: 408,
     TOO_MANY_REQUESTS: 429,
     CLIENT_ERROR_RANGE_END: 499,
+    SERVER_ERROR: 500,
```
neha-bhargava, line 79: Duplicate of INTERNAL_SERVER_ERROR at line 82

neha-bhargava (changes requested):
Since the managed identity calls use GET request, please evaluate how this timeout affects MSI calls.

lib/msal-common/src/network/INetworkModule.ts
```diff
@@ -31,7 +31,7 @@ export interface INetworkModule {
     sendGetRequestAsync<T>(
         url: string,
         options?: NetworkRequestOptions,
-        cancellationToken?: number
+        timeout?: number
```
bgavrilMS, line 34: Can you please add a comment on the public API that is deprecated? 

https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/23df98f6747f47f9ba1bb458feba9f183f4a902c/lib/msal-common/src/error/AuthError.ts#L18

Robbie-Microsoft: This isn't necessary. This is only changing the parameter name, and cancellationToken/timeout is only used internally.

neha-bhargava (approved):

PR #7008, opened 2024-04-05
Skip "login_hint" opaque claim if domainHint param is set

### LINKED ISSUE #6993: AADSTS1002014: Unable to complete request. The request was invalid since domain_hint and opaque login_hint cannot be used together.
### Core Library

MSAL.js (@azure/msal-browser)

### Core Library Version

2.38.3

### Wrapper Library

### Wrapper Library Version

None

Public

### Description

When use MSAL React in SAP project.  Business need use Domain_Hint to set the default domain at first. and then,  The Entra ID registered App also enabled the ID Claim ( login_hint ) when issue ID_Token. 

once the local cache expired.  will call acquireTokenSilent, acquireTokenRedirect or LoginRedirect methods to refresh token or relogin.   

At this conditions, will get "AADSTS1002014: Unable to complete request. The request was invalid since domain_hint and opaque login_hint cannot be used together." error. 

image

Troubleshooting details
If you contact your administrator, send this info to them.
Copy info to clipboard
Request Id: 0283f61b-fdf0-470e-98be-507221a55800
Correlation Id: 018e55c0-e998-764a-b50c-bdff2f23814b
Timestamp: 2024-03-19T08:08:52Z
Message: AADSTS1002014: Unable to complete request. The request was invalid since domain_hint and opaque login_hint cannot be used together.
Flag sign-in errors for review: Enable flagging
If you plan on getting help for this problem, enable flagging and try to reproduce the error within 20 minutes. Flagged events make diagnostics available and are raised to admin attention.

### Error Message

AADSTS1002014: Unable to complete request. The request was invalid since domain_hint and opaque login_hint cannot be used together.

### MSAL Logs

_No response_

### Network Trace (Preferrably Fiddler)

- [ ] Sent
- [ ] Pending

### MSAL Configuration

```javascript
import { LogLevel } from "@azure/msal-browser";
// Browser check variables
// If you support IE, our recommendation is that you sign-in using Redirect APIs
// If you as a developer are testing using Edge InPrivate mode, please add "isEdge" to the if check
const ua = window.navigator.userAgent;
const msie = ua.indexOf("MSIE ");
const msie11 = ua.indexOf("Trident/");
const msedge = ua.indexOf("Edge/");
const firefox = ua.indexOf("Firefox");
const isIE = msie > 0 || msie11 > 0;
const isEdge = msedge > 0;
const isFirefox = firefox > 0; // Only needed if you need to support the redirect flow in Firefox incognito

export const msalConfig = {
    auth: {
        clientId: process.env.REACT_APP_CLIENT_ID,
        authority: process.env.REACT_APP_AUTHORITY,
        redirectUri: "/mcd-boss-h5/otpLogin",
        postLogoutRedirectUri: "/portal",
        navigateToLoginRequestUrl: false
    },
    cache: {
        storeAuthStateInCookie: isIE || isEdge || isFirefox,
    },
    system: {
        allowNativeBroker: false, // Disables WAM Broker
        loggerOptions: {
                if (containsPii) {
                    return;
                }
                switch (level) {
                    case LogLevel.Error:
                        console.error(message);
                        return;
                    case LogLevel.Info:
                        console.info(message);
                        return;
                    case LogLevel.Verbose:
                        console.debug(message);
                        return;
                    case LogLevel.Warning:
                        console.warn(message);
                        return;
                    default:
                        return;
                }
            },
        },
    },
};

export const loginRequest = {
    scopes: ["User.Read"],
    domainHint: "xxxxxx.cn"
};
```

### Relevant Code Snippets

```javascript
if (!accessToken) {
        const account = msalInstance.getActiveAccount();
        if (!account) {
            throw Error("No active account! Verify a user has been signed in and setActiveAccount has been called.");
        }
        console.log("the output loginrequest : " + loginRequest);
        console.log(loginRequest)
        console.log("the output account: " + account);
        console.log(account);

        const response = await msalInstance.acquireTokenSilent({
            ...loginRequest,
            //scopes: loginRequest.scopes,
            account: account
        }).catch((error) => {
            console.log('getCasToken error', error);
       
            console.log(error);
            //delete account["idTokenClaims"]["login_hint"];
            const responsenew = msalInstance.acquireTokenRedirect({
                ...loginRequest,
                //scopes: loginRequest.scopes,
                account: account
            }).catch(error => {
                console.log('acquireTokenRedirect error', error);
            })

            console.log(responsenew)
        });


// For Login Methods.

    const { instance } = useMsal();

    const onReLogin = () => {

        const account = instance.getActiveAccount();
        //delete account["idTokenClaims"]["login_hint"];

        //instance.loginRedirect({ ...loginRequest, account: account });
        instance.loginRedirect(loginRequest);
    };
```

Update some Screenshot:

image

and the Account Login_hint :
image

### Reproduction Steps

1. Call loginRedirect or acquireTokenRedirect or acquireTokenSilent if you input domain_hint & login_hint together.

### Expected Behavior

As those methods allow input domain_hint , login_hint parameters together. It's cannot throw the confilct error.   

We expected the no error and login successfully. 

### Identity Provider

Entra ID (formerly Azure AD) / MSA

Chrome, Firefox, Edge, Safari

### Regression

_No response_

### Source

Internal (Microsoft)

konstantin-msft:
- Skip "login_hint" opaque claim if domainHint param is set.

This PR addresses AADSTS1002014 `Unable to complete request. The request was invalid since domain_hint and opaque login_hint cannot be used together`.


lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -557,9 +557,11 @@ export class AuthorizationCodeClient extends BaseClient {
                 parameterBuilder.addSid(request.sid);
             } else if (request.account) {
                 const accountSid = this.extractAccountSid(request.account);
-                const accountLoginHintClaim = this.extractLoginHint(
-                    request.account
-                );
+                // Skip opaque login_hint claim if domain hint is present
+                const accountLoginHintClaim = request.domainHint
```
tnorling: Can we log a warning when we do this? Opaque loginHint should be most preferred so we should indicate to developers they should stop passing domainHint

konstantin-msft: Good point, thanks! Added a warning message.

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -557,9 +557,13 @@ export class AuthorizationCodeClient extends BaseClient {
                 parameterBuilder.addSid(request.sid);
             } else if (request.account) {
                 const accountSid = this.extractAccountSid(request.account);
-                const accountLoginHintClaim = this.extractLoginHint(
-                    request.account
-                );
+                let accountLoginHintClaim = this.extractLoginHint(request.account);
+
+                if (accountLoginHintClaim && request.domainHint) {
+                    this.logger.warning(`AuthorizationCodeClient.createAuthCodeUrlQueryString: "domainHint" param is set, skipping opaque "login_hint" claim`);
```
```suggestion
                    this.logger.warning(`AuthorizationCodeClient.createAuthCodeUrlQueryString: "domainHint" param is set, skipping opaque "login_hint" claim. Please consider not passing domainHint`);
```

tnorling (approved):

lib/msal-common/src/client/AuthorizationCodeClient.ts
```diff
@@ -560,7 +560,7 @@ export class AuthorizationCodeClient extends BaseClient {
                 let accountLoginHintClaim = this.extractLoginHint(request.account);
 
                 if (accountLoginHintClaim && request.domainHint) {
-                    this.logger.warning(`AuthorizationCodeClient.createAuthCodeUrlQueryString: "domainHint" param is set, skipping opaque "login_hint" claim`);
+                    this.logger.warning(`AuthorizationCodeClient.createAuthCodeUrlQueryString: "domainHint" param is set, skipping opaque "login_hint" claim. Please consider not passing domainHint`);
```
sameerag: Shouldn't this be `opaqueLoginHint` is preferred over `domain_hint`? 
Also I am a little concerned that a set of users who deem to *need* domain_hint will now keep seeing more interaction?

konstantin-msft: Given that ESTS does not allow both "domain_hint" and opaque "login_hint" to be present I can think of 2 options here:
1. Skip opaque "login_hint" and set it to "account.username".
2. Remove "domain_hint" if opaque "login_hint" claim is present.

I would prefer the first one as per recommendation from ESTS.

tnorling: Ideally an opaque login_hint should be priority #1 and domain_hint is priority #4 (after sid and upn) but this gets a little weird in scenarios where the developer explicitly provides domain_hint but one of the other hints is provided implicitly via account (provided or via active account). Another less opinionated approach might be to mirror eSTS behavior and simply throw forcing the developer to be clearer about their intention. 

I'm on the fence about what I like better so I'm ok with whatever you decide.

sameerag: Isn't legacy login_hint `account.username` and the opaque login_hint, a hash of certain properties? In that case, isn't removing `domain_hint` the better option?

Can we get this clarity for STS?

sameerag (approved):

PR #7014, opened 2024-04-09
Client Assertion implementation now accepts an async callback as well as a string argument

### LINKED ISSUE #7013: [msal-node] Confidential Client - Client Assertion implementation should accept callback instead of string argument
### Core Library

MSAL Node (@azure/msal-node)

### Wrapper Library

Not Applicable

Confidential

### Description

The client assertion implementation for confidential client should take a user callback that returns a string instead of accepting a string assertion. This will extend the usability of the API to allow users for refreshing the assertion token etc. .NET and Java implementation of MSAL already support this.

### Source

Internal (Microsoft)

Robbie-Microsoft:
Client assertion can currently be provided by a developer as a string. This PR allows a developer to provide an async callback (which will resolve to a string) as the client assertion.

The client assertion initialization has been removed from ConfidentialClient's constructor and is now initialized inside of buildOauthClientConfiguration, called during every acquireToken call.

Note: Applied Boy-Scout-Rule on UsernamePasswordClient.spec.ts. Only the last test in the file is relevant for this review. Now this test file is up-to-date and mirrors ClientCredentialsClient.spec.ts and OnBehalfOfClient.spec.ts.


lib/msal-common/src/account/ClientCredentials.ts
```diff
@@ -3,11 +3,14 @@
...
  */
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type ClientAssertionCallbackFunction = (...args: any[]) => string;
+
 /**
  * Client Assertion credential for Confidential Clients
  */
 export type ClientAssertion = {
-    assertion: string;
```
bgavrilMS, line 10: breaking change. You have to support both

lib/msal-common/src/account/ClientCredentials.ts
```diff
@@ -3,11 +3,14 @@
  * Licensed under the MIT License.
  */
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type ClientAssertionCallbackFunction = (...args: any[]) => string;
```
bgavrilMS: 1. Why `..args: any[]` ? Are you trying to keep it generic? If you are trying to be generic, how about using a typed object instead, smth like `(assertionRequestOptions: AssertionRequestOptions) => string` (the linter also suggests to avoid `any`)
2. callback has to be async

lib/msal-common/src/account/ClientCredentials.ts
Robbie-Microsoft: ah. The GitHub issue says "instead of".

lib/msal-common/src/account/ClientCredentials.ts
Robbie-Microsoft: IDK what params will be passed into the callback function, so I used `any`

Robbie-Microsoft: I'm not sure we can allow async callbacks. In the changes in this PR, inside of`lib/msal-node/src/client/ConfidentialClientApplication.ts`, the `clientAssertion` callback function is executed in a function that's called from the constructor. There's no way to make this callback function async without changing the instantiation of `ConfidentialClient` to `new await ConfidentialClient()`. This will be a breaking change for all developers.

Please see [this stack overflow post](https://stackoverflow.com/questions/43431550/async-await-class-constructor) to get a better idea of what I'm talking about.

bgavrilMS: > IDK what params will be passed into the callback function, so I used `any`

I don't think this is in the spirit of Typescript. It breaks the promise of "typed languages". 

If there are no input parameters, then use `() => string`. If there are input parameters, then use `(a: SomeInputObjectThatIsTyped)=>string`. 

I recommend the second option to be consistent with MSAL.NET. See https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/AppConfig/AssertionRequestOptions.cs#L14

You will use 2 properties on the input param: client id and token endpoint.

bgavrilMS: For the async, I don't understand the issue with async and ctors. 

All you have to do is to use a promise.

`(a: AssertionRequestOptions)=>Promise<string>`

bgavrilMS: You can have a look at the IAppTokenProvider extensiblity point I defined a while back https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/cd48cb747404831026ba7af6bc22cf34ce2cafed/lib/msal-node/test/client/ClientCredentialClient.spec.ts#L915

I am not sure an interface is needed.

lib/msal-node/docs/initialize-confidential-client-application.md
```diff
@@ -38,7 +39,7 @@ const clientConfig = {
             thumbprint: process.env.thumbprint,
             privateKey: process.env.privateKey,
         }, // OR
-        clientAssertion: "assertion",
+        clientAssertion: assertionCallback,
```
bgavrilMS: As stated before, both of these must be supported. Ideally you can reuse the "clientAssertion" key, but if not possible, it's acceptable to have an additional "clientAssertionCallback: my_callback"

lib/msal-common/src/account/ClientCredentials.ts
KarishmaGhiya: sorry, yeah supporting both can be an option too :)

KarishmaGhiya: yeah so what about a constructor overload?

Robbie-Microsoft: [I don't think TypeScript supports that](https://github.com/microsoft/TypeScript/issues/54397). Do you have evidence otherwise?

Robbie-Microsoft: Anyways, I'd like Bogdan to leave his feedback on the implementation I just finished up today before I look into that.

lib/msal-common/src/account/ClientCredentials.ts
```diff
@@ -3,11 +3,20 @@
  * Licensed under the MIT License.
  */
 
+export type ClientAssertionConfig = {
+    clientId: string;
+    tokenEndpoint?: string;
+};
+
+export type ClientAssertionCallbackFunction = (
```
bgavrilMS: Maybe call it just `ClientAssertionCallback`

lib/msal-common/test/request/RequestParameterBuilder.spec.ts
```diff
@@ -407,14 +416,92 @@ describe("RequestParameterBuilder unit tests", () => {
...
+            requestQueryString.includes(
+                AADServerParamKeys.CLIENT_ASSERTION_TYPE
+            )
+        ).toBe(false);
+    });
+
+    it("adds clientAssertion (ClientAssertionCallbackFunction) and assertionType if they are provided by the developer", async () => {
+        const clientAssertionCallbackFunction: ClientAssertionCallbackFunction =
+            (_config: ClientAssertionConfig) => {
+                return Promise.resolve("testAssertion");
```
bgavrilMS: You can assert that _config is populated with clientId and tokenEndpoint here.

lib/msal-node/test/client/ConfidentialClientApplication.spec.ts
```diff
@@ -203,7 +203,7 @@ describe("ConfidentialClientApplication", () => {
             auth: {
                 clientId: TEST_CONSTANTS.CLIENT_ID,
                 authority: TEST_CONSTANTS.AUTHORITY,
-                clientAssertion: "testAssertion",
+                clientAssertion: () => "testAssertion",
```
bgavrilMS: Why did this have to change? Returning a string should still work?

lib/msal-node/test/client/ClientAssertion.spec.ts
```diff
@@ -8,15 +8,15 @@ const jsonwebtoken = require("jsonwebtoken");
...
 describe("Client assertion test", () => {
     const cryptoProvider = new CryptoProvider();
     const issuer = "client_id";
     const audience = "audience";
 
     test("creates ClientAssertion From assertion", () => {
-        const assertion = ClientAssertion.fromAssertion(
-            TEST_CONSTANTS.CLIENT_ASSERTION
-        );
+        const assertion = ClientAssertion.fromAssertion(assertionCallback());
```
bgavrilMS: This change is  no longer needed ?

lib/msal-common/src/account/ClientCredentials.ts
bgavrilMS: Not a JS / TS expert, I think this is ok. We welcome your feedback @KarishmaGhiya

lib/msal-node/api/msal-node.api.md
```diff
@@ -3,57 +3,61 @@
 > Do not edit this file. It is a report generated by [API Extractor](https://api-extractor.com/).
```
bgavrilMS: Don't edit this file?

lib/msal-node/docs/initialize-confidential-client-application.md
```diff
@@ -53,7 +63,7 @@ const pca = new msal.ConfidentialClientApplication(clientConfig);
 -   A Client credential is mandatory for confidential clients. Client credential can be a:
     -   `clientSecret` is secret string generated set on the app registration.
     -   `clientCertificate` is a certificate set on the app registration. The `thumbprint` is a X.509 SHA-1 thumbprint of the certificate, and the `privateKey` is the PEM encoded private key. `x5c` is the optional X.509 certificate chain used in [subject name/issuer auth scenarios](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/docs/sni.md).
-    -   `clientAssertion` is string that the application uses when requesting a token. The certificate used to sign the assertion should be set on the app registration. Assertion should be of type urn:ietf:params:oauth:client-assertion-type:jwt-bearer.
+    -   `clientAssertion` is a ClientAssertion object containing an assertion string or a callback function that returns an assertion string that the application uses when requesting a token, as well as the assertion's type (urn:ietf:params:oauth:client-assertion-type:jwt-bearer). The certificate used to sign the assertion should be set on the app registration.
```
bgavrilMS:
```suggestion
    -   `clientAssertion` is a ClientAssertion object containing an assertion string or a callback function that returns an assertion string that the application uses when requesting a token, as well as the assertion's type (urn:ietf:params:oauth:client-assertion-type:jwt-bearer). The callback is invoked every time MSAL needs to acquire a token from the token issuer. App developers should generally the callback because assertions expire and new assertions need to be created. App developers are responsabile for the assertion lifetime. Use this mechanism to get tokens for a downstream API using a Federated Identity Credential - see https://learn.microsoft.com/entra/workload-id/workload-identity-federation-create-trust
```

bgavrilMS (changes requested):
Insufficient tests

lib/msal-common/test/request/RequestParameterBuilder.spec.ts
Robbie-Microsoft: I'll be updating the unit tests shortly. I wanted to see if you approved of the proof-of-concept in this PR.

lib/msal-node/test/client/ConfidentialClientApplication.spec.ts
Robbie-Microsoft: I'll be updating the unit tests shortly. I wanted to see if you approved of the proof-of-concept in this PR.

lib/msal-node/test/client/ClientAssertion.spec.ts
Robbie-Microsoft: I'll be updating the unit tests shortly. I wanted to see if you approved of the proof-of-concept in this PR.

Robbie-Microsoft:
> Insufficient tests

I'll be updating the unit tests shortly. I wanted to see if you approved of the proof-of-concept in this PR.

bgavrilMS (approved):
approved with comments

lib/msal-node/test/client/ClientAssertion.spec.ts
```diff
@@ -3,6 +3,7 @@ import { TEST_CONSTANTS } from "../utils/TestConstants";
 import { CryptoProvider } from "../../src/crypto/CryptoProvider";
```
bgavrilMS, line 6: Can you please add tests for AcquireTokenSilent and AcquireTokenByAuthorization code for this? These are both part of confidential client.

bgavrilMS (approved):

lib/msal-node/test/client/ClientAssertion.spec.ts
Robbie-Microsoft: I think you're looking at the wrong file. The test for AcquireTokenByAuthorization already exists in ConfidentialClientApplication.spec.ts. However, a AcquireTokenSilent test is missing from that file. I will add one.

lib/msal-node/src/request/ClientCredentialRequest.ts
```diff
@@ -3,15 +3,18 @@
...
+} from "@azure/msal-common";
 
 /**
  * CommonClientCredentialRequest
  * - scopes                  - Array of scopes the application is requesting access to. Typically contains only the .default scope for a single resource. See: https://learn.microsoft.com/azure/active-directory/develop/scopes-oidc#the-default-scope
  * - authority               - URL of the authority, the security token service (STS) from which MSAL will acquire tokens.
  * - correlationId           - Unique GUID set per request to trace a request end-to-end for telemetry purposes.
  * - skipCache               - Skip token cache lookup and force request to authority to get a a new token. Defaults to false.
- * - clientAssertion         - A Base64Url-encoded signed JWT assertion string used in the Client Credential flow
+ * - clientAssertion         - A callback function that returns a Base64Url-encoded signed JWT assertion string used in the Client Credential flow
```
neha-bhargava:
```suggestion
 * - clientAssertionCallback         - A callback function that returns a Base64Url-encoded signed JWT assertion string used in the Client Credential flow
```

neha-bhargava: either the name of the variable is clientAssertionCallback or the description should be string or callback function?

neha-bhargava (approved):

lib/msal-node/src/request/ClientCredentialRequest.ts
Robbie-Microsoft: the latter

PR #7021, opened 2024-04-10
Capture and instrument cache errors

konstantin-msft:
- Add CacheError.
- Capture and throw cache errors as CacheError.
- Instrument the number of tokens in the cache if CacheError occurs.

**Note**: Cache errors, like "cache quota exceeded", are not captured and reported as empty errors at the moment.


tnorling (approved):

hectormmg (approved):

sameerag (commented):
Thanks for getting this in, like the coverage. I assume this will also mean we can get the percentage of these errors in dashboards too?

konstantin-msft:
> Thanks for getting this in, like the coverage. I assume this will also mean we can get the percentage of these errors in dashboards too?

Yes, we will be able to pivot on cache errors too,

PR #7030, opened 2024-04-15
Instrument scenario id for tracking custom user prompts

konstantin-msft:
- Instrument scenario id for tracking custom user prompts.


lib/msal-common/src/request/RequestParameterBuilder.ts
```diff
@@ -471,6 +476,12 @@ export class RequestParameterBuilder {
         );
         Object.keys(sanitizedEQParams).forEach((key) => {
             this.parameters.set(key, eQParams[key]);
+            if (key === "scenarioId" && this.correlationId) {
```
tnorling, line 479: Can we make `scenarioId` a first class param if we're going to take dependency on it? 2 reasons:
1. EQP should be opaque
2. Allows us to define type

PR #7036, opened 2024-04-16
Instrument server error number

konstantin-msft:
- Instrument server error number.

### DIFF
--- a/lib/msal-common/src/response/ResponseHandler.ts
+++ b/lib/msal-common/src/response/ResponseHandler.ts
@@ -56,7 +56,7 @@ import * as CacheHelpers from "../cache/utils/CacheHelpers";
 function parseServerErrorNo(
     serverResponse: ServerAuthorizationCodeResponse
 ): string | undefined {
-    const errorCodePrefix = "error_code=";
+    const errorCodePrefix = "code=";
     const errorCodePrefixIndex =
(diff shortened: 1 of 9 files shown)


hectormmg (approved):

lib/msal-common/src/response/ResponseHandler.ts
```diff
@@ -53,6 +53,19 @@ import {
...
 
+function parseServerErrorNo(
+    serverResponse: ServerAuthorizationCodeResponse
+): string | undefined {
+    const errorCodePrefix = "code=";
+    const errorCodePrefixIndex =
+        serverResponse.error_uri?.lastIndexOf(errorCodePrefix);
+    return errorCodePrefixIndex && errorCodePrefixIndex >= 0
+        ? serverResponse.error_uri?.substring(
+              errorCodePrefixIndex + errorCodePrefix.length
```
tnorling, line 64: Are we sure it will always be the last item? Would it be more robust to take all numbers between "AADSTS" and ":" in the message?

tnorling (approved):

lib/msal-common/src/response/ResponseHandler.ts
konstantin-msft: `error_uri` looks like a better option in terms of performance. It is not mutable AFAIK.

PR #7039, opened 2024-04-18
Added initialize() to isAuthenticated() function to correctly handle redirects correctly

exhaler:
After logging in, if we input into the address bar `http://localhost:3000/profile` this would fail authentication and redirect to `/failed`. The reason is because `isAuthenticated()` does not have `.initialize()` and throws an error

> {
>     "errorCode": "uninitialized_public_client_application",
>     "errorMessage": "You must call and await the initialize function before attempting to call any other MSAL API.  For more visit: aka.ms/msaljs/browser-errors",
>     "subError": "",
>     "name": "BrowserAuthError"
> }

### DIFF
--- a/samples/msal-browser-samples/vue3-sample-app/src/router/Guard.ts
+++ b/samples/msal-browser-samples/vue3-sample-app/src/router/Guard.ts
@@ -17,7 +17,10 @@ export function registerGuard(router: Router) {
     });
 }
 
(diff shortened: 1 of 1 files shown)


tnorling:
This should not be needed here. If you're getting an uninitialized error it means you haven't verified inProgress is 'none' before invoking your login/acquireToken APIs.

PR #7049, opened 2024-04-22
Instrument account type

konstantin-msft:
- Instrument account type
- Fix acquireTokenByCode native flow instrumentation.

**NOTE** Account type is instrumented before (if exists) and after network/cache call to make it defined for failed requests when no response is returned.

### DIFF
--- a/lib/msal-browser/src/controllers/StandardController.ts
+++ b/lib/msal-browser/src/controllers/StandardController.ts
@@ -84,6 +84,23 @@ import { ClearCacheRequest } from "../request/ClearCacheRequest";
 import { createNewGuid } from "../crypto/BrowserCrypto";
 import { initializeSilentRequest } from "../request/RequestHelpers";
 
+function getAccountType(
+    account?: AccountInfo
+): "AAD" | "MSA" | "B2C" | undefined {
+    const idTokenClaims = account?.idTokenClaims;
+    if (idTokenClaims?.tid) {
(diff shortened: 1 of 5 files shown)


lib/msal-browser/src/controllers/StandardController.ts
```diff
@@ -84,6 +84,22 @@ import { ClearCacheRequest } from "../request/ClearCacheRequest";
 import { createNewGuid } from "../crypto/BrowserCrypto";
 import { initializeSilentRequest } from "../request/RequestHelpers";
 
+function getAccountType(
+    account?: AccountInfo
+): "AAD" | "MSA" | "B2C" | undefined {
+    const idTokenClaims = account?.idTokenClaims;
+    if (idTokenClaims?.tfp || idTokenClaims?.acr) {
+        return "B2C";
```
tnorling, line 92: Do we ever expect this condition to be hit when collecting telemetry? AFAIK no 1P apps are using B2C

tnorling (approved):

lib/msal-browser/src/controllers/StandardController.ts
konstantin-msft: We do not expect this condition to be hit for 1P apps. However, i would prefer to handle this scenario for 3P apps injecting custom performance clients.

hectormmg (approved):

PR #7056, opened 2024-04-24
Fix empty telemetry error code for the redirect flow when server response is empty

konstantin-msft:
- Fix empty telemetry error code for the redirect flow when server response is empty.

**NOTE** Empty error code is caused by parent measurement empty error code overriding child one.

### DIFF
--- a/lib/msal-browser/src/controllers/StandardController.ts
+++ b/lib/msal-browser/src/controllers/StandardController.ts
@@ -455,11 +455,7 @@ export class StandardController implements IController {
                 this.logger,
                 this.performanceClient,
                 rootMeasurement.event.correlationId
-            )(
-                hash,
-                this.performanceClient,
-                rootMeasurement.event.correlationId
-            );
+            )(hash);
         }
 
         return redirectResponse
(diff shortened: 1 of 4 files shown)


lib/msal-browser/src/controllers/StandardController.ts
```diff
@@ -497,7 +493,13 @@ export class StandardController implements IController {
                     EventType.HANDLE_REDIRECT_END,
                     InteractionType.Redirect
                 );
-                rootMeasurement.end({ success: false });
+
+                rootMeasurement.end(
+                    {
+                        success: false,
+                    },
+                    new AuthError("no_server_response")
```
tnorling, line 501: I'm a bit confused by this. No response is not an error. Also in the if above we're setting success to true, but it looks like this will overwrite and success will always be false no matter what?

lib/msal-browser/src/interaction_client/RedirectClient.ts
```diff
@@ -220,12 +216,6 @@ export class RedirectClient extends StandardInteractionClient {
                 this.browserStorage.cleanRequestByInteractionType(
                     InteractionType.Redirect
                 );
-                if (performanceClient && correlationId) {
-                    performanceClient?.addFields(
-                        { errorCode: "no_server_response" },
```
tnorling, line 225: This was actually the correct place to put this because here we know we were expecting a response but didn't get one.

lalimasharda (approved):

Issue #6918, opened 2024-02-22
`useIsAuthenticated` hook potentially may return `false` even if user is actually authenticated


bohdanzhylavskyi:
### Core Library

MSAL.js (@azure/msal-browser)

### Wrapper Library

MSAL React (@azure/msal-react)

### Public or Confidential Client?

Public

### Description

Here is the current implementation of the `useIsAuthenticated` hook (https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/msal-react-v2.0.12/lib/msal-react/src/hooks/useIsAuthenticated.ts#L32):

```
export function useIsAuthenticated(matchAccount?: AccountIdentifiers): boolean {
    const { accounts: allAccounts, inProgress } = useMsal();

    const [hasAuthenticated, setHasAuthenticated] = useState<boolean>(() => {
        if (inProgress === InteractionStatus.Startup) {
            return false;
        }
        return isAuthenticated(allAccounts, matchAccount);
    });

    useEffect(() => {
        setHasAuthenticated(isAuthenticated(allAccounts, matchAccount));
    }, [allAccounts, matchAccount]);

    return hasAuthenticated;
}
```

If you look carefully at the implementation, you may notice that there is a chance that on subsequent renders (all renders except the initial) that hook may return `false` even if user is authenticated and there are actually presented accounts in msal instance (`isAuthenticated(allAccounts, matchAccount)  === true`). It happens because on each render we return the value (`hasAuthenticated`) calculated at the previous render (to be precise calculated right after the previous render, in `useEffect` hook).

That behavior may cause problems, for example, lately, I had a scenario where `UnauthenticatedTemplate` component (which uses `useIsAuthenticated` under the hood) was rendering its content (`children` props) even when the user was actually authenticated (so the account was presented in msal instance).

Below you can see pseudo-code to demonstrate that scenario:
```
<MsalProvider instance={msal}>
    <AuthenticatedTemplate>
        <App />
    </AuthenticatedTemplate>
    <UnauthenticatedTemplate>
        <DoLoginRedirect msal={msal} /> // that component will just call msal.redirectLogin() on initial render
    </UnauthenticatedTemplate>
</MsalProvider>
```

In my case, `DoLoginRedirect` was rendered for a short time after completing authentication (after redirecting from authentication server to be precise) and as a result, one more redundant login redirect step/loop happened.
[...]

PR #7057, opened 2024-04-25
Fix `useIsAuthenticated` hook briefly returning false even though the user is authenticated

hatched-kade:
Addresses https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/6918

The current `useIsAuthenticated` has an unnecessary `useEffect`, which causes it to briefly return `false` in some cases even when an account is present. See https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state and https://react.dev/learn/you-might-not-need-an-effect#caching-expensive-calculations 

This PR switches to use the `useMemo` recommendation

I added a unit test that passes now, but fails with the old implementation:
```
 FAIL  test/hooks/useIsAuthenticated.spec.tsx
  ● withMsal tests › useAuthenticated always returns true if user has an account

    expect(jest.fn()).toHaveBeenCalledTimes(expected)

    Expected number of calls: 0
    Received number of calls: 1

      78 |         expect(await screen.findByText("Has accounts")).toBeInTheDocument();
      79 |         expect(await screen.findByText("Is authed")).toBeInTheDocument();
    > 80 |         expect(invalidAuthStateCallback).toHaveBeenCalledTimes(0);
         |                                          ^
      81 |     });
      82 | });
      83 |

      at Object.<anonymous> (test/hooks/useIsAuthenticated.spec.tsx:80:42)
```

I also had to tweak some act / await / async code in other tests, presumably because now it takes fewer renders for the hook to start returning the correct value. Without the tweaks a couple tests were failing, and others printed warning about updating state outside of an `act(...)`.

### DIFF
--- a/change/@azure-msal-react-5355b5e3-0c1c-4e23-ae9d-a6d96ef4285d.json
+++ b/change/@azure-msal-react-5355b5e3-0c1c-4e23-ae9d-a6d96ef4285d.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Fix useIsAuthenticated returning incorrect value during useEffect update",
+  "comment": "Fix useIsAuthenticated returning incorrect value during useEffect update #7057",
   "packageName": "@azure/msal-react",
   "email": "kade@hatchedlabs.com",
   "dependentChangeType": "patch"
--- a/change/@azure-msal-react-5355b5e3-0c1c-4e23-ae9d-a6d96ef4285d.json
+++ b/change/@azure-msal-react-5355b5e3-0c1c-4e23-ae9d-a6d96ef4285d.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix useIsAuthenticated returning incorrect value during useEffect update",
+  "packageName": "@azure/msal-react",
+  "email": "kade@hatchedlabs.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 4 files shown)


hatched-kade:
@microsoft-github-policy-service agree company="Giant Eagle"

lib/msal-react/test/hooks/useIsAuthenticated.spec.tsx
```diff
@@ -0,0 +1,82 @@
+import React from "react";
+import { render, screen, waitFor, act } from "@testing-library/react";
+import "@testing-library/jest-dom";
+import { Configuration, PublicClientApplication } from "@azure/msal-browser";
+import { TEST_CONFIG, testAccount } from "../TestConstants";
+import { MsalProvider, useIsAuthenticated, withMsal } from "../../src/index";
+
+describe("withMsal tests", () => {
```
tnorling:
```suggestion
describe("useIsAuthenticated tests", () => {
```

tnorling (approved):
Thanks for the PR! I imagine there's other places we can make similar changes.

change/@azure-msal-react-5355b5e3-0c1c-4e23-ae9d-a6d96ef4285d.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix useIsAuthenticated returning incorrect value during useEffect update",
```
jo-arroyo:
```suggestion
  "comment": "Fix useIsAuthenticated returning incorrect value during useEffect update #7057",
```

jo-arroyo (approved):
Looks great, small nit

aleks-sch:
thanks @hatched-kade!

@tnorling @jo-arroyo could we (you) do a release of msal-react please?

we are also impacted by #6918 and would rather not install from dev branch. thanks

PR #7058, opened 2024-04-25
Discard empty redirect telemetry events with no error codes

konstantin-msft:
- Discard empty redirect telemetry events with no error codes. 
- Replace performanceClient with parentMeasurement param to track child errors without throwing.

### DIFF
--- a/lib/msal-browser/src/controllers/StandardController.ts
+++ b/lib/msal-browser/src/controllers/StandardController.ts
@@ -494,9 +494,17 @@ export class StandardController implements IController {
                     InteractionType.Redirect
                 );
 
+                /*
+                 * Instrument an event only if an error code is set. Otherwise, discard it when the redirect response
+                 * is empty and the error code is missing.
+                 */
                 if (rootMeasurement.event.errorCode) {
                     rootMeasurement.end({ success: false });
                 } else {
+                    /*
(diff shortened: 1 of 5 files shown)


lib/msal-browser/src/controllers/StandardController.ts
```diff
@@ -497,7 +493,12 @@ export class StandardController implements IController {
                     EventType.HANDLE_REDIRECT_END,
                     InteractionType.Redirect
                 );
-                rootMeasurement.end({ success: false });
+
+                if (rootMeasurement.event.errorCode) {
+                    rootMeasurement.end({ success: false });
+                } else {
+                    rootMeasurement.discard();
```
tnorling: Will this not discard all successful measurements as well? Or once a measurement has been ended it can no longer be discarded?

Can you also please add a comment describing the what & why we're doing here?

konstantin-msft: Measurements are grouped by a correlation id. We are discarding a single measurement here including all sub-measurements.

konstantin-msft: Added a comment.

tnorling (approved):

jo-arroyo (approved):

PR #7059, opened 2024-04-26
Removed Managed Identity Resource URI Validation

Robbie-Microsoft:
Removed Managed Identity Resource URI Validation. URI's can start with strings other than "https". "api://", for example.

After a discussion with Bogdan, we decided it would be best to remove this validation entirely.

### DIFF
--- a/lib/msal-node/src/client/ManagedIdentityApplication.ts
+++ b/lib/msal-node/src/client/ManagedIdentityApplication.ts
@@ -122,17 +122,6 @@ export class ManagedIdentityApplication {
     public async acquireToken(
         managedIdentityRequestParams: ManagedIdentityRequestParams
     ): Promise<AuthenticationResult> {
-        const resourceUrlString = new UrlString(
-            managedIdentityRequestParams.resource.replace("/.default", "")
-        );
-        try {
-            resourceUrlString.validateAsUri();
-        } catch (e) {
-            throw createManagedIdentityError(
-                ManagedIdentityErrorCodes.invalidResource
-            );
-        }
-
         const managedIdentityRequest: ManagedIdentityRequest = {
             forceRefresh: managedIdentityRequestParams.forceRefresh,
             resource: managedIdentityRequestParams.resource.replace(
--- a/lib/msal-node/src/client/ManagedIdentityApplication.ts
+++ b/lib/msal-node/src/client/ManagedIdentityApplication.ts
@@ -16,7 +16,8 @@ import {
     ProtocolMode,
     StaticAuthorityOptions,
     AuthenticationResult,
-    UrlString,
+    createClientConfigurationError,
+    ClientConfigurationErrorCodes,
 } from "@azure/msal-common";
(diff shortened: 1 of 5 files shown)


lib/msal-node/src/client/ManagedIdentityApplication.ts
```diff
@@ -122,17 +122,6 @@ export class ManagedIdentityApplication {
     public async acquireToken(
         managedIdentityRequestParams: ManagedIdentityRequestParams
     ): Promise<AuthenticationResult> {
-        const resourceUrlString = new UrlString(
```
neha-bhargava, line 125: You still need to check and remove the /.default suffix. MSI endpoints don't accept this suffix. But you don't have to validate the resource as a URI. It can be a string.

Robbie-Microsoft: I am. It happens on line 132, directly below this.

lib/msal-node/test/client/ManagedIdentitySources/Imds.spec.ts
```diff
@@ -779,33 +779,6 @@ describe("Acquires a token successfully via an IMDS Managed Identity", () => {
...
-                })
-            ).rejects.toMatchObject(
-                createManagedIdentityError(
-                    ManagedIdentityErrorCodes.invalidResource
-                )
-            );
-
-            await expect(
-                systemAssignedManagedIdentityApplication.acquireToken({
-                    resource: "",
```
neha-bhargava, line 800: You should still throw if an empty string is passed as a resource.

lib/msal-node/test/client/ManagedIdentitySources/Imds.spec.ts
```diff
@@ -822,7 +812,7 @@ describe("Acquires a token successfully via an IMDS Managed Identity", () => {
 
             expect(() => {
                 new ManagedIdentityApplication(badUserAssignedClientIdConfig);
-            }).toThrowError(
+            }).toThrow(
```
Robbie-Microsoft, line 815: toThrowError is deprecated

neha-bhargava (approved):

lib/msal-node/src/client/ManagedIdentityApplication.ts
```diff
@@ -16,7 +16,8 @@ import {
     ProtocolMode,
     StaticAuthorityOptions,
     AuthenticationResult,
-    UrlString,
+    createClientConfigurationError,
```
pmaytak, line 19: nit: should this be capitalized?

pmaytak (approved):

lib/msal-node/src/client/ManagedIdentityApplication.ts
sameerag: This is inline with all other error naming, and this is probably a `function` exported.

sameerag (approved):

PR #7064, opened 2024-04-27
Add additional logging for NAA initialization

codexeon:
Add additional logging for NAA initialization to help diagnose if NestAppAuthController is being used for the session, or if it is falling back to StandardController. Also log exception that occurred in trying to use Nested App Auth bridge to get further details on failure cases. The change also avoids the time spent trying to initialize the bridge if caller does not opt-in to supportsNestedAppAuth.


sameerag (approved):

baljesingh (approved):
Looks good to me.

PR #7069, opened 2024-04-30
Do not register duplicate performance callbacks

konstantin-msft:
- Do not register duplicate performance callbacks

### DIFF
--- a/lib/msal-common/src/telemetry/performance/PerformanceClient.ts
+++ b/lib/msal-common/src/telemetry/performance/PerformanceClient.ts
@@ -818,6 +818,15 @@ export abstract class PerformanceClient implements IPerformanceClient {
      * @returns {string}
      */
     addPerformanceCallback(callback: PerformanceCallbackFunction): string {
+        for (const [id, cb] of this.callbacks) {
+            if (cb.toString() === callback.toString()) {
+                this.logger.warning(
+                    `PerformanceClient: Performance callback is already registered with id: ${id}`
+                );
+                return id;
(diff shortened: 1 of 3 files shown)


jo-arroyo (approved):

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
```diff
@@ -818,6 +818,15 @@ export abstract class PerformanceClient implements IPerformanceClient {
      * @returns {string}
      */
     addPerformanceCallback(callback: PerformanceCallbackFunction): string {
+        for (const [id, cb] of this.callbacks) {
+            if (cb.toString() === callback.toString()) {
```
tnorling, line 822: This could potentially be very expensive depending on how many callbacks are registered / how many times we attempt to add a duplicate callback. We don't necessarily need to proactively optimize this as I do expect usage to be fairly low but it might be worthwhile to at least measure and track the cost

tnorling (approved):
Approving but let's consider measuring the cost of this comparison to help inform future decisions around optimizing or alternative approaches

lib/msal-common/src/telemetry/performance/PerformanceClient.ts
konstantin-msft: It's always a single callback based on my experience. We can definitely maintain a separate set of callbacks if the number grows to reduce complexity from O(N) to O(1). I think we are fine with O(N) for now.

PR #7072, opened 2024-05-01
Supports caching in NAA apps

sameerag:
NAA apps currently always request the bridge for tokens, adding load to the host caching hence blocking the deployment on scale. This PR adds support for:

* Caching NAA app tokens (idToken and accessToken) in the naa app storage (based on `cacheConfig`)
* Allows for NAA app to set `initContext` with `accountContext` to look for specific account in cache when available.

MetaOS apps create and delete the NAA apps independent of the session. Hence, we always assume the Bridge will add the account context and only look for those tokens. If the account is not found, we always go for the hub to fetch tokens.

**Minor additions:**
* AccountManager created to not duplicate account lookup code and make them functions (saving size).
* If a token is expired while cache lookup, it is cleaned.

**Note and TBD:** 
Since there is no `logout` and naa app life cycle does not map to session life time, cache clean up is not comprehensive. However, this PR clears the cache when a token is identified as expired. The logout equivalent cache clearance (if the parent logs out, nested app should clean up cache) will be handled as a separate use case.


lib/msal-browser/src/naa/BridgeProxy.ts
```diff
@@ -33,6 +33,11 @@ export class BridgeProxy implements IBridgeProxy {
     sdkName: string;
     sdkVersion: string;
     capabilities?: BridgeCapabilities;
+    accountContext: {
```
codexeon: accountContext should be optional, to consider the scenario it doesn't exist

lib/msal-browser/src/naa/BridgeProxy.ts
```diff
@@ -33,6 +33,11 @@ export class BridgeProxy implements IBridgeProxy {
...
+        homeAccountId: string;
+        environment: string;
+        tenantId: string;
+    };
```
codexeon: Can this just be AccountContext instead of redefining the type?

lib/msal-browser/src/naa/BridgeProxy.ts
```diff
@@ -134,15 +139,18 @@ export class BridgeProxy implements IBridgeProxy {
...
-    public async getActiveAccount(): Promise<AccountInfo> {
-        const result = await this.sendRequest("GetActiveAccount");
-        return BridgeProxy.validateBridgeResultOrThrow(result.account);
-    }
-
     public getHostCapabilities(): BridgeCapabilities | null {
         return this.capabilities ?? null;
     }
 
+    public getAccountContext(): AccountContext {
```
codexeon: AccountContext | null

lib/msal-browser/src/naa/BridgeProxy.ts
```diff
@@ -134,15 +139,18 @@ export class BridgeProxy implements IBridgeProxy {
...
+        return {
```
codexeon: can this just return this.accountContext ?? null

lib/msal-browser/src/controllers/ControllerFactory.ts
```diff
@@ -23,15 +23,18 @@ export async function createController(
...
 
-    const operatingContexts = [standard.initialize(), teamsApp.initialize()];
+    const operatingContexts = [standard.initialize(), nestedApp.initialize()];
 
     await Promise.all(operatingContexts);
 
-    if (teamsApp.isAvailable()) {
+    if (
+        nestedApp.isAvailable() &&
+        nestedApp.getConfig().auth.supportsNestedAppAuth
```
codexeon: I moved this check (nestedApp.getConfig().auth.supportsNestedAppAuth) to the TeamsAppOperatingContext itself. Can you make sure the refactor keeps that change (and additional logging).

lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -169,7 +222,10 @@ export class NestedAppAuthController implements IController {
                     reqTimestamp
                 );
 
-            this.operatingContext.setActiveAccount(result.account);
+            // cache the tokens in the response
+            await this.hydrateCache(result, request);
+
+            this.browserStorage.setActiveAccount(result.account);
```
codexeon, line 233: not sure we need to automatically set the cache active account after token acquisition. I think we were setting operating context just to ensure next call for the session had a fully populated account.

lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -276,28 +353,193 @@ export class NestedAppAuthController implements IController {
...
 
+    /**
+     * acquires tokens from cache
+     * @param request
+     * @returns
+     */
+    private async acquireTokenFromCache(
+        request: SilentRequest
+    ): Promise<AuthenticationResult | null> {
+        const atsMeasurement = this.performanceClient.startMeasurement(
```
codexeon, line 369: should this measurement be in acquireTokenSilent itself, this function may not just be for acquireTokenSilent?

lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -276,28 +353,193 @@ export class NestedAppAuthController implements IController {
...
+
+    /**
+     *
+     * @param request
+     * @returns
+     */
+    private async acquireTokenFromCacheInternal(
+        request: SilentRequest
+    ): Promise<AuthenticationResult | null> {
+        const cachedAccount = AccountManager.getAccountByHomeId(
```
codexeon: home id may not be correct for guest scenario. Shouldn't this use getAccount({tenantId, homeAccountId, environment})

lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -344,77 +592,116 @@ export class NestedAppAuthController implements IController {
...
-                this.nestedAppAuthAdapter.fromNaaAccountInfo(currentAccount),
-            ];
-        } else {
-            return [];
-        }
+
+    /**
+     * Gets the currently active account
+     */
+    getActiveAccount(): AccountInfo | null {
```
codexeon, line 712: this might not return the right value for embedded scenario. I think it should also try to filter account from accountContext

lib/msal-browser/src/controllers/ControllerFactory.ts
sameerag: Yes, when merging, should be covered. Will double check.

lib/msal-browser/src/controllers/NestedAppAuthController.ts
sameerag: Okay either way. Because we cannot guarantee the NAA app's lifetime, it is needed to check the `activeAccount` matches the `accountContext` always, so removing this does not make a difference, I guess. Will make the changes.

lib/msal-browser/src/controllers/NestedAppAuthController.ts
sameerag: This was temporary, will be `getAccount` in the final version. We are testing a theory for a bug in naa code. Please check in the next iteration.

lib/msal-browser/src/controllers/NestedAppAuthController.ts
sameerag: Yes, will always do the filter.

lib/msal-browser/src/naa/BridgeProxy.ts
sameerag: Addressed

lib/msal-browser/src/naa/BridgeProxy.ts
sameerag: Sorry, missed in the sequence of coding. Addressed.

lib/msal-browser/src/naa/BridgeProxy.ts
sameerag: Done

lib/msal-browser/src/controllers/NestedAppAuthController.ts
sameerag: It always is only for `acquireTokenSilent`, we do not look into cache for other cases. I have pushed this here for the case where we do not want to go to the host to `silently` fetch tokens - which is essentially `ssoSilent` on the host end.

lib/msal-browser/src/controllers/NestedAppAuthController.ts
sameerag: Done

lib/msal-browser/src/controllers/ControllerFactory.ts
sameerag: I have moved it back here, to make sure we delete it later when NAA gets removed from `PCANext`.

lib/msal-browser/src/cache/AccountManager.ts
```diff
@@ -0,0 +1,184 @@
...
+    if (!username) {
+        logger.warning("getAccountByUsername: No username provided");
+        return null;
+    }
+
+    const account = browserStorage.getAccountInfoFilteredBy({
+        username,
+    });
+    if (account) {
+        logger.verbose(
```
konstantin-msft, line 77: I think the logging part should be moved to `browserStorage.getAccountInfoFilteredBy()` that can capture the whole `accountFilter` object.

lib/msal-browser/src/cache/AccountManager.ts
```diff
@@ -0,0 +1,184 @@
...
+    if (!homeAccountId) {
+        logger.warning("getAccountByHomeId: No homeAccountId provided");
+        return null;
+    }
+
+    const account = browserStorage.getAccountInfoFilteredBy({
+        homeAccountId,
+    });
+    if (account) {
+        logger.verbose(
```
konstantin-msft, line 114: Same here. The logging message here is similar with a single change to the input parameter. It would be better to move this to `browserStorage.getAccountInfoFilteredBy()`

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
```diff
@@ -249,4 +265,53 @@ export class NestedAppAuthAdapter {
             return new AuthError("unknown_error", "An unknown error occurred");
         }
     }
+
+    /**
+     *
```
konstantin-msft, line 271: Please add a function description

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
```diff
@@ -249,4 +265,53 @@ export class NestedAppAuthAdapter {
...
+     * @param account
+     * @param idToken
+     * @param accessToken
+     * @param reqTimestamp
+     * @returns
+     */
+    public fromCachedTokens(
```
konstantin-msft: Probably should be named something like `makeAuthResult` given that the function expects not just the tokens.

sameerag: Let me check if this is public interface, else can make the change.

konstantin-msft (approved):

lib/msal-browser/src/cache/AccountManager.ts
sameerag: I don't think we want to log the filter as it could be very lengthy.  The individual params like `username` and `localAccountId` is already shared by the customer for this particular API, so making them Pii makes sense.

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
sameerag: Done

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
sameerag: Done. Mot so great name though - and may be mapping isn't the call. Will tackle finding a new home to this in a later PR.

lib/msal-browser/src/cache/AccountManager.ts
sameerag: Added a comment above.

lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -169,7 +227,10 @@ export class NestedAppAuthController implements IController {
                     reqTimestamp
                 );
 
-            this.operatingContext.setActiveAccount(result.account);
+            // cache the tokens in the response
+            await this.hydrateCache(result, request);
+
+            // this.browserStorage.setActiveAccount(result.account);
```
codexeon: remove comment

lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -276,28 +358,201 @@ export class NestedAppAuthController implements IController {
...
+        const accountContext = this.bridgeProxy.getAccountContext();
+        let currentAccount = null;
+        if (accountContext) {
+            const hubAccount = AccountManager.getAccount(
+                accountContext,
+                this.logger,
+                this.browserStorage
+            );
+            // always prioritize for hub account context, the reqirement of `request.account` will be removed soon
+            currentAccount = hubAccount || request.account;
```
codexeon, line 429: Probably should prefer request.account if they explicitly pass it in the call. Also, if accountContext is null but request.account is populated, currentAccount can be set to request.account

lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -276,28 +358,201 @@ export class NestedAppAuthController implements IController {
...
+
+        this.logger.verbose(
+            "active account found, attempting to acquire token silently"
+        );
+
+        const authRequest: BaseAuthRequest = {
+            ...request,
+            correlationId:
+                request.correlationId || this.browserCrypto.createNewGuid(),
+            authority: request.authority || currentAccount.environment,
```
codexeon, line 448: can environment be passed as authority property? Realm is not included, does it default to value?

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
```diff
@@ -249,4 +265,54 @@ export class NestedAppAuthAdapter {
...
+    /**
+     * Returns an AuthenticationResult from the given cache items
+     *
+     * @param account
+     * @param idToken
+     * @param accessToken
+     * @param reqTimestamp
+     * @returns
+     */
+    public toAuthenticationResultFromCache(
```
codexeon, line 278: the building of authenticationResult could likely share a lot of code with fromNaaTokenResponse. Could you create a common function with AccountInfo/idToken/accessToken/request parameters that both of these public functions can call into?

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
```diff
@@ -249,4 +265,54 @@ export class NestedAppAuthAdapter {
...
+        const authenticationResult: AuthenticationResult = {
+            authority: accessToken.environment || account.environment,
+            uniqueId: account.localAccountId,
+            tenantId: account.tenantId,
+            scopes: scopes.split(" "),
+            account,
+            idToken: idToken.secret,
+            idTokenClaims: idTokenClaims || {},
+            accessToken: accessToken.secret,
+            fromCache: true,
```
codexeon, line 305: I think fromNaaTokenResponse always returns fromCache: true as well. Maybe always consider fromNaaTokenResponse fromCache: false.

lib/msal-browser/src/naa/InitContext.ts
```diff
@@ -4,9 +4,11 @@
  */
 
 import { BridgeCapabilities } from "./BridgeCapabilities";
+import { AccountContext } from "./BridgeAccountContext";
 
 export interface InitContext {
     capabilities?: BridgeCapabilities;
     sdkName: string;
     sdkVersion: string;
+    accountContext: AccountContext;
```
codexeon: Make accountContext optional so we don't forget the null checks

baljesingh (approved):

lib/msal-browser/src/controllers/NestedAppAuthController.ts
sameerag: Done. We are ignoring `getActiveAccount()`, so I am not sure about the value add here.

lib/msal-browser/src/controllers/NestedAppAuthController.ts
sameerag: The sample usage is trouble some. request gets the account from getActiveAccount() - which is active in the context of the embedded account but not necessarily the active account in the host. If we can avoid that, we can revert to `request.account`.

This also can change once we make `account` optional for `acquireTokenSilent` which we plan to soon.

lib/msal-browser/src/controllers/NestedAppAuthController.ts
sameerag: Realm is tenant. Here `authority` is a string from the user, later we do alias mapping and add it as `environment`. Realm is derived from the cloud discovery.

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
sameerag: Yes. Changed it.

lib/msal-browser/src/naa/mapping/NestedAppAuthAdapter.ts
sameerag: I think most of it is on the fly. I do not see optimization in claims, as they are always a util call. If we think we can do more, let me take up a clean up PR later.

lib/msal-browser/src/naa/InitContext.ts
sameerag: Done

lib/msal-browser/src/naa/BridgeProxy.ts
```diff
@@ -210,6 +214,7 @@ export class BridgeProxy implements IBridgeProxy {
         return new BridgeProxy(
             response.sdkName,
             response.sdkVersion,
+            response.accountContext || undefined,
```
codexeon: || undefined should not be necessary, if property doesn't exist it should already be undefined

lib/msal-browser/src/naa/BridgeProxy.ts
```diff
@@ -194,10 +196,12 @@ export class BridgeProxy implements IBridgeProxy {
     private constructor(
         sdkName: string,
         sdkVersion: string,
+        accountContext?: AccountContext | undefined,
         capabilities?: BridgeCapabilities
     ) {
         this.sdkName = sdkName;
         this.sdkVersion = sdkVersion;
+        this.accountContext = accountContext || undefined;
```
codexeon: || undefined should not be necessary, if property doesn't exist it should already be undefined

lib/msal-browser/src/naa/BridgeProxy.ts
```diff
@@ -194,10 +196,12 @@ export class BridgeProxy implements IBridgeProxy {
     private constructor(
         sdkName: string,
         sdkVersion: string,
+        accountContext?: AccountContext | undefined,
```
codexeon: using | undefined here is redundant, it is already an optional parameter

lib/msal-browser/src/controllers/NestedAppAuthController.ts
codexeon: account is optional, but still allowed to be passed. account should be used as the fallback if provided.

lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -344,77 +605,116 @@ export class NestedAppAuthController implements IController {
...
-            ];
-        } else {
-            return [];
-        }
+
+    /**
+     * Gets the currently active account
+     */
+    getActiveAccount(): AccountInfo | null {
+        return AccountManager.getActiveAccount(this.browserStorage);
```
codexeon, line 713: I think this API should consider accountContext, and not directly read from browser storage.

lib/msal-browser/src/controllers/NestedAppAuthController.ts
```diff
@@ -243,7 +322,10 @@ export class NestedAppAuthController implements IController {
                     reqTimestamp
                 );
 
-            this.operatingContext.setActiveAccount(result.account);
+            // cache the tokens in the response
+            await this.hydrateCache(result, request);
+
+            this.browserStorage.setActiveAccount(result.account);
```
codexeon, line 328: for NAA scenario writing active account to localStorage is likely to cause issues. I think I prefered the operating context active account, that is for this session only.

lib/msal-browser/src/error/NestedAppAuthError.ts
```diff
@@ -29,4 +34,11 @@ export class NestedAppAuthError extends AuthError {
             NestedAppAuthErrorMessage.unsupportedMethod.desc
         );
     }
+
+    public static createNoAccountContextError(): NestedAppAuthError {
+        return new NestedAppAuthError(
+            NestedAppAuthErrorMessage.noAccountContext.code,
+            NestedAppAuthErrorMessage.noAccountContext.desc
+        );
+    }
```
codexeon: is this error used anywhere?

codexeon (approved):

lib/msal-browser/src/naa/BridgeProxy.ts
sameerag: Done

lib/msal-browser/src/naa/BridgeProxy.ts
sameerag: Agree, sorry made changes in reverse and my ts fixer auto added.

lib/msal-browser/src/naa/BridgeProxy.ts
sameerag: Done

lib/msal-browser/src/controllers/NestedAppAuthController.ts
sameerag: Should I just completely get rid of this for now and we can land on improvements on default account once hosts test it?

lib/msal-browser/src/error/NestedAppAuthError.ts
sameerag: Good catch. This was before we made `accountContext` optional.

lib/msal-browser/src/controllers/NestedAppAuthController.ts
sameerag: Will do a follow up PR on account management.

PR #7078, opened 2024-05-06
Update regional-authorities.md

bgavrilMS:
@trwalke and @Robbie-Microsoft - please review.

### DIFF
--- a/lib/msal-node/docs/regional-authorities.md
+++ b/lib/msal-node/docs/regional-authorities.md
@@ -1,6 +1,6 @@
 # Enabling regional authorities
 
-> NOTE: This legacy feature is only available for internal Microsoft services and the client credential flow. It is recommended to use [Managed Identity](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/docs/managed-identity.md) instead
+> NOTE: This legacy feature is only available for internal Microsoft services and the client credential flow. It is recommended to use [Managed Identity](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/docs/managed-identity.md) instead.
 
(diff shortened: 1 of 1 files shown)


Robbie-Microsoft (approved):

trwalke (approved):
LGTM

lib/msal-node/docs/regional-authorities.md
```diff
@@ -47,7 +49,7 @@ cca.acquireTokenByClientCredential(clientCredentialRequest)
     });
 ```
 
-> NOTE: If you provide the value `"TryAutoDetect"` in the `azureRegion` field, the msal library which will try to discover the region the application has been deployed to and use that region. If no region is auto discovered the library will fall back to using the global authority.
+> NOTE: If you provide the value `"TryAutoDetect"` in the `azureRegion` field, the msal library which will try to discover the region the application has been deployed to and use that region. This is un-reliable and should be avoided.
```
pmaytak:
```suggestion
> NOTE: If you provide the value `"TryAutoDetect"` in the `azureRegion` field, the msal library which will try to discover the region the application has been deployed to and use that region. This is unreliable and should be avoided.
```

lib/msal-node/docs/regional-authorities.md
```diff
@@ -1,5 +1,7 @@
 # Enabling regional authorities
 
+> NOTE: This legacy feature is only available for internal Microsoft services and the client credential flow. It is recommended to use [Managed Identity](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/docs/managed-identity.md) instead
```
pmaytak:
```suggestion
> NOTE: This legacy feature is only available for internal Microsoft services and the client credential flow. It is recommended to use [Managed Identity](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/docs/managed-identity.md) instead.
```

sameerag (approved):

PR #7081, opened 2024-05-06
Add missing sendTime prop to NAA request

salaman:
`sendTime` prop is present on `BridgeResponseEnvelope` but not used for nested app auth requests. Adds value here which is useful to have on host side.


sameerag:
Is this to manage clock skew?

salaman:
@sameerag I assumed it was for telemetry, @codexeon?

konstantin-msft:
> @sameerag I assumed it was for telemetry, @codexeon?

Telemetry SDK adds event time by default. You don't need to add another date,

salaman:
@konstantin-msft agreed, but this is for our host-side telemetry outside of MATS, which every host is currently implementing as MATS isn't fully fleshed out for NAA brokering flows. This prop has existed on the NAA schema, just unpopulated for some reason.

codexeon (approved):

PR #7085, opened 2024-05-08
Migrate to Lab Vault Cert for E2E Tests

hectormmg:
This PR:
- Adds `3p-e2e.yml` which defines a pipeline to run 3P MSAL JS E2E tests to replace existing GitHub workflows
- Updates `gen_env.sh` to use a new PowerShell script called `gen_env.ps1` that creates and populates a .env file with E2E Test environment variables (Express session secret, ClientID, TenantID and the absolute path of the client certificate used to authenticate with the Lab App)
- Updates `e2e-test-utils/LabClient` to use a `CertificateCredential` instead of `ClientSecretCredential`
- Updates the `b2c-user-flows` Node sample to use a SESSION_SECRET env variable instead of the AZURE_CLIENT_SECRET


.pipelines/3p-e2e.yml
```diff
@@ -0,0 +1,86 @@
+variables:
+    CDP_DEFINITION_BUILD_COUNT: $[counter('', 0)] # needed for onebranch.pipeline.version task https://aka.ms/obpipelines/versioning
+    LinuxContainerImage: "mcr.microsoft.com/onebranch/cbl-mariner/build:2.0" # Docker image which is used to build the project https://aka.ms/obpipelines/containers
+    DEBIAN_FRONTEND: noninteractive
+    ${{ if startsWith(variables['Build.SourceBranch'], 'refs/heads/') }}:
+        sourceBranchName: ${{ replace(variables['Build.SourceBranch'], 'refs/heads/', '') }}
+    ${{ if startsWith(variables['Build.SourceBranch'], 'refs/pull/') }}:
```
konstantin-msft: Should this be `${{ if startsWith(variables['System.PullRequest.SourceBranch'], 'refs/pull/') }}:` instead?

hectormmg: https://learn.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#build-variables-devops-services

They'll have the same value if it's coming from a PR tag

.pipelines/3p-e2e.yml
```diff
@@ -0,0 +1,87 @@
...
+                        poolType: "linux"
+                        publishFlagName: "publishMsalAngular"
+                        sourceBranchName: ${{ variables.sourceBranchName }}
+                        stage: "PR"
+                        override: true
+                        samples:
+                            - "angular15-sample-app"
+                            - "angular16-sample-app"
+                            - "angular-b2c-sample-app"
+                            - "angular-standalone-sample"
```
jo-arroyo, line 87: Were you going to add `angular17-standalone-sample` here too?

bgavrilMS (approved):

konstantin-msft (approved):

.pipelines/3p-e2e.yml
hectormmg: Done

PR #7093, opened 2024-05-10
API for Managed Identity to detect the current environment

### LINKED ISSUE #7083: [Feature Request] Need an API for Managed Identity to detect the current environment
### Core Library

MSAL Node (@azure/msal-node)

### Wrapper Library

Not Applicable

Confidential

### Description

Please refer to the "MSAL implementation" part of [this internal PR](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?version=GBdev&path=/ManagedIdentity/FacilitateImdsProbing.md&_a=preview&anchor=msal-implementations).

### Source

Internal (Microsoft)

Robbie-Microsoft:
This feature is requested by Azure Identity SDK. Please see the description in the linked issue for more info.


lib/msal-node/src/utils/Constants.ts
```diff
@@ -43,6 +43,19 @@ export const ManagedIdentitySourceNames = {
...
     (typeof ManagedIdentitySourceNames)[keyof typeof ManagedIdentitySourceNames];
 
+/**
+ * Azure Identity SDK defined identifiers for the Managed Identity Sources
+ */
+export const AzureIdentitySdkManagedIdentitySourceNames = {
+    APP_SERVICE: "APP_SERVICE",
+    AZURE_ARC: "ARC",
+    CLOUD_SHELL: "CLOUD_SHELL",
+    IMDS: "DEFAULT_TO_VM",
```
rayluo: The caller would mainly use the const's names rather than their values. Better also name the const same `DEFAULT_TO_VM` here. The meaning we want to convey here is that the new helper `GetManagedIdentitySource()` does not have confidence on whether the underlying environment is a VM, therefore the caller probably want to probe.

lib/msal-node/src/client/ManagedIdentityApplication.ts
```diff
@@ -183,4 +186,8 @@ export class ManagedIdentityApplication {
             );
         }
     }
+
+    public getManagedIdentitySource(): AzureIdentitySdkManagedIdentitySourceNames {
```
bgavrilMS: It's a public API so it needs comments. The comments should clearly state that this API is for Azure SDK.

Also, maybe it could live somewhere else? Like some helper that we don't document?

bgavrilMS: Why doesn't this just return a string?

lib/msal-node/src/client/ManagedIdentityClient.ts
```diff
@@ -73,6 +74,40 @@ export class ManagedIdentityClient {
...
+            }
+        );
+    }
+
+    /**
+     * Determine the Managed Identity Source based on available environment variables. This API is consumed by Azure Identity SDK.
+     * @returns AzureIdentitySdkManagedIdentitySourceNames - Azure Identity SDK defined identifiers for the Managed Identity Sources
+     */
+    public getManagedIdentitySource(): AzureIdentitySdkManagedIdentitySourceNames {
+        return this.allEnvironmentVariablesAreDefined(
```
bgavrilMS: Is it not possible to just call the existing `selectManagedIdentitySource` ? and to check the type of the object returned?

lib/msal-node/src/client/ManagedIdentityApplication.ts
Robbie-Microsoft: It does, essentially. The AzureIdentitySdkManagedIdentitySourceNames is a string enum. This is just extra type checking.

lib/msal-node/src/client/ManagedIdentityClient.ts
Robbie-Microsoft: Way back when, that's sort of how I wanted to originally write this class. I wanted to call `selectManagedIdentitySource` on `ManagedIdentityClient.ts`'s initialization. However, you said we shouldn't determine the source type before the acquireToken call.

What you've suggested is possible if you're ok with saving the source type when azure sdk calls `selectManagedIdentitySource`. Otherwise, there's no point to calling `selectManagedIdentitySource`.

In my opinion, my solution is better because it's just null checks, and the source isn't actually being created via each source's `tryCreate`. Plus, I'm able to further modularize each source's `tryCreate` by removing the null checks that are now re-used by `getManagedIdentitySource`.

lib/msal-node/src/client/ManagedIdentityApplication.ts
Robbie-Microsoft: I'll add comments.

However, it needs to live here because only `ManagedIdentityApplication.ts` has a reference to `ManagedIdentityClient.ts`, which is not part of the public api.

lib/msal-node/src/utils/Constants.ts
neha-bhargava: I added an enum to the list instead as renaming the IMDS enum. Renaming will be a breaking change in case anyone already have a check on this. We do send back the source in Error scenarios. See https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4761/files/de6bc4bb5cca106e3bc3c9137908e1c3af45f494#diff-13058fddef4867b4d36dca3120b5b961ae64b039b32bcf5ad410a364b36fbfa8R83

lib/msal-node/src/client/ManagedIdentityClient.ts
neha-bhargava: I did refactor a bit to use the getManagedIdentitySource in the select method. There are more things happening in the selectManagedIdentitySource that does not need to happen in get. See https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4761/files#diff-13058fddef4867b4d36dca3120b5b961ae64b039b32bcf5ad410a364b36fbfa8R43

lib/msal-node/src/utils/Constants.ts
rayluo: > Renaming will be a breaking change in case anyone already have a check on this.

Removing existing names would be a breaking change. But you may always add a new alias for existing name, and then the old name can be considered deprecated.

lib/msal-node/src/client/ManagedIdentityClient.ts
```diff
@@ -73,6 +74,40 @@ export class ManagedIdentityClient {
...
+                return environmentVariable !== undefined;
+            }
+        );
+    }
+
+    /**
+     * Determine the Managed Identity Source based on available environment variables. This API is consumed by Azure Identity SDK.
+     * @returns AzureIdentitySdkManagedIdentitySourceNames - Azure Identity SDK defined identifiers for the Managed Identity Sources
+     */
+    public getManagedIdentitySource(): AzureIdentitySdkManagedIdentitySourceNames {
```
bgavrilMS: Consider caching the value here (in a static variable). It cannot change.

bgavrilMS (approved):

lib/msal-node/src/utils/Constants.ts
Robbie-Microsoft: In regards to the original comment, this string enum is not public api. Only the string values get exposed. 

The keys represent what the sources are called in msal-node, and the values represent what they're called in Azure Identity SDK.

I think the name of the enum is clear enough to convey this, for the developers sake.

Robbie-Microsoft: Implemented Ray's feedback, per offline discussion.

Robbie-Microsoft: @neha-bhargava, I didn't rename any existing constants. I made a new string enum just for the api that Azure Identity SDK consumes.

lib/msal-node/src/client/ManagedIdentityClient.ts
```diff
@@ -73,6 +75,43 @@ export class ManagedIdentityClient {
...
+        return Object.values(environmentVariables).every(
+            (environmentVariable) => {
+                return environmentVariable !== undefined;
+            }
+        );
+    }
+
+    /**
+     * Determine the Managed Identity Source based on available environment variables. This API is consumed by ManagedIdentityApplication's getManagedIdentitySource.
+     * @returns AzureIdentitySdkManagedIdentitySourceNames - Azure Identity SDK defined identifiers for the Managed Identity Sources
```
neha-bhargava:
```suggestion
     * @returns AzureIdentitySdkManagedIdentitySourceNames - Identifying Azure source.
```

neha-bhargava: Or something. These are not defined by Azure Identity

Robbie-Microsoft: Good catch, I forgot to update that description

neha-bhargava (approved):

PR #7103, opened 2024-05-15
Update E2E Test Templates

hectormmg:
This PR:
- Updates npm scripts in samples used for E2E testing to standardize the test:e2e command
- Updates `3p-e2e.yml` and 1`p-e2e.yml` to use the new unified `e2e-tests.yml` template instead of each package having it's own e2e template to reduce duplication and pipeline fix overhead
- Adds stage, branch and repo source logic to correctly handle branch name substitutions for every combination of 1P and 3P builds


.pipelines/1p-e2e.yml
```diff
@@ -16,27 +17,29 @@ resources:
         - repository: 1P
           type: git
           name: IDDP/msal-javascript-1p
-          ref: master
+          ref: update-validation-templates
```
tnorling: Is this intentional?

.pipelines/3p-e2e.yml
```diff
@@ -16,7 +17,7 @@ resources:
         - repository: 1P
           type: git
           name: IDDP/msal-javascript-1p
-          ref: master
+          ref: update-validation-templates
```
tnorling: Same q here

tnorling (approved):

peterzenz (approved):

.pipelines/1p-e2e.yml
hectormmg: Yes, it's temporary because the pipeline definitions aren't on master yet. I'll update before merging (builds will fail for that unfortunately)

PR #7112, opened 2024-05-17
Support CreatePCA pattern for NAA apps

sameerag:
This PR adds two new public APIs:

* `createNestablePublicClientApplication` and `createStandardPublicClientApplication` as exportable functions.
* Moving these functions out of PCA class helps us to eventually clean up and provide a better size output for NAA only apps.

The usage pattern still remains the same:
- Create nestedPCA if the OperatingContext matches
- Fall back to standardPCA as default

`PublicClientNext` will eventually see removal in any NAA support as we make changes. We will add notice that `PCANext` will not be following semver in the upcoming releases, so we can experiment with the future changes using that class once this pattern is adapted.


konstantin-msft (approved):

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -417,3 +415,41 @@ export class PublicClientApplication implements IPublicClientApplication {
...
+        const controller = new NestedAppAuthController(nestedAppAuth);
+        return new PublicClientApplication(configuration, controller);
+    }
+
+    return CreateStandardPublicClientApplication(configuration);
+}
+
+/**
+ * creates NestedAppAuthController and passes it to the PublicClientApplication,
+ * falls back to StandardController if NestedAppAuthController is not available
```
```suggestion
 * creates PublicClientApplication using StandardController
```

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -417,3 +415,41 @@ export class PublicClientApplication implements IPublicClientApplication {
...
+): Promise<IPublicClientApplication> {
+    const nestedAppAuth = new NestedAppOperatingContext(configuration);
+    await nestedAppAuth.initialize();
+
+    if (nestedAppAuth.isAvailable()) {
+        const controller = new NestedAppAuthController(nestedAppAuth);
+        return new PublicClientApplication(configuration, controller);
+    }
+
+    return CreateStandardPublicClientApplication(configuration);
```
```suggestion
    return createStandardPublicClientApplication(configuration);
```

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -417,3 +415,41 @@ export class PublicClientApplication implements IPublicClientApplication {
...
+ */
+export async function createNestablePublicClientApplication(
+    configuration: Configuration
+): Promise<IPublicClientApplication> {
+    const nestedAppAuth = new NestedAppOperatingContext(configuration);
+    await nestedAppAuth.initialize();
+
+    if (nestedAppAuth.isAvailable()) {
+        const controller = new NestedAppAuthController(nestedAppAuth);
+        return new PublicClientApplication(configuration, controller);
```
tnorling, line 435: Do we need to call initialize on this here too before returning?

lib/msal-browser/src/app/PublicClientNext.ts
```diff
@@ -35,6 +35,9 @@ import { UnknownOperatingContext } from "../operatingcontext/UnknownOperatingCon
  * The goals of these changes are to provide a clean separation of behavior between different operating contexts (Nested App Auth, Platform Brokers, Plain old Browser, etc.)
  * while still providing a consistent API surface for developers.
  *
+ * Please use PublicClientApplication for any prod/real-world scenarios.
+ * Note: We will use PublicClientNext for any internal exiperiments and testing, and any breaking changes will be made here first.
```
tnorling:
```suggestion
 * Note: PublicClientNext is experimental and subject to breaking changes without following semver
```

lib/msal-browser/src/app/PublicClientApplication.ts
sameerag: We should. Without initialize the `nestedApp.isAvailable` is not populated.

tnorling: Sorry, I'm asking about pca.initialize, which I don't see here.

lib/msal-browser/docs/create-pca.md
```diff
@@ -0,0 +1,34 @@
+# Creating Client Application instances for MSAL JS
```
tnorling: There's already an[ initialization doc ](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/initialization.md)- can we update that with these changes?

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -417,3 +415,40 @@ export class PublicClientApplication implements IPublicClientApplication {
...
+ * falls back to StandardController if NestedAppAuthController is not available
+ *
+ * @param configuration
+ * @returns IPublicClientApplication
+ *
+ */
+export async function createNestablePublicClientApplication(
+    configuration: Configuration
+): Promise<IPublicClientApplication> {
+    const nestedAppAuth = new NestedAppOperatingContext(configuration);
```
codexeon, line 430: is supportsNestedAppAuth:true still necessary if developer creates pca with createNestablePublicClientApplication, they want to use NAA. I think we could either remove the config option, or default the value based on calling this function.

sameerag: Fine with me. I wanted check this with the hub sdk, if this is a dependency. As of now, NestedAppOperatingContext checks for this config in the `initialize` method. If all hosts/hub agree, we can remove it. All we will look for is the bridge being present.

sameerag (commented):
> Fine with me. I wanted check this with the hub sdk, if this is a dependency. As of now, NestedAppOperatingContext checks for this config in the `initialize` method. If all hosts/hub agree, we can remove it. All we will look for is the bridge being present.

lib/msal-browser/src/app/PublicClientApplication.ts
sameerag: For NAA, in `NestedAppAuthController` the initialize does nothing. It is an empty function added for the sake of consistency with the other controllers. If we see a future need for it, we can add now, but it will be an empty function.  I can check if it affects functionality, but I don't think it does.

lib/msal-browser/src/app/PublicClientApplication.ts
sameerag: Also Is there any scenario where Bridge is enabled but NAA instantiation is not required? If not, we can always check for Bridge only and update the guidance accordingly.

sameerag: Removed the config check and added deprecation notice.

lib/msal-browser/docs/create-pca.md
sameerag: Done.

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -28,6 +28,8 @@ import { AuthenticationResult } from "../response/AuthenticationResult";
 import { EventCallbackFunction } from "../event/EventMessage";
 import { ClearCacheRequest } from "../request/ClearCacheRequest";
 import { EndSessionPopupRequest } from "../request/EndSessionPopupRequest";
+import { NestedAppAuthController } from "../controllers/NestedAppAuthController";
```
salaman, line 31: just confirming -- because these imports are in the main PCA.ts file, they _may_ increase bundle size for folks that don't have tree shaking configured correctly. Since msal-browser specifies sideEffects=false, most cases should be smart enough to shake this out of the bundle if they don't call  `createNestablePublicClientApplication` directly, but just fyi

lib/msal-browser/src/app/PublicClientApplication.ts
```diff
@@ -417,3 +415,40 @@ export class PublicClientApplication implements IPublicClientApplication {
...
+}
+
+/**
+ * creates PublicClientApplication using StandardController
+ *
+ * @param configuration
+ * @returns IPublicClientApplication
+ *
+ */
+export async function createStandardPublicClientApplication(
```
salaman, line 448: why not reuse `createPublicClientApplication`? seems like `createPublicClientApplication` and `createStandardPublicClientApplication` do basically the same thing and end up using StandardController?

lib/msal-browser/src/app/PublicClientApplication.ts
salaman: I would add the init call _just in case_ someone adds some logic to NAAController in the future and we forget to call it here...

lib/msal-browser/src/controllers/ControllerFactory.ts
```diff
@@ -3,38 +3,36 @@
...
-    return controller.StandardController.createController(standard);
+    return StandardController.createController(standard);
 }
 
 export async function createController(
     config: Configuration
 ): Promise<IController | null> {
     const standard = new StandardOperatingContext(config);
-    const teamsApp = new TeamsAppOperatingContext(config);
+    const nestedApp = new NestedAppOperatingContext(config);
```
salaman, line 26: this is just called from PCANext and could be removed when we drop PCANext, right? is it worth just getting rid of the NAA stuff from PCANext right now and slimming this down?

lib/msal-browser/src/app/PublicClientApplication.ts
tnorling: Forward looking. We're thinking long term the plan should be to use these create___ functions and the static method on PCA will be removed. That allows us to provide a way to treeshake out any and all controllers that aren't being used by a given app, which isn't possible today with the constructor & static create method referencing the Standard Controller.

lib/msal-browser/src/app/PublicClientApplication.ts
tnorling: This is acceptable IMO. If you haven't configured treeshaking you likely don't care about size anyway.

lib/msal-browser/src/controllers/ControllerFactory.ts
```diff
@@ -3,38 +3,36 @@
...
 
-    const operatingContexts = [standard.initialize(), teamsApp.initialize()];
+    const operatingContexts = [standard.initialize(), nestedApp.initialize()];
 
     await Promise.all(operatingContexts);
 
-    if (teamsApp.isAvailable()) {
-        const controller = await import("./NestedAppAuthController");
-        return controller.NestedAppAuthController.createController(teamsApp);
+    if (nestedApp.isAvailable()) {
```
codexeon: probably need to add back check to config.supportsNestedAppAuth here. I get this is an experimental class, but since NestedAppOperatingContext no longer checks for this value, it will always think developer opts-in if they use PublicClientNext, and it isn't obvious.

lib/msal-browser/src/app/PublicClientApplication.ts
sameerag: +1 and we will try doing whatever size optimization we can eventually but the focus is to make sure at least tree shaking works as intended and will be improved.

lib/msal-browser/src/app/PublicClientApplication.ts
sameerag: Exactly. I added `createStandard ...` to eventually remove the static reference. I will follow up with cleaning controller factory and other code that is not relevant soon.

lib/msal-browser/src/controllers/ControllerFactory.ts
sameerag: Yes, added comment above. Will follow up with a PR to clean controller creation altogether.

salaman:
any chance you can merge https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/7081 into here while you're changing this? I don't seem to have whatever permissions are required to get a working build.

sameerag:
> any chance you can merge #7081 into here while you're changing this? I don't seem to have whatever permissions are required to get a working build.

Sure. I can.

tnorling (approved):

lib/msal-browser/src/controllers/ControllerFactory.ts
sameerag: The idea is to clean up NAA altogether from `PCANext`. I added the check in `ControllerFactory` itself which we plan to remove soon. Let us make an announcement that `PCANext` will stop supporting NAA once this PR is merged.

sameerag:
> > any chance you can merge #7081 into here while you're changing this? I don't seem to have whatever permissions are required to get a working build.
> 
> Sure. I can.

I pulled these into the cache PR.

salaman (approved):

codexeon (approved):

